PR #1, opened 2016-03-03
Add expression MLangs

bshepherdson:
With decently complete tests against a real DAO.

Plus some small core tweaks to fix some small bugs encountered.

### DIFF
--- a/src/core/stdlib.js
+++ b/src/core/stdlib.js
@@ -289,6 +289,12 @@ foam.LIB({
       if ( ! a || ! b ) return false;
       if ( a.equals ) return a.equals(b);
       return a == b;
+    },
+    function compare(a, b) {
+      if ( a.compareTo ) return a.compareTo(b);
+      if ( b.compareTo ) return - b.compareTo(a);
+      if ( foam.util.equals(a, b) ) return 0;
+      return a.$UID.compareTo(b.$UID);
     }
   ]
 });
--- a/src/core/stdlib.js
+++ b/src/core/stdlib.js
@@ -291,6 +291,10 @@ foam.LIB({
       return a == b;
     },
     function compare(a, b) {
+      if ( typeof a === 'number' && typeof b === 'number' )
+        return a < b ? -1 : a > b ? 1 : 0;
+      if ( typeof a === 'string' && typeof b === 'string' )
+        return a < b ? -1 : a > b ? 1 : 0;
       if ( a.compareTo ) return a.compareTo(b);
       if ( b.compareTo ) return - b.compareTo(a);
       if ( foam.util.equals(a, b) ) return 0;
--- a/src/lib/dao.js
+++ b/src/lib/dao.js
@@ -612,6 +612,8 @@ foam.CLASS({
   methods: [
     function put(o) {
       this.a.push(o);
+    },
+    function eof() {
     }
   ]
 });
@@ -621,8 +623,9 @@ foam.CLASS({
   package: 'foam.dao',
   name: 'ArrayDAO',
   extends: 'foam.dao.AbstractDAO',
-
-  requires: [ 'foam.dao.ArraySink' ],
+  requires: [
+    'foam.dao.ArraySink',
+  ],
 
   properties: [
     {
--- a/src/lib/dao.js
+++ b/src/lib/dao.js
@@ -691,7 +691,7 @@ foam.CLASS({
     function select(sink, options) {
       var resultSink = sink || this.ArraySink.create();
 
-      sink = this.decorateSink_(sink || this.ArraySink.create(), options);
+      sink = this.decorateSink_(resultSink, options);
 
       var promise = this.Promise.create();
 
--- a/src/lib/dao.js
+++ b/src/lib/dao.js
@@ -612,8 +612,6 @@ foam.CLASS({
   methods: [
     function put(o) {
       this.a.push(o);
-    },
-    function eof() {
     }
   ]
 });
--- a/src/lib/mlang.js
+++ b/src/lib/mlang.js
@@ -17,11 +17,12 @@
 
 
 
+// TODO(braden): Port the partialEval() code over here.
 
 foam.CLASS({
   package: 'foam.mlang',
   implements: ['foam.dao.Sink'],
-  name: 'CountExpr',
+  name: 'Count',
   properties: [
     {
       name: 'value',
@@ -31,7 +32,10 @@ foam.CLASS({
   methods: [
     function put() {
       this.value++;
-    }
+    },
+    function toString() {
+      return 'COUNT()';
+    },
   ]
 });
 
@@ -42,15 +46,15 @@ foam.CLASS({
   properties: [
     {
       name: 'of',
-      defaultValue: 'foam.mlang.Expr'
+      defaultValue: 'foam.mlang.predicate.Expr'
     },
     {
       name: 'adaptArrayElement',
       defaultValue: function(o) {
-        if ( typeof o !== "object" ) return foam.mlang.ConstantExpr.create({ value: o });
-        if ( Array.isArray(o) ) return foam.mlang.ConstantExpr.create({ value: o });
-        if ( o === true ) return foam.mlang.TrueExpr.create();
-        if ( o === false ) return foam.mlang.FalseExpr.create();
+        if ( typeof o !== "object" ) return foam.mlang.predicate.Constant.create({ value: o });
+        if ( Array.isArray(o) ) return foam.mlang.predicate.Constant.create({ value: o });
+        if ( o === true ) return foam.mlang.True.create();
+        if ( o === false ) return foam.mlang.False.create();
         return o;
       }
     }
@@ -65,7 +69,8 @@ foam.CLASS({
     {
       name: 'adapt',
       defaultValue: function(_, o) {
-        if ( typeof o !== "object" ) return foam.mlang.ConstantExpr.create({ value: o });
+        if ( typeof o !== "object" ) return foam.mlang.predicate.Constant.create({ value: o });
+        if ( o instanceof Date ) return foam.mlang.predicate.Constant.create({ value: o });
         return o;
       }
     }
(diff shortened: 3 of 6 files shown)


src/core/stdlib.js
```diff
@@ -260,6 +260,12 @@ foam.LIB({
       if ( ! a || ! b ) return false;
       if ( a.equals ) return a.equals(b);
       return a == b;
+    },
+    function compare(a, b) {
+      if ( a.compareTo ) return a.compareTo(b);
```
adamvy-google, line 298: Did we add compareTo to the primitive types?  And if so, wont' this convert numbers into Numbers and end up being slow?

src/lib/dao.js
```diff
@@ -572,6 +573,8 @@ foam.CLASS({
   methods: [
     function put(o) {
       this.a.push(o);
+    },
+    function eof() {
```
adamvy-google: This shouldn't be necessary, the Sink interface provides a default empty implementation of each method.

src/lib/mlang.js
```diff
@@ -147,108 +259,207 @@ foam.CLASS({
...
 foam.CLASS({
-  package: 'foam.mlang',
+  package: 'foam.mlang.predicate',
+  name: 'ContainsICExpr',
+  extends: 'foam.mlang.predicate.BinaryExpr',
+  methods: [
+    function f(o) {
+      var s1 = this.arg1.f(o).toUpperCase();
+      var s2 = this.arg2.f(o).toUpperCase();
+      return s1 ? s1.indexOf(s2) !== -1 : false;
```
adamvy-google: If you going to guard against s1 being undefined you need to do that before you call toUpperCase()

src/lib/mlang.js
```diff
@@ -147,108 +259,207 @@ foam.CLASS({
...
+/** Unary expression that checks the given property is well-defined. */
+foam.CLASS({
+  package: 'foam.mlang.predicate',
+  name: 'HasExpr',
+  extends: 'foam.mlang.predicate.UnaryExpr',
+
+  methods: [
+    function f(obj) {
+      var value = this.arg1.f(obj);
+      var notHas = value === undefined || value === null || value === '' ||
```
adamvy-google, line 424: Are these the right semantics?  Or should HAS be essentially hasOwnProperty()

If I want to query for things that have a property defined rather than things that are the defaultValue.  I'm not sure what is more useful or expected.

test/js/lib/mlang.js
```diff
@@ -0,0 +1,496 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+require('../../../src/lib/dao');
+require('../../../src/lib/mlang');
+
+function buildPersonDAO() {
```
adamvy-google: I strongly prefer setup code like this to just live in the beforeEachTest.

Helper methods in tests just obfuscate what exactly you are testing and make it harder to update tests if the helper method ever gets shared between two tests.

src/core/stdlib.js
bshepherdson, line 298: I'm not sure. I don't think we did. They're slow, so it's probably better to add `typeof a === "number"` checks to `equals` and `compare` to handle this.

Kevin, thoughts?

src/lib/dao.js
bshepherdson: Ah, yep, that's true. I added `eof()` before realizing that ti didn't implement `Sink`.

Removed.

src/lib/mlang.js
bshepherdson: Whoops, fixed.

kgrgreer: Is the 'Expr' suffix redundant?  It's already in the foam.mlang.predicate package.  The code was originally written before we had packages, which add extra semantic information already.

src/lib/mlang.js
bshepherdson, line 424: We talked about this a while back. `HAS` is intended to mean "has an interesting, non-NULL value, even if that value is falsy". If there's a `defaultValue` that isn't `null` or `undefined`, then `HAS` is always true and that's fine.

Put another way, `HAS()` is roughly SQL `IS NOT NULL`.

test/js/lib/mlang.js
bshepherdson: When I wrote that I expected it to go in multiple different blocks. That didn't actually happen, so I've pulled this into the main describe's `beforeEach`.

src/lib/mlang.js
kgrgreer: This can be made much much faster by using a regex with the ignore case option.  At least leave a TODO:  See foam/apps/quickbug/DefaultQuery.js

src/lib/mlang.js
kgrgreer, line 424: Database semantics rather than JS semantics.  I think this is correct.

adamvy-google, line 424: It's not really JS semantics so much as FOAM semantics.

Databases don't have default values other than NULL usually.  What about a DEFINED and HAS as separate expressions, one for when an object has a set value for a property and isn't using the default, and another for when a property is an interesting value or not, regardless of defaults.

src/core/stdlib.js
kgrgreer, line 298: Yes, add a "number" check.  Also, check if Strings are slower as well, I think they might be, and then check for them if needed also.

bshepherdson, line 298: I've added a special case when both values have `typeof` `'number'`.

src/lib/mlang.js
bshepherdson, line 424: I'd be happy with `DEFINED` that was roughly equivalent to `hasOwnProperty`, once we had a cause for it. Right now, I'm not sure it's worth adding.

src/lib/mlang.js
```diff
@@ -73,72 +75,182 @@ foam.CLASS({
...
+ */
 foam.CLASS({
-  package: 'foam.mlang',
-  name: 'Expr'
+  package: 'foam.mlang.predicate',
+  name: 'Expr',
+  abstract: true,
+  // PORT: There should be several interfaces here eventually, eg.
+  // Functor/Projectable/whatever we call f()'s interface.
+  // implements: ['ExprProtocol'],
```
adamvy-google: Delete this for now.  ExprProtocol was a thing added for swift, we should be able to do this better.

src/lib/mlang.js
```diff
@@ -73,72 +75,182 @@ foam.CLASS({
...
+    /** Converts this expression into SQL, suitable for use in WHERE clauses. */
+    function toSQL() {
+      return this.cls_.label;
+    },
+    /** Converts this expression to a human-readable string for debugging. */
+    function toString() {
+      return this.toMQL();
+    },
+    /** Recursively adds all inputs of an expression to an array. */
+    function collectInputs(terms) {
```
adamvy-google: Delete this.

src/lib/mlang.js
```diff
@@ -73,72 +75,182 @@ foam.CLASS({
...
+      ],
+    },
+    /**
+     * Converts this expression into MQL syntax, as used in eg. Gmail searches.
+     */
+    function toMQL() {
+      return this.cls_.label;
+    },
+    /** Converts this expression into SQL, suitable for use in WHERE clauses. */
+    function toSQL() {
```
adamvy: Delete this

src/lib/mlang.js
```diff
@@ -73,72 +75,182 @@ foam.CLASS({
...
+      args: [
+        {
+          name: 'obj',
+        },
+      ],
+    },
+    /**
+     * Converts this expression into MQL syntax, as used in eg. Gmail searches.
+     */
+    function toMQL() {
```
adamvy-google: Delete toMQL and toSQL, we'll add them as refinements when we need them.

src/lib/mlang.js
```diff
@@ -73,72 +75,182 @@ foam.CLASS({
...
+    // PORT: pipe() is kind of the wrong name, this is a different meaning and
+    // signature from DAO.pipe(), which could be very confusing. chain()?
+    // wrap()?
+    /**
+     * Returns a Sink which applies this expression to each value put or
+     * removed, calling the provided sink only for those that match this
+     * expression.
+     */
+    /*
+    function pipe(sink) {
```
adamvy-google: I would rather delete this as well and put this as a TODO in the bottom of the file, like we did for mm.js/dao.js

src/lib/mlang.js
```diff
@@ -73,72 +75,182 @@ foam.CLASS({
...
+  package: 'foam.mlang.predicate',
+  name: 'Expr',
+  abstract: true,
+  // PORT: There should be several interfaces here eventually, eg.
+  // Functor/Projectable/whatever we call f()'s interface.
+  // implements: ['ExprProtocol'],
+
+  methods: [
+    /**
+     * Abstract projection method. Given an object, extracts the value of a
```
adamvy-google: This isn't always a projection method, in some cases it's a predicate function, in others it's a projection.

src/core/stdlib.js
```diff
@@ -289,6 +289,14 @@ foam.LIB({
       if ( ! a || ! b ) return false;
       if ( a.equals ) return a.equals(b);
       return a == b;
+    },
+    function compare(a, b) {
+      if ( typeof a === 'number' && typeof b === 'number' )
```
adamvy-google, line 294: Add typeof a === 'string' checks.

src/lib/mlang.js
```diff
@@ -73,72 +75,182 @@ foam.CLASS({
...
+    function toMQL() {
+      return this.cls_.label;
+    },
+    /** Converts this expression into SQL, suitable for use in WHERE clauses. */
+    function toSQL() {
+      return this.cls_.label;
+    },
+    /** Converts this expression to a human-readable string for debugging. */
+    function toString() {
+      return this.toMQL();
```
adamvy-google: toString should return a EQ(AND(....)) style looking thing

src/lib/mlang.js
bshepherdson: Done.

src/lib/mlang.js
bshepherdson: Done. Fixed `toString()` to not use `toMQL()` as well.

src/lib/mlang.js
bshepherdson: Done.

src/lib/mlang.js
bshepherdson: Done.

src/lib/mlang.js
bshepherdson: Adjusted the comment.

src/core/stdlib.js
bshepherdson, line 294: Done.

bshepherdson:
Now with `toString` support and tests for same. PTAL

src/lib/mlang.js
```diff
@@ -150,108 +232,259 @@ foam.CLASS({
...
-    function _nary_(name, args) { return this[name].create({ args: foam.array.argsToArray(args) }); },
-    function _binary_(name, arg1, arg2) { return this[name].create({ arg1: arg1, arg2: arg2 }); },
-    function OR() { return this._nary_("OrExpr", arguments); },
-    function AND() { return this._nary_("AndExpr", arguments); },
-    function IN(a, b) { return this._binary_("InExpr", a, b); },
-    function CONTAINS(a, b) { return this._binary_("ContainsExpr", a, b); },
-    function EQ(a, b) { return this._binary_("EqExpr", a, b); },
-    function GT(a, b) { return this._binary_("GtExpr", a, b); }
+    function _nary_(name, args) {
+      return this[name].create({ args: foam.array.argsToArray(args) });
```
kgrgreer, line 467: FYI, it is faster to do:
var r = this[name].create();
r.args = foam.array.argsToArray(args);
return r;
But since the cost of creating an mLang is orders of magnitude less than using it, it doesn't matter.

adamvy-google:
LGTM

adamvy-google:
Can this PR be deleted now?

bshepherdson:
Yeah. Rebasing to sort out some conflicts made Github lose track of this PR; all the commits are merged.

PR #2, opened 2016-03-11
Drop src/lib/promise.js and port DAOs to use standard Promises

bshepherdson:


src/lib/dao.js
```diff
@@ -662,9 +661,7 @@ foam.CLASS({
     },
 
     function put(obj, sink) {
-      sink = sink || this.Promise.create();
-      var promise = this.Promise.create();
-      promise.fulfill(sink);
+      sink = sink || this.ArraySink.create();
```
adamvy-google: An ArraySink doesn't seem useful here.

src/lib/dao.js
```diff
@@ -677,26 +674,24 @@ foam.CLASS({
       sink.put(obj);
       this.on.put.publish(obj);
 
-      return promise;
+      return Promise.resolve(obj);
```
adamvy-google: This isn't the exact behaviour we had before.  Before we returned a promise of the sink, and if the sink wasn't provided then we made the sink a promise of the object.

You could make sink optional and then return Promise.resolve(sink || obj);

src/lib/dao.js
```diff
@@ -704,24 +699,20 @@ foam.CLASS({
...
-          promise.reject(fc.errorEvt);
-          return promise;
+          return Promise.reject(fc.errorEvt);
         }
 
         sink.put(this.array[i], null, fc);
       }
 
-      sink.eof();
+      sink && sink.eof && sink.eof();
```
adamvy-google: Don't change this, let's force sinks to have eof for the time being.  I'm working on a plan that will allow uesrs to pass sinks with only a subset of the interface without having to do sink && sink.eof &&... every single time.

PR #5, opened 2016-04-04
Object pool axiom added.

jacksonic:


src/core/patterns.js
```diff
@@ -59,3 +59,61 @@ foam.CLASS({
...
+          nu.destroyed = false;
+          nu.initArgs(args, X);
+          nu.init && nu.init();
+        } else {
+          nu = oldCreate.apply(this, arguments);
+        }
+        return nu;
+      }
+
+      var oldDestroy = cls.prototype.destroy;
```
adamvy-google: What about doing cls.installAxiom(foam.core.Method.create({ name: 'destroy', code: function() { ... this.SUPER(); } }));  Then it will show in .describe() and IDEs nicely.

src/core/patterns.js
```diff
@@ -59,3 +59,61 @@ foam.CLASS({
...
+foam.CLASS({
+  package: 'foam.pattern',
+  name: 'Pooled',
+  axioms: [ foam.pattern.Singleton.create() ],
+
+  methods: [
+    function installInClass(cls) {
+      // Keeping the object pools in an accessible location allows them
+      // to be cleared out.
+      if ( ! foam.__objectPools__ ) { foam.__objectPools__ = {}; }
```
adamvy-google: Would it make sense to store the objectPool one the class itself?  This would make it possible to garbage collect a class if all the references to it and its instances are removed, regardless of if every instance was .destryoed().

Plus then it removes an extra lookup during create.

src/core/patterns.js
```diff
@@ -59,3 +59,61 @@ foam.CLASS({
...
+        return nu;
+      }
+
+      var oldDestroy = cls.prototype.destroy;
+      cls.prototype.destroy = function() {
+        if ( this.destroyed ) return;
+
+        // Run destroy process on the object, but leave its privates empty but intact
+        // to avoid reallocating them
+        var inst_ = this.instance_;
```
adamvy-google: I feel like this cleanup shouldn't be here.

If an object needs to clean up in .destroy() it should already do so correctly.  The object pool should just add pooling of instances.  Cleanup should be a responsibility of the object.

src/core/patterns.js
jacksonic: The only reason to keep them in one place is so you could wipe out all retained pools at once. Otherwise they would be better placed in the class.

An alternative would be to make a list of pooled classes in a shared location (and maybe it's optional) so you could iterate and clear them if necessary.

adamvy-google: In what instance would you want to wipe out all retained pools at once?  I can't imagine ever wanting to say "remote all pooled objects no matter what they are." That seems pretty dangerous compared to wiping out pools I know I no longer need.

jacksonic: The pool arrays would only contain destroyed objects that aren't being used. If you created a thousand things, destroyed them all and then moved on to something else, emptying the pools would free up those thousand retained things.

src/core/patterns.js
jacksonic: Done

src/core/patterns.js
jacksonic: Which reminds me that a size limit would be a smart thing to have!

adamvy-google: Ah right, of course, forgot they only contain destroyed objects.

I see what you're saying, and I'm not sure which would be better in the long run.  I still think it's a more useful property to be able to garbage collect a class and its associated pool, it just feels better componentized than creating a special global value.

Actually in regards to the global, what of instead of putting **objectPools** on "foam" add it as a property to foam.pattern.Pooled.  Pooled is afterall a singleton, so seems a sensible place to store "global" values only used by this pattern without mucking with the global foam object.

jacksonic: There's actually a potential problem: var pool is captured in create and destroy, so if a user went into **objectPools** and just assigned a new [], they wouldn't clear anything and lose the reference to the real pool array.

jacksonic: I'll give Pooled a property and stick the references there

PR #6, opened 2016-04-05
Idbdao

jacksonic:
Also includes Async generic DAO tests to support IDB cleanup for each test.

### DIFF
--- a/src/lib/IDBDAO.js
+++ b/src/lib/IDBDAO.js
@@ -61,6 +61,11 @@ foam.CLASS({
 //    data store.  Obviously this will get slow if you store large amounts
 //    of data in the database.
 //  */},
+  constants: {
+    /** Global cache of the current transaction reference. Only element 0 is used. */
+    __TXN__: [],
+  },
+
   properties: [
     {
       name:  'of',
@@ -153,20 +158,21 @@ foam.CLASS({
     },
 
     function withStore_(mode, fn) {
-      if ( global.__TXN__ && global.__TXN__.store ) {
+      var self = this;
+      if ( self.__TXN__[0] ) {
         try {
-          fn.call(this, __TXN__.store);
+          fn.call(self, self.__TXN__[0]);
           return;
         } catch (x) {
-          global.__TXN__ = undefined;
+          self.__TXN__[0] = undefined;
         }
       }
-      this.openDB((function (db) { // TODO: memoize like in foam1
-        var tx = db.transaction([this.name], mode);
-        var os = tx.objectStore(this.name);
-        if ( global.__TXN__ ) global.__TXN__.store = os;
-        fn.call(this, os);
-      }).bind(this));
+      self.openDB((function (db) { // TODO: memoize like in foam1
+        var tx = db.transaction([self.name], mode);
+        var os = tx.objectStore(self.name);
+        self.__TXN__[0] = os;
+        fn.call(self, os);
+      });
     },
 
     function put(value, sink) {
--- a/src/lib/IDBDAO.js
+++ b/src/lib/IDBDAO.js
@@ -167,7 +167,7 @@ foam.CLASS({
           self.__TXN__[0] = undefined;
         }
       }
-      self.openDB((function (db) { // TODO: memoize like in foam1
+      self.openDB(function (db) { // TODO: memoize like in foam1
         var tx = db.transaction([self.name], mode);
         var os = tx.objectStore(self.name);
         self.__TXN__[0] = os;
--- a/src/lib/IDBDAO.js
+++ b/src/lib/IDBDAO.js
@@ -302,8 +302,8 @@ foam.CLASS({
     },
 
     function select(sink, options) {
-      sink = sink || this.ArraySink.create();
-      sink = this.decorateSink_(sink, options);
+      var resultSink = sink || this.ArraySink.create();
+      sink = this.decorateSink_(resultSink, options);
 
       var fc = this.FlowControl.create();
       var self = this;
@@ -326,7 +326,7 @@ foam.CLASS({
 
             if (!cursor) {
               sink.eof && sink.eof();
-              resolve(sink);
+              resolve(resultSink);
               return;
             }
 
--- a/src/lib/IDBDAO.js
+++ b/src/lib/IDBDAO.js
@@ -52,10 +52,6 @@ foam.CLASS({
       }
     },
     {
-      name: 'useSimpleSerialization',
-      value: true
-    },
-    {
       name: 'indicies',
       factory: function() { return []; }
     }
@@ -73,30 +69,14 @@ foam.CLASS({
         if ( prop.shortName ) propKeys[prop.shortName] = prop.name;
       }
 
-      if ( this.useSimpleSerialization ) {
-        this.serialize = this.SimpleSerialize;
-        this.deserialize = this.SimpleDeserialize;
-      } else {
-        this.serialize = this.FOAMSerialize;
-        this.deserialize = this.FOAMDeserialize;
-      }
-
       this.withDB = foam.fn.memoize1(this.openDB.bind(this));
     },
 
-    function FOAMDeserialize(json) {
+    function deserialize(json) {
       return foam.json.parse(foam.json.parseString(json)); //TODO: serialization
     },
 
-    function FOAMSerialize(obj) {
-      return foam.json.stringify(obj); //TODO: serialization
-    },
-
-    function SimpleDeserialize(json) {
-      return foam.json.parse(foam.json.parseString(json)); //TODO: serialization
-    },
-
-    function SimpleSerialize(obj) {
+    function serialize(obj) {
       return foam.json.stringify(obj); //TODO: serialization
     },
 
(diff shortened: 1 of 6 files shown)


src/lib/IDBDAO.js
```diff
@@ -0,0 +1,377 @@
...
+    'async',
+  ],
+
+//   function documentation() {/*
+//   Usage:<br/>
+//    <code>var dao = IDBDAO.create({model: Issue});<br/>
+//    var dao = IDBDAO.create({model: Issue, name: 'ImportantIssues'});<br/></code>
+//   <br/>
+//    TODO:<br/>
+//    Optimization.  This DAO doesn't use any indexes in indexeddb yet, which
```
adamvy-google: I don't know if this is even an TODO worth having.  Optimizing against IDB is a losing battle.  Maybe some minor improvements can be made but in general you're not going to get very far, better off caching in an MDAO.

jacksonic: That's from the old IDBDAO, I'll edit that doc comment

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,377 @@
...
+
+    function openDB(cc) {
+      var indexedDB = window.indexedDB ||
+        window.webkitIndexedDB         ||
+        window.mozIndexedDB;
+
+      var request = indexedDB.open("FOAM:" + this.name, 1);
+      var self = this;
+
+      request.onupgradeneeded = (function(e) {
```
adamvy-google: Why the extra parenthesis around this function and below?

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,377 @@
...
+      if ( ! this.q_ ) {
+        var q = [fn];
+        this.q_ = q;
+        this.async(function() {
+          self.withStore_(mode, function(store) {
+            // console.log('q length: ', q.length);
+            if ( self.q_ == q ) self.q_ = undefined;
+            for ( var i = 0 ; i < q.length ; i++ ) q[i](store);
+          });
+        }, this.X)();
```
adamvy-google: async doesn't take a second context parameter.

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,377 @@
...
+      var self = this;
+      if ( self.__TXN__[0] ) {
+        try {
+          fn.call(self, self.__TXN__[0]);
+          return;
+        } catch (x) {
+          self.__TXN__[0] = undefined;
+        }
+      }
+      self.openDB(function (db) { // TODO: memoize like in foam1
```
adamvy-google: You could probably memoize this by making withDB a promise of the return value of openDB.

this.withDB = new Promise(function(resolve, reject) { this.openDB(resolve); };

Then this.withDB.then(function(db) { <db - work> });

src/lib/IDBDAO.js
jacksonic: Not sure what's up there. Removing

src/lib/IDBDAO.js
jacksonic: Removing

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,377 @@
...
+            function(e) {
+              self.pub('on','put', value);
+              sink && sink.put && sink.put(value);
+              resolve(sink || value);
+            });
+          request.transaction.addEventListener(
+            'error',
+            function(e) {
+              sink && sink.error && sink.error('put', value, e); // TODO: err message
+              reject(self.IDBException.create({ id: value.id, error: e }));
```
adamvy-google: Let's add a TODO somewhere to sort out which error events are internal vs external.

For now I believe we should treat them all as internal errors, but maybe things like DataCloneError could be considered external ( although if you get that error its probably a bug with IDB serialization and thus internal ).

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,377 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'IDBException',
+  extends: 'foam.dao.ExternalException',
```
adamvy-google: Let's make this an InternalException and rename to IDBInternalException.  Most exceptions that indexeddb throws are likely internal.

Internal means there's something going wrong inside IDB, like it can't talk to the disk/network.  External means there's something wrong with the data you are trying to store, such as it doesn't pass validation, or you don't have permissions.

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,377 @@
...
+              reject(self.IDBException.create({ id: value.id, error: e }));
+            });
+        });
+      });
+    },
+
+    function find(key) {
+      var self = this;
+
+      if ( foam.mlang.predicate.Expr.isInstance(key) ) {
```
adamvy-google: I don't think we should support finding with an expression any more.  It was rarely used, and not always clear.  If you really want to find based off an expression you can do limit(1).where().select() yourself.

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,377 @@
...
+                if (!request.result) {
+                  reject(self.ObjectNotFoundException.create({ id: key }));
+                  return;
+                }
+                var result = self.deserialize(request.result);
+                resolve(result);
+              });
+            request.onerror = function(e) {
+              // TODO: Parse a better error out of e
+              reject(e); // TODO: err message
```
adamvy-google: Build an IDB internal exception here and reject with that

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,359 @@
...
+
+    function remove(obj) {
+      var self = this;
+      var key = obj.id != undefined ? obj.id : obj;
+      return new Promise(function(resolve, reject) {
+        self.withStore("readwrite", function(store) {
+          var getRequest = store.get(key);
+          getRequest.onsuccess = function(e) {
+            if (!getRequest.result) {
+              resolve(null); // not found? as good as removed!
```
adamvy-google: To be consistent with ArrayDAO resolve with the obj passed in to remove().  We can discuss if this is correct at next design meeting.

src/lib/IDBDAO.js
jacksonic: Not sure how I missed that one, with two TODOs on it :P

src/lib/IDBDAO.js
jacksonic: Done

src/lib/IDBDAO.js
jacksonic: Done

src/lib/IDBDAO.js
adamvy-google: Change of plans from FOAM discussion.

Resolve with nothing just resolve() on delete, regardless of whether the object was there before or not.

src/lib/IDBDAO.js
jacksonic: Done

src/lib/IDBDAO.js
jacksonic: Done, no object passed to resolve() in any case. Also fixed missing pub and return in not-found case.

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,350 @@
...
+        });
+      }
+    },
+  ],
+
+  methods: [
+
+    function init() {
+
+      var propKeys = this.propKeys_ = {};
```
adamvy-google: Probably safe to delete this.  It looks like propKeys_ was only used in the FOAM1 IDBDAO for simple serialization.  But we should just make foam.json smart enough to use shortnames if available.

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,350 @@
...
+        self.withStore("readonly", function(store) {
+          var request = store.get(key);
+          request.transaction.addEventListener(
+            'complete',
+            function() {
+              if (!request.result) {
+                reject(self.ObjectNotFoundException.create({ id: key }));
+                return;
+              }
+              var result = self.deserialize(request.result);
```
adamvy-google, line 194: Let's put a todo somewhere that we should catch errors from deserialization/serialization treat them as internal errors and surface them to the users as such.

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,350 @@
...
+          getRequest.onsuccess = function(e) {
+            if (!getRequest.result) {
+              // not found? as good as removed!
+              self.pub('on','remove', data);
+              resolve();
+              return;
+            }
+            var data = self.deserialize(getRequest.result);
+            var delRequest = store.delete(key);
+            delRequest.transaction.addEventListener('complete', function(e) {
```
adamvy: What happens if we call delete an the object isn't there?  Can we detect that case and treat it as a success.  Then we can skip the entire getRequest step.

adamvy: Disregard, we need the get request for the event to work.

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,350 @@
...
+            };
+          };
+          getRequest.onerror = function(e) {
+            reject(self.IDBInternalException.create({ id: key, error: e }));
+          };
+        });
+      });
+    },
+
+    function removeAll(sink, skip, limit, order, predicate) {
```
adamvy-google: Remove the sink parameter to match upstream now

src/lib/IDBDAO.js
```diff
@@ -0,0 +1,350 @@
...
+      var query = predicate || this.True.create();
+
+      var self = this;
+
+      // If the caller doesn't care to see the objects as they get removed,
+      // then just nuke them in one go.
+      if ( ! predicate && ! ( sink && sink.remove ) ) {
```
adamvy-google: We need to check if anyone is listening for remove events.  If nobody is listening, and we have no predicate, then its safe to .clear().  This appears to have been a bug in FOAM1, thankfully few people actually nuke an entire dao.

PR #7, opened 2016-04-08
Basic Polymer integration

bshepherdson:
<foam-select> binds a DAO (currently an eval'd expression - ugh!) to a Polymer
array, and then syncs changes between the DAO and the array. (That means both
upstream changes to the DAO and manipulating the array through your Polymer UI).

Check out polymer/index.html for a todo-list demo app.

### DIFF
--- a/polymer/foam-select.html
+++ b/polymer/foam-select.html
@@ -1,0 +1,156 @@
+<link rel="import" href="./bower_components/polymer/polymer.html">
+
+<!--
+Two-way binds a FOAM DAO to an array, for use in eg. `dom-repeat`.
+
+Similar to `iron-localstorage` and `firebase-collection`.
+
+Input: the `dao` property should the path from `window` to a DAO object.
+
+The DAO is watched for changes (edits, new items and removed items) which will
+be updated in the output array.
+
+Output: `data` is an Array. It is watched for changes, which will be written
+back to the DAO. That includes new elements, removed elements, and edits to
+particular elements.
+-->
+<dom-module id="foam-select">
+  <script>
+  Polymer({
+    is: 'foam-select',
+
+    properties: {
+      dao: String,
+      data: {
+        type: Array,
+        notify: true
+      }
+    },
+
+    observers: [
+      '_onDataUpdate(data.*)'
+    ],
+
+    ready: function() {
+      this._ids = {};
+
+      this._sink = {
+        put: this._put.bind(this),
+        remove: this._remove.bind(this),
+        reset: this._reset.bind(this),
+        eof: function() {}
+      };
+
+      this.feedback_ = true;
+      this.data = [];
+      this.feedback_ = false;
+
+      this._dao = this.get(this.dao, window);
+      this._dao.pipe(this._sink);
+    },
+
+    _put: function(obj) {
+      obj = obj.clone();
+      obj.propertyChange.sub(function(sub) {
+        sub.destroy();
+        this._dao.put(obj);
+      }.bind(this));
+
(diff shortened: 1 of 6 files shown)


bshepherdson:
I updated my email for this repository to be my @google.com one, so hopefully the CLA thing won't whine about it. In any case, I'm a licensed contributor.

polymer/foam-select.html
```diff
@@ -0,0 +1,138 @@
...
+        this.dao.put(obj);
+      }.bind(this));
+
+      if (typeof this._ids[obj.id] !== 'undefined') {
+        // Replace that user with this new one.
+        this.feedback_ = true;
+        this.splice('data', this._ids[obj.id], 1, obj);
+        this.feedback_ = false;
+        return;
+      } else {
```
kgrgreer: No need for else if you've already returned.

polymer/foam-select.html
```diff
@@ -0,0 +1,138 @@
...
+    },
+
+    _put: function(obj) {
+      obj = obj.clone();
+      obj.propertyChange.sub(function(sub) {
+        sub.destroy();
+        this.dao.put(obj);
+      }.bind(this));
+
+      if (typeof this._ids[obj.id] !== 'undefined') {
```
kgrgreer: Match coding standard.

bshepherdson:
PTAL, all addressed.

Two coding style questions:
- My understanding is that `if ( foo ) return;` and similar are allowed, but if it needs to wrap onto two lines, then it needs braces. Correct?
- Loops are spelled `for ( var i = 0; i < foo.length; i++ ) {`, with the semicolons hugging left and spaced right.

adamvy-google:
Loops should be spelled for ( var i = 0 ; i < foo.length ; i++ ) {

spaces on both sides of the semicolon

polymer/foam-select.html
```diff
@@ -0,0 +1,138 @@
+<link rel="import" href="./bower_components/polymer/polymer.html">
+
+<dom-module id="foam-select">
```
adamvy-google, line 17: Would be good to have documentation about what this element does as foam-select is not a very descriptive name.

polymer/foam-select.html
```diff
@@ -0,0 +1,138 @@
...
+        for ( var j = 0 ; j < splice.addedCount ; j++ ) {
+          addedKeys[splice.object[splice.index + j].id] = true;
+        }
+        // Now everything that's in removed but not in addedKeys needs to be
+        // removed from the DAO.
+        for ( var j = 0 ; j < splice.removed.length ; j++ ) {
+          var obj = splice.removed[i];
+          if ( ! addedKeys[obj.id] ) {
+            delete this._ids[obj.id];
+            window.setTimeout(this._asyncRemove.bind(this, obj), 0);
```
adamvy-google: Design wise, I would move the setTimeout into the _asyncPut/_asyncRemove functions so you can just call this._asyncPut(obj) here.

polymer/foam-select.html
```diff
@@ -0,0 +1,138 @@
...
+        remove: this._remove.bind(this),
+        reset: this._reset.bind(this),
+        eof: function() {}
+      };
+
+      this.feedback_ = true;
+      this.data = [];
+      this.feedback_ = false;
+
+      this.dao = eval(this.dao);
```
adamvy-google: It looks like you could change the type of the dao property to string and do this.get(this.dao, window);

https://www.polymer-project.org/1.0/api/#Polymer.Base:method-get

adamvy-google:
LGTM overall, see if my alternative to eval will work.

Feel free to merge this before documenting foam-select

PR #10, opened 2016-04-12
MAP mlang, Func expression

jacksonic:
MAP can take a function directly, since adapt will now convert it to a Func expr.


adamvy-google:
LGTM

PR #15, opened 2016-04-25
Query parser port from FOAM 1

bshepherdson:


src/core/parse.js
```diff
@@ -301,16 +307,22 @@ foam.CLASS({
...
   methods: [
     function step() {
       var ps1 = this.ps;
       var ps = this.ps;
-      var str = this.s;
+      var str = this.case ? this.s : this.s.toLowerCase();
       for ( var i = 0 ; i < str.length ; i++, ps = ps.tail ) {
-        if ( str.charAt(i) !== ps.head ) {
+        if ( ! ps.head ||
+            str.charAt(i) !== (this.case ? ps.head : ps.head.toLowerCase()) ) {
```
adamvy-google: I'm concerned about the performance this will have the parsers, adding an extra branch point every character even if we're not in a literal_ic situation.

Let's make LiteralIC it's own parser like it was in foam1

src/core/parse.js
```diff
@@ -342,6 +354,36 @@ foam.CLASS({
...
+  extends: 'foam.parse.compiled.State',
+
+  properties: [
+    {
+      name: 'from',
+      final: true,
+    },
+    {
+      name: 'to',
+      final: true,
```
adamvy-google: remove tailing commas

src/lib/QueryParser.js
```diff
@@ -0,0 +1,438 @@
...
+              typeof v[2] === 'number' ? new Date(+v[2] + 1, 0, 1) : v[2];
+          return [start, end];
+        },
+
+        'quoted string': compactToString,
+        word: compactToString,
+      };
+
+      var g = this.ImperativeGrammar.create({
+        symbols: function() { return grammar; },
```
adamvy-google: You can just do symbols: grammar, it will do the right thing.

adamvy-google:
Overall LGTM, minus some minor parser changes.

src/core/parse.js
bshepherdson: Done.

src/core/parse.js
bshepherdson: Done. All jscs-backed style complaints cleaned up for files other than `core/parse.js`.

src/core/parse.js
```diff
@@ -280,17 +272,47 @@ foam.CLASS({
...
+      final: true
+    },
+    {
+      name: 'value',
+      final: true
+    }
+  ],
+  methods: [
+    function step() {
+      var str = this.s.toLowerCase();
```
kgrgreer: Shouldn't we do this on the postSet() of 's' so that we only have to do it once?

src/core/parse.js
```diff
@@ -280,17 +272,47 @@ foam.CLASS({
...
-      this.success.ps = ps.setValue(this.value || str);
+      this.success.ps = ps.setValue(this.value || this.s);
       return this.success;
     }
   ]
 });
 
+foam.CLASS({
+  package: 'foam.parse.compiled',
+  name: 'LiteralICWithValue',
```
kgrgreer, line 294: Docs to explain what this is/does.

src/core/parse.js
```diff
@@ -321,6 +343,34 @@ foam.CLASS({
...
+      name: 's',
+      final: true
+    }
+  ],
+
+  methods: [
+    function step() {
+      var ps1 = this.ps;
+      var ps = this.ps;
+      var str = this.s.toLowerCase();
```
kgrgreer: PostSet or adapt 's' to avoid repetition.

src/lib/QueryParser.js
```diff
@@ -0,0 +1,446 @@
...
+    'foam.mlang.predicate.True',
+    'foam.parse.Alternate',
+    'foam.parse.ImperativeGrammar',
+    'foam.parse.Literal',
+    'foam.parse.Parsers',
+    'foam.parse.StringPS'
+  ],
+
+  imports: [
+    'me'
```
kgrgreer: Add comment.

src/lib/QueryParser.js
```diff
@@ -0,0 +1,446 @@
...
+      var fields = [];
+      var properties = cls.getAxiomsByClass(foam.core.Property);
+      for ( var i = 0; i < properties.length; i++ ) {
+        var prop = properties[i];
+        fields.push(this.Literal.create({
+          s: prop.name,
+          case: false,
+          value: prop
+        }));
+        // TODO(braden): Support aliases and shortName when Property does.
```
kgrgreer: Property already has shortName.

src/lib/mlang.js
```diff
@@ -339,6 +339,46 @@ foam.CLASS({
...
 });
 
+foam.CLASS({
+  package: 'foam.mlang.predicate',
+  name: 'InIC',
+  extends: 'foam.mlang.predicate.Binary',
+
+  properties: [
+    {
+      name: 'arg2',
```
kgrgreer, line 352: Can we make this more efficient by pre-toUpperCase()'ing?

bshepherdson, line 352: We do; see how valueSet_ is constructed below. We can't pre-uppercase if `arg2` is not an array; it might be projected from the input object.

bshepherdson:
All fixed, except the `InIC` one replied to above.

src/core/parse.js
```diff
@@ -1231,7 +1420,7 @@ foam.CLASS({
   package: 'foam.parse',
   name: 'PSymbol',
 
-  properties: [ 'name', 'parser' ]
+  properties: ['name', 'parser']
```
adamvy-google, line 1453: I believe we actually do want those spaces to be there.  The linter config I have in the real repo requires those spaces.  I've backported it to this repo (something I should have done a while ago).

I wouldn't worry too much about style in this repo though, we can fix it up when we move stuff to the real repo.

bshepherdson:
This is still passing tests. The coding style isn't perfect. Should I fix it up, or are we ignoring that for now, to be fixed as it lands in the real repo?

adamvy-google:
We're not ignoring coding style, but also don't spend a lot of time on it.

If the code is readable, don't worry too much about specific spacing for now

bshepherdson:
Freshly updated, no conflicts now.

We really should get this checked in. Kevin/Alex, can we get final word on whether I should make sure this is style-clean now, or later?

src/core/parse.js
```diff
@@ -278,23 +270,63 @@ foam.CLASS({
...
       var ps = this.ps;
       for ( var i = 0 ; i < str.length ; i++, ps = ps.tail ) {
-        if ( str.charAt(i) !== ps.head ) {
+        if ( ! ps.head || str.charAt(i) !== ps.head.toLowerCase() ) {
           this.fail.ps = this.ps;
           return this.fail;
         }
       }
-      this.success.ps = ps.setValue(this.value || str);
+      this.success.ps = ps.setValue(this.value || this.s);
```
adamvy-google: Why the change to this.s ? str was already set to this.s

src/core/parse.js
```diff
@@ -635,19 +725,69 @@ foam.CLASS({
...
-      })
+      });
     },
 
-    function parse(ps, obj) {
+    function parse(ps) {
       var str = this.s;
       for ( var i = 0 ; i < str.length ; i++, ps = ps.tail ) {
-        if ( str.charAt(i) !== ps.head ) return undefined;
+        if ( ! ps.head || str.charAt(i) !== ps.head ) {
```
adamvy-google: I don't see why all these ! ps.head checks were added to the literal parsers.  FOAM1 parsers did not include them.

src/core/parse.js
```diff
@@ -635,19 +725,69 @@ foam.CLASS({
...
+          return undefined;
+        }
       }
-      return ps.setValue(this.value !== undefined ? this.value : str);
+      return ps.setValue(this.value !== undefined ? this.value : this.s);
```
adamvy-google: Again, str is already set to this.s

src/core/parse.js
bshepherdson: They're actually needed only for the `LiteralIC` versions. (Which end up calling `undefined.toLowerCase()` without the check.) I've removed these where they were unnecessary.

src/lib/QueryParser.js
```diff
@@ -0,0 +1,459 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+foam.CLASS({
+  package: 'foam.parse',
+  name: 'QueryParser',
+  axioms: [foam.pattern.Singleton.create()],
```
adamvy-google: Alternative design idea.

What if we make this a multiton based upon the id of the class you pass in?

Then we can just do foam.parse.QueryParser.create({ of: 'some.class' }), in every spot we need the query parser, no need to pass it around or make sure we only call parserForCls() once per class.

Then QueryParser could extend Grammar and could be used like this.

this.QueryParser.create({ of: 'some.Class' }).parseString(query)

src/lib/QueryParser.js
```diff
@@ -0,0 +1,459 @@
...
+    'foam.mlang.predicate.Or',
+    'foam.mlang.predicate.True',
+    'foam.parse.Alternate',
+    'foam.parse.ImperativeGrammar',
+    'foam.parse.LiteralIC',
+    'foam.parse.Parsers',
+    'foam.parse.StringPS'
+  ],
+
+  imports: [
```
adamvy-google: I'm not sure me should be imported, especially if this is a singleton/multiton.  Maybe it should be a parameter to the parse method, or should be part of the .create() arguments if we make this a multiton.

bshepherdson:
Updated to use a Multiton. PTAL

adamvy-google:
LGTM

PR #17, opened 2016-04-29
Flyweight pattern added, used in MDAO index

jacksonic:


src/core/patterns.js
```diff
@@ -152,3 +152,78 @@ foam.CLASS({
     }
   ]
 });
+
+/** Flyweight classes create instances that can create flyweight copies.
```
kgrgreer, line 156: This is not what a "Flyweight" is.  You could call it a "lightweight" or something else, but Flyweight has a specific meaning, which is different from this.

src/lib/IndexBinaryTree.js
```diff
@@ -29,8 +30,8 @@ foam.CLASS({
     { class: 'Simple', name: 'left'  },
     { class: 'Simple', name: 'right' },
 
-    { name: 'index' },
-    { name: 'selectCount' }, // compare, dedupe, nullNode
+    { name: 'index' },
```
kgrgreer, line 33: Extra space at EOL.

src/core/patterns.js
jacksonic, line 156: Yes, but it's flyweight in that "If a flyweight pattern is what's appropriate, use this instead".

A Java implementation would look very different on the output side, but ideally we could generate it  from the same models. (though an axiom alone might not be enough, and if that won't work then there's no reason to call this Flyweight and get people's hopes up).

PR #18, opened 2016-05-05
Adapting a property value in a getter is a pain

jacksonic:
Here's a Class property that does the lookup at the last possible moment. Decorating the default getter is a pain, though.


jacksonic:
Another way to do it might be to run the installInProto super first, then decorate the getter and redefine the property.

PR #20, opened 2016-05-12
Many improvements to the Angular integration

bshepherdson:
These are in support of the Tracker demo app.

Highlights:
- `<foam-details>` is an automatic detail view
- `<foam-dao-controller>` clones, and handles the object changing properly.
- `<foam-view>` can handle its DAO not being set right away, and its DAO changing.


bshepherdson:
I signed it! Frigging @googlebot ; I'm signed up for the Google Corporate CLA, of course. I just added my @google.com email to my Github account though, so maybe that will fix it?

PR #21, opened 2016-05-13
Angular issue tracker demo app

bshepherdson:


angular/tracker/src/models.js
```diff
@@ -0,0 +1,224 @@
...
+    {
+      name: 'isOpen',
+      value: false
+    }
+  ],
+
+  values: [
+    {
+      name: 'NEW',
+      values: { label: 'New', isOpen: true }
```
adamvy-google, line 47: The seaprate values:{} field is no longer required for enums.

Just

{
name: 'NEW',
label: 'New',
isOpen: true
}

angular/tracker/src/models.js
```diff
@@ -0,0 +1,224 @@
...
+    {
+      name: 'OBSOLETE',
+      values: { label: 'Obsolete' }
+    }
+  ]
+});
+
+
+foam.CLASS({
+  package: 'ng.dao',
```
adamvy-google, line 86: Some sort of documentation for this class?  What does it do?

adamvy-google:
I think we should move this to com.google.foam.demos packages, and src/com/google/foam/demos/...

It needs to be in a demo directory somewhere, and src/com/google/foam/demos is the canonical one now.

angular/tracker/src/models.js
```diff
@@ -0,0 +1,224 @@
...
+});
+
+foam.CLASS({
+  package: 'ng.tracker',
+  name: 'User',
+
+  properties: [
+    { class: 'String', name: 'id', required: true },
+    { class: 'String', name: 'name', required: true },
+    { class: 'String', name: 'imageUrl' }
```
kgrgreer, line 221: We have URL type.

PR #22, opened 2016-05-26
Api

jacksonic:
XHRMethod for easy implementation of DAOs or adapters for REST calls.


src/lib/HTTPMethod.js
```diff
@@ -0,0 +1,179 @@
...
+    },
+
+    function installInProto(p) {
+      // generate body to call through to xhr
+      var axiom = this;
+
+      // set up function with correct args, pass them into the
+      // actual implementation, callRemote_()
+      // ALTERNATE: just pass args in and read from arguments
+      var code = "(function "+axiom.name+"_"+"(";
```
adamvy-google: Might be better to just create an anonymous function here in a closure and return that.

I know in other places I used new Function() to create a function, but I tried to limit those to places where the function code was very very short and readably, and it was partly an experiment to see if it would change performance.

new Function() has the advantage that when creating the function it does _not_ create a closure.  This theoretically results in a function that runs faster, but I never verified that, and the JIT probably makes this not true.

This is a pretty complicated eval, I have a hard time looking at this and understanding what the code at the end will look like.  Plus eval is not great for startup performance.  I would favour just returning an anonymous function.

test/helpers/generic_dao.js
```diff
@@ -24,8 +24,13 @@
 
 
 global.genericDAOTestBattery = function(daoFactory) {
+  var oldContext;
+
   describe('generic DAO tests', function() {
     beforeEach(function() {
+      oldContext = foam.__context__;
```
adamvy-google: Is this really part of this CL, or just a bad merge?  These things should be covered by testcontext.js now.

src/lib/HTTPMethod.js
jacksonic: The only advantage the generated function has is the named arguments, if you happened to debug it, but otherwise we can add type checking or debug help to whatever is generated either way. I'll swap in an args array version

test/helpers/generic_dao.js
jacksonic: Nested suites were a problem at one point, but I'm not seeing errors related to that anymore. Removing

src/lib/HTTPMethod.js
```diff
@@ -64,6 +64,27 @@ foam.CLASS({
         create a partially filled request object. */
       name: 'HTTPRequestFactoryName',
       value: 'HTTPRequestFactory'
+    },
+    {
+      name: 'code',
+      required: false,
+      transient: true,
+      factory: function() {
```
adamvy-google: This should probably be an expression off of this.args

src/lib/HTTPMethod.js
```diff
@@ -79,25 +100,8 @@ foam.CLASS({
...
-        var opt_args = {};
-        for ( var i = 0; i < arguments.length && i < argNames.length; i ++ ) {
-          opt_args[argNames[i]] = arguments[i];
-        }
-        return axiom.callRemote_(opt_args, this);
-      }
-
       // set code on proto
-      p[axiom.name] = this.code;
+      p[this.name] = this.code;
```
adamvy-google, line 104: Can probably delete this entire installInProto method now right?  The parent class impl should be sufficient.

PR #24, opened 2016-06-14
U2: Import U2, modernized and ready for FOAM 2

bshepherdson:


src/lib/u2.js
```diff
@@ -0,0 +1,1419 @@
...
+    __ID__: [ 0 ],
+    NEXT_ID: function() {
+      return 'u2v' + this.__ID__[ 0 ]++;
+    }
+  },
+
+  properties: [
+    {
+      name: 'factory',
+      value: function() { return foam.u2.EID.NEXT_ID(); }
```
adamvy-google, line 42: This could probably be changed to just use 'u2v' + foam.next$UID();

No need to have multiple unique id generators.

src/lib/u2.js
```diff
@@ -0,0 +1,1419 @@
...
+    // Initial State of an Element
+    INITIAL: {
+      output: function(out) {
+        this.initE();
+        this.output_(out);
+        this.state = this.OUTPUT;
+        return out;
+      },
+      load:          function() {
+        console.error('Must output before loading.');
```
adamvy-google: Switch these to this.error() which exists on all FOBjects in FOAM2

src/lib/u2.js
```diff
@@ -0,0 +1,1419 @@
...
+    },
+
+    // State of an Element after it has been output (to a String) but before it is loaded.
+    // This should be only a brief transitory state, as the Element should be loaded
+    // almost immediately after being output.  It is an error to try and mutate the Element
+    // while in the OUTPUT state.
+    OUTPUT: {
+      output: function(out) {
+        // Only warn because it could be useful for debugging.
+        console.warn('Duplicate output.');
```
adamvy-google: this.warn()

PR #27, opened 2016-07-18
Introduce optional event tracing components (and a few tests for them)

mdittmer:
Includes core changes:
  src/core/FObject.js : Implement mLang expression-based event matching
  src/core/Listeners.js : Add metadata to listener functions to support accurate async tracing
  src/core/Window.js : Rework async listener abstractions to facilitate tracing refinement

### DIFF
--- a/src/core/FObject.js
+++ b/src/core/FObject.js
@@ -132,6 +132,12 @@ foam.CLASS({
       removing subscriptions be done quickly by connecting next to prev
       and prev to next.
 
+      Additional properties support mLang matching of topics. 'expr'
+      contains the expression to match against at this element of the
+      topic list; 'exprs' contains a map of expression-match children,
+      and 'hasExprs' provides an efficient check for whether 'exprs' is
+      empty.
+
       Note that both the head structure and the nodes themselves have a
       'next' property. This simplifies the code because there is no
       special case for handling when the list is empty.
@@ -147,10 +153,20 @@ foam.CLASS({
           subTopic1: <same structure>,
           ...
           subTopicn: <same structure>
-      }
+      },
+      expr: <mLang-expr or null>,
+      exprs: {
+        <expr1.toString()>: <same structure: expr=expr1>,
+        ...
+        <exprm.toString()>: <same structure: expr=exprm>
+      },
+      hasExprs: <bool: is exprs empty?>
     */
     function createListenerList_() {
-      return { next: null };
+      return {
+        next: null,
+        expr: null
+      };
     },
 
     /** Return the top-level listener list, creating if necessary. */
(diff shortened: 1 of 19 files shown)


src/core/FObject.js
```diff
@@ -169,36 +185,70 @@ foam.CLASS({
...
+      case 7: l(s, a[0], a[1], a[2], a[3], a[4], a[5], a[6]); break;
+      case 8: l(s, a[0], a[1], a[2], a[3], a[4], a[5], a[6], a[7]); break;
+      case 9: l(s, a[0], a[1], a[2], a[3], a[4], a[5], a[6], a[7], a[8]); break;
+      default: l.apply(l, [s].concat(Array.from(a)));
+      }
+    },
+
     function hasListeners(/* args */) {
       /** Return true iff there are listeners for the supplied message. **/
+      var args = Array.from(arguments);
```
kgrgreer: Isn't this going to cause the creation of garbage on every property set, thus greatly decreasing our performance?

mdittmer: @kgrgreer, you're right. Prefer passing |arguments| as only argument to |has(Keyed|Expr)Listeners_| below?

src/core/tracing/Window.js
```diff
@@ -0,0 +1,91 @@
...
+      value: null
+    }
+  ],
+
+  methods: [
+    function getAsyncListener_(state) {
+      var ctx = this;
+      var baseListener = this.SUPER(state);
+      return function asyncListener() {
+        var tps = foam.lookup('foam.core.FObject').TRACING_PUB_SUB;
```
kgrgreer: Better to require: FObject then just do this.FObject.

src/core/tracing/Window.js
```diff
@@ -0,0 +1,91 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Override some bits of foam.core.Window to support more accurate tracing.
+foam.CLASS({
+  package: 'foam.core.tracing',
+  name: 'Window',
+  extends: 'foam.core.Window',
```
kgrgreer, line 22: Why does this need to extend Window?
When I create a new Window, how will I know to add this context as well?

test/src/core/Boot.js
```diff
@@ -755,6 +755,7 @@ describe('Slots', function() {
     expect(t.a$.isDefined()).toBe(true);
   });
   it('binds property slots', function() {
+    debugger;
```
kgrgreer: Is this supposed to be here?

test/src/core/Boot.js
```diff
@@ -801,6 +802,7 @@ describe('Slots', function() {
   it('unsubs manual listeners', function() {
     var last_args, last_value;
     var l = function() {
+      debugger;
```
kgrgreer: Is this supposed to be here?

src/core/tracing/Window.js
mdittmer, line 22: @kgrgreer, some context: This PR initially included an axiom called "overwrites" that extends a base class (including SUPER support), then registers the new class under the old class's name. After discussing the complexity of this solution, Adam and I agreed that "overwrites" was yet-another-way-to-do-inheritance that we don't want muddying the waters.

In the case of FObject, we switched to "refines" and used constants to store base method definitions (pseudo-SUPER-support). In the case of Window, Adam pointed out that overwriting the global context would probably be enough to capture calls to |merged| and |framed|; we can use SUPER as usual (and we do).

You correctly point out that this means when creating a new window, I have to know that tracing is installed and decide whether or not to use it (create a foam.core.Window or foam.core.tracing.Window). Should we make window tracing a system-wide all-or-nothing kind of thing by implementing the refines+pseudo-SUPER strategy used for FObject? @adamvy, what do you think?

src/core/tracing/Window.js
mdittmer: @kgrgreer, sure. Prefer 'FObject' or 'foam.core.FObject'?

test/src/core/Boot.js
mdittmer: @kgrgreer, no. Fixed.

test/src/core/Boot.js
mdittmer: @kgrgreer, no. Fixed.

src/core/FObject.js
```diff
@@ -169,36 +185,70 @@ foam.CLASS({
...
+      return this.hasKeyedListeners_(args) || this.hasExprListeners_(args);
```
kgrgreer: Isn't this inefficient because both methods walk the same list?  Could they be combined?

src/core/FObject.js
```diff
@@ -169,36 +185,70 @@ foam.CLASS({
...
+
+      return false;
+    },
+
+    function hasExprListeners_(args) {
+      /** Return true iff there are expr-matched listeners for message. **/
+      var listeners = this.getPrivate_('listeners');
+
+      for ( var i = 0; listeners; i++ ) {
+        if ( listeners.hasExprs ) {
```
kgrgreer, line 240: Expressions don't get to predicate on the payload, only the topic?  Is that what you intended?

src/core/FObject.js
kgrgreer: Yes.  Why do you need to convert it to an Array?

src/core/tracing/Window.js
kgrgreer: You can just require: [ 'FObject' ](since it's in foam.core. you don't need to specify it), then refer to it as 'this.FObject'.

mdittmer: Done.

src/core/FObject.js
mdittmer: I do not. Done.

src/core/FObject.js
mdittmer: @kgrgreer, could do. I separated the two for clarity. Each method is responsible for one traversal strategy. My intuition regarding performance was that expression-based listeners are sufficiently rare that invoking |hasExprListeners_| and having it fail fast is not very expensive.

What are your thoughts?

src/core/FObject.js
mdittmer, line 240: @kgrgreer, that is my intention, yes. It serves the only use case we have so far. To clarify, don't most listeners include matchable data in the topic? E.g., most common event topic looks like ['propertyChange', someName, someObj], so I can use .sub('propertyChange', 'foo', myExpr) for a filtered subscription to foo-property-changes.

PR #28, opened 2016-07-20
Add CSS axiom. No parser or prefixing for now.

bshepherdson:


src/core/Window.js
```diff
@@ -75,6 +77,10 @@ foam.CLASS({
     {
       name: 'console',
       factory: function() { return this.window.console; }
+    },
+    {
+      name: 'installedCSS',
```
kgrgreer: Add docs.

src/lib/u2.js
```diff
@@ -1437,3 +1437,52 @@ foam.CLASS({
...
+    },
+    {
+      name: 'name',
+      factory: function() { return 'CSS-' + this.$UID; }
+    }
+  ],
+
+  methods: [
+    function installInClass(cls) {
+      // There might be several CSS axioms on each class.
```
kgrgreer: Why would you need to install more than one?  I guess if you had a trait/mixin, then you could have more than one.

src/lib/u2.js
```diff
@@ -1437,3 +1437,52 @@ foam.CLASS({
...
+      if ( cls.create.__cssHandled ) return;
+
+      var oldCreate = cls.create;
+      var axiom = this;
+
+      cls.create = function(args, X) {
+        // Install our own CSS, and then all parent models as well.
+        var c = this;
+        while ( c && c.private_ ) {
+          if ( X.installedCSS[c.id] ) break;
```
kgrgreer: If a class has more than one create-decorating-axiom (like Singleton or Multiton, although, you probably wouldn't use those for Views, but something like them), then this could break.

Why not have each Axiom decorate create to install itself?  This would be the most common case and it would be the fastest and simplest.  If you have more than one axiom, then it gets decorated more than once.

kgrgreer: I would rather see the installed check and install method combined.  I shouldn't have to know the mechanism.  Like a maybeInstallCSS().  Or, just a registerClass(), which could do more than one thing, like install CSS and global document listeners or install gesture manager.  It could just be a isFirstInstanceInDocument() method, then the implementation could be entirely in the Axiom, rather than shared between the Axiom and Window.

kgrgreer:
Added design notes.

src/lib/u2.js
bshepherdson: Adam and I said the same thing. It's not hard to add, and a trait that adds particular `display: flex` et al might be really handy.

src/lib/u2.js
bshepherdson: 1. Suppose I did have two or three `CSS` axioms, and each decorated `create` to install only itself. The outermost one works fine, but then an inner one would see that this class has already been installed in the document, and not install further CSS. It seemed simpler to combine multiple CSS axioms into a single install step.
2. I can make the once-per-document logic less CSS-specific. The `Window` can have `isClassInstalled()` and `installClass()` methods, rather than the specific mechanism.

bshepherdson:
All addressed. (Show the last block of outdated comments.)

PTAL

kgrgreer:
LGTM

PR #29, opened 2016-07-21
CSS: Add the ^ shorthand. The CSS_NAME constant can be set to override it

bshepherdson:
With tests.


src/lib/u2.js
```diff
@@ -584,7 +584,7 @@ foam.CLASS({
         Without an extra, results in eg. 'foam-u2-Input-'.
         With an extra of "foo", results in 'foam-u2-Input-foo'.
       */
-      var base = this.CSS_CLASS || foam.String.cssClassize(this.model_.id);
+      var base = this.cls_.CSS_NAME || foam.String.cssClassize(this.cls_.id);
```
kgrgreer, line 587: cls_. not needed

PR #31, opened 2016-08-03
Add MD browser components, backed by Polymer

bshepherdson:
This is still pretty rough, but it was a working (not very well whitespaced) `DetailView`, with MD toolbars with all the various actions.

It comes with a basic demo port of the Chrome APIs app from FOAM 1, which is very, very incomplete.

### DIFF
--- a/src/com/chrome/apis/Controller.js
+++ b/src/com/chrome/apis/Controller.js
@@ -1,0 +1,110 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+foam.CLASS({
+  package: 'foam.u2.md',
+  name: 'DAOController',
+  extends: 'foam.u2.DAOController',
+
+
+  methods: [
+    function initE() {
+      // TODO(braden): Once toPropertyE is generalized to actions etc., use that
+      // here if practical.
+      var list = this.listFactory$f({
+        rowFactory: this.rowFactory,
+        data: this.data
+      });
+    },
+  ]
+});
+
+foam.CLASS({
+  package: 'com.chrome.apis',
+  name: 'Controller',
+  extends: 'foam.u2.Element',
+  requires: [
+    //'com.chrome.apis.Experiment',
+    'com.chrome.apis.Origin',
+    //'com.chrome.apis.User',
+    'foam.dao.CachingDAO',
+    'foam.dao.GUIDDAO',
+    'foam.dao.IDBDAO',
+    'foam.dao.MDAO',
+    'foam.u2.Stack',
+    'foam.u2.md.DAOController',
+    'foam.u2.md.StackView',
+    'foam.u2.md.Toolbar',
+    'foam.u2.md.ToolbarContainer'
+  ],
+
+  exports: [
+    'originDAO',
+    'stack'
+  ],
+
+  properties: [
+    {
+      name: 'originDAO',
+      factory: function() {
+        return this.GUIDDAO.create({
+          of: this.Origin,
+          delegate: this.CachingDAO.create({
+            of: this.Origin,
+            src: this.IDBDAO.create({ of: this.Origin }),
+            cache: this.MDAO.create({ of: this.Origin })
+          })
+        });
+      }
+    },
+    {
+      name: 'stack',
+      factory: function() {
+        return this.Stack.create();
+      }
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.cssClass('flex').cssClass('layout').cssClass('vertical');
+      var sv = this.StackView.create();
+      this.add(sv);
+      var inner = this.ToolbarContainer.create({
+        toolbar: this.Toolbar.create({
+          title: 'Chrome Experiments'
+        }),
+        body: this.DAOController.create({ data: this.originDAO })
+      });
+
+      //inner.cssClass('layout').cssClass('vertical');
+
+      this.stack.pushOnly(inner);
+    }
+  ],
+
+  axioms: [
+    foam.u2.CSS.create({
+      code: function CSS() {/*
+        body {
+          height: 100vh;
+          margin: 0;
+          padding: 0;
+        }
+      */}
+    })
+  ]
+});
--- a/src/com/chrome/apis/Controller.js
+++ b/src/com/chrome/apis/Controller.js
@@ -19,6 +19,13 @@ foam.CLASS({
   name: 'DAOController',
   extends: 'foam.u2.DAOController',
 
+  requires: [
+    'foam.u2.md.Button',
+    'foam.u2.md.DAOCreateController',
+    'foam.u2.md.DAOUpdateController',
+    'foam.u2.md.Toolbar',
+    'foam.u2.md.ToolbarContainer'
+  ],
 
   methods: [
     function initE() {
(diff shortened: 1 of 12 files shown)


src/com/chrome/apis/Controller.js
```diff
@@ -0,0 +1,299 @@
...
+  ],
+
+  methods: [
+    function initE() {
+      // TODO(braden): Once toPropertyE is generalized to actions etc., use that
+      // here if practical.
+      var list = this.listFactory$f({
+        rowFactory: this.rowFactory,
+        data: this.data
+      }, this);
```
adamvy-google: Isn't 'this' as the second argument the default?

src/com/chrome/apis/Controller.js
```diff
@@ -0,0 +1,299 @@
...
+        icon: 'add',
+        type: 'fab'
+      }).cssClass(this.myCls('fab'));
+      this.add(fab);
+    },
+
+    function buildCreateController(X) {
+      var toolbar = this.Toolbar.create();
+      var Y = X.createSubContext({ toolbar: toolbar });
+      return this.ToolbarContainer.create({
```
kgrgreer, line 56: Couldn't the ToolbarContainer create it's own Toolbar if not supplied?

src/com/chrome/apis/Controller.js
```diff
@@ -0,0 +1,299 @@
...
+      //var dv = this.DetailView.create({ of: this. of });
+      //dv.data$ = this.data$;
+      //this.add(dv);
+
+      this.toolbar.title$ = this.title$;
+      this.toolbar.addLeftAction(this.buildButton(this.CANCEL));
+      this.toolbar.addRightAction(this.buildButton(this.SAVE));
+    },
+
+    function buildButton(action) {
```
kgrgreer: Why not createButton()?  Isn't that more like our convention?

src/com/chrome/apis/Controller.js
```diff
@@ -0,0 +1,299 @@
...
+          this.innerData = nu.clone();
+        }.bind(this));
+      }
+    }
+  ],
+
+
+  methods: [
+    function initE() {
+      var dv = this.DetailView.create({ of: this. of });
```
kgrgreer: Remove the space between 'this.' and 'of'.

src/com/chrome/apis/Controller.js
```diff
@@ -0,0 +1,299 @@
...
+      dv.data$ = this.innerData$;
```
kgrgreer: This can be done in the create() call now.  Then you no longer need to specify 'of'.

src/com/chrome/apis/Controller.js
```diff
@@ -0,0 +1,299 @@
...
+  exports: [
+    'originDAO',
+    'stack'
+  ],
+
+  properties: [
+    {
+      name: 'originDAO',
+      factory: function() {
+        return this.GUIDDAO.create({
```
kgrgreer: We have EasyDAO now.

src/com/chrome/apis/polymer.js
```diff
@@ -0,0 +1,207 @@
...
+      factory: function() {
+        return [];
+      }
+    },
+    [ 'nodeName', 'paper-toolbar' ]
+  ],
+
+  methods: [
+    function initE() {
+      this.add(this.slot('leftActions_'));
```
kgrgreer: Why not just this.leftActions_$?

src/com/chrome/apis/polymer.js
```diff
@@ -0,0 +1,207 @@
...
+      this.start('span').cssClass('title').add(this.title$).end();
+      this.add(this.slot('rightActions_'));
+    },
+
+    function addLeftAction(button) {
+      this.leftActions_.push(button);
+      this.propertyChange.pub('leftActions_', this.leftActions_);
```
kgrgreer: This is not the correct signature for propertyChange events.  You should be pushing the slot, not the value itself.
If instead, you do this.leftActions_ = this.leftActions_.concat([button]); then the system will fire the proper event for you.

src/com/chrome/apis/polymer.js
```diff
@@ -0,0 +1,207 @@
...
+
+        var event = this.onKey ? 'keypress' : 'change';
+        var self = this;
+
+        e.onload.sub(function() {
+          var raw = e.el();
+          // Need to wrap this in a setTimeout, or raw.value isn't updated yet.
+          raw.addEventListener(event, function() {
+            self.setTimeout(function() {
+              e.data$.set(raw.value);
```
kgrgreer: Why not just e.data = raw.value?

src/com/chrome/apis/polymer.js
```diff
@@ -0,0 +1,207 @@
...
+        var e = X.lookup('foam.u2.PolymerWrapper').create({
+          nodeName: 'paper-input'
+        }, X);
+
+        e.attrs({ label: this.label$ });
+
+        var event = this.onKey ? 'keypress' : 'change';
+        var self = this;
+
+        e.onload.sub(function() {
```
kgrgreer: Why do you need to wait for onload to add event listeners?

src/lib/u2.js
```diff
@@ -47,6 +47,59 @@ foam.CLASS({
...
+        // Now call through to the original create.
+        return oldCreate.call(this, args, X);
+      };
+    },
+
+    function expandCSS(cls, text) {
+      /* Performs expansion of the ^ shorthand on the CSS. */
+      // TODO(braden): Parse and validate the CSS.
+      // TODO(braden): Add the automatic prefixing once we have the parser.
+      return text.replace(/\^/g,
```
kgrgreer: I don't think the '-' should be included if this isn't used as a prefix.

src/lib/u2/browser.js
```diff
@@ -95,6 +90,22 @@ foam.CLASS({
           .add(this.NEW_ITEM)
           .add(list)
           .endContext();
+    },
+
+    // These two functions are designed to be overridden by subclasses that
+    // need to wrap or replace the DAOUpdateController.
+    function buildUpdateController(X, data) {
+      return this.DAOUpdateController.create({
+        of: this.of,
```
kgrgreer: Is 'of' redundant?

src/lib/u2/md/DetailView.js
```diff
@@ -0,0 +1,63 @@
...
+        return of.model_.label;
+      }
+    },
+    [ 'showTitle', true ],
+    [ 'nodeName', 'div' ]
+  ],
+
+  methods: [
+    function initE() {
+      var self = this;
```
kgrgreer: Not used.

src/com/chrome/apis/Controller.js
bshepherdson, line 56: Except for the contexts, yes, that would work fine. The vital part is that the `toolbar` be in the context before the `this.SUPER(Y)` call, so that the buttons get added to the correct toolbar.

src/com/chrome/apis/Controller.js
bshepherdson: Done.

src/com/chrome/apis/Controller.js
bshepherdson: Whoops, done.

src/com/chrome/apis/Controller.js
bshepherdson: Much simpler now.

src/com/chrome/apis/Controller.js
bshepherdson: Done.

src/com/chrome/apis/polymer.js
bshepherdson: It makes the style checker angry. It thinks `foo_$` is fine, but `fooBar_$` is mixing `camelCase` with `snake_case`.

We can patch the style enforcer in time, but for now I was trying to keep these files style-clean.

src/com/chrome/apis/polymer.js
bshepherdson: Fancy. Done.

src/com/chrome/apis/polymer.js
bshepherdson: I'm doing more than adding event listeners. I'm also wiring up data and trying to set it.

The Polymer element isn't ready for that until it's really loaded in the DOM.

src/lib/u2.js
bshepherdson: This is how it worked in FOAM1 U2. It's hard to tell if a prefix is following, especially in the parser-integrated logic for this in FOAM1.

I can try to avoid the trailing `-` here if you think it's worthwhile.

src/lib/u2/browser.js
bshepherdson: Yes.

src/lib/u2/md/DetailView.js
bshepherdson: Removed.

src/com/chrome/apis/polymer.js
kgrgreer: I would rather patch the style-checker than our code.

src/com/chrome/apis/polymer.js
kgrgreer: How is that different than regular U2 Elements?  They also aren't ready to add listeners to until they've been loaded, so U2 remembers the listeners and then adds them once it is loaded.  I still don't see why this is needed.  Can't the PolymerWrapper handle this?

src/lib/u2.js
kgrgreer: I also complained about this in FOAM1 as well ;-).
Yes, please remove trailing -'s.

src/com/chrome/apis/polymer.js
bshepherdson: I've corrected the code, causing style violations for now.

We'll fix the checker separately.

src/com/chrome/apis/polymer.js
bshepherdson: This code is now much improved. The only really messy bit is the `setTimeout`, but that is required and I've included the reason why in a comment.

src/com/chrome/apis/polymer.js
```diff
@@ -0,0 +1,201 @@
...
+        e.attrs({
+          label: this.label$,
+          value: e.data$
+        });
+
+        var event = this.onKey ? 'keypress' : 'change';
+        var self = this;
+
+        e.addEventListener(event, function() {
+          // setTimeout is required because the event fires right before the
```
adamvy: Is the setTimeout only necessary due to the keypress event?

Perhaps there's a better event like 'input'.

src/lib/u2.js
bshepherdson: Done.

bshepherdson:
PTAL

bshepherdson:
Conflicts repaired. PTAL.

src/com/chrome/apis/bower.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "chrome-experiments-framework",
```
adamvy-google: The real name of the project is now "Origin Trials"

src/foam/u2/Element.js
```diff
@@ -52,6 +52,66 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
```
adamvy: Is this diff out of date?  There already exists a CSS axiom in Element.js, which looks similar to this except for expandCSS()

adamvy-google:
LGTM.

I think we should add a TODO to try to eliminate explicit usage of the context when Elements support .start(SomeModel, args)

src/com/chrome/apis/bower.json
bshepherdson: This is supposed to be the machine-readable name. I've changed it to `chrome-origin-trials`.

PR #101, opened 2016-09-14
Improved Olympics: Material Design, better layout/scrolling, etc.

bshepherdson:


src/com/google/foam/demos/olympics/Controller.js
```diff
@@ -30,8 +30,13 @@ foam.CLASS({
     'foam.u2.TableView'
   ],
```
adamvy-google, line 32: The package name of this class should be updated to reflect its new home in come.google.foam.demos

bshepherdson: Done.

src/com/google/foam/demos/olympics/Controller.js
```diff
@@ -81,34 +88,57 @@ foam.CLASS({
...
+    'scroller_',
+    'container_',
+    'scrollExtent',
+    'scrollHeight'
   ],
 
   methods: [
+    function init() {
+      this.SUPER();
+      if ( this.polymer ) {
```
adamvy-google: Rather than have this as a parameter, I think I'd rather see a separate sub-class called the PolymerOlympicDemo that makes the necessary registrations and modifications to be polymer.

The olympic demo is a key example that we want to point users to to base their work off of, so we should keep the code as straight forward and declarative as possible.

bshepherdson:
I've cut this apart into a `PolymerController` and `polymer.html`.

src/foam/u2/ActionView.js
```diff
@@ -63,6 +63,7 @@ foam.CLASS({
   properties: [
     'data',
     'action',
+    [ 'nodeName', 'button'],
```
adamvy-google: Keeping this in initE is actually the preferred style.

src/foam/u2/TableView.js
```diff
@@ -133,7 +138,26 @@ foam.CLASS({
 
         e = e.start('tr');
         for ( var i = 0 ; i < properties_.length ; i++ ) {
-          e = e.start('td').add(properties_[i].label).end();
+          var sorting$ = this.sortOrder$.map(function(prop, order) {
```
adamvy-google: At this point we should refactor the header into its own model, this is getting too large for my tastes.

src/foam/u2/ActionView.js
bshepherdson: Sorry, what? When did that happen?

More importantly, why? Isn't a property you can override more flexible?

adamvy-google: Happened one day in a design meeting.

I think both are equally flexibly, when you extend an element you're probably going to write a custom initE() so you can just set the nodeName there.

I personally like the style of initE() more or less completely defining the element.  It also keeps it more consistent when nesting elements.  If you have a big block of DSL, you get used to configuring elements though DSL, it becomes an impedance mis-match to configure your element through properties but configure all element you create in initE() through DSL.

That was one of my major pain points with foam U1, was all half the time you had to remember which things to do as properties, which as toHTML, which as toInnerHTML.  In U2 its all in initE()

It's why .setNodeName() was added to the DSL.

src/com/google/foam/demos/olympics/Controller.js
```diff
@@ -30,8 +30,13 @@ foam.CLASS({
     'foam.u2.TableView'
   ],
 
+  imports: [
+    'window'
+  ],
+
   constants: {
-    CSS_NAME: 'foam-u2-search-FilterController'
+    CSS_NAME: 'foam-u2-search-FilterController',
```
kgrgreer, line 38: What is this for?  Where is it used?

src/foam/u2/TableView.js
kgrgreer: Why the '$' suffixes?

src/foam/u2/TableView.js
```diff
@@ -133,7 +138,26 @@ foam.CLASS({
...
+            if ( ! order ) return '';
+            var desc = this.Desc.isInstance(order);
+            var baseOrder = desc ? order.arg1 : order;
+            if ( prop.name === baseOrder.name ) {
+              return desc ? this.Entity.create({ name: 'darr' }) :
```
kgrgreer: Shorter to put the '?' inside the Entity.create().

src/foam/u2/md/ActionView.js
```diff
@@ -0,0 +1,76 @@
...
+    {
+      name: 'label',
+      expression: function(action) { return action.label; }
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.cssClass(this.myCls())
+          .attrs({
```
kgrgreer: Better to move the isEnabled check below and just not add disabled: at all.  Less DOM and less work.

src/foam/u2/md/ActionView.js
```diff
@@ -0,0 +1,76 @@
...
+            disabled: ! this.action.isEnabled ? false :
+                this.action.createIsEnabled$(this.data$).map(function(e) {
+                  return e ? false : 'disabled';
+                })
+          })
+          .on('click', this.click)
+          .add(this.label$);
+
+      if ( this.action.isAvailable ) {
```
kgrgreer, line 45: See the recent changes to the regular ActionView. unavailable is now the CSS class we use to toggle and if isAvailable isn't set then you don't need to do anything.

src/foam/u2/md/ActionView.js
```diff
@@ -0,0 +1,76 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+foam.CLASS({
+  package: 'foam.u2.md',
+  name: 'ActionView',
+  extends: 'foam.u2.View',
```
kgrgreer: Make extend 'Element' instead then just add a 'data' property.

src/foam/u2/ActionView.js
bshepherdson: Done.

src/foam/u2/TableView.js
bshepherdson: Done.

src/com/google/foam/demos/olympics/Controller.js
bshepherdson: Combining all the CSS from the three models (`PolymerController`, `Controller` and `FilterController` under the same name. Otherwise it would only pick up the CSS for the childmost class, and not have the parents'.

I agree it sucks, but we don't have a better plan right now.

src/foam/u2/TableView.js
bshepherdson: Done.

src/foam/u2/md/ActionView.js
bshepherdson: Done.

src/foam/u2/md/ActionView.js
bshepherdson: Done.

src/foam/u2/md/ActionView.js
bshepherdson: Why? Isn't this a textbook example of a `View`?

src/com/google/foam/demos/olympics/Controller.js
adamvy-google: So is this a problem with CSS inheritence?  I'm PolymerController and I extend FilterController, but this.myCls() resolves to 'PolymerController' and the CSS classes installed by FilterController don't get applied to my dom nodes?

bshepherdson: Yes, that's correct according to my understanding. Let me check that my memory corresponds to reality... yes, it's badly broken without that.

src/foam/u2/md/ActionView.js
kgrgreer: It doesn't export 'data'. Or perhaps I should remove the export from View.

src/com/google/foam/demos/olympics/Controller.js
adamvy-google: Make a note on the next design meeting agenda for us to discuss this and try to find something cleaner or more predictable.

src/com/google/foam/demos/olympics/Controller.js
```diff
@@ -81,34 +88,43 @@ foam.CLASS({
...
+        extent$: this.scrollExtent$,
+        height$: this.scrollHeight$
       }, canvas);
       canvas.cview = scroller;
       this.scroller_ = scroller;
 
+      this.window.addEventListener('resize', this.onResize);
+
       parent.on('wheel', function(e) {
         scroller.value += e.deltaY > 0 ? 3 : -3;
```
adamvy-google: I think the returned value should be a function of deltaY.  If the browser batches up a bunch of wheel events and delivers one with a deltaY that is large, I'd expect my table to scroll an equal amount because that more cloesly matches the distance I actually turned the wheel.

adamvy-google:
lgtm

src/com/google/foam/demos/olympics/Controller.js
bshepherdson: Done. It's now converted to a number of rows and rounded up. It's relatively hard, at least with my mousewheel, to get more than the 3 rows at a time, but it is possible. This also makes scrolling by one row work.

src/foam/u2/TableView.js
```diff
@@ -158,9 +223,34 @@ foam.CLASS({
       name: 'onDAOUpdate',
       isFramed: true,
       code: function() {
-        this.data.select(foam.u2.TableBodySink.create({ properties_: this.properties_ }, this)).then(
-          function(a) { this.body = a.body; }.bind(this));
+        var dao = this.data;
+        if ( this.sortOrder ) {
```
adamvy-google, line 227: I feel like the more FOAM way of doing this, would be to make a property called orderedDAO, which is an expression of dao and sortOrder.

src/foam/u2/md/ActionView.js
adamvy-google: The non-md ActionView doesn't extend view.  These should be kept consistent.

src/foam/u2/TableView.js
adamvy-google: Ignore this for now.  If my DAOSlot work pans out we'll be replacing this anyways.

src/foam/u2/md/ActionView.js
bshepherdson: Done.

PR #108, opened 2016-09-16
TableView: Add Selection support

bshepherdson:


src/foam/u2/TableView.js
```diff
@@ -46,6 +49,14 @@ foam.CLASS({
   methods: [
     function addObj(obj) {
       var e = this.start('tr')
+          .on('click', function() {
+            this.tableView.selection = obj;
```
adamvy-google: Is there a way we could attach this listener only once at the table element?  Otherwise we have to make a DOM access for every row which could be slow when scrolling.

bshepherdson: In principle, yes. In practice, scrolling is still instant, so I'm not sure it's worth the complexity of determining the target row (and corresponding object) from the click event.

adamvy-google: Is scrolling instant on mobile/arm chromebooks?

bshepherdson: It's fast on my phone (although since we're binding directly to the wheel events, it doesn't work for touch except on the scrollbar).

I don't have an underpowered ARM Chromebook handy to try it on. It's acceptable on this underpowered Intel one, though it could still definitely be faster.

kgrgreer: Each extra listener is 11k. So if you're scrolling 100 rows at 60fps that's ~ 10k_100rows_60frames = 60000 or 60M / sec

bshepherdson: You can't scroll anywhere near that fast in practice, but your point is taken.

If we're concerned about generating garbage here, I'll write the grungy logic to figure out which row was clicked and which object it corresponds to, and use that instead.

kgrgreer: I wouldn't hold up the CL for this, but add a TODO to try using a single listener for performance/memory.

bshepherdson: Already done, PTAL at the new code.

adamvy-google:
LGTM

src/com/google/foam/demos/olympics/Controller.js
```diff
@@ -126,16 +126,15 @@ foam.CLASS({
         scroller.value += negative ? -rows : rows;
       });
       this.container_ = parent.start().cssClass(this.myCls('table-container'));
-      this.container_.start(this.TableView, {
+      this.table = this.container_.start(this.TableView, {
```
adamvy-google: You can use the .call(function() { self.table = this; }) style as well here.

src/foam/u2/search/FilterController.js
```diff
@@ -248,11 +251,14 @@ foam.CLASS({
...
-      parent.start(this.TableView, {
+      // NB: This function should render the table into the parent Element, and
+      // should also set this.table to be the TableView.
+      this.table = parent.start(this.TableView, {
         data$: this.filteredDAO$,
         editColumnsEnabled: true,
         title$: this.title$
-      }).end();
+      });
+      this.table.end();
```
kgrgreer: Why is this line needed?

src/com/google/foam/demos/olympics/Controller.js
bshepherdson: I've restructured this whole block to be fully fluent, using the `.call()` style.

src/foam/u2/search/FilterController.js
bshepherdson: I've restructured this to use the `.call()` style instead.

bshepherdson (commented):
PTAL

adamvy-google:
LGTM

PR #146, opened 2016-10-05
Use the FilterController for the Origin Trials app

bshepherdson:


src/com/chrome/origintrials/model/Application.js
```diff
@@ -56,7 +59,8 @@ foam.CLASS({
     {
       class: 'Boolean',
       name: 'agreedToTerms',
-      label: 'I accept the Google Chrome Terms of Service',
+      //label: 'I accept the Google Chrome Terms of Service',
+      label: 'Terms of Service',
```
adamvy-google: Can we fit something about "agreed" in the label?  Something like "Agreed to terms?"

adamvy-google:
LGTM

src/com/chrome/origintrials/model/Application.js
bshepherdson: Will do.

PR #151, opened 2016-10-11
TableView selection support

bshepherdson:
Uses this new support in the Origin Trials as well.

### DIFF
--- a/src/foam/u2/TableView.js
+++ b/src/foam/u2/TableView.js
@@ -63,15 +63,22 @@ foam.CLASS({
   extends: 'foam.u2.Element',
 
   requires: [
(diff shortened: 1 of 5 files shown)


src/foam/u2/TableView.js
```diff
@@ -63,15 +63,22 @@ foam.CLASS({
   extends: 'foam.u2.Element',
 
   requires: [
+    'foam.mlang.predicate.And',
```
kgrgreer, line 66: Why not just implement Expressions?

bshepherdson: I had forgotten about that. I won't change it here, but I've added a `TODO`.

Issue #163, opened 2016-10-27
Index multi-property ordering or compound comparator


jacksonic:
Sort by PropA, then by PropB, etc.

Current TreeIndex select/selectReverse do not support this.

jacksonic:
Working impl: Comparator mlangs supply a sort-direction chain, that chain determines which way select() at a given level runs. The next tree down the index takes the next direction indicator in the chain. selectReverse() is not needed on Index anymore, but tree nodes still use it.

Indexes also can report if they and their tails are capable of sorting by a given ordering, cleaning up TreeIndex's plan() where it figures out what to sort.

Issue #128, opened 2016-09-20
AutoIndex predicate support


jacksonic:
Build indexes based on incoming predicate in addition to existing ordering. #126 for OR support is required.

Also check whether multiple column ordering is supported.

jacksonic:
Working, full testing pending. See AutoIndex branch for current state.

jacksonic:
Current state of the art for nested auto indexes on AutoIndex2.
This will be benchmarked to determine the potential performance impact.

Issue #109, opened 2016-09-16
Tree index: remove shared properties


jacksonic:
Remove nullNode/treeNodeFactory and pass as parameters

jacksonic:
AutoIndex PR will fix

Issue #155, opened 2016-10-17
Remove uses of obj.id in MDAO/indexes


jacksonic:
Calling obj.id on MultiPartIDs will cause object creation as the array is populated, so use ID.compare and don't get the ID value unless necessary.

i.e. avoid: 
- mapStore[obj.id] = obj;
- foam.util.compare(obj.id, key);

Check places like:
- [ ] MDAO find()
- [ ] index bulkLoad()
- [ ] TreeIndex compare fn, bulkLoad, put's key

PR #169, opened 2016-11-03
Auto index, the grand merge

jacksonic:
New features, in order as you read through the github diff:

- Index split into Index (definition of index structure) and IndexNode (instances that hold data)
   - IndexNode is lightweight, uses no context, requires, or imports. Has access to 'creator' to get those things.
   - IndexNode.select() gains a cache argument, for storing data that only exists for the life of the query. AltIndex uses it to store the chosen delegate.
- MDAO gains ability to DNF transform the predicate and handle an OR using MergePlan.
- AATree now supports selecting with multiple orderings by popping the topmost order and checking the direction to scan.
   - getAll() supports TreeIndex's CONTAINS check.
- AltIndex now has separate notions of the indexes that define the sub-indexes vs. the many instances of AltIndexNode holding actual instances of those sub-indexes, possibly incomplete.
   - addIndex() registers the new index with the creator, so a complete list of potential sub-indexes is available for estimate(), but only populates the particular AltIndexNode where the add was requested
   - pickDelegate() picks and caches the best index to use for a select scan with ordering, using static method Index.estimate() to pick the best sub-index. If the index has not been built yet for each node that is hit, it will be built during select() execution.
- AutoIndex now supports building full indexes for ordering and predicates.
   - Jank is allowed when adding root indexes or large sub-indexes
   - predicate.toIndex() and order.toIndex() both asked for their best index, estimate() from each and the current index compared to choose whether to build anything
- Index interface changes:
   - estimate() added. Runs as a static method, given a size and query, how well this index type would generally perform. Used by AutoIndex for picking, AltIndex for picking, TreeIndex to find tail's ordering support.
   - toPrettyString() returns a readable printout of the index structure
- IndexNode interface (takes on the data operations of Index):
   - lightweight by convention (no imports, requires, etc. not enforced. Accesses these things through a creator property)
   - plan() gains a "root" argument, the root index instance. This gives access to the entire index, including the total size of the DAO for autoindex decision making.
   - select() gains a cache parameter, to pass temporary query state along.
   - selectReverse() no longer required since select() determines its scan direction.
- MergePlan merges potentially duplicate results from an OR query and sorts them into the result sink. Future optimizations could speed up the sorting stage.
- ProxyIndex manually proxies the lightweight index classes
- TreeIndex
   - isOrderSelectable() checks the order property and tail estimate, if the tail estimate is higher than sort cost, the tail is sorting the hard way and this index isn't good for ordered scanning.
   - estimate() mirrors the checks in plan(), to provide an estimate of how the given ordering and predicate would be handled by this index. Node count is picked arbitrarily, and certain parameters in this method could be tweaked if the estimate is not accurate relative to other index types.
   - plan() gains the CONTAINS(_IC) handler
- mlangs
   - toIndex() calls through to the Property's toIndex() in most cases, except AND where args are sorted by their estimates. This means the fastest indexes are executed first, restricting the result set quickly. (i.e. EQ first, then CONTAINS, then LT scan). AND also accepts a 'depth' parameter to limit how many subindexes it chains.
   - toDisjunctiveNormalForm() creates a new predicate in DNF, or returns the original if already DNF or OR-less.
   - ordering.Comparator provides:
      - orderTail() to get the next (if any) ordering to sort by
      - orderPrimaryProperty() extracts the first property being sorted
      - orderDirection() returns 1 to indicate ascending order, -1 for descending. 0 would imply no ordering, but is not used
   - ThenBy allows multiple property ordering, THEN_BY(propA, propB), or THEN_BY(a, THEN_BY(b, c)).

- And a bunch of tests and benchmark updates.


jacksonic:
New idea to simplify mlang order.orderDirection(): eliminate Direction class, just return 1, -1. Select() can pass the order in as before, and move index-specific state to a new argument.

jacksonic:
Open meeting TODOs in recent commits

src/foam/dao/index/Plan.js
```diff
@@ -166,3 +166,79 @@ foam.CLASS({
...
+    /**
+      Executes sub-plans, limiting results from each, then merges results,
+      removes duplicates, sorts, skips, and limits.
+    */
+    function execute(promise, sink, skip, limit, order, predicate) {
+      // TODO: Investigate pre-sorted results from subqueries being
+      //   zipped together quickly
+
+      var resultSink = this.DedupSink.create({
+        delegate: this.decorateSink_(sink, skip, limit, order)
```
kgrgreer, line 198: This is going to be very inefficient when used with skip and limit since it will require visiting and the resorting all data, not just the data in the skip->skip+limit range. The data is already sorted properly, so you shouldn't need to sort it again. If instead of executing the subPlans in series, you instead execute them all in parallel and merge the results, then you won't need to resort or visit any data after skip+limit.

test/benchmarks/Index.js
```diff
@@ -95,9 +65,21 @@ describe("Index benchmarks", function() {
...
+
     subjectDAO = foam.dao.MDAO.create({ of: Subject })
-      .addPropertyIndex(Subject.INT_P, Subject.BOOL_P)
-      .addPropertyIndex(Subject.STRING_P, Subject.INT_P)
+      //.addPropertyIndex(Subject.INT_P)
+      //.addPropertyIndex(Subject.INT_P, Subject.BOOL_P)
+      .addIndex(
+        foam.dao.index.AltIndex.create({
+          delegates: [ idx1, idx2 ],
+          delegates: [ idx1.createNode(), idx2.createNode() ]
```
kgrgreer: Why is delegates: specified twice?

src/foam/dao/index/AltIndex.js
```diff
@@ -32,21 +41,79 @@ foam.CLASS({
...
   methods: [
-    function addIndex(index) {
-      // Populate the index
-      var a = foam.dao.ArraySink.create();
-      this.plan(a).execute([Promise.resolve()], a);
 
-      index.bulkLoad(a);
-      this.delegates.push(index);
+    /** Returns smallest estimate from the delegates */
+    function estimate(size, sink, skip, limit, order, predicate) {
```
kgrgreer, line 58: Why do we need estimate() in addition to plan()?

src/foam/dao/index/Plan.js
jacksonic: Yes, the TODO indicates this improvement should be implemented. I'll add an issue for it

test/benchmarks/Index.js
jacksonic: That's a refactoring bug, will fix

src/foam/dao/index/AltIndex.js
jacksonic: estimate() is the 'static' version of plan, that estimates performance without having any real data. Estimates should be relatively accurate when compared to each other, but won't be directly comparable to plan.cost since that will be much more accurate.

src/foam/dao/index/AATree.js
```diff
@@ -272,32 +260,40 @@ foam.CLASS({
...
         return;
       }
 
-      this.left.select(sink, skip, limit, order, predicate);
-      this.value.select(sink, skip, limit, order, predicate);
-      this.right.select(sink, skip, limit, order, predicate);
+      this.left.select(sink, skip, limit, order, predicate, cache);
+
+      this.value.select(sink, skip, limit,
+        order && order.orderTail(), predicate, cache);
```
kgrgreer, line 275: Was this a bug in previous versions?

src/foam/dao/index/AATree.js
```diff
@@ -272,32 +260,40 @@ foam.CLASS({
...
         skip[0] -= this.size;
         return;
       }
 
-      this.right.selectReverse(sink, skip, limit, order, predicate);
-      this.value.selectReverse(sink, skip, limit, order, predicate);
-      this.left.selectReverse(sink,  skip, limit, order, predicate);
+      this.right.selectReverse(sink, skip, limit, order, predicate, cache);
+
+      // select() will pick reverse or not based on order
```
kgrgreer, line 292: How will it know that this query is reversed?

src/foam/dao/index/AATree.js
jacksonic: Yes, selecting with multiple properties ordered never worked correctly. Issue #163

src/foam/dao/index/AATree.js
jacksonic: The comparator specified will give a 1, -1 direction indicator. Select() moves to the next ordering down the chain and passes that on to subsequent tail indexes.

src/foam/dao/index/AutoIndex.js
```diff
@@ -20,57 +20,191 @@
...
   ],
 
   methods: [
-    function put() { },
+    function estimate(size, sink, skip, limit, order, predicate) {
+      return this.delegate.estimate(size, sink, skip, limit, order, predicate);
+    },
+
+    function toPrettyString(indent) {
+      var ret = "";
```
kgrgreer: Google and FOAM standard is to use single quotes for string literals unless string contains a single quote.

src/foam/dao/index/Index.js
```diff
@@ -15,52 +15,114 @@
...
+  many IndexNode instances:
+
+<pre>
+                 1---------> TreeIndex(id)
+  MDAO: AltIndex 2---------> TreeIndex(propA) ---> TreeIndex(id) -------------> ValueIndex
+        | 1x AltIndexNode    | 1x TreeIndexNode    | 14x TreeIndexNodes         | (DAO size)x ValueIndexNodes
+           (2 alt subindexes)     (14 nodes)             (each has 0-5 nodes)
+</pre>
+  The base AltIndex has two complete subindexes (each holds the entire DAO).
+  The The TreeIndex on property A has created one TreeIndexNode, holding one tree of 14 nodes.
```
kgrgreer: Remove extra "The".

src/foam/dao/index/Index.js
```diff
@@ -15,52 +15,114 @@
...
-  index multiplexer, or any other MDAO select() assistance class. */
+*/
 foam.CLASS({
   package: 'foam.dao.index',
   name: 'Index',
 
+  properties: [
+    {
+      class: 'Class',
+      name: 'tailClass',
```
kgrgreer: Add docs to explain what a tailClass is. What is a tailClass?

src/foam/dao/index/Index.js
```diff
@@ -15,52 +15,114 @@
...
+  The Index interface for an ordering, fast lookup, single value,
+  index multiplexer, or any other MDAO select() assistance class.
 
-/*
- * TODO:
- *  update(oldValue, newValue)
- *  reuse plans
- *  add ability for indices to pre-populate data
- */
+  Each Index subclass also defines an IndexNode class. Index defines
```
kgrgreer, line 22: Is this over specified? Do we need to know that Indices have nodes? Isn't this an implementation detail?

src/foam/dao/index/Index.js
```diff
@@ -15,52 +15,114 @@
...
-    function create(args) {
-      var c = Object.create(this);
-      args && c.copyFrom(args);
-      c.init && c.init();
-      return c;
+
+    function estimate(size, sink, skip, limit, order, predicate) {
+      /** Estimates the performance of this index given the number of items
+        it will hold and the planned parameters. */
+      return size * size; // n^2 is a good worst-case estimate by default
```
kgrgreer, line 66: Is this ever used? When would it ever be worse than nlg(n)?

src/foam/dao/index/AATree.js
```diff
@@ -198,7 +186,7 @@ foam.CLASS({
       return s.split(locked).skew(locked);
     },
 
-    function removeKeyValue(key, value, compare, locked) {
+    function removeKeyValue(key, value, compare, locked, nullNode) {
```
kgrgreer, line 189: Wouldn't it be more flexible to pass the index itself rather than nullNode?

src/foam/dao/index/AltIndex.js
```diff
@@ -32,21 +41,79 @@ foam.CLASS({
...
+
+    function toString() {
+      return 'Alt([' + (this.delegates.join(',')) + '])';
+    },
+  ]
+});
+
+foam.CLASS({
+  package: 'foam.dao.index',
+  name: 'AltIndexNode',
```
kgrgreer, line 86: Why does an AltIndex need nodes?

src/foam/dao/index/Index.js
jacksonic: Refactoring name change missed: should be 'nodeClass', docs added (it's the type of nodes this index creates in createNode())

src/foam/dao/index/Index.js
jacksonic: Every index has nodes, since any index may be the tail of a Tree (or other type that splits into many sub-nodes). Indexes always worked this way, but now the shared definition object exists as Index, and the existing data-searching/manipulating methods moved to IndexNode.

src/foam/dao/index/Index.js
jacksonic: It's only there as a default if you write an index and don't specify. Could be infinity, or we could throw if you try to leave it out.

jacksonic: It does need to be worse than sorting, since if you're sorting you should specify that you are doing so with an accurate estimate. An alternative index that really does sort, and accurately reports it, should win if estimates are compared.

src/foam/dao/index/AATree.js
jacksonic: This was an efficiency optimization you wanted a long time ago (passing things as args rather than shared state). There might be a performance impact from the extra lookup, but as nullNode isn't used all that often compared to the other args I'll have to profile and see. Either way, this API is very tightly coupled to the way TreeIndex works so flexibility gains will be limited.

src/foam/dao/index/AltIndex.js
jacksonic: Every index needs nodes. In the simple case you could assemble an index that would have one chain of Indexes, with one single instance of IndexNode for each.

If a TreeIndex appears in the chain, then each TreeNode will store a separate IndexNode for the tail index. So in that case a Tree -> Alt setup would have one TreeIndexNode holding many AltIndexNodes, each of which may or may not have the same set of delegates.

The AltIndex instance that manages all those nodes knows about every possible delegate type, and can estimate which one to use (and build, if missing from a particular delegate node) for selects.

src/foam/dao/index/Index.js
```diff
@@ -15,52 +15,114 @@
...
+  ]
+});
+
+/**
+  The index tail interface represents a piece of the index that actually
+  holds data. A tree will create a tail for each node, so one Index will
+  create many tail instances, each operating on part of the data in the DAO.
+
+  For creation speed, do not require or import anything in a tail class.
+  Use the 'creator' property to access requires and imports on the
```
kgrgreer, line 90: Why do we split creators out of the Index?

jacksonic: The creator is the Index instance that both defines the type and parameters for that part of the index, and contains any shared state or context access the lightweight nodes need.

So for nodes, the creator reference lets them access requires without themselves triggering a context build on themselves, and access estimate() or other methods that don't depend on the actual data in the index.

MyIndex.create() // allowed to do heavy setup operations
MyIndex.createNode() --> MyIndexNode.create() // not allowed to be slow
myNodeInstance.creator.ThingRequired // access to things that were set up by MyIndex 

(this avoids the Progenitor implementation that previously modelled and enforced the lightweightness of nodes)

jacksonic: Noticed the docs here are still referencing tail, will fix

jacksonic: From a user's point of view, they will only ever handle Index instances. Constructing a chain of Indexes and handing that to MDAO is the deepest manual assembly developers typically need (and should be rare given you can just toIndex() a predicate, ordering, or property).

Only when developing a new index, which requires knowledge of how they are optimized, do you need to know how the nodes work. We need a guide document outlining the whole process (there's obviously some clarity missing from the existing docs in code)

jacksonic:
Fixes #163, Fixes #128, Fixes #109, 
Unblocks #135, #155

PR #181, opened 2016-11-22
Fixes for all Missing Context warnings, mostly tests but some core fixes

jacksonic:
Some core changes where context was not passed through
Enum is using default foam.__context__ for created values
Tests look a bit messier, but could be refactored with a create() helper

FoamByExample test disabled pending major refactor where test failures and missing context warnings can be fixed


jacksonic:
Rolled back for anything inside the "foam" root package. This means DAOs don't give a missing context warning.

PR #190, opened 2016-12-14
TypeChecking on LIBs active

jacksonic:
Memoize0/1 retains the original function's toString() in order to support type checking on memo'd functions.

### DIFF
--- a/src/foam/core/debug.js
+++ b/src/foam/core/debug.js
@@ -331,8 +331,6 @@ foam.__context__ = foam.debug.Window.create(null, foam.__context__).__subContext
 
 
 
-/** The types library deals with type safety. */
-// ???: Should this go in foam.Function along with formalArgs() and be renamed just functionArgs?
 foam.LIB({
   name: 'foam.Function',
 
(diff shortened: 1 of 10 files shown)


src/foam/core/debug.js
```diff
@@ -368,44 +393,78 @@ foam.LIB({
         }
 
         // If nothing threw an exception, we are free to run the function
-        var retVal = fn.apply(this, arguments);
+        var typeCheckerVal = fn.apply(this, arguments);
 
-        // check the return value
-        if ( args.returnType ) args.returnType.validate(retVal);
+        // check the typeCheckerurn value
```
kgrgreer: What is 'typeCheckerurn'?

src/foam/core/stdlib.js
```diff
@@ -381,12 +371,9 @@ foam.LIB({
 
     {
       name: 'camelize',
```
kgrgreer: Move out of stdlib.

jacksonic: Done on master

src/foam/core/debug.js
jacksonic: fixed... search and replace typp

PR #200, opened 2017-01-09
Float parser

jlhughes:


adamvy-google:
This also overlaps with the Long parser pull request.  Maybe we'll just batch all these changes together into that request?

jlhughes:
Unable to create multiple pull-request on same fork, so reviewing in LongParser.

PR #207, opened 2017-01-17
Sort keys in foam.core.JSON to get predictable cross-platform output

mdittmer:


kgrgreer:
Why don't we already get predictable cross-platform output? Isn't the order dependent on the models, which are the same for all platforms?

mdittmer:
Key order is notoriously un-spec'd on the web. I don't know whether all major browsers now, by implementation accident, do the same thing, but it certainly didn't used to be the case that they did.

kgrgreer:
Do you have a use-case where you specifically need this? It seems like people shouldn't be writing software which is fragile to the ordering of something that's specified to be unordered (although, V8 does order).

mdittmer:
We have checked-in mini-databases that contain pretty-printed JSON (that we plan to move to pretty-printed FON). Re-gathering said data and writing it out should result in no diff, regardless of the browser/platform used to re-gather the data.

adamvy-google:
The current output will follow the order that the properties are declared
in the model, which will keep the output stable if you're not re-arranging
properties in the model.

adamvy-google:
Oh this is not a model you're serializing, just some plain javascript
object.

Even still, it seems the only benefit of keeping the output stable is that
your diff's will look nicer if you are checking these values into source
control.  I'm not sure that's worth it.

mdittmer:
I discussed this with Kevin offline; he feels okay with this on a default-false config parameter. I've pushed a change to reflect this strategy.

PR #208, opened 2017-01-17
Listeners return a value, so that the value can be used...

mdittmer:
… in the case of synchronous listeners


kgrgreer:
I think this could be a bad idea since it would break for framed or merged listeners and would prevent you from making a listener async without breaking things.
What do you need this for (DOM listeners?)?

mdittmer:
Yes. Listeners returning values is used in SMW.

PR #210, opened 2017-01-17
stdlib: add 'localMMethod' for package-local model name support

mdittmer:


src/foam/core/stdlib.js
```diff
@@ -693,6 +693,15 @@ foam.LIB({
         }
         return ( type[uid] || opt_defaultMethod ).apply(this, arguments);
       };
+    },
+    function localMMethod(packageName, localMap,
```
kgrgreer, line 721: Docs? What does this do? What is it for?

mdittmer: Pushed documentation that contains an example to clarify use case.

mdittmer:
That makes sense, especially since mmethod design is not finalized yet.

PR #213, opened 2017-01-19
Long parser

jlhughes:
Includes changes IntParser which now fails on Integer MAX_VALUE overflow. 
LongParser was moved before IntParser in AnyParser to support predicates with a 'Long' value.


src/lib/JavaClass.js
```diff
@@ -998,8 +998,55 @@ foam.CLASS({
...
+    ['javaType', 'long'],
+    ['javaInfoType', 'foam.core.AbstractLongPropertyInfo'],
+    ['javaJSONParser', 'foam.lib.json.LongParser']
+  ],
+
+  methods: [
+    function createJavaPropertyInfo_(cls) {
+      var info = this.SUPER(cls);
+      var m = info.getMethod('cast');
+      m.body = 'return ( o instanceof Long ) ?'
```
adamvy-google: Shorter to just check o instanceof Number, then o.longValue()

src/lib/JavaClass.js
```diff
@@ -1020,6 +1067,24 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
+  refines: 'foam.core.Date',
+  properties: [
+    ['javaType', 'int'],
```
adamvy-google: This should probably be a long.  int is too short for some timestamps.

src/lib/box.js
```diff
@@ -1475,7 +1475,7 @@ foam.CLASS({
         req.then(function(resp) {
           return resp.payload;
         }).then(function(p) {
-          this.me.send(foam.json.parse(foam.json.parseString(p, null, this)));
+          this.me.send(foam.json.parse(foam.json.parseString(p, this, this)));
```
adamvy-google: This line should actually just read

this.me.send(foam.json.parseString(p, this));

since parseString() calls parse() internally.  When I wrote this originally parseString() just did the eval() portion and didn't call parse(), or at least that's what I thought when I first wrote it.

Doing this should also remove the need for your nested try{}catch{} earlier.

src/foam/core/JSON.js
```diff
@@ -481,15 +481,26 @@ foam.LIB({
     },
 
     function parseString(jsonStr, opt_ctx) {
-      return this.parse(eval('(' + jsonStr + ')'), undefined, opt_ctx);
+        console.debug('parseString: '+jsonStr);
```
adamvy-google: This shouldn't be necessary, the call in HTTPBox to foam.json.parse() is wrong, and if we fix that, then we shouldn't need this change.

src/foam/u2/FoamTagLoader.js
```diff
@@ -57,4 +57,4 @@ foam.CLASS({
   ]
 });
 
-foam.u2.FoamTagLoader.create();
+//foam.u2.FoamTagLoader.create();
```
adamvy-google: Please exclude this from this pull request.

Also what do you need this change for?  Perhaps this file should receive it's own flag in files.js so we can avoid loading it in all contexts except a visible web page.

adamvy-google (changes requested):

src/lib/box.js
adamvy-google: I've gone ahead and made this change upstream.

src/foam/u2/FoamTagLoader.js
jlhughes: re: FoamTagLoader
Users/jhughes/workspace/jbbmobile/repos/sefoam-quote/build/foam-java-bin.js:32589
      this.window.addEventListener('load', this.onLoad, false);
                  ^

TypeError: this.window.addEventListener is not a function

flags: ['web']

other 'web' files are needed, so you would need an additional flag.

adamvy-google:
Feel free to put foam.u2.FoamTagLoader on an additional flag :)

java_src/foam/lib/json/FloatParser.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+        decimalFound = true;
+        n.append(c);
+      } else {
+        break;
+      }
+      ps = ps.tail();
+    }
+
+    return ps.setValue(n.length() > 0 ? Float.valueOf(n.toString()) : 0.0);
```
adamvy-google, line 43: If n.length() == 0, should we really treat that as successfully parsing 0?  Seems like we should just fail.

src/files.js
```diff
@@ -168,7 +172,7 @@ FOAM_FILES([
   { name: "foam/u2/search/TextSearchView", flags: ['web'] },
   { name: "foam/u2/stack/Stack", flags: ['web'] },
   { name: "foam/u2/stack/StackView", flags: ['web'] },
-  { name: "foam/u2/FoamTagLoader", flags: ['web'] },
+  { name: "foam/u2/FoamTagLoader", flags: ['loader'] },
```
adamvy-google, line 175: Please also enable the loader flag in src/foam.js the same way that the web flag is enabled.

java_src/foam/mlang/predicate/Has.java
```diff
@@ -0,0 +1,19 @@
+// DO NOT MODIFY BY HAND.
```
adamvy-google: Don't check in generated code.

src/foam/core/types.js
```diff
@@ -89,6 +89,12 @@ foam.CLASS({
 
         return foam.Date.compare(o1, o2);
       }
+    },
+    {
+      name: 'toJSON',
+      value: function(o) {
+        return o && o.toISOString();
```
adamvy-google, line 96: So I realize that the DateParser I wrote is trying to parse these strings, but that was written to interop with an existing system I couldn't change.

It would be better I think to format dates just as an epoch timestamp with o.getTime(), and parse it the same way.  This is simpler and more robust.

I would rather have this code do return o && o.getTime(), and then update the Date property to have a different parser that parses a long int and converts it to a date.

jlhughes:
Pull Request updated. 
The Pull Request, for reasons only known to git, now also includes changes made on foam master than I never modified.

adamvy-google:
I noticed that as well, not sure why github is struggling so much.

Is this ready to be reviewed again?

jlhughes:
Yes, ready to be reviewed again.

jlhughes:
Adam
If it helps, this is a list of the files I'm explicitly aware of changes.
Anything else is a result of the many merges.
Joel

java_src/foam/core/AbstractLongPropertyInfo.java
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-bab62dea580d2211ae3b048321b3e0e4>

java_src/foam/lib/json/AnyParser.java
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-1af03dca8e2c43c343650621eb80689e>

java_src/foam/lib/json/FloatParser.java
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-54ec9ae2963b9ba07649559f30f66ed5>

java_src/foam/lib/json/IntParser.java
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-4c4b2a4c5969c3035f2eb64132942ce0>

java_src/foam/lib/json/LongParser.java
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-fea5ade8d5027d9c1e06809a81f6162e>

java_src/gen.sh
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-e1762e8ed456793919a500db7ad2bab1>

src/files.js
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-ecdbeb0b77447177e8165fbaf62b5a09>

src/foam/core/JSON.js
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-bf6b8c3849eeb3ae35d4149d3b8ac21c>

src/foam/core/types.js
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-99ce6310b435d998b4e86b010fc30033>

src/lib/JavaClass.js
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-45f2aa1e421069bd162f660f38e7d801>

src/lib/box.js
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-1026476588fe09153344c0f56172ca9d>

tools/genjava.js
<https://github.com/foam-framework/foam2-experimental/pull/213/files#diff-1b61b929e30c71a1b909025bc11a2a55>

adamvy-google:
The reason the pull request appears to be including changes that are
already present in upstream FOAM is because of this commit here:

https://github.com/jlhughes/foam2-experimental/commit/51017ce12bdb93c5323e7ac5ea39af56dd19f4bf

This commit is not actually a merge commit, even though the message implies
it is.  It only has one parent commit, unlike other merge commits which
generally have 2.  What I assume happened is you started a merge, but
somehow reset the merge status, maybe via trying to fix conflicts, and then
when you committed it, it no longer commit as a merge commit.

As a result it thinks you wrote a bunch of lines that actually came from
changes I made.

I would prefer not to merge this broken history to our repo, partly because
now git blame will blame you as the author for a large chunk of java code
that I wrote :).


You can do the following, assuming that your pull request is on a branch
named LongParser, and you have a remote named "upstream" that points to the
foam2-experimental repo.

$ git checkout LongParser
$ git fetch upstream
$ git diff upstream/master > 001_LongParser.patch
$ git checkout -b LongParser upstream/master
$ path -p1 -i 001_LongParser.patch
$ git commit -am "Various fixes for upstream FOAM."

And now you have a branch named LongParser2 which contains the actual
things you've.  Set that up as a pull request and I'll use that.

src/foam/u2/PopupView.js
```diff
@@ -53,7 +53,7 @@ foam.CLASS({
       var parent   = this.parentNode;
       var close    = function() {
         self.remove();
-        bg.remove();
+        //bg.remove();
```
adamvy-google, line 56: Rather than removing the background for all users of the PopupView, if you don't like the background, can we make it a configurable parameter of the PopupView?  Then your code can just say "PopupView.create({ modal: true })" or "fullscreen: true" or "overlay: true"

I think overlay: true works best.

src/lib/java_hacks.js
```diff
@@ -0,0 +1,32 @@
+foam.CLASS({
```
adamvy-google, line 1: I would rather not add a file named java_hacks.js

Additionally are these changes necessary?  I fixed the issue with booleans not serializing properly and mentioned it to you on Jan 11.

tools/genjava.js
```diff
@@ -75,7 +80,7 @@ var classes = [
   'foam.dao.SkipSink',
   'foam.dao.FlowControl',
   'foam.mlang.order.Comparator',
-  'foam.mlang.sink.Count'
+  'foam.mlang.sink.Count',
```
adamvy-google, line 83: Leave out the trailing comma.

adamvy-google (changes requested):

jlhughes:
Hey Adam
That didn't work at all, unfortunately.
Still has some of your changes and is missing most of mine.

Other thoughts on moving forward?
I could try and create individual patch files for each model and we forgo
the pull-request process.

Joel

jlhughes:
Replaced by PullRequest LongParser2

PR #214, opened 2017-01-19
mlang predicate toSQL support

jlhughes:
I don't know how to create branches and pull requests which don't include everything that came before them. 
The only changes for this pull request are:
- sql.js
- files.js

Probably best just to copy it out. 
NOTE: haven't tested all the operations, but it's sufficient for review.


jlhughes:
I've changed use of SQLSupport from 'implements' to Singleton to avoid complications with Java compilation.  And the changes, though pushed to the same branch, have not been added to this pull request.  I'll create another.

PR #232, opened 2017-02-02
Create NodeModelDAO, ClassLoader, a nodejs helper, and some i18n support.

mcarcaso:


src/foam/core/ClassLoader.js
```diff
@@ -17,48 +17,92 @@
...
+
+  methods: [
+    function arequire() {
+      var p = Promise.resolve();
+      if (this.extends) {
+        p = p.then(this.__subContext__.arequire(this.extends));
+      }
+      for (var i = 0, a; a = this.axioms_[i]; i++) {
+        if (a.arequire) {
+          p = p.then(a.arequire());
```
adamvy-google: I don't believe this is correct.  You can't pass a promise to p.then().

What you really want is to either build an array of promises

promises.push(a.arequire());

and then do return Promise.all(promises);

In fact I would restructure this whole method as such

var promises = [];

if ( this.extends ) promises.push(this.__subContext__.arequire(this.extends));

for ( var i = 0, a; a = this.axioms... ) {
  if ( a.arequire ) promises.push(a.arequire());
}

return Promise.all(promises);

src/foam/core/ClassLoader.js
```diff
@@ -17,48 +17,92 @@
...
+
   methods: [
     {
       name: 'arequire',
       class: 'foam.core.ContextMethod',
       code: function(X, modelId) {
-        // TODO: This method only adds requires. Make it add extends,
-        // properties, and anything else that might be needed.
+        if (this.__subContext__.isRegistered(modelId)) return Promise.resolve();
+        if (this.pending[modelId]) return this.pending[modelId];
```
adamvy-google: I would test that this actually handles loops correctly.  I'm not sure it will, because by returning the previous promise you could end up in a situation where arequire('Foo') is waiting on arequire('Bar') which is waiting on the exisiting promise arequire('Foo');

I think you want to return an empty promise if pending is set.

src/foam/core/Requires.js
```diff
@@ -88,10 +88,11 @@ foam.CLASS({
           var path  = a[0];
           var parts = path.split('.');
           var name  = a[1] || parts[parts.length-1];
-          return foam.core.Requires.create({name: name, path: path});
+          return foam.core.Requires.create(
+              {name: name, path: path}, this.__subContext__);
```
adamvy-google: just pass `this` as the second parameter.  It will automatically pull out this.__subContext__ if it needs it.

src/foam/core/Requires.js
```diff
@@ -88,10 +88,11 @@ foam.CLASS({
...
         }
 
-        return foam.core.Requires.create(o);
+        return foam.core.Requires.create(o, this.__subContext__);
```
adamvy-google: ditto

src/foam/dao/NodeModelFileDAO.js
```diff
@@ -29,17 +29,25 @@ foam.CLASS({
...
-          return Promise.reject(
-              'Unable to load at ' + path + '. Error: ' + e.stack);
-        }
+      var foamCLASS = foam.CLASS;
+      var X = this.__subContext__;
+
+      var model;
+      foam.CLASS = function(m) {
+        var cls = m.class ? foam.lookup(m.class) : foam.core.Model;
+        model = cls.create(m, X);
```
adamvy-google: ditto, just pass this as second parameter, or in this case do var self = this; cls.create(m, self);

tools/node.js
```diff
@@ -73,11 +73,13 @@ foam.CLASS({
...
       var modelArgs = this.modelArgs;
       var X = this.__subContext__;
-      return X.arequire(this.modelId).then(function(model) {
-        return model.create(modelArgs, X).execute();
-      }).catch(function(err) {
-        console.log(err.stack);
-      });
+      var modelId = this.modelId;
+      return X.arequire(modelId)
+        .catch(console.log)
```
adamvy-google: Did you test this?

I'm suprised you don't need to do console.log.bind(console)

tools/node.js
```diff
@@ -73,11 +73,13 @@ foam.CLASS({
...
+        .then(function() {
+          return foam.lookup(modelId).create(modelArgs, X).execute();
```
adamvy-google: Again, use the parent object as a second parameter, not the context directly.

adamvy-google (changes requested):

src/foam/core/ClassLoader.js
mcarcaso: Good call! Done.

tools/node.js
kgrgreer: They changed this in Chrome a couple releases ago, but this will break in browsers other than Chrome. Actually, I see that it works in Firefox also.

src/foam/core/Requires.js
mcarcaso: Done

src/foam/core/Requires.js
mcarcaso: Done

src/foam/dao/NodeModelFileDAO.js
mcarcaso: Done

tools/node.js
mcarcaso: I think so. I can run the following line from the browser:
new Promise(function() { throw 'hello' }).catch(console.log);
and it just logs 'hello'

tools/node.js
mcarcaso: Done.

src/foam/core/ClassLoader.js
mcarcaso: I think this works ok but maybe I didn't run the right test. I created a circular dependency in the models located in demos/node/js and ran both the node script and the browser at http://localhost:8000/?model=nodetooldemo.Test&classpath=/src/,/demos/node/js/ and it seemed to be fine.

tools/node.js
mcarcaso: This script is only to be used when running node.js so I don't think we need to worry about how this works in browsers.

demos/node/js/nodetooldemo/TestCustomProperty.js
```diff
@@ -0,0 +1,28 @@
...
+foam.CLASS({
+  package: 'nodetooldemo',
+  name: 'TestCustomProperty',
+  extends: 'foam.core.Property',
+  properties: [
+    {
+      name: 'value',
+      value: 'This is the custom property default value.',
+    },
+  ],
```
adamvy-google: FOAM style is to not have any trailing commas.

demos/node/run_script.sh
```diff
@@ -0,0 +1,19 @@
...
+{
+  classpaths: [
+    '$DIR/../../src',
+    '$DIR/js'
+  ],
+  locale: 'fr',
+  modelId: 'nodetooldemo.Test',
+  modelArgs: {
+    name: 'Joe',
+  },
```
adamvy-google: Trailing commas are technically illegal in JSON.  Although FOAM's json is a bit more permissive.

index.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+  ],
+  methods: [
+    function execute() {
+      var self = this;
+      var X = this.__subContext__;
+      var modelId = this.model;
+      var viewId = this.view;
+      Promise.all([
+        X.arequire(viewId),
```
adamvy-google: Can't we use this.arequire, since we imported it?

demos/android/run_script.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+rm $DIR/gen/*
+mkdir -p $DIR/gen
+
+read -r -d '' ARGS << EOM
```
adamvy-google: Technically read -d is a bashism, so change the first line of these scripts to #!/bin/bash

demos/android/run_script.sh
```diff
@@ -0,0 +1,42 @@
...
+  modelArgs: {
+    models: [
+      'nodetooldemo.Test',
+    ],
+    outfile: '$DIR/gen/strings.xml',
+  },
+  locale: 'en',
+}
+EOM
+node $DIR/../../tools/node.js "$ARGS"
```
adamvy-google: I'm also starting to think it might simplify things if we made these scripts just javascript scripts instead of shell scripts calling into foam.

You can do 

#!/usr/bin/env node

// JS code here
require('path/to/foam.js');

Something.create().execute();

etc.
etc.


Might be worth trying at some point to see if it simplifies things a bit.

adamvy-google: We could even do some other fun things along this line, but I'll talk to you about that later.

index.js
```diff
@@ -0,0 +1,107 @@
...
+        X.arequire(modelId),
+      ]).then(function() {
+        var model = X.lookup(modelId).create({}, self);
```
adamvy-google: Don't pass the empty object as argument one.  It's faster to just pass null if you don't have any args to pass.

index.js
```diff
@@ -0,0 +1,107 @@
...
+        var view = X.lookup(viewId).create({showActions: true}, self);
```
adamvy-google: If viewId was not specified we should do view = model.toE(null, self);

index.js
```diff
@@ -0,0 +1,107 @@
...
+        view.data = model;
+        view.write();
+      })
+    }
+  ],
+});
+
+var search = /([^&=]+)=?([^&]*)/g;
```
adamvy-google: I kind of think that this code should be part of WebModelExecutor rather than top level code.

src/foam/android/GenResources.js
```diff
@@ -0,0 +1,72 @@
...
+  methods: [
+    function execute() {
+      var self = this;
+      if (!this.outfile) {
+        console.log('ERROR: outfile not specified');
+        process.exit(1);
+      }
+      var p = Promise.resolve();
+      for (var i = 0; i < this.models.length; i++) {
+        p = p.then(this.arequire(this.models[i]))
```
adamvy-google: Same comments I made previously about building an array of promises and then doing Promise.all. I dont' think p.then(arequire()) does what you are expecting.

src/foam/android/GenStrings.js
```diff
@@ -0,0 +1,92 @@
...
+        },
+        {
+          class: 'String',
+          name: 'message'
+        },
+      ]
+    }
+  ],
+  methods: [
+    function modelToResources(model) {
```
adamvy-google: Might want to rename this to be classToResources() and name the parameter cls, since technically you're taking a Class not a Model.

src/foam/core/ClassLoader.js
```diff
@@ -0,0 +1,112 @@
...
+ */
+
+foam.CLASS({
+  package: 'foam.core',
+  name: 'ModelARequireExtension',
+  refines: 'foam.core.Model',
+
+  methods: [
+    function arequire() {
+      var X = this.__subContext__;
```
adamvy-google: I think we should use `this.__context__;`

Things you import should be taken from `__context__`, `__subContext__` is for things that we export.  And you are essentially importing arequire() here.

src/foam/core/ClassLoader.js
```diff
@@ -0,0 +1,112 @@
...
+});
+
+foam.CLASS({
+  package: 'foam.core',
+  name: 'RequiresARequireExtension',
+  refines: 'foam.core.Requires',
+
+  methods: [
+    function arequire() {
+      return this.__subContext__.arequire(this.path);
```
adamvy-google: Again `this.__context__`

src/foam/core/ClassLoader.js
```diff
@@ -0,0 +1,112 @@
...
+    },
+  ],
+
+  methods: [
+    {
+      name: 'arequire',
+      class: 'foam.core.ContextMethod',
+      code: function(X, modelId) {
+        if (X.isRegistered(modelId)) return Promise.resolve();
+        if (this.pending[modelId]) return this.pending[modelId];
```
adamvy-google: Okay so this _would_ hang if you have a cycle, except that since you're registering the model before you call model.arequire(), this case is never hit via a circular dependency.

That being said, it feels more correct to me to do m.arequire() before we register the latch, as otherwise isRegistered() will return true for models that arn't quite loaded yet.  So I would make that change and then also fix this if to just return Promise.resolve()

src/foam/dao/NodeModelFileDAO.js
```diff
@@ -0,0 +1,54 @@
...
+    {
+      name: 'classpath',
+    },
+  ],
+
+  methods: [
+    function find(id) {
+      var foamCLASS = foam.CLASS;
+      var self = this;
+      var X = this.__subContext__;
```
adamvy-google: Is this variable used anywhere?

src/foam/core/ClassLoader.js
```diff
@@ -0,0 +1,112 @@
...
+      },
+    },
+  ],
+
+  methods: [
+    {
+      name: 'arequire',
+      class: 'foam.core.ContextMethod',
+      code: function(X, modelId) {
+        if (X.isRegistered(modelId)) return Promise.resolve();
```
adamvy-google: FOAM style is to put spaces after the "(" and before the ")"

src/foam/dao/NodeModelFileDAO.js
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+foam.CLASS({
+  package: 'foam.dao',
```
adamvy-google: I'm not sure this it the right package for this.  Especially since its such an impoverished DAO.  Perhaps a foam.classloader package would be more appropriate.

I also think this should be added to files.js with a nodejs flag.

src/foam/dao/WebModelFileDAO.js
```diff
@@ -0,0 +1,71 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+foam.CLASS({
+  package: 'foam.dao',
```
adamvy-google: Same comment as NodeModelFileDAO

src/foam/dao/WebModelFileDAO.js
```diff
@@ -0,0 +1,71 @@
...
+      factory: function() {
+        return this.window.location.protocol + '//' + this.window.location.host + '/src/';
+      },
+    },
+  ],
+
+  methods: [
+    function find(id) {
+      var self = this;
+      var X = this.__subContext__;
```
adamvy-google: Unused variable I assume?

tools/node.js
```diff
@@ -0,0 +1,89 @@
...
+      },
+    },
+  ],
+  methods: [
+    function execute() {
+      var self = this;
+      var modelArgs = this.modelArgs;
+      var X = this.__subContext__;
+      var modelId = this.modelId;
+      return X.arequire(modelId)
```
adamvy-google: You imported arequire, so you can call it with just this.arequire(modelId)

adamvy-google (changes requested):
Leaving the i18n stuff for now, will review that later.

demos/android/gen_xml.js
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
```
adamvy-google: Look at that, 8 lines shorter! :D

mcarcaso:
Addressed most of your comments in your review individually but my comments seem to have been lost. In summary:
I've accounted for the circular dependency (if there's anything you want me to test, let me know)
I moved all of the classloader stuff into its own directory
I replaced bash scripts with js scripts

I'll remove the i18n stuff from this branch and re-introduce it on a different branch. I'll update this when that's all done.

index.html
```diff
@@ -1,8 +1,13 @@
...
     <script src="src/foam.js"></script>
-    <script src="src/foam/dao/OrDAO.js"></script>
-    <script src="src/foam/dao/WebModelFileDAO.js"></script>
-    <script src="index.js"></script>
+    <script src="src/foam/classloader/OrDAO.js"></script>
+    <script src="src/foam/classloader/WebModelFileDAO.js"></script>
+    <script src="src/foam/classloader/WebModelDisplayer.js"></script>
+    <script>
+      var executor = foam.classloader.WebModelDisplayer.create()
+      executor.fromQuery(this.window.location.search.substring(1));
```
adamvy-google: Pretty sure in this context "this" and "window" are the same thing.

src/foam/classloader/ClassLoader.js
```diff
@@ -67,14 +67,25 @@ foam.CLASS({
     {
       name: 'arequire',
       class: 'foam.core.ContextMethod',
-      code: function(X, modelId) {
-        if (X.isRegistered(modelId)) return Promise.resolve();
-        if (this.pending[modelId]) return this.pending[modelId];
+      code: function(X, modelId, deps) {
```
adamvy-google: Maybe name it opt_deps

src/foam/classloader/ClassLoader.js
```diff
@@ -67,14 +67,25 @@ foam.CLASS({
...
 
         var modelDao = X[foam.String.daoize(foam.core.Model.name)];
-        this.pending[modelId] = modelDao.find(modelId).then(function(m) {
+        var m;
+        this.pending[modelId] = modelDao.find(modelId).then(function(model) {
+          m = model;
           m.validate();
-          if (X.isRegistered(modelId)) return m;
+          return m.arequire(deps);
+        }).then(function() {
```
adamvy-google: It might be useful to have model.arequire() resolve to the model.

Would make this slightly shorter as you could just get m as a parameter to this promise.

Not a big deal though, but something you could do if you wanted.

adamvy-google (commented):
LGTM overall.  Few small comments that you can choose to address at your discretion.

mcarcaso:
Thanks. Moved out the i18n stuff and addressed just about all of the comments.

PR #233, opened 2017-02-05
Correct v1 assignment in Unary and Binary, and add comments regarding…

jlhughes:
… use of SUPER for validation.

### DIFF
--- a/src/foam/dao/SQL.js
+++ b/src/foam/dao/SQL.js
@@ -84,12 +84,17 @@ foam.CLASS({
 });
 
 
+/**
+   The base models Unary and Binary are used to validate the arguments
+   and throw a SQLException when invalid.
+   Each child model shall call SUPER() to validate.
+ */
 foam.CLASS({
   refines: 'foam.mlang.predicate.Unary',
   requires: [ 'foam.dao.SQLException' ],
   methods: [
     function toSQL() {
-      var v1 = this.arg1.toSQL ? v1 = this.arg1.toSQL() :
+      var v1 = this.arg1.toSQL ? this.arg1.toSQL() :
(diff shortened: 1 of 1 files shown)


src/foam/dao/SQL.js
```diff
@@ -84,12 +84,17 @@ foam.CLASS({
 });
 
 
+/**
+   The base models Unary and Binary are used to validate the arguments
```
adamvy-google, line 97: I'm personally not wild about this setup.

This means that when Eq goes to generate its SQL, it will call SUPER() which calls arg1.toSQL() and arg2.toSQL(), then it goes to call SQLSupport.values() which also calls arg1.toSQL() and arg2.toSQL().  We're generating twice as much SQL as we need not to mention checking for the existence of arg1, arg2, arg1.toSQL, arg1.toString, arg2.toSQL, arg2.toString, multiple times.

In what cases would arg1 or arg2 not have a toSQL() method?

If we refine toSQL() on all things that can be in an expression (Properties and all mlangs) then we can reliably call toSQL() on all arguments.

arg1 and arg2 already have validation to ensure you don't set them to invalid values.

jlhughes: OK, If you refine toSQL on all things, then we can forgo the validation.

jlhughes: Also, regardless of changing/removing the validation, this pull-request is a patch to fix the MLang SQL that is already part of the foam package.  So these fixes need to be applied and we should discuss the validation refactoring separately.

adamvy-google: I'm okay with discussing validation refactoring later, but if we do that then let's leave a TODO here for now to revisit this design.

PR #234, opened 2017-02-05
ChoiceView - Add support for 'optional' or 'no-selection'

jlhughes:
If the ChoiceView is 'optional' then the 'placeholder' is presented as the first list entry and if an explicit choice is not set, the 'placeholder' is selected. The ChoiceView returns null for the placeholder selection.


src/foam/u2/tag/Select.js
```diff
@@ -45,8 +47,15 @@ foam.CLASS({
...
 
-        if ( placeholder ) {
-          cs.push(self.E('option').attrs({disabled: 'disabled'}).add(self.placeholder));
+          if ( placeholder ) {
+            if (this.optional) {
+              cs.push(self.E('option').attrs({
+                selected: self.data == -1 ? true : undefined,
+                value: -1
+            }).add(self.placeholder));
+          } else {
```
kgrgreer: Your indentation makes it look like the else is for the first if statement, but it's actually for the inner one.

kgrgreer:
Is there any case where you would would not want it to be 'optional'? Is this needed, or can we just change the default behaviour?

jlhughes:
I'm unable to 'comment' directly on the pull-request at the moment.

re: optional - default behaviour
I have a roughly 70/30 split between optional/non-optional for the 20 or so
ChoiceViews.  For me I think of 'enabling' optional rather than 'disabling'
it.   Also, for your examples and use-cases you've had no need for
optional.

PR #237, opened 2017-02-07
Some i18n support

mcarcaso:
Adds the Message axiom and an android string generator that makes use of it.


src/foam/android/GenResources.js
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+foam.CLASS({
+  package: 'foam.android',
```
adamvy: Might make more sense for things like this to live in foam.android.tools

I sure in the future we'll have foam.android.dao, foam.android.ui etc etc.

src/foam/i18n/Messages.js
```diff
@@ -0,0 +1,83 @@
...
+  name: 'MessagesExtension',
+  refines: 'foam.core.Model',
+
+  properties: [
+    {
+      name: 'messages',
+      class: 'AxiomArray',
+      of: 'foam.i18n.MessageAxiom',
+      adaptArrayElement: function(o, prop) {
+        return foam.lookup(prop.of).create(o);
```
adamvy-google: I believe this is the default adapt of AxiomArray, so you can leave it out.

mcarcaso:
Sorry I forgot about this pull request. I'm pretty sure I addressed Adam's comments way back when. I rebased this on the RolledBackModelDAOs branch because it depends on the behaviour of the rolled back DAOs to get the right dependencies in place.

jacksonic:
Was GoogleBot confused by the RolledBackModelDAOs base branch swap? I don't see any non-mcarcaso commits now.

kgrgreer:
I think it was confused because I hadn't set my GIT username property on my
new machine.

PR #238, opened 2017-02-08
Refactor SQL, moving validation into SQLSupport

jlhughes:
Move validation into SQLSupport which removes the use of SUPER for validation and results in arg1/arg2 toSQL/toString only being called once. 
Unary/Binary throw SQLExceptions reporting that toSQL has not be implemented for the particular mlang.  This could be removed if we decide to refine toSQL into all mlangs.

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -117,6 +117,7 @@ FOAM_FILES([
   { name: "foam/dao/TimingDAO" },
   { name: "foam/dao/LoggingDAO" },
   { name: "foam/dao/IDBDAO", flags: ['web'] },
+  { name: "foam/dao/SQL" },
   { name: "foam/parse/QueryParser" },
   { name: "foam/physics/Physical" },
   { name: "foam/physics/Collider" },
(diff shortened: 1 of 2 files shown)


src/files.js
```diff
@@ -117,6 +117,7 @@ FOAM_FILES([
   { name: "foam/dao/TimingDAO" },
   { name: "foam/dao/LoggingDAO" },
   { name: "foam/dao/IDBDAO", flags: ['web'] },
+  { name: "foam/dao/SQL" },
```
kgrgreer: Can we put this behind a 'sql' flag, since it isn't useful as a core library function, but only to those who are also loading some kind of SQL DAO.

src/foam/dao/SQL.js
```diff
@@ -161,9 +124,7 @@ foam.CLASS({
   requires: [ 'foam.mlang.predicate.SQLSupport' ],
   methods: [
     function toSQL() {
-      this.SUPER();
-
-      var values = this.SQLSupport.create().values(this.arg1, this.arg2);
+      var values = this.SQLSupport.create().values(this);
```
kgrgreer: Needs porting.

PR #240, opened 2017-02-09
Inversion of control for diffing of properties: 

mdittmer:
`FObject` delegates to `property.compareProperty(o1, o2, diffObj)`; `Property` implements `compareProperty(o1, o2, diffObj)` which delegates to `comparePropertyValues(f(o1), f(2), diffObj)`.

This _almost_ mirrors how comparison works, except that `diffObj` is a writable parameter that must be passed around because `o2` could 'add value: `undefined`', complicating the issue of returning a 'no diff' value (which would otherwise allow us to avoid passing `diffObj` around). This corner case was caught by existing tests :)

### DIFF
--- a/src/foam/core/Property.js
+++ b/src/foam/core/Property.js
@@ -205,6 +205,36 @@ foam.CLASS({
           return comparePropertyValues(f(o1), f(o2));
         };
       }
+    },
+    {
+      name: 'diffPropertyValues',
+      transient: true,
+      factory: function() {
+        var name = this.name;
+        return function(v1, v2, diff) {
+          if ( Array.isArray(v1) ) {
+            var subdiff = foam.Array.diff(v1, v2);
+            if ( subdiff.added.length !== 0 || subdiff.removed.length !== 0 ) {
+              diff[name] = subdiff;
+            }
+          } else if ( ! foam.util.equals(v1, v2) ) {
+            // if the primary value is undefined, use the compareTo of the other
+            diff[name] = v2;
+          }
+          return diff;
+        };
+      }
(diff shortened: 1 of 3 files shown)


mdittmer:
@kgrgreer PTAL

src/foam/core/Property.js
```diff
@@ -205,6 +205,36 @@ foam.CLASS({
           return comparePropertyValues(f(o1), f(o2));
         };
       }
+    },
+    {
+      name: 'diffPropertyValues',
+      transient: true,
+      factory: function() {
```
adamvy-google: Rather than build a new diff closure for every property, just use a default value and pass the property as the last parameter.

Same pattern we use for adapt/preSet/postSet etc.

Also we could refine all this in from a diff.js package

mdittmer: Clarification from offline chat: @adamvy proposes that _the whole diff API_ live in a refinement package.

src/foam/core/Property.js
```diff
@@ -205,6 +205,29 @@ foam.CLASS({
           return comparePropertyValues(f(o1), f(o2));
         };
       }
+    },
+    {
+      name: 'diffPropertyValues',
+      transient: true,
+      value: function(v1, v2, diff) {
+        if ( Array.isArray(v1) ) {
+          var subdiff = foam.Array.diff(v1, v2);
```
kgrgreer, line 214: Would be better to not perform this check here but instead have a different implementation of diffPropertyValues in Array Property classes, but I'll accept this now because it's still better than what we have.

PR #242, opened 2017-02-23
Implement hashing of float values in Javascript...

mdittmer:
... Also consolidate hash-combining logic.


src/foam/core/types.js
```diff
@@ -129,7 +130,32 @@ foam.CLASS({
...
+      function(value) {
+        this.float64Array[0] = value;
+        return foam.util.hashTogether(
+            foam.util.hashCode(this.int32Array[0]),
+            foam.util.hashCode(this.int32Array[1]));
+      }
+    ],
+    [
+      'arrayBuffer',
+      new ArrayBuffer(8)
```
adamvy-google: This will make all floats in the system share a single array buffer.

mdittmer:
@jacksonic, I'm not sure why indexes are affected here. PTAL: https://travis-ci.org/foam-framework/foam2/jobs/204634955

src/foam/core/types.js
mdittmer: That's okay, right? We want to use this array buffer for synchronous hashing.

adamvy-google: I guess so, but then why use a separate Float64Array for every property?

mdittmer:
It was to avoid using a closure. The typed arrays need to pass in the
buffer that backs them. We could, instead, use a closure.

src/foam/core/types.js
mdittmer: [dup] Email replies don't stay threaded:

It was to avoid using a closure. The typed arrays need to pass in the
buffer that backs them. We could, instead, use a closure.

adamvy-google:
Not saying this is the right solution, but what about just adding a better hashCode method in foam.Number stdlib?

src/foam/core/stdlib.js
```diff
@@ -338,19 +338,31 @@ foam.LIB({
...
+    name: 'foam.Number',
+    methods: [
+      function isInstance(o) { return typeof o === 'number'; },
+      function clone(o) { return o; },
+      function equals(a, b) { return a === b; },
+      function compare(a, b) {
+        return ( b === null || b === undefined ) ? 1 :
+            a < b ? -1 : a > b ? 1 : 0;
+      },
+      function hashCode(n) {
```
adamvy-google: Personally I would find it a bit cleaner to build the closure here just for the hashCode function.

But it's not a big deal.

mdittmer: Done.

adamvy-google (approved):

mdittmer:
@jacksonic PTAL at changes in latest push. I think the test expectations were changed correctly, and I added a new one, but I'd like verification before I merge.

jacksonic (approved):

PR #243, opened 2017-02-23
Hash plain objects by visiting keys...

mdittmer:
... also, differentiate between falsey values.


src/foam/core/stdlib.js
```diff
@@ -619,7 +619,14 @@ foam.LIB({
     function compare(a, b) {
       return foam.Number.compare(a.$UID, b ? b.$UID : -1);
     },
-    function hashCode(o) { return 0; },
+    function hashCode(o) {
+      var hash = 19;
+      for (var key in o) {
```
kgrgreer: This line and next are non-standard spacing.

jacksonic (approved):

src/foam/core/stdlib.js
mdittmer: Fixed in next push.

PR #244, opened 2017-02-28
mlang Has for java

jlhughes:
Need to discuss if we should test for empty string as well as null.


src/foam/mlang/mlangJava.js
```diff
@@ -323,6 +323,17 @@ foam.CLASS({
 });
 
 
+foam.CLASS({
+  refines: 'foam.mlang.predicate.Has',
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: 'return getArg1().f(obj) != null && getArg1().f(obj) != "";'
```
adamvy-google: So JS checks for undefined, null, empty string and empty array.

I'm not sure what the correct behaviour is, but we should be consistent in any case.

jlhughes: Testing for an empty array is not straight forward.  Testing for an array can be accomplished with 'value.getClass().isArray()' or 'instanceof _Type_[]', but then to get the length, it needs to be cast or you have to know the _Type_.

jlhughes:
Hi  Adam
The test for empty string and empty Array was added some time ago, appears my comment regarding it was hidden. 
Joel

src/foam/mlang/mlangJava.js
```diff
@@ -323,6 +323,20 @@ foam.CLASS({
...
 
 
+foam.CLASS({
+  refines: 'foam.mlang.predicate.Has',
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: 'Object value = getArg1().f(obj); '+
+        'return ! (value == null || '+
```
adamvy, line 333: Are you sure this compiles?

I don't believe isEmpty() is a method on all Objects in java.

I also don't think we should compare the output of toString() as that may require building strings we don't need which is slow.

jlhughes: yes, I commented on this previously, but it was lost/hidden:
"Testing for an empty array is not straight forward. Testing for an array can be accomplished with 'value.getClass().isArray()' or 'instanceof Type[]', but then to get the length, it needs to be cast or you have to know the Type."
How would you like proceed?

adamvy: We can use getClass().isArray() && java.lang.reflect.Array.getLength(value)

**https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/Array.html#getLength(java.lang.Object)**

jlhughes:
Will re-open on new fork.

PR #245, opened 2017-03-01
Split sqlValue into sqlName and sqlValue to facilitate property name …

jlhughes:
…adjustments during SQL WHERE clause construction.


adamvy-google:
This control flow is getting more complicated.

I still think we should just refine toSQL() onto Property and then we just do

return this.arg1.toSQL() + ' IS NOT NULL'

and similar throughout all the expressions.  If all expressions implement toSQL(), then we don't need so many checks.

jlhughes:
Abandoned.

PR #248, opened 2017-03-03
Universal http request / response types

mdittmer:
... also changed `(XHR = XMLHttpRequest)(Request|Response)` to `XMLHTTP(Request|Response)`


src/com/google/foam/demos/olympics/Controller.js
```diff
@@ -24,7 +24,7 @@ foam.CLASS({
     'foam.dao.MDAO',
```
mdittmer: Ignore file: Feature required universal HTTPRequest. Will merge in #248 after it's reviewed.

src/com/google/net/OAuth2.js
```diff
@@ -117,8 +117,8 @@ foam.CLASS({
```
mdittmer, line 117: Ignore file: Feature required universal HTTPRequest. Will merge in #248 after it's reviewed.

src/lib/firebase.js
```diff
@@ -345,7 +345,7 @@ foam.CLASS({
   extends: 'com.firebase.FirebaseDAO',
```
mdittmer: Ignore file: Feature required universal HTTPRequest. Will merge in #248 after it's reviewed.

src/lib/net.js
```diff
@@ -175,7 +175,7 @@ foam.CLASS({
           return this.resp.json();
```
mdittmer: Ignore file: Feature required universal HTTPRequest. Will merge in #248 after it's reviewed.

src/lib/node/net.js
```diff
@@ -685,9 +685,9 @@ foam.CLASS({
       if ( this.port ) options.port = this.port;
```
mdittmer: Ignore file: Feature required universal HTTPRequest. Will merge in #248 after it's reviewed.

src/lib/net.js
```diff
@@ -681,3 +681,65 @@ foam.CLASS({
     }
   ]
 });
+
+
+foam.ENUM({
```
mdittmer: TODO: We should early-bind environment. We can't load things for web in NodeJS in a 3rd party tool right now, anyway.

src/lib/net.js
```diff
@@ -681,3 +681,65 @@ foam.CLASS({
...
+      var runtimeCls = this.getRuntimeClass(cls);
+      cls.create = function(args, opt_parent) {
+        return runtimeCls.create(args, opt_parent);
+      };
+    }
+  ]
+});
+
+
+foam.CLASS({
```
mdittmer: TODO: Just register the appropriate model(s) under `universal` package name, statically. (using `foam.register()`).

mdittmer: TODO: Move `foam.net.AnyModel` to `foam.net.web`. This should just be `foam.net`.

mdittmer:
`foam.net` and `foam.net.web` changes applied, as per design review. @kgrgreer PTAL.

src/lib/box.js
```diff
@@ -1037,7 +1037,7 @@ foam.CLASS({
   extends: 'foam.box.PromisedBox',
 
   requires: [
-    'foam.net.Socket',
+    'foam.net.web.Socket',
```
kgrgreer: Ask Adam.

mdittmer: So, this is definitely wrong. I pushed a fix that qualifies `Socket` and `SocketService` as `foam.net.node`, because that's the only environment they will work in.

@adamvy anything I'm missing there?

adamvy: Fine with me.

src/lib/net.js
```diff
@@ -15,731 +15,19 @@
...
-});
+(function() {
+  var pkg = 'foam.net.' + (typeof process === 'undefined' ? 'web' : 'node');
+  console.log('PKG:', pkg);
+  var clss = [
+    'HTTPRequest',
+    'HTTPResponse'
+  ];
+
+  // For each class with a "web" (browser) and "node" (Node JS)
```
adamvy: Add a TODO this should be provided via some sort of ContextFactory IMO.

Also your comment is not entirely accurate, since you are not doing this for the WebSocketService class, which might be good to do.

mdittmer: Do you mean that `node.SocketService` and `web.WebSocketService` have the same API and should be unified under `foam.net.SocketService`?

mdittmer: ... if so, perhaps we should keep the `web` package but drop the `Web` classname prefix on `web.WebSocketService`. @adamvy WDYT?

adamvy: No, there is a WebsocketService for both web and nodejs.  The nodejs one extends the web one.

mdittmer: ... nvm. Didn't realize we had a `node.WebSocket(Service)`.

PR #252, opened 2017-03-08
Client-side REST DAO unit tests & implementation

mdittmer:


src/foam/core/JSON.js
```diff
@@ -435,6 +435,13 @@ foam.CLASS({
             a[i] = this.objectify(o[i]);
           }
           return a;
+        },
+        Object: function(o) {
+          var ret = {};
+          for ( var key in o ) {
+            if ( o.hasOwnProperty(key) ) ret[key] = this.objectify(o[key]);
```
kgrgreer: Add a note that a faster version could only create 'ret' when the first objectified object is different than o[key]. This would avoid creating new objects when not necessary.

src/foam/dao/RestDAO.js
```diff
@@ -0,0 +1,169 @@
...
+    'foam.net.HTTPRequest'
+  ],
+
+  properties: [
+    {
+      class: 'String',
+      name: 'baseURL',
+      documentation: 'URL for most rest calls. Some calls add "/<some-info>".',
+      final: true,
+      factory: function() {
```
kgrgreer: Better to mark required: true, then call this.validate() before trying to use.

mdittmer: Is `required: true` anything more than a UI hint? (@jingt06 this applies to comments in one of your open PRs).

Also, what's meant by "before trying to use"? Should we put something here, or does client code need to know to do this?

jacksonic: Should this value be force-supplied at object creation time, or is there a "configuration phase" you want where the object is partially invalid and verified before use?

jacksonic (approved):

src/foam/dao/RestDAO.js
mdittmer: No such initialization phase. A `RestDAO` with no `baseURL` "has no meaning". If you don't know the `baseURL` yet, you're not ready to construct a `RestDAO` yet.

kgrgreer: @jacksonic There is sort of a "configuration phase". You can do whatever you want and then you call obj.validate(), at which point it validates everything. There should be a ValidatingDAO (or we can write one if there isn't already) that validates before put(). We don't know when the data needs to be valid for your use-case (when going into the DAO, across the network, before enabling the UPDATE button, etc.), so it's up to you to call .validate() (or have some standard code like a ValidatingDAO or standard Controller do it).

mdittmer: OK. I set `required: true` and setup `createRequest_()` to invoke `validate()`.

PR #259, opened 2017-03-14
Register named refinements in context

mdittmer:
Modify Boot.phase2 to register refinements under <package>.<name>, but only when name is provided in class/refinement definition.

### DIFF
--- a/src/foam/core/Boot.js
+++ b/src/foam/core/Boot.js
@@ -192,28 +192,49 @@ foam.LIB({
       /* Creates a Foam class from a plain-old-object definition.
           @method CLASS
           @memberof module:foam */
-      foam.CLASS = function(m) {
-        var cls   = m.class ? foam.lookup(m.class) : foam.core.Model;
-        var model = cls.create(m);
-        model.validate();
-        var cls = model.buildClass();
+      foam.CLASS = function(clsDef) {
+        var classOfModel = clsDef.class ?
+            foam.lookup(clsDef.class) : foam.core.Model;
+        var modelOfClass = classOfModel.create(clsDef);
+        modelOfClass.validate();
+        var cls = modelOfClass.buildClass();
         cls.validate();
 
-        if ( ! m.refines ) {
+        if ( ! clsDef.refines ) {
+          // Register class in global context.
           foam.register(cls);
 
           // Register the class in the global package path.
           foam.package.registerClass(cls);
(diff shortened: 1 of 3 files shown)


src/foam/core/Boot.js
```diff
@@ -192,28 +192,49 @@ foam.LIB({
       /* Creates a Foam class from a plain-old-object definition.
           @method CLASS
           @memberof module:foam */
-      foam.CLASS = function(m) {
-        var cls   = m.class ? foam.lookup(m.class) : foam.core.Model;
-        var model = cls.create(m);
-        model.validate();
-        var cls = model.buildClass();
+      foam.CLASS = function(clsDef) {
```
kgrgreer: Why the need to rename everything? Makes the change look artificially larger than it is and conflates two things.

src/foam/core/Boot.js
```diff
@@ -192,28 +192,49 @@ foam.LIB({
...
-        var m = foam.lookup(key).model_;
-        m.refines = m.id;
-        foam.CLASS(m);
+        var clsDef = foam.lookup(key).model_;
+
+        // classModel.buildClass() expects 'refines' if we are upgrading an
+        // existing class.
+        clsDef.refines = clsDef.id;
+
+        // Same as phase2()'s foam.CLASS() above, except skip registration.
```
kgrgreer: Rather than duplicating logic, why not just have a second optional parameter called skipRegistration and then call foam.CLASS(m, true)?

src/foam/core/Boot.js
mdittmer: That's true. I renamed things b/c I felt it clarified things, and is consistent with proposed changes in #263. I'll definitely change it back of the other proposed changes get dropped.

src/foam/core/Boot.js
mdittmer: My rationale was:

(1) It's not much code, and keeping things in sync is easy b/c it's just a few lines up, and
(2) we only need the skip registration case once as a part of bootstrap; no need to pay for the check every time `CLASS` is called at "actual runtime" (after bootstrap).

That said, I don't feel strongly. I can add the param if you like.

kgrgreer: Yes, please do. We had this logic duplicated already in the Enum code and then it got updated in one place and not the other which caused a bug.

PR #261, opened 2017-03-15
Better environment detection (node js vs. web)

mdittmer:
Use` typeof window` instead of `typeof process`, and set `global` to `window` in web environments

### DIFF
--- a/src/foam.js
+++ b/src/foam.js
@@ -17,7 +17,7 @@
 
 (function() {
 
-  var isServer = typeof process === 'object';
+  var isServer = typeof window === 'undefined';
   var isWorker = typeof importScripts !== 'undefined';
 
   var flags = this.FOAM_FLAGS || {};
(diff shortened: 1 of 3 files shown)


mdittmer:
@jingt06 FYI

src/foam.js
```diff
@@ -17,7 +17,7 @@
 
 (function() {
 
-  var isServer = typeof process === 'object';
+  var isServer = typeof window === 'undefined';
```
kgrgreer, line 20: How about just '! window'?

src/foam/core/lib.js
```diff
@@ -19,7 +19,7 @@
  * Top-Level of foam package
  */
 foam = {
-  isServer: typeof process === 'object',
+  isServer: typeof window === 'undefined',
```
kgrgreer, line 22: '! winddow' ?

Issue #262, opened 2017-03-15
ClassLoader/ModelFileDAO: Support for custom classes and facets


mdittmer:
The `ClassLoader` and `*ModelFileDAO` implementations do not currently deal appropriately with custom property classes or `of:`s. Here's the issue:

- The current design walks axioms (including `requires`) on models it `find()`s from the `ModelFileDAO`
- This requires `ModelFileDAO` to instantiate models (not just pass their plain-object spec up to the `ClassLoader`)
- However, when `ModelFileDAO` invokes `foam.core.Model.create(clsDef)`, it eventually gets around to `adapt`ing the `properties` `AxiomArray`; while adapting elements that have a `class:` that's not loaded yet, an error is thrown

It looks like properties, and anything else that can be specified with a custom `class:` need:

- A way to add a new class-load-promise to the current chain of promises to be resolved on `ModelFileDAO.find()`
- Some sort of `upgrade()` interface that allows placeholders to be upgraded. (Perhaps the `AxiomArray` controls the `upgrade`, and knows which placeholder to swap out?)

This allows the `ModelFileDAO` to return a model only after all its constituent parts are ready, and deals with the fact that the code "realizes" that the class is missing in the middle of synchronous `Model.create()`.

adamvy:
This could also be handled by async json parsing.

If the JSON parser had an async interface, it could pause whenever it encountered a class that wasn't available yet.

PR #263, opened 2017-03-15
ClassLoader/ModelFileDAO fixes

mdittmer:
This was work towards fixing up ClassLoader/ModelFileDAO before #262 was identified.

(1) account for require() only running code the first time;
(2) validate models at the appropriate time relative to requires axioms;
(3) check model ids before returning them from find().

### DIFF
--- a/src/foam/classloader/ClassLoader.js
+++ b/src/foam/classloader/ClassLoader.js
@@ -82,7 +82,6 @@ foam.CLASS({
 
         var modelDao = X[foam.String.daoize(foam.core.Model.name)];
         this.pending[modelId] = modelDao.find(modelId).then(function(m) {
-          modelId;
           return m.arequire(deps).then(function() {
             m.validate();
             return m;
--- a/src/foam/classloader/ClassLoader.js
+++ b/src/foam/classloader/ClassLoader.js
@@ -82,6 +82,8 @@ foam.CLASS({
 
         var modelDao = X[foam.String.daoize(foam.core.Model.name)];
         this.pending[modelId] = modelDao.find(modelId).then(function(m) {
+          // Model validation may make use of deps. Require them first, then
+          // validate the model.
           return m.arequire(deps).then(function() {
             m.validate();
             return m;
--- a/src/foam/classloader/ClassLoader.js
+++ b/src/foam/classloader/ClassLoader.js
@@ -82,8 +82,11 @@ foam.CLASS({
 
         var modelDao = X[foam.String.daoize(foam.core.Model.name)];
         this.pending[modelId] = modelDao.find(modelId).then(function(m) {
-          m.validate();
-          return m.arequire(deps).then(function() { return m; });
+          modelId;
+          return m.arequire(deps).then(function() {
+            m.validate();
+            return m;
+          });
         }).then(function(m) {
           if ( X.isRegistered(modelId) ) return m;
 
--- a/src/foam/classloader/NodeModelFileDAO.js
+++ b/src/foam/classloader/NodeModelFileDAO.js
@@ -23,6 +23,14 @@ foam.CLASS({
   properties: [
     {
       name: 'classpath'
+    },
+    {
+      name: 'loadedModels',
+      documentation: `Retain a copy of loaded models in case they are requested
+        later. This can happen, for example, if a modl is loaded twice in
+        different contexts. Subsequent require() calls will not re-invoke
+        foam.CLASS(), hence the model loaded the first time must be retained.`,
+      value: {}
     }
   ],
 
@@ -30,13 +38,25 @@ foam.CLASS({
     function find(id) {
       var foamCLASS = foam.CLASS;
       var self = this;
-      var model;
+      var model = this.loadedModels[id] || null;
 
-      foam.CLASS = function(m) {
-        var cls = m.class ? foam.lookup(m.class) : foam.core.Model;
-        model = cls.create(m, self);
-        foam.CLASS = foamCLASS;
-      }
+      if ( model ) return Promise.resolve(model);
+
+      foam.CLASS = function(clsDef) {
+        var classOfModel = clsDef.class ? foam.lookup(clsDef.class) :
+              foam.core.Model;
+        var modelOfClass = classOfModel.create(clsDef, self);
+        if ( modelOfClass.id === id ) {
+          model = modelOfClass;
+        } else {
+          // TODO(markdittmer): We should do something more reasonable here, but
+          // the DAO API only allows us to deliver one model in response to
+          // find().
+          console.warn(
+            'Class', id, 'created via arequire, but never built or registered');
+        }
+        self.loadedModels[modelOfClass.id] = modelOfClass;
+      };
 
       var sep = require('path').sep;
       var path = this.classpath + sep + id.replace(/\./g, sep) + '.js';
--- a/src/foam/classloader/NodeModelFileDAO.js
+++ b/src/foam/classloader/NodeModelFileDAO.js
@@ -32,12 +32,15 @@ foam.CLASS({
     function find(id) {
       var foamCLASS = foam.CLASS;
       var self = this;
+      // Return null if model not found.
       var model = null;
 
       foam.CLASS = function(m) {
         var cls = m.class ? foam.lookup(m.class) :
               foam.core.Model;
         var mdl = cls.create(m, self);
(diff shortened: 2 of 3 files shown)


src/foam/classloader/ClassLoader.js
```diff
@@ -82,8 +82,11 @@ foam.CLASS({
 
         var modelDao = X[foam.String.daoize(foam.core.Model.name)];
         this.pending[modelId] = modelDao.find(modelId).then(function(m) {
-          m.validate();
-          return m.arequire(deps).then(function() { return m; });
+          modelId;
```
adamvy: Why are you just referencing modelId here?

src/foam/classloader/NodeModelFileDAO.js
```diff
@@ -23,20 +23,40 @@ foam.CLASS({
   properties: [
     {
       name: 'classpath'
+    },
+    {
+      name: 'loadedModels',
```
adamvy: Mike had this support originally but I had him remove it.  It's not the ModelDAO's job to cache IMO.

src/foam/classloader/WebModelFileDAO.js
```diff
@@ -52,10 +52,19 @@ foam.CLASS({
         var model;
         var foamCLASS = foam.CLASS;
 
-        foam.CLASS = function(m) {
-          var cls = m.class ? foam.lookup(m.class) : foam.core.Model;
-          model = cls.create(m, self);
-          foam.CLASS = foamCLASS;
+        foam.CLASS = function(clsDef) {
+          var classOfModel = clsDef.class ? foam.lookup(clsDef.class) :
```
adamvy: I do not find this wordier version easier to read.  In fact I find it less readable.

src/foam/classloader/ClassLoader.js
mdittmer: Leftover from debugging. Will fix.

src/foam/classloader/NodeModelFileDAO.js
mdittmer: Who tracks whether code has already been `require`d, then?

src/foam/classloader/WebModelFileDAO.js
mdittmer: It's wordier, but it describes the meaning of the model/class back-flips we're doing here.

@jacksonic @mcarcaso @shepheb @kgrgreer am I the only one that finds this easier to parse? If so, I'll happily change it back, and maybe add a short comment instead.

src/foam/classloader/NodeModelFileDAO.js
adamvy: Context.arequire() does the check.

https://github.com/foam-framework/foam2/blob/master/src/foam/classloader/ClassLoader.js#L78

It's actually more complicated than just caching the loaded models.  Cause we also do special bookkeeping to resolve circular dependencies.

src/foam/classloader/WebModelFileDAO.js
adamvy: classOfModel followed by modelOfClass is confusing, because the modelOfClass is not the model of the class referenced in classOfModel.

The confusing bit here (IMO) is that we're overriding foam.CLASS() to extract the model definition from a file which contains a foam.CLASS call.   These descriptive names do not adequately describe the "backflips" we're doing, they just make it longer which makes it even harder to parse IMO.   I would rather keep the code as is and add a small comment explaining what we're doing.

Once you understand what this code is trying to achieve I think it reads fine as is.

src/foam/classloader/NodeModelFileDAO.js
mdittmer: That's contextualized. `require()` is global. Re-`arequire`-ing in different contexts is what instigated this change.

adamvy: Ah I see.

Maybe require wasn't the best thing to use here, but rather readFile and eval/vm.runInThisContext

But I guess this is okay too.

src/foam/classloader/WebModelFileDAO.js
mdittmer: You've convinced me. PTAL at comments added. The only "rename" change I kept is not double-declaring `cls`. I think using the same variable to mean something entirely different was what instigated the confusion for me.

mdittmer: Oops. New changes were added to #259. PTAL there. I'll make the code here match.

src/foam/classloader/NodeModelFileDAO.js
mdittmer: I would generally agree. Shall I add a TODO?

src/foam/classloader/WebModelFileDAO.js
mdittmer: Now updated here.

src/foam/classloader/NodeModelFileDAO.js
mdittmer: nvm; just changed it. PTAL.

src/foam/classloader/ClassLoader.js
```diff
@@ -82,8 +82,10 @@ foam.CLASS({
 
         var modelDao = X[foam.String.daoize(foam.core.Model.name)];
         this.pending[modelId] = modelDao.find(modelId).then(function(m) {
-          m.validate();
-          return m.arequire(deps).then(function() { return m; });
+          return m.arequire(deps).then(function() {
+            m.validate();
```
kgrgreer: Add comment to explain why this is done.

src/foam/classloader/NodeModelFileDAO.js
```diff
@@ -21,36 +21,51 @@ foam.CLASS({
...
+    { name: 'fs',  factory: function() { return require('fs');       } },
+    { name: 'vm',  factory: function() { return require('vm');       } }
   ],
 
   methods: [
     function find(id) {
       var foamCLASS = foam.CLASS;
       var self = this;
-      var model;
+      var model = null;
```
kgrgreer: Why is this change needed?

src/foam/classloader/NodeModelFileDAO.js
```diff
@@ -21,36 +21,51 @@ foam.CLASS({
...
-        return Promise.resolve(null);
-      } finally {
-        foam.CLASS = foamCLASS;
-      }
-
-      return Promise.resolve(model);
+        var cls = m.class ? foam.lookup(m.class) :
+              foam.core.Model;
+        var mdl = cls.create(m, self);
+        if ( mdl.id === id ) {
```
kgrgreer: When would mdl.id !== id?

src/foam/classloader/WebModelFileDAO.js
```diff
@@ -53,9 +53,18 @@ foam.CLASS({
         var foamCLASS = foam.CLASS;
 
         foam.CLASS = function(m) {
-          var cls = m.class ? foam.lookup(m.class) : foam.core.Model;
-          model = cls.create(m, self);
-          foam.CLASS = foamCLASS;
+          var cls = m.class ? foam.lookup(m.class) :
+                foam.core.Model;
+          var mdl = cls.create(m, self);
+          if ( mdl.id === id ) {
```
kgrgreer: When is this not the case?

src/foam/classloader/NodeModelFileDAO.js
mdittmer: The DAO should `resolve(null)` if the required code doesn't actually provide the requested class model.

src/foam/classloader/NodeModelFileDAO.js
mdittmer: If the file name matches, but it contains multiple `CLASS` calls (some of which may not be of the requested id).

src/foam/classloader/WebModelFileDAO.js
mdittmer: If the file name matches, but it contains multiple `CLASS` calls (some of which may not be of the requested id).

src/foam/classloader/NodeModelFileDAO.js
kgrgreer: Add a comment to explain that.

mdittmer:
@kgrgreer @adamvy this is getting stale. Anything else before we merge it in?

kgrgreer:
LGTM

PR #265, opened 2017-03-16
Clean up src/files.js: Dedup and add a comment

mdittmer:


adamvy:
I'm not sure this is the right fix.  Wouldn't every file need  js flag?

mdittmer:
D'oh. Yes. Well, or web or node (assuming it is deployed in JS
environments). Maybe we need some kind of "options" in addition to flags. A
"js_only" option could be checked by non-js evvironmnents. Or even "env" if
we want the meaning to be "when set, the collection of environments this
file supports".

adamvy:
K slow down.  What problem are you trying to fix with an "options" field?

PR #267, opened 2017-03-17
HTML "Parser" (Lexer for text + tags)

mdittmer:


mdittmer:
I think I'm overflowing @kgrgreer 's review queue. @adamvy @shepheb @jacksonic or @jingt06 want to take a look before I merge? The lexer is intended for cases when we want to do little or no open/close tag matching, but want to parse otherwise valid HTML documents.

src/foam/parsers/HTMLLexer.js
```diff
@@ -0,0 +1,337 @@
...
+        var grammar = this.ImperativeGrammar.create({symbols: this.symbols});
+        grammar.addActions(this.actions);
+        return grammar;
+      }
+    },
+    {
+      name: 'ps',
+      factory: function() {
+        return this.StringPS.create();
+      },
```
kgrgreer: No trailing comments.

kgrgreer:
LGTM

PR #268, opened 2017-03-17
Fix Node JS http redirect

mdittmer:
for cases such as 'http://example.com/frobinator' redirect to '/new-frobinator' (notice no protocol or host in new URL)


src/lib/node/net.js
```diff
@@ -656,16 +656,36 @@ foam.CLASS({
       class: 'Boolean',
       name: 'followRedirect',
       value: true
+    },
+    {
+      name: 'urlLib',
+      value: require('url')
```
adamvy: This isn't serializable.  I would just keep doing require('url') where needed unless it needs to be saved for performance reasons.

src/lib/node/net.js
```diff
@@ -656,16 +656,36 @@ foam.CLASS({
...
+      if ( data.protocol ) this.protocol = data.protocol.slice(0, -1);
+      if ( data.hostname ) this.hostname = data.hostname;
       if ( data.port ) this.port = data.port;
-      this.path = data.path;
+      if ( data.path ) this.path = data.path;
+
+      return this;
+    },
+
+    function copyUrlFrom(other) {
```
adamvy, line 677: Why does the standard copyFrom semantics not work for this?  Can we fix the model so that it does?

mdittmer: Well, for one thing, we don't want to copy everything, only the URL-related stuff. So using copyFrom would be something like

```js
foo.copyFrom({
  protocol: bar.protocol,
  /* ... */
})
```

For another thing, port is copied conditionally, which further mucks with a `copyFrom()` call. That's why I wanted to consolidate the careful-imperative-copy in to a helper method.

src/lib/node/net.js
kgrgreer: Can you make it a factory:?

PR #269, opened 2017-03-20
Fix some bug in cloud datastore DAO and allow multi-part ids.

jingt06:
@mdittmer take a look if you have time. Not urgent, we can discuss when you back.

1.
```javascript
var name = new Array(ids.length);
for ( var i = 0; i < ids.length; i++ ) {
     name = ids[i] + ( i < ids.length - 1 ? ':' : '' );
}
```
name was created as an Array, but was overwritten in the loop every time. Did you mean `name[i] = ...`?

2.
```javascript 
return { kind: this.getOwnDatastoreKind(), name: name };
```
As stated in the [API](https://cloud.google.com/datastore/docs/reference/rest/v1/Key#PathElement), name is a string, and is unique for each elements in dataStore. So I used ids.join('_') to create the unique id.

3.
```javascript
if ( cls.ids && cls.ids.length > 1 ) {
    ...
}
```
I didn't find ids exists in cls. DId you mean `cls.model_.ids`?

4.
```javascript
throw new Error('Not implemented: Deserialization of Cloud Datastore multi-part ids');
```
This will break in the browserAPI case. browserAPI uses ids = [brwserName, browserVersion, osName, osVersion, interfaceName, apiName]. I think the cloud datastore DAO should support multi-part ids.
Thus, for multi-part ids, I just update property values which belongs to the ids array. (An alternative way is do nothing if cls.ids.length is not 1, since the properties will be updated in the second for loop.)

### DIFF
--- a/src/com/google/cloud/datastore/types.js
+++ b/src/com/google/cloud/datastore/types.js
@@ -146,17 +146,15 @@ foam.LIB({
       var id = key.name;
 
       var o = cls.create();
-
-      if ( cls.model_.ids && cls.model_.ids.length > 1 ) {
-        throw new Error('Not implemented: Deserialization of Cloud Datastore ' +
-            'multi-part ids');
+      if ( cls.model_.ids && cls.model_.ids.length === 1 ) {
+        o[cls.model_.ids[i]] = id;
+      } else {
+        for ( var i = 0; i < cls.model_.ids.length; i++ ) {
+          idName = cls.model_.ids[i];
+          o[idName] = com.google.cloud.datastore
+            .fromDatastoreValue(entity.properties[idName]);
+        }
       }
-
-      var idProp = cls.model_.ids && cls.model_.ids.length === 1 ?
-          cls.getAxiomByName(cls.model_.ids[0]) :
-          cls.getAxiomByName('id');
-
-      if ( idProp ) o[idProp.name] = id;
 
       var props = entity.properties;
       for ( var name in props ) {
--- a/src/com/google/cloud/datastore/types.js
+++ b/src/com/google/cloud/datastore/types.js
@@ -146,8 +146,8 @@ foam.LIB({
       var id = key.name;
 
       var o = cls.create();
-      if ( cls.model_.ids && cls.model_.ids.length === 1 ) {
-        o[cls.model_.ids[i]] = id;
+      if ( ! cls.model_.ids || cls.model_.ids.length === 1 ) {
+        o.id = id;
       } else {
         for ( var i = 0; i < cls.model_.ids.length; i++ ) {
           idName = cls.model_.ids[i];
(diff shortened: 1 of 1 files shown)


src/com/google/cloud/datastore/types.js
```diff
@@ -146,18 +146,16 @@ foam.LIB({
       var id = key.name;
 
       var o = cls.create();
-
-      if ( cls.ids && cls.ids.length > 1 ) {
-        throw new Error('Not implemented: Deserialization of Cloud Datastore ' +
-            'multi-part ids');
+      if ( cls.model_.ids && cls.model_.ids.length === 1 ) {
+        o[cls.model_.ids[i]] = id;
```
mdittmer: `i` uninitialized.

src/com/google/cloud/datastore/types.js
```diff
@@ -146,18 +146,16 @@ foam.LIB({
...
+      } else {
+        for ( var i = 0; i < cls.model_.ids.length; i++ ) {
+          idName = cls.model_.ids[i];
+          o[idName] = com.google.cloud.datastore
```
mdittmer: Lookup `com.google.cloud.datastore.fromDatastoreValue` outside loop.

src/com/google/cloud/datastore/types.js
```diff
@@ -262,9 +260,9 @@ foam.CLASS({
 
       var name = new Array(ids.length);
```
mdittmer: I think this is meant to be `name = ''`, and use `+=` on the string to avoid creating garbage with arrays.

src/com/google/cloud/datastore/types.js
```diff
@@ -262,9 +260,9 @@ foam.CLASS({
...
       for ( var i = 0; i < ids.length; i++ ) {
-        name = ids[i] + ( i < ids.length - 1 ? ':' : '' );
+        name[i] = this[ids[i]];
       }
-      return { kind: this.getOwnDatastoreKind(), name: name };
+      return { kind: this.getOwnDatastoreKind(), name: name.join('_') };
```
mdittmer: `_` seems like it would be more likely to show up in an id part than `:`.

mdittmer (changes requested):
@kgrgreer PTAL; you're more familiar with FOAM2's multi-part key handling. Also, The Datastore REST API supports mulit-part keys [1], so we should probably use that rather than stringifying key parts.

[1] https://cloud.google.com/datastore/docs/reference/rest/v1beta3/projects/lookup

src/com/google/cloud/datastore/types.js
kgrgreer: This line is broken in two parts: [i] should be [0] and it isn't even safe anyway, since you don't know if cls.model_.ids event exists.
 
This line should be just: o.id = id;

src/com/google/cloud/datastore/types.js
kgrgreer: Why is this for loop needed? Would this be the same as just:

return { kind: this.getOwnDatastoreKind(), name: this.id.join('_') };

Since 'id' is set to an array where each element is a part of the multi-part key, just like you were creating.
Also, couldn't you just use foam.json.stringify(this.id) to avoid issues with keys containing either _ or :?

PR #272, opened 2017-03-21
Fix HTTPRequest.payload: Make consistently Promisey

mdittmer:
Some implementations were Promisey, some were not. Now all are Promisey. Also changed tests and DAOs that depended on non-Promisey behaviour.


mdittmer:
@jingt06 FYI

kgrgreer:
LGTM

PR #273, opened 2017-03-23
DatastoreDAO: multi-batch select() handling

mdittmer:
Run test with `npm run testThirdParty`. If you don't have `gcloud` Datastore emulator installed it should auto-install (which may take a little while).


mdittmer:
@jingt06 FYI

kgrgreer:
LGTM

PR #274, opened 2017-03-23
Rest dao remote sink

mdittmer:
@jingt06 FYI

### DIFF
--- a/src/foam/dao/RestDAO.js
+++ b/src/foam/dao/RestDAO.js
@@ -89,15 +89,9 @@ foam.CLASS({
        */
       var query = [];
 
-      // Proxy non-serializable sinks (with either "sink" or an ArraySink).
-      // When "sink" is serializable, make "proxySink" falsey as a signal to:
-      // (1) Send the serializable sink (" ! proxySink" below);
-      // (2) Signal to "onSelectResponse" that sink in payload should be passed
-      //     back to caller.
-      var proxySink = this.Serializable.isInstance(sink) ? null :
-          sink ? sink : this.ArraySink.create();
-      if ( ! proxySink )
-        query.push('sink=' + encodeURIComponent(this.jsonify_(sink)));
+      var networkSink = this.Serializable.isInstance(sink) && sink;
+      if ( networkSink )
+        query.push('sink=' + encodeURIComponent(this.jsonify_(networkSink)));
 
       if ( typeof skip !== 'undefined' )
         query.push('skip=' + encodeURIComponent(this.jsonify_(skip)));
@@ -112,7 +106,8 @@ foam.CLASS({
         method: 'GET',
         url: this.baseURL + ':select?' + query.join('&')
       }).send().then(this.onResponse.bind(this, 'select'))
-          .then(this.onSelectResponse.bind(this, proxySink));
+          .then(this.onSelectResponse.bind(
+              this, sink || this.ArraySink.create()));
     },
 
     function removeAll(skip, limit, order, predicate) {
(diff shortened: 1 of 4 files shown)


src/foam/dao/RestDAO.js
```diff
@@ -87,6 +88,14 @@ foam.CLASS({
        * Each key's value is network-foam-jsonified.
        */
       var query = [];
+
+      // Proxy non-serializable sinks. Otherwise, serialize and send the sink,
+      // then instantiate the results when they come back from the server.
+      var proxySink = this.Serializable.isInstance(sink) ? null :
```
kgrgreer: This logic doesn't look correct to me. Are you sure this is what you meant?

kgrgreer: Wouldn't this work (renaming to networkSink)?

var networkSink = this.Serializable.isInstance(sink) && sink;

mdittmer: That would work for `networkSink`, but this is the local proxy. We want it to be:
(1) `null` iff `sink` is serializable (i.e., no local proxy; just use what comes back); otherwise:
(2) `sink` if `sink` is defined; otherwise:
(3) an `ArraySink`.

Further down, we only send `sink` when `proxySink` is **not** defined; i.e., `sink` is serializable.

mdittmer: ... maybe needs better comments?

test/src/dao/RestDAO.js
```diff
@@ -100,13 +100,14 @@ describe('RestDAO', function() {
             } catch (err) {
               return Promise.resolve(createResponse({ status: 500 }));
             }
+            var sink = data.sink || this.ArraySink.create();
```
kgrgreer, line 103: Isn't this redundant? Won't all DAO's treat an undefined sink as an ArraySink anyway?

mdittmer: The current implementation only sends a sink param if a serializable sink was passed to `select()` on the client.

src/foam/dao/RestDAO.js
```diff
@@ -87,6 +88,17 @@ foam.CLASS({
        * Each key's value is network-foam-jsonified.
        */
       var query = [];
+
+      // Proxy non-serializable sinks (with either "sink" or an ArraySink).
+      // When "sink" is serializable, make "proxySink" falsey as a signal to:
+      // (1) Send the serializable sink (" ! proxySink" below);
+      // (2) Signal to "onSelectResponse" that sink in payload should be passed
+      //     back to caller.
+      var proxySink = this.Serializable.isInstance(sink) ? null :
```
kgrgreer: Isn't this design simpler and more direct:

 var networkSink = this.Serializable.isInstance(sink) && sink;
      if ( networkSink )
        query.push('sink=' + encodeURIComponent(this.jsonify_(networkSink)));

...

     .then(this.onSelectResponse);


 function onSelectResponse(localSink, payload) {
      var wasSerializable =  this.Serializable.isInstance(localSink);
      var remoteSink = foam.json.parse(payload);

      // If not proxying a local unserializable sink, just return the remote.
      if ( wasSerializable ) return remoteSink; // Or could have been passed as a parameter

      var array = remoteSink.a;
      if ( ! array )
        throw new Error('Expected ArraySink from REST endpoint when proxying local sink');

      if ( localSink.put ) {
        for ( var i = 0; i < array.length; i++ ) {
          localSink.put(array[i]);
        }
      }
      if ( localSink.eof ) localSink.eof();

      return localSink;
    },

mdittmer: This works. Needed to pass along `sink || this.ArraySink.create()` to `onSelectResponse`, but that's not problem.

PR #275, opened 2017-03-23
Fix documentation and code mismatch in mlang.CONTAINS

jingt06:
Documentations for Contains and ContainsIC in mlang are wrong.
The array.indexOf(element) method returns the first index at which a given element can be found in the array, or -1 if it is not present.


jingt06:
@mdittmer

mdittmer (approved):
LGTM

PR #276, opened 2017-03-23
Implement Datastore DAO handling of multi-part IDs…

mdittmer:
… via Property inversion-of-control.

Thanks to @jingt06 for clearly identifying the issues here (in #269). This is an alternative fix that gives `Property` objects control over how they contribute to mutli-part IDs.

There's one thing that I'm not thrilled with here, but I think the implementation is reasonably good: We're stuck having to support `find(objWithId)` as well as `find(id)`. The former case gives properties control over conversion-to-datastore. In the latter case, we cannot be sure of the right class + property to delegate to for inversion of control. In the interest of...

    It’s often better to make things distinctly different than to make them almost the same.
      ~Henry Spencer, in The Art of Unix Programming

I ground the key name multi-method in an implementation that throws for `Array`. The intended effect is that `find(id)` is supported for simple types, but *not* for multi-part id *values*. Lookup on multi-part ids require an object with property values to ensure that we don't lookup/generate/coerce two *different* ids for `find(objWithMultiPartId)` and `find(objWithMultiPartId.id)`.

@kgrgreer and @jingt06 PTAL.

### DIFF
--- a/src/com/google/cloud/datastore/types.js
+++ b/src/com/google/cloud/datastore/types.js
@@ -233,7 +233,7 @@ foam.LIB({
     {
       name: 'toDatastoreKeyName',
       code: foam.mmethod({
-        Number: function(n) { return n.toString(); },
+        Number: function(n) { return n + ''; },
         String: function(str) { return str; },
         FObject: function(o) {
           var idProp = o.cls_.ID;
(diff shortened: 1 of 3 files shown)


src/com/google/cloud/datastore/types.js
```diff
@@ -229,6 +229,82 @@ foam.LIB({
...
+            throw new Error('Attempt to construct datastore key from ' +
+                'unidentified object');
+          }
+          return idProp.toDatastoreKeyName(o);
+        },
+        Array: function(a) {
+          throw new Error('Multi-part keys must be derived from objects, not values');
+        },
+        Object: function(o) {
+          throw new Error('Cannot convert plain object to datastore key name');
```
mdittmer, line 250: I just realized, we should probably special case `Date` objects here; provide an implementation consistent with `foam.core.Date.toDatastoreKeyNamePart(o)` below.

src/com/google/cloud/datastore/types.js
```diff
@@ -229,6 +229,82 @@ foam.LIB({
               JSON.stringify(v));
         };
       })()
+    },
+    {
+      name: 'toDatastoreKeyName',
+      code: foam.mmethod({
+        Number: function(n) { return n.toString(); },
```
kgrgreer: n+'' is 10% faster and shorter, not that it matters too much.

src/com/google/cloud/datastore/types.js
```diff
@@ -229,6 +229,82 @@ foam.LIB({
...
+//
+// Refine properties and multi-part ids to support conversion:
+// property-on-object => datastore-key-name
+//
+
+foam.CLASS({
+  refines: 'foam.core.Property',
+
+  methods: [
+    function toDatastoreKeyName(o) {
```
kgrgreer: Document this.

src/com/google/cloud/datastore/types.js
```diff
@@ -229,6 +229,82 @@ foam.LIB({
...
+      return this.toDatastoreKeyNamePart(o);
+    },
+    function toDatastoreKeyNamePart(o) {
```
kgrgreer: Document this.

kgrgreer:
Minor comments. LGTM

PR #279, opened 2017-03-25
Delegate to foam.util when cloning property values that do not…

mdittmer:
… implement `clone()`


mdittmer:
@jingt06 FYI. @jacksonic PTAL @ tests. This change seems right, but it introduces a couple failures in index-related tests.

The particular issue @jingt06 found was that `GROUP_BY`'s use of `someSink.clone()` did not clone the array in an `ArraySink` to get properly separated groups. We could, instead have `foam.core.Array` implement `cloneProperty`, but I think that, in general, we want `foam.util.clone()` to be a reasonable implementation for property values.

PR #280, opened 2017-03-27
Clean up json_dao test and fix raceyness

mdittmer:
This is causing flake in `npm test`. @shepheb PTAL. @jacksonic FYI: Looks like this is what was causing failures in Travis.

### DIFF
--- a/src/lib/node/json_dao.js
+++ b/src/lib/node/json_dao.js
@@ -20,6 +20,8 @@ foam.CLASS({
   name: 'JSONFileDAO',
   extends: 'foam.dao.ArrayDAO',
 
+  imports: [ 'warn' ],
+
   properties: [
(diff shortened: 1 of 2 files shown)


src/lib/node/json_dao.js
```diff
@@ -20,6 +20,8 @@ foam.CLASS({
   name: 'JSONFileDAO',
   extends: 'foam.dao.ArrayDAO',
```
kgrgreer, line 21: Why ArrayDAO instead of an MDAO? With ArrayDAO, find() and remove() are going to be O(n) without a cache, and remove() is going to be O(n) even with a cache.

kgrgreer:
Added a single comment, which actually relates to old code, not this CL.

PR #281, opened 2017-03-31
Fix ManyToManyRelationshipDAO

mdittmer:
`ManyToManyRelationshipDAO` was not loading the appropriate filter predicate for `select()`ing over what it is `of`. This PR attempts to fix that logic error and add some other niceness such as `skip, ..., predicate` support and `TRUE`/`FALSE` exported by `Expressions`.

### DIFF
--- a/src/foam/dao/ManyToManyRelationshipDAO.js
+++ b/src/foam/dao/ManyToManyRelationshipDAO.js
@@ -25,17 +25,17 @@ foam.CLASS({
   documentation: 'Adapts a DAO based on a *:* Relationship.',
 
   properties: [
-    'junctionProperty', // TODO(markdittmer): Is this needed?
+    'junctionProperty',
     'joinDAOKey',
-    'targetProperty', // TODO(markdittmer): Is this needed?
+    'targetProperty',
     {
       name: 'predicate',
       documentation: `ManyToMany filtered querys are always "backward" to
-        match sourceId and source object's id.`,
+        match inverse-namd property and source object's id.`,
       getter: function() {
-        // TODO(markdittmer): Can the desired property ever be something other
-        // than SOURCE_ID?
-        return this.EQ(this.of.SOURCE_ID, this.obj.id);
+        return this.EQ(
+          this.of[foam.String.constantize(this.relationship.inverseName)],
+          this.obj.id);
       }
     }
   ],
@@ -48,7 +48,7 @@ foam.CLASS({
       return self.SUPER(self.MAP(self.junctionProperty)).then(function(map) {
         return joinDAO.select(sink, skip, limit, order, self.AND(
           predicate || self.TRUE,
-          self.IN(joinDAO.of.ID, map.delegate.a)));
+          self.IN(self.targetProperty, map.delegate.a)));
       });
     }
   ]
--- a/src/foam/dao/ManyToManyRelationshipDAO.js
+++ b/src/foam/dao/ManyToManyRelationshipDAO.js
@@ -25,20 +25,30 @@ foam.CLASS({
   documentation: 'Adapts a DAO based on a *:* Relationship.',
 
   properties: [
-    'junctionProperty',
+    'junctionProperty', // TODO(markdittmer): Is this needed?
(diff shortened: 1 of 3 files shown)


mdittmer:
@jingt06 FYI. This patch is necessary to get all-but-one test fixed on https://github.com/mdittmer/confluence/tree/relationalDAO-find-select. I'm not sure what's going on with the broken integration test.

src/foam/mlang/mlang.js
```diff
@@ -1685,6 +1687,9 @@ foam.CLASS({
       return this[name].create({ arg1: arg1, arg2: arg2 });
     },
 
+    function TRUE() { return this.True.create(); },
```
kgrgreer: Would these be better as constants instead of methods?

src/foam/dao/Relationship.js
```diff
@@ -293,6 +293,8 @@ foam.CLASS({
       */
     },
 
+    // TODO(markdittmer): Use of "forward" here doesn't seem right. No call
+    // sites pass a second parameter.
```
kgrgreer: This was used by *:1 support, which I've removed (and plan to put back, but haven't yet). I've pushed a CL to remove 'forward'.

src/foam/dao/ManyToManyRelationshipDAO.js
```diff
@@ -25,20 +25,30 @@ foam.CLASS({
...
+      }
+    }
   ],
 
   methods: [
     function select(sink, skip, limit, order, predicate) {
-      var self    = this;
-      var joinDAO = this.__context__[this.joinDAOKey];
+      var self = this;
+      var joinDAO = self.__context__[self.joinDAOKey];
```
kgrgreer: Why make this change?

src/foam/dao/ManyToManyRelationshipDAO.js
```diff
@@ -25,20 +25,30 @@ foam.CLASS({
   documentation: 'Adapts a DAO based on a *:* Relationship.',
 
   properties: [
-    'junctionProperty',
+    'junctionProperty', // TODO(markdittmer): Is this needed?
```
kgrgreer: Yes, because you don't know if this is going from target to source or from source to target.

src/foam/dao/ManyToManyRelationshipDAO.js
```diff
@@ -25,20 +25,30 @@ foam.CLASS({
...
     'joinDAOKey',
-    'targetProperty'
+    'targetProperty', // TODO(markdittmer): Is this needed?
+    {
+      name: 'predicate',
+      documentation: `ManyToMany filtered querys are always "backward" to
+        match sourceId and source object's id.`,
+      getter: function() {
+        // TODO(markdittmer): Can the desired property ever be something other
+        // than SOURCE_ID?
```
kgrgreer: Yes, *:* is symmetrical, so there's a ManyToManyRelationshipDAO going each way.

src/foam/dao/ManyToManyRelationshipDAO.js
```diff
@@ -25,20 +25,30 @@ foam.CLASS({
   documentation: 'Adapts a DAO based on a *:* Relationship.',
 
   properties: [
-    'junctionProperty',
+    'junctionProperty', // TODO(markdittmer): Is this needed?
     'joinDAOKey',
-    'targetProperty'
+    'targetProperty', // TODO(markdittmer): Is this needed?
```
kgrgreer: Used to make the correct 'predicate'.

src/foam/mlang/mlang.js
jingt06: I got 'Uncaught (in promise) TypeError: Cannot read property 'create' of undefined' in this line. Is this function used in other places without a bind?

src/foam/dao/ManyToManyRelationshipDAO.js
mdittmer: Right. This one makes sense to me know; it's needed in the `select()` implementation.

PR #282, opened 2017-04-03
RestDAOHandler-on-select(): Accept URL query or payload params.

mdittmer:
Some large query predicates were maxing out Node JS URL+headers size
limit. RestDAOHandler now accepts params in query or payload, and default
RestDAO uses payload to avoid exceeding limits on large queries.

Dependent changes:
- Tests revealed that:
  + HTTPRequest was not specifying Content-Type and Accept, even when
    configuration implied them
  + Handler was not encoding response text in JSON when responding to a JSON
    API
- HTTPRequest now sets 'Content-Type' and 'Accept' when they can be deduced
  + HTTPRequest.contentType defaults to HTTPRequest.responseType
- Extra Handler helper method looks for 'application/json' in request
  'Accept' header, and wraps textual responses in {"message":...} JSON


mdittmer:
Notes from offline review:

- We are passing `sink`s to the server
- Such `sink`s may make *writing* (statefu) changes on the server
- [GET payload should have no semantics](http://stackoverflow.com/questions/978061/http-get-with-request-body)
- Ergo: Switch `select()` to from `GET` to `PUT`

adamvy:
I believe post is better semantics than put.                                                                                                                                                                                                                                                                                                                                                    From: notifications@github.comSent: April 4, 2017 12:17 PMTo: foam2@noreply.github.comReply-to: reply@reply.github.comCc: adam.vany@gmail.com; review_requested@noreply.github.comSubject: Re: [foam-framework/foam2] RestDAOHandler-on-select(): Accept URL query or payload params. (#282)  Notes from offline review:
We are passing sinks to the serverSuch sinks may make writing (statefu) changes on the serverGET payload should have no semanticsErgo: Switch select() to from GET to PUT

—You are receiving this because your review was requested.Reply to this email directly, view it on GitHub, or mute the thread.

jacksonic:
A non-idempotent (POST) example from the near future might be:

`select(UPDATE("daoname", INCREMENT(Model.PROP_A, 3)))`

Which could be set up to delegate the work of modifying the results and replacing them into a named DAO (though exact way to support that is TBD).

adamvy:
PUT is idempotent, POST is not.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.1.2

adamvy:
Which I realize now is what you meant by saying "A non-idempotent (POST)
example"

So I think we're still in agreement that select() should be a POST :)

jacksonic:
Yessir :D

mdittmer:
Just to clarify, this conversation had nothing to do with `put()` using HTTP `PUT`, right? Because it still does with this PR...

jacksonic:
Right, my comments were only about select().

PR #284, opened 2017-04-10
Expose generic DAO tests in read-only and write-only flavours

mdittmer:


jacksonic:
Which DAOs need the read vs. write tests?

mdittmer:
The `GoogleChrome` `confluence and `webidl-diff` repositories will both need read-only DAOs. [Here's one](https://github.com/GoogleChrome/webidl-diff/blob/master/lib/org/chromium/webidl/LocalGitIDLFileDAO.js) (that currently, incorrectly, doesn't throw on write).

jacksonic (approved):

jacksonic:
Generic testing on Read-only DAOs will be tough, since the test won't be able to have hard coded data in it. But some injection of "what to expect in there" could still make for usable generic tests.

mdittmer:
Good point. Maybe this split isn't the right way to go for read-only tests.

PR #285, opened 2017-04-10
MultiPartID index support

### LINKED ISSUE #135: MDAO.find, MultiPartID toIndex()
MutliPartID can have a new toIndex() added to generate one chained tree index per property indexed.
MDAO.find_(key) takes a key that is either a value or multipart array, depending on the of:class.

In general MDAO shouldn't need to know what kind of index the MultiPartID generates. The old code that  dissected the key and called get() repeatedly relied on knowing the structure of the idIndex. MultiPartID can create a KeyIndex (that accepts an array key for .get()) that hides the internal tree details. Thus calling mdao.find(obj.id) will always work, regardless of the type of ID used. It also keeps the key's array-aware code out of MDAO and under MutliPartID.

jacksonic:
Fixes #135.

Some tests became more efficient as they gained a default index on the test model's Int property.

### DIFF
--- a/src/foam/dao/index/MultiKeyIndex.js
+++ b/src/foam/dao/index/MultiKeyIndex.js
@@ -1,0 +1,69 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+foam.CLASS({
+  package: 'foam.dao.index',
+  name: 'MultiKeyIndex',
+  extends: 'foam.dao.index.ProxyIndex',
+
+  methods: [
+    function toPrettyString(indent) {
+      var ret = '  '.repeat(indent) + 'MultiKey(' + this.$UID + ')\n';
+      ret += this.delegate.toPrettyString(indent + 1);
+      return ret;
+    }
+  ]
+
+});
+
+foam.CLASS({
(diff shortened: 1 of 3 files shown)


src/foam/dao/index/MultiKeyIndex.js
```diff
@@ -0,0 +1,69 @@
...
+  extends: 'foam.dao.index.ProxyIndexNode',
+
+  methods: [
+    function get(key) {
+      // MultiKey indexes use an array of values as the key. To find the
+      // sub-index, we must dig into the delegate's tail indexes. Abort if
+      // any key part is not found, returning undefined.
+      // TODO(jacksonic): validate key array
+      var subIndex = this.delegate.get(key[0]);
+      for (var i = 1; (i < key.length && subIndex); ++i) {
```
mdittmer: Moar spacings.

src/foam/dao/index/MultiKeyIndex.js
```diff
@@ -0,0 +1,69 @@
...
+  refines: 'foam.core.MultiPartID',
+
+  requires: [ 'foam.dao.index.MultiKeyIndex' ],
+
+  methods: [
+    function toIndex(tail) {
+      // build index, innermost to outermost
+      var index = tail;
+      var ps = this.props;
+      for (var i = ps.length - 1; i >= 0; --i) {
```
mdittmer: ditto.

test/src/lib/Index.js
```diff
@@ -950,16 +949,17 @@ describe('AutoIndex', function() {
 
     pred = pred.toDisjunctiveNormalForm();
 
-    // the results end up small enough that the first index is good enough
-    // for all subpred cases
     for ( var i = 0; i < pred.args.length; i++ ) {
       var subpred = pred.args[i];
-      idxInstance
+      var plan = idxInstance
```
mdittmer: Can we get comments in the `beforeEach` and/or in the tests explaining why we expect certain `length`s and `size()`s? It's hard to review test changes like this without much context on what's being setup and why we expect what we expect.

mdittmer (changes requested):
Just nits & request for comments/explanation.

test/src/lib/Index.js
jacksonic: Added details about what the test is doing

jacksonic:
Comments addressed, PTAL for anyone else interested

mdittmer (approved):
This LGTM

PR #286, opened 2017-04-12
Batched Cloud Datastore DAO implementation

mdittmer:
@kgrgreer PTAL. Tests coming soon.


mdittmer:
@jingt06 FYI. Should be able to use this instead of a rate-limited DAO for web API confluence bulk updates.

PR #290, opened 2017-04-15
Added Array support to SchemaImport

kirkeaton:
This adds support for converting mongoose Array properties to FOAM Array properties. Supports conversion of String, Boolean, Date, Number, and Object Reference arrays.

### DIFF
--- a/src/foam/tools/schemaimport/SchemaImport.js
+++ b/src/foam/tools/schemaimport/SchemaImport.js
@@ -47,15 +47,28 @@ foam.CLASS({
         var s = schema[key];
         var p = {};
 
+        var isArray = Array.isArray(s);
+        if ( isArray ) s.type = Array.prototype.constructor;
+
         switch ( s.type ) {
           case String:  p.class = 'String';  break;
           case Boolean: p.class = 'Boolean'; break;
           case Date:    p.class = 'Date';    break;
           case Number:  p.class = 'Double';  break;
+          case Array:
+            s = s[0];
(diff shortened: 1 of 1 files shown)


src/foam/tools/schemaimport/SchemaImport.js
```diff
@@ -47,15 +47,28 @@ foam.CLASS({
...
+
         switch ( s.type ) {
           case String:  p.class = 'String';  break;
           case Boolean: p.class = 'Boolean'; break;
           case Date:    p.class = 'Date';    break;
           case Number:  p.class = 'Double';  break;
+          case Array:
+            s = s[0];
+            p.class = 'Array';
+            switch ( s ) {
```
kgrgreer: This could just be replaced with: p.of = s;

src/foam/tools/schemaimport/SchemaImport.js
```diff
@@ -47,15 +47,28 @@ foam.CLASS({
         var s = schema[key];
         var p = {};
 
+        var isArray = Array.isArray(s);
```
adamvy, line 50: In FOAM code it might be better to use the FOAM standard library for checking the type.  foam.Array.isInstance().

FOAM stdlib.js provides a set of standard types for JS primitives that all follow the same interface.  foam.Number.isInstance() foam.String.isInstance(), etc etc.  Plus all FOAM classes also implemented isInstance(), which makes things nicely consistent.

PR #292, opened 2017-04-16
mlang: fix typos on documentation fields.

drish:


src/foam/mlang/mlang.js
```diff
@@ -642,7 +642,7 @@ foam.CLASS({
   extends: 'foam.mlang.predicate.Binary',
   implements: [ 'foam.core.Serializable' ],
 
-  documentation: 'Predicate returns true iff second arg found in first array argument.',
+  documentation: 'Predicate returns true if second arg found in first array argument.',
```
kgrgreer, line 645: iff means "if and only if", which means the implication is bi-directional.

kgrgreer:
Not a typo. iff means "if and only if", which means the implication is bi-directional.

PR #293, opened 2017-04-17
ManyToManyRelationship fixes

mcarcaso:
1. Add 'dest' property to ManyToManyRelationshipDAO that's used for the
predicate because the 'of' property should be the resulting type and not the
junction type.
2. Set the 'of' property for the ManyToManyRelationshipDAOs that are created
when doing a foam.RELATIONSHIP.
3. Use foam.dao.DAOProperty when a 1.* relationship is being created instead of
foam.core.Property.
4. Pass along targetProperty and sourceProperty in *.* relationships.


src/foam/dao/ManyToManyRelationshipDAO.js
```diff
@@ -28,13 +28,14 @@ foam.CLASS({
     'junctionProperty',
     'joinDAOKey',
     'targetProperty',
+    'dest',
```
adamvy: I know I recommend "dest" first, but now that I think about it, "junctionCls" would be a more appropriate name.

src/foam/dao/Relationship.js
```diff
@@ -254,18 +257,21 @@ foam.CLASS({
...
           inverseName: 'targetId',
           sourceDAOKey: this.targetDAOKey,
           targetDAOKey: this.junctionDAOKey,
-          relationshipDAOFactory: function(source) {
+          relationshipDAOFactory: function(s) {
             return foam.dao.ManyToManyRelationshipDAO.create({
-              obj: source,
+              dest: jModel,
+              obj: s,
+              of: source.id,
```
adamvy: You renamed "source" to "s" but didn't change it on this line.  Why not just leave it as "source" so the diff is smaller.

mcarcaso: Oh meant to ask about this. So in the previous code, the source var from the method is clobbered by the source var that's passed into this function and I wasn't sure if I should be using that source.

adamvy: ah I see.  Yeah i think you did the right thing then

src/foam/dao/ManyToManyRelationshipDAO.js
mcarcaso: done

PR #296, opened 2017-04-18
NoSelectAllDAO - improving validation and testing order and predicate

marcogbarcellos:
- before inserting the required predicates `True` and `False`, I was getting `cannot set isInstance of undefined`. 

- created unit test to validate predicate and orderby.

### DIFF
--- a/src/foam/dao/NoSelectAllDAO.js
+++ b/src/foam/dao/NoSelectAllDAO.js
@@ -31,8 +31,8 @@ foam.CLASS({
   methods: [
     function select(sink, skip, limit, order, predicate) {
         if ( predicate &&
-             ( ! this.True.isInstance(predicate) &&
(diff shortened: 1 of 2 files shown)


kgrgreer:
Something else is wrong. The " this.True && " check shouldn't be required since True and False are required on lines 25 and 26.

adamvy:
Had you pulled recently when you were getting those errors?  I added the requires earlier tonight.

kgrgreer:
Hahaha, that would explain it. You don't pull for a couple hours and someone goes and fixes the code ahead of you.

marcogbarcellos:
You're right guys! perhaps only the test worths it. thanks.

kgrgreer:
Update your patch to just contain the test and I'll merge it.

PR #297, opened 2017-04-18
mlang: implements min()

drish:
this PR:

- implements the MIN() functionality for mlang.
- adds tests for MIN().

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -1613,9 +1613,7 @@ foam.CLASS({
 
   methods: [
     function put(obj) {
-      if ( ! this.hasOwnProperty('value') ) {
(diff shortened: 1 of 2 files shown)


src/foam/mlang/mlang.js
```diff
@@ -1592,6 +1592,35 @@ foam.CLASS({
...
+  properties: [
+    {
+      name: 'value',
+      value: 0
+    }
+  ],
+
+  methods: [
+    function put(obj) {
+      if ( ! this.hasOwnProperty('value') ) {
```
kgrgreer: If you OR'ed the two if clauses then you wouldn't need to duplicate the body twice.

drish: agreed, will do the same for the Max function too, in another pr.

PR #306, opened 2017-05-02
Added support for Short data type

kirkeaton:
1. Added Short type to foam/core/types.js
2. Added refinement for Short data type
3. Added AbstractShortPropertyInfo
4. Added ShortParser
5. Added ShortParser to AnyParser


kgrgreer:
Having an Alt() parser with Long, Int, and Short looks broken to me, but it was already broken before your change. The three are syntactically the same, so it should be the same parser in all cases and just coersed to an int or short when necessary. Will fix.

PR #308, opened 2017-05-02
NotFoundPlan should have cost 0, not the max

marcogbarcellos:
As I talked to @kgrgreer , the `cost` property of the NotFoundPlan is zero.

### DIFF
--- a/src/foam/dao/index/Plan.js
+++ b/src/foam/dao/index/Plan.js
@@ -42,7 +42,7 @@ foam.CLASS({
   axioms: [ foam.pattern.Singleton.create() ],
 
   properties: [
-    { name: 'cost', value: Number.MAX_VALUE }
+    { name: 'cost', value: 0 }
   ],
 
   methods: [


jacksonic:
What about an Alt with some subplans that fail to find and one that does? The particular case I'm thinking about is Autoindex partially populating indexes.

kgrgreer:
In FOAM1, the NotFoundPlan meant that I couldn't find it because I know it doesn't exist. Perhaps a better name would have been "DoesntExist'. This is different than NoPlan, which means that I don't have a plan for how to perform this query, which is what I think you (jackson) are talking about.

jacksonic:
Yes that's what I was thinking of. Much easier to optimize when we can stop after the first not-found indication.

kgrgreer:
Okay, so do you approve of the change? It doesn't break any tests.

jacksonic:
Yes, merge it

PR #309, opened 2017-05-03
MDAO.find(): Support object (instance of this.of) as argument

mdittmer:


adamvy:
Sounds good, but I find the argument name objOrId ugly.  But I have no better idea.

src/foam/dao/ArrayDAO.js
```diff
@@ -29,6 +29,14 @@ foam.CLASS({
 
   properties: [
     {
+      class: 'Class',
+      name: 'of',
+      factory: function() {
+        if (this.array.length === 0) return foam.lookup('foam.core.FObject');
```
kgrgreer: Why the lookup()?

mdittmer: I thought it was standard practice. I suppose it should be `imports: ['lookup']` and `this.lookup()`.

mdittmer:
@adamvy what about `key` as the arg, then set `id = <maybe extract id from obj>`? I don't think it's as clear as `objOrId`, but it's shorter. WDYT?

src/foam/dao/ArrayDAO.js
mdittmer: or not: https://travis-ci.org/foam-framework/foam2/jobs/228443299. What's the right way to pull in the `lookup()` function from the context?

adamvy: this.lookup is fine without an import currently.

I think Kevin was implying that you could just reference foam.core.FObject globally directly, since it will always be defined.  But a lookup doesn't really hurt I don't think.

adamvy:
I'd be okay with it.

PR #311, opened 2017-05-03
Test and implement: Logger interface and SimpleLogger for Javascript

mdittmer:
Upstreaming implementation from https://github.com/GoogleChrome/webidl-diff/pull/34


src/foam/log/SimpleLogger.js
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+foam.CLASS({
+  package: 'foam.log',
```
kgrgreer: Let's move this to foam.nanos.log, since I want to put all runtime services there.

src/foam/log/SimpleLogger.js
```diff
@@ -0,0 +1,82 @@
...
+  name: 'SimpleLogger',
```
kgrgreer: Naming this something like ConsoleLogger would be more meaningful.

src/foam/log/LogLevel.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+  ],
+
+  values: [
+    {
+      name: 'DEBUG',
+      shortName: 'DEBG',
+      label: 'Debug',
+      consoleMethodName: 'debug'
+    },
```
bshepherdson: What happened to `LOG`? The code below uses `.log()` alongside `.info()`. Are they synonyms? It should say so, in that case.

src/foam/log/SimpleLogger.js
bshepherdson: +1

src/foam/log/LogLevel.js
mdittmer: The test is wrong. It shouldn't be invoking `.log()`.

mdittmer: Wait. I was mistaken. There was a whole conversation about `log` as a log level being a weirdness of Javascript when landing this in the other repo. They `.log()` should be documented as a synonym for `.info()`. That's a documentation error.

mdittmer:
@shepheb please take another look.

PR #313, opened 2017-05-04
removing unnecessary if

marcogbarcellos:
I think this `if` was included for debugging purposes. I didn't find a good reason for keeping it. Am I wrong? thanks.


adamvy:
LGTM

PR #315, opened 2017-05-05
Add abstract getWidth function to AbstractStringPropertyInfo.java

kirkeaton:


src/foam/java/refinements.js
```diff
@@ -537,7 +537,9 @@ foam.CLASS({
         name: 'getWidth',
         visibility: 'public',
         type: 'int',
-        body: 'return ' + this.width + ';'
+        body: ( this.width > 0 ) ?
```
kgrgreer: See my last CL for a better way to do this. This isn't DRY, since if some other part of the code assumes a different default value, then the two will be out of sync. Just remove this change and keep the one in AbstractStringPropertyInfo.

PR #316, opened 2017-05-05
FilteredDAO: Test and implement filtered find()

mdittmer:


src/foam/dao/AbstractDAO.js
```diff
@@ -284,6 +284,17 @@ foam.CLASS({
   ],
 
   methods: [
+    function find(key) {
+      if ( this.of && this.of.isInstance(key) ) {
+        if ( this.predicate.f(key) ) return this.delegate.find(key);
```
kgrgreer: I'm not a fan of this style of if/else, also if both branches are returning anyway, just do:
return this.predicate.f(key) ? this.delegate.find(key) : Promose.resolve(null);

kgrgreer: Also, this logic is a security bug. Say I'm given a filtered DAO where I only get emails which are to or from me. I can craft an email with the id of some email that doesn't belong to me, but set to to or from to be me and the predicate will pass and give me an email which doesn't belong to me. The predicate needs to be applied to the actual found object.

src/foam/dao/AbstractDAO.js
```diff
@@ -284,6 +284,12 @@ foam.CLASS({
   ],
 
   methods: [
+    function find(key) {
```
adamvy, line 287: Would it make more sense to pass predicate as an extra parameter to find() the same way we do on .select()?

Slight advantage here is if you're going over the network we don't pay to send objects that you're going to filter out on the client.  Less important for find() since you're getting only a single object.

mdittmer: I think that would make sense, but I don't have time to implement it for `FilteredDAO` + all our network DAOs right now. Maybe file an issue?

adamvy: File your own issues :p

Or at least leave a TODO in the code.

adamvy: We don't have that many DAO implementations,

PR #317, opened 2017-05-05
Many-to-many DAO: Code clean up and implement find()

mdittmer:
This creates a copy of the 7-line `FilteredDAO.select()` implementation so that it can be referred to directly by `find()` (where `SUPER()` will not refer to the correct function). See comments in `find()` for rationale. I felt this simpler and clearer than storing "the super implementation of select" with, say `this.__proto__._proto__.select` or `this.lookup('foam.dao.FilteredDAO').getAxiomByName('select').code`.


mdittmer:
I'm merging this to unbreak Confluence's build. Another acceptable implementation (against the new relationships API) will land in #318. @adamvy FYI.

PR #318, opened 2017-05-05
Reactions, InitAgents, Comics, ContextMultipleInheritence, Nested Exports and Relationships

adamvy:
Lots of different changes together in this pull request.  Some are self contained features, each of which will be described in detail.  There are other scattered fixes throughout.

I'm particularly looking for review of Reactions, Context Multiple Inheritence, Init Agents and Relationships, but feel free to comment on anything else you see.

# Reactions
 - New axiom for declarative event bindings
 - Implemented in foam/core/Reaction.js

## Example

```
properties: [
  {
    class: 'FObjectProperty',
    of: 'foam.comics.DAOController',
    name: 'data'
  },
],

reactions: [
  [ 'data', 'action,create', 'onCreate' ]
],

listeners: [
  function onCreate(e) {
    // Show create view!
  }
]
```

This is roughly equivalent to adding a postSet to .data which automatically subscribes to

`this.data.sub('action', 'create', this.onCreate);`

and updates this subscription when data changes.

# Init Agents
 - Axioms are given a chance to run code at object initialization time.
 - Used by Reactions to initialize their event bindings.
 - Implemented by by changes to FObject.js

# COMICS
 - Lots of refactoring happening, comics is now able to comfortably edit objects with 1:* and *:* relationships.
 - Changed main entry point from DAOControllerView to BrowserView and InlineBrowserView.
 - DAOControllerView is now just a detail view of a DAOController as the name would imply, to create a view of a DAO use foam.comics.BrowserView ( with reciprical search ) or foam.comics.InlineBrowserView ( without reciprical search) .

# Context Multiple Inheritence
 - Implemented in foam/core/ContextMultipleInheritence.js
 - Contexts can now handle a limited form of multiple inheritence.
 - context.createSubContext() can now either take a Map or an FObject, if a Map is passed, a sub context is created exactly how it was before, but if an FObject is passed, the exports of that FObject are mixed into the context hierarchy.

## Example
```
foam.CLASS({
  name: 'ContextA',
  exports: [ 'a' ],
  properties: [ 'a' ]
});

foam.CLASS({
  name: 'ContextB',
  exports: [ 'b' ],
  properties: [ 'b' ]
});

foam.CLASS({
  name: 'ContextC',
  exports: [ 'b', 'c' ],
  properties: [ 'b' ]
});

var a = ContextA.create({ a: 1 });
var x = a.__subContext__;

var b = ContextB.create({ b: 2 }, a);
var y = b.__subContext__;

var c = ContextC.create({ b: 3, c: 4 });
var z = b.__subContext__.createSubContext(c);

console.log(z.a, z.b, z.c);
// OUTPUT: 1 3 4
// ContextC's exports are mixed into b's subContext.
```

 - foam.u2.Stack.push() was updated to take an optional parent object as a second parameter.  This allows elements on the stack to export things via the context to views that will be subsequently pushed onto the stack.

# Nested Exports
 - You can now export properties of your properties.

## Example

```
foam.CLASS({
  name: 'Abc',
  properties: [ 'a' ]
});

foam.CLASS({
  name: 'Controller',
  exports: [
    'abc.a as a'
  ],
  properties: [
    {
      name: 'abc',
      factory: function() { return Abc.create({ a: 12 }); }
    }
  ]
});

var c = Controller.create();

console.log(c.__subContext__.a);
// OUTPUT: 12

c.abc.a = 24;

console.log(c.__subContext__.a);
// OUTPUT: 24
```

# Relationships rework
 - Many to Many relationships were reworked pretty extensively.
 - The relationship property intalled onto an object is no longer a relationship DAO.  Instead it is a relationship object which has a few properties including the related DAO.

```
foam.CLASS({
  package: 'foam.dao',
  name: 'RelationshipPropertyValue',
  properties: [
    {
      name: 'dao'
    },
    {
      name: 'junctionDAO'
    },
    {
      name: 'targetDAO'
    },
  ],
  methods: [
    function add(obj) {},
    function remove(obj) {}
  ]
});
```

### Many to many

```
foam.CLASS({
  name: 'User'
  properties: [ 'id' ]
});

foam.CLASS({
  name: 'Conversation'
  properties: [ 'id' ]
});

foam.RELATIONSHIP({
  sourceModel: 'User',
  targetModel: 'Conversation',
  cardinality: '*:*',
});

...

// Select all Conversations that the user is a member of.
user.conversations.dao.select()

// Select all Conversations
user.conversations.targetDAO.select()

// Select all UserConversationJunction objects
user.conversations.junctionDAO.select()

// Associate a particular conversation with this user
user.conversations.add(someConversationObject)

// De-associate this user with a particular conversation
user.conversations.remove(someConversationObject)


user.conversations.dao.put(someConversationObject) // throws
user.conversations.dao.remove(someConversationObject) // throws
user.conversations.dao.removeAll() // throws


// Select all users in this converstaion
conversation.users.dao.select()

// Select all conversations
conversation.users.targetDAO.select()

// Select all UserConversationJunction objects.
conversation.users.junctionDAO.select()

// Associate this conversation with a particular user.
conversation.users.add(someUserObject)

// De-associate this conversation with a particular user.
conversation.users.remove(someUserObject)
```

### One to many
```
foam.CLASS({
  name: Parent',
  properties: [ 'id' ]
});

foam.CLASS({
  name: 'Child',
  properties: [ 'id' ]
});

foam.RELATIONSHIP({
  sourceModel: 'Parent',
  targetModel: 'Child',
  cardinality: '1:*'
});

...

// Select all Children objects that are related to parent.
parent.children.dao.select();

// Select all Children objects.
parent.children.targetDAO.select();

// junctionDAO is null as there is no junction required for a one to many relationship
parent.children.junctionDAO == null;

parent.children.add(someChildObject) // throws
parent.children.remove(someChildObject) // throws

// Create or update a child object, and relate it to parent.
parent.children.dao.put(someChildObject);

// Remove someChildObject from the DAO
parent.children.dao.remove(someChildObject);

```

# Expected Regressions

The main API change in this PR is the relationships interface.  Most existing code can probably update just by chaging references from obj.children. to obj.children.dao.

*:* relationships require calling .add() on the relationship object, or writing into the junctionDAO manually to add related objects.  If you need to create a new object that you then want to associate with the relationship, you must call something like `obj.related.targetDAO.put(RelatedObject.create()).then(function(o) { obj.related.add(o); })`

Users of COMICS will probably want to change uses of DAOControllerView to BrowserView.  DAOControllerView now expects it's data to be a DAOController.


adamvy:
I signed it!

adamvy:
Jesus christ googlebot, at least point to the commits you think are the problem.

adamvy:
So pretty sure googlebot is complaining because the commits were authored as adam@vany.ca but I signed the CLA and sent the PR as adam.vany@gmail.com

src/foam/core/ImportsExports.js
```diff
@@ -163,34 +163,45 @@ foam.CLASS({
...
+          var path = b.key.split('.');
+
+          var a = this.cls_.getAxiomByName(path[0]);
+
+          foam.assert(!!a, 'Unknown axiom: "', path[0], '" in model: ',
+                      this.cls_.id, ", trying to export: '", b.key, "'");
+
+          // Axioms have an option of wrapping a value for export.
+          // This could be used to bind a method to 'this', for example.
+          var e = a.exportAs ? a.exportAs(this, path.slice(1)) : this[path[0]];
```
mdittmer, line 182: In combination with `Property`'s `exportAs()` (which returns a `Slot`), doesn't this mean that `exports: ['foo']` exports the current value of `this.foo`, but `exports: ['foo.bar']` exports a `Slot` that fers to `this.foo.bar`?

src/foam/dao/ManyToManyRelationshipDAO.js
```diff
@@ -18,51 +18,38 @@
...
   package: 'foam.dao',
   name: 'ManyToManyRelationshipDAO',
-  extends: 'foam.dao.RelationshipDAO',
+  extends: 'foam.dao.ProxyDAO',
 
   implements: [ 'foam.mlang.Expressions' ],
 
   documentation: 'Adapts a DAO based on a *:* Relationship.',
 
   properties: [
```
mdittmer, line 27: Can we please document these? Every time I go to debug relationships I get confused about "is this 'target' always 'target', or is it 'target, unless some state indicates that this relationship is inverted'?", and "this 'junction*' property, does it refer to the whole junction DAO? A filtered DAO with only `this.obj`'s junctions?", likewise for source/target DAOs, etc..

src/foam/dao/ManyToManyRelationshipDAO.js
```diff
@@ -18,51 +18,38 @@
...
-      var self = this;
-      return self.__context__[foam.String.daoize(self.of.name)].put(obj)
-          .then(function(obj) {
-            return self.delegate.put(self.relationship.adaptTarget(self.obj, obj))
-          })
-    },
     function find(id) {
-      return this.joinDAO.find(id);
+      var self = this;
+      return self.junctionDAO.select(self.MAP(self.junctionProperty)).then(function(map) {
```
mdittmer: This can be really expensive if the number of `this.obj` -> `targetObj` relations is large. That's why I was constructing a key to `find()` over delegate in https://github.com/foam-framework/foam2/pull/317/files#diff-6cb3faaa1e42ab1c377ce592ca9cb453R63. @kgrgreer you mentioned that the linked code isn't quite right. Is there state that we currently store (or could store) to use the find-on-delegate strategy correctly?

src/foam/dao/ManyToManyRelationshipDAO.js
mdittmer: ... we should also include documentation about what this DAO is `of`, and what it's `delegate` is `of`. They used to be `of` different things (which was a bit confusing, but workable).

src/foam/dao/Relationship.js
```diff
@@ -23,7 +23,9 @@ foam.CLASS({
   documentation: 'An Axiom for defining Relationships between models.',
 
   requires: [
-    'foam.dao.RelationshipDAO'
+    'foam.dao.RelationshipDAO',
+    'foam.dao.ManyToManyRelationshipDAO',
+    'foam.dao.ReadOnlyDAO',
   ],
 
   properties: [
```
mdittmer, line 31: Ditto: Even if it's terse, we should take this opportunity to add documentation to these properties.

src/foam/dao/Relationship.js
```diff
@@ -149,28 +144,39 @@ foam.CLASS({
...
+      var sourceDAOKey   = this.sourceDAOKey;
+      var targetDAOKey   = this.targetDAOKey;
 
       if ( cardinality === '1:*' ) {
         if ( ! sourceProps.length ) {
           sourceProps = [
-            foam.dao.RelationshipDAOProperty.create({
+            foam.dao.RelationshipProperty.create({
               name: forwardName,
+              cloneProperty: function(value, map) {},
```
mdittmer: The need for this is non-obvious. Can we get a comment here?

src/foam/dao/ManyToManyRelationshipDAO.js
```diff
@@ -18,51 +18,38 @@
 foam.CLASS({
   package: 'foam.dao',
   name: 'ManyToManyRelationshipDAO',
-  extends: 'foam.dao.RelationshipDAO',
+  extends: 'foam.dao.ProxyDAO',
```
mdittmer, line 21: This drops storing a predicate (and superclass implementation to filter by the predicate). Does that potentially eliminate the possibility of the `find()` implementation I propose below?

src/foam/dao/Relationship.js
```diff
@@ -324,3 +320,57 @@ foam.LIB({
     }
   ]
 });
+
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'RelationshipPropertyValue',
+  properties: [
```
mdittmer, line 329: Documentation: Are `DAOs` "complete" DAOs of each type, or filtered according to the source object (`o` in `o.relationshipPropertyName`)?

mdittmer (changes requested):
Review of `foo.bar` exports and relationships. Can we get some automated tests of relationships? As far as I know we just have browser-based console-prints-examples test.

src/foam/core/ImportsExports.js
adamvy: exports: [ 'foo' ] exports this.foo$ as a slot, we always exports slots when we can.

src/foam/dao/ManyToManyRelationshipDAO.js
adamvy: It _can_ be really expensive, but it has the added benefit of being correct :).

I pushed a different version that does a .find() on the junction to see if a relation exists, then moves on to the delegate, should be faster.

mdittmer:
Tests in #321 now passing. Just documentation requests (and maybe tests for `1:*` relationships?) outstanding.

src/foam/dao/ManyToManyRelationshipDAO.js
mdittmer: Thanks!

src/foam/core/ImportsExports.js
mdittmer: Yes. I remember that now. It is still inconsistent though that 

`foo` => `foo` (value), `foo$` (slot)
and
`foo.bar` => `foo` (slot)

Isn't that what happens here?

adamvy: No.

Consider that in both cases of exporting `foo` and `foo.bar` we call `exportAs` on the `foo` axiom.  So why would this become inconsistent?

The code here isn't responsible for making sure both foo and foo$ are available in the context.  That's done in createSubContext.  All we do here is let the axiom decide what it wants to export itself as and choose the key we want to export with.

kgrgreer:
I think the reaction subject syntax should change from using commas to perios, ie. from:

reactions: [
  [ 'data', 'action,create', 'onCreate' ]
],

to:

reactions: [
  [ 'data', 'action.create', 'onCreate' ]
],

Multiple inheritance for contexts is good, but the access point: providing an FObject to createSubContext() seems non-obvious / odd. Do you think this is necessary, vs just creating a MultiContext object explicitly, possibly with ability to provide more than two parent contexts at once?

Would:
exports: [ 'abc.a as a' ],
be the same as: 
exports: [ 'abc.a' ] ?

Could I also do:
exports: [ 'abc.*' ] ?

Why do you introduce the extra level of indirection with the RelationshipProperty for 1:* Relationships? Didn't the old design already work well for 1:* Relationships?

src/foam/blob/Blob.js
```diff
@@ -560,7 +565,7 @@ foam.CLASS({
     function put(file) {
       var id = this.nextId++;
       this.blobs[id] = file;
-      return Promise.resolve(id);//this.NumberedFile.create({ id: id }));
+      return Promise.resolve(self.IdentifiedBlob.create({ id }));
```
kgrgreer: Where is self defined?

src/foam/comics/DAOController.js
```diff
@@ -20,45 +20,83 @@ foam.CLASS({
...
+      hidden: true
+    },
+    {
+      name: 'mode',
+      value: 'browse'
     }
   ],
 
+  constants: {
+    MODE_SELECT: 'select',
```
kgrgreer: Can you document these two modes.

src/foam/comics/DAOControllerView.js
```diff
@@ -21,52 +21,86 @@ foam.CLASS({
...
-              start(this.DAOController.PREDICATE, {dao$: this.data$}).end().
-            end().
-            start('td').style({ 'vertical-align': 'top', 'width': '100%' }).add(this.DAOController.FILTERED_DAO).end().
-          end().
         start('tr').
-          show(this.mode$.map(function(m) { return m == foam.u2.DisplayMode.RW; })).
-          start('td').end().start('td').add(this.DAOController.CREATE).end().
+          start('td').add(this.cls.PREDICATE).end().
+          start('td').style({ 'vertical-align': 'top', 'width': '100%' }).
+            add(this.cls.FILTERED_DAO).end().
```
kgrgreer: Line the end() up with the start() if they aren't on the same line, otherwise it looks like a bug at first.

src/foam/comics/DAOControllerView.js
```diff
@@ -21,52 +21,86 @@ foam.CLASS({
...
         end().
-        endContext();
+        start('tr').
+          show(this.mode$.map(function(m) { return m == foam.u2.DisplayMode.RW; })).
+        start('td').end().start('td').
```
kgrgreer: Odd indenting.

src/foam/core/ContextMultipleInheritence.js
```diff
@@ -0,0 +1,32 @@
...
+          var obj = opt_args;
+
+          var exports = obj.cls_.getAxiomsByClass(foam.core.Export);
+
+          if ( ! exports ) return X;
+
+          opt_args = exports[0].getExportMap.call(obj);
+        }
+
+        return this.__context__.createSubContext.call(X, opt_args, opt_name);
```
kgrgreer, line 24: So, this is actually more like a mix-in than inheritance?

src/foam/core/FObject.js
```diff
@@ -279,6 +284,17 @@ foam.LIB({
       return as;
     },
 
+    function getInitAgents() {
```
kgrgreer, line 287: maybe renamed getInitAgents_().

src/foam/blob/Blob.js
adamvy: nowhere, fixed.

src/foam/comics/DAOControllerView.js
adamvy: fixed

src/foam/comics/DAOControllerView.js
adamvy: fixed

src/foam/core/FObject.js
```diff
@@ -959,6 +975,10 @@ foam.CLASS({
       // Distinguish between prototypes and instances.
       return this.cls_.id + (
           this.cls_.prototype === this ? 'Proto' : '');
+    },
+
+    function dot(name) {
```
kgrgreer, line 980: document

src/foam/core/ContextMultipleInheritence.js
adamvy: yeah more like a mix-in, but I think I want to re-do it to have real multiple inheritence.  I might do that at a later date, as this is sufficient for my needs at the moment.

src/foam/core/FObject.js
adamvy: done

src/foam/core/Property.js
```diff
@@ -593,9 +593,16 @@ foam.CLASS({
       return prop;
     },
 
-    function exportAs(obj) {
+    function exportAs(obj, path) {
```
kgrgreer: What's the use-case for the path?

adamvy: for exporting 'abc.a.b.c', we walk path to create this.abc$.dot(a).dot(b).dot(c)

src/foam/core/Reaction.js
```diff
@@ -0,0 +1,72 @@
...
+      class: 'StringArray',
+      name: 'topic'
+    },
+    {
+      name: 'listener'
+    },
+  ],
+  methods: [
+    function initObject(obj) {
+      var path = this.target.split('.');
```
kgrgreer: Would be more efficient if target were a StringArray with an adapt method to do the split if you just wanted to provide a string rather than an array. Would be faster to init an create fewer objects.

src/foam/core/Reaction.js
```diff
@@ -0,0 +1,72 @@
...
+      var slot = obj;
+
+      for ( var i = 0 ; i < path.length ; i++ ) {
+        slot = slot.dot(path[i]);
+      }
+
+      var listener = obj[this.listener];
+      var topic = this.topic;
+
+      if ( topic.length ) {
```
kgrgreer: When would we not want to have a topic?

src/foam/u2/stack/MultiStackView.js
```diff
@@ -37,7 +37,11 @@ foam.CLASS({
     },
     function wrap(s) {
       return this.slot(function(s) {
-        return foam.u2.ViewSpec.createView(s, null, this, this.__subSubContext__);
+        var view = s[0];
+        var parent = s[1];
+
+        var X = parent ? this.__subSubContext__.createSubContext(parent) : this.__subSubContext__;
```
kgrgreer: Document this.

src/foam/u2/stack/Stack.js
```diff
@@ -62,12 +62,12 @@ foam.CLASS({
       return i < 0 ? this.stack_[this.pos + i + 1] : this.stack_[i];
     },
 
-    function push(v) {
+    function push(v, parent) {
```
kgrgreer, line 65: Document usage.

src/foam/u2/stack/StackView.js
```diff
@@ -47,7 +47,12 @@ foam.CLASS({
         this.add(this.Stack.BACK, this.Stack.FORWARD);
 
       this.add(this.slot(function(s) {
-        return foam.u2.ViewSpec.createView(s, null, this, this.__subSubContext__);
+        var view = s[0];
+        var parent = s[1];
+        var X = parent ? this.__subSubContext__.createSubContext(parent) : this.__subSubContext__;
```
kgrgreer: Why?

adamvy: If I push something onto the stack, I want it to be able to pick up things that I have exported, but I also want it to be in a sub-context of its parent node like always.

This was the motivation for doing Context Multiple Inheritence.

For instance DAOControllerView exports 'data.dao as dao', and DAOUpdateControllerView imports 'dao'

Without this, DAOUpdateController would only pick up the context of the stack view it's being rendered in and be unable to find it's 'dao' import.

src/foam/core/Reaction.js
adamvy: If I want to react every time .data changes to a new value.

src/foam/core/ContextMultipleInheritence.js
kgrgreer: Okay, I'm also not keen on the instantiation method, but I'll let it stand for now.

src/foam/core/Property.js
kgrgreer: So it's really the source path and not the export path.

src/foam/u2/stack/StackView.js
kgrgreer: Okay, then add a comment to explain.

src/foam/comics/DAOController.js
```diff
@@ -20,45 +20,82 @@ foam.CLASS({
...
+      hidden: true
+    },
+    {
+      name: 'mode',
+      value: 'browse'
     }
   ],
 
+  constants: {
+    MODE_SELECT: 'select'
```
kgrgreer: Document. We also have the foam.u2.ControllerMode enum, so is it confusing to have two controller modes?

src/foam/core/Property.js
adamvy: Yes it's really the sourcePath.

src/foam/u2/stack/StackView.js
adamvy: done

src/foam/comics/DAOController.js
adamvy: Removed it and added a documented boolean instead.

src/foam/core/Property.js
adamvy: Renamed to sourcePath

adamvy:
I added a TODO about doing a MultiContext.

I'm going to submit this later today.  I will not be documenting the remaining things that need documentation right now, however any corporate entities are welcome to sponsor documentation if they need it for their purposes.

PR #320, opened 2017-05-06
test: add basic test for journaled dao + nodeFileDao

drish:


test/src/lib/node/JournaledDAO.js
```diff
@@ -0,0 +1,75 @@
...
+      name: 'Car',
+      properties: [
+        { name: 'id' },
+        { name: 'color' }
+      ]
+    });
+
+    cars = foam.dao.ArrayDAO.create();
+    for ( var i = 0; i < 50; ++i ) {
+      cars.put(
```
kgrgreer: Put to the jDAO instead, otherwise you put()'s wont be journalled.

test/src/lib/node/JournaledDAO.js
```diff
@@ -0,0 +1,75 @@
...
+  it('should persist to file', function(done) {
+
+    cars.select({ put: function(sub, car) { jDAO.put(car)}, eof: function () {} })
+    .then(function() {
+      fs.readFile(filename, 'utf8', function (err, data) {
+       if (err) {
+         return done(err);
+       }
+       var lines = data.split('\n').filter(Boolean);
+       expect(lines.length).toBe(50);
```
kgrgreer: You're testing an assumed implementation detail, not the contract of the DAO interface. It could pretty print the content to be more than 50 lines for write in some binary format that puts all data on one line. Create a second JournaledDAO and see if it has the data you put in the first one. You care that it's a valid DAO, not how its implemented.

test/src/lib/node/JournaledDAO.js
```diff
@@ -71,5 +76,32 @@ describe('JournaledDAO', function() {
...
+
+    foam.dao.JournaledDAO.create({
+      delegate: cars,
+      journal: foam.dao.NodeFileJournal.create({
+        fd: fs.openSync(filename, 'r+')
+      })
+    })
+    .select({
+      put: function(sub, data) {
+        cars.put(data);
```
kgrgreer: This should be redundant. The JournaledDAO will already copy all of the journal data into the delegate (cars) when it's created.

test/src/lib/node/JournaledDAO.js
```diff
@@ -58,7 +62,8 @@ describe('JournaledDAO', function() {
 
   it('should persist to file', function(done) {
 
-    cars.select({ put: function(sub, car) { jDAO.put(car)}, eof: function () {} })
+    inboundDAO
```
kgrgreer: I disagree that this whole block should exist.

PR #321, opened 2017-05-07
Many-to-many tests + small bug fix

mdittmer:
@adamvy PTAL. I fixed one small error but 4/8 tests are currently failing. I think `foo.bars.dao.select()` isn't filtering to include only `bars` associated with `foo`. May be related to dropping inheritance from `RelationshipDAO` and (transitively) `FilteredDAO`.


mdittmer:
Moved to #348

PR #322, opened 2017-05-08
Enum Support - Java Code Generation

TanayParikh:
- Added Enum code generation support for Java
- Updated src/foam/java/Enum.js starter file
- Added refinements in src/foam/java/refinements.js
- Updated  src/foam/java/Class.js and src/foam/java/Method.js to facilitate enum definitions


kgrgreer:
@adamvy Can you please review. Thanks

adamvy:
@kgrgreer Only if you review #318 :)

src/foam/java/Enum.js
```diff
@@ -1,73 +1,55 @@
...
     else {
-%>  <%= value.name %>(<%= value.index %>, <%= value.javaValue %>, <%= labelForValue(value) %>) <%
+%>  <%= value.name %>(<%= i %>, <%= value.definition.ordinal %>, <%= labelForValue(value) %>) <%
     }
     if ( i == this.values.length - 1 ) {%>;<%}
     else {%>,<%}
   }
 %>
 
   private final int index_;
```
adamvy: Do we really need index_ and ordinal_ ?  Ordinal should be sufficient.

src/foam/java/Enum.js
```diff
@@ -1,73 +1,55 @@
-// Test with: foam.u2.ControllerMode.model_.javaSource();
+// Defines template for the EnumModel
```
adamvy: Overall I wonder if we can do some better code sharing by making a foam.java.Enum class, and having it extend foam.java.Class

An Enum is in essence a Class with a fixed number of implementations.  If we extend foam.java.Class and update the outputJava method to output the VALUES, as well as whatever tweaks we need to match java enum syntax, we should also get methods/fields for free.


I would start by defining foam.java.Enum as a foam.CLASS() that extends foam.java.Class, and then trying to get the output correct.  Then refine buildJavaClass() onto foam.core.EnumModel, which should build a foam.java.Enum.

adamvy (changes requested):

src/foam/java/Enum.js
TanayParikh: Thanks, that's a good point. I've removed the index property.

src/foam/java/Enum.js
```diff
@@ -1,97 +1,64 @@
...
-      `
-    }
-  ]
-})
-
-/*
-package <%= this.package %>;
 <%
 var labelForValue = function(value) {
   if (this.androidResource) {
```
kgrgreer: Do we need this now? When we do, can we integrate it with FOAM's Message axiom support and then integrate the message support with Android Resources.

TanayParikh: The androidResource is directly from the FOAM 1 enum template. Should I just go ahead and remove it?

kgrgreer: Yes, just remove for now.

TanayParikh: Removed

src/foam/java/Enum.js
TanayParikh: Hi Adam,

I've made foam.java.Enum and extended it from foam.java.Class. I've also updated the general code generation logic to move away from templates and more along the lines of the FOAM2 structure.

src/foam/java/Enum.js
```diff
@@ -1,106 +1,166 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
 foam.CLASS({
-  refines: 'foam.core.EnumModel',
+  package: 'foam.java',
+  name: 'EnumModel',
```
adamvy: This can just be called Enum, same way foam.java.Class is call Class not ClassModel.

src/foam/java/Enum.js
```diff
@@ -1,106 +1,166 @@
...
+  extends: 'foam.java.Class',
 
-  methods: [
+  properties: [
+    ['javaType', 'java.lang.Enum'],
```
adamvy: Are these used anywhere?  I'm pretty sure these properties should be refined on foam.core.Enum

src/foam/java/Enum.js
```diff
@@ -1,106 +1,166 @@
...
+      o.indent();
+      o.out('}\n');
+
+      this.extras.forEach(function(c) { o.out(c, '\n'); });
+      o.decreaseIndent();
+      o.indent();
+      o.out('}');
+    },
+
+    function toJavaSource() {
```
adamvy: This should be inherited from foam.java.Class

src/foam/java/Enum.js
adamvy: Ah reading further, you did refine them onto foam.core.Enum, which is good, these properties here should probably be deleted.

src/foam/java/refinements.js
```diff
@@ -483,13 +484,66 @@ foam.CLASS({
...
+    ['javaType', 'java.lang.Enum'],
+    ['javaInfoType', 'foam.core.AbstractFObjectPropertyInfo'],
+    ['javaJSONParser', 'foam.lib.json.FObjectParser']
+  ]
+});
+
+foam.CLASS({
+  refines: 'foam.core.AbstractEnum',
+
+  properties: [
```
adamvy: I don't believe these properties are needed here.

src/foam/java/refinements.js
```diff
@@ -483,13 +484,66 @@ foam.CLASS({
...
+        };
+      }
+    }
+  ]
+});
+
+foam.CLASS({
+  refines: 'foam.core.EnumModel',
+
+  properties: [
```
adamvy: Again with these properties :), I don't think you need them here either

src/foam/java/Enum.js
```diff
@@ -1,106 +1,166 @@
...
+  extends: 'foam.java.Class',
 
-  methods: [
+  properties: [
+    ['javaType', 'java.lang.Enum'],
+    ['javaInfoType', 'foam.core.AbstractFObjectPropertyInfo'],
+    ['javaJSONParser', 'foam.lib.json.FObjectParser'],
   ],
+  methods: [
+    function outputJava(o) {
```
adamvy: This is a good start.

Next I would look at if you can refactor this to share some code with outputJava in foam.java.Class.  There is a fair amount of duplication here that can probably be cleaned up.

kirkeaton:
@adamvy Are there any more changes required for this PR?

adamvy:
Sorry for the delay, this looks good to me, thanks!

PR #324, opened 2017-05-09
Fix for broken console functions in examples

m-cheung:


test/browser/FOAMByExampleRunner.js
```diff
@@ -191,4 +191,14 @@ foam.async.repeat(FBE.length, function runExemplar(index) {
     writeOutput();
     return result;
   }
-})();
+})().then(function() {
+  // reverting logging strategy performed during async calls
+  console = {
+    __proto__: oldConsole,
```
mdittmer: This looks right, although we don't need to override `oldConsole`'s implementations with `oldConsole`'s implementations. `console = oldConsole` should suffice.

mdittmer (changes requested):
Got the async/promisey pattern right, but we don't need the extra overrides.

test/browser/FOAMByExampleRunner.js
m-cheung: The change has been made as requested

m-cheung (commented):
The change has been made as requested

mdittmer (approved):
LGTM

PR #328, opened 2017-05-09
NanoServices + Boot + HttpServer

diaswrd:
- Implemented `NanoService` interface
- Created `Boot` class to start all nano services on startup
- Created `HttpServer` that listens on port 80 and forward requests to the desired service. E.g: `GET /logger/foo/bar` will pass `/foo/bar` to `Logger` service.
- Minor changes to Java MapDAO and AbstractDAO


java_src/foam/nanos/boot/NSpec.java
```diff
@@ -0,0 +1,143 @@
+// DO NOT MODIFY BY HAND.
```
kgrgreer: This shouldn't be checked in. Should only be generated as part of the build and then discarded.

kgrgreer:
Don't delete the origintrials files. There's a way now to specify your custom file list for genjava. Ask Tanay how.

diaswrd:
Sorry about that @kgrgreer, it should be good now. 👍

PR #329, opened 2017-05-09
minor tweaks on AATree Node methods

marcogbarcellos:


src/foam/dao/index/AATree.js
```diff
@@ -358,7 +358,7 @@ foam.CLASS({
         return copy;
       }
 
-      if ( r < 0 ) return s.right.lte(key, compare, nullNode);
+      if ( r < 0 ) return s.left.lte(key, compare, nullNode);
```
mdittmer, line 361: This looks like a "how did this ever work?" kind of thing. @marcogbarcellos can we get a test that would have failed with the old code?

kgrgreer:
> This looks like a "how did this ever work?" kind of thing. 

That's exactly what I thought. Scary. The old code definitely looks wrong.

marcogbarcellos:
@mdittmer yes I'll try to work on tests for the whole AATree.

jacksonic:
Goes all the way back to FOAM1: https://github.com/foam-framework/foam/blob/c321d7d5ffc621e84a67fab85cf45155d1af947c/core/index.js#L446

LT case looks to be the correct way, so we may just have never (throroughly) tried an LTE.

PR #334, opened 2017-05-10
Auth

MarcRoopchand:
Split up files so that they can be generated for Java


kgrgreer:
You've added back the class list in genjava which was previously removed.

tools/genjava.js
```diff
@@ -15,123 +15,123 @@
  * limitations under the License.
  */
 
-// enable FOAM java support.
-global.FOAM_FLAGS = { 'java': true, 'debug': true };
+ // enable FOAM java support.
```
kgrgreer: Why did you add a blank space to the beginning of each line?

MarcRoopchand: My editor was being very weird. I just reverted to the last commit. Not sure where that white space came from

PR #341, opened 2017-05-11
mlang Has for Java

jlhughes:
Continuation of pull-request mlang.Has from a previous fork.
Our discussion regarding testing for empty string and array:
_Adam_: I don't think we should compare the output of toString() as that may require building strings we don't need which is slow

_Joel_: Testing for an empty array is not straightforward. Testing for an array can be accomplished with 'value.getClass().isArray()' or 'instanceof Type[]', but then to get the length, it needs to be cast or you have to know the Type. How would you like proceed?

_Adam_: We can use getClass().isArray() && java.lang.reflect.Array.getLength(value)
https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/Array.html#getLength(java.lang.Object)


kgrgreer:
Do we need to know if it's an array or not if we add .has() as a method to Property and then let polymorphism figure it out for us by having Array properties have a different implementation than non-array properties?

jlhughes:
Yes, .has() on Property would simplify things. 
The test for empty array was ensuring the Java **Has** was consistent with the javascript **Has**, so both would benefit.

src/foam/mlang/mlangJava.js
```diff
@@ -350,6 +350,21 @@ foam.CLASS({
 });
 
 
+foam.CLASS({
+  refines: 'foam.mlang.predicate.Has',
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: 'Object value = getArg1().f(obj); '+
```
mdittmer: @kgrgreer we've started using backtick strings for documentation. Should we allow/encourage the same for generated code strings?

mdittmer: nit: Space before `+`es or backtick string.

jlhughes: Can someone point me to an example of this backtick thing, thanks.

mdittmer:
Backtick multiline string example: https://github.com/foam-framework/foam2/blob/3bac563915e2d031b859d8687c8c5ab0742c39e2/src/foam/core/Slot.js#L22

src/foam/mlang/mlangJava.js
mdittmer: To be clear, meant:

```js
'some long ' +
//          ^--- this space
'kind ' +
//     ^--- and this space
'of string'

// OR

`some long
kind
of string`
// This one only works if having the newlines be a part of the string is acceptable.
```

jlhughes:
OK, so we are converting all generated java code to use backticks?

kgrgreer:
You can use regular quoted strings if it all fits on one line, but ES6's new multi-line backquoteed strings: `fblah blah
blah` prevent the need for using the function() {/* blah
blah */} hack. Unless you're targetting IE9, but even then, you can run the code through an ES6 transpiler for the production build.

PR #342, opened 2017-05-11
Modified java_src/gen.sh for external file list generation command

pavneetskapoor:
Changed gen.sh command for file generation in regards to previous change of genjava.js where an external file list can be used to generate classes and such.


kgrgreer:
@jlhughes

Issue #301, opened 2017-04-25
Revert sink.put() argument re-ordering


mdittmer:
When flow controls were removed and a subscription added to `sink.put()`, the payload was moved from the first arg to the second arg. Subsequent discussion of this issue has built consensus that the payload should be moved back to the first arg.

mdittmer:
Fixed in 930fc935fc14d58372e61f51c0d16584630a731d

PR #344, opened 2017-05-11
Sink interface change: put, remove args: (sub, obj) => (obj, sub)

mdittmer:
Towards #301. 25abf392efdc6db3d5b4e303ac771bed19d68748 changed the `Sink.put()` and `Sink.remove()` to accept a subscription object as the first paramter. The idea was to make these methods closer to the `Listener` interface. Subsequent discussions built a consensus that similarity to `Listener` is less important that putting the most-used paramter first (and not breaking existing consumers of `Sink`). This change effectively reverts the "change the Sink interface" parts of 25abf392efdc6db3d5b4e303ac771bed19d68748.


jacksonic:
I agree it is certainly simpler to put the optional argument last.

mdittmer:
FWIW, I'm not sure whether the argument is strictly optional (in the interface contract sense). However, it is rarely used by `Sink` implementations.

jacksonic:
Right, "frequently unused" from the Sink's point of view. Callers must
supply something.

adamvy:
fine with me

PR #350, opened 2017-05-11
Add Aware Interface files and build flags for Java compilation

jlhughes:
Each of EnabledAware, LastModifiedAware, LastModifiedByAware must generate a corresponding Java interface file, via foam.INTERFACE, as these models are being 'implemented' in Java.
nanos.js specifies build flags 'js' and 'java'. Two flags are required as build.js creates a working set from non-flagged files and those matching the passed flags. This requires non-java foam builds to pass flag 'js' to build a foam-bin.js

### DIFF
--- a/src/foam/dao/SinkJava.js
+++ b/src/foam/dao/SinkJava.js
@@ -38,13 +38,13 @@ foam.INTERFACE({
       javaReturns: 'void',
       args: [
         {
+          name: 'obj',
+          javaType: 'foam.core.FObject'
+        },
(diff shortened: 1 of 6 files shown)


kgrgreer:
There is no 'js' flag, so this broke our web apps. I guess what goes around comes around ;-). Fixed now.

jlhughes:
Ah, that's where your flags are; tried to find them.  Never thought to look
in foam.js, as we don't use it.

PR #351, opened 2017-05-12
Test for null 'sub' before detaching.

jlhughes:


adamvy:
It would be good to know where this is happening.  Almost all DAOs should pass a real object as sub.

I think MDAO doesn't because it handles limit/skip on its own, but it probably should be updated to respect the sub as well.

PR #353, opened 2017-05-12
DAOs enabledModAware

kennyqiyenkan:
Added the following:
- EnabledAwareDAO
  - both .js and .java
- LastModifiedDAO
  - both .js and .java


kennyqiyenkan:
I signed it!

src/foam/nanos/auth/LastModifiedAwareDAO.java
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.auth;
+import foam.dao.*;
+import foam.core.*;
+import java.time.LocalDateTime;
+
+public abstract class LastModifiedAwareDAO
```
kgrgreer: Shouldn't be abstract.

src/foam/nanos/auth/LastModifiedAwareDAO.js
```diff
@@ -0,0 +1,22 @@
...
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'LastModifiedAwareDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  methods: [
+    {
+      name: 'put',
+      code: function(value) {
+        value.lastModified = new Date().now();
```
kgrgreer: lastModified is a Date, so remove the .now().

tools/classes.js
```diff
@@ -48,6 +48,7 @@ var classes = [
   'foam.mlang.sink.Count',
   'foam.nanos.boot.NSpec',
   'foam.nanos.auth.EnabledAware',
+  'foam.nanos.auth.EnabledAwareDAO',
```
kgrgreer: Remove from here but put in src/foam/files.js.

kgrgreer:
Now add the Java decorators, which are the primary goal.

src/foam/nanos/auth/LastModifiedAwareDAO.js
```diff
@@ -0,0 +1,22 @@
...
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+// TODO: Test
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'LastModifiedAwareDAO',
+  extends: 'foam.dao.ProxyDAO',
```
adamvy: I think these would be better served as DAODecorators rather than ProxyDAOs.

src/foam/nanos/auth/LastModifiedAwareDAO.java
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.auth;
```
adamvy: Are these classes really nanos specific? Or would they do better in a more generate purpose package?

src/foam/nanos/auth/LastModifiedAwareDAO.js
kgrgreer: DAODecorator makes modifying the data on the way out easier, but these decorators are only adding behaviour on put(), so I don't think it buys anything here.

src/foam/nanos/auth/LastModifiedAwareDAO.java
kgrgreer: LastModifiedAware and EnabledAware aren't but LastModifiedByAware is, and it seemed like they should be in the same directory. I thought about putting them in foam.pattern, but you generally think of patterns as being more complex. Perhaps just putting them in foam.dao would be better, and then add a constant to LastModifiedByAware for the PRINCIPAL_KEY, so that it's not directly tied to the auth system.

kgrgreer:
@kennyqiyenkan Please move this code  to the foam.dao package instead of foam.nanos.auth. Thx

java_src/foam/dao/EnabledAwareDAO.java
```diff
@@ -0,0 +1,26 @@
...
+import foam.dao.*;
+import foam.mlang.MLang.*;
+import foam.mlang.predicate.Predicate;
+import foam.mlang.order.Comparator;
+import foam.dao.Sink;
+import foam.nanos.auth.EnabledAware;
+
+public class EnabledAwareDAO
+  extends ProxyDAO
+  implements EnabledAware
```
kgrgreer: Doesn't need to implement EnabledAware. Only the objects it's filtering need to.

java_src/foam/dao/EnabledAwareDAO.java
```diff
@@ -0,0 +1,26 @@
...
+{
+  public Sink select(Sink s, Integer skip, Integer limit, Comparator order, Predicate predicate) {
+    return super.select(s, skip, limit, order, EQ(this.EnabledAware.ENABLED, true));
```
kgrgreer: this could be a constant so that you don't need to create a new one for each call

java_src/foam/dao/LastModifiedAwareDAO.java
```diff
@@ -0,0 +1,23 @@
...
+import foam.nanos.auth.LastModifiedAware;
+import java.time.LocalDateTime;
+
+public class LastModifiedAwareDAO
+  extends ProxyDAO
+  implements LastModifiedAware
+{
+  public FObject put(FObject value) {
+    ((LastModifiedAware)value).lastModified = LocalDateTime.now();
+    super.put(value);
```
kgrgreer: The delegate could return a different object, so the correct semantics would be:
return super.put(value);

src/foam/dao/LastModifiedAwareDAO.js
```diff
@@ -0,0 +1,21 @@
...
+  package: 'foam.dao',
+  name: 'LastModifiedAwareDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  methods: [
+    {
+      name: 'put',
+      code: function(value) {
+        value.lastModified = new Date();
+        SUPER.put(value);
```
kgrgreer: return SUPER(value);

java_src/foam/dao/EnabledAwareDAO.java
```diff
@@ -14,13 +14,13 @@
 
 public class EnabledAwareDAO
   extends ProxyDAO
-  implements EnabledAware
 {
+  private Predicate PREDICATE_ = EQ(this.EnabledAware.ENABLED, true);
```
kgrgreer: Should be 'final' and 'static' to be a constant. Otherwise you're still creating multiple instants and users could update the value.

PR #355, opened 2017-05-12
Flags 'js' and 'java' are mutually exclusive.

jlhughes:
genjava.js will fail if foam.js flag 'js' is true. The 'js' and 'java' flags control which FooAware model is used: the class or interface.


kgrgreer:
I already (which is to say, 15 minutes after, without noticing your CL) fixed this in:
https://github.com/foam-framework/foam2/commit/1bf12b6085276a50f43ad0a691907edaafb63a51

PR #359, opened 2017-05-16
Added support for Java Array List

TanayParikh:
## Java Array Lists
- Added custom comparison template for Java Array Lists to refinements.js

### Before:
```java
      public int compare(Object o1, Object o2) {
        
          java.util.ArrayList values1 = get_(o1);
          java.util.ArrayList values2 = get_(o2);
                if ( values1.size() > values2.size() ) return 1;
                if ( values1.size() < values2.size() ) return -1;
        
                int result;
                for ( int i = 0 ; i < values1.size() ; i++ ) {
                  result = ((Comparable)values1.get(i)).compareTo(values2.get(i));
                  if ( result != 0 ) return result;
                }
                return 0;
      }
```

### After:
```java
      public int compare(Object o1, Object o2) {
        
          java.util.ArrayList values1 = get_(o1);
          java.util.ArrayList values2 = get_(o2);
        
          if ( values1.size() > values2.size() ) return 1;
          if ( values1.size() < values2.size() ) return -1;
        
          int result;
          for ( int i = 0 ; i < values1.size() ; i++ ) {
            result = ((Comparable)values1.get(i)).compareTo(values2.get(i));
            if ( result != 0 ) return result;
          }
          return 0;
      }
```

### DIFF
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -627,15 +627,16 @@ foam.CLASS({
       template: function() {/*
   <%= this.javaType %> values1 = get_(o1);
   <%= this.javaType %> values2 = get_(o2);
-        if ( values1.length > values2.length ) return 1;
-        if ( values1.length < values2.length ) return -1;
+  
+  if ( values1.length > values2.length ) return 1;
+  if ( values1.length < values2.length ) return -1;
 
-        int result;
-        for ( int i = 0 ; i < values1.length ; i++ ) {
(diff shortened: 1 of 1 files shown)


TanayParikh:
I'm getting the following warning on generation:
```
Property foam.core.FObjectProperty.of "value" hidden by "getter"
```

Can anyone verify issue is present in their environment as well?

kgrgreer:
This isn't an error, it's just letting us know that this could be a
potential problem, but it's what we intend.

PR #365, opened 2017-05-16
Extends interfaces

marcogbarcellos:
I was creating an interface to be extended by another interface and noticed it didn't really extended it, it was just getting its methods. 
So I first change the `refinements.js` to stop ignoring the `model_.extends` and some other properties. 
Also changed the `Context.js` because the `lookout` method was considering only `string` types, but, for example the `extends` property of an interface is an array of string, so now we are checking whether it's an array or a string.

### DIFF
--- a/src/foam/core/Context.js
+++ b/src/foam/core/Context.js
@@ -43,8 +43,22 @@
      * @param opt_suppress Suppress throwing an error.
      **/
     lookup: function(id, opt_suppress) {
-      var ret = typeof id === 'string' && this.__cache__[id];
-
+      // If it's an interface extending other interfaces, it won't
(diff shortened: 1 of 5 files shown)


src/foam/core/Context.js
```diff
@@ -43,8 +43,22 @@
      * @param opt_suppress Suppress throwing an error.
      **/
     lookup: function(id, opt_suppress) {
-      var ret = typeof id === 'string' && this.__cache__[id];
-
+      // If it's an interface extending other interfaces, it won't
```
kgrgreer, line 46: I don't think this should be handled here. Lookup should only support a String id. Supporting arrays of extends values should be handled in the interface code, not here.

Issue #366, opened 2017-05-17
foam.util.compare() is broken


m-cheung:
There are several observed instances where foam.util.compare will not return the expected value
For example:

1. Given `a = undefined` and `b = anything`
`foam.util.compare(a, b) => 1` and `foam.util.compare(b, a) => 1`

2. Given `a = null` and `b = anything`
`foam.util.compare(a, b) => 1` and `foam.util.compare(b, a) => 1`

3. Given `a = true` and `b = FObject` (This one is really weird...)
`foam.util.compare(a, b) => 0` and `foam.util.compare(b, a) => 1`

m-cheung:
Here are some examples of cases that remain unresolved by #368. These really don't have an intuitive answer to them and design choices will likely need to be made. 

For all cases below, the current result is `foam.util.compare(a, b) == foam.util.compare(b, a) == 1`

1. `a = []` and `b = 1`
2. `a = []` and `b = null`
3. `a = ""` and `b = null`
4. `a = function() {}` and `b = null`
5. `a = function() {}` and `b = undefined`
...

The core idea behind the issue stems from the fact that the primitive types usually perform a type check (`typeOf(b) == typeOf(a)`) and returns 1 when the types are not the same. Thus, we get 1 no matter what order is specified when these elements are compared.

m-cheung:
To resolve the remaining issues, precedence has been assigned to each type. 

In the first attempt, precedence was determined by alphabetically (with `foam.Object` and `foam.core.FObject` being placed last). As you can see in this [Travis build](https://travis-ci.org/foam-framework/foam2/builds/233686152), this breaks some of the tests concerning the DAO. The main issue stems from the fact that `undefined` is the first parameter in a call to `foam.util.compare` in `foam.mlang.predicate.Gt`

In the second attempt, precedence was given to `foam.Undefined` and `foam.Null`, with the rest remaining in the same order as above. However, as you can see in this [Travis build](https://travis-ci.org/foam-framework/foam2/jobs/233688892), a test with regards to Date was broken. The failing test is `expect(foam.util.compare(new Date(234324), null)).toEqual(1);`

Alternatively, all of the tests will pass if priority is given just to `foam.Undefined` while the rest remain in the initial order that was described.

@mdittmer has suggested that the second approach makes the most sense and it gives precedence to the "simplest" types. @kgrgreer I would appreciate any input you can give on this.

mdittmer:
@m-cheung fixed this in #368

PR #368, opened 2017-05-17
Adjustments to foam.util.compare() and addition of related tests

m-cheung:
This pull request attempts to resolve the issue described in #366. If you have any suggestions on any changes I have made so far, I would be more than happy to incorporate them

### DIFF
--- a/src/foam/core/FObject.js
+++ b/src/foam/core/FObject.js
@@ -809,8 +809,8 @@ foam.CLASS({
     function compareTo(other) {
       if ( other === this ) return 0;
 
-      // Use primitive type comparator if available
-      if (foam.typeOf(other).id !== "foam.core.FObject") {
+      // Use comparator of other if it is not an FObject
+      if (!foam.core.FObject.isInstance(other)) {
         return -foam.typeOf(other).compare(other, this);
       }
 
--- a/src/foam/core/FObject.js
+++ b/src/foam/core/FObject.js
@@ -808,6 +808,12 @@ foam.CLASS({
 
     function compareTo(other) {
       if ( other === this ) return 0;
+
+      // Use primitive type comparator if available
+      if (foam.typeOf(other).id !== "foam.core.FObject") {
+        return -foam.typeOf(other).compare(other, this);
+      }
+
       if ( ! other        ) return 1;
 
       if ( this.model_ !== other.model_ ) {
--- a/src/foam/core/FObject.js
+++ b/src/foam/core/FObject.js
@@ -811,7 +811,7 @@ foam.CLASS({
 
       // Use comparator of other if it is not an FObject.
       if ( ! foam.core.FObject.isInstance(other) ) {
-        return -foam.typeOf(other).compare(other, this);
+        return foam.util.compare(this, other);
       }
 
       if ( ! other        ) return 1;
--- a/src/foam/core/FObject.js
+++ b/src/foam/core/FObject.js
@@ -810,7 +810,7 @@ foam.CLASS({
       if ( other === this ) return 0;
 
       // Use comparator of other if it is not an FObject
-      if (!foam.core.FObject.isInstance(other)) {
+      if ( ! foam.core.FObject.isInstance(other) ) {
         return -foam.typeOf(other).compare(other, this);
       }
 
--- a/src/foam/core/FObject.js
+++ b/src/foam/core/FObject.js
@@ -809,7 +809,7 @@ foam.CLASS({
     function compareTo(other) {
       if ( other === this ) return 0;
 
-      // Use comparator of other if it is not an FObject
+      // Use comparator of other if it is not an FObject.
       if ( ! foam.core.FObject.isInstance(other) ) {
         return -foam.typeOf(other).compare(other, this);
       }
--- a/src/foam/core/stdlib.js
+++ b/src/foam/core/stdlib.js
@@ -69,7 +69,7 @@ foam.LIB({
     function isInstance(o) { return o === undefined; },
     function clone(o) { return o; },
     function equals(_, b) { return b === undefined; },
-    function compare(_, b) { return b === undefined ? 0 : 1; },
+    function compare(_, b) { return b === undefined ? 0 : -1; },
     function hashCode() { return -2; }
   ]
 });
@@ -82,7 +82,7 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(_, b) { return b === null; },
     function compare(_, b) {
-      return b === null ? 0 : b === undefined ? -1 : 1;
+      return b === null ? 0 : -1;
     },
     function hashCode() { return -3; }
   ]
@@ -95,7 +95,10 @@ foam.LIB({
     function isInstance(o) { return typeof o === 'boolean'; },
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
-    function compare(a, b) { return a ? (b ? 0 : 1) : (b ? -1 : 0); },
+    function compare(a, b) {
+      if ( ! this.isInstance(b) ) return -1;
+      return a ? (b ? 0 : 1) : (b ? -1 : 0);
+    },
     function hashCode(o) { return o ? 1 : -1; }
   ]
 });
@@ -108,6 +111,7 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return b ? a.toString() === b.toString() : false; },
     function compare(a, b) {
+      if ( ! this.isInstance(b) ) return -1;
       return b ? foam.String.compare(a.toString(), b.toString()) :  1;
     },
     function hashCode(o) { return foam.String.hashCode(o.toString()); },
@@ -361,8 +365,9 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
     function compare(a, b) {
-      return ( ! this.isInstance(b) ) ? 1 :
-          a < b ? -1 : a > b ? 1 : 0
+      if ( ! this.isInstance(b) || ( isNaN(a) && ! isNaN(b)) ) return -1;
+      else if ( ! isNaN(a) && isNaN(b) ) return 1;
+      return a < b ? -1 : a > b ? 1 : 0;
     },
     (function() {
       var bufForHash = new ArrayBuffer(8);
@@ -388,7 +393,10 @@ foam.LIB({
     function isInstance(o) { return typeof o === 'string'; },
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
-    function compare(a, b) { return b != null ? a.localeCompare(b) : 1 ; },
+    function compare(a, b) {
+      if ( ! this.isInstance(b) ) return -1;
+      return b != null ? a.localeCompare(b) : 1 ;
+    },
     function hashCode(s) {
       var hash = -4;
 
@@ -532,7 +540,7 @@ foam.LIB({
       return true;
     },
     function compare(a, b) {
-      if ( ! b || ! Array.isArray(b) ) return 1;
+      if ( ! b || ! Array.isArray(b) ) return -1;
       var l = Math.min(a.length, b.length);
       for ( var i = 0 ; i < l ; i++ ) {
         var c = foam.util.compare(a[i], b[i]);
@@ -568,10 +576,10 @@ foam.LIB({
     function getTime(d) { return ! d ? 0 : d.getTime ? d.getTime() : d ; },
     function equals(a, b) { return this.getTime(a) === this.getTime(b); },
     function compare(a, b) {
-      if ( ! this.isInstance(b) && ! foam.Number.isInstance(b) ) return 1;
+      if ( ! this.isInstance(b) && ! foam.Number.isInstance(b) ) return -1;
       a = this.getTime(a);
-      b = this.getTime(b);
-      return a < b ? -1 : a > b ? 1 : 0;
+      b = foam.Number.isInstance(b) ? b : this.getTime(b);
+      return foam.Number.compare(a, b);
     },
     // Hash n & n: Truncate to 32 bits.
     function hashCode(d) { var n = d.getTime(); return n & n; },
@@ -648,6 +656,7 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
     function compare(a, b) {
+      if ( ! this.isInstance(b) ) return -1;
       return foam.Number.compare(a.$UID, b ? b.$UID : -1);
     },
     function hashCode(o) {
@@ -698,9 +707,9 @@ foam.typeOf = (function() {
   };
 })();
 
-foam.typeOrder = [ foam.Undefined, foam.Null, foam.Array,
-    foam.Boolean, foam.Date, foam.Function, foam.String,
-    foam.Number, foam.core.FObject, foam.Object ];
+foam.typeOrder = [ foam.Undefined, foam.Null, foam.Boolean,
+    foam.Date, foam.Number, foam.String, foam.Array,
+    foam.Function, foam.Object, foam.core.FObject ];
 
 foam.LIB({
   name: 'foam',
--- a/src/foam/core/stdlib.js
+++ b/src/foam/core/stdlib.js
@@ -69,7 +69,7 @@ foam.LIB({
     function isInstance(o) { return o === undefined; },
     function clone(o) { return o; },
     function equals(_, b) { return b === undefined; },
-    function compare(_, b) { return b === undefined ? 0 : -1; },
+    function compare(_, b) { return b === undefined ? 0 : 1; },
     function hashCode() { return -2; }
   ]
 });
@@ -82,7 +82,7 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(_, b) { return b === null; },
     function compare(_, b) {
-      return b === null ? 0 : -1;
+      return b === null ? 0 : 1;
     },
     function hashCode() { return -3; }
   ]
@@ -96,7 +96,7 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
     function compare(a, b) {
-      if ( ! foam.Boolean.isInstance(b) ) return -1;
+      if ( ! foam.Boolean.isInstance(b) ) return 1;
       return a ? (b ? 0 : 1) : (b ? -1 : 0);
     },
     function hashCode(o) { return o ? 1 : -1; }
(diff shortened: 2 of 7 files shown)


src/foam/core/FObject.js
```diff
@@ -808,6 +808,12 @@ foam.CLASS({
 
     function compareTo(other) {
       if ( other === this ) return 0;
+
+      // Use primitive type comparator if available
+      if (foam.typeOf(other).id !== "foam.core.FObject") {
```
m-cheung: There might be an easier way to check for primitive types... but this was the shortest solution that occurred to me

src/foam/core/stdlib.js
```diff
@@ -361,7 +361,7 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
     function compare(a, b) {
-      return ( b === null || b === undefined ) ? 1 :
+      return ( b === null || b === undefined || foam.typeOf(b).id === 'foam.core.FObject') ? 1 :
```
m-cheung: Due to the inversion of control introduced above, when foam.util.compare is called with an object and number, this method will be called. Without this check, we would get 0 since (a !< b && a !> b).

src/foam/core/FObject.js
kgrgreer: ! foam.core.FObject.isInstance(other)

src/foam/core/stdlib.js
kgrgreer: foam.core.FObject.isInstance(other)

test/src/core/FObject.js
```diff
@@ -214,3 +214,49 @@ describe('copyFrom', function() {
...
+    var someObj = SomeClass.create();
+
+    // foam.Boolean does a loose comparison. Object is considered true
+    expect(foam.util.compare(true, someObj)).toBe(0);
+    expect(foam.util.compare(someObj, true)).toBe(0);
+
+    // Unlike other comparators, foam.Boolean return -1 when object is in 2nd
+    // param and 1 when object is in 1st param.
+    expect(foam.util.compare(false, someObj)).toBe(-1);
+    expect(foam.util.compare(someObj, false)).toBe(1);
```
m-cheung: foam.Boolean behaves slightly differently than the other comparators. Observe that the the comparison function of the other types return 1 when object is the 2nd parameter; the boolean comparator returns a -1 instead.

src/foam/core/FObject.js
```diff
@@ -809,8 +809,8 @@ foam.CLASS({
     function compareTo(other) {
       if ( other === this ) return 0;
 
-      // Use primitive type comparator if available
-      if (foam.typeOf(other).id !== "foam.core.FObject") {
+      // Use comparator of other if it is not an FObject
+      if (!foam.core.FObject.isInstance(other)) {
```
kgrgreer: See the FOAM style guide for proper formatting.

src/foam/core/stdlib.js
```diff
@@ -568,6 +568,7 @@ foam.LIB({
     function getTime(d) { return ! d ? 0 : d.getTime ? d.getTime() : d ; },
     function equals(a, b) { return this.getTime(a) === this.getTime(b); },
     function compare(a, b) {
+      if (!this.isInstance(b) && !foam.Number.isInstance(b)) return 1;
```
kgrgreer: formatting.

src/foam/core/stdlib.js
```diff
@@ -361,8 +361,8 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
     function compare(a, b) {
-      return ( b === null || b === undefined ) ? 1 :
-          a < b ? -1 : a > b ? 1 : 0;
+      return ( b === null || b === undefined || foam.Object.isInstance(b) ||
```
mdittmer: Does this file `'use strict';`? If not, `undefined` could have been reassigned to anything. It may be sufficient to `'use strict';` this function.

src/foam/core/stdlib.js
```diff
@@ -361,8 +361,8 @@ foam.LIB({
...
+          foam.core.FObject.isInstance(b) ) ? 1 : a < b ? -1 : a > b ? 1 : 0;
```
mdittmer: This seems really complex. Shouldn't we have something like `Number.isNumber(b) ? a > b : 1;`? Anything that comes before this in the precedence list will never hit this code path, and anything that comes after aught to return `1`, right?

src/foam/core/stdlib.js
```diff
@@ -746,7 +747,26 @@ foam.LIB({
     methods: [
       function clone(o)      { return typeOf(o).clone(o); },
       function equals(a, b)  { return typeOf(a).equals(a, b); },
-      function compare(a, b) { return typeOf(a).compare(a, b); },
+      function compare(a, b) {
+        // To ensure that symmetry is present when comparing,
+        // we will always use the comparator of higher precedence.
+        var types = [ foam.Undefined, foam.Null, foam.Array,
```
mdittmer: Let's not redefine this every time `compare` runs. Can we store it somewhere like `foam.typeOrder`? I don't love that name. @kgrgreer WDYT, re, storing this somewhere else and/or what to name it?

src/foam/core/FObject.js
```diff
@@ -808,6 +808,12 @@ foam.CLASS({
 
     function compareTo(other) {
       if ( other === this ) return 0;
+
+      // Use comparator of other if it is not an FObject.
+      if ( ! foam.core.FObject.isInstance(other) ) {
```
mdittmer, line 813: This can be simpler now, right? With the addition of precedence + `FObject` being last, we can use `if (...other not FObj...) return foam.util.compare(this, other);` which will take the right path for us.

mdittmer (commented):
Some initial thoughts on `Number`'s complexity and updating some "early ideas" in this space to fit neatly with the implementation we later chose.

src/foam/core/stdlib.js
m-cheung: Somewhere in `mlang`, there is a direct call to `foam.Number.compare`, which is why I ended up with all those checks. In the next commit, I have changed the use of `foam.Number.compare` to `foam.util.compare` and replaced it with ` ( ! this.isInstance(b) ) ? 1 : a < b ? -1 : a > b ? 1 : 0;`

The latter part of the ternary is still needed, since we need to return 1 and -1 as required for numbers, not just a boolean value.

src/foam/core/stdlib.js
m-cheung: I don't think this is needed anymore due to the change below

src/foam/core/FObject.js
m-cheung: I think it can be simplified to `if (...other not FObj...) return -foam.util.compare(other, this);`. 
We cannot write `return foam.util.compare(this, other);` or we get an may run into an infinite loop (this is due to the fact FObject is an Object and it causes complications on this matter).

I have kept this as is for now as I'm not sure if the snippet I mentioned above actually avoids all potential infinite loops as well

src/foam/core/stdlib.js
```diff
@@ -361,8 +361,8 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
     function compare(a, b) {
-      return ( b === null || b === undefined ) ? 1 :
-          a < b ? -1 : a > b ? 1 : 0;
+      return ( ! this.isInstance(b) ) ? 1 :
+          a < b ? -1 : a > b ? 1 : 0
```
mdittmer: nit: missing semicolon

src/foam/core/FObject.js
mdittmer: My understanding is that `FObject` is the last thing we would put on the left. Hence, if `other` is an `Object`, then `foam.util.compare(this, other)` will eventually lead to `-foam.Object.compare(other, this)`.

src/foam/mlang/mlang.js
```diff
@@ -564,7 +564,7 @@ foam.CLASS({
           sortedArgs[costKey] = arg;
           costs.push(costKey);
         }
-        costs = costs.sort(foam.Number.compare);
+        costs = costs.sort(foam.util.compare);
```
mdittmer: In hindsight, I think we don't need to change this. using `foam.<type>.compare(a, b)` directly is safe iff `a` and `b` are known to be of type `<type>`. That is the case here. It's also the case [here](https://github.com/foam-framework/foam2/blob/2868ed4916cc7a3477708e7538f7a4e27421b888/src/foam/core/types.js#L101).

test/src/core/FObject.js
```diff
@@ -214,3 +214,62 @@ describe('copyFrom', function() {
     expect(obj.d).toBe(true);
   });
 });
+
+describe('FObject compare', function() {
```
mdittmer, line 218: Let's tighten this up. We should be able to get away with something like:

```js
var things = [ undefined, null, false, true, 0, 1, '', 'foo', {}, new Date(), foam.core.FObject.create(), foam.core.FObject.create() ];
for ( var i = 0; i < things.length; i++ ) {
  for ( var j = i; j < things.length; j++ ) {
    expect(foam.util.compare(things[i], things[j]).toBe(-foam.util.compare(things[j], things[i]));
  }
}
```

src/foam/core/stdlib.js
```diff
@@ -697,6 +698,9 @@ foam.typeOf = (function() {
   };
 })();
 
+foam.typeOrder = [ foam.Undefined, foam.Null, foam.Array,
```
mdittmer: I think this should go from "simplest type" to "most complex type". To me, this would be:

`undefined, null, boolean, number, string, array, function, date, object, fobject`

I'm unsure about the ordering of `string`/`array` and `function`/`date`, but I think the others are "genuinely simpler/more complex".

If using this order introduces problems we should discuss the nature of the problems.

src/foam/core/stdlib.js
```diff
@@ -746,7 +759,24 @@ foam.LIB({
     methods: [
       function clone(o)      { return typeOf(o).clone(o); },
       function equals(a, b)  { return typeOf(a).equals(a, b); },
-      function compare(a, b) { return typeOf(a).compare(a, b); },
+      function compare(a, b) {
+        // To ensure that symmetry is present when comparing,
+        // we will always use the comparator of higher precedence.
+        var types = foam.typeOrder;
+        var typeID = function(obj) {
```
kgrgreer: This is going to be slow. Comparators are some of the most heavily called code in the system. Maybe add an 'ordinal' property to each of the type libs so that we can instead do:
var ta = foam.typeOf(a), tb = foam.typeOf(b);
return foam.Number.compare(ta.ordinal, tb.ordinal) || ta.compare(a, b);

src/foam/core/stdlib.js
```diff
@@ -568,9 +576,10 @@ foam.LIB({
     function getTime(d) { return ! d ? 0 : d.getTime ? d.getTime() : d ; },
     function equals(a, b) { return this.getTime(a) === this.getTime(b); },
     function compare(a, b) {
+      if ( ! this.isInstance(b) && ! foam.Number.isInstance(b) ) return -1;
```
kgrgreer: Aren't we assuming that type-specific comparators are of the same type? If so, are these kinds of checks needed?

src/foam/mlang/mlang.js
kgrgreer: Don't use foam.util.compare if not needed since it will be much slower than specific comparators.

src/foam/core/stdlib.js
mdittmer: We re not. We assume that `b` will not be of a type "prior to" (in this case) `foam.Date` in `foam.typeOrder` (or equivalent `ordinal` markings). That means that, if `b` isn't a `Date` or something that can be interpreted as a `Date`, we wish to return "`a` comes before `b`": `-1`. The recommendation I make for calling these directly from outside of `foam.util.compare` only when `a` and `b` are (in this case) `Date`s is so that the behaviour is independent of these assumptions that come from type ordering.

src/foam/core/stdlib.js
mdittmer: Do you mean `(ta.ordinal <= tb.ordinal ? ta : tb).compare(a, b)`?

src/foam/core/stdlib.js
```diff
@@ -361,8 +365,9 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return a === b; },
     function compare(a, b) {
-      return ( b === null || b === undefined ) ? 1 :
-          a < b ? -1 : a > b ? 1 : 0;
+      if ( ! this.isInstance(b) || ( isNaN(a) && ! isNaN(b)) ) return -1;
+      else if ( ! isNaN(a) && isNaN(b) ) return 1;
+      return a < b ? -1 : a > b ? 1 : 0;
```
mdittmer: `Math.sign(a - b)`?

src/foam/core/stdlib.js
kgrgreer: No, that's not what I mean. I first compare the types and if they're different, I don't need to compare the actual values. If they are the same type the first compare returns 0 so I then do the type-specific compare().

src/foam/core/stdlib.js
m-cheung: `Math.sign(a - b)` works for every case except for `a = NaN` and ` b = NaN`

test/src/core/FObject.js
m-cheung: This has been implemented in `test/src/core/stdlib.js`

test/src/core/stdlib.js
```diff
@@ -166,10 +166,10 @@ describe('foam.Boolean', function() {
...
     expect(foam.Boolean.compare(false, false)).toBe(0);
 
-    expect(foam.Boolean.compare(true, 0)).toBe(1);
+    expect(foam.Boolean.compare(true, 0)).toBe(-1);
     expect(foam.Boolean.compare(false, 66)).toBe(-1);
-    expect(foam.Boolean.compare(true, 9)).toBe(0);
-    expect(foam.Boolean.compare(false, 0)).toBe(0);
+    expect(foam.Boolean.compare(true, 9)).toBe(-1);
+    expect(foam.Boolean.compare(false, 0)).toBe(-1);
   });
```
m-cheung: Notice that due to the recent changes, `foam.Boolean.compare` is no longer performing a loose comparison

test/src/core/Boot.js
```diff
@@ -194,30 +194,33 @@ describe('Property default comparators', function() {
...
   it('accepts double falsey', function() {
-    expect(c(false, false)).toEqual(0);
-    expect(c(NaN, NaN)).toEqual(0);
-    expect(c(0, 0)).toEqual(0);
-    expect(c('', '')).toEqual(0);
+    expect(c(false, false)).toBe(0);
+    expect(c(NaN, NaN)).toBe(0);
+    expect(c(0, 0)).toBe(0);
+    expect(c('', '')).toBe(0);
   });
```
m-cheung, line 204: Changed from `toEqual(0)` to `toBe(0)` since `-0` will fail the validation

test/src/core/Boot.js
```diff
@@ -194,30 +194,33 @@ describe('Property default comparators', function() {
...
   it('falls back on $UID.compareTo', function() {
-    var o1 = {};
-    var o2 = {};
-    expect(c(o1, o2)).toEqual(-1);
+    var o1 = Object.create(null);
+    o1.$UID = 26;
+    var o2 = Object.create(null);
+    o2.$UID = 27;
+    expect(c(o1, o2)).toBe(-1);
+    expect(c(o2, o1)).toBe(1);
```
m-cheung, line 223: There were a few instances on my local machine where `o1` seemed to have a greater UID than `o2`. I have explicitly defined them to avoid this from causing a false test failure

test/src/core/stdlib.js
```diff
@@ -167,9 +167,9 @@ describe('foam.Boolean', function() {
     expect(foam.Boolean.compare(false, false)).toBe(0);
 
     expect(foam.Boolean.compare(true, 0)).toBe(1);
-    expect(foam.Boolean.compare(false, 66)).toBe(-1);
-    expect(foam.Boolean.compare(true, 9)).toBe(0);
-    expect(foam.Boolean.compare(false, 0)).toBe(0);
+    expect(foam.Boolean.compare(false, 66)).toBe(1);
+    expect(foam.Boolean.compare(true, 9)).toBe(1);
+    expect(foam.Boolean.compare(false, 0)).toBe(1);
```
m-cheung, line 172: `foam.Boolean.compare` is no longer performing loose comparisons

src/foam/core/stdlib.js
```diff
@@ -746,7 +774,17 @@ foam.LIB({
     methods: [
       function clone(o)      { return typeOf(o).clone(o); },
       function equals(a, b)  { return typeOf(a).equals(a, b); },
-      function compare(a, b) { return typeOf(a).compare(a, b); },
+      function compare(a, b) {
+        // To ensure that symmetry is present when comparing,
+        // we will always use the comparator of higher precedence.
+        var aOrdinal = typeOf(a).ordinal;
+        var bOrdinal = typeOf(b).ordinal;
+        foam.assert(aOrdinal !== undefined && bOrdinal !== undefined,
```
kgrgreer: Remove this assert(). Compare() performance is too important for this.

src/foam/core/stdlib.js
```diff
@@ -746,7 +774,17 @@ foam.LIB({
...
+            'comparator cannot operate on unknown types');
+
+        return (aOrdinal > bOrdinal) ? typeOf(a).compare(a, b) :
```
kgrgreer: typeOf() is expensive, which is why in the code I proposed I stored type type in a variable rather than the ordinal, so that I wouldn't have to look it up again here.

kgrgreer: You are missing the case where aOrdinal == bOrdinal, which my original code handled.

src/foam/core/stdlib.js
mdittmer: @kgrgreer Don't asserts get turned into no-ops in production?

src/foam/core/stdlib.js
```diff
@@ -777,13 +775,14 @@ foam.LIB({
...
-
-        return (aOrdinal > bOrdinal) ? typeOf(a).compare(a, b) :
-            -typeOf(b).compare(b, a);
+        var aType = typeOf(a);
+        var bType = typeOf(b);
+        return (aType.ordinal >= bType.ordinal) ? aType.compare(a, b) :
+            this.comparatorNegate(bType.compare(b, a));
+      },
+      function comparatorNegate(num) {
+        if (num === 0) return 0;
```
mdittmer: Nit: please add a comment explaining handling of -0

mdittmer (approved):
One nit. Otherwise LGTM!

src/foam/core/stdlib.js
```diff
@@ -781,6 +781,9 @@ foam.LIB({
             this.comparatorNegate(bType.compare(b, a));
       },
       function comparatorNegate(num) {
+        // When negating comparisons, a -0 may be returned and
```
mdittmer: I'm not sure this says exactly what we mean. I think the main points are:

- `num` may be `-0`
- `-0 === 0 ---> true`
- We wish to avoid comparators to return `-0`

I feel the middle point is what's missing from the current explanation.

PR #370, opened 2017-05-17
Added Test and Script to nano

waseem-chamaa:


waseem-chamaa:
I signed it!

java_src/gen.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-node ../tools/genjava.js ../tools/classes.js .
+node ../tools/genjava.js ../tools/classes.js ../java_src
```
kgrgreer, line 3: Don't make this change. I don't want to break other people's builds.

PR #376, opened 2017-05-18
HTTP auth agents and requests: Google 2-legged OAuth

mdittmer:
This introduces abstractions for:

- HTTP authentication agents
- HTTP credentials
- Authenticated HTTP requests

Concrete implementations are:

- An agent for Google 2-legged (server-to-server) OAuth
- Token bearer credentials
- Token bearer authenticated requests

The intended use case is as follows:

```js
var ctlr = foam.net.auth.EasyAuthController.create({
  authClient: some.auth.Agent.create({/*...*/})
})
some.AuthenticatedRequestMaker.create({/*...*/}, ctlr);
```

UPDATE: Controller dropped. New usage:

```js
some.AuthorizedRequestMaker.create({
  /*... class makes requests via 'foam.net.HTTPRequest'...*/
}, some.concrete.AuthClient.create({
  /*...*/
}));
```


mdittmer:
In hindsight, until we support more than one credential type, we could probably get away with requiring `AuthAgent` implementations to `exports: [ 'as authAgent' ]` and register their `HTTPRequest` decorator in their sub-context. This would simplify the use case down to `some.AuthenticatedRequestMaker.create({...}, authAgent)`. Would that simpler scheme be preferable?

mdittmer:
... I'm not a big fan of "export as <some name> + register your companion class as 'foam.net.HTTPRequest'" being a part of an interface-like contract. How are other implementers supposed to know that's what's expected of them?

PR #380, opened 2017-05-22
Implement dao.select() optimizations for DatastoreDAO.select(COUNT())

mdittmer:
(1) Perform key-only query (full data objects not required);
(2) Use results.length to update count.value, rather than constructing objects and issuing put()s.


mdittmer:
TBR @shepheb

PR #381, opened 2017-05-23
Fixed argument merging behaviour for abstract methods

TanayParikh:
- Previously argCount logic could cause access to index out of range in `this.args`

## Previous Error
```
foam2/src/foam/java/refinements.js:242
        result.args[i] = this.args[i].clone().copyFrom(child.args[i]);
                                     ^

TypeError: Cannot read property 'clone' of undefined
    at Object.createChildMethod_ 
```

## Error Replication
Got the above error when I tried overriding the default object equality method (`equals`).

### DIFF
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -239,7 +239,7 @@ foam.CLASS({
       // Special merging behaviour for args.
       var argCount = Math.max(this.args.length, child.args.length)
       for ( var i = 0 ; i < argCount ; i++ ) {
-        result.args[i] = this.args[i].clone().copyFrom(child.args[i]);
+        result.args[i] = (this.args.length <= i) ? child.args[i] : this.args[i].clone().copyFrom(child.args[i]);
       }
 
       return result;
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -237,9 +237,8 @@ foam.CLASS({
       }
 
       // Special merging behaviour for args.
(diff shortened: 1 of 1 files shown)


src/foam/java/refinements.js
```diff
@@ -239,7 +239,7 @@ foam.CLASS({
       // Special merging behaviour for args.
       var argCount = Math.max(this.args.length, child.args.length)
       for ( var i = 0 ; i < argCount ; i++ ) {
-        result.args[i] = this.args[i].clone().copyFrom(child.args[i]);
+        result.args[i] = (this.args.length <= i) ? child.args[i] : this.args[i].clone().copyFrom(child.args[i]);
       }
```
kgrgreer: Or, how about this which avoids the max() and the check on each iteration:
var i = 0;
for ( ; i < this.args.length   ; i++ ) result.args[i] = this.args[i].clone().copyFrom(child.args[i]);
for ( ; i < child.args.length ; i++ ) result.args[i] = child.args[i];

TanayParikh: Sure, the second loop can even be removed I believe (the entire `child` object is already being cloned into `result` on line 229). 

```js
      // Special merging behaviour for args.
      for (var i = child.args.length ; i < this.args.length ; i++ ) {
        result.args[i] = this.args[i].clone().copyFrom(child.args[i]);
      }
```

Edit: My bad, could lead to data loss. Two loops it is.

Issue #325, opened 2017-05-09
Implement attribute parsing in HTMLLexer


mdittmer:
We need to address [this TODO](https://github.com/foam-framework/foam2/blob/master/src/foam/parsers/HTMLLexer.js#L232).

PR #383, opened 2017-05-23
Implementing attribute parsing in HTMLLexer

m-cheung:
[Please do not merge. Refactoring in progress]
Purposed solution to #325

### DIFF
--- a/src/foam/parsers/HTMLLexer.js
+++ b/src/foam/parsers/HTMLLexer.js
@@ -219,7 +219,8 @@ foam.CLASS({
             return Embed.create({
               nodeName: v[0][2],
               content: [ v[1] ],
-              attributes: v[0][4]
+              attributes: v[0][4],
+              type: OPEN_CLOSE
             });
           },
 
@@ -234,7 +235,8 @@ foam.CLASS({
             var str = v[1];
             var ret = Embed.create({
               nodeName: nodeName,
-              attributes: attributes
+              attributes: attributes,
+              type: OPEN_CLOSE
             });
 
             // Attempt to parse maybeEmbeds. Returns "html" parse or string.
--- a/src/foam/parsers/HTMLLexer.js
+++ b/src/foam/parsers/HTMLLexer.js
@@ -29,63 +29,6 @@ foam.ENUM({
 
 foam.CLASS({
   package: 'foam.parsers',
-  name: 'Attribute',
-
-  properties: [
-    {
-      class: 'String',
-      name: 'element'
-    },
-    {
-      class: 'String',
-      name: 'value'
-    }
-  ]
-});
-
-
-foam.CLASS({
-  package: 'foam.parsers',
-  name: 'Tag',
-
-  properties: [
-    {
-      class: 'Enum',
-      of: 'foam.parsers.TagType',
-      name: 'type',
-      factory: function() { return foam.parser.TagType.OPEN; }
-    },
-    {
-      class: 'String',
-      name: 'nodeName',
-      value: 'div'
-    },
-    {
-      class: 'Array',
-      name: 'attributes',
-    }
-  ]
-});
-
-
-foam.CLASS({
-  package: 'foam.parsers',
-  name: 'Embed',
-  extends: 'foam.parsers.Tag',
-
-  properties: [
-    {
-      name: 'type',
-      factory: function() { return foam.parser.TagType.OPEN_CLOSE; }
-    },
-    'content'
-  ]
-
-});
-
-
-foam.CLASS({
-  package: 'foam.parsers',
   name: 'HTMLLexer',
 
   documentation: `Parse an HTML string into a flat sequence of tags and
@@ -95,9 +38,9 @@ foam.CLASS({
     'foam.parse.ImperativeGrammar',
     'foam.parse.Parsers',
     'foam.parse.StringPS',
-    'foam.parsers.Attribute',
-    'foam.parsers.Embed',
-    'foam.parsers.Tag',
+    'foam.parsers.html.Attribute',
+    'foam.parsers.html.Embed',
+    'foam.parsers.html.Tag',
     'foam.parsers.TagType'
   ],
 
@@ -315,7 +258,7 @@ foam.CLASS({
           comment: function(v) { return null; },
           attributes: function(v) { return v; },
           attribute: function(v) {
-            return Attribute.create({ element: v[0], value: v[1] });
+            return Attribute.create({ name: v[0], value: v[1] });
           },
           value: function(v) { return v; }
         };
--- a/src/foam/parsers/HTMLLexer.js
+++ b/src/foam/parsers/HTMLLexer.js
@@ -134,8 +134,8 @@ foam.CLASS({
           value: str(alt(
               plus(notChars('\'" \t\r\n<>')),
               // TODO(markdittmer): Support proper escaping inside strings.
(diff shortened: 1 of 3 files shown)


src/foam/parsers/HTMLLexer.js
```diff
@@ -29,6 +29,23 @@ foam.ENUM({
 
 foam.CLASS({
   package: 'foam.parsers',
+  name: 'Attribute',
+
+  properties: [
+    {
+      class: 'String',
+      name: 'element'
```
mdittmer: Element sounds confusing wrt html elements. Maybe "name" or "key"?

src/foam/parsers/HTMLLexer.js
```diff
@@ -42,8 +59,11 @@ foam.CLASS({
       class: 'String',
       name: 'nodeName',
       value: 'div'
+    },
+    {
+      class: 'Array',
```
mdittmer: We  want class: 'FObjectArray', of: '(fully qualified attribute class name)'

src/foam/parsers/HTMLLexer.js
```diff
@@ -29,6 +29,23 @@ foam.ENUM({
 
 foam.CLASS({
   package: 'foam.parsers',
+  name: 'Attribute',
```
mdittmer: This and Tag should probably either be called HTML(Classname) or move to foam.parsers.html

mdittmer (commented):
Some initial thoughts. Still need to check semantic actions and think about tests when I'm not reviewing from a phone.

src/foam/parsers/HTMLLexer.js
m-cheung: I have placed `Attributes`, `Embed` and `Tag` under `foam.parsers.html` in the next change

mdittmer:
We should add tests that address [issues identified here](https://stackoverflow.com/questions/4015345/how-to-properly-escape-quotes-inside-html-attributes). In particular:

- Single or double quotes
- Converting `&name;`-style escapes

src/foam/parsers/HTMLLexer.js
```diff
@@ -283,9 +256,11 @@ foam.CLASS({
           doctypePart: function(v) { return null; },
           cdata: function(v) { return null; },
           comment: function(v) { return null; },
-          attributes: function(v) { return null; },
-          attribute: function(v) { return null; },
-          value: function(v) { return null; }
+          attributes: function(v) { return v; },
```
mdittmer: No need for this line; `v => v` is the default action.

src/foam/parsers/HTMLLexer.js
```diff
@@ -283,9 +256,11 @@ foam.CLASS({
...
+          attribute: function(v) {
+            return Attribute.create({ name: v[0], value: v[1] });
```
mdittmer: Will `v[1]` be `undefined` or `null`? We should probably set `Attribute.value` to `null`.

src/foam/parsers/HTMLLexer.js
```diff
@@ -283,9 +256,11 @@ foam.CLASS({
...
+          },
+          value: function(v) { return v; }
```
mdittmer: Ditto: Don't need default action.

test/src/parsers/HTMLLexer.js
```diff
@@ -139,6 +139,38 @@ describe('HTMLLexer', function() {
...
+        '<div id=1 name="TestDiv2" visible=true />',
+        '<label name="TestLabel" id=2 visible />',
+        '<pre class="idl" visible=false id=3><div>Some more content</div></pre>'
+    ].forEach(function(str, index) {
+      var value = testParse('htmlPart', str);
+      if ( ! value ) return;
+
+      expect(Array.isArray(value.attributes)).toBe(true);
+      expect(value.attributes).toBeDefined();
+      expect(value.attributes.length > 2).toBe(true);
```
mdittmer: Let's use `toBeGreaterThan` here.

test/src/parsers/HTMLLexer.js
```diff
@@ -139,6 +139,38 @@ describe('HTMLLexer', function() {
...
+      var value = testParse('htmlPart', str);
+      if ( ! value ) return;
+
+      expect(Array.isArray(value.attributes)).toBe(true);
+      expect(value.attributes.length).toBe(1);
+      expect(value.attributes[0]).toBeDefined();
+    });
+  });
+
+  it('should parse tags multiple attributes', function() {
```
mdittmer: I think we should have more tests, each that tests one thing (or a particular combination of things). Maybe something like:

- Multiple attributes
- Attribute without value
- Attribute without quotes
- Attribute in single quotes
- Attribute with `&name;` escape of some kind
- Attribute containing whitespace (maybe include ` \t\r\n`)

mdittmer (changes requested):
Finished reviewing all code. Mostly just need a few more tests.

m-cheung:
All requested changes have been made and should be good to go now

src/foam/parsers/HTMLLexer.js
```diff
@@ -168,10 +132,9 @@ foam.CLASS({
                    sym('value')))),
 
           value: str(alt(
-              plus(notChars('\'" \t\r\n<>')),
-              // TODO(markdittmer): Support proper escaping inside strings.
-              seq1(1, '"', repeat(notChars('"', anyChar())), '"'),
-              seq1(1, "'", repeat(notChars("'", anyChar())), "'"))),
+              plus(alt(sym('escape'), notChars('\'" \t\r\n<>'))),
```
m-cheung: I was required to alter this case as well, since Chrome parses `<option value=&amp;Potato />`

src/foam/parsers/HTMLLexer.js
```diff
@@ -219,7 +219,8 @@ foam.CLASS({
             return Embed.create({
               nodeName: v[0][2],
               content: [ v[1] ],
-              attributes: v[0][4]
+              attributes: v[0][4],
+              type: OPEN_CLOSE
             });
```
m-cheung: I noticed that `Embed` and `maybeEmbed` did not previously have a type. However, I noticed that the way the grammar is defined, these nodes act like self closing items (i.e., there is only one node for each set of opening and closing tags for these types. Thus, they essentially act as `OPEN_CLOSE`).

Although this change can be discarded, it does make my work a lot easier when performing tag matching.

test/src/parsers/HTMLLexer.js
```diff
@@ -96,29 +121,27 @@ describe('HTMLLexer', function() {
...
+      var value = testParse('html', str);
       if ( ! value ) return;
 
-      expect(foam.parsers.Embed.isInstance(value)).toBe(true);
-      expect(Array.isArray(value.content)).toBe(true);
-      expect(value.content.length).toBe(1);
-      expect(value.content[0]).toBe('sc < ri ? p &gt; t : en');
+      expect(Array.isArray(value)).toBe(true);
+      expect(value.length).toBe(3);
+      expect(value[1]).toBe('sc < ri ? p > t : en');
```
m-cheung, line 144: I'm not sure how the lexer would know whether or not to convert the escaped character. If we place this into Chrome, the escaped character does not get converted (similar to what the test used to be).

I need confirmation on whether or not this test should be changed

mdittmer:
I believe this PR is being made obsolete by a simpler "true lexer" that @m-cheung is working on.

PR #385, opened 2017-05-24
Changing Value from int to long on mlang Count and minor tweaks

marcogbarcellos:
- Changing Value from int to long on mlang Count
- Decreasing foam authors headers size
- Changing `for` loop to be on the FOAM coding standard

### DIFF
--- a/java_src/foam/dao/index/AltIndex.java
+++ b/java_src/foam/dao/index/AltIndex.java
@@ -19,9 +19,8 @@ public class AltIndex implements Index {
   protected ArrayList<Index> delegates_ = new ArrayList();
 
   public AltIndex(Index... indices) {
-    for (Index indice : indices) {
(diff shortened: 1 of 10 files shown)


java_src/foam/dao/index/AltIndex.java
```diff
@@ -30,9 +19,8 @@
   protected ArrayList<Index> delegates_ = new ArrayList();
 
   public AltIndex(Index... indices) {
-    for (Index indice : indices) {
-      addIndex(indice);
-    }
+    for ( int i = 0 ; i < indices.length ; i++ )
```
kgrgreer, line 22: FYI: I was okay with the kind of loop, just not the spacing.

marcogbarcellos: got it, thanks

PR #389, opened 2017-05-24
Expression and validation enhancements

mcarcaso:
Some features:

1. Expressions can now look something like this:
function(foo$bar) { /* code */ }
Which, under the hood, will create an expression with this.foo$.dot('bar') being the slot.

2. You can now pass an array to FObject's slot method of the form: [args, code] where args is an array of strings and code is the function to call with the value of these args.

3. Give validateObj a default value when "required" is true (this makes use of #2).

### DIFF
--- a/src/foam/core/FObject.js
+++ b/src/foam/core/FObject.js
@@ -454,6 +454,7 @@ foam.CLASS({
       if ( this.hasOwnProperty(name) ) {
         var oldValue = this[name];
         this.instance_[name] = undefined;
+        this.clearPrivate_(name);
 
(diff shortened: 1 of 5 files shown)


adamvy:
@kgrgreer 

FYI: This commit PR adds support for subscribing to nested values in expressions.

Example:

{
  name: 'foo',
   factory: function() { return this.Foo.create({ a: 12 }); }
},
{
  name: 'someProp',
  expression: function(foo$a) {
    return foo$a *2;
  }
}

I looked over the code with mike, I'm pretty happy with it so we're merging as is but feel free to add your own comments.

PR #391, opened 2017-05-24
Enable nano services lazy loading

diaswrd:
We were having issues with the initialization order of the services using the Boot class.

This PR will make all services lazy by default, and start them when they get accessed for the first time. To avoid this behaviour, set `lazy` flag as `false` on your services.


diaswrd:
Sorry about all the merge commit logs, I will clean up them for the next PR.

java_src/foam/nanos/boot/NSpecFactory.java
```diff
@@ -0,0 +1,37 @@
...
+    spec_ = spec;
+  }
+
+  public Object create(X x) {
+    NanoService ns = null;
+
+    try {
+      ns = spec_.createService();
+
+      ((ContextAwareSupport) ns).setX(x);
```
kgrgreer, line 25: We should make NanoService extend ContextAware (waiting for bug fix by @marcogbarcellos), then this cast won't be needed.

kgrgreer:
Nice.

PR #394, opened 2017-05-26
Added refinement for Max sink

kirkeaton:


src/foam/mlang/mlang.js
```diff
@@ -1568,17 +1568,14 @@ foam.CLASS({
...
-  ],
+  implements: ['foam.core.Serializable'],
 
   documentation: 'A Sink which remembers the maximum value put().',
 
   properties: [
     {
-      name: 'value',
-      value: 0
+      class: 'FObjectProperty',
```
kgrgreer: class: 'Object' 

(not FObject)

kirkeaton: What's the difference between FObjectProperty & ObjectProperty?

kgrgreer:
FObject is a FOAM-Object, ie. an Object which has been modelled. All
modelled objects in Java extend FObject. Object is a regular Java Object.

PR #395, opened 2017-05-26
java interfaces now can extend interfaces

marcogbarcellos:
I was creating an interface to be extended by another interface and noticed it didn't really extended it, it was just getting its methods.

There's a thing this fix does not solve and we must have it as well. here is a practical example: 

- Inside the mlang.js we have the 'Expr' Interface that now should extend an interface called `F`. If we use `extends: 'foam.mlang.F'` it generates the content properly for both F and Expr. 
- But for the the `Constant`(also defined on the mlang) that extends the `AbstractExpr` that implements `Expr` and in this case, the f method(that comes from the `F` interface) is "losing" its type and returning void instead of returning the same defined on the `F` interface as it should.
- If instead of using `extends: 'foam.mlang.F'` inside the `Expr` we use `implements: ['foam.mlang.F']` everything works fine(including the `Constant` thing). that's what this branch is doing for now.


src/foam/core/Interface.js
```diff
@@ -81,7 +81,24 @@ foam.LIB({
 
   methods: [
     function INTERFACE(m) {
+
       m.class = m.class || 'foam.core.InterfaceModel';
+      // if m.implements not defined, add it as an array, otherwise add its content
+      // in an array
+      if ( !m.implements ) {
```
kgrgreer, line 88: if ( ! m.implements ) {

PR #396, opened 2017-05-26
add java implementation of journaled dao 

drish:
- the usage of the java side usage implementation is a bit different from the javascript implementation, I plan to tweak the javascript implementation accordingly.

In javascript:

```js
 var journaledDAO = foam.dao.JournaledDAO.create({
  delegate: delegate,
  journal: foam.dao.NodeFileJournal({
    fd: require('fs').openSync('journal-file', 'rw');
  })
});

journaledDAO.put . . .
```

In java I simplified the call a little bit to, example:

```java

DAO delegate = new DAO(); // example
JournaledDAO jdAO = new JournaledDAO(delegate, "journal-file"); // wrapped all the JavaScript calls into the JournaledDAO constructor, making it simpler to use.
jDAO.put . . .
```

Also.

- use `p` and `r` as operations instead of `put` and `remove` in the journal file for more file space efficiency, will need to create another PR for updating the Javascript land implementation to use `p` and `r` too.

from: 
```txt
put(foam.parse.json({class: "foam.dao.Guitar", brand: "fender"}));
remove(foam.parse.json({class: "foam.dao.Guitar", brand: "fender"}));
```

to:
```txt
p(foam.parse.json({class: "foam.dao.Guitar", brand: "fender"}));
r(foam.parse.json({class: "foam.dao.Guitar", brand: "fender"}));
```


java_src/foam/dao/FileJournal.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param obj
+     * @param sub
+     */
+    @Override
+    public void put(FObject obj, Detachable sub) {
+        try {
+            Outputter outputter = new Outputter();
+            this.bw.write("p(foam.json.parse(" + outputter.stringify(obj) + "))");
+            this.bw.newLine();
+            this.bw.flush();
```
kgrgreer, line 44: This will kill performance, but I'll show you how to fix it later.

java_src/foam/dao/FileJournal.java
```diff
@@ -0,0 +1,102 @@
...
+     * Persists data into the File System.
+     *
+     * @param obj
+     * @param sub
+     */
+    @Override
+    public void put(FObject obj, Detachable sub) {
+        try {
+            Outputter outputter = new Outputter();
+            this.bw.write("p(foam.json.parse(" + outputter.stringify(obj) + "))");
```
kgrgreer, line 42: Why do we need '(foam.json.parse(' and '))'?

java_src/foam/lib/json/JournalParser.java
```diff
@@ -0,0 +1,49 @@
...
+ * put(foam.json.parse({"class":"foam.nanos.boot.NSpec","name":"Gibs1" }))
+ * remove(foam.json.parse({"class":"foam.nanos.boot.NSpec","name":"Gibs" }))
+ */
+public class JournalParser extends ContextAwareSupport {
+
+    protected JSONParser jsonParser;
+
+    public JournalParser() {
+        // do not fail if context is null
+        X ctx = (getX() == null) ? new ProxyX() : getX();
```
kgrgreer, line 21: Why ProxyX() instead of EmptyX.instance()?

java_src/foam/lib/json/JournalParser.java
```diff
@@ -0,0 +1,49 @@
...
+        return this.jsonParser.parseString(object);
+    }
+
+    /**
+     * Extract the operation (put/remove) from the journal file line.
+     *
+     * @param journalLine
+     * @return operation
+     */
+    public String parseOperation(String journalLine) {
```
kgrgreer, line 43: This will always just be the first character.

java_src/foam/lib/json/JournalParser.java
```diff
@@ -0,0 +1,49 @@
...
+        this.jsonParser = new JSONParser();
+        jsonParser.setX(ctx);
+    }
+
+
+    // TODO(drish): do not generate a second copy of the journalLine
+    public FObject parseObject(String journalLine) {
+
+        // get the actual object
+        int objectIndex = journalLine.indexOf("{");
```
kgrgreer, line 31: This should always just be the 3rd char. ie. after skipping 'p ' or 'r '.

java_src/foam/dao/FileJournal.java
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public void removeAll() {
+        this.file.delete();
+    }
+
+    @Override
+    public void remove(FObject obj, Detachable sub) {
+        try {
+            Outputter outputter = new Outputter();
+            this.bw.write("r(foam.json.parse(" + outputter.stringify(obj) + "))");
```
kgrgreer, line 61: Wasteful to write whole object. Just need to write the 'id'.

drish: I don't see how the entire object would be reconstructed if we only save the `id` in the journal file, for example when the journal file already exist and it's being replayed back to the delegated DAO. Am I missing something ?

java_src/foam/dao/FileJournal.java
drish: The javascript JournaledDAO uses `eval();` to load it. That's why it's stored that way.

kgrgreer: We don't need to be file compatible with the JS version, so there's no need for Java to do that.

java_src/foam/dao/FileJournal.java
kgrgreer: We can just store the id, then do dao.remove(dao.find(id)), which will save a lot of disk space because it's redundant to restore the whole object just so that you can remove it. The id for some objects could be orders of magnitude smaller than the full object.

drish:
gonna open a clean pr, this branch was really old and it was going to take me a while to fix some conflicts.

PR #399, opened 2017-05-29
Added SequenceNumberDAO implementation in Java

kirkeaton:
Implementation assumes that the property being used as the sequence is an integer. (although may be better to use long instead).

### DIFF
--- a/java_src/foam/dao/SequenceNumberDAO.java
+++ b/java_src/foam/dao/SequenceNumberDAO.java
@@ -1,0 +1,62 @@
+package foam.dao;
+
+import foam.core.ClassInfo;
+import foam.core.FObject;
+import foam.core.PropertyInfo;
+import foam.mlang.sink.Max;
+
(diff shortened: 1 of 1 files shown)


java_src/foam/dao/SequenceNumberDAO.java
```diff
@@ -0,0 +1,62 @@
...
+    return this;
+  }
+  
+  public SequenceNumberDAO setValue(int value) {
+    this.value = value;
+    return this;
+  }
+  
+  private PropertyInfo getProperty_() {
+    if (property_ != null) {
```
kgrgreer: Simpler to do:

  if ( property_ == null ) {
    property_ = (PropertyInfo) this.of_.getAxiomByName(property);
  }
     
  return property_;

kirkeaton: Thanks. I've made the changes.

PR #408, opened 2017-05-31
Implemented fclone to FObject, and made MapDAO return a clone

nick-prat:


java_src/foam/core/AbstractFObject.java
```diff
@@ -7,6 +7,23 @@
...
+    FObject ret;
+    try {
+      ret = (FObject) getClassInfo().getObjClass().newInstance();
+    } catch (InstantiationException | IllegalAccessException e) {
+      e.printStackTrace();
+      return null;
+    }
+
+    List<PropertyInfo> props = getClassInfo().getAxiomsByClass(PropertyInfo.class);
+    for(PropertyInfo pi : props) {
```
kgrgreer: See FOAM style guideline: https://github.com/foam-framework/foam/wiki/FOAM2-Style-Guidelines

This line should be:

for ( PropertyInfo pi : props ) {

java_src/foam/dao/MapDAO.java
```diff
@@ -51,9 +51,9 @@ public FObject remove(FObject obj) {
 
   public FObject find(Object o) {
     if( getOf().getObjClass().isInstance(o) ) {
-      return getData().get(getPrimaryKey().get(o));
+      return getData().get(getPrimaryKey().get(o)).fclone();
```
kgrgreer: This will throw a NPE if the object isn't found.

kgrgreer: Rewrite as:

return maybeClone(getOf().getObjClass().isInstance(o) ?
    getData().get(getPrimaryKey().get(o)):
    getData.get(o));

where maybeClone() is a method that you add which clones if not null.
isInstance() could also be added to ClassInfo so you could just do getOf().isInstance()

java_src/foam/core/EmptyClassInfo.java
```diff
@@ -23,6 +23,10 @@ public ClassInfo addProperty(PropertyInfo p) {
     return this;
   }
 
+  public boolean isInstance(Object o) {
+    return o instanceof EmptyClassInfo;
```
kgrgreer: Just return 'false', since you can't create objects of this type.

java_src/foam/core/AbstractFObject.java
```diff
@@ -7,6 +7,28 @@
...
 {
+
+  public static FObject maybeClone(FObject fo) {
+    return (fo == null ? null : ((AbstractFObject)fo).fclone());
+  }
+
+  public FObject fclone() {
+    FObject ret;
+    try {
+      ret = (FObject) getClassInfo().getObjClass().newInstance();
```
kgrgreer: Add newInstance as a method to ClassInfo the same way you did for isInstance().

java_src/foam/core/AbstractFObject.java
```diff
@@ -7,6 +7,22 @@
...
   implements FObject
 {
+
+  public static FObject maybeClone(FObject fo) {
+    return (fo == null ? null : ((AbstractFObject)fo).fclone());
+  }
+
+  public FObject fclone() {
+    FObject ret = (FObject) getClassInfo().newInstance();
+    if(ret != null) {
```
kgrgreer, line 17: Should be:
if ( ret != null ) {

See the FOAM2 style guide.

PR #409, opened 2017-05-31
Fixed issue with newLines and Outputter 

waseem-chamaa:


kgrgreer:
All code is written by waseem-chamaa who has signed the CLA.

PR #410, opened 2017-05-31
Auth java code

MarcRoopchand:
AuthService
- Permission check completed 
- Also Added Caching


java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,102 @@
...
+    Group group = (Group) user.getGroup();
+    if (group == null) {
+      return false;
+    }
+
+    /**
+     * Check if data is already cached, if it is return this data
+     * If not, save it to the maps
+     * */
+    if (permissionMap.containsKey(permission.getName())) {
```
kgrgreer: See the https://github.com/foam-framework/foam/wiki/FOAM2-Style-Guidelines 

This line should be:
if ( permissionMap.containsKey(permission.getName()) ) {

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,102 @@
...
+      ConcurrentHashMap<String, Boolean> userMap = permissionMap.get(permission.getName());
```
kgrgreer: If you just declare the type to be Map<String, Boolean> then if/when you change the type in the future you wouldn't have the change the type here. Code to the interface not the implementation.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,102 @@
...
+
+      if (userMap.containsKey(user.getId())) {
+        return userMap.get(user.getId());
+      }
+
+      Boolean permissionCheck = group.implies(permission.getName());
```
kgrgreer: Should be 'boolean', not 'Boolean'.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,102 @@
...
+      userMap.put(user.getId(), permissionCheck);
+      permissionMap.put(permission.getName(), userMap);
+      return permissionCheck;
+    }
+    else {
```
kgrgreer: No need for an 'else' if you've already returned. 
if ( ) {

  return;
}  else {
  foo();
}

Can just be written:

if ( ) {

  return;
}

foo();

Which reduces nesting.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,102 @@
...
+      ConcurrentHashMap<String, Boolean> userMap = new ConcurrentHashMap<>();
```
kgrgreer: If you're not using any specific parts of the ConcurrentHashMap class, just declare the type as Map.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,102 @@
...
+
+      return permissionCheck;
+    }
+  }
+
+  /**
+   * On logout, walk through entire map and remove all instances of the user
+   * */
+  @Override
+  public X logout(X x) {
```
kgrgreer: I think the return type should be 'void'.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,102 @@
...
+    if (x == null) {
+      return null;
+    }
+
+    User user = (User) x.get("user");
+    if (user == null) {
+      return null;
+    }
+
+    for (String key: permissionMap.keySet()) {
```
kgrgreer: This seems very inefficient. If you had a map of maps then you could just remove one entry without iterating.

src/foam/nanos/auth/Group.js
```diff
@@ -38,5 +44,25 @@ foam.CLASS({
...
+      args: [
+        {
+          name: 'name',
+          javaType: 'String'
+        }
+      ],
+      javaCode:
+        'if (getPermissions() == null) {return false;} \n' +
+        'for (int i = 0; i < permissions_.length; i++) { \n' +
+        '  if (permissions_[i].getId().equals(name)) { \n' +
```
kgrgreer: Don't use equality, create a java.security.Permission and call its implies() method. This adds wildcard support.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
MarcRoopchand: @kgrgreer What if your user belongs to multiple permissions? 
Say your cached data looked something like this 
{
  permission1: {
    user1: true
  }
  permission2: {
    user1: true
  }
}

Now a user logs out, and I need to remove this user from all the permissions.

java_src/foam/nanos/auth/LRULinkedHashMap.java
```diff
@@ -0,0 +1,20 @@
+package foam.nanos.auth;
```
kgrgreer: Maybe put in foam.nanos.util, since it will be useful for others also.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,83 @@
...
+   *  }
+   *  ...
+   *}
+   */
+  protected Map<String, Map<String, Boolean>> permissionMap;
+
+  @Override
+  public void start() {
+    super.start();
+    permissionMap = new LRULinkedHashMap<>(1000);
```
kgrgreer: This could just be set it permissionMap's declaration.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,83 @@
...
+
+    Group group = (Group) user.getGroup();
+    if ( group == null ) return false;
+
+    /**
+     * Check if data is already cached, if it is return this data
+     * If not, save it to the maps
+     * */
+    if ( permissionMap.containsKey(permission.getName()) ) {
+      Map<String, Boolean> userMap = permissionMap.get(permission.getName());
```
kgrgreer: The code below could be restructured to avoid duplication.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,83 @@
...
+   * On logout, walk through entire map and remove all instances of the user
+   * */
+  @Override
+  public void logout(X x) {
+    if ( x == null ) return;
+
+    User user = (User) x.get("user");
+    if ( user == null ) return;
+
+    for ( String key: permissionMap.keySet() ) {
```
kgrgreer: Won't this be a lot more efficient if you make the userMap be the top-level collection instead?

java_src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -195,7 +171,7 @@ public Boolean validateUser(User user) throws IllegalStateException {
    * Just return a null user for now. Not sure how to handle the cleanup
    * of the current context
    */
-  public X logout(X x) {
-    return this.getX().put("user", null);
+  public void logout(X x) {
+    this.getX().put("user", null);
```
kgrgreer: Contexts are immutable, so this line doesn't do anything. Which is good, cause it isn't needed.

src/foam/nanos/auth/Group.js
```diff
@@ -38,5 +44,25 @@ foam.CLASS({
...
+      args: [
+        {
+          name: 'permission',
+          javaType: 'java.security.Permission'
+        }
+      ],
+      javaCode:
+        'if ( getPermissions() == null ) return false; \n' +
+        'for ( int i = 0; i < permissions_.length; i++ ) { \n' +
+        '  if ( permission.implies(new javax.security.auth.AuthPermission(permissions_[i].getId())) ) { \n' +
```
kgrgreer: I think you have the arguments of implies the wrong way round.

java_src/foam/nanos/auth/CachedUserAndGroupAuthService.java
MarcRoopchand: I can do that but permission map contains multiple userMaps. 
Don't I still need to find which userMap I should be accessing?

src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,74 @@
...
+   *    ...
+   *  }
+   *  ...
+   *}
+   */
+  protected Map<String, Map<String, Boolean>> userMap = new LRULinkedHashMap<>(1000000);
+  protected Map<String, Boolean> permissionMap;
+
+  @Override
+  public void start() {
```
kgrgreer: Why is this needed?

src/foam/nanos/auth/CachedUserAndGroupAuthService.java
```diff
@@ -0,0 +1,74 @@
...
+    if ( user == null ) return false;
+
+    Group group = (Group) user.getGroup();
+    if ( group == null ) return false;
+
+    if ( userMap.containsKey(user.getId()) ) {
+      permissionMap = userMap.get(user.getId());
+    }
+    else {
+      permissionMap = new LRULinkedHashMap<>(1000);
```
kgrgreer: I think 1000 needs to be larger, or even just make this a regular HashMap. A user can easily be checked against thousands of permissions if you have one for each menu and one for each property and mode. The users will get cleaned up anyway, so this is redundant.

src/foam/nanos/auth/Group.js
```diff
@@ -38,5 +44,25 @@ foam.CLASS({
...
+    {
+      name: 'implies',
+      javaReturns: 'Boolean',
+      args: [
+        {
+          name: 'permission',
+          javaType: 'java.security.Permission'
+        }
+      ],
+      javaCode:
```
kgrgreer, line 58: Use multi-line backquote (`) to avoid having to concatenate lines together.

src/foam/nanos/auth/CachedUserAndGroupAuthService.java
MarcRoopchand: I assumed that we will need to start this service from Boot.

How do you envision this caching fitting into the system ?

src/foam/nanos/auth/UserAndGroupAuthServiceTest.java
```diff
@@ -1,72 +1,164 @@
...
+  public void addTestUsers() {
+    System.out.println("Registering " + numUsers + " Users");
+    long startTime  = System.nanoTime();
+    ListSink sink   = (ListSink) groupDAO_.select(new ListSink());
+
+    /**
+     * For each user, randomly select a group from the groups created
+     * and assign the user to this group
+     * */
+    for ( int i = 0; i < numUsers; i++ ) {
```
kgrgreer: See FOAM style-guide for spacing:
for ( int i = 0 ; i < numUsers ; i++ ) {

src/foam/util/LRULinkedHashMap.java
```diff
@@ -0,0 +1,20 @@
+package foam.nanos.util;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Created by marcroopchand on 2017-05-24.
+ */
+public class LRULinkedHashMap<K, V> extends LinkedHashMap<K, V> {
+  private final int MAX_SIZE;
```
kgrgreer: Rename MAX_SIZE to maxSize_. CONSTANT naming convention implies it's a constant, but it's just final.

src/foam/nanos/auth/Group.js
MarcRoopchand: We should add this to the style guide as well. Seems throughout the project all javaCode has concatenation

PR #414, opened 2017-06-05
NanoLogger Added

Akenne:
added changes from Kevin's comments

### DIFF
--- a/java_src/foam/nanos/logger/NanoLogger.java
+++ b/java_src/foam/nanos/logger/NanoLogger.java
@@ -1,0 +1,91 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.logger;
+
+import foam.core.*;
+import foam.nanos.NanoService;
+import java.util.logging.*;
+import java.sql.Timestamp;
+import java.text.SimpleDateFormat;
+import java.io.IOException;
+
+public class NanoLogger extends ContextAwareSupport implements NanoService {
+  Logger logger;
+  StringBuffer sb;
+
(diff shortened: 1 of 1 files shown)


java_src/foam/nanos/logger/NanoLogger.java
```diff
@@ -0,0 +1,85 @@
...
+    logger.warning(msg);
+  }
+
+  public void error(String msg) {
+    logger.severe(msg);
+  }
+
+  // can't normally do .debug() with custom formatter: append prefix to indicate debug
+  public void debug(String msg) {
+    logger.warning("~DEBUG~" + msg);
```
kgrgreer: Isn't it cheaper/easier to just encode that it's a debug message in the severity?
Like:
  logger.log(Level.FINE, msg)

java_src/foam/nanos/logger/NanoLogger.java
```diff
@@ -0,0 +1,85 @@
...
+
+import foam.core.*;
+import foam.nanos.NanoService;
+import java.util.logging.*;
+import java.sql.Timestamp;
+import java.text.SimpleDateFormat;
+import java.io.IOException;
+
+public class NanoLogger extends ContextAwareSupport implements NanoService {
+  private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyy.MM.dd.HH.mm.ss");
```
kgrgreer: SimpleDateFormat is not thread-safe. So, if the logger is reentrant, this will also have to change. We should make a ThreadSafeDateFormat using ThreadLocal SimpleDateFormat objects.

java_src/foam/nanos/logger/NanoLogger.java
Akenne: I couldn't find any way to implement that using the custom formatter. the custom formatter only allows a single variable (msg) the be passed. The 3 methods I used (info,error,warning) are the only ones that it implements as well.

Akenne:
i fixed both of the above problems @kgrgreer

PR #415, opened 2017-06-06
HTMLLexer behaves more like a lexer

m-cheung:
`HTMLLexer` should now produce a stream of tokens instead of trying to act like a parser. This should resolve many of the issues that we were experiencing previously (such as unmatched code tags).

Note: Items produced by the lexer are no longer escaped. Since the lexer is not context-aware, escaping is no longer performed. It is up to the user to perform the escaping as needed.

Also solves #325


src/foam/parsers/HTMLLexer.js
```diff
@@ -105,30 +68,21 @@ foam.CLASS({
...
-              sym('maybeEmbed'),
-
               sym('closeTag'),
-              sym('openTag')),
+              sym('openTag'),
+              sym('text')),
 
           openTag: seq(
               '<',
-              sym('whitespace'),
```
m-cheung, line 119: The HTML Standard does not allow for whitespaces to exist between the opening angled bracket and the identifier

src/foam/parsers/HTMLLexer.js
```diff
@@ -105,30 +68,21 @@ foam.CLASS({
...
               sym('tagName'),
               sym('whitespace'),
               sym('attributes'),
               sym('whitespace'),
               optional('/'),
-              sym('whitespace'),
```
m-cheung, line 125: No whitespace is allowed between `/` and the closing angled bracket either

mdittmer (commented):
A couple things to rework:

- It seems confusing that HTMLLexer + an enum (but not other supporting classes) are in `HTMLLexer.js`. Let's put everything in `package/path/Location`; move files to `src/foam/parsers/html` and do class/enum-per-file with `Name.js`
- I think `openTag_` and `closeTag_` (and the variables that refer to them) are now obsolete. Let's delete them.

src/files.js
```diff
@@ -246,7 +246,10 @@ FOAM_FILES([
   { name: "foam/classloader/WebModelFileDAO", flags: ['web', 'debug'] },
   { name: "foam/classloader/NodeModelFileDAO", flags: ['node', 'debug'] },
   { name: "foam/classloader/NodeJsModelExecutor", flags: ['node', 'debug'] },
-  { name: "foam/parsers/html" },
-  { name: "foam/parsers/HTMLLexer" },
+  { name: "foam/parsers/html/Attribute" },
```
mdittmer: I think proper deps order should demands Attribute and TagType before Tag; Tag before HTMLLexer. Maybe put lib first, too?

mdittmer (approved):
Just file ordering in `src/files.js`; otherwise LGTM

PR #416, opened 2017-06-06
Performance Monitors Fix

nick-prat:
Made the stored DAO a service so it can be properly accessed again.

### DIFF
--- a/java_src/foam/nanos/pm/PMDAO.java
+++ b/java_src/foam/nanos/pm/PMDAO.java
@@ -1,5 +1,6 @@
 package foam.nanos.pm;
 
+import foam.dao.MapDAO;
 import foam.dao.ProxyDAO;
 import foam.nanos.NanoService;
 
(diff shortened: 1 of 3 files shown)


java_src/foam/nanos/pm/PMDAO.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Created by nick on 06/06/17.
+ */
+public class PMDAO
+  extends ProxyDAO
+  implements NanoService
+{
+
+  public static final String ServiceName = "pmDAO";
```
kgrgreer: Constants should use constant naming convention. ie. SERVICE_NAME;

java_src/foam/nanos/pm/PMDAO.java
```diff
@@ -0,0 +1,18 @@
+package foam.nanos.pm;
+
+import foam.dao.ProxyDAO;
+import foam.nanos.NanoService;
+
+/**
+ * Created by nick on 06/06/17.
+ */
+public class PMDAO
+  extends ProxyDAO
```
kgrgreer, line 11: Where do you set the delegate to this class?

nick-prat: I changed it from DAOPMLogger to PMDAO in the last commit

PR #418, opened 2017-06-07
Implemented Cron & CronScheduler classes

kirkeaton:
### Cron
Implemented Cron class is loosely based on syntax of a crontab file. It has properties for minute, hour, day of month, month, and day of week, where -1 represents a wildcard value. It has a property for the scheduled time which is calculated based on the other properties via a factory method. It also contains a function to calculate the next scheduled time. It attempts to calculate the next time until the year 3000 (which is the exit condition in a similar function in the Quartz java library). This function throws an exception if the next scheduled time cannot be found. 

### CronScheduler
Added a CronScheduler class which performs the following:
1. Loads Cron jobs from a DAO
2. Selects the Cron jobs where the scheduled time is <= the current time
3. Runs the cron job and updates the next scheduled time
4. Sleeps until the minimum scheduled time.

### DIFF
--- a/java_src/foam/nanos/cron/CronScheduler.java
+++ b/java_src/foam/nanos/cron/CronScheduler.java
@@ -60,6 +60,8 @@ public class CronScheduler
     cron.setLastRun(new Date());
     ps.flush();
     cron.setOutput(baos.toString());
+    // set to next scheduled time
+    cron.setScheduledTime(cron.getNextScheduledTime());
   }
 
   private final Thread cronJobs = new Thread() {
@@ -94,8 +96,8 @@ public class CronScheduler
     Cron cron = new Cron()
         .setHour(23)
         .setMinute(59)
-        .setMonth(6)
-        .setDayOfWeek(6);
+        .setDayOfMonth(31)
+        .setMonth(1);
     System.out.println(cron.getNextScheduledTime());
   }
 }
--- a/java_src/foam/nanos/cron/CronScheduler.java
+++ b/java_src/foam/nanos/cron/CronScheduler.java
@@ -8,6 +8,9 @@ package foam.nanos.cron;
 
 import foam.core.ContextAwareSupport;
 import foam.dao.MapDAO;
+import foam.nanos.NanoService;
+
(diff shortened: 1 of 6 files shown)


java_src/foam/nanos/cron/CronScheduler.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  /**
+   * Runs a Cron job script
+   *
+   * @param cron the cron job to run
+   * @return the updated cron job
+   */
+  private void runCronJob(Cron cron) {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
kgrgreer: The first part of this method could be added to the Script class/model as runScript() so that it could be reused.

src/foam/nanos/cron/Cron.js
```diff
@@ -8,8 +8,108 @@ foam.CLASS({
...
+      name: 'dayOfWeek',
+      value: -1,
+      documentation: `Day of week to execute script.
+          Ranges from 0 - 6, where 0 is Sunday. -1 for wildcard`
+    },
+    {
+      class: 'DateTime',
+      name: 'scheduledTime',
+      documentation: `Scheduled time to run Cron script.`,
+      javaFactory: function() {
```
kgrgreer: Instead of javaFatory: function() { /*  code goes here */ }, you can use
javaFactory: `
  code goes here
`

src/foam/nanos/cron/Cron.js
```diff
@@ -8,8 +8,108 @@ foam.CLASS({
...
+/*Calendar now = Calendar.getInstance();
```
kgrgreer: Why is this different than nextScheduledTime?

kirkeaton: It's not I guess.

java_src/foam/nanos/cron/CronScheduler.java
kirkeaton: Which lines are you referring to?

PR #419, opened 2017-06-07
foam.net.RetryHTTPRequest: Retry requests that reject-on-send()

mdittmer:


src/foam/net/RetryHTTPRequest.js
```diff
@@ -0,0 +1,69 @@
...
+  listeners: [
+    function onError(error) {
+      this.currentTry_++;
+      this.warn('RetryHTTPRequest: Try #' + this.currentTry_ +
+                ' failed on ' + error);
+      if ( this.currentTry_ < this.numTries ) {
+        return this.send();
+      } else {
+        this.error('RetryHTTPRequest: Max tries reached');
+        throw new Error('RetryHTTPRequest: Max tries reached');
```
jacksonic, line 65: Should the underlying error also be included?

mdittmer: Append it to the message? AFAICT JS errors don't have a great model for this.

mdittmer: Also, I'm not accumulating the errors on each attempt, so the last error may prove to be a red herring...

mdittmer: @jacksonic LMK if feels strongly we should fix this in a follow-up PR. I'm going to land this with some other changes that I want to merge cleanly into beta-1.

jacksonic: True.

jacksonic (approved):

PR #420, opened 2017-06-07
Changed exception handling with newInstance()

nick-prat:
newInstance() throws instantiation exceptions instead of handling internally

### DIFF
--- a/java_src/foam/core/AbstractFObject.java
+++ b/java_src/foam/core/AbstractFObject.java
@@ -21,7 +21,6 @@ public abstract class AbstractFObject
       }
       return ret;
     } catch (IllegalAccessException | InstantiationException e) {
-      e.printStackTrace();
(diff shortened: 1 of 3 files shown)


java_src/foam/core/AbstractFObject.java
```diff
@@ -9,18 +9,21 @@
...
-    if(ret != null) {
+    try {
+      FObject ret = (FObject) getClassInfo().newInstance();
       List<PropertyInfo> props = getClassInfo().getAxiomsByClass(PropertyInfo.class);
       for( PropertyInfo pi : props ) {
         pi.set(ret, pi.get(this));
       }
+      return ret;
+    } catch (IllegalAccessException | InstantiationException e) {
+      e.printStackTrace();
```
kgrgreer: remove

PR #423, opened 2017-06-07
Adding unescapeString function to HTML lib

m-cheung:
HTMLLexer no longer automatically escape text. This added function allows the user to unescape strings they wished to unescape.

### DIFF
--- a/src/foam/parsers/html/lib.js
+++ b/src/foam/parsers/html/lib.js
@@ -515,6 +515,19 @@
       },
       function isSelfClosing(nodeName) {
         return selfClosingNodeNames[nodeName];
+      },
+      function unescapeString(str) {
+        if ( ! foam.String.isInstance(str) ) return '';
+        var escapeKeys = Object.keys(escapes).map(function(key) {
+          return `&${key};`
+        });
+
+        var exp = RegExp(`(?=(${escapeKeys.join('|')}))\\1`, 'g');
+        return str.replace(exp, function(m) {
+          // m is in the form of &id; We drop first and last character.
+          var id = m.slice(1, -1);
+          return escapes[id];
+        });
       }
     ]
   });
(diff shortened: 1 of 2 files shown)


src/foam/parsers/html/lib.js
```diff
@@ -515,6 +515,19 @@
...
       function isSelfClosing(nodeName) {
         return selfClosingNodeNames[nodeName];
+      },
+      function unescapeString(str) {
+        if ( ! foam.String.isInstance(str) ) return '';
+        var escapeKeys = Object.keys(escapes).map(function(key) {
+          return `&${key};`;
+        });
+
+        var exp = RegExp(`(?=(${escapeKeys.join('|')}))\\1`, 'g');
```
mdittmer: Is this RegExp stateful (i.e., can it not be reused)? If it can be reused, then we should store the RegExp, rather than compute it every time.

test/src/parsers/HTMLlib.js
```diff
@@ -0,0 +1,35 @@
+describe('HTML Lib', function() {
+  var lib;
+
+  beforeAll(function() {
```
mdittmer, line 4: This is extraneous. We should be able to just have `beforeAll(function() { lib = foam.parsers.html' });`.

mdittmer (approved):
LGTM

src/foam/parsers/html/lib.js
```diff
@@ -506,6 +506,12 @@
     'yen': '¥',
   };
 
+  var escapeKeys = Object.keys(escapes).map(function(key) {
```
mdittmer: One last thing: Drop extra line between `escapeKeys` and `escapeSequencesRegExp`, and add a comment above the two:

```js
// FUTURE: Lazily instantiate RegExp to save memory.
```

PR #430, opened 2017-06-08
Added HistoryDAO decorator

kirkeaton:
Added a HistoryDAO decorator which logs all of the changes that have been made to a model with the timestamp of when it was changed and the User who changed the model.

### DIFF
--- a/java_src/foam/dao/history/HistoryDAO.java
+++ b/java_src/foam/dao/history/HistoryDAO.java
@@ -69,7 +69,7 @@ public class HistoryDAO
     FObject current = this.find(obj);
 
     // add new history record
-    String objectId = (String) ((PropertyInfo) obj.getClassInfo().getAxiomByName("id")).f(obj);
+    Object objectId = ((PropertyInfo) obj.getClassInfo().getAxiomByName("id")).f(obj);
     HistoryRecord historyRecord = new HistoryRecord();
     historyRecord.setObjectId(objectId);
     historyRecord.setUser(formatUserName(user));
@@ -83,7 +83,7 @@ public class HistoryDAO
   @Override
   public FObject remove(FObject obj) {
     // TODO: use context-oriented context when available.
-    String objectId = (String) ((PropertyInfo) obj.getClassInfo().getAxiomByName("id")).f(obj);
+    Object objectId = ((PropertyInfo) obj.getClassInfo().getAxiomByName("id")).f(obj);
     SequenceNumberDAO historyDAO = (SequenceNumberDAO) getX().get("historyDAO");
     historyDAO.removeAll(null, null, null, EQ(HistoryRecord.OBJECT_ID, objectId));
     return super.remove(obj);
--- a/java_src/foam/dao/history/HistoryDAO.java
+++ b/java_src/foam/dao/history/HistoryDAO.java
@@ -68,6 +68,7 @@ public class HistoryDAO
 
     // add new history record
     HistoryRecord historyRecord = new HistoryRecord();
+    historyRecord.setObjectId(obj.getClassInfo().getId());
     historyRecord.setUser(formatUserName(user));
     historyRecord.setTimestamp(new Date());
     historyRecord.setUpdates(getUpdatedProperties(current, obj));
--- a/java_src/foam/dao/history/HistoryDAO.java
+++ b/java_src/foam/dao/history/HistoryDAO.java
@@ -1,0 +1,66 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.dao.history;
+
+import foam.core.FObject;
+import foam.core.PropertyInfo;
+import foam.dao.MapDAO;
+import foam.dao.ProxyDAO;
+import foam.nanos.auth.User;
+
+import java.util.Date;
+import java.util.Iterator;
+import java.util.Map;
+
+public class HistoryDAO
+  extends ProxyDAO
+{
+
+  /**
+   * Returns an array of updated properties
+   *
+   * @param currentValue current value
+   * @param newValue new value
+   * @return array of PropertyUpdate objects
+   */
+  private PropertyUpdate[] getUpdatedProperties(FObject currentValue, FObject newValue) {
+    Map diff = currentValue.diff(newValue);
+    Iterator i = diff.keySet().iterator();
+
+    int index = 0;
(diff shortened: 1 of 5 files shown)


src/foam/dao/history/HistoryRecord.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao.history',
+  name: 'HistoryRecord',
+  documentation: `Contains an array of property updates`,
```
kgrgreer: Can use single-quotes ' instead of ` when only one line.

src/foam/dao/history/HistoryRecord.js
```diff
@@ -0,0 +1,27 @@
...
+  properties: [
+    {
+      class: 'FObjectProperty',
```
kgrgreer: Should just be a String. We don't want to store the whole User, which contains all of their contact information and can be quite large. Just their name. Maybe something like:
Greer, Kevin (kgr)
last-name, first-name (id)

src/foam/dao/history/HistoryRecord.js
```diff
@@ -0,0 +1,27 @@
...
+      of: 'foam.nanos.auth.User',
+      name: 'user'
+    },
+    {
+      class: 'Long',
```
kgrgreer: class: 'DateTime'

src/foam/dao/history/HistoryRecord.js
```diff
@@ -0,0 +1,27 @@
...
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao.history',
+  name: 'HistoryRecord',
+  documentation: `Contains an array of property updates`,
+  properties: [
```
kgrgreer: You also need to store the 'id' of the object which was updated. Maybe call it 'objectId' or 'key'.

kgrgreer: You need to specify the primaryKey of HistoryRecord. Something like:
ids: [ 'key', 'timestamp' ] (but then you have to make sure there aren't two updates the same ms), or
ids: [ 'key', 'seqNo' ] (and add a seqNo property)

src/foam/dao/history/HistoryRecord.js
kirkeaton: Done

src/foam/dao/history/HistoryRecord.js
kirkeaton: Done

src/foam/dao/history/HistoryRecord.js
kirkeaton: Done

java_src/foam/dao/history/HistoryDAO.java
```diff
@@ -0,0 +1,66 @@
...
+
+    // add new history record
+    HistoryRecord historyRecord = new HistoryRecord();
+    historyRecord.setUser(user);
+    historyRecord.setTimestamp(new Date().getTime());
+    historyRecord.setUpdates(getUpdatedProperties(current, obj));
+    historyDAO.put(historyRecord);
+
+    return super.put(obj);
+  }
```
kgrgreer, line 81: Implement remove() to also remove all of an object's HistoryRecords when it is removed.

kirkeaton: Why would we want to remove the history records for an object when it's removed?

src/foam/dao/history/HistoryRecord.js
kirkeaton: If I add the seqNo to the property then can I assume that the DAO for the HistoryRecord model is a SequenceNumberDAO?

java_src/foam/dao/history/HistoryDAO.java
```diff
@@ -69,8 +70,9 @@ public FObject put(FObject obj) {
     FObject current = this.find(obj);
 
     // add new history record
+    String objectId = (String) ((PropertyInfo) obj.getClassInfo().getAxiomByName("objectId")).f(obj);
```
kgrgreer: Object id = obj.getClassInfo().getId(obj);  (you don't know if it will be a String).

java_src/foam/dao/history/HistoryDAO.java
```diff
@@ -70,9 +69,8 @@ public FObject put(FObject obj) {
     FObject current = this.find(obj);
 
     // add new history record
-    String objectId = (String) ((PropertyInfo) obj.getClassInfo().getAxiomByName("objectId")).f(obj);
     HistoryRecord historyRecord = new HistoryRecord();
-    historyRecord.setObjectId(objectId);
+    historyRecord.setObjectId(obj.getClassInfo().getId());
```
kgrgreer: I was wrong, that's the id of the ClassInfo, not the object.
Your original code was right, except that it should have been 'id' instead of 'objectId'.

kirkeaton: That's assuming that the object has a property called id. If it doesn't then a NPE will be thrown.

java_src/foam/dao/history/HistoryDAO.java
kgrgreer: We can't keep data forever. When I delete a record, I want to delete it's history also. Think of a bug/issue tracking system or a patient record. When I delete the issue or patient, I want to have a cascading delete remove all of their data. If I want to keep the data around then I just don't delete it.

kirkeaton: Okay sure. I've added the remove function.

java_src/foam/dao/history/HistoryDAO.java
kgrgreer: All FObjects stored in a DAO have an id property, either directly or indirectly. If you don't have a property called 'id', then you do ids: [ 'keypart1', ..., 'keypartn' ] and it makes you a psedo-property named 'id' which is an alias to your id(s).

kirkeaton: Are you sure that this is the case in Java?

java_src/foam/dao/history/HistoryDAO.java
kirkeaton: Okay changed it.

PR #431, opened 2017-06-08
Java @Override Flag Support

TanayParikh:
- Added support for `@Override` flag. Flag permits for better error handling.
> When overriding a method, you might want to use the `@Override` annotation that instructs the compiler that you intend to override a method in the superclass. If, for some reason, the compiler detects that the method does not exist in one of the superclasses, then it will generate an error. For more information on `@Override`, see Annotations.
> https://docs.oracle.com/javase/tutorial/java/IandI/override.html

- Fixed `documentation` typo in nanos

### DIFF
--- a/src/foam/java/Method.js
+++ b/src/foam/java/Method.js
@@ -30,12 +30,22 @@ foam.CLASS({
       of: 'foam.java.Argument',
       name: 'args'
     },
+    {
+      class: 'Boolean',
+      name: 'override',
+      value: false
+    },
     { class: 'StringArray', name: 'throws' },
     { class: 'foam.java.CodeProperty', name: 'body' }
   ],
 
   methods: [
     function outputJava(o) {
+      if (this.override) {
+        o.indent();
+        o.out("@Override\n");
+      }
+
       o.indent();
       o.out(this.visibility, this.visibility ? ' ' : '',
         this.static ? 'static ' : '',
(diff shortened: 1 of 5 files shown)


src/foam/java/Method.js
```diff
@@ -30,12 +30,22 @@ foam.CLASS({
...
     { class: 'StringArray', name: 'throws' },
     { class: 'foam.java.CodeProperty', name: 'body' }
   ],
 
   methods: [
     function outputJava(o) {
+      if (this.override) {
+        o.indent();
+        o.out("@Override\n");
+      }
```
kirkeaton, line 47: Are you missing an `o.indent();` here or does InterfaceMethod.js just have an extra one?

kirkeaton (approved):

src/foam/java/Method.js
TanayParikh: Tested again, extra indent is needed for proper formatting. Nice catch though :)

kgrgreer:
I disagree with this approach. Would be better to validate at the Model level rather than leaving for the Java compiler to detect. This method only works for Java, not JS or Swift, and catches the error much later in the pipeline than if we just did it in FOAM (which in the future, could be flagged right in the IDE).

TanayParikh:
Thanks for the clarification, closing for now.

PR #434, opened 2017-06-08
Indexes java

marcogbarcellos:


java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.Comparator;
+
+public class TreeIndex implements Index {
+  protected Index tail;
+  protected PropertyInfo prop;
+  protected long selectCount;
+  protected TreeNode tree;
+  
+  public TreeIndex(PropertyInfo prop) {
+    this.prop = prop;
```
kgrgreer: The whole body of this method should just be:
  this(prop, ValueIndex.instance());

java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+    this.tree = TreeNode.getNullNode();
+  }
+  
+  public TreeIndex(PropertyInfo prop, Index tail) {
+    this.prop = prop;
+    this.selectCount = 0;
+    this.tail = tail;
+  }
+  
+  public void bulkLoad(FObject... a) {
```
kgrgreer: I think this should be (FObject[] a)

java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+    if ( this.tail instanceof ValueIndex ) {
```
kgrgreer: Why does it matter what your tail is?

java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+      this.tree = (TreeNode) this.put(this.tree, a[i]);
+    }
+  }
+  // predicate from plan, prop from index itself, Mlang Expression
+  protected Binary isExprMatch(Predicate predicate, Predicate model) {
+    Predicate foc = new Eq();
+    
+    if ( predicate != null && this.prop != null && model != null ) {
+      if ( predicate instanceof Binary && model instanceof Binary &&
+           ((Binary)model).getClassInfo().equals(((Binary)predicate).getClassInfo()) &&
```
kgrgreer: Add spaces after casts: ((Binary) model).

java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+    if ( this.tail instanceof ValueIndex ) {
+      this.tree = (TreeNode)this.tree.bulkLoad(this.prop, 0, a.length-1, a);
+    }
+    for ( int i = 0 ; i < a.length ; i++ ) {
+      this.tree = (TreeNode) this.put(this.tree, a[i]);
+    }
+  }
+  // predicate from plan, prop from index itself, Mlang Expression
+  protected Binary isExprMatch(Predicate predicate, Predicate model) {
+    Predicate foc = new Eq();
```
kgrgreer: What is this for?

java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+            Binary b = new Binary() {};
+            b.setArg2(((Binary)p).getArg2());
+            return b;
+          }
+        }
+      }
+    }
+    return null;
+  }
+  public Object get(Object key) {
```
kgrgreer: Blank line before methods.

java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+  
+  public void bulkLoad(FObject... a) {
+    if ( this.tail instanceof ValueIndex ) {
+      this.tree = (TreeNode)this.tree.bulkLoad(this.prop, 0, a.length-1, a);
+    }
+    for ( int i = 0 ; i < a.length ; i++ ) {
+      this.tree = (TreeNode) this.put(this.tree, a[i]);
+    }
+  }
+  // predicate from plan, prop from index itself, Mlang Expression
```
kgrgreer: Better docs and maybe better names needed here. Shouldn't one of the arguments be a Class?

java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+    return null;
+  }
+  public Object get(Object key) {
+    return this.tree.get(this.tree, key, this.prop);
+  }
+  
+  public Object put(Object state, FObject value) {
+    this.tree = this.tree.putKeyValue((TreeNode)state, this.prop, this.prop.f(value), value);
```
kgrgreer: Why does the index have a 'tree' instance variable? It should be stateless, that's why we're passing the state in and out. If it were just an instance variable, then we wouldn't need to bother.

Consider an Index on firstName with a tail index on lastName. The tail index can be re-used for every node of the first index, but if it has a 'tree', which tree is it?

java_src/foam/dao/index/TreeIndex.java
```diff
@@ -0,0 +1,152 @@
...
+//      Object result = this.get(key);
+//      if ( result == null) {
+//        return NotFoundPlan.instance();
+//      }
+//      Plan[] subPlans = { planSelect(result, sink, skip, limit, order, predicate) };
+//      return new AltPlan(subPlans,this.prop);
+//    }
+    TreeNode subTree = this.tree;
+    
+    Binary expr = isExprMatch(predicate, new Gt());
```
kgrgreer: It is expensive to create all of these new objects, like 'new Gt()', when we could just pass the class 'Gt'.

java_src/foam/dao/index/TreeNode.java
```diff
@@ -0,0 +1,352 @@
...
+ */
+package foam.dao.index;
+
+import foam.core.FObject;
+import foam.core.PropertyInfo;
+
+
+public class TreeNode {
+  
+  protected Index tail;
```
kgrgreer: Why does the TreeNode need to have the Index. You'll have 100's of millions of TreeNodes, so any extra size adds up.

java_src/foam/dao/index/TreeNode.java
```diff
@@ -0,0 +1,352 @@
...
+  }
+  
+  public Object get(TreeNode s, Object key, PropertyInfo prop) {
+    if ( s == null ) {
+      return s;
+    }
+    int r = prop.compare(s.key, key);
+    if ( r == 0 ) {
+      return s.value;
+    } else if ( r > 0 ) {
```
kgrgreer: No need for an 'else' when you've already return'ed.

java_src/foam/dao/index/ValueIndex.java
```diff
@@ -0,0 +1,71 @@
...
+package foam.dao.index;
+
+import foam.core.FObject;
+import foam.dao.Sink;
+import foam.mlang.predicate.Predicate;
+import java.util.Comparator;
+
+public class ValueIndex implements Index {
+  
+  protected static ValueIndex instance_ = null;
```
kgrgreer: Just assign this to new ValueIndex() here and then remove the check below.

java_src/foam/dao/MDAO.java
```diff
@@ -0,0 +1,82 @@
...
+import foam.dao.index.SelectPlan;
+import foam.dao.index.TreeIndex;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+import foam.mlang.predicate.True;
+
+public class MDAO extends AbstractDAO {
+  
+  protected ClassInfo of_;
+  protected Index index;
```
kgrgreer, line 23: End instance variables with _. ie index_ and state_.

java_src/foam/dao/MDAO.java
```diff
@@ -0,0 +1,82 @@
...
+  protected Object state;
+  
+  public MDAO(ClassInfo of_) {
+    this.of_ = of_;
+    state = null;
+    index = new AltIndex(new TreeIndex((PropertyInfo) this.of_.getAxiomByName("id")));
+  }
+  
+  public void addIndex(PropertyInfo prop) {
+    if ( ! (index instanceof AltIndex) ) {
```
kgrgreer, line 33: You don't need this check since you already set index to an AltIndex in the constructor. You could also just make index_ be of type AltIndex so you don't have to cast it since you know that's always the case.

java_src/foam/dao/MDAO.java
```diff
@@ -0,0 +1,82 @@
...
+      this.index = new AltIndex(this.index);
+    }
+    ((AltIndex) index).addIndex(new TreeIndex(prop));
+  }
+  
+  public FObject put(FObject obj) {
+    FObject oldValue = find(obj);
+    if ( oldValue != null ) {
+      index.remove(state, oldValue);
```
kgrgreer, line 42: This is a bug. You didn't update the state, so the object isn't really removed.

java_src/foam/dao/MDAO.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public FObject find(Object id) {
+    if ( id == null ) {
+      return null;
+    }
+    return (FObject)index.get(state, id);
+  }
+
+  public Sink select(Sink sink, Integer skip, Integer limit, Comparator order, Predicate predicate) {
+    Plan plan = index.planSelect(state, sink, skip, limit, order, predicate);
```
kgrgreer, line 67: The return type of planSelect() should be a SelectPlan.

java_src/foam/dao/MDAO.java
```diff
@@ -0,0 +1,82 @@
...
+    
+    if ( plan instanceof SelectPlan ) {
```
kgrgreer, line 69: This check and cast shouldn't be necessary.

java_src/foam/dao/index/CountPlan.java
```diff
@@ -8,7 +8,7 @@
 import foam.dao.Sink;
 import foam.mlang.predicate.Predicate;
 import foam.mlang.sink.Count;
-import java.util.Comparator;
+import foam.mlang.order.Comparator;
```
kgrgreer, line 11: Why was this switched from java.util.Comparator?

java_src/foam/dao/index/Index.java
```diff
@@ -18,6 +19,7 @@
   public Object put(Object state, FObject value);
   public Object remove(Object state, FObject value);
   public Object removeAll();
+  public Object get(Object state, Object key);
```
kgrgreer, line 22: Why is this needed? Maybe there should be a sub-interface called PrimaryIndex which has this extra method.

java_src/foam/dao/index/ValueIndex.java
```diff
@@ -7,12 +7,12 @@
...
+import foam.mlang.order.Comparator;
 import foam.mlang.predicate.Predicate;
-import java.util.Comparator;
 
 public class ValueIndex implements Index {
   
-  protected static ValueIndex instance_ = null;
+  protected static ValueIndex instance_ = new ValueIndex();
   
   protected FObject value;
```
kgrgreer: ValueIndex should not have a value. That's stored in the state which is passed it.

java_src/foam/dao/index/ValueIndex.java
```diff
@@ -23,16 +23,16 @@ protected ValueIndex() {
...
-      return instance_;
-    }
-    instance_ = new ValueIndex();
     return instance_;
   }
   
   public void onAdd(Sink sink) {
   }
   
+  public Object get(Object state, Object key) {
```
kgrgreer: This shouldn't be needed. See comment above about primaryIndex.

PR #440, opened 2017-06-09
fobject: get and set property methods for helping runtime class management.

drish:
created these two methods for helping with runtime class management.

i.e:

before
```java
FObject put(FObject o) { 
  PropertyInfo prop = (PropertyInfo) o.getClassInfo().getAxiomByName("brand"); 
  prop.set(o, "blue");
}
```

now:
```java
FObject put(FObject o) { 
  o.setProperty("brand", "blue");
}
```

### DIFF
--- a/java_src/foam/core/AbstractFObject.java
+++ b/java_src/foam/core/AbstractFObject.java
@@ -69,6 +69,6 @@ public abstract class AbstractFObject
 
   public Object getProperty(String prop) {
     PropertyInfo property = ((PropertyInfo) getClassInfo().getAxiomByName(prop));
-    return property == null ? null: property.get(this);
+    return property == null ? null : property.get(this);
   }
 }
--- a/java_src/foam/core/AbstractFObject.java
+++ b/java_src/foam/core/AbstractFObject.java
@@ -63,18 +63,12 @@ public abstract class AbstractFObject
 
   public FObject setProperty(String prop, Object value) {
     PropertyInfo property = ((PropertyInfo) getClassInfo().getAxiomByName(prop));
-    if (property == null) {
-      return null;
-    }
-    property.set(this, value);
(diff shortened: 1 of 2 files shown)


java_src/foam/core/AbstractFObject.java
```diff
@@ -60,4 +60,14 @@ public int compareTo(Object o) {
 
     return 0;
   }
+
+  public FObject setProperty(String prop, Object value) {
+    PropertyInfo property = ((PropertyInfo) getClassInfo().getAxiomByName(prop));
+    property.set(this, value);
```
kgrgreer: Add check that 'property' exists, otherwise you'll get NPE's.

java_src/foam/core/AbstractFObject.java
```diff
@@ -63,11 +63,18 @@ public int compareTo(Object o) {
 
   public FObject setProperty(String prop, Object value) {
     PropertyInfo property = ((PropertyInfo) getClassInfo().getAxiomByName(prop));
+    if (property == null) {
```
kgrgreer: This will also break users that chain setProperty() calls because this method is expected to return 'this'.

if ( property != null ) property.set(this, value);

return this;

java_src/foam/core/AbstractFObject.java
```diff
@@ -63,11 +63,18 @@ public int compareTo(Object o) {
...
+      return null;
+    }
     property.set(this, value);
     return this;
   }
 
   public Object getProperty(String prop) {
-    return ((PropertyInfo) getClassInfo().getAxiomByName(prop)).get(this);
+    PropertyInfo property = ((PropertyInfo) getClassInfo().getAxiomByName(prop));
+    if (property == null) {
```
kgrgreer: return property == null ? null : property.get(this);

PR #441, opened 2017-06-10
Mult-process Node JS servers using using boxes and sockets

mdittmer:
Example script:

```js
require('./src/foam.js');

var server = foam.box.node.ServerBox.create({
  port: 7000
}, ctx);

var numPeers = 3;

// Create a full mesh network:
// Wait for all peers to bind to this server, then bind them all to each other.
server.peers$.sub(function(sub, _, __, prop) {
  var peers = prop.get();
  if ( peers.length !== numPeers ) return;

  sub.detach();

  for ( var i = 0; i < peers.length; i++ ) {
    for ( var j = i + 1; j < peers.length; j++ ) {
      server.bindPeers(peers[i], peers[j]);
      server.bindPeers(peers[j], peers[i]);
    }
  }
});

// Start process-local server.
server.start();

// Create peer servers.
for ( var i = 0; i < numPeers; i++ ) {
  server.createLocalPeer();
}
```


mdittmer:
@adamvy PTAL. I'm interested in design feedback and good-idea/bad-idea on my in-comment `TODO`s.

mdittmer:
Simplified implementation. New sample script:

```js
require('./src/foam.js');

var socketBoxToChild = foam.box.node.ForkBox.create(null, ctx);
// Do what you will with a socketBox connected to child process('s registry).
// Note: Default script will not store socket to parent that was used to expose socket to child.
```

kgrgreer:
Hi Mark,

  What does this do? This is for when your data is stored in some server
other than than Node itself and Node is just acting as a stateless relay?

Thanks,

KGR

mdittmer:
This allows you to federate services on the same machine with multiple NodeJS instances. The intention is that parallel NodeJS applications can wire up particular services to run on particular "workers" that are forked and accessed from the parent using a `ForkBox`.

PR #443, opened 2017-06-12
Box pipeline

mdittmer:
Add support for composing box `Runnable`s into a pipelines. Merging + forking pipeline parts is supported, but cycles are not. Each `Runnable` is registered once as a box service; overriding registry in context allows for creating multiple workers and picking one for units of work.

Tests coming soon.

Example script:

```js
require('./src/foam.js');
require('./src/foam/nanos/nanos.js');
foam.CLASS({ name: 'MyInt', properties: [ { class: 'Int', name: 'value' } ] });
foam.CLASS({
  name: 'SquareIt',
  extends: 'foam.box.Runnable',

  methods: [
    function run(x) {
      this.output(MyInt.create({ value: x.value * x.value }));
    }
  ]
});
foam.CLASS({
  name: 'AddOne',
  extends: 'foam.box.Runnable',

  methods: [
    function run(x) {
      this.output(MyInt.create({ value: x.value + 1 }));
    }
  ]
});
// ctx = ctx.__subContext__.createSubContext({
//   registry: foam.box.LogBox.create({ name: 'registry', delegate: ctx.registry }, ctx)
// });

// Construct:
// ^2 --           -- ^2
//       >-- ^2 --<
// +1 --           -- +1

var builder = foam.box.pipeline.PipelineBuilder.create(null, ctx);
var shared = builder.then(SquareIt.create());
shared.then(SquareIt.create());
shared.then(AddOne.create());
var squareItFirst = builder.first(SquareIt.create()).build();
var addOneFirst = builder.first(AddOne.create()).build();

var initialMessage = foam.box.Message.create({
  object: MyInt.create({ value: 2 })
});

var boxes = builder.buildAll();

//      -- ^2 --          -- ^2
// 2 --<        >-- ^2 --<
//      -- +1 --          -- +1
//
// ==> 256
// ==> 17
// ==> 81
// ==> 10
boxes[0].send(initialMessage);
boxes[1].send(initialMessage);
```


src/foam/box/pipeline/Pipeline.js
```diff
@@ -0,0 +1,82 @@
...
+      of: 'foam.box.PromisedBox',
+      name: 'remoteInput',
+      factory: function() { return this.PromisedBox.create(); }
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.box.Box',
+      name: 'output',
+      postSet: function(old, nu) {
+        this.runnable && this.runnable.output = nu;
```
m-cheung: This gives you an `Invalid left-hand side in assignment` error. Brackets are required.
`this.runnable && (this.runnable.output = nu);`

src/foam/box/Runnable.js
```diff
@@ -0,0 +1,96 @@
...
+      and sends them to its outputBox, reporting any errors to errorBox.`,
+
+  requires: [ 'foam.box.Message' ],
+
+  properties: [
+    {
+      class: 'String',
+      name: 'ioRelationshipType',
+      documentation: 'The n:m relationship type of input-to-output.',
+      value: '1:1'
```
m-cheung, line 33: What uses this property? What if `n` and/or `m` is unknown? For example, what if I input one file which can result in a variable (`m`) number of computations?

mdittmer:
@m-cheung PTAL at latest commit. Is this the right fix to the issue we discussed offline?

src/foam/box/pipeline/PipelineBuilder.js
```diff
@@ -243,12 +243,12 @@ foam.CLASS({
             the merge points in the pipeline.`
       },
       code: function() {
-        if ( this.parents.length === 0 ) return [ this.build_() ];
+        if ( this.parents.length === 0 ) return [ this.buildAll() ];
```
m-cheung: This change is not present in my local version (this change actually causes an infinite loop)

PR #448, opened 2017-06-13
Fileserver

nick-prat:
File server, used as {server}/file/{filename}. Limited to only being able to send files in or below it's directory

### DIFF
--- a/src/foam/nanos/http/FileServlet.java
+++ b/src/foam/nanos/http/FileServlet.java
@@ -1,0 +1,63 @@
+package foam.nanos.http;
+
+import com.google.api.client.util.IOUtils;
+
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.*;
+
+/**
+ * Created by nick on 12/06/17.
+ */
+public class FileServlet
+    extends HttpServlet
+{
+
+  public static final String SERVLET_NAME = "file";
+
+  private void fileNotFoundError(HttpServletResponse resp, String file) {
+    resp.setStatus(HttpServletResponse.SC_NOT_FOUND);
+    resp.setContentType("application/json");
+
+    try {
+      PrintWriter pw = resp.getWriter();
+      pw.write(
+          "{\"error\": \"File not found\"," +
+              "\"filename\": \""+file+"\"}");
+    } catch (IOException ignore) {}
+  }
+
+  @Override
+  protected void doGet(HttpServletRequest req, HttpServletResponse resp) {
+    String pathInfo = req.getPathInfo();
+    String filePath = pathInfo.substring(SERVLET_NAME.length() + 2);
+    String fileName = "";
+
+    String[] paths = filePath.split("/");
+    if(paths.length > 0) {
+      fileName = paths[paths.length - 1];
+    } else {
+      fileNotFoundError(resp, filePath);
+      return;
+    }
+
+    try {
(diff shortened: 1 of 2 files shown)


src/foam/nanos/http/FileServlet.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  protected void doGet(HttpServletRequest req, HttpServletResponse resp) {
+    String pathInfo = req.getPathInfo();
+    String filePath = pathInfo.substring(SERVLET_NAME.length() + 2);
+    String fileName = "";
+
+    String[] paths = filePath.split("/");
+    if(paths.length > 0) {
```
kgrgreer: See FOAM style guideline. Formatting should be:
if ( paths.length > 0 ) {

src/foam/nanos/http/FileServlet.java
```diff
@@ -0,0 +1,63 @@
...
+    } catch (IOException ignore) {}
+  }
+
+  @Override
+  protected void doGet(HttpServletRequest req, HttpServletResponse resp) {
+    String pathInfo = req.getPathInfo();
+    String filePath = pathInfo.substring(SERVLET_NAME.length() + 2);
+    String fileName = "";
+
+    String[] paths = filePath.split("/");
```
kgrgreer, line 58: If I insert "../" into the path, does it let me read from a parent/ancestor directory?

src/foam/nanos/http/FileServlet.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+public class FileServlet
+    extends HttpServlet
+{
+
+  public static final String SERVLET_NAME = "file";
```
kgrgreer: s/file/static/

src/foam/nanos/http/FileServlet.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    try {
+      File srcFile = new File(filePath);
+
+      String path = srcFile.getAbsolutePath();
+      String cwd = System.getProperty("user.dir");
+
+      if ( srcFile.isFile() && srcFile.canRead() && cwd.equals(path.substring(0, cwd.length())) ) {
+        resp.setContentType("application/octet-stream");
```
kgrgreer: I think the content type should be set based on the extension of the file requested. Add a map to constants: which includes the types we're likely to use (ie. .js, .gif, .xvg, .png, .csv, .txt, .html, etc.)

See: http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types

src/foam/nanos/http/FileServlet.java
nick-prat: It seems like `..` messes up the `URI.getPath` call in `HttpHandler` and it doesn't even reach the `FileServlet`

src/foam/nanos/http/FileServlet.java
nick-prat: Servlet's can't contain a / in their name in the way it's setup currently

kgrgreer: s/file/static/ means: replace "file" with "static"

It's a short-hand programmer notation.

src/foam/nanos/http/FileServlet.java
kgrgreer: Even if the ../ is after the servlet name?  So /file/../foo.txt

PR #455, opened 2017-06-15
PostgresDAO

drish:
This is a work in progress.


src/foam/dao/PgConnectionPool.java
```diff
@@ -0,0 +1,32 @@
+package foam.dao;
+
+import org.apache.commons.dbcp2.BasicDataSource;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public abstract class PgConnectionPool {
+
+    protected static String base = "jdbc:postgresql://";
```
kgrgreer: See FOAM style guide: https://github.com/foam-framework/foam/wiki/FOAM2-Style-Guidelines
Indentation is two spaces and spacing is like:
if ( foobar ) {

src/foam/dao/PgConnectionPool.java
```diff
@@ -0,0 +1,32 @@
+package foam.dao;
```
kgrgreer: Maybe foam.dao.postgres or .pg would be better to not fill up the main package.

src/foam/dao/PgConnectionPool.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.commons.dbcp2.BasicDataSource;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public abstract class PgConnectionPool {
```
kgrgreer: Is this Postgres specific, or is this just generic JDBC? Would the same code work for MySQL or Oracle?

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+    public FObject remove(FObject o) {
+
+        try {
+
+            SQLData data = new SQLData(o);
+            Connection c = PgConnectionPool.getConnection();
+
+            PreparedStatement smt = c.prepareStatement(data.createDeleteStatement());
+
+            smt.setLong(1, ((Long) o.getProperty("id")));
```
kgrgreer: We don't know that the 'id' property is a Long. (Which is okay for now, but at least add a TODO.)

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+
+            int removed = smt.executeUpdate();
+            if (removed == 0) {
```
kgrgreer: When does this happen. The DAO interface does not throw an exception if an object wasn't removed but not found. Only if there was an issue connecting to the database.

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * maps a database result row to an FObject
+     *
+     * @param row
+     * @return FObject
+     * @throws SQLException
+     */
+    private FObject createFObject(ResultSet row) throws Exception {
+
+        if (getOf() == null) {
```
kgrgreer: 'of' should be set and enforced in the constructor instead of here.

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+            throw new Exception("`Of` is not set");
+        }
+
+        List<PropertyInfo> props = getOf().getAxioms();
```
kgrgreer: Call getAxionsByClass(Property.class) to only get the Property Axioms.

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+        FObject result = (FObject) getOf().getObjClass().newInstance();
+
+        // set fields
+        for(int i = 0; i < props.size(); i++ ) {
+            String propName = props.get(i).getName();
+            Object value = row.getObject(i + 1);
+            result.setProperty(propName, value);
```
kgrgreer: If you call the set method on props.get(i) it will avoid looking up the property by name and be faster.

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+        return result;
+    }
+
+    @Override
+    public FObject find(Object o) {
+
+        try {
+
+            String tableName = getOf().getObjClass().getSimpleName().toLowerCase();
```
kgrgreer: This generates garbage and should just be done once in the constructor.

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+            Connection c = PgConnectionPool.getConnection();
+
+            String sql = "select * from " + tableName;
+            sql += " where id = ?";
+
+            PreparedStatement smt = c.prepareStatement(sql);
+            smt.setLong(1, ((Long) o));
```
kgrgreer: Add TODO to support other id types.

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public FObject find(Object o) {
+
+        try {
+
+            String tableName = getOf().getObjClass().getSimpleName().toLowerCase();
+            Connection c = PgConnectionPool.getConnection();
+
+            String sql = "select * from " + tableName;
+            sql += " where id = ?";
```
kgrgreer: Should you add 'limit 1' to the query?

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+                // put data into statement preserving order.
+                int i = 1;
+                for (Object value: data.getValues()) {
+                    smt.setObject(i, value);
+                    i++;
+                }
+
+                int inserted = smt.executeUpdate();
+                if (inserted == 0) {
+                    throw new SQLException("Error while inserting.");
```
kgrgreer: If you throw this exception, or others, how is the connection returned to the pool?

src/foam/dao/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Override
+    public FObject put(FObject obj) {
+
+        try {
+            Connection c = PgConnectionPool.getConnection();
+
+            // updating existing one
+            // when FObject isIdSet is null it returns 0 , is it ok ?
+            if ((Long) obj.getProperty("id") != 0) {
```
kgrgreer: Better to check that the 'id' doesn't equal the defaultValue, since 0 could be a valid value. Can be done later, just add a TODO.

src/foam/dao/SQLData.java
```diff
@@ -0,0 +1,146 @@
+package foam.dao;
+
+import foam.core.FObject;
+import foam.core.PropertyInfo;
+
+import java.util.*;
+import java.util.ArrayList;
```
kgrgreer: Next two imports are redundant if you've already imported java.util.*

src/foam/dao/SQLData.java
```diff
@@ -0,0 +1,146 @@
+package foam.dao;
+
+import foam.core.FObject;
+import foam.core.PropertyInfo;
+
```
kgrgreer: remove blank lines between imports

src/foam/dao/PgConnectionPool.java
drish: Yeah, gonna wrap all the database files on its own package.

src/foam/dao/SQLData.java
```diff
@@ -0,0 +1,146 @@
...
+  //TODO(drish): use uuids ?
+  protected Object id;
+  protected String table;
+  protected ArrayList<String> columnNames;
+  protected ArrayList<Object> values;
+
+  public SQLData(FObject obj) throws IllegalStateException, IllegalAccessException {
+
+    table = obj.getClass().getSimpleName().toLowerCase();
```
kgrgreer: Using simplename will cause conflicts between DAOS for models in different packages with same name.

src/foam/dao/PostgresDAO.java
kgrgreer: Faster and generates less garbage to combine this with the above line. += is expensive in Java

src/foam/dao/PgConnectionPool.java
drish: Yes, It's a generic JDBC database pool implementation, and yes it will work with any other relational database compatible with jdbc.

@kgrgreer , do you think the pool size number should be defined at runtime based on runtime properties, such as memory size , cpus etc ? also, do you think it should be something configurable on the model ?

https://commons.apache.org/proper/commons-dbcp/

src/foam/dao/PostgresDAO.java
drish: Ok, just used Long for initial implementation.

I think the default data types for ids are: `serial` which is translated to Long in Java, or uuids.

src/foam/dao/PgConnectionPool.java
drish: gonna also rename to just ConnectionPool

src/foam/mlang/mlangJava.js
```diff
@@ -217,6 +238,10 @@ foam.CLASS({
                 + '  if ( ! getArgs()[i].f(obj) ) return false;\n'
                 + '}\n'
                 + 'return true;'
+    },
+    {
+      name: 'toSQL',
+      javaCode: 'return null;'
```
kgrgreer: I guess this is just a place-holder to be done later.

src/foam/mlang/mlangJava.js
```diff
@@ -280,6 +305,10 @@ foam.CLASS({
       name: 'f',
       // TODO(adamvy): Is there a better option than all the Comparable casts?
       javaCode: 'return ((Comparable)getArg1().f(obj)).compareTo((Comparable)getArg2().f(obj)) == 0;'
+    },
+    {
+      name: 'toSQL',
+      javaCode: 'return getArg1().f(obj).toString() + "=" + getArg2().f(obj).toString();'
```
kgrgreer: Call toSQL() on the args, not toString(). toSQL() for ConstantExpr will have to escape HTML so that we don't get SQL injection errors.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * maps a database result row to an FObject
+     *
+     * @param row
+     * @return FObject
+     * @throws SQLException
+     */
+    private FObject createFObject(ResultSet row) throws Exception {
+
+        if (getOf() == null) {
```
kgrgreer: See the FOAM coding standard: https://github.com/foam-framework/foam/wiki/FOAM2-Style-Guidelines
Indent is two spaces and this line should be:
    if ( getOf() == null ) {   (extra spaces)

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,175 @@
...
+            throw new Exception("`Of` is not set");
+        }
+
+        List<PropertyInfo> props = getOf().getAxioms();
+        FObject result = (FObject) getOf().getObjClass().newInstance();
+
+        // set fields
+        for(int i = 0; i < props.size(); i++ ) {
```
kgrgreer: formatting

src/foam/mlang/mlangJava.js
drish: yeah, its just a place holder for now, will add  a TODO, otherwise would java wouldnt compile because the classes are not meeting the interface methods.

src/foam/core/AbstractPropertyInfo.java
```diff
@@ -36,6 +38,16 @@ public void toJSON(foam.lib.json.Outputter outputter, StringBuilder out, Object
   }
 
   @Override
+  public String createStatement() {
+    return getName();
+  }
+
+  @Override
+  public Object prepareStatement() {
```
kgrgreer, line 46: prepareStatement should return void and be passed the prepared statement. This method should do nothing.

src/foam/dao/EnabledAware.java
```diff
@@ -1,12 +1,15 @@
...
+
+import java.sql.PreparedStatement;
 import java.util.Date;
 
 public interface EnabledAware {
   public final static Predicate ENABLED = new Predicate() {
     public boolean f(foam.core.FObject obj) { return true; }
-
+    public String createStatement(String table) { return ""; }
+    public void prepareStatement(PreparedStatement stmt) { return; }
```
kgrgreer, line 12: return isn't needed;

src/foam/dao/pg/ConnectionPool.java
```diff
@@ -0,0 +1,32 @@
...
+import java.sql.SQLException;
+
+public abstract class ConnectionPool {
+
+    protected static String base = "jdbc:postgresql://";
+    // TODO(drish) get an accurate number for pool size.
+    protected static int poolSize = 4;
+    protected static BasicDataSource pool;
+
+    public static void setup(String host, String port, String dbName, String username, String password) {
```
kgrgreer, line 15: This shouldn't be static, otherwise how are you going to have more than one?

src/foam/dao/pg/ConnectionPool.java
```diff
@@ -0,0 +1,32 @@
...
+            pool.setUsername(username);
+            pool.setDriverClassName("org.postgresql.Driver");
+            pool.setUrl(url);
+            pool.setInitialSize(poolSize);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
+    public static Connection getConnection() throws SQLException {
```
kgrgreer, line 29: Shouldn't be static.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,176 @@
...
+
+import java.sql.*;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+
+public class PostgresDAO extends ProxyDAO {
```
kgrgreer: It shouldn't extend ProxyDAO, it should extend AbstractDAO. This is a real DAO implementation (Strategy), not a decorator or proxy.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,176 @@
...
+
+    public PostgresDAO(DAO delegate, String host, String port, String dbName, String username, String password) {
+        setDelegate(delegate);
+
+        if (dbName == null || username == null) {
```
kgrgreer: Follow the FOAM coding standard for spacing throughout this PR.

src/foam/dao/pg/SQLData.java
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * creates a prepared statement for update
+   * in the following format:
+   *
+   *  update tableName set (col1, col2) = (?, ?, ?) where id = id;
+   *
+   * @return String
+   */
+  public String createUpdateStatement() {
```
kgrgreer, line 67: These prepared statements will never change for a particular DAO, so it's very inefficient to generate them for each statement rather than just reuse them.

src/foam/dao/pg/SQLData.java
```diff
@@ -0,0 +1,144 @@
...
+import foam.core.PropertyInfo;
+
+import java.util.*;
+import java.util.ArrayList;
+import java.util.StringJoiner;
+
+/**
+ * prepares sql info based on a FObject
+ */
+public class SQLData {
```
kgrgreer, line 13: Why is this class needed? Can't this just be done in the Postgres DAO?

kgrgreer: This builds a collection of columnNames, which should be the same every time, so should be calculated in the DAO constructor rather than once per update, which is wasteful.
Also, copying the values out of the object itself before processing just generates extra work and garbage, especially for primitive values like numbers and booleans which will need to be promoted to real objects and then garbage collected.

kirkeaton:
@drish @kgrgreer This can be closed now as PostgresDAO based on this PR has now been merged with some modifications

PR #457, opened 2017-06-16
On network message payload creation, null out context so it's not sent.

jlhughes:


adamvy:
On June 16, 2017 12:42:09 AM EDT, Joel Hughes <notifications@github.com> wrote:
>
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/foam-framework/foam2/pull/457
>
>-- Commit Summary --
>
>* On network message payload creation, null out context so it's not
>sent.
>
>-- File Changes --
>
>    M src/lib/Stub.js (6)
>
>-- Patch Links --
>
>https://github.com/foam-framework/foam2/pull/457.patch
>https://github.com/foam-framework/foam2/pull/457.diff
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/foam-framework/foam2/pull/457

This will null out the first argument for all network communications, but we use RPCs for more than just DAOs, all those other uses will break with this change.

Would be safer to just update ClientDAO.  It might make more sense for the network RPCs to call methods which don't contain a context, just regular select/put/remove.  That way the server dao can set up its own context.

jlhughes:
Adam, thanks for pointing me to ClientDAO, I wasn't aware of it.  Yes, this is a much better place for the change.

jlhughes:
This is incomplete; need the serializable consideration in the other DAO operations put, remove,...

src/lib/box.js
```diff
@@ -698,6 +698,15 @@ foam.CLASS({
     'foam.dao.BoxDAOListener'
   ],
   methods: [
+    function put_(x, obj) {
+      return this.SUPER(this.Serializable.isInstance(x) ? x : null, obj);
```
kgrgreer: I think we can just pass null, since we would never want to serialize the context.

PR #459, opened 2017-06-17
Add a hook to JSON stringify that makes it only output own properties.

mcarcaso:
I noticed stringify outputting values from expressions and that didn't seem right to me. How do you feel about this change? I could change the default value to preserve the current behaviour but the current behaviour feels wrong to me.

### DIFF
--- a/src/foam/core/JSON.js
+++ b/src/foam/core/JSON.js
@@ -155,6 +155,11 @@ foam.CLASS({
     },
     {
       class: 'Boolean',
+      name: 'outputOwnPropertiesOnly',
+      value: true
+    },
+    {
+      class: 'Boolean',
       name: 'outputClassNames',
       value: true
     },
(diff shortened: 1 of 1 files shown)


src/foam/core/JSON.js
```diff
@@ -155,6 +155,11 @@ foam.CLASS({
     },
     {
       class: 'Boolean',
+      name: 'outputOwnPropertiesOnly',
+      value: true
```
kgrgreer, line 159: This is probably the wrong default value since it changes the behaviour for others.

mcarcaso:
You're right that it changes the behavior but I don't think that behavior
is worth preserving. It seems that the special case would be to output
expressions as I don't know why you'd want that behavior normally.

mdittmer:
Is "own" here "belonging to to concrete class", "set on instance_", or
something else entirely?

mcarcaso:
It's the result of hasOwnProperty which I think is just it being set on
instance_

PR #460, opened 2017-06-19
ENUM Property Getter Return Type Includes Class Name

TanayParikh:
Property getters for Enum types now include the class name when returning the default value. Resolves compilation errors. 

## Before
```java
public java.lang.Enum getAdrTp() {
  if ( ! AdrTpIsSet_ ) {
   return ADDR;
  }
  return AdrTp_;
}
```

## After
```java
public java.lang.Enum getAdrTp() {
  if ( ! AdrTpIsSet_ ) {
   return com.acme.AddressCode.ADDR;
  }
  return AdrTp_;
}
```

### DIFF
--- a/src/foam/core/Enum.js
+++ b/src/foam/core/Enum.js
@@ -315,6 +315,12 @@ foam.CLASS({
         return of && of.VALUES[0];
       },
     },
+    {
+      name: 'javaValue',
+      expression: function(of) {
+        return this.of.package + '.' + this.of.name + '.' + (of && of.VALUES[0]);
+      },
+    },
     [
       'adapt',
       function(o, n, prop) {
--- a/src/foam/core/Enum.js
+++ b/src/foam/core/Enum.js
@@ -318,7 +318,9 @@ foam.CLASS({
     {
       name: 'javaValue',
       expression: function(of) {
-        return this.of.package + '.' + this.of.name + '.' + (of && of.VALUES[0]);
+        var val = (of && of.VALUES[0].name);
+        var prefix = this.of.package + '.' + this.of.name + '.';
+        return val.includes(prefix) ? val : prefix + val;
       },
     },
     [
--- a/src/foam/core/Enum.js
+++ b/src/foam/core/Enum.js
@@ -318,9 +318,9 @@ foam.CLASS({
     {
       name: 'javaValue',
       expression: function(of) {
-        var val = (of && of.VALUES[0].name);
+        var val = (of && of.VALUES[0]);
         var prefix = this.of.package + '.' + this.of.name + '.';
-        return val.includes(prefix) ? val : prefix + val;
+        return foam.String.isInstance(val) ? val : prefix + val;
       },
     },
     [
--- a/src/foam/core/Enum.js
+++ b/src/foam/core/Enum.js
@@ -319,8 +319,7 @@ foam.CLASS({
       name: 'javaValue',
       expression: function(of) {
         var val = (of && of.VALUES[0]);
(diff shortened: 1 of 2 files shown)


kirkeaton (approved):

kirkeaton:
@TanayParikh does this still work if the Enum is in a different package? I'm thinking now if it might be necessary to include the entire package name.

TanayParikh:
@eatonk Thanks, that's a good point. I've updated to include both the package and class names within the return statement.

kgrgreer:
I think this logic should only be in the EnumProperty. Also, you need to account for the case where the 'value:' is already fully qualified, ex: value: com.acme.DayOfWeek.MONDAY.

kirkeaton:
@kgrgreer There is no "EnumProperty" class defined anywhere.

kgrgreer:
foam.core.Enum extends Property.

TanayParikh:
Thanks @eatonk, @kgrgreer. Have updated accordingly. 

**Note:** To check for the fully qualified name is `foam.String.isInstance` the correct method, or should I just check for inclusion of the package + class in the existing value (`value.includes(package + '.' + class)`)?

src/foam/core/Enum.js
```diff
@@ -315,6 +315,14 @@ foam.CLASS({
         return of && of.VALUES[0];
       },
     },
+    {
+      name: 'javaValue',
+      expression: function(of) {
+        var val = (of && of.VALUES[0]);
+        var prefix = this.of.package + '.' + this.of.name + '.';
```
kgrgreer: Isn't this the same as this.of.id + '.' ?

src/foam/core/Enum.js
```diff
@@ -315,6 +315,14 @@ foam.CLASS({
...
+        return foam.String.isInstance(val) ? val : prefix + val;
```
kgrgreer: Are the values of the conditional reversed?

TanayParikh: Does `isInstance` check if it is a fully qualified name?

src/foam/core/Enum.js
TanayParikh: Yes it is! Didn't see that, updated.

src/foam/core/Enum.js
```diff
@@ -315,6 +315,13 @@ foam.CLASS({
         return of && of.VALUES[0];
       },
     },
+    {
+      name: 'javaValue',
+      expression: function(of) {
+        var val = (of && of.VALUES[0]);
+        return foam.String.isInstance(val) ? val : this.of.id + '.' + val;
```
kgrgreer: Are the conditional values reversed?

TanayParikh: L322 is based on the understanding that `isInstance` checks for the fully qualified name (so if it is fully qualified just returns that, otherwise adds package + class -> id). If that understanding of `isInstance` is incorrect, and it only checks for a valid type, then the conditional values need to be reversed. Please let me know.

src/foam/core/Enum.js
```diff
@@ -315,6 +315,12 @@ foam.CLASS({
         return of && of.VALUES[0];
       },
     },
+    {
+      name: 'javaValue',
+      expression: function(of) {
+        return this.of.id + '.' + this.value;
```
kgrgreer: You don't need to do 'this.of' when 'of' is already a parameter. Also, you should add 'value' as a parameter, so:
expression: function(of, value) { return of.id + '.' value; }

TanayParikh: Thanks, updated.

PR #462, opened 2017-06-19
NoDisjunctionDAO

mdittmer:
*Main change*: DAO decorator for DAOs that do not support disjunction (e.g., Google Cloud Datastore).

1. Turn predicate into DNF
2. predicates = OR'd parts of DNF predicate
3. select() against each of predicates into shared DAOSink backed by synchronous
   DAO implementation
4. select() original predicate over DAOSink's dao into original sink

*Other necessary changes*: Better `toDisjunctiveNormalForm()` and `partialEval()` in some parts of `mlang.js`.


src/foam/dao/NoDisjunctionDAO.js
```diff
@@ -0,0 +1,65 @@
...
+  methods: [
+    function select_(x, sink, skip, limit, order, predicate) {
+      if ( ! predicate )
+        return this.SUPER(x, sink, skip, limit, order, predicate);
+
+      var prevPredicate;
+      var dnfPredicate = predicate;
+      // This might as well be while ( ! equals(prev, dnf) ), but counter
+      // limits time spent simplifying expression.
+      for ( var i = 0; i < 5; i++ ) {
```
jacksonic: Why is this loop needed?

src/foam/dao/NoDisjunctionDAO.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+      dnfPredicate = dnfPredicate.partialEval();
+
+      if ( ! this.Or.isInstance(dnfPredicate) )
+        return this.SUPER(x, sink, skip, limit, order, dnfPredicate);
+
+      var predicates = dnfPredicate.args;
+      var dao = this.MDAO.create({of: this.of});
+      // TODO(markdittmer): Create indices based on predicate.
+      var sharedSink = this.DAOSink.create({ dao: dao });
```
jacksonic: Why is there an MDAO in here? What about normal external caching?

src/foam/mlang/mlang.js
```diff
@@ -317,7 +317,8 @@ foam.CLASS({
       return foam.String.constantize(this.cls_.name) + '(' +
           this.arg1.toString() + ', ' +
           this.arg2.toString() + ')';
-    }
+    },
+    function contradicts(other) { return false; }
```
jacksonic: should this be a more general simplification method (that AND could call, as here, but returns a new simpler mlang rather than just bool).

jacksonic (changes requested):

src/foam/dao/NoDisjunctionDAO.js
jacksonic: Or put another way, do we really need to collect the results into an MDAO when it's going to be using a MergePlan to collect OR'd results anyway?

src/foam/mlang/mlang.js
```diff
@@ -807,6 +840,18 @@ foam.CLASS({
       }
 
       return rhs ? rhs.indexOf(lhs) !== -1 : false;
+    },
+    function toDisjunctiveNormalForm() {
+      foam.assert(this.Constant.isInstance(this.arg2),
```
jacksonic: A non-constant arg2 could still produce a viable array to use for IN, so this is going to limit its functionality. Breaking IN down into EQ's would be fine for cases where it is possible, but the fallback would be to leave the IN where it is (for indexes, for instance, IN can be optimized)

jacksonic (changes requested):

src/foam/dao/NoDisjunctionDAO.js
mdittmer: Do you mean "why not an ArrayDAO"? We won't get the right results from simply streaming the shared sink to the original sink. There may be duplicates and/or ordering issues.

src/foam/mlang/mlang.js
mdittmer: I'm open to ideas. I'm worried that in the "do not change" case, we may make expressions more complex; e.g., if `a.comparisonMethodOfSomeKind(b)` wishes to return an equivalent mLang when embedded in an `AND`, we may get: `AND(a, b, c) ==> a.comparisonMethodOfSomeKind(b) ==>  AND(AND(a, b), c)`

src/foam/dao/NoDisjunctionDAO.js
mdittmer: Sometimes `toDisjunctiveNormalForm()` will produce optimizable predicates (hence multiple passes). Given that, I was hoping the comment would explain why the loop is bounded by a counter.

src/foam/mlang/mlang.js
jacksonic: With an AND specific a.reduceWithAnd(b), just avoid returning AND.
This method is basically giving mlangs a way to express their AND partialEval rules (FOAM1 had these embedded inside AND). "When combined with other mlang b, what happens?"

src/foam/dao/NoDisjunctionDAO.js
jacksonic: MDAO is going to do a DNF and use a MergePlan to combine the results. This is basically doing the same thing, but doing the DNF and recombine multiple times. If we could skip the MDAO and re-use the merge process (this might call for a refactor of MergePlan to create a sibling MergeSink, though). In the end it's just an optimization to cut out some extra work and memory overhead.

src/foam/dao/NoDisjunctionDAO.js
jacksonic: We should do this kind of thing inside partialEval. The DNF should produce an OR of ANDs, and partialEval can do what it can recursively. If partialEval is introducing ORs, that sounds like a problem.

mdittmer:
@jacksonic PTAL

jacksonic (approved):

PR #466, opened 2017-06-20
Export service build fixes

kirkeaton:
1. Added missing import statements & try catch
2. Added java generation of `ExportDriverRegistry` class

### DIFF
--- a/src/foam/nanos/export/ExportService.java
+++ b/src/foam/nanos/export/ExportService.java
@@ -12,6 +12,8 @@ import foam.dao.MapDAO;
 import foam.dao.ProxyDAO;
 import foam.nanos.NanoService;
 
+import java.io.IOException;
+
 public class ExportService
   extends    ProxyDAO
   implements NanoService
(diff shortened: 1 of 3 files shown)


src/foam/nanos/export/ExportService.java
```diff
@@ -6,14 +6,26 @@
...
   extends    ProxyDAO
   implements NanoService
 {
 
   public void start() {
-    DAO dao = new JournaledDAO(new MapDAO(), "ExportDriverRegistry.txt");
-    dao.setOf(ExportDriverRegistry.class);
-    setDelegate(dao);
+    try {
+      DAO dao = new JournaledDAO(new MapDAO(), "ExportDriverRegistry.txt");
```
kgrgreer: Remove .txt extension.

kirkeaton: Removed. Should I be handling the IOException like this? I wasn't sure.

PR #469, opened 2017-06-21
Added cmd and cmd_ to DAO Interface

kennyqiyenkan:
- the actual function has no implementation yet


src/foam/dao/AbstractDAO.java
```diff
@@ -109,4 +109,13 @@ public DAO inX(X x) {
     dao.setX(x);
     return dao;
   }
+
+  public Object cmd_(X x, Object obj) {
+    // TODO
+    return true;
```
kgrgreer: Return null to indicate that it wasn't handled.

src/foam/dao/AbstractDAO.js
```diff
@@ -258,6 +258,14 @@ foam.CLASS({
       return this.find_(this.__context__, id);
     },
 
+    function cmd_(x, obj) {
+      return true;
```
kgrgreer: return null

PR #473, opened 2017-06-21
Java Constant Converter

kennyqiyenkan:
Users can now use the `constants` to add constants to their converted java code.


src/foam/java/Class.js
```diff
@@ -100,6 +112,11 @@ foam.CLASS({
       return this;
     },
 
+    function constant(c) {
+      this.constants.push(foam.java.constant.create(c));
```
kgrgreer: Should this be foam.java.Constant? How is this working?

src/foam/java/Constant.js
```diff
@@ -0,0 +1,72 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.java',
+  name: 'Constant',
+
+  properties: [
+    'name',
+    'javaType',
```
kgrgreer: This is already java specific, so we just need 'type' with java being implied.
Also, we only need 'value', which is the 'javaValue'.

src/foam/java/Constant.js
```diff
@@ -0,0 +1,72 @@
...
+    'documentation'
+  ],
+
+  methods: [
+    function outputJava(o) {
+      o.indent();
+
+      if ( this.documentation != undefined ) {
+        o.out('/* ', this.documentation, ' */\n');
+        o.indent();
```
kgrgreer, line 35: What are the next two lines for?

src/foam/java/Constant.js
```diff
@@ -0,0 +1,72 @@
...
+        o.indent();
+      }
+
+      o.out('public static final ');
+
+      if ( this.type == undefined ) {
```
kgrgreer: This logic should be in the factory for the 'type' property.

src/foam/java/Constant.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      o.out(' ' + this.name);
+      o.out(' = ');
+
+      // if the user has declared a type that requires a custom instantiation,
+      // but not declare a javaType for it, the Java code should break on compile
+      if ( this.javaValue == undefined ) {
+        if ( foam.String.isInstance(this.value) ) {
+          o.out('\"' + this.value + '\"');
```
kgrgreer: You'll need to escape this.value in case it contains "'s.

src/foam/java/Constant.js
kennyqiyenkan: the tabbing doesn't come out well after the `\n`. The two indents fixes that.

src/foam/java/Constant.js
kennyqiyenkan: Should that not be handled by the user?

src/foam/java/Constant.js
kennyqiyenkan: what would you mean by that? As in we do not need to check if the type is undefined?

src/foam/java/Constant.js
kennyqiyenkan: type is understandable.

as for javaValue/value, how would we be able to discern if they have their own custom instantiation or they are simply using a primitive?

src/foam/java/Constant.js
kgrgreer: The user could do something like:
constants: { QUOTE_CHARACTER: '"' } OR
constants: { QUOTE_CHARACTER: "\"" }

Which you would output as """. How could the user handle this?

kennyqiyenkan: mm.. I believe I see your point.

PR #475, opened 2017-06-22
FOAM FObject ⇔ CSV Parsing

TanayParikh:
- Implements two directional FObject to CSV parsing.

### Can be used with the following commands
```js
foam.lib.csv.stringify(FObject)
```

```js
foam.lib.csv.objectify(String)
```

## ⇔
```js
foam.lib.csv.objectify(foam.lib.csv.stringify(FObject))
```


src/foam/core/CSV.js
```diff
@@ -0,0 +1,382 @@
...
+          for ( var i = 0 ; i < ps.length ; i++ ) {
+            this.outputProperty(o, ps[i]);
+          }
+
+          this.removeTrailingComma();
+        },
+        Array: function(o) { /* Ignore arrays in CSV */ },
+        Function: function(o) { /* Ignore functions in CSV */ },
+        Object: function(o) {
+          debugger;
```
kirkeaton: Remove debugger statement

TanayParikh: Thanks, removed and replaced with TODO

src/foam/core/CSV.js
```diff
@@ -0,0 +1,382 @@
...
+  properties: [
+    {
+      name: 'fromCSV',
+      value: function fromCSV(value, ctx, prop, csv) {
+        return foam.csv.parse(value, null, ctx);
+      }
+    },
+    {
+      name: 'toCSV',
+      value: function toCSV(value, outputter) { return value; }
```
kgrgreer: What is the outputter used for?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,382 @@
...
+      value: '\n'
+    },
+    {
+      class: 'String',
+      name: 'undefinedStr',
+      value: ''
+    },
+    {
+      class: 'Boolean',
+      name: 'outputDefaultValues',
```
kgrgreer: I don't think this works for CSV, since the location is always present, making it impossible to tell the difference between the default value and an empty string.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,382 @@
...
+    // TODO: Copied from foam.core.json (move to common library)
+    function nl() {
+      if ( this.nlStr && this.nlStr.length ) {
+        this.out(this.nlStr);
+      }
+
+      return this;
+    },
+
+    // TODO: Copied from foam.core.json (move to common library)
```
kgrgreer: Do these still apply? Do we need to escape commas or whatever we configured the delimiter to be?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,382 @@
...
+
+    // End a nested object, 1 fewer `__` within object name
+    function end() {
+      this.nestedObjectNames.pop();
+      return this;
+    },
+
+    // TODO: Copied from foam.core.json (move to common library)
+    function nl() {
+      if ( this.nlStr && this.nlStr.length ) {
```
kgrgreer: With JSON, nl's are optional, but with CSV they're mandatory.

src/foam/core/CSV.js
TanayParikh: Removed.

src/foam/core/CSV.js
TanayParikh: Removed.

src/foam/core/CSV.js
TanayParikh: Removed.

src/foam/core/CSV.js
TanayParikh: Partially. I removed everything except the escaping of `"`. We still need that as all CSV properties are surrounded by `"` by default (configurable), and it would cause an issue if they are present in the string values.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.csv',
+  name: 'Outputer',
```
kgrgreer: Correct spelling is Outputter. Two t's.

TanayParikh: Should I correct it in JSON.js as well?

TanayParikh: Fixed for `CSV`

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+
+    function externalProperty(o, p) {
+      if ( ! this.propertyPredicate(o, p) ) return false;
+      if ( p.isDefaultValue(o[p.name]) ) return false;
+
+      return true;
+    },
+
+    function outputHeaderTitle(o) {
+      this.out(this.includeQuotes ? '"' : '').writePrefix().out(o, this.includeQuotes ? '"' : '', this.delimiter);
```
kgrgreer: If enabled, you'll need to escape " characters.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+
+  properties: [
+    {
+      class: 'String',
+      name: 'buf_',
+      value: ''
+    },
+    {
+      class: 'Boolean',
+      name: 'includeQuotes',
```
kgrgreer: I don't think we need this.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+      value: true
+    },
+    {
+      class: 'String',
+      name: 'delimiter',
+      value: ','
+    },
+    {
+      class: 'String',
+      name: 'nestedObjectSeperator',
```
kgrgreer: What is this needed for?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+      value: '__'
+    },
+    {
+      class: 'StringArray',
+      name: 'nestedObjectNames'
```
kgrgreer: Why is this needed?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+    function outputHeader(o) {
+      if ( ! this.outputHeaderRow ) return;
+
+      this.outputPropertyFilteredName(o);
+      this.removeTrailingComma().out(this.nlStr);
+    },
+
+    function removeTrailingComma() {
+      // Removes trailing comma
+      if ( this.buf_.charAt(this.buf_.length - 1) == ',' ) this.buf_ = this.buf_.slice(0, -1);
```
kgrgreer: This is very expensive. Instead, just output a , before each column except the first.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+          // Gets and recurses through object properties
+          var ps = o.cls_.getAxiomsByClass(foam.core.Property);
+
+          for ( var i = 0 ; i < ps.length ; i++ ) {
+            this.outputPropertyName(o, ps[i]);
+          }
+        },
+        Array: function(o) { /* Ignore arrays in CSV */ },
+        Function: function(o) { /* Ignore functions in CSV */ },
+        Object: function(o) {
```
kgrgreer: I don't think CSV can support generic Objects since you don't know how many keys they're going to have so then you don't know which column means what.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+    },
+
+    function outputHeaderTitle(o) {
+      this.out(this.includeQuotes ? '"' : '').writePrefix().out(o, this.includeQuotes ? '"' : '', this.delimiter);
+    },
+
+    {
+      name: 'outputPropertyFilteredName',
+      code: foam.mmethod({
+        // Is there a `switch` like fall through so the function call isn't unecessarily repeated?
```
kgrgreer: function mmethod(map, opt_defaultMethod) {

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+      })
+    },
+
+    function outputPropertyName(o, p) {
+      if ( ! this.externalProperty(o, p) ) return;
+
+      // Checks if property is enum, or object with properties of its own
+      // TODO: Is this correct way to check if object has relevant sub-properties
+      if ( foam.core.AbstractEnum.isInstance( o[p.name] ) ) {
+        this.outputPropertyFilteredName(p.name + '__ordinal');
```
kgrgreer: Why is __ordinal required?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+      this.removeTrailingComma().out(this.nlStr);
+      ret += this.buf_;
+      this.reset();
+
+      return ret;
+    },
+
+    function externalProperty(o, p) {
+      if ( ! this.propertyPredicate(o, p) ) return false;
+      if ( p.isDefaultValue(o[p.name]) ) return false;
```
kgrgreer: I think we need to output default values. We don't need to do that for JSON because there we're usually loading the data into another FOAM system which already knows the default values. For CSV though, FOAM is very unlikely to be the recipient. Most like it will be a spreadsheet, which doesn't know the default values.

src/foam/core/CSV.js
TanayParikh: The configurability or the functionality itself? What should the default be?

src/foam/core/CSV.js
TanayParikh: Its a stack of object names. Helps print the header line for nested objects.

src/foam/core/CSV.js
TanayParikh: Perfect, I'll remove it. I was unsure on how to test that as well.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,359 @@
...
+  name: 'foam.csv',
+
+  constants: {
+    Compact: foam.csv.Outputter.create({
+      includeQuotes: true
+    }),
+  },
+
+  methods: [
+    function stringify(o) {
```
kgrgreer: I think toCSV() / fromCSV() would be easier to understand.

TanayParikh: Updated, thanks.

src/foam/core/CSV.js
TanayParikh: Removed.

src/foam/core/CSV.js
TanayParikh: It helps permit nesting logic for the reverse `fromCSV` functionality.

src/foam/core/CSV.js
TanayParikh: Updated to output default properties.

src/foam/core/CSV.js
TanayParikh: Just tried that, functions inside the map (ex. FObject) seem to be losing scope. Tried binding `this` but still throwing an error. Is there another way to pass in `this` into an `mmethod`?

src/foam/core/CSV.js
kgrgreer: How?

src/foam/core/CSV.js
kgrgreer: mmethod should preserve 'this' scope

src/foam/core/CSV.js
kgrgreer: The functionality.

src/foam/core/CSV.js
kgrgreer: But why do you need to store it? This will cause conflicts if you reuse the CSV outputter to output more than one class. So if someone used foam.csv.Compact, which is persistent, then the headers would be corrupted.

src/foam/core/CSV.js
TanayParikh: Thanks, updated.

For reference `function (o) { this.xyzMethod(o); }` will need to be used. Arrow functions do not work in this case due to scoping issues, (thanks @kgrgreer).

src/foam/core/CSV.js
TanayParikh: Removed.

src/foam/core/CSV.js
TanayParikh: Default to include quotes? Or nothing at all?

kgrgreer: I would say nothing at all, but test with Google sheets and/or Excel if you have it to see how it handles importing the two styles.

src/foam/core/CSV.js
TanayParikh: Refactored and removed. 	64bce2b

src/foam/core/CSV.js
TanayParikh: Google Sheets supports both. Should FOAM do the same?

src/foam/core/CSV.js
TanayParikh: Updating.

Is there a way to pass in a param into a `mmethod` (I'll need to pass in the `nestedObjectNames` stack so that it is preserved through recursive calls.

src/foam/core/CSV.js
kgrgreer: Just support without quotes. We can add if we ever need it.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,279 @@
...
+      return model;
+    }
+  ]
+});
+
+foam.LIB({
+  name: 'foam.csv',
+
+  constants: {
+    Compact: foam.csv.Outputter.create({
```
kgrgreer: s/Compact/Standard/

TanayParikh: Updated.

src/foam/core/CSV.js
TanayParikh: Removed quoted field/value support.

TanayParikh:
@kgrgreer All requested changes have been made, thanks for your help.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+
+      this.outputPropertyFilteredName(o, '');
+      this.out(this.nlStr);
+    },
+
+    function toCSV(o) {
+      // Outputs object headers
+      this.outputHeader(o);
+      var ret = this.buf_;
+      this.reset(); // reset to avoid retaining garbage
```
kgrgreer: Calling reset() here loses the contents of the buffer, forcing you to concatenate below, which doubles the amount of garbage that you generate. Wouldn't just reset()-ing at the top of the method be simpler and more efficient?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+    function outputHeaderTitle(o, prefix) {
+      this.out(this.buf_.length ? this.delimiter : '')
+          .out(prefix, o);
+    },
+
+    function outputPropertyName(o, p, prefix) {
+      if ( ! this.propertyPredicate(o, p) ) return;
+
+      // Checks if property is enum, or object with properties of its own
+      if ( p.of && ( ! foam.core.AbstractEnum.isInstance( o[p.name] ) ) ) {
```
kgrgreer: p.of will also pickup DAOProperties
Why do you care if it's an Enum? Enum's still only have one value.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+        // Appends object name to prefix for CSV Header
+        prefix += p.name + this.nestedObjectSeperator;
+        this.outputPropertyFilteredName(o[p.name], prefix);
+      } else {
+        this.outputPropertyFilteredName(p.name, prefix);
+      }
+    },
+
+    {
+      name: 'outputPropertyFilteredName',
```
kgrgreer: Can you come up with a better name for this method?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+      this.output(o[p.name]);
+    },
+
+    function reset() {
+      this.buf_ = '';
+      return this;
+    },
+
+    function outputPrimitive(val) {
+      this.out(this.buf_.length ? this.delimiter : '', val);
```
kgrgreer: I think it would be easier to handle the delimiter where you're doing the looping.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+        Null:      function(o) { this.outputPrimitive(null); },
+        String:    function(o) { this.outputPrimitive(o); },
+        FObject:   function(o) {
+          if ( o.outputCSV ) {
+            o.outputCSV(this)
+            return;
+          }
+          
+          // Outputs only the ordinal of ENUM
+          if ( foam.core.AbstractEnum.isInstance(o) ) {
```
kgrgreer: Why are enums treated differently?

kgrgreer: You can put AbstractEnum as one of your mmethod cases.

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+        Function: function(o) { /* Ignore functions in CSV */ },
+        Object: function(o) { /* Ignore generic objects in CSV */ }
+      }, function(o) { this.outputPrimitive(o); })
+    },
+
+    function fromCSV(className, s) {
+      if ( ! s ) throw 'Invalid CSV input to convert. Arguments must be (className, csvString).'
+      var lines = s.split('\n');
+
+      if ( lines.length == 0 ) throw 'Insufficient CSV Input'
```
kgrgreer: Missing ;

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+
+      // Trims quotes and splits CSV row into array
+      var props = lines[0].split(',');
+
+      // Account for array of values (TODO?)
+      var values = lines[1].split(',');
```
kgrgreer: What about strings with ,'s in them?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+  constants: {
+    Standard: foam.csv.Outputter.create(),
+  },
+
+  methods: [
+    function toCSV(o) {
+      return foam.csv.Standard.toCSV(o);
+    },
+
+    function fromCSV(className, s) {
```
kgrgreer: Why pass the className rather than just the class/model itself?

src/foam/core/CSV.js
kgrgreer: Shouldn't it be valid to have no lines?

kgrgreer: Where/how do you process the lines after the first line?

src/foam/core/CSV.js
```diff
@@ -0,0 +1,233 @@
...
+                                          values.slice(i, j), prop.of.id));
+              
+              i = j - 1;
+              break;
+            }
+          }
+        // Adds regular prop
+        } else {
+          var prop = model.cls_.getAxiomByName(p);
+          prop.set(model, prop.of ? foam.lookup(prop.of.id).create({ ordinal: v }) : v);
```
kgrgreer: Isn't foam.lookup(prop.of.id) the same as just prop.of?

src/foam/core/CSV.js
TanayParikh: There isn't really a way to handle that unless we use a special delimiter (~), or we surround the fields with a symbol ("), as far as I know. What would you recommend?

TanayParikh: What would the output be in that case? Empty FObject?

TanayParikh: I process only the first (header) line and the second (value) line. I was not sure if we were to add support for multiple value lines, so I left it as a potential TODO on line 170. I can add it in if required.

src/foam/core/CSV.js
TanayParikh: Fixed.

src/foam/core/CSV.js
kgrgreer: I don't think there should be any return value. You need to return more than one value, one per row, so I would just have this method return void and pass in a Sink as a third argument for it to put() objects to.

kgrgreer: Yes, that's absolutely required. CSV files usually have many lines.

src/foam/core/CSV.js
TanayParikh: Added `first` flag, as done in `JSON.js`.

src/foam/core/CSV.js
TanayParikh: Fixed.

src/foam/core/CSV.js
TanayParikh: Added to mmethod case.

src/foam/core/CSV.js
TanayParikh: Updated, thanks.

src/foam/core/CSV.js
kgrgreer: Google Sheets wraps the whole value in ""'s.

TanayParikh: Implemented Sink.

TanayParikh: Multiple lines are processed, and put into Sink as recommended.

src/foam/core/CSV.js
TanayParikh: `p.of` Was used to check if the property was an 'inner object' (with properties of its own). Is there a better/more conventional way to achieve this?

I'm handling Enum's slightly differently to ensure only the ordinal is printed. Otherwise, if treated like a regular object, the ordinal, label and name are all printed. (`JSON.js` prints only the ordinal).

src/foam/core/CSV.js
```diff
@@ -166,11 +166,14 @@ foam.CLASS({
       // Trims quotes and splits CSV row into array
       var props = lines[0].split(',');
 
-      // Account for array of values (TODO?)
-      var values = lines[1].split(',');
+      for (var i = 1; i < lines.length; ++i) {
```
kgrgreer: spaces before ;'s

TanayParikh: Fixed, thanks.

src/foam/core/CSV.js
TanayParikh: Renamed to `outputPropertyName`.

TanayParikh:
Please note, the sink is currently being converted to an array, and then returned. This functionality will be removed on the final build once testing is completed. 

**Edit:** Removed 34aa8c2. Kept return of sink (could be useful in certain cases).

src/foam/core/CSV.js
TanayParikh: Added custom CFG parser.

FObject -> CSV
- Fields with commas are surrounded in quotes. Quotes within the string itself are escaped with a second quote char. (Google Sheets approach)

src/foam/core/CSV.js
TanayParikh: I tried passing in the model (via `foam.lookup('fullQualifier')`), however, it gives the following error: `customModel.create is not a function`

When I pass in just the className, `foam.lookup('fullQualifier')` is exactly how I convert it into a model. I’m not sure what’s causing the error in that case.

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+    function outputHeaderTitle(o, prefix, first) {
+      this.out(first ? '' : this.delimiter)
+          .out(this.escapeString(prefix + o));
+    },
+
+    function outputPropertyName_(o, p, prefix, first) {
+      if ( ! this.propertyPredicate(o, p) ) return;
+
+      // Checks if property is enum, or object with properties of its own
+      if ( p.of && ( ! foam.core.AbstractEnum.isInstance( o[p.name] ) ) ) {
```
kgrgreer: The above comment and the code don't agree.

kgrgreer: Better to just check if the property is an FObjectProperty than looking for 'of'.

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+          }
+        },
+        Array: function(o) { /* Ignore arrays in CSV */ },
+        Function: function(o) { /* Ignore functions in CSV */ },
+        Object: function(o) { /* Ignore generic objects in CSV */ }
+      }, function(o, prefix, first) { this.outputHeaderTitle(o, prefix, first); })
+    },
+
+    function outputProperty(o, p, first) {
+      if ( ! this.propertyPredicate(o, p) ) return;
```
kgrgreer: A little more straight-forward, and less code, to just do:
if ( this.propertyPredicate(o, p) ) this.output(o[p.name], first);

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+      }
+      
+      return s;
+    },
+
+    {
+      name: 'output',
+      code: foam.mmethod({
+        Undefined:    function(o, first) { this.outputPrimitive(this.undefinedStr, first); },
+        Null:         function(o, first) { this.outputPrimitive(null, first); },
```
kgrgreer: Is this line redundant?

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+        sink.put(this.createModel(props, values, className));
+      }
+
+      return sink;
+    },
+
+    function splitIntoValues(csvString) {
+      if ( ( csvString == undefined ) || ( csvString.length == 0) ) return [];
+
+      var parser = foam.lookup('foam.lib.csv.CSVParser').create();
```
kgrgreer: Why do you have to look this up?

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+            this.outputProperty(o, ps[i], (i == 0 && first));
+          }
+        },
+        Array:        function(o) { /* Ignore arrays in CSV */ },
+        Function:     function(o) { /* Ignore functions in CSV */ },
+        Object:       function(o) { /* Ignore generic objects in CSV */ }
+      }, function(o, first) { this.outputPrimitive(o, first); })
+    },
+
+    function fromCSV(className, s, sink) {
```
kgrgreer: Why passing className rather than just the class?

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+    },
+
+    function splitIntoValues(csvString) {
+      if ( ( csvString == undefined ) || ( csvString.length == 0) ) return [];
+
+      var parser = foam.lookup('foam.lib.csv.CSVParser').create();
+      return parser.parseString(csvString).map(field => field.value == undefined ? '' : field.value);
+    },
+
+    function createModel(props, values, className) {
```
kgrgreer: Why passing className instead of just class?

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+      foam.assert(props.length == values.length,
+        'Invalid CSV Input, header and value rows must be the same number of cells');
+
+      var model = foam.lookup(className).create();
+
+      for ( var i = 0 ; i < props.length ; ++i ) {
```
kgrgreer: i++

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+        var p = props[i];
+        var v = values[i];
+
+        // Adds nested prop
+        if ( p.includes('__') ) {
```
kgrgreer: Shouldn't this use nestedObjectSeperator instead?

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+          p = p.split('__');
```
kgrgreer: Shouldn't this use nestedObjectSeperator instead?

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+          foam.assert(p.length >= 2, 'Invalid CSV object nesting, properties of inner objects are identified by `__`');
+          var prefix = p[0] + '__';
```
kgrgreer: Shouldn't this use nestedObjectSeperator instead?

src/foam/lib/csv/CSV.js
TanayParikh: I tried passing in the model (via `foam.lookup('fullQualifier')`), however, it gives the following error: `customModel.create is not a function`

When I pass in just the className, `foam.lookup('fullQualifier')` is exactly how I convert it into a model. I’m not sure what’s causing the error in that case.

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,256 @@
...
+  constants: {
+    Standard: foam.lib.csv.Outputter.create(),
+  },
+
+  methods: [
+    function toCSV(o) {
+      return foam.lib.csv.Standard.toCSV(o);
+    },
+
+    function fromCSV(className, csvString, sink) {
```
kgrgreer: Why pass className instead of just class?

src/foam/lib/csv/CSV.js
TanayParikh: Yes, updated.

src/foam/lib/csv/CSV.js
TanayParikh: Yes, updated.

src/foam/lib/csv/CSV.js
TanayParikh: Yes, updated.

src/foam/lib/csv/CSV.js
TanayParikh: Updated.

src/foam/lib/csv/CSV.js
TanayParikh: Was redundant, removed.

src/foam/lib/csv/CSV.js
TanayParikh: Removed.

src/foam/lib/csv/CSV.js
TanayParikh: Thanks, updated.

src/foam/lib/csv/CSV.js
TanayParikh: Reworded to better reflect the code.

src/foam/lib/csv/CSV.js
kgrgreer: When you pass in the class (not the model, which you actually rarely use in FOAM2) you should be able to call create on it, so let's switch to that and then diagnose your "is not a function" error. Also, why do you need to lookup? Just declare that you require: the class then refer to it as this.Class. This has the advantage that it lets the dynamic code-loading support know what classes it should load when load the calling class.

TanayParikh: Thanks, that really helped! I've updated accordingly.

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,257 @@
...
+    },
+
+    function outputPropertyName_(o, p, prefix, first) {
+      if ( ! this.propertyPredicate(o, p) ) return;
+
+      // Checks if property is an inner object (and not an enum)
+      if ( p.of && ( ! foam.core.AbstractEnum.isInstance( o[p.name] ) ) ) {
+        // Gets new empty object if FObjectProperty is currently undefined
+        // Done to permit appropriate headers for multi-line CSVs (multiple objects to convert)
+        if ( o[p.name] == undefined ) o[p.name] = foam.lookup(p.of.id);
```
kgrgreer: foam.lookup(p.of.id) is the same as just p.of, isn't it?

src/foam/lib/csv/CSV.js
```diff
@@ -0,0 +1,257 @@
...
+    function outputHeaderTitle(o, prefix, first) {
+      this.out(first ? '' : this.delimiter)
+          .out(this.escapeString(prefix + o));
+    },
+
+    function outputPropertyName_(o, p, prefix, first) {
+      if ( ! this.propertyPredicate(o, p) ) return;
+
+      // Checks if property is an inner object (and not an enum)
+      if ( p.of && ( ! foam.core.AbstractEnum.isInstance( o[p.name] ) ) ) {
```
kgrgreer: Can't you just check if p is an FObjectProperty?

PR #481, opened 2017-06-23
Rewrote Boot.java to load services from a JDAO

kirkeaton:


src/foam/nanos/boot/Boot.java
```diff
@@ -23,11 +21,15 @@
...
-    MapDAO serviceDAO = new MapDAO();
-    serviceDAO.setOf(NSpec.getOwnClassInfo());
-    serviceDAO.setX(root_);
-    serviceDAO_ = serviceDAO;
+    try {
+      // Used for all the services that will be required when Booting
+      DAO nSpecDAO = new MapDAO();
+      ((MapDAO) nSpecDAO).setOf(NSpec.getOwnClassInfo());
+      ((MapDAO) nSpecDAO).setX(root_);
+      serviceDAO_ = new JournaledDAO(nSpecDAO, "services");
```
kgrgreer, line 29: I should be able to just replace all of this with:
DAO nSpecDAO = new JDAO(NSpec, "services");

kirkeaton: I haven't made the changes to JDAO to accept ClassInfo as a parameter yet but I'll do that now.

PR #485, opened 2017-06-26
Sum java

kirkeaton:


src/foam/mlang/mlang.js
```diff
@@ -1724,6 +1724,7 @@ foam.CLASS({
       name: 'arg1'
     },
     {
+      class: 'Long',
```
kgrgreer: This isn't going to work with Double or Float properties. Better to make it a double.

kirkeaton: Good point. Thanks. Made the change

PR #486, opened 2017-06-26
Rework box pipeline into simpler intended-for-client-consumption API

mdittmer:


src/foam/box/pipeline/PipelineNode.js
```diff
@@ -59,7 +59,7 @@ foam.CLASS({
       name: 'errorBox',
       documentation: 'Error box for RPC-related errors external to runnable.',
       factory: function() {
-        return this.LogBox.create({
+        return this.defaultErrorBox || this.LogBox.create({
```
m-cheung, line 62: Just a heads up: this triggers a warning (`Access missing import: defaultErrorBox`) when `defaultErrorBox` is not provided. Is there a way to prevent the warning from showing up?

src/foam/box/pipeline/PipelineManager.js
```diff
@@ -0,0 +1,258 @@
...
+
+  properties: [
+    {
+      class: 'FObjectProperty',
+      of: 'foam.box.Box',
+      name: 'defaultOutputBox',
+      documentation: `Output box used for end-of-computation (when no
+          next-to-run delegates bound to this step in the pipeline).`,
+      factory: function() {
+        return this.ctxDefaultOutputBox || this.LogBox.create();
```
m-cheung, line 110: A similar warning (`Access missing import: ctxDefaultOutputBox`) will be displayed here as well when `defaultOutputBox` is not provided

mdittmer:
Let's ignore the warning. I believe it is dev-mode-only. @kgrgreer is there
a warning-free pattern for this?

PR #488, opened 2017-06-27
Fix Number.compare() and update QueryParser to manually adapt numbers…

mdittmer:
…, rather than expecting foam.Number.compare() to do it.

@m-cheung PTAL compare() change.

@jacksonic PTAL test change.

@kgrgreer PTAL QueryParser change.


src/foam/parse/QueryParser.js
```diff
@@ -500,6 +505,22 @@ foam.CLASS({
     function parseString(str, opt_name) {
       var query = this.grammar_.parseString(str, opt_name);
       return query && query.partialEval ? query.partialEval() : query;
+    },
+    // Adapt binary predicate values.
```
kgrgreer: Add more docs to explain what this does and why.

src/foam/core/stdlib.js
```diff
@@ -372,8 +372,7 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return foam.Number.compare(a, b) == 0; },
     function compare(a, b) {
-      if ( ! foam.Number.isInstance(b) && isNaN(parseFloat(b))
-          || ( isNaN(a) && ! isNaN(b)) ) return 1;
+      if ( ! foam.Number.isInstance(b) || ( isNaN(a) && ! isNaN(b)) ) return 1;
       else if ( ! isNaN(a) && isNaN(b) ) return -1;
```
kgrgreer: No need for 'else', since you've already returned in the 'if'.

test/src/lib/QueryParser.js
```diff
@@ -89,6 +89,7 @@ describe('Query parser', function() {
       var seen = {};
       for ( var i = 0; i < actual.array.length; i++ ) {
         seen[actual.array[i].id] = true;
+        if ( ! ( expected.indexOf(actual.array[i].id) >-1 ) ) debugger;
```
m-cheung: This looks like a statement that was included for debugging purposes. Can it be removed?

mdittmer:
Added documentation and removed code-for-debugging. @kgrgreer anything else?

m-cheung (approved):
LGTM (with regards to comparisons / tests)

src/foam/parse/QueryParser.js
```diff
@@ -506,21 +506,36 @@ foam.CLASS({
...
-      if ( ! ( opts.arg1 && opts.arg1.adapt && opts.hasOwnProperty('arg2') ) )
-        return opts;
-
-      if ( Array.isArray(opts.arg2) ) {
-        var arg2 = opts.arg2;
-        for ( var i = 0; i < arg2.length; i++ ) {
-          arg2[i] = opts.arg1.adapt(undefined, opts.arg2[i]);
+    {
+      name: 'bin_',
+      documentation: `For binary ops like EQ(prop, value), the QueryParser may
```
kgrgreer: We build the parser dynamically by walking the Property axioms of the Model. At that point we know the type and specific parser for the property, so we could just build a type-sound parser.

mdittmer:
@m-cheung and @kgrgreer, some more deep changes in `compare()` now included. PTAL.

src/foam/core/FObject.js
```diff
@@ -824,7 +824,7 @@ foam.CLASS({
 
       // Use comparator of other if it is not an FObject.
       if ( ! foam.core.FObject.isInstance(other) ) {
```
kgrgreer: Is this still needed?

src/foam/core/stdlib.js
```diff
@@ -586,10 +586,8 @@ foam.LIB({
     function getTime(d) { return ! d ? 0 : d.getTime ? d.getTime() : d ; },
     function equals(a, b) { return this.getTime(a) === this.getTime(b); },
     function compare(a, b) {
-      if ( ! foam.Date.isInstance(b) && ! foam.Number.isInstance(b) ) return 1;
-      a = this.getTime(a);
-      b = foam.Number.isInstance(b) ? b : this.getTime(b);
-      return foam.Number.compare(a, b);
+      if ( ! foam.Date.isInstance(b) ) return 1;
```
kgrgreer, line 589: Couldn't this line be replaced with foam.util.compare(a,b)?

mdittmer: That would blow up the stack if this predicate is true and b is of lower precedence than Date, I think.

src/foam/core/stdlib.js
```diff
@@ -795,8 +793,9 @@ foam.LIB({
         // we will always use the comparator of higher precedence.
         var aType = typeOf(a);
         var bType = typeOf(b);
-        return (aType.ordinal >= bType.ordinal) ? aType.compare(a, b) :
-            this.comparatorNegate(bType.compare(b, a));
+        return aType.ordinal > bType.ordinal ? 1 : aType.ordinal < bType.ordinal ? -1 : aType.compare(a, b);
```
kgrgreer: Can't we get rid of this and only compare by ordinal if types are different?

src/foam/core/FObject.js
mdittmer: The reason I kept it is because `myFObject.compareTo(nonFObject)` may be invoked directly  (without `foam.util.compare()` on the stack). Is that case considered misusing the API? If so, I think it's pretty subtle. If not, then this is still needed.

src/foam/core/stdlib.js
mdittmer: Isn't that what this does?

src/foam/core/stdlib.js
```diff
@@ -796,8 +793,9 @@ foam.LIB({
...
```
m-cheung: I believe some of the other comparators have a check that is performed (e.g. `if ( ! foam.<type>.isInstance(b) ) return 1`)
The only case that I can think of that depends on those checks now is `return -foam.typeOf(other).compare(other, this);` (the change you made in FObject.js).

I believe those can be removed if FObject utilize `foam.util.compare` instead of the comparators directly

src/foam/core/stdlib.js
```diff
@@ -587,10 +586,8 @@ foam.LIB({
     function getTime(d) { return ! d ? 0 : d.getTime ? d.getTime() : d ; },
     function equals(a, b) { return this.getTime(a) === this.getTime(b); },
     function compare(a, b) {
-      if ( ! foam.Date.isInstance(b) && ! foam.Number.isInstance(b) ) return 1;
-      a = this.getTime(a);
-      b = foam.Number.isInstance(b) ? b : this.getTime(b);
-      return foam.Number.compare(a, b);
+      if ( ! foam.Date.isInstance(b) ) return 1;
```
m-cheung, line 589: Double checking to make sure this is intentional. This change makes the Date comparator a strict comparison. Previously, you could perform an operation such as
```
var x = new Date();
foam.Date.compare(x, x.getTime()); // Returning a 0, implying equivalence.
```
This was also the reason for the weird ordering of the types originally.

src/foam/core/stdlib.js
```diff
@@ -372,9 +372,8 @@ foam.LIB({
     function clone(o) { return o; },
     function equals(a, b) { return foam.Number.compare(a, b) == 0; },
     function compare(a, b) {
-      if ( ! foam.Number.isInstance(b) && isNaN(parseFloat(b))
-          || ( isNaN(a) && ! isNaN(b)) ) return 1;
-      else if ( ! isNaN(a) && isNaN(b) ) return -1;
+      if ( ! foam.Number.isInstance(b) || ( isNaN(a) && ! isNaN(b)) ) return 1;
+      if ( ! isNaN(a) && isNaN(b) ) return -1;
```
m-cheung, line 376: Are the guards in the comparators still required at this point? Do we expect any users to directly call type compare with different types?

kgrgreer:
Looks much better. Thanks

PR #490, opened 2017-06-27
Sequence DAO fixes

kirkeaton:
1. Made value be a long
2. Add two constructors:
 - One that takes in just the delegate
 - One that takes in both the delegate and  the property name.
3. Simplified calcDelegateMax_ function
4. Added `put` function which just forwards the call to `put_`

### DIFF
--- a/src/foam/dao/SequenceNumberDAO.java
+++ b/src/foam/dao/SequenceNumberDAO.java
@@ -10,8 +10,10 @@ import foam.mlang.sink.Max;
 public class SequenceNumberDAO
     extends ProxyDAO
 {
-  protected String property = "id";
   protected long value_ = 1;
+  protected boolean isValueSet_ = false;
+
+  protected String property = "id";
   private PropertyInfo property_ = null;
 
   public SequenceNumberDAO(DAO delegate) {
@@ -21,6 +23,7 @@ public class SequenceNumberDAO
   public SequenceNumberDAO(PropertyInfo property, DAO delegate) {
     setDelegate(delegate);
     this.property_ = property;
+    this.property = property.getName();
   }
 
   public AbstractDAO setOf(ClassInfo of) {
@@ -37,6 +40,7 @@ public class SequenceNumberDAO
 
   public SequenceNumberDAO setValue(long value) {
     value_ = value;
+    isValueSet_ = true;
     return this;
   }
 
@@ -58,11 +62,12 @@ public class SequenceNumberDAO
   }
 
   public FObject put(FObject obj) {
+    if ( ! isValueSet_ )
+      calcDelegateMax_();
     return put_(getX(), obj);
   }
 
   public FObject put_(X x, FObject obj) {
-    calcDelegateMax_();
     Number val = (Number) obj.getProperty(property);
     if ( val.longValue() < 1 ) {
       getProperty_().set(obj, value_++);
--- a/src/foam/dao/SequenceNumberDAO.java
+++ b/src/foam/dao/SequenceNumberDAO.java
@@ -14,6 +14,15 @@ public class SequenceNumberDAO
   protected long value = 1;
   private PropertyInfo property_ = null;
 
+  public SequenceNumberDAO(DAO delegate) {
+    setDelegate(delegate);
+  }
+
+  public SequenceNumberDAO(DAO delegate, String property) {
+    setDelegate(delegate);
+    this.property = property;
+  }
+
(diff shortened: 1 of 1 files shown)


src/foam/dao/SequenceNumberDAO.java
```diff
@@ -4,15 +4,25 @@
...
   protected String property = "id";
-  protected int value = 1;
+  protected long value = 1;
   private PropertyInfo property_ = null;
 
+  public SequenceNumberDAO(DAO delegate) {
+    setDelegate(delegate);
+  }
+
+  public SequenceNumberDAO(DAO delegate, String property) {
```
kgrgreer: Reverse the order of the arguments so that if you have a deeply decorated DAO the property is closer to the SequenceNumberDAO constructor and the code is easier to read.

src/foam/dao/SequenceNumberDAO.java
```diff
@@ -42,21 +52,23 @@ private PropertyInfo getProperty_() {
...
+  public FObject put(FObject obj) {
+    return put_(getX(), obj);
   }
 
   public FObject put_(X x, FObject obj) {
     calcDelegateMax_();
-    int val = (int) getProperty_().f(obj);
-    if (val < 1) {
+    Number val = (Number) obj.getProperty(property);
+    if (val.longValue() < 1) {
```
kgrgreer: spacing

src/foam/dao/SequenceNumberDAO.java
```diff
@@ -42,21 +52,23 @@ private PropertyInfo getProperty_() {
...
       getProperty_().set(obj, this.value++);
-    } else if (val >= this.value ) {
-      this.value = (int) val + 1;
+    } else if (val.longValue() >= this.value ) {
```
kgrgreer: spacing

src/foam/dao/SequenceNumberDAO.java
```diff
@@ -4,15 +4,25 @@
...
+import foam.mlang.MLang;
 import foam.mlang.sink.Max;
 
 public class SequenceNumberDAO
-  extends ProxyDAO
+    extends ProxyDAO
 {
   protected String property = "id";
-  protected int value = 1;
+  protected long value = 1;
```
kgrgreer: rename to value_

src/foam/dao/SequenceNumberDAO.java
```diff
@@ -4,15 +4,25 @@
...
 import foam.core.X;
 import foam.core.PropertyInfo;
+import foam.mlang.MLang;
 import foam.mlang.sink.Max;
 
 public class SequenceNumberDAO
-  extends ProxyDAO
+    extends ProxyDAO
 {
   protected String property = "id";
```
kgrgreer: Rename to property_.

src/foam/dao/SequenceNumberDAO.java
kgrgreer: Less error prone if you let the user specify using a PropertyInfo rather than a String. Then, if the property is removed they'll get a compile-time error.

src/foam/dao/SequenceNumberDAO.java
```diff
@@ -42,21 +52,23 @@ private PropertyInfo getProperty_() {
...
+    getDelegate().select(sink);
+    this.setValue((long) (((Max) sink).getValue() + 1.0));
+  }
+
+  public FObject put(FObject obj) {
+    return put_(getX(), obj);
   }
 
   public FObject put_(X x, FObject obj) {
     calcDelegateMax_();
```
kgrgreer: This is going to be very slow.  You should only calcDelegateMax on the first put(), not on every put.

kirkeaton: Okay done.

src/foam/dao/SequenceNumberDAO.java
```diff
@@ -25,8 +38,9 @@ public SequenceNumberDAO setProperty(String property) {
     return this;
   }
 
-  public SequenceNumberDAO setValue(int value) {
-    this.value = value;
+  public SequenceNumberDAO setValue(long value) {
+    value_ = value;
+    isValueSet_ = true;
```
kgrgreer: This should be set in calcDelegateMax_(), not in the constructor.

kirkeaton: Done.

PR #493, opened 2017-06-28
Web auth

MarcRoopchand:
Web Auth Service 
Implemented  the following functions to accept a userId instead of X
- login
- check 
- updatePassword
- logout


src/foam/nanos/auth/WebAuthService.java
```diff
@@ -0,0 +1,62 @@
+package foam.nanos.auth;
```
kgrgreer, line 1: Don't forget copyright header.

PR #495, opened 2017-06-28
Composite Logger

Akenne:


src/foam/nanos/logger/ILogger.java
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.logger;
+
+public interface ILogger {
+  public void start();
```
kgrgreer, line 10: Why is start() needed?

src/foam/nanos/logger/ILogger.java
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.logger;
+
+public interface ILogger {
```
kgrgreer, line 9: Why ILogger instead of just Logger?

src/foam/nanos/logger/ILogger.java
```diff
@@ -0,0 +1,17 @@
...
+  public void start();
+  public String combine(Object[] args);
```
kgrgreer, line 11: Why does combine() need to be part of the interface?

src/foam/nanos/logger/NanoLogger.java
```diff
@@ -6,112 +6,58 @@
...
-    }
-
-    @Override
-    public StringBuilder get() {
-      StringBuilder b = super.get();
-      b.setLength(0);
-      return b;
-    }
-  };
+  private List<ILogger> childLoggers = new ArrayList<ILogger>();
```
kgrgreer, line 20: Use protected instead of private. Makes easier to debug and test.

kgrgreer, line 20: s/childLoggers/delegates_/

src/foam/nanos/logger/NanoLogger.java
```diff
@@ -6,112 +6,58 @@
...
-      str.append(n.toString());
-    }
-    return str.toString();
+  public void remove(ILogger logger) {
+    childLoggers.remove(logger);
   }
 
   public void log(Object... args) {
-    logger.info(combine(args));
+    for (ILogger logger : childLoggers) {
```
kgrgreer, line 35: Spacing: for ( Logger logger : delegates_ )

src/foam/nanos/logger/FileLogger.java
```diff
@@ -0,0 +1,115 @@
...
+      if ( prevTime / 1000 != System.currentTimeMillis() / 1000 ) {
+        prevTime = System.currentTimeMillis();
+        prevTimestamp = sdf.get().format(new Timestamp(prevTime));
+      }
+
+      str.append(prevTimestamp);
+      str.append(',');
+
+      // debug special case, fine level == 500
+      if (lev == 500) {
```
kgrgreer, line 74: formatting: if ( lev == 500 ) {

kgrgreer:
ping

src/foam/nanos/logger/ILogger.java
Akenne: nanologger implements nanoservice, which requires it has a start() function.

src/foam/nanos/logger/ILogger.java
Akenne: the name Logger is already used by java, so i made it ILogger to signify its an (I)nterface

I can change it to something else but not Logger

src/foam/nanos/logger/NanoLogger.java
Akenne: I don't quite understand what you mean by this

Akenne: Oh I see

src/foam/nanos/logger/ILogger.java
kgrgreer: It isn't required to be a NanoService now. Also, you don't declare that it extends NanoService, so having a start() method doesn't mean that it's a NanoService if you don't declare it also.

src/foam/nanos/logger/ILogger.java
kgrgreer: But then to be consistent we would need to prefix all of our other interfaces with "I" as well, which we don't do. Better to name it NanoLogger or just name it Logger, and then refer to the java Logger with the full pathname in our code where we need to use it.

src/foam/nanos/logger/ILogger.java
Akenne: oh sorry, start is not needed because of the nanoservice, it is needed because filelogger and stdoutlogger have start() functions, and this function needs to be called in nanologger.

like this (code snipped from nanologger.java)

```
public void add(ILogger logger) {

    logger.start();

    delegates_.add(logger);

  }
```

Akenne: @kgrgreer 

this is the only thing left and then i can recommit

kgrgreer: There are still about half a dozen unresolved comments.

Akenne: So it is okay to leave in the start() function here?

kgrgreer:
Already merged in different PR.

PR #499, opened 2017-06-29
formatting and put id on quotes

### LINKED ISSUE #496: JDAO Doesn't Output String Keys Properly
String keys are output like:

r({"id":key})

But key isn't in quotes so isn't valid JSON.

@drish

drish:
fixes journal dao parser file formatting
fixes #496


src/foam/dao/FileJournal.java
```diff
@@ -63,7 +63,7 @@ public void remove(FObject obj, Detachable sub) {
 
   public void remove(Object id, Detachable sub) {
     try {
-      bw.write("r({\"id\":" + id + "})");
+      bw.write("r({\"id\":\"" + id + "\"})");
```
kgrgreer: But if the id isn't a string, then this will be wrong.

JSON.stringify({id:'foo'});
"{"id":"foo"}"
JSON.stringify({id:42});
"{"id":42}"

drish: It should be formatted based on the `of` `id` type ?

something like:
```
if instanceof getOf().ID == string then put on quotes
else don't put on quotes  

```

or should all ids be treated as strings ?

kgrgreer:
The PropertyInfo will have a toJSON() method, but it's not done yet, so for now just check if 'id' is a String.

PR #500, opened 2017-06-29
HTTP and FileServlet Fixes

nick-prat:
Added getX() to boot. Fixed if/else chain in HttpHandler, File servlet tries to get URI information from getRequestURI first, then getPathInfo.


src/foam/nanos/http/FileServlet.java
```diff
@@ -47,9 +47,13 @@ private void fileNotFoundError(HttpServletResponse resp, String file) {
 
   @Override
   protected void doGet(HttpServletRequest req, HttpServletResponse resp) {
-    String pathInfo = req.getPathInfo();
+    String pathInfo;
+    try {
+      pathInfo = req.getRequestURI();
+    } catch (Exception e) {
```
kgrgreer: What exception are you getting here?

nick-prat: UnsupportedOperationException, getReqquestURI throws it when launched using Java's built in Http server, while getPathInfo fails when using tomcat. I can switch it to be that error

kgrgreer: Does this help: 
https://github.com/foam-framework/foam2/commit/d5b8a094d1e5d30548ecc5ee70634928b78ee83f

PR #501, opened 2017-06-29
Void property setters

kirkeaton:
1. Modified auto-generated Java code to return void for property setters.
2. Fixed up classes that were chaining setter calls to make them use separate setter function calls.


src/foam/dao/AbstractDAO.java
```diff
@@ -16,19 +16,27 @@
   protected PropertyInfo primaryKey_        = null;
 
   public DAO where(Predicate predicate) {
-    return new FilteredDAO().setPredicate(predicate).setDelegate(this);
+    FilteredDAO filteredDAO = new FilteredDAO().setPredicate(predicate);
```
kgrgreer: Shouldn't we switch setPredicate() also?

kgrgreer: FilteredDAO should have a two-arg constructor that takes predicate and delegate, since it only has two properties and isn't useful unless they're both set.

src/foam/dao/AbstractDAO.java
```diff
@@ -16,19 +16,27 @@
...
+    filteredDAO.setDelegate(this);
+    return filteredDAO;
   }
 
   public DAO orderBy(Comparator comparator) {
-    return new OrderedDAO().setOrder(comparator).setDelegate(this);
+    OrderedDAO orderedDAO = new OrderedDAO().setOrder(comparator);
```
kgrgreer: Should also have a two-arg constructor, just like SkipDAO and LimitedDAO.

src/foam/dao/history/HistoryDAO.java
```diff
@@ -52,10 +52,10 @@ private String formatUserName(User user) {
       String key = (String) i.next();
       PropertyInfo prop = (PropertyInfo) currentValue.getClassInfo().getAxiomByName(key);
 
-      PropertyUpdate update = new PropertyUpdate()
-          .setName(key)
-          .setOldValue(prop.f(currentValue))
-          .setNewValue(diff.get(key));
+      PropertyUpdate update = new PropertyUpdate();
```
kgrgreer: Better to have a three-arg constructor.

src/foam/mlang/MLang.java
```diff
@@ -19,42 +19,69 @@
 {
 
   public static Expr prepare(Object o) {
-    return o instanceof Expr ? (Expr) o : new Constant().setValue(o);
+    if ( o instanceof Expr )
```
kgrgreer: Better if Constant() has a 1-arg constructor.

src/foam/mlang/MLang.java
```diff
@@ -19,42 +19,69 @@
...
+      return (Expr) o;
+    Constant constant = new Constant();
+    constant.setValue(o);
+    return constant;
   }
 
   public static Predicate LT(Object o1, Object o2) {
-    return new Lt().setArg1(MLang.prepare(o1)).setArg2(MLang.prepare(o2));
+    Lt lt = new Lt();
```
kgrgreer: Better if binary predicates have two arg constructors.

src/foam/mlang/MLang.java
```diff
@@ -19,42 +19,69 @@
...
-    return new Gt().setArg1(MLang.prepare(o1)).setArg2(MLang.prepare(o2));
+    Gt gt = new Gt();
+    gt.setArg1(MLang.prepare(o1));
+    gt.setArg2(MLang.prepare(o2));
+    return gt;
   }
 
   public static Predicate AND(Predicate... args) {
-    return new And().setArgs(args);
+    And and = new And();
```
kgrgreer: Should have a one arg constructor.

src/foam/dao/AbstractDAO.java
kirkeaton: Done.

src/foam/dao/AbstractDAO.java
kirkeaton: Done.

PR #505, opened 2017-06-30
DatastoreDAO breaking storage change: Store multi-part IDs and adapt binary predicates

mdittmer:
MultiPartIDs are generally transient. However, storing these IDs as a property in Datastore allows IDs to be used as an order or predicate. In order to support MultiPartID in a predicate, additional changes to adapt predicate constants according to their corresponding property type (similar to Binary.adapt() implementation).

An alternative to this design would be to adapt incoming predicates to DatastoreDAO to convert `MLANG(Cls.ID, value)` to `AND(MLANG(Cls.FIRST_ID_PROP, valuePart1), ..., MLANG(Cls.LAST_ID_PROP, valuePartN))`. In fact, this might be the right thing to do universally, since `MultiPartID`'s storage transience should prevent it from being stored anywhere (hence mLangs that contain them need adaptation before being passed to a storage mechanism). If so, what's the best place to do this? Baking a special case into `predicate.Binary` for something so specific as `MultiPartID` feels wrong, but `partialEval() / f()` on `MultiPartID` is too low in the predicate tree to perform the transformation. @kgrgreer WDYT?

### DIFF
--- a/src/com/google/cloud/datastore/DatastoreDAO.js
+++ b/src/com/google/cloud/datastore/DatastoreDAO.js
@@ -313,7 +313,7 @@ foam.CLASS({
         deletes[i] = { delete: arr[i].getDatastoreKey() };
       }
 
-      this.sendRequest('commit', {
+      return this.sendRequest('commit', {
         mode: 'TRANSACTIONAL',
         mutations: deletes,
         transaction: transaction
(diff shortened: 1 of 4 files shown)


mdittmer:
I'm closing this, as the application it was intended for will take a different approach.

PR #507, opened 2017-07-04
Box readme

mdittmer:
@kgrgreer PTAL in general. @adamvy PTAL in general + there's a section I'd like you to write ^_^

### DIFF
--- a/src/foam/box/README.md
+++ b/src/foam/box/README.md
@@ -1,0 +1,110 @@
+# Boxes
+
+The `Box` interface is a message-passing interface. Its various
+implementations facilitate message passing over various boundaries such as
+FOAM contexts, operating system processes, and network layers.
+
+## Box and Message
+
+The box interface contains one method:
+
+```
+interface Box {
+  void send(message);
+}
+```
+
+The message interface contains a payload and optional routing information:
+
+```
+interface Message {
+  Object object;     // payload
+  Object attributes; // routing information
+}
+```
+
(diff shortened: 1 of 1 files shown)


src/foam/box/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+```
+interface BoxRegistry {
+  Box  register(opt_name, opt_clientServerService, box);
+  void unregister(nameOrBox);
+}
+```
+
+The `register` method is used to register the input `box` under the name
+`opt_name`. If `opt_name` is omitted, the registry selects an available
```
kgrgreer, line 57: Would 'random' be better than 'available' or 'random available'?

src/foam/box/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+```
+interface StubFactory {
+  Class get(Class);
+}
+```
+
+The `get()` method returns a stub class for the input class. Stub class
+instances have a `delegate` property that should be set to a box that accepts
+RPC messages over the unerlying class's interface.
```
kgrgreer: s/underlying/underlying

kgrgreer:
Added a couple comments, but LGTM.

src/foam/box/README.md
mdittmer: The default implementation is not random. It uses a counter. https://github.com/foam-framework/foam2/blob/master/src/lib/box.js#L235

PR #510, opened 2017-07-05
ForkBox: Add 'detached' parameter passed to child_process.spawn().

mdittmer:
Also, add some documentation and rename replyBox to replyBox_ to clarify that it is an internal property (not to be configured by consumer).


mdittmer:
TBR @m-cheung

PR #512, opened 2017-07-05
Desc Java Ordering

TanayParikh:
1. Added Desc class to classes.js
2. Removed orderTail, orderPrimaryProperty, and orderDirection functions
3. Fixed compare function to match the interface. It should return an int and take in two Objects o1 and o2 as parameters.


src/foam/mlang/mlangJava.js
```diff
@@ -399,24 +399,22 @@ foam.CLASS({
...
+        }
+      ],
       javaCode: 'return -1 * getArg1().compare(o1, o2);'
     },
     {
       name: 'toString',
       javaCode: 'return "DESC(" + getArg1().toString() + ")";'
-    },
-    {
-      name: 'orderTail',
```
eatonk: You should remove the following functions from the `foam.mlang.order.Comparator` interface as well. Otherwise you will have issues with this class not implementing all the functions of that interface.

src/foam/mlang/order/ComparatorJava.js
```diff
@@ -8,5 +8,22 @@
    package: 'foam.mlang.order',
    name: 'Comparator',
 
-   javaImplements: [ 'java.lang.Comparator' ]
+   javaExtends: [ 'java.lang.Comparable' ],
```
kirkeaton, line 11: I think this needs to be java.util.Comparator instead

TanayParikh: Thanks, created a PR with the modification https://github.com/foam-framework/foam2/pull/516

TanayParikh: Never mind, I see it is included in your PR https://github.com/foam-framework/foam2/pull/514

Closed #516

PR #513, opened 2017-07-05
Updated xmldao

pavneetskapoor:
- Created XMLDAO and XMLSupport for writing from FObject to XML and vice versa
- Modified foam.core.Date and foam.core.DateTime to incorporate various date formats in the future.
- For now, it is set to ISO8601. Property for these types is called dateFormat
- Future Improvements: Reading primitive typed arrays still needs to be supported internally


src/foam/java/refinements.js
```diff
@@ -55,6 +55,7 @@ foam.CLASS({
         propName: this.name,
         propType: this.javaType,
         propRequired: this.required,
+        of: this.of,
```
kgrgreer: Change to: this.of.id, since you just want to store the package and name of the 'of', not the JS Object itself, which doesn't make any sense to Java.

src/foam/java/PropertyInfo.js
```diff
@@ -106,6 +106,12 @@ foam.CLASS({
             visibility: 'public',
             type: 'boolean',
             body: 'return ' + Boolean(this.propRequired) + ';'
+          },
+          {
+            name: 'of',
+            visibility: 'public',
+            type: 'String',
```
kgrgreer: Shouldn't this be ClassInfo?

src/foam/java/refinements.js
kgrgreer: Also, most PropertyInfo's don't have an 'of' property, which is where you were getting the 'undefined' from. I think this should only be added to PropertyInfo sub-classes that use 'of'.

pavneetskapoor: So this should be added to separate abstractPropertyInfos such as `abstractFObjectPropertyInfo` ?

src/foam/java/PropertyInfo.js
pavneetskapoor: This wouldn't return the `of` field that would be filled by the specific properties if it were ClassInfo.

TanayParikh:
Confirmed.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -0,0 +1,94 @@
...
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import java.util.ArrayList;
+import java.util.List;
+
+
+public abstract class AbstractArrayPropertyInfo
+  extends AbstractPropertyInfo {
+
+  public int compareValues(boolean b1, boolean b2) {
```
kgrgreer: Why does this take boolean arguments?

src/foam/core/AbstractPropertyInfo.java
```diff
@@ -6,8 +6,14 @@
 
 package foam.core;
 
+import org.w3c.dom.Document;
```
kgrgreer: sort imports

src/foam/dao/XMLDAO.java
```diff
@@ -0,0 +1,88 @@
+package foam.dao;
+
+import org.w3c.dom.Document;
```
kgrgreer: sort imports

src/foam/dao/XMLDAO.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Iterator;
+
+public class XMLDAO
+  extends MapDAO
+{
+
+  protected String fileName;
+
+  public void setFileName(String filename) {
+    if ( filename.contains(".xml") ) {
```
kgrgreer: .endsWith() would be better than .contains()

src/foam/java/refinements.js
```diff
@@ -434,8 +433,11 @@ foam.CLASS({
       var info = this.SUPER(cls);
       var m = info.getMethod('cast');
       m.body = 'return ( o instanceof Number ) ?'
-            + '((Number)o).intValue() :'
-            + '(int)o;';
+        + ' ((Number)o).intValue() :'
```
kgrgreer: Would be nicer as a `-quoted multi-line string. Other similar code blocks below as well.

pavneetskapoor: Done

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+
+public abstract class AbstractArrayPropertyInfo
+  extends AbstractPropertyInfo {
+
+  public int compareValues(Object[] b1, Object[] b2) {
+    return (b1 == b2) ? 1 : 0;
```
kgrgreer: if ( b1 == b2 ) return 0;
if ( b2 == null ) return 1;
if ( b1 == null ) return -1;
int h1 = b1.hashCode();
int h2 = b2.hashCode();
return h1 == h2 ? 0 : h1 > h2 ? 1 : -1;

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -0,0 +1,93 @@
...
+    String startTag = reader.getLocalName();
+    try {
+      int eventType;
+      while ( reader.hasNext() ) {
+        eventType = reader.next();
+        switch ( eventType ) {
+          case XMLStreamConstants.START_ELEMENT:
+            // Nested object in array
+            if ( reader.getLocalName().equals("object") ) {
+              FObject clsInstance = XMLSupport.createObj(x, reader);
```
kgrgreer: I find the name 'clsInstance' confusing, since the name would make me suspect it were an instance of a class. Maybe a better/simpler name would just be 'o'. When would this value be null?

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -0,0 +1,93 @@
...
+    objElement.appendChild(prop);
+
+    // FObject Array check
+    if ( this.f(obj) instanceof FObject[] ) {
+      FObject[] nestedArray = (FObject[]) this.f(obj);
+      for ( int j = 0; j < nestedArray.length; j++ ) {
+        XMLSupport.toXML(nestedArray[j], doc, prop);
+      }
+      return;
+    } else {
```
kgrgreer: No need to wrap in an 'else' if you've already returned in the 'if'. Or, no need for the 'return'. Either way.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -0,0 +1,98 @@
...
+import java.lang.UnsupportedOperationException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.xml.stream.XMLStreamConstants;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+
+public abstract class AbstractArrayPropertyInfo
```
kgrgreer, line 19: This is used to handle FObjectArray and Array, which will be handled completely differently (simple values vs. objects). Since this doesn't work with primitive types anyway, I would just divide it into two different classes (or three, if you count Map). Also, how is Map being supported?

src/foam/core/AbstractEnumPropertyInfo.java
```diff
@@ -0,0 +1,58 @@
...
+  public Object fromXML(X x, XMLStreamReader reader) {
+    FObject obj = null;
+    NanoLogger logger = (NanoLogger) x.get("logger");
+    try {
+
+      while ( reader.hasNext() ) {
+        switch ( reader.getEventType() ) {
+          case XMLStreamConstants.START_ELEMENT:
+            // Enum Specific Case
+            if ( reader.getLocalName() == this.getName() ) {
```
kgrgreer, line 32: Add some comments to explain this.

src/foam/core/AbstractObjectPropertyInfo.java
```diff
@@ -6,11 +6,25 @@
 
 package foam.core;
 
+import javax.xml.stream.XMLStreamReader;
+import javax.xml.stream.XMLStreamWriter;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+
 public abstract class AbstractObjectPropertyInfo
```
kgrgreer, line 14: Why is this class needed? Just for compareValues()?

src/foam/core/AbstractPropertyInfo.java
```diff
@@ -47,4 +53,30 @@ public void diff(FObject o1, FObject o2, Map diff, PropertyInfo prop) {
...
   }
 
+  public void setFromString(Object obj, String value) {
+    // TODO: Need to write
+  }
+
+  @Override
+  public Object fromXML(X x, XMLStreamReader reader) {
+    // Moves reader to characters state in order for value reading for various data types (date, boolean, short ...)
+    NanoLogger logger = (NanoLogger) x.get("logger");
```
kgrgreer: Faster to not lookup until needed.

src/foam/dao/XMLDAO.java
```diff
@@ -0,0 +1,88 @@
...
+    saveToXML();
+    return s;
+  }
+
+  public void removeAll() {
+    super.removeAll();
+    saveToXML();
+  }
+
+  public void saveToXML () {
```
kgrgreer: Needs to be synchronized so that you don't write the file twice and the same time and corrupt it.

src/foam/core/AbstractObjectPropertyInfo.java
pavneetskapoor: This was a class that was existing before I worked on the XMLDAO so essentially I just added the methods it would override as part of a generic abstract object class

src/foam/core/AbstractArrayPropertyInfo.java
pavneetskapoor: Finished dividing into two different classes. Map is not currently supported at the moment. That would be a TODO in the near future.

src/foam/core/AbstractEnumPropertyInfo.java
pavneetskapoor: Done

PR #522, opened 2017-07-09
License Header Tool, Removal of Property Refinement

TanayParikh:
- Removed `PropertyRefinement.js` (empty file)
- Added licenses to files missing them
- Added synchronized dummy property to silence warnings
- Added a shell script which detects files within the `src/foam` directory which don't have `/**` at the start (indicating a missing license).

### DIFF
--- a/src/foam/java/InterfaceMethod.js
+++ b/src/foam/java/InterfaceMethod.js
@@ -40,6 +40,12 @@ foam.CLASS({
       setter: function() {},
       getter: function() {}
     },
+    {
+      name: 'synchronized',
+      documentation: 'Dummy property to silence warnings',
+      setter: function() {},
+      getter: function() {}
+    },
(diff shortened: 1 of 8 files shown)


tools/licenseFinder.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
kirkeaton, line 1: Nice tool

src/foam/java/InterfaceMethod.js
```diff
@@ -40,6 +40,12 @@ foam.CLASS({
       setter: function() {},
       getter: function() {}
     },
+    {
+      name: 'synchronized',
```
kirkeaton, line 44: This should actually be removed. Java 8 doesn't allow the `synchronized` as part of interfaces.

PR #524, opened 2017-07-10
Slot: Drop detach-on-class-change logic. Affords cases such as:

mdittmer:
view.add(view.data$.('foo');
...
view.data = undefined;
...
view.data = somethingWithFoo;


mdittmer:
TBR @kgrgreer

PR #528, opened 2017-07-10
Java Implements Flag, IN Predicate & MAP Sink Implementation

TanayParikh:
- Added optional `java` flag to `implements` array object to filter out mixins from Java output (as they are not necessarily interfaces and do cause conflicts. 

### Demo:
**Before:**
```js
  implements: [ 'foam.mlang.Expressions' ],
```

**After:**
```js
  implements: [ { path: 'foam.mlang.Expressions', java: false } ],
```
- Note, the `java` flag defaults to `true`
- The old method can still be used for the implements array.

### `IN` Predicate
- Implemented the `IN` predicate
   - Needed for Java Relationships (WIP)

### `MAP` Sink Decorator
- Implemented the `MAP` sink decorator
   - Needed for Java Relationships (WIP)

### Other Changes
- Added `foam.mlang.predicate.ArrayBinary` to `classes.js`

- Filtered `foam.mlang.Expressions` mixin from JS `IN`, `implements` array using java implements flag

- Added `.idea/` to gitignore (can be removed if it is an issue)


.gitignore
```diff
@@ -9,3 +9,4 @@ test/third_party/google-cloud-sdk*
 build
 foam-bin.js
 target/
+.idea/
```
kirkeaton, line 12: 👍

src/foam/java/refinements.js
```diff
@@ -136,7 +136,7 @@ foam.CLASS({
   refines: 'foam.core.Implements',
   methods: [
     function buildJavaClass(cls) {
-      cls.implements = (cls.implements || []).concat(this.path);
+      if (this.java) cls.implements = (cls.implements || []).concat(this.path);
```
kgrgreer: Follow FOAM style guide.

src/foam/mlang/MLang.java
```diff
@@ -17,6 +17,8 @@
  */
 public class MLang
 {
+  public static final Predicate TRUE = new True();
```
kgrgreer, line 20: This makes me realize that we have a bug since True and False should be Singletons. I'll fix.

src/foam/java/refinements.js
TanayParikh: Sorry about that, updated.

src/foam/mlang/MLang.java
TanayParikh: Ok, thanks!

PR #529, opened 2017-07-10
XML Js ( FOAM FObject <==> XML ) 

pavneetskapoor:
- Created parser and writer for XML to FObject and vice versa

Can be used through the following commands:

 ```javascript 
       foam.xml.stringify(FObject)
       // For Pretty Printing
       foam.xml.Pretty.stringify(FObject)  
```
 ```javascript 
       foam.xml.objectify(String)
```


src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+      return foam.xml.Pretty.stringify(this);
+    }
+  ]
+});
+
+
+/** XML Outputter **/
+foam.CLASS({
+  package: 'foam.xml',
+  name: "Outputer",
```
kirkeaton: Should be Outputter

kirkeaton:
Could you write the XML Driver as well: https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/export/XMLDriver.js

It should be very similar to the other drivers like this one for example: https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/export/CSVDriver.js

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+      return foam.xml.Pretty.stringify(this);
+    }
+  ]
+});
+
+
+/** XML Outputter **/
+foam.CLASS({
+  package: 'foam.xml',
+  name: "Outputter",
```
adamvy, line 50: Can we make an abstract base class that both JSON and XML share?  There is a lot of code overlap between the two.

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+        Date:      function(o) { this.outputDate(o); },
+        Function:  function(o) { this.outputFunction(o); },
+        AbstractEnum: function(o) { },
+        FObject:   function(o) {
+          if ( o.outputXML ) {
+            o.outputXML(this)
+            return;
+          }
+
+          var clsName = o.cls_.id;
```
adamvy, line 290: We should pass clsName through a local method like "escapeAttribute" or "outputAttributeValue" that ensures that any nested quotes are escaped.  Not that we expect quotes in class id's but better to be safe.

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+            if ( !o[i].cls_ ) this.out("</value>");
+            if ( o.length - i != 1 ) this.nl().indent();
+          }
+        },
+        Object: function(o) {
+          if ( o.outputXML ) {
+            o.outputXML(this);
+          } else {
+            var oName = o.name;
+            this.start("<object name='" + oName + "'>");
```
adamvy, line 314: Same here, make sure we pass oName through some sort of escaping method.

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+      }
+      var ret = this.buf_;
+      this.reset(); // reset to avoid retaining garbage
+      return ret;
+    },
+
+    {
+      name: 'createFObj',
+      code: foam.mmethod({
+        Object: function (o) {
```
adamvy: Pass a context parameter in like we do in foam.core.JSON.parse

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+          // Create FObject
+          var className = o.className;
+          var obj = foam.lookup(className).create();
```
adamvy: And the use that context for looking up and creating the class.

See JSON.js:546

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+      } else {
+        this.output(o);
+      }
+      var ret = this.buf_;
+      this.reset(); // reset to avoid retaining garbage
+      return ret;
+    },
+
+    {
+      name: 'createFObj',
```
adamvy: Would this be better named parse rather than createFObject?

It's responsible for taking some sort of document representation and then turning it into real FObject instances, much like foam.core.JSON.parse is.

In particular, it doesn' tjust create one FObject like the name would imply, if given an array it may create a whole bunch.

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+        Object: function (o) {
+          // Create FObject
+          var className = o.className;
+          var obj = foam.lookup(className).create();
+          var props = o.children;
+
+          // Populate FObject with properties
+          for ( var propIndex = 0; propIndex < props.length; propIndex++ ) {
+
+            var currentProp = props[propIndex];
```
adamvy: The naming here is confusing.  currentProp is not actually a property, it's a node.  In a function that has variables both holding nodes and Property objects I would expect currentProp to hold a property.  Can we rename to currentNode or just node?

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+    },
+
+    {
+      name: 'objectify',
+      code: foam.mmethod({
+        String: function (o) {
+          if ( o ) {
+            // Convert xml string into an xml DOM object for node traversal
+            var parser = new DOMParser();
+            o = o.replace(/\t|\n|\r|↵/g, "");
```
adamvy: Does parseFromString not support these special characters?  They are allowed in the XML spec as far as I know and they may have been meaningful parts of property values.  What happens if we serialize/deserialize a multi-line string with this?

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+
+            // Nested Object
+            if ( childName === 'object' ) {
+              var nestObj = this.createFObj(currentProp.firstChild);
+              prop.set(obj, nestObj);
+              continue;
+            }
+
+            // Regular Prop
+            if ( currentProp.firstChild.nodeValue ) {
```
adamvy: Comment says "regular prop" but this appears to be code special casing for Enums?

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+            var childName = currentProp.firstChild.localName;
+            // Specific case for array
+            if ( currentProp.className === 'Array' ) {
+              // Array of FObjects
+              if ( childName === 'object' ) {
+                var nestObjArray = Array.from(currentProp.childNodes);
+                prop.set(obj, this.createFObj(nestObjArray));
+              } else {
+                // Array of other objects with 'value' tag names
+                var arrayValue = (Array.from(currentProp.children)).map( function (x) { return x.innerHTML; });
```
adamvy: innerHTML returns the inner HTML as a string does it not?  Should we be digging in to each child and parsing it?

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+          if ( o ) {
+            // Convert xml string into an xml DOM object for node traversal
+            var parser = new DOMParser();
+            o = o.replace(/\t|\n|\r|↵/g, "");
+            var xmlDoc = parser.parseFromString(o, "text/xml");
+            var rootName = xmlDoc.firstChild.nodeName;
+            // Check if multiple objects
+            if ( rootName === 'objects' ) {
+              return this.createFObj(Array.from(xmlDoc.firstChild.childNodes));
+            } else if ( rootName === 'object' ) {
```
adamvy: This conditional looks very similar to the one on line 359, can we refactor this a bit?

What about adding a parseNode method which takes a node and checks its name to decide whether to create an array or not.  Then this just becomes.

return this.parseNode(xmlDoc.firstChild);

and line 359 becomes  return prop.set(obj, this.parseNode(xmlDoc.firstChild));

Parse node would look something like.

```
parseNode(node) {
  if ( node.nodeName === 'objects' ) return this.parse(Array.from(node.childNodes));
  else if ( node.nodeName === 'object') return this.parse(node);
  
  warn("unknown node encountered, continuing.");
  return undefined;
}
```

Something like that to help reduce the conditionals in createFObj which is a bit difficult to follow.

src/foam/core/XML.js
pavneetskapoor: Agreed. Better to keep it consistent and clear with `parseFObj` as the name

adamvy: I would just call it parse() to be consistent with foam.core.JSON.  Although I agree it's probably not the clearest name I think this should deviate from foam.core.JSON as little as possible since they are almost  the same just slightly different markup.

src/foam/core/XML.js
```diff
@@ -0,0 +1,466 @@
...
+@license
+Copyright 2017 The FOAM Authors. All Rights Reserved.
+http://www.apache.org/licenses/LICENSE-2.0
+*/
+
+foam.CLASS({
+  refines: 'foam.core.Property',
+
+  properties: [
+    { class: 'String', name: 'shortName' },
```
kgrgreer: I moved this into the standard Property definition so it can be reused by both XML and JSON support. So just remove this line.

src/foam/core/XML.js
kgrgreer: Since Pavi's last day, add a TODO and create an issue.

src/foam/core/XML.js
kgrgreer: I don't think this (or the previous) escaping is needed since if the property contains a quote character, then you code isn't going to run anyway. Better to validate Property names to exclude quotes than add checks all over the code, especially when you would already be broken anyway.

src/foam/core/XML.js
pavneetskapoor: I assumed that when we set the property of an Object, there was an `adapt` method within PropertyInfo that would parse correctly from String and convert the correct type.

src/foam/core/XML.js
pavneetskapoor: Done

src/foam/core/XML.js
pavneetskapoor: Added more comments to explain.

src/foam/core/XML.js
pavneetskapoor: Done.

src/foam/core/XML.js
pavneetskapoor: Done

src/foam/core/XML.js
pavneetskapoor: Done

src/foam/core/XML.js
pavneetskapoor: On line 362, the conditional requires for a mapping of the innerHTML within the children of the currentNode which is not common amongst the case shown here. I don't know if refactoring would work here.

pavneetskapoor: On line 358, the check is for FObjects that are nested within the object itself. If it's not an array of nested FObjects, then it will be primitive typed values that have to be parsed. In the case of line 411, it is checking whether the XML string contains multiple objects or a single object to parse.

src/foam/core/XML.js
adamvy: I think its good practice to pass all XML attribute values through an escape step, regardless of their source.  That way if code ever gets copy/pasted or moved and the new author doesn't know what assumptions the original author made about the source of the value.  There could be other XML escaping rules that don't apply to property names.

kgrgreer:
Please fix merge conflict.

Issue #484, opened 2017-06-25
JDAO.java bug in removeAll()


kgrgreer:
removeAll_() journals a remove-all operation but doesn't check the predicate or limit parameters. So, something like dao.limit(0).where(false).removeAll() would remove all data from the journal, whereas no data should actually be removed.
@drish

drish:
on it

kgrgreer:
MapDAO has the same problem.

drish:
@kgrgreer can we close this issue ?

PR #530, opened 2017-07-10
make jdao removeAll Sink aware

drish:
#484


drish:
not ready yet to be merged
wip

src/foam/dao/RemoveSink.java
```diff
@@ -0,0 +1,30 @@
+package foam.dao;
+
+import foam.core.Detachable;
+import foam.core.FObject;
+import foam.core.X;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+
+public class RemoveSink {
```
kgrgreer: This isn't a Sink since it doesn't implement the Sink interface.

src/foam/dao/JDAO.java
```diff
@@ -50,10 +51,17 @@ public FObject remove_(X x, FObject obj) {
   }
 
   @Override
-  public void removeAll_(X x, long skip, long limit, Comparator order, Predicate predicate) {
+  public void removeAll_(final X x, long skip, final long limit, Comparator order, Predicate predicate) {
     // TODO: this is wrong, should only call journal.removeAll() if neither limit nor predicate
     // are set.
-    journal.removeAll();
+
+    getDelegate().select_(x, new AbstractSink() {
```
kgrgreer: Instead of doing "new AbstractSink()" you should be doing "new RemoveSink()".

src/foam/dao/RemoveSink.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public RemoveSink(final DAO dao) {
```
kgrgreer: RemoveSink's constructor should just store the 'dao' and do nothing else. It should have a put() method that does dao.remove(x, obj);

src/foam/dao/RemoveSink.java
```diff
@@ -6,25 +6,32 @@
...
+
+  protected DAO dao;
+
+  public RemoveSink(DAO dao) {
+    this.dao = dao;
+  }
+
+  @Override
+  public void put(FObject obj, Detachable sub) {
+    final X x = ((AbstractDAO) this.dao).getX();
```
kgrgreer: Replace these two lines with just: dao.remove(obj);

dao already has a context and you don't need "this." in Java.

src/foam/dao/RemoveSink.java
```diff
@@ -6,25 +6,32 @@
...
 
-  public RemoveSink(final DAO dao, long skip, long limit, Comparator order, Predicate predicate) {
-    final X x = ((AbstractDAO) dao).getX();
-    dao.select_(x, new AbstractSink() {
-      @Override
-      public void put(FObject obj, Detachable sub) {
-        dao.remove_(x, obj);
-      }
-    }, skip, limit, order, predicate);
+  @Override
```
kgrgreer: If you extend AbstractSink, then you don't need to implement these NOP methods.

drish: oh ok, thanks. 
done.

src/foam/dao/JDAO.java
```diff
@@ -50,10 +51,11 @@ public FObject remove_(X x, FObject obj) {
   }
 
   @Override
-  public void removeAll_(X x, long skip, long limit, Comparator order, Predicate predicate) {
+  public void removeAll_(final X x, long skip, final long limit, Comparator order, Predicate predicate) {
```
kgrgreer, line 54: If you move this to AbstractDAO instead of JDAO, then it will fix MapDAO also.

PR #535, opened 2017-07-12
Dedicated and Shared Worker Registry

m-cheung:
Please do not merge - Work in progress.
Input / suggestions requested.


src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,116 @@
...
+  requires: [
+    'foam.core.StubFactorySingleton',
+    'foam.box.BoxRegistry',
+    'foam.box.node.ForkBox'
+  ],
+
+  imports: [ 'registry' ], // __context__.registry is the context it is declared in
+
+  exports: [ 'as registry' ], // __subContext__.registry === this
+
```
m-cheung: I am not sure if this is the appropriate action. Previous, it was discussed that we would do something along the lines of
```
var ctx = foam.box.Context();
ctx.registry = foam.box.DedicatedWorkerRegistry.create({
  delegate: ctx.registry
}, ctx);
```
But this causes a cyclic reference to itself. If we create them separately
```
var ctx = foam.box.Context();
var registry = foam.box.DedicatedWorkerRegistry.create({
  delegate: ctx.registry
}, ctx);
```
then use `registry.register`, registrations will be ok, but issues will come up later on. An example:
```
var stub = foam.core.StubFactorySingleton.create().get(foam.dao.DAO)
   .create({ delegate: register }, ctx);
stub.put(...)
```
As this will register will a different registry, and then the stub call will not be able to find the box in the registry, (in the current implementation)

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,116 @@
...
+        stubFactory: function() {
+          return self.stubFactory_.create({
+            delegate: self.workerFactory(this)
+          }, this);
+        },
+
+        registerWorker: function(name, service, box) {
+          return self.dedicatedWorkers_[key].register(name, service, box);
+        }
+      });
```
m-cheung: Perhaps, line 64-74 is where my misinterpretation begins? I don't think you intended me to handle the context's in this sense.

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,116 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+foam.CLASS({
+  package: 'foam.box',
+  name: 'DedicatedWorkerRegistry',
+  extends: 'foam.box.ProxyBox',
+  implements: [ 'foam.box.BoxRegistryBox' ],
```
m-cheung: I naturally thought that this is more of an extension of `foam.box.BoxRegistryBox` and it implements `foam.box.ProxyBox` but doing so is not possible with importing registry as well.

(Additionally, most places tend to extend `foam.box.ProxyBox`, so I believe this should be the right way of doing this)

src/foam/box/SharedWorkerRegistry.js
```diff
@@ -0,0 +1,83 @@
+/**
```
m-cheung, line 1: This file can be ignored for now. It will be updated once `DedicatedWorkerRegistry` is fixed.

test/src/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,195 @@
...
+
+    MockRegistry = foam.lookup('foam.box.DedicatedWorkerRegistry.Test.MockRegistry');
+    Registry = foam.lookup('foam.box.DedicatedWorkerRegistry');
+    Context = foam.lookup('foam.box.Context');
+    LogBox = foam.lookup('foam.box.LogBox');
+  });
+
+  it('should forward all registrations to delegate by default', function() {
+    var ctx = Context.create();
+    var mockRegistry = MockRegistry.create(null, ctx);
```
m-cheung: Assume that `mockRegistry` is acting as the "default" registry throughout these tests

test/src/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,195 @@
...
+
+    ctx.registry = Registry.create({
+      delegate: mockRegistry
+    }, ctx);
```
m-cheung: Is this how we intend on using `DedicatedWorkerRegistry`? Or do we intend to use the registry separately from the context?

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+      this.register = oldReg;
+      return ret;
+    },
+    function unregister(name) {
+      // Attempt to unregister in ourselves first.
+      if ( foam.box.Box.isInstance(name) ) {
+        var key = this.getDedicatedWorkerKey(name);
+
+        for (var key in this.dedicatedWorkers_ ) {
+          if ( this.dedicatedWorkers_[key] === name ) {
```
m-cheung: Do we need to worry about removing `StubFactory<class>` from `dedicatedWorkers_`?
If not, we can simply this code

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+foam.CLASS({
+  package: 'foam.box',
+  name: 'DedicatedWorkerRegistry',
+  extends: 'foam.box.BoxRegistryBox', //'foam.box.ProxyBox',
```
mdittmer: nit: Here and elsewhere: Drop commented-out code.

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+    'foam.box.node.ForkBox'
+  ],
+
+//  imports: [ 'registry' ], // __context__.registry is the context it is declared in
+
+  exports: [ 'as registry' ], // __subContext__.registry === this
+
+  properties: [
+    {
+      name: "getDedicatedWorkerKey",
```
mdittmer: `class: 'Function'`. You may need to use a `factory` instead of `value` if `foam.core.Function` has a factory.

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+      value: function(box) { return null; }
+    },
+    {
+      name: 'workerFactory',
```
mdittmer: `class: 'Function',`

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+    'foam.core.StubFactorySingleton',
+    'foam.box.BoxRegistry',
+    'foam.box.node.ForkBox'
+  ],
+
+//  imports: [ 'registry' ], // __context__.registry is the context it is declared in
+
+  exports: [ 'as registry' ], // __subContext__.registry === this
+
+  properties: [
```
mdittmer: Per-property documentation strings, please :smile:

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+foam.CLASS({
+  package: 'foam.box',
+  name: 'DedicatedWorkerRegistry',
```
mdittmer, line 20: Documentation with example, please 😸

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+      factory: function() {
+        return this.StubFactorySingleton.create().get(this.BoxRegistry);
+      }
+    }
+  ],
+
+  methods: [
+    function register(name, service, box) {
+      var key = this.getDedicatedWorkerKey(box);
+      if ( ! key ) return this.delegate.register(name, service, box);
```
mdittmer: if `delegate` isn't marked as `required: true` in `ProxyBox`, then we should add that in this class:

```js
properties: [
  {
    name: 'delegate',
    documentation: `...`,
    required: true
  }
]
```

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+
+      // Redirecting all registers of following statements with our registry.
```
mdittmer: Wait, what?

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+  exports: [ 'as registry' ], // __subContext__.registry === this
+
+  properties: [
+    {
+      name: "getDedicatedWorkerKey",
+      value: function(box) { return null; }
+    },
+    {
+      name: 'workerFactory',
+      value: function(ctx) {
```
mdittmer: Let's make this receive `(args, ctx)`, like the `create` interface.

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+
+      // Redirecting all registers of following statements with our registry.
+      // If this is not done, this register method will be called recursively,
+      // and registrations will be forwarded to delegate.
+      var oldReg = this.register;
+      this.register = this.SUPER;
+
+      if ( ! this.dedicatedWorkers_[key] )
+        this.dedicatedWorkers_[key] = this.stubFactory_.create({
+          delegate: this.workerFactory(this)
```
mdittmer: Shouldn't `(this)` be `(this.__context__)` to avoid the aforementioned recursion? Such a strategy instantiates `ForkBox` (or whatever `workerFactory` is injected to do) in a context that has `this`'s `registry`, not `this`'s subcomponents' registry.

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+  methods: [
+    function register(name, service, box) {
+      var key = this.getDedicatedWorkerKey(box);
+      if ( ! key ) return this.delegate.register(name, service, box);
+
+      // Redirecting all registers of following statements with our registry.
+      // If this is not done, this register method will be called recursively,
+      // and registrations will be forwarded to delegate.
+      var oldReg = this.register;
+      this.register = this.SUPER;
```
m-cheung: Is there a safer way of doing this? The following function needs to run in the same context, except with a different register method. I tried using subContext's but I could not find a way to change just the method

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+      var ret = this.dedicatedWorkers_[key].register(name, service, box);
+
+      // Register is set back to this method for future calls.
+      this.register = oldReg;
+      return ret;
+    },
+    function unregister(name) {
+      // Attempt to unregister in ourselves first.
+      if ( foam.box.Box.isInstance(name) ) {
+        var key = this.getDedicatedWorkerKey(name);
```
mdittmer: I don't think this will work reliably. `name` here is the box returned above (`ret`), not the box passed in to `register()` above.

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+
+        for (var key in this.dedicatedWorkers_ ) {
```
mdittmer: ditto.

src/foam/box/DedicatedWorkerRegistry.js
mdittmer: To do this correctly we would need to keep a local collection of services (still) registered on this dedicated worker. When the size of the collection reaches zero, then we could delete this. For now, I'd say it can be a `TODO` to clean up dedicated workers.

mdittmer (changes requested):
Reviewed `foam.box.DedicatedWorkerRegistry`. In addition to inline comments, `unregister()` will be unsound without additional state tracking.

`register()` needs mutate some kind of  bookkeeping that associates workers with names and/or exported boxes (`ret` in `register()`). Getting the name when name was unspecified is a an edge case here. Another edge case is when the return value from `dedicatedWorker_[key].register(...)` has a `PromisedBox` in its delegate chain (so its `Promise` could reject, and that aforementioned bookkeeping needs to be cleaned up). Then, `unregister()` can use the bookkeeping to determine whether or not it should handle the call or delegate. It should never do both.

We can chat about this offline if you like.

src/foam/box/DedicatedWorkerRegistry.js
m-cheung: I may still have something wired incorrectly then. I checked both `this.__context__.registry` and `this.__subContext.registry` and they are both referring to `this` (i.e. `this.__context__.registry === this`)

src/foam/box/DedicatedWorkerRegistry.js
m-cheung: I have actually simplified this to just
```
this.SUPER(name); 
this.delegate.unregister(name);
```

This does assume that name of the box is unique (I believe this is true, but I may have to add a check on register) and should only be registered in one of the registries (this should be true).

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -18,36 +18,81 @@
...
 
-//  imports: [ 'registry' ], // __context__.registry is the context it is declared in
-
-  exports: [ 'as registry' ], // __subContext__.registry === this
+  exports: [ 'as registry' ],
 
   properties: [
     {
+      class: 'Function',
       name: "getDedicatedWorkerKey",
```
mdittmer: nit: single quotes

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -18,36 +18,81 @@
...
-      value: function(box) { return null; }
+      documentation: `Gets the service name from the provided box.
+        If the value is not null, a dedicated worker will be used for the box.
```
mdittmer: nit: +4 indent after first line.

src/foam/box/DedicatedWorkerRegistry.js
```diff
@@ -57,7 +102,9 @@ foam.CLASS({
   methods: [
     function register(name, service, box) {
       var key = this.getDedicatedWorkerKey(box);
-      if ( ! key ) return this.delegate.register(name, service, box);
+      if ( ! key ) {
+        return this.delegate.register(name, service, box);
+      }
```
mdittmer: Lines 109-114, 121-123 should be gone now, right?

test/src/box/DedicatedWorkerRegistry.js
mdittmer: It should be usable anywhere you would use a `BoxRegistry`. In practice, it's likely that a `PipelineBuilder` will have `imports: [ 'registry' ]` which will probably contain `DedicatedWorkerRegistry -> SharedWorkerRegistry`, possibly with `-> BoxRegistry` on the end.

test/src/box/DedicatedWorkerRegistry.js
```diff
@@ -70,10 +70,14 @@ describe('dedicated worker registry', function() {
       delegate: mockRegistry
     }, ctx);
 
+    /*
```
mdittmer: nit: Clean this up once test code is as you want it.

src/foam/box/BoxRegistry.js
```diff
@@ -70,6 +70,13 @@ foam.CLASS({
       ]
     },
     {
+      name: 'getRegisteredNames',
+      returns: 'Array',
+      code: function() {
+        return Object.keys(this.registry);
+      }
+    },
+    {
```
m-cheung: Having `DedicatedWorkerRegistry` keep track of registrations is not sufficient. This is because we pass `localRegistry` to functions, which make registrations that do not go through `DedicatedWorkerRegistry` (but technically belonging to `DedicatedWorkerRegistry`).

Thus, name conflicts would still be possible if we only relied on a map property that is updated upon register and unregister.

src/foam/box/BoxRegistryBox.js
```diff
@@ -45,7 +45,7 @@ foam.CLASS({
         if ( this.SubBoxMessage.isInstance(msg.object) ) {
           var name = msg.object.name;
 
-          if ( this.registry[name].localBox ) {
+          if ( this.registry[name] && this.registry[name].localBox ) {
```
m-cheung, line 48: Previous check was insufficient. If `this.registry[name]` is not defined, it will not move on to attempting to send an error back through `msg.attributes.errorBox`

src/foam/box/DedicatedWorkerRegistry.js
m-cheung: Addressed below.

test/src/box/DedicatedWorkerRegistry.js
```diff
@@ -0,0 +1,256 @@
...
+    var dao = foam.dao.ArrayDAO.create({ of: 'Object' });
+    var box = foam.box.SkeletonBox.create({ data: dao });
+    box.serviceName = "ArrayDAO"; // Enabling dedicated service worker.
+
+    // Perform remote registration of service.
+    var register = ctx.registry.register(null, null, box);
+
+    // Creating a stub on local for accessing method.
+    var stub = foam.core.StubFactorySingleton.create().get(foam.dao.DAO)
+        .create({ delegate: register }, ctx.registry.localRegistry);
```
m-cheung, line 135: Should the user need to know which registry the box above was registered in? If `ctx` is provided instead (i.e. allowing `DedicatedWorkerRegistry` to handle the registration), it will perform registrations for the stub in the delegate's registry. This results in issues when we call `stub.put(obj)` below (it will not be able to find the ReplyBox).

A stack trace captured at the lookup failure:
```
code | @ | BoxRegistryBox.js:48
send | @ | DedicatedWorkerRegistry.js:146
send | @ | NamedBox.js:53
code | @ | SubBox.js:42
(anonymous function) | @ | Promised.js:56
_tickCallback | @ | internal/process/next_tick.js:109
_Promise.resolve (async)_
(anonymous function) | @ | ForkBox.js:81
send | @ | ReplyBox.js:42
code | @ | BoxRegistryBox.js:51
(anonymous function) | @ | net.js:471
```

To be clear: providing `ctx.registry.localRegistry` will work fine. I'm just not sure if the user should be required to specify the correct registry.

m-cheung:
@mdittmer I noticed while writing my tests that processes spawned by ForkScript are not killed if the exit status is 0. When running Jasmine tests, the child processes will continue to run after, even when all of the tests have completed and Jasmine / Node has exited.

The child processes are killed if we send a kill signal (e.g. `Ctrl`+`C` or `SIGTERM`) is sent, but that is about it. Here is a snippet of code you can use to reproduce this behaviour.

```
require('./src/foam.js');                                                                                              
                                                                                                                                                                                                         
var remoteRegistry = stubFactory.get(foam.box.BoxRegistry).create({                                                    
  delegate: foam.box.node.ForkBox.create(null, ctx)                                                                    
}, ctx);                                                                                                               
                                                                                                                       
var serviceBox = remoteRegistry.register(null, null, foam.box.LogBox.create());                                                                     
serviceBox.send(foam.box.Message.create({ object: 'Hello world' }));

setTimeout(function() {
  process.exit();
}, 2000);
```

TLDR: I think we need a mechanism to clean up child processes in ForkBox

m-cheung:
Issues were addressed in #547

PR #536, opened 2017-07-12
Added of function to AbstractFObjectPropertyInfo

kirkeaton:
1. Added of function to AbstractFObjectPropertyInfo.
2. Added of method to FObjectProperty, Array, Enum, Proxy.
3. Consolidated duplicate FObjectProperty refinements.

### DIFF
--- a/src/foam/core/AbstractFObjectPropertyInfo.java
+++ b/src/foam/core/AbstractFObjectPropertyInfo.java
@@ -9,5 +9,5 @@ package foam.core;
 public abstract class AbstractFObjectPropertyInfo
   extends AbstractObjectPropertyInfo
 {
-  public abstract String of();
+  public abstract String getOf();
 }
--- a/src/foam/core/AbstractFObjectPropertyInfo.java
+++ b/src/foam/core/AbstractFObjectPropertyInfo.java
@@ -9,7 +9,5 @@ package foam.core;
 public abstract class AbstractFObjectPropertyInfo
   extends AbstractObjectPropertyInfo
 {
-  //  public int compareValues(FObject o1, FObject o2) {
-  //    return o1.compareTo(o2);
-  //  }
+  public abstract String of();
 }
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -533,6 +533,17 @@ foam.CLASS({
 foam.CLASS({
   refines: 'foam.core.FObjectProperty',
 
+  properties: [
+    {
+      name: 'javaType',
+      expression: function(of) {
+        return of ? of.id : 'foam.core.FObject';
+      }
+    },
+    ['javaInfoType', 'foam.core.AbstractFObjectPropertyInfo'],
+    ['javaJSONParser', 'foam.lib.json.FObjectParser']
+  ],
+
   methods: [
     function createJavaPropertyInfo_(cls) {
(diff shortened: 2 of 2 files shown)


src/foam/core/AbstractFObjectPropertyInfo.java
```diff
@@ -9,7 +9,5 @@
 public abstract class AbstractFObjectPropertyInfo
   extends AbstractObjectPropertyInfo
 {
-  //  public int compareValues(FObject o1, FObject o2) {
-  //    return o1.compareTo(o2);
-  //  }
+  public abstract String of();
```
kgrgreer: I think this would be more useful if it returned ClassInfo instead of String.

kgrgreer: DAO's use getOf(), so you could rename it to that for consistency, but I can live with either.

kirkeaton: How would I resolve to the ClassInfo given the class String?

kirkeaton: Changed it to getOf()

src/foam/java/refinements.js
```diff
@@ -555,6 +573,19 @@ foam.CLASS({
...
+  ],
+
+  methods: [
+    function createJavaPropertyInfo_(cls) {
+      var info = this.SUPER(cls);
+      info.method({
+        name: 'getOf',
+        visibility: 'public',
+        type: 'String',
+        body: 'return "' + (this.of ? this.of.id ? this.of.id : this.of : null) + '";'
```
kgrgreer, line 585: 'of' is required for Enums, so would just 'this.of'id' be sufficient?

src/foam/java/refinements.js
```diff
@@ -555,6 +573,19 @@ foam.CLASS({
...
     ['javaJSONParser', 'foam.lib.json.FObjectParser']
+  ],
+
+  methods: [
+    function createJavaPropertyInfo_(cls) {
+      var info = this.SUPER(cls);
+      info.method({
+        name: 'getOf',
+        visibility: 'public',
+        type: 'String',
```
kgrgreer, line 584: Wouldn't ClassInfo be a more useful return type than String?

src/foam/java/refinements.js
```diff
@@ -672,6 +687,13 @@ foam.CLASS({
       var info = this.SUPER(cls);
       var compare = info.getMethod('compare');
       compare.body = this.compareTemplate();
+
+      info.method({
+        name: 'getOf',
```
kgrgreer, line 692: How/where is this used? Is it needed?

If 'of' is set, shouldn't it change the 'javaType' above?

src/foam/java/refinements.js
```diff
@@ -821,6 +843,19 @@ foam.CLASS({
...
     ['javaJSONParser', 'foam.lib.json.FObjectParser']
+  ],
+
+  methods: [
+    function createJavaPropertyInfo_(cls) {
+      var info = this.SUPER(cls);
+      info.method({
+        name: 'getOf',
+        visibility: 'public',
+        type: 'String',
```
kgrgreer, line 854: Wouldn't ClassInfo be a more useful return type?

PR #538, opened 2017-07-12
Java Long Epoch Date Parsing

TanayParikh:
- Converts an Epoch timestamp to a Java Date

### DIFF
--- a/src/foam/lib/json/DateParser.java
+++ b/src/foam/lib/json/DateParser.java
@@ -12,23 +12,24 @@ public class DateParser
   extends ProxyParser
 {
   public DateParser() {
-    super(new Seq(
-                  new Literal("\""),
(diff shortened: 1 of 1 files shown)


src/foam/lib/json/DateParser.java
```diff
@@ -39,12 +40,18 @@ public PStream parse(PStream ps, ParserContext x) {
 
     System.out.println("Succcess.");
 
+    // Checks if Long Date (Timestamp from epoch)
+    if (ps.value() instanceof java.lang.Long) {
```
kgrgreer: if ( ps.value() instanceof java.lang.Long ) {

TanayParikh: Updated, thanks.

PR #539, opened 2017-07-13
Renaming "registry" property in BoxRegistry

m-cheung:
This change attempts to reduce confusing and complications in other scenarios. For example, when import registry, overwriting the property.


mdittmer (approved):
This LGTM. @adamvy PTAL. Rationale: Naming the local map "registry" (same name imported to access a `BoxRegistry` implementation) was causing imports/exports issues with composable registries that @m-cheung is working on.

PR #541, opened 2017-07-13
Added logging & PM to CronScheduler

kirkeaton:


src/foam/nanos/cron/CronScheduler.java
```diff
@@ -36,31 +38,33 @@ private Date getMinScheduledTime() {
...
-          });
+  public void start() {
+    cronDAO_ = (DAO) getX().get("cronDAO");
+    new Thread(this).start();
+  }
+
+  @Override
+  public void run() {
+    NanoLogger logger = (NanoLogger) getX().get("logger");
+    final PM pm = new PM(this.getClass(), "cronScheduler");
```
kgrgreer: Move this to line 56.5 so that each cron job gets a new PM start time.

src/foam/nanos/cron/CronScheduler.java
```diff
@@ -36,31 +38,33 @@ private Date getMinScheduledTime() {
...
-        e.printStackTrace();
+    try {
+      while ( true ) {
+        Date now = new Date();
+        cronDAO_.where(MLang.LTE(Cron.SCHEDULED_TIME, now)).select(new AbstractSink() {
+          @Override
+          public void put(FObject obj, Detachable sub) {
+            ((Cron) obj).runScript(CronScheduler.this.getX());
+            cronDAO_.put(obj);
+            pm.log(getX());
```
kgrgreer: Put this in a finally block so that it PM's even if the cronjob throws an Exception.;

src/foam/nanos/cron/CronScheduler.java
```diff
@@ -36,31 +38,33 @@ private Date getMinScheduledTime() {
...
+          }
+        });
+
+        Date minScheduledTime = getMinScheduledTime();
+        Thread.sleep(minScheduledTime.getTime() - now.getTime());
       }
+    } catch (InterruptedException e) {
```
kgrgreer: Generalize this to Throwable so that we catch all errors. Also, this will prevent exceptions from shutting down the cron-service, which would be bad.

src/foam/nanos/cron/CronScheduler.java
kirkeaton: Would I need to change the serviceKey in this case? Maybe to be something that represents the cron job being run instead of the cron scheduler as a whole.

src/foam/nanos/cron/CronScheduler.java
kirkeaton: Done

src/foam/nanos/cron/CronScheduler.java
kirkeaton: Done

PR #542, opened 2017-07-13
Fixing MDAO.find to look through Id only and adding TreePlan

marcogbarcellos:


src/foam/dao/index/AltIndex.java
```diff
@@ -67,6 +61,11 @@ public Object removeAll() {
     return s;
   }
 
+  public Object get(Object state, Object key) {
```
kgrgreer: Why is this method needed?

src/foam/dao/index/TreeNode.java
```diff
@@ -256,7 +256,8 @@ public Object get(TreeNode s, Object key, PropertyInfo prop) {
     if ( s == null ) {
       return s;
     }
-    int r = prop.compare(s.value, key);
+    int r = java.lang.Long.compare((Long)prop.get(s.value), (Long)key);
```
kgrgreer: Doesn't this change break use of String ids or String keys?

src/foam/dao/index/TreePlan.java
```diff
@@ -0,0 +1,25 @@
+package foam.dao.index;
+
+import foam.core.FObject;
+import foam.core.PropertyInfo;
+
+public class TreePlan implements FindPlan {
```
kgrgreer: TreePlan doesn't seem like a very descriptive name since there will be more than one Tree related Plan. Maybe TreeLookupFindPlan.

src/foam/dao/index/TreeIndex.java
```diff
@@ -71,14 +64,14 @@ public Object put(Object state, FObject value) {
   public Object remove(Object state, FObject value) {
     return ((TreeNode) state).removeKeyValue((TreeNode) state, prop_, prop_.f(value), value, tail_);
   }
-
+
```
kgrgreer: remove extra space

src/foam/dao/index/TreeIndex.java
```diff
@@ -71,14 +64,14 @@ public Object put(Object state, FObject value) {
   public Object remove(Object state, FObject value) {
     return ((TreeNode) state).removeKeyValue((TreeNode) state, prop_, prop_.f(value), value, tail_);
   }
-
+ 
   public Object removeAll() {
     return TreeNode.getNullNode();
   }
-
+
```
kgrgreer: remove extra spaces

src/foam/dao/index/TreeLookupFindPlan.java
```diff
@@ -0,0 +1,29 @@
...
+public class TreeLookupFindPlan implements FindPlan {
+  
+  protected PropertyInfo prop_;
+  
+  public TreeLookupFindPlan(PropertyInfo prop) {
+    prop_ = prop;
+  }
+  //TODO
+  public long cost() {
+    return 1;
```
kgrgreer: Cost is log2(size)

src/foam/dao/index/TreeIndex.java
marcogbarcellos: I don't think there's an extra space here. I just put a blank line between the methods as you told me to do before. That's right, isn't it?

adamvy: The github diff looks like the line has some whitespace at the beginning.

src/foam/java/refinements.js
```diff
@@ -861,11 +922,16 @@ foam.CLASS({
     ['javaJSONParser', 'foam.lib.json.BooleanParser'],
     ['javaInfoType', 'foam.core.AbstractBooleanPropertyInfo']
   ],
+
   methods: [
     function createJavaPropertyInfo_(cls) {
       var info = this.SUPER(cls);
+      var c = info.getMethod('comparePropertyValue');
+      c.body = 'return compareValues((boolean) key, (boolean) f(o));';
```
adamvy: Isn't this just a duplicate of the implementation already provided by AbstractBooleanPropertyInfo?

src/foam/core/AbstractBooleanPropertyInfo.java
```diff
@@ -12,4 +12,8 @@
   public int compareValues(boolean b1, boolean b2) {
     return Boolean.compare(b1, b2);
   }
+  
+  public int comparePropertyValue(Object key, FObject o) {
```
kgrgreer: A better name might be comparPropertyToObject().

src/foam/core/AbstractBooleanPropertyInfo.java
```diff
@@ -12,4 +12,8 @@
...
+    return compareValues((boolean) key, (boolean) f(o));
```
kgrgreer: This is very expensive because f() is going to create an object for each comparison. Which if inserting or finding and millions objects a second would be 20M objects/s. Instead of f(), calling get_() wouldn't create an object.

src/foam/dao/index/TreeIndex.java
kgrgreer: According to github, there's a blank line, but it contains one space.

src/foam/dao/index/TreeLookupFindPlan.java
```diff
@@ -0,0 +1,29 @@
+package foam.dao.index;
```
kgrgreer, line 6: Missing copyright header.

src/foam/java/refinements.js
```diff
@@ -458,6 +462,10 @@ foam.CLASS({
       m.body = 'return ( o instanceof Number ) ?'
             + '((Number)o).byteValue() :'
             + '(byte)o;';
+
+      var c = info.getMethod('comparePropertyValue');
+      c.body = 'return compareValues((byte) key, (byte) f(o));';
```
kgrgreer: Here and below call get_() instead of f().

src/foam/core/AbstractDoublePropertyInfo.java
```diff
@@ -12,4 +12,8 @@
   public int compareValues(double d1, double d2) {
     return Double.compare(d1, d2);
   }
+  
+  public int comparePropertyValue(Object key, FObject o) {
```
kgrgreer: Why are you including these comparePropertyValue() methods both here and below in the createJavaPropertyInfo_() methods? (I think it should only be below.)

src/foam/core/AbstractLongPropertyInfo.java
```diff
@@ -12,7 +12,7 @@
   public int compareValues(long o1, long o2) {
     return java.lang.Long.compare(o1, o2);
   }
-
+
```
kgrgreer: There are two spaces on this line.

src/foam/dao/index/TreeLookupFindPlan.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public long cost() {
+    return ((Double) Math.log(Long.valueOf(size_).doubleValue())).longValue();
+  }
+
+  public FObject find(Object state, Object key) {
+    if ( state != null && state instanceof TreeNode ) {
+      return (FObject)((TreeNode) state).get(((TreeNode) state), key, prop_);
+    }
+    if ( state != null && state instanceof Object[] &&
```
kgrgreer: When is state an Object[]?

src/foam/java/refinements.js
```diff
@@ -436,6 +436,10 @@ foam.CLASS({
       m.body = 'return ( o instanceof Number ) ?'
             + '((Number)o).intValue() :'
             + '(int)o;';
+
+      var c = info.getMethod('comparePropertyToObject');
+      c.body = 'return compareValues((int) key, (int) get_(o));';
```
kgrgreer: The second cast isn't required because that's already the declared return value of get_(). Here and in other implementations also.

PR #543, opened 2017-07-17
Java ENUM - JSON IO

TanayParikh:
- Implements `JSON <=> Java ENUM` conversion
- To be used in conjunction with the custom PropertyInfo for enum's that @pavneetskapoor is working on.


TanayParikh:
Some overlap w/#513, should still merge successfully. Ex. `of`, `forOrdinal`, `getOrdinal`.

src/foam/java/refinements.js
```diff
@@ -554,7 +554,78 @@ foam.CLASS({
...
     ['javaType', 'java.lang.Enum'],
     ['javaInfoType', 'foam.core.AbstractFObjectPropertyInfo'],
-    ['javaJSONParser', 'foam.lib.json.FObjectParser']
+    ['javaJSONParser', 'foam.lib.json.IntParser']
+  ],
+  methods: [
+    function createJavaPropertyInfo_(cls) {
+      var info = this.SUPER(cls);
+
+      const OF = (this.of ? this.of.id ? this.of.id : this.of : null);
```
kgrgreer: Can't this be simplified since 'of' is required for Enums?

src/foam/java/refinements.js
```diff
@@ -554,7 +554,78 @@ foam.CLASS({
...
+
+      info.method({
+        name: 'of',
+        visibility: 'public',
+        type: 'String',
```
kgrgreer: Wouldn't this be more useful as a ClassInfo that as a String?

src/foam/java/refinements.js
TanayParikh: @pavneetskapoor (overlap w/#513)

src/foam/java/refinements.js
TanayParikh: @pavneetskapoor (overlap w/#513)

pavneetskapoor: It would be more useful as ClassInfo, agreed. However, for arrays that are of a primitive type (ie. "Int") , returning ClassInfo leads to errors.

pavneetskapoor: Until primitive types have support for nested arrays, it could be consistently returning a string for now.

src/foam/java/refinements.js
pavneetskapoor: In that case would it just be ```this.of ? this.of.id : null```

kgrgreer: Doesn't just 'this.of.id' work?

src/foam/java/refinements.js
kgrgreer: But this is for EnumPropertyInfo. What does that have to do with primitive types?

pavneetskapoor: Yes for enums we could have of return class info instead.

src/foam/java/refinements.js
pavneetskapoor: Yes that would also work.

src/foam/java/refinements.js
TanayParikh: Removed the `of` function for now, it is covered in #513

Issue #544, opened 2017-07-17
JSONParser breaks when spaces are not included


MarcRoopchand:
JSONParser.parseString breaks while trying to parse the line below. 
{class:"foam.box.Message",attributes:{},object:{class:"foam.box.RPCMessage",name:"login",args:["asdf","adsf"]}}

The fix is to put a spaces after each comma or update the parser to parse no spaces
{class:"foam.box.Message", attributes:{}, object:{class:"foam.box.RPCMessage", name:"login", args:["asdf","adsf"]}}

@TanayParikh @kgrgreer suggested this be assigned to you

TanayParikh:
Thanks, I'll take a look.

PR #545, opened 2017-07-18
JSONParser Empty Map Support + Output

TanayParikh:
- Added support for empty maps
- Adding closing brace for map output

Resolves https://github.com/foam-framework/foam2/issues/544

### DIFF
--- a/src/foam/lib/json/MapParser.java
+++ b/src/foam/lib/json/MapParser.java
@@ -11,20 +11,23 @@ import foam.lib.parse.*;
 public class MapParser extends ProxyParser {
   public MapParser() {
     super(
-          new Seq1(2,
+          new Seq1(1,
                    new Whitespace(),
-                   new Literal("{"),
-                   new Repeat(
-                              new Seq2(1, 5,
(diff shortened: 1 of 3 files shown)


src/foam/lib/json/MapParser.java
```diff
@@ -11,27 +11,36 @@
...
-                                       new AnyKeyParser(),
-                                       new Whitespace(),
-                                       new Literal(":"),
-                                       new Whitespace(),
-                                       new AnyParser()),
-                              new Seq0(new Whitespace(), new Literal(","))),
-                   new Whitespace(),
-                   new Literal("}"))
+                   new Alt(
+                      new Seq1(1, new Literal("{"), new Whitespace(), new Literal("}")),
```
TanayParikh: Can't simply do `Alt` with `Whitespace` and `Repeat` as `Whitespace` matches all chars (even empty string) so a map key/value would never be picked up (and it would be ambiguous with `Repeat`'s first sequence which begins with a `Whitespace`).

PR #547, opened 2017-07-18
Selector registry

mdittmer:
A `BoxRegistry` implementation that accepts a `selector` that will select a delegate registry where service should be registered. In order for `NamedBox` and `SubBox` relays to work here, the `SelectorRegistry` expects its selector (a `RegistrySelector` implementation) to provide registries from a different (and differently named) `foam.box.Context`.

These caveats are documented in `(SelectorRegistry|RegistrySelector).documentation`.


src/foam/box/SelectorRegistry.js
```diff
@@ -0,0 +1,123 @@
...
+      registries according to its "selector".
+
+      NOTE: SelectorRegistry's delegation strategy expects registries returned
+      by "selector" to reside in a different foam.box.Context (with a different
+      foam.box.Context.myname) than the SelectorRegistry.`,
+
+  requires: [
+    'foam.box.NameAlreadyRegisteredException',
+    'foam.box.node.ForkBox'
+  ],
```
m-cheung: I don't believe you need these requirements anymore

src/foam/box/SelectorRegistry.js
```diff
@@ -0,0 +1,123 @@
...
+
+  methods: [
+    function register(name, service, box) {
+      var delegate = this.selector.select(name, service, box);
+      var delegateRegisteredBox = delegate.register(null, null, box);
+      this.selectorRegistrations_.push(this.Registration.create({
+        registry: delegate,
+        box: delegateRegisteredBox
+      }));
+      return this.SUPER(name, service, delegateRegisteredBox);
```
m-cheung, line 83: Heads up, you can still register two boxes with the same name right now. Neither `BoxRegistry` nor `BoxRegistryBox` implement a check to see if a box with a given name has been registered before. It will currently overwrite the previous registration of the same name with the new one.

src/foam/box/SelectorRegistry.js
```diff
@@ -0,0 +1,123 @@
...
+    },
+    function unregister(nameOrBox) {
+      // Similar to BoxRegistry implementation, but grab local box for
+      // unregistration in delegate.
+      var delegateRegisteredBox;
+      if ( foam.box.Box.isInstance(nameOrBox) ) {
+        for ( var key in this.registry ) {
```
m-cheung: Did you mean `for ( var key in this.registry_ )` by any chance?

mdittmer: Indeed.

src/foam/box/SelectorRegistry.js
mdittmer: That's definitely a bug. I'm not going to block this PR on it though.

mdittmer:
Filed #549 about underlying registration issue.

adamvy:
What is this used for?

mdittmer:
Pipelines use this to register services in different contexts. For example,
it facilitates registering a particular service in a different process, or
registering a service in multiple processes and returning a box that
implements a load balancing strategy.

PR #556, opened 2017-07-21
Tomcat integration

nick-prat:
Upped javax.sevlet version to 3.1.0, and implemented a ServletContextListener


src/foam/nanos/http/FileServlet.java
```diff
@@ -48,8 +48,15 @@ private void fileNotFoundError(HttpServletResponse resp, String file) {
...
-    String filePath = pathInfo.substring(SERVLET_NAME.length() + 2);
+    String pathInfo;
+    try {
+      pathInfo = req.getRequestURI();
+    } catch(Exception e) {
+      e.printStackTrace();
+      pathInfo = req.getPathInfo();
+    }
+
+    String filePath = pathInfo.substring(SERVLET_NAME.length() + (pathInfo.startsWith("/static/") ? 2 : 1));
```
kgrgreer: What is this check for? I could install multiple instances under multiple paths other than static.

src/foam/nanos/http/ServletHandler.java
```diff
@@ -219,9 +234,45 @@ public void handle(final HttpExchange ex) throws IOException {
...
-        return newInput.read();
+      public void setReadListener(ReadListener readListener) {
+        this.readListener = readListener;
+      }
+
+      @Override
+      public int read() {
+        int res = 0;
+        try {
+          if(isReady()) {
```
kgrgreer: Space next two lines according to coding standard.

src/foam/nanos/http/FileServlet.java
nick-prat: it lets you access the website with localhost:8080/static or localhost:8080/static/

kgrgreer: But how do you know that I've installed this servlet under /static? I could put it anywhere, including having multiple instances, like /lib and /img.

nick-prat: I just assumed that FileServlet would always be installed under /static

kgrgreer: Have your Servlet implement the foam.nanos.boot.NSpecAware interface then the router will call setNSpec(nspec) to provide you with your generating NSpec, from which you can get your actual path to use instead of 'static'.

PR #558, opened 2017-07-24
__Property__ Parsing + Decorated Sink Context + Ordered Sink Comparator

TanayParikh:
- Updated Java JSON parsing to use `ExprParser` instead of `FObjectParser`. The `ExprParser` has the `FObjectParser` as one of the `Alt` options.
- Added functionality for decorated sinks to have context
- Added explicit comparator to ordered sink `eof` sorting
- Java JSON array parsing (null class) bug fix

### DIFF
--- a/src/foam/dao/AbstractDAO.java
+++ b/src/foam/dao/AbstractDAO.java
@@ -61,6 +61,13 @@ public abstract class AbstractDAO
     return sink;
   }
 
+  protected Sink decorateSink_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+    Sink decoratedSink = decorateSink_(sink, skip, limit, order, predicate);
+    ((AbstractFObject) decoratedSink).setX(x);
+
+    return decoratedSink;
+  }
+
   public ClassInfo getOf() {
     return of_;
   }
--- a/src/foam/dao/AbstractDAO.java
+++ b/src/foam/dao/AbstractDAO.java
@@ -41,7 +41,7 @@ public abstract class AbstractDAO
     throw new UnsupportedOperationException();
   }
 
-  protected Sink decorateSink_(Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+  protected Sink decorateSink_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
     if ( limit < this.MAX_SAFE_INTEGER ) {
       sink = new LimitedSink(limit, 0, sink);
     }
@@ -58,14 +58,9 @@ public abstract class AbstractDAO
       sink = new PredicatedSink(predicate, sink);
     }
 
+    ((AbstractFObject) sink).setX(x);
+
     return sink;
-  }
-
-  protected Sink decorateSink_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
-    Sink decoratedSink = decorateSink_(sink, skip, limit, order, predicate);
-    ((AbstractFObject) decoratedSink).setX(x);
-
-    return decoratedSink;
   }
 
   public ClassInfo getOf() {
--- a/src/foam/dao/AbstractDAO.java
+++ b/src/foam/dao/AbstractDAO.java
@@ -46,13 +46,19 @@ public abstract class AbstractDAO
       sink = new LimitedSink(limit, 0, sink);
     }
 
+    ((AbstractFObject) sink).setX(x);
+
     if ( skip > 0 ) {
       sink = new SkipSink(skip, 0, sink);
     }
 
+    ((AbstractFObject) sink).setX(x);
+
     if ( order != null ) {
       sink = new OrderedSink(order, null, sink);
     }
+
+    ((AbstractFObject) sink).setX(x);
 
     if ( predicate != null ) {
       sink = new PredicatedSink(predicate, sink);
--- a/src/foam/dao/AbstractDAO.java
+++ b/src/foam/dao/AbstractDAO.java
@@ -41,30 +41,22 @@ public abstract class AbstractDAO
     throw new UnsupportedOperationException();
   }
 
-  protected Sink decorateSink_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+  protected Sink decorateSink_(Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
     if ( limit < this.MAX_SAFE_INTEGER ) {
       sink = new LimitedSink(limit, 0, sink);
     }
-
-    ((AbstractFObject) sink).setX(x);
 
     if ( skip > 0 ) {
       sink = new SkipSink(skip, 0, sink);
     }
 
-    ((AbstractFObject) sink).setX(x);
-
     if ( order != null ) {
       sink = new OrderedSink(order, null, sink);
     }
 
-    ((AbstractFObject) sink).setX(x);
-
     if ( predicate != null ) {
       sink = new PredicatedSink(predicate, sink);
     }
-
-    ((AbstractFObject) sink).setX(x);
 
     return sink;
   }
--- a/src/foam/dao/SinkJava.js
+++ b/src/foam/dao/SinkJava.js
@@ -167,7 +167,7 @@ foam.CLASS({
       name: 'eof',
       javaCode: 'if ( getArray() == null ) setArray(new java.util.ArrayList());\n'
                 + 'java.util.Collections.sort(getArray(), getComparator());\n'
-                + 'foam.dao.Subscription sub = getX().create(foam.dao.Subscription.class);\n'
+                + 'foam.dao.Subscription sub = new foam.dao.Subscription();\n'
                 + 'for ( Object o : getArray() ) {\n'
                 + '  if ( sub.getDetached() ) {\n'
                 + '    break;\n'
--- a/src/foam/dao/SinkJava.js
+++ b/src/foam/dao/SinkJava.js
@@ -166,7 +166,7 @@ foam.CLASS({
     {
       name: 'eof',
       javaCode: 'if ( getArray() == null ) setArray(new java.util.ArrayList());\n'
-                + 'java.util.Collections.sort(getArray());\n'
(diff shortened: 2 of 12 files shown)


src/foam/dao/AbstractDAO.java
```diff
@@ -61,6 +61,13 @@ protected Sink decorateSink_(Sink sink, long skip, long limit, Comparator order,
     return sink;
   }
 
+  protected Sink decorateSink_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
```
kgrgreer: Just move the body of the other decorateSink_ (the one without the X) here and remove the other.

src/foam/dao/SinkJava.js
```diff
@@ -111,21 +111,23 @@ foam.CLASS({
...
     {
       name: 'put',
-      javaCode: 'if ( getCount() >= getLimit() ) {\n'
-              + '  if ( sub != null ) sub.detach();\n'
-              + '} else {\n'
-              + '  setCount(getCount() + 1);\n'
-              + '  getDelegate().put(obj, sub);\n'
-              + '}\n'
+      javaCode: 
+`if ( getCount() >= getLimit() ) {
```
kgrgreer: I prefer this indented over.

src/foam/dao/SinkJava.js
```diff
@@ -111,21 +111,23 @@ foam.CLASS({
...
     },
     {
       name: 'remove',
-      javaCode: 'if ( getCount() >= getLimit() ) {\n'
-              + '  if ( sub != null ) sub.detach();\n'
-              + '} else {'
-              + '  setCount(getCount() + 1);\n'
-              + '  getDelegate().put(obj, sub);\n'
-              + '}\n'
+      javaCode:
```
kgrgreer: FYI: Best not to mix formatting and code changes in same PR. Then for the formatting changes are easier to review.

src/foam/lib/json/AnyParser.java
```diff
@@ -18,6 +18,6 @@ public AnyParser() {
         new BooleanParser(),
         new LongParser(),
         new DoubleParser(),
-        new FObjectParser()));
+        new ExprParser()));
```
kgrgreer: What is this change for? What if I want to parse an FObject with the AnyParser?

src/foam/lib/json/FObjectArrayParser.java
```diff
@@ -15,7 +15,7 @@ public FObjectArrayParser() {
                    new Literal("["),
                    new Whitespace(),
                    new Repeat(
-                              new FObjectParser(),
+                              new ExprParser(),
```
kgrgreer: Why?

src/foam/lib/json/JSONParser.java
```diff
@@ -12,7 +12,7 @@
 public class JSONParser
   extends foam.core.ContextAwareSupport
 {
-  private Parser   parser   = new FObjectParser();
+  private Parser   parser   = new ExprParser();
```
kgrgreer: Use protected instead of private. Makes easier to test and debug. Why the change?

src/foam/lib/json/Outputter.java
```diff
@@ -90,7 +90,9 @@ public void output(StringBuilder out, Object value) {
       outputFObject(out, (FObject)value);
     } else if ( value instanceof Number ) {
       outputNumber(out, (Number)value);
-    } else if ( value.getClass().isArray() ) {
+    } else if ( ( value != null ) &&
```
kgrgreer: Move expression to an isArray() method.

src/foam/dao/SinkJava.js
TanayParikh: Reverted multiline formatting for now

src/foam/dao/SinkJava.js
TanayParikh: Thanks, noted. I've reverted the commit for now. I'll create a separate PR for it.

src/foam/lib/json/AnyParser.java
TanayParikh: The `ExprParser` consists of an `Alt` with the two options being an `FObjectParser` and a custom parser for properties (`{ class: "__Property__", forClass_: "xyz" }`). This is needed to ensure properties are also parsed and handled correctly (the `__Property__` class isn't available from `foam.lookup()`, so the `FObjectParser` crashes if left as is).

src/foam/lib/json/JSONParser.java
TanayParikh: Updated. 

The `ExprParser` consists of an `Alt` with the two options being an `FObjectParser` and a custom parser for properties (`{ class: "__Property__". forClass_: "xyz" }`). This is needed to ensure properties are also parsed and handled correctly (the `__Property__` class isn't available from `foam.lookup()`, so the `FObjectParser` crashes if left as is).

src/foam/lib/json/FObjectArrayParser.java
TanayParikh: The `ExprParser` consists of an `Alt` with the two options being an `FObjectParser` and a custom parser for properties (`{ class: "__Property__", forClass_: "xyz" }`). This is needed to ensure properties are also parsed and handled correctly (the `__Property__` class isn't available from `foam.lookup()`, so the `FObjectParser` crashes if left as is).

src/foam/dao/AbstractDAO.java
TanayParikh: Updated, **possible breaking change**. Fixed one internal issues, could impact other users.

https://github.com/foam-framework/foam2/pull/558/commits/d6d001cd5c50b9cc6157f27587de787f59862059

src/foam/lib/json/Outputter.java
TanayParikh: Updated.

src/foam/lib/json/AnyParser.java
kgrgreer: Then, add a comment to explain that here or else add each of those Alt choices here directly.

src/foam/lib/json/FObjectArrayParser.java
kgrgreer: Add comment.

src/foam/lib/json/AnyParser.java
TanayParikh: So remove the existing `ExprParser`, make a new parser for `__Property__` and then replace L:21 with `new Alt(new __PropertyParser__(), new FObjectParser());`?

Or add the `Alt` choices in the comment?

kgrgreer: Yes, except L:21 is already an AltParser, so you don't need another one. Just call it PropertyParser.

src/foam/dao/AbstractDAO.java
```diff
@@ -58,6 +58,8 @@ protected Sink decorateSink_(Sink sink, long skip, long limit, Comparator order,
       sink = new PredicatedSink(predicate, sink);
     }
 
+    ((AbstractFObject) sink).setX(x);
```
adamvy: What do we need this for?

Can we change the previous "new FooSink()" lines to read "x.create(FooSink.class)" instead?

src/foam/dao/SinkJava.js
```diff
@@ -166,7 +166,7 @@ foam.CLASS({
     {
       name: 'eof',
       javaCode: 'if ( getArray() == null ) setArray(new java.util.ArrayList());\n'
-                + 'java.util.Collections.sort(getArray());\n'
+                + 'java.util.Collections.sort(getArray(), comparator_);\n'
```
adamvy: Use getComparator() instead.  It's preferable to access properties through the getFoo() interface, even if they are local properties so we don't skip any additional getter logic that FOAM properties may provide.

src/foam/dao/AbstractDAO.java
TanayParikh: This is mainly to stay consistent with the other methods within the abstract class which accept a context. Ex. `pipe`, `removeAll_`.

src/foam/dao/SinkJava.js
TanayParikh: Updated, thanks

src/foam/dao/AbstractDAO.java
adamvy: I see.

I feel like we should be using x.create() in all those places rather than new Foo(), but I will get caught up on that code to see if we did it this way for any particular reason.

In this case in particular, setX() will only set the context on the latest decorated sink, but if we did a skipped, ordered, predicated sink, the skip and ordered Sinks will not have their context set.  I'm okay if we leave it as new Foo() but add the setX() calls after each new Foo() statement.

TanayParikh: I've updated accordingly for now. https://github.com/foam-framework/foam2/pull/558/commits/1af82c0f78fca2ed2c01401be2c53e0169a2cc47

kgrgreer: I don't want to incur the extra overhead of calling x.create() for internal functionality where I may never need it. Especially now that the Context hasn't been heavily optimized yet. Also, I'm not sure if I want people to redefine underlying predicates like EQ for security reasons.

adamvy: Went back if forth in my head for a while.  I get what you're saying.  Although in this context people couldn't redefine EQ() to have any ill effect since your PredicatedDAOs are already created, but they could redefine PredicatedSink.

In general I think things like DAOs which are both Context Aware and Context Oriented, should use the Context-Aware context for creating internal things, and not their context oriented context, which my suggestion would not be doing.

In this particular case, I think we should remove setX(), and also update OrderedSink to use "new foam.dao.Subscription()" rather than getX().create(foam.dao.Subscription.class); and everything should work for now.

src/foam/lib/json/FObjectArrayParser.java
TanayParikh: Added

src/foam/dao/AbstractDAO.java
TanayParikh: Updated.

PR #567, opened 2017-07-28
Sync DAO overhaul

mdittmer:
Many of these changes are aesthetic (updating documentation strings; following latest conventions for public/private method names, etc.). Largest functional changes are synchronization:

- `sync()` passes do not overlap; they wait for the previous pass to complete
- `delegate.put()/remove()` + `syncRecordDAO.put()` are transactional

These changes appear to fix issues identified in standalone SyncDAO stress test: `test/node/syncDAOStress.js`

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -140,6 +140,7 @@ FOAM_FILES([
   { name: "foam/dao/SequenceNumberDAO" },
   { name: "foam/dao/ContextualizingDAO" },
   { name: "foam/dao/VersionNoDAO" },
+  { name: "foam/dao/sync/SyncRecord" },
   { name: "foam/dao/SyncDAO" },
   { name: "foam/dao/EasyDAO" },
   { name: "foam/dao/NoSelectAllDAO" },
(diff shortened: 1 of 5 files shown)


mdittmer:
Another important change missed in description:

It is assumed that, after initial sync, `syncFromRemote_()` is only necessary when `polling`. Similarly, `remoteDAO$Proxy.sub(...)` only when **not** polling. I was prompted to make this change when I noticed that polling DAOs hooked up to a remote that supports push were getting updates twice.

mdittmer:
`test/node/syncDAOStress.js` does not currently test the `polling` case. This implementation fails to pick up other clients' `remove()`s when `polling`. I plan to implement that in a follow-up PR that will:

- Manage `VersionNoDAO` backlog with a proper `StoreAndForwardDAO`
- Add a `deletedProperty` to `VersionNoDAO` and `SyncDAO`
- Implement `VersionNoDAO.remove(All)_()` to save default record with only `id` and `deleted` set
- Implement `SyncDAO.syncFromRemote_()` to check `deletedProperty` on objects, then either `putFromRemote_()` or `removeFromRemote_()`

This leaves deleted records in the remote indefinitely, but ensures that clients can always sync.

mdittmer:
FYI: Above mentioned follow-up work on branch `sync-dao__next`

PR #570, opened 2017-07-31
SyncDAO/VersionNoDAO/StoreAndForwadDAO

mdittmer:
These changes are built on top of https://github.com/foam-framework/foam2/pull/567 (see it for notes on initial overhaul). Many changes to `SyncDAO` and friends:

- `SyncDAO` and `VersionNoDAO` both have `deletedProperty` for marking records as deleted
- `SyncDAO` property rename: `syncProperty` is now `versionProperty` for consistency with `VersionNoDAO` which is (loosely speaking) its server-side analog
- `VersionNoDAO` should be wrapped in a `StoreAndForwardDAO` if its initial delegate may be non-empty and/or operations may be sent to it immediately
- `VersionNoDAO` now converts `remove()` to `put(<empty-object-marked-as-deleted>)`; this allows polling sync clients to pick up `remove()`s from other clients

The value of these changes is that (according to `syncDAOStress` test) both polling and non-polling `SyncDAO`s stay in sync with multiple clients under moderate load.

### DIFF
--- a/src/foam/dao/SyncDAO.js
+++ b/src/foam/dao/SyncDAO.js
@@ -61,10 +61,20 @@ foam.CLASS({
     },
     {
       name: 'syncProperty',
+      of: 'Property',
       documentation: `The property to use to store the object version. This
           value on each object will be incremented each time it is put() into
           the SyncDAO.`,
       required: true,
+      hidden: true,
+      transient: true
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'Property',
+      name: 'deletedProperty',
+      required: true,
+      hidden: true,
       transient: true
     },
     {
@@ -285,7 +295,11 @@ foam.CLASS({
                     var promises = [];
 
                     for ( var i = 0 ; i < array.length ; i++ ) {
-                      promises.push(self.putFromRemote_(array[i]));
+                      if ( array[i][self.deletedProperty.name] ) {
+                        promises.push(self.removeFromRemote_(array[i]));
+                      } else {
+                        promises.push(self.putFromRemote_(array[i]));
+                      }
                     }
 
                     return Promise.all(promises);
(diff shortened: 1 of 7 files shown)


src/foam/dao/SyncDAO.js
```diff
@@ -16,248 +16,391 @@
...
+        var self = this;
+        var ret;
+        return self.withSyncRecordTx_(function() {
+          return self.delegate.remove(obj).then(function(o) {
+            ret = o;
+            // console.log('removeFromRemote_: store syncRecord for', foam.json.stringify(obj));
+            return self.syncRecordDAO.put(self.SyncRecord.create({
+              id: obj.id,
+              syncNo: obj[self.versionProperty.name],
+              deleted: true
```
jacksonic, line 252: This is putting a sync record for the removed object back in? Shouldn't this be after the remote and cache have dropped the object, so we can lose the sync record?

mdittmer: We don't want to remove the sync record. It represents a version number we have now seen. This is being updated after the server has marked the record as deleted and the cache has removed its copy. I'm not sure for what else we should wait.

jacksonic: Isn't this the time when syncRecords can get cleaned up? Why remember versioning for an item the server won't send us again (unless it is re-created, in which case the version bump server-side would hit us on the next poll)

mdittmer: I think what you propose would probably be okay. I'd have to run it through the stress test a few times to be more confident. One down side of that approach is that, until we see another non-deletion update, we will re-process this deletion on each sync in the polling case. Since we push our updates first, that shouldn't be a problem, but as I say, I want to think about it some more and run it through the stress test a few times before implementing the change.

jacksonic: External chat about this: yes, it should work to persist the last-seen id. A future PR can optimize the sync record storage.

jacksonic (approved):

PR #573, opened 2017-07-31
MongoDB DAO (Java)

TanayParikh:
- Implemented Java MongoDAO
- Implemented `getValueClass` to PropertyInfo to get the property's value class

### DIFF
--- a/src/foam/core/PropertyInfo.java
+++ b/src/foam/core/PropertyInfo.java
@@ -20,7 +20,7 @@ public interface PropertyInfo
 
   public boolean getTransient();
   public boolean getRequired();
-  public Class getPropertyClass();
+  public Class getValueClass();
   public String getName();
   public Object get(Object obj);
   public void set(Object obj, Object value);
--- a/src/foam/core/PropertyInfo.java
+++ b/src/foam/core/PropertyInfo.java
@@ -20,6 +20,7 @@ public interface PropertyInfo
 
   public boolean getTransient();
   public boolean getRequired();
+  public String getPropertyType();
   public String getName();
   public Object get(Object obj);
   public void set(Object obj, Object value);
--- a/src/foam/core/PropertyInfo.java
+++ b/src/foam/core/PropertyInfo.java
@@ -20,7 +20,7 @@ public interface PropertyInfo
 
   public boolean getTransient();
   public boolean getRequired();
-  public String getPropertyType();
+  public Class getPropertyClass();
   public String getName();
   public Object get(Object obj);
   public void set(Object obj, Object value);
--- a/src/foam/dao/MongoDAO.java
+++ b/src/foam/dao/MongoDAO.java
@@ -34,9 +34,12 @@ import static com.mongodb.client.model.Filters.*;
 import com.mongodb.client.result.DeleteResult;
 import static com.mongodb.client.model.Updates.*;
 import com.mongodb.client.result.UpdateResult;
+
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Arrays;
+import java.util.Map;
 
 
 public class MongoDAO
@@ -86,6 +89,7 @@ public class MongoDAO
 
     String collectionName = (String) x.get("collectionName");
     String collectionClass = (String) x.get("collectionClass");
+    Map<String, String> embeddedObjClasses = (Map<String, String>) x.get("embeddedObjClasses");
 
     if ( collectionName == null || collectionName.isEmpty() ) {
       throw new IllegalArgumentException("Invalid collection name in context. Please provide a string name.");
@@ -98,7 +102,7 @@ public class MongoDAO
 
     try {
       while (cursor.hasNext()) {
-        sink.put(createFObject(collectionClass, cursor.next()), null);
+        sink.put(createFObject(collectionClass, embeddedObjClasses, cursor.next()), null);
       }
     } finally {
       cursor.close();
@@ -107,12 +111,18 @@ public class MongoDAO
     return sink;
   }
 
-  private FObject createFObject(String cls, Document d) {
+  private FObject createFObject(String cls, Map<String, String> embeddedObjClasses, Document d) {
     JsonWriterSettings writerSettings = new JsonWriterSettings(JsonMode.SHELL, true);
     String jsonStr = d.toJson(writerSettings);
 
     // Trims initial `"_id" : ObjectId("[24 HEX Chars]"),`
     jsonStr = "{ class: \"" + cls + "\", " + jsonStr.substring(MONGO_OBJECT_PREFIX_LENGTH, jsonStr.length() - 1) + " }";
+
+
+    // TODO: Find better technique to add embedded object classes
+    for (Map.Entry<String, String> entry : embeddedObjClasses.entrySet()) {
+      jsonStr = jsonStr.replace("\"" + entry.getKey() + "\" : {", "\"" + entry.getKey() + "\" : { class: \"" + entry.getValue() + "\",");
+    }
 
     JSONParser parser = new JSONParser();
     parser.setX(EmptyX.instance());
--- a/src/foam/dao/MongoDAO.java
+++ b/src/foam/dao/MongoDAO.java
@@ -82,10 +82,13 @@ public class MongoDAO
   }
 
   @Override
-  public foam.dao.Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+  public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
     if ( sink == null ) {
       sink = new ListSink();
     }
+
+    Sink         decorated = decorateSink_(sink, skip, limit, order, predicate);
+    Subscription sub       = new Subscription();
 
     String collectionName = (String) x.get("collectionName");
     String collectionClass = (String) x.get("collectionClass");
@@ -102,11 +105,19 @@ public class MongoDAO
 
     try {
       while (cursor.hasNext()) {
-        sink.put(createFObject(collectionClass, props, cursor.next()), null);
+        if ( sub.getDetached() ) break;
+
+        FObject obj = createFObject(collectionClass, props, cursor.next());
+        
+        if ( predicate == null || predicate.f(obj) ) {
+          decorated.put(obj, sub);
+        }
       }
     } finally {
       cursor.close();
     }
+
+    decorated.eof();
 
     return sink;
   }
--- a/src/foam/dao/MongoDAO.java
+++ b/src/foam/dao/MongoDAO.java
@@ -6,7 +6,6 @@
 
 package foam.dao;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Arrays;
 import java.util.Iterator;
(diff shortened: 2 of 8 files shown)


src/foam/lib/json/StringParser.java
```diff
@@ -16,10 +16,22 @@ public StringParser() {
     escape_ = '\\';
   }
 
+  public PStream isNull(PStream ps) {
+    // Is there a neater way to do this?
+    if ( ps.head() == 'n' &&
```
TanayParikh: Is there a neater way to do this?

src/foam/dao/MongoDAO.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    Sink         decorated = decorateSink_(sink, skip, limit, order, predicate);
+    Subscription sub       = new Subscription();
+
+    String collectionName = (String) x.get("collectionName");
+    String collectionClass = (String) x.get("collectionClass");
+    List props = getFObjectProperties(collectionClass, x);
+
+    if ( collectionName == null || collectionName.isEmpty() ) {
```
adamvy: Just out of curiostiry, why are collectionName and collectionClass coming from the context rather than being properties?

src/foam/dao/MongoDAO.java
```diff
@@ -0,0 +1,176 @@
...
+    if ( sink == null ) {
+      sink = new ListSink();
+    }
+
+    Sink         decorated = decorateSink_(sink, skip, limit, order, predicate);
+    Subscription sub       = new Subscription();
+
+    String collectionName = (String) x.get("collectionName");
+    String collectionClass = (String) x.get("collectionClass");
+    List props = getFObjectProperties(collectionClass, x);
```
adamvy: Most DAOs have an 'of' property which represents the class that they are a DAO of.

We should follow that pattern in this DAO as well.

src/foam/lib/json/StringParser.java
TanayParikh: Cleaned 👍

src/foam/lib/json/StringParser.java
```diff
@@ -16,10 +16,22 @@ public StringParser() {
...
+    }
+
+    return null;
+  }
+
   public PStream parse(PStream ps, ParserContext x) {
     delim_ = ps.head();
     if ( delim_ != '"' && delim_ != '\'' ) {
-      return null;
+      return isNull(ps);
```
adamvy: But what if I really did want to store the string "null" for a string property?

src/foam/dao/MongoDAO.java
TanayParikh: This was to allow the same DAO to access multiple collections of the same database. If that isn't of too much use, I can refactor to make collectionName a property and collectionClass would be using the DAO's `of` property.

src/foam/lib/json/StringParser.java
TanayParikh: Then it would be `"null"`, whereas here it only checks for `null` (the if condition would be false)

src/foam/dao/MongoDAO.java
adamvy: That would be of some use, but as a result the MongoDAO would behave differently than other DAOs.  It would be better to be consistent so someone looking to use a MongoDAO doesn't have to learn a different way of accessing using the DAO.

src/foam/dao/MongoDAO.java
```diff
@@ -0,0 +1,176 @@
...
+      }
+
+      return props;
+    } catch (Exception ex) {
+      throw new RuntimeException(clsName + " was not found.");
+    }
+  }
+
+  private FObject createFObject(String clsName, List props, Document d) {
+    JsonWriterSettings writerSettings = new JsonWriterSettings(JsonMode.SHELL, true);
```
adamvy: This method will be really really slow.  You're creating a lot of garbage, first by serializing the Mongo Document to JSON, and then appending a bunch more strings together to build a FOAM JSOn string, and then parsing that back to a FOAM object.

This is being called for every single object enumerated in a .select() call, meaning we're creating a lot of garbage in a tight loop, that will be bad for performance.


A better approach would be to create the FObject first with X.create(getOf()), and then set its properties individually as you enumerate the fields of the document.

You also may need to do a little bit of serialization/deserialization work by hand here for certain property values that MongoDB can't store out of box.

src/foam/dao/MongoDAO.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    throw new UnsupportedOperationException("put_ is not yet supported for MongoDAO");
+  }
+
+  @Override
+  public FObject find_(X x, Object id) {
+    throw new UnsupportedOperationException("find_ is not yet supported for MongoDAO");
```
adamvy, line 219: It might be easier to start with put_ and find_ together rather than select_(), that way you can iterate on proper serialization for a single object to/from MongoDB, after which you should be able to add enumeration without too much effort.

src/foam/dao/MongoDAO.java
```diff
@@ -0,0 +1,176 @@
...
+    // Trims initial `"_id" : ObjectId("[24 HEX Chars]"),`
+    jsonStr = "{ class: \"" + clsName + "\", " + 
+                 jsonStr.substring(MONGO_OBJECT_PREFIX_LENGTH, jsonStr.length() - 1) + 
+              " }";
+
+    Iterator i = props.iterator();
+
+    while ( i.hasNext() ) {
+      PropertyInfo prop = (PropertyInfo) i.next();
+      jsonStr = jsonStr.replace("\"" + prop.getName() + "\" : {", "\"" + prop.getName() + "\" : { class: \"" + prop.getPropertyType() + "\",");
```
adamvy: I don't think prop.getPropertyType() is necessary.  And it may lead to bugs.  When an FObjectProperty is serialized, the FObject it contains should serialize class: '' into the JSON and that's the class we should use when deserializing.

Consider this example

```
foam.CLASS({
  name: 'Abc',
  properties: ['a']
});

foam.CLASS({
  name: 'Def',
  extends: 'Abc',
  properties: ['b']
});

foam.CLASS({
  name: 'SomeClass',
  properties: [
  { class: 'FObjectProperty', of: 'Abc', name: 'foo' }
  ]
});
```

If I have an instance of SomeClass and I see foo = Def.create({ a: 1, b: 2 });  After saving and re-loading this instance from your DAO, I will instead get back foo == Abc.create({ a: 1 }), the fact that I had a Def instance stored will be lost.

src/foam/dao/MongoDAO.java
TanayParikh: That makes sense, thanks, updated.

src/foam/dao/MongoDAO.java
TanayParikh: Updated accordingly.

src/foam/dao/MongoDAO.java
TanayParikh: I see what you mean, thanks for the explanation. In that case, what would be a good technique to get the `ClassInfo` of a particular `FObjectProperty`?

src/foam/dao/MongoDAO.java
TanayParikh: Discussed, N/A - (Just `select_` is needed for now).

src/foam/dao/MongoDAO.java
TanayParikh: I need this so I can create and populate nested FObjects.

adamvy: Okay I understand this better.

I'm not wild about it returning a string that we then use to lookup the actual class.  What if instead we return an actual Class object.  We can even use this for primitive types as I just discovered, int.class or java.lang.Integer.TYPE is a thing in Java now.

In most cases people will want the actual Class object IMO, and if they need the name they can do getName().

src/foam/lib/json/DateParser.java
```diff
@@ -12,7 +12,27 @@
   extends ProxyParser
 {
   public DateParser() {
-    super(new Alt(new LongParser(), 
+    super(new Alt(new LongParser(),
+                  new Seq1(1,
+                    new Literal("ISODate("),
```
adamvy: If this is only to accomodate existing data in MongoDB, I'd rather these parser here not be modified, and custom parser be added to a different package like foam.dao.mongo or something, or even just inside the MongoDAO.

Once the MongoDB data is migrated to a new FOAM DAO, these parsers won't be needed I believe.

src/foam/lib/json/StringParser.java
```diff
@@ -19,7 +19,7 @@ public StringParser() {
   public PStream parse(PStream ps, ParserContext x) {
     delim_ = ps.head();
     if ( delim_ != '"' && delim_ != '\'' ) {
-      return null;
+      return (new Literal("null", "")).parse(ps, x);
```
adamvy: This will create a new literal every time we try to parse a string.  I also don't think we should be transforming a literal null into empty string.  If some JSON contains "foo": null, then I expect later getFoo() on my object to return null.

At the very least it's not the responsibility of the StringParser to make these decisions.

src/foam/dao/MongoDAO.java
TanayParikh: I've updated the method to parse the BSON document directly instead of the previous technique. This leverages the internal mongodb-driver APIs, and doesn't require the JSON parsers at all. Note, I've reverted the updates to the StringParser and DateParser, but kept the update to the BooleanParser (as that may be useful elsewhere as well). 

I've implemented it such that it creates an object from context, and then parses the BSON Document fields, setting the value of individual properties. This will also handle nested objects well.

src/foam/dao/MongoDAO.java
TanayParikh: Thanks, I've updated accordingly.

src/foam/lib/json/DateParser.java
TanayParikh: Removed as the MongoDAO no longer uses JSON parsers.

src/foam/lib/json/StringParser.java
TanayParikh: Removed as the MongoDAO no longer uses JSON parsers.

src/foam/lib/json/BooleanParser.java
```diff
@@ -12,6 +12,8 @@
   public BooleanParser() {
     super(new Alt(
                   new Literal("true", true),
-                  new Literal("false", false)));
+                  new Literal("false", false),
+                  new Literal("\"true\"", true),
```
adamvy: I'm still wary of these, if they ever get used in an context like Alt(Boolean(), String()), then it will return the boolean value true, if parsing a String "true", which is unlikely what you wanted in that case.

Can we leave these out until we have a concrete use case?

src/foam/core/PropertyInfo.java
```diff
@@ -20,6 +20,7 @@
 
   public boolean getTransient();
   public boolean getRequired();
+  public Class getPropertyClass();
```
adamvy: Sorry to be pedantic, but I'm not super happy with this name.  What about getValueClass() ?  getPropertyClass sounds like the class of the property, but what we we're returning is the class of the value we store in a given property.

src/foam/lib/json/BooleanParser.java
TanayParikh: Removed.

src/foam/core/PropertyInfo.java
TanayParikh: Sure, that sounds like it would be clearer. I've updated to `getValueClass()`. Thanks.

PR #574, opened 2017-07-31
Less invasive versioning \ diffbase \ SyncDAO/VersionNoDAO/StoreAndForwadDAO

mdittmer:
SyncDAO/VersionNoDAO to use generated "Versioned" classes

This change generates "versioned" classes with properties `version_` and `deleted_` for interal use with `SyncDAO` and `VersionNoDAO`. Currently, `VersionNoDAO` expects a versioned `of` (and does not automatically create one, but maybe it should?).

The value of this change is that client code outside of the `SyncDAO` can operate on classes that do not contain versioning information; this becomes an implementation detail hidden inside: (Sync DAO ==> ... VersionNoDAO => ... ==> <storage mechanism>).


mdittmer:
@jacksonic PTAL. I am not as happy with this as I thought I would be for two reasons:

1. Look at the stress test code. It feels like enough layers need to know about versioning that it is hardly worth keeping it out of the one layer that doesn't. Also, the developer now needs to keep track of which layers do and don't know about versioning. Maybe I just feel that way because what's being tested is the interface between versioned and unversioned objects; WDYT?
2. I have a use case that uses two layers of SyncDAO. I'm not sure whether this change is compatible with that use case. (Or at least, `SyncDAO` may need *another* change to detect that `of` is already versioned, and don't bother adapting client objects).

LMK what you think.

src/foam/dao/SyncDAO.js
```diff
@@ -112,6 +118,18 @@ foam.CLASS({
       factory: function() { return Promise.resolve(); }
     },
     {
+      name: 'versionedOf_',
+      expression: function(of) {
+        return this.versionedClassFactory_.get(of);
+      }
```
jacksonic: I think what's making this seem complicated is the inclusion of the versioning adapter here. If a DAO decorator was adapting the instances on the client end of SyncDAO, then everything in here could be implemented as before, and a facade or factory class could configure the parts of the sync system to use the versioned class. That would mean the versionedClassFactory would reside in the facade, using a generic TypeAdapterDAO + generic (but expecting intrinsic versioning) sync system

mdittmer: I agree. At the moment, I'm using this `SyncDAO` implementation (so I need to land it), but I do not depend on `SyncDAO` owning this adaptation (current usage is `of: theVersionedClass` in server <--> server interactions).

All that is to say, pulling this out to `TypeAdapterDAO` should be a prerequisite to integrating with `EasyDAO`, but I don't have time to implement it right now.

jacksonic: Offline discussion: yes, we'll refactor this stuff out in a future PR

jacksonic (approved):

mdittmer:
Manually merged to master: https://github.com/foam-framework/foam2/commit/5d61f913e0f9720ebb9f726d8633f47f3781a231

PR #578, opened 2017-08-02
Action command java

pavneetskapoor:
- Added `javaCode` support for actions in cases of business logic being executed on server side for specific actions. 
- Added `ActionCommand` for sending command including object and action method (javaCode) to be performed on object.
- Added `ActionCommandDAO` which listens for actionCommand instances and invokes the necessary methods. 

Currently the action methods generated in Java are all static and do not contain meta-information such as label, isEnabled() and such.


src/foam/core/ActionCommand.js
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+foam.CLASS({
+  package: 'foam.core',
```
adamvy: I don't think this is "core" enough to be in foam.core.

Can we move this to foam.nanos somewhere?

src/foam/dao/ActionCommandDAO.java
```diff
@@ -0,0 +1,39 @@
...
+import foam.core.ClassInfo;
+import foam.core.FObject;
+import foam.core.X;
+import java.lang.IllegalAccessException;
+import java.lang.NoSuchMethodException;
+import java.lang.SecurityException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class ActionCommandDAO
```
adamvy: Ditto, I think this can live in foam.nanos for now

src/foam/java/refinements.js
```diff
@@ -321,6 +321,34 @@ foam.CLASS({
...
+
+  methods: [
+    function buildJavaClass(cls) {
+      if ( ! this.javaCode ) return;
+
+      cls.method({
+        visibility: 'public',
+        name: this.name,
+        type: 'void',
+        static: true,
```
adamvy: I don't think the method should be static, we want to apply it to instances of the class we are generating.

adamvy (changes requested):

adamvy:
@kgrgreer Can you advise on the use case for this? I'm having trouble with the design because I don't know what were designing it for.

Do we want to execute actions on objects that are already live instances on the Java side?  Or do we want to execute actions on objects that live in a DAO on the server?  Or are we sending complete objects to the server upon which we want the server to execute actions?

src/foam/nanos/actioncommand/ActionCommand.js
```diff
@@ -0,0 +1,36 @@
...
+      name: 'object',
+      required: true
+    },
+    {
+      class: 'String',
+      name: 'actionName',
+      required: true
+    }
+  ]
+});
```
adamvy: Missing newline at end of file

src/foam/nanos/actioncommand/ActionCommandDAO.java
```diff
@@ -0,0 +1,40 @@
...
+      }
+    }
+    return null;
+  }
+
+  public Object cmd_(X x, Object obj) {
+    this.cmd(obj);
+    return null;
+  }
+}
```
adamvy: Missing newline at end of file

tools/classes.js
```diff
@@ -5,6 +5,7 @@
  */
 
 var classes = [
+  'foam.core.Action',
```
adamvy: I don't think we need this.  We're not using the actual foam.core.Action class on the java side anywhere.

src/foam/nanos/actioncommand/ActionCommandDAO.java
kgrgreer: @adamvy The idea is that the action would be applied to an object already stored in a DAO. So we could have action buttons in the GUI but when you press them the action is actually performed on the server. So in a GUI in a UpdateView you would have actions that you could trigger which would then be run on the server. The server would find the object and then call it's server-side Java action (which is really just a method there).

src/foam/comics/DAOUpdateControllerView.js
```diff
@@ -63,16 +63,11 @@ foam.CLASS({
...
-      this.startContext({data: this.data.obj})
-        .add(this.data.dao.of.getAxiomsByClass(foam.core.Action))
-      .endContext()
-      */
-      this.add(
-        this.DAOUpdateController.SAVE,
-        this.DAOUpdateController.CANCEL,
-        this.DAOUpdateController.DELETE,
-        this.DAOUpdateController.OBJ);
+      this.
```
kgrgreer: Can you put the comment that you deleted back.

src/foam/java/refinements.js
```diff
@@ -321,6 +321,33 @@ foam.CLASS({
...
+  refines: 'foam.core.Action',
+
+  properties: [
+    {
+      class: 'String',
+      name: 'javaCode'
+    },
+    {
+      class: 'String',
+      name: 'javaReturns'
```
kgrgreer: Is this used or required?

src/foam/nanos/actioncommand/ActionCommand.js
```diff
@@ -0,0 +1,36 @@
...
+foam.CLASS({
+  package: 'foam.nanos.actioncommand',
+  name: 'ActionCommand',
+
+  documentation: 'Command object containing an object and pending action to be done on it',
+
+  properties: [
+    {
+      class: 'FObjectProperty',
+      name: 'object',
```
kgrgreer: Should this be the whole object or just its 'id'? I thought we would just send the id.

src/foam/nanos/actioncommand/ActionCommandDAO.java
```diff
@@ -0,0 +1,42 @@
...
+      try {
+        // Calling method
+        Method action = actionObj.getClass().getDeclaredMethod(((ActionCommand)obj).getActionName());
+        action.invoke(actionObj);
+      } catch ( SecurityException | NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+
+      }
+      return true;
+    }
+    super.setDelegate(this);
```
kgrgreer: Replace next two lines with:

return super.cmd(obj);

src/foam/java/refinements.js
pavneetskapoor: This was added in because I was not sure whether the javaCode would allow for methods to be created that would return a specific type instead of being void all the time.

src/foam/nanos/actioncommand/ActionCommandDAO.java
adamvy: Okay that's what I thought, that's for the clairification

src/foam/nanos/actioncommand/ActionCommand.js
adamvy: Yeah this should just be the 'id' of the object.

src/foam/nanos/actioncommand/ActionCommandDAO.java
adamvy: *thanks.

So should this really be a DAO then?  Or I feel like this could be its own self contained service, where you give it a daoName, objectId and actionName, and it finds the object by ID and calls actionName.

src/foam/java/refinements.js
kgrgreer: For JS, Actions (like most implementations of the Command Design Pattern) don't return values.

src/foam/nanos/actioncommand/ActionCommand.js
```diff
@@ -0,0 +1,36 @@
+/**
```
kgrgreer, line 1: Wrong copyright header. Use the smaller non-Google one for code that NP contributes.

pavneetskapoor: Done

src/foam/nanos/actioncommand/ActionCommandDAO.java
kgrgreer: Since there would always be a 1:1 mapping between the services and the DAO's, it would be easier to just have the DAO be the service, via the cmd() method. Then the controller/JS-Action is assured to already have the DAO so that it can execute the remote Java actions. Also, different DAO's could be on different servers and the normal DAO routing/client-server support would handle this.

PR #579, opened 2017-08-02
Java Object to Primitive Casting Support

TanayParikh:
Necessary to avoid invalid cast errors from a nullable Object to a primitive.

### DIFF
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -435,10 +435,10 @@ foam.CLASS({
       var m = info.getMethod('cast');
       m.body = 'return ( o instanceof Number ) ?'
             + '((Number)o).intValue() :'
-            + '(int)o;';
+            + '((Integer) o).intValue();';
 
       var c = info.getMethod('comparePropertyToObject');
-      c.body = 'return compareValues((int) key, get_(o));';
+      c.body = 'return compareValues(cast(key), get_(o));';
 
       return info;
     }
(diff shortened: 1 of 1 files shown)


src/foam/java/refinements.js
```diff
@@ -435,10 +435,10 @@ foam.CLASS({
       var m = info.getMethod('cast');
       m.body = 'return ( o instanceof Number ) ?'
             + '((Number)o).intValue() :'
-            + '(int)o;';
+            + '((Integer) o).intValue();';
```
adamvy: Integer, Byte, Short etc are already instances of Number are they not?  When would the second case ever be hit?

TanayParikh: It wouldn't, however the compiler throws an error for attempting to cast a generic object to a primitive. @waseem-chamaa

TanayParikh: It wouldn't, however, the compiler throws an error for attempting to cast a generic object to a primitive.

TanayParikh: Though, this code has been around for a while, not sure why its causing an issue now.

adamvy: I'm not sure the old code was ever being triggered, it doesn't look like it should, if your parameter is Object java will automatically box primitive types for you.

I think we should just remove the conditional and do `((Number)o).intValue()` and similar for all cases.  No sense updating and keeping a conditional that literally never happens.

TanayParikh: Sounds good, removed the conditional.

PR #581, opened 2017-08-02
MongoDAO - Added NanoLogger for unknown property support

TanayParikh:
**Note:** Documents of unknown property/class still need to be parsed to ensure the parser remains in a valid state. As such null checks for class/prop are needed.

### DIFF
--- a/src/foam/dao/MongoDAO.java
+++ b/src/foam/dao/MongoDAO.java
@@ -204,7 +204,7 @@ public class MongoDAO
         break;
 
       default:
-        System.out.println(reader.getCurrentBsonType() + " parsing is not yet implemented.");
+        logger.error(reader.getCurrentBsonType(), "parsing is not yet implemented.");
     }
 
     return value;
--- a/src/foam/dao/MongoDAO.java
+++ b/src/foam/dao/MongoDAO.java
@@ -18,6 +18,8 @@ import foam.core.ClassInfo;
 import foam.mlang.order.Comparator;
 import foam.mlang.predicate.Predicate;
 
+import foam.nanos.logger.NanoLogger;
+
 // FObject JSON Parsing
 import foam.core.EmptyX;
 import foam.lib.json.JSONParser;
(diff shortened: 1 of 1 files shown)


src/foam/dao/MongoDAO.java
```diff
@@ -112,8 +115,8 @@ public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Pre
     return sink;
   }
 
-  private FObject createFObject(X x, BsonDocumentReader reader, Class cls) {
-    FObject obj = (FObject) x.create(cls);
+  private FObject createFObject(X x, BsonDocumentReader reader, Class cls, NanoLogger logger) {
+    FObject obj = ( cls == null ) ? null : (FObject) x.create(cls);
```
adamvy, line 119: If cls is null that should probably be a hard error.

src/foam/dao/MongoDAO.java
```diff
@@ -185,7 +196,11 @@ private Object getValue(X x, BsonDocumentReader reader, Class cls) {
...
       case DOCUMENT:
-        value = createFObject(x, reader, cls);
+        value = createFObject(x, reader, cls, logger);
+        break;
+
+      case OBJECT_ID:
+        value = reader.readObjectId();
         break;
 
       default:
```
adamvy, line 206: Use logger for the default case here as well.

src/foam/dao/MongoDAO.java
TanayParikh: This was to handle the case when the mongo field name read in isn't recognized. In that case the value still needs to be read (and ignored) in order to ensure the parser state is maintained. `cls` can be null in the case the unknown field is a nested document. Here, if `cls` is null, it continues parsing through the nested document and just ignores the values (thus `cls` wouldn't be needed).

The `cls` we care about would've been fetched from the DAO's `of` property on line 103.

src/foam/dao/MongoDAO.java
TanayParikh: Updated, thanks.

src/foam/dao/MongoDAO.java
adamvy: Ah I understand, sounds good to me.

PR #584, opened 2017-08-03
Multi-transaction BatchMutationDatastoreDAO

mdittmer:
Bulk updates to Datastore are excruciatingly slow with 25-record transactions in series. This change add support for a configurable number of multiple parallel write transactions. Of course, the config parameter should be set to 1 if the order of mutations matters.


mdittmer:
@jacksonic did you get a chance to look at this when you looked at the transactional IDBDAO? The core of the logic is the same.

jacksonic (approved):

Issue #585, opened 2017-08-03
IE11 broken by missing method names


mdittmer:
Loading FOAM in IE11 fails right now because `foam.LIB()` (and I suspect `foam.CLASS()`) cannot handle IE11's non-naming of functions. In IE11:

```js
(function foo() {}).name // => undefined
```

Right now, code expects `methodObject.name` to be defined (either because it was set with `{name: 'foo', code: ...}` or provided by Javascript on `function foo() { ... }`).

PR #586, opened 2017-08-03
Ie11 fix

mdittmer:
Towards #585

@kgrgreer and @adamvy PTAL. We have clients that need IE11 support. Some of these fixes are non-template-string usage of ES6/ES2015 features. Isn't our policy not to use such features? We may need a PSA.

### DIFF
--- a/src/foam/blob/Blob.js
+++ b/src/foam/blob/Blob.js
@@ -574,7 +574,7 @@ foam.CLASS({
     function put(file) {
       var id = this.nextId++;
       this.blobs[id] = file;
-      return Promise.resolve(this.IdentifiedBlob.create({ id }));
+      return Promise.resolve(this.IdentifiedBlob.create({ id: id }));
     },
     function find(id) {
       return Promise.resolve(this.blobs[id] || null);
--- a/src/foam/core/lib.js
+++ b/src/foam/core/lib.js
@@ -70,19 +70,6 @@ foam.assert = function assert(cond) {
 };
 
 
-foam.getMethodName = (function named() {}).name === 'named' ?
-    function(method) { return method.name;} :
-    function(method) {
-      if (typeof method !== 'function') return method.name;
-
-      // IE11 does not support named functions. Extract name from f.toString().
-      var match = method.toString()
-          .match(/^function\s+([A-Za-z_$][0-9A-Za-z_$]*)\s*\(/);
-      foam.assert(match, 'Unable to deduce method name from function');
-      return match[1];
-    };
-
-
 /**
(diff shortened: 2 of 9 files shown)


src/foam/blob/Blob.js
```diff
@@ -574,7 +574,7 @@ foam.CLASS({
     function put(file) {
       var id = this.nextId++;
       this.blobs[id] = file;
-      return Promise.resolve(this.IdentifiedBlob.create({ id }));
+      return Promise.resolve(this.IdentifiedBlob.create({ id: id }));
```
adamvy, line 577: lol oops

src/foam/core/lib.js
```diff
@@ -70,6 +70,19 @@ foam.assert = function assert(cond) {
 };
 
 
+foam.getMethodName = (function named() {}).name === 'named' ?
```
adamvy: I think this should be part of the foam.Function library.

What if we move this to the stop of stdlib.js and define

foam.LIB({
  name: 'foam.Function',
  methods: [ { name: 'getName', code: ... } ]
});

mdittmer:
On Aug 3, 2017 1:35 PM, "adamvy" <notifications@github.com> wrote:

*@adamvy* commented on this pull request.
------------------------------

In src/foam/core/lib.js
<https://github.com/foam-framework/foam2/pull/586#discussion_r131209989>:

> @@ -70,6 +70,19 @@ foam.assert = function assert(cond) {
 };


+foam.getMethodName = (function named() {}).name === 'named' ?

I think this should be part of the foam.Function library.

What if we move this to the stop of stdlib.js and define

foam.LIB({
name: 'foam.Function',
methods: [ { name: 'getName', code: ... } ]
});

The trouble with that is that foam.LIB needs it.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/foam-framework/foam2/pull/586#pullrequestreview-54167491>,
or mute the thread

adamvy:
Mark Dittmer <notifications@github.com> writes:

> The trouble with that is that foam.LIB needs it.

foam.LIB only needs it when defining methods in the short form.  The
long form takes the name from the plain map, as per my example.

mdittmer:
I see what you mean.

jacksonic:
If there's no other comments here I'll merge this so we can update the users seeing the compatibility issues.

mdittmer:
@jacksonic please merge this so we can ship to those users. @kgrgreer PTAL and LMK if we should have a PSA about features we do and don't use. Perhaps we should have a doc for that too?

PR #589, opened 2017-08-08
Symmetric parser/generator patterns for boxes

mdittmer:
Also implements a "real JSON" (as apposed to "FON") parser/generator.

The purpose is to support use cases such as:

```
var ctx = foam.box.Context.create(...);
ctx.parser = ctx.generator = foam.json.JSONParserGenerator.create(null, ctx);
```

and get native JSON parsing/stringifying
outputter.objectify()/foam.json.parse(). This performs better for some use
cases and eases publishing JSON APIs.


src/foam/core/JSON.js
```diff
@@ -523,6 +570,30 @@ foam.LIB({
...
+      outputDefaultValues: false,
+      // TODO: No deserialization support for shortnames yet.
+      //      useShortNames: true,
+      useShortNames: false,
+      strict: false,
+      propertyPredicate: function(o, p) { return ! p.networkTransient; }
+    }),
+
+    // JSON (not FON): Short, but exclude storage-transient properties.
+    Storage: foam.json.JSONParserGenerator.create({
```
adamvy: You're overriding the existing Storage option, I assume you meant StorageJSON ?

src/foam/box/HTTPBox.js
```diff
@@ -26,8 +26,9 @@ foam.CLASS({
   ],
 
   imports: [
+    'generator? as ctxGenerator',
```
adamvy: Boxes are not generally interoperable with existing systems.  Why do we need the ability to inject other serializations into boxes?

src/foam/box/MessagePortBox.js
```diff
@@ -27,25 +27,33 @@ foam.CLASS({
     'foam.box.Message'
   ],
 
-  imports: [ 'messagePortService', 'me' ],
+  imports: [
+    'generator? as ctxGenerator',
```
adamvy: Same question, also I'm not a fan of the name "generator" in the context.  It's a serializer or marshaller.  Generator has other connotations in comp sci.

src/foam/core/JSON.js
mdittmer: I did! Thanks.

src/foam/box/HTTPBox.js
mdittmer: For example, I've verified that native JSON parsing performs better when storing a cache in a worker and using `MessagePortBox` to access the DAO on the main thread. I cannot do that unless I can choose `JSONParserGenerator`.

src/foam/box/MessagePortBox.js
mdittmer: I'd be fine with a rename of "generator". I was thinking of the "parser/generator" jargon, but it is true that generator means something else in other CS contexts.

You have a preference? Serializer?

mdittmer: Maybe stringifier? stringify() is the only method that importers use.

src/foam/box/HTTPBox.js
adamvy: If your new JSON parser can properly serialize/deserialize all serializable FObjects properly and also performs faster than we should switch to it or we should optimize FON to be faster.

I don't like the idea of adding yet another knob, especially if that knob turns between "correct" and "fast."

If you really need a special serialization case in your application then I would replace MessagePortBox/Service with a custom class that uses your own serialization, only in your application.

mdittmer: I would have to subclass all <transport>Box/Service classes to do that. I would also risk having my code break if implementation details of the base class change underneath me. All I want to do is customize serialization, not the entire method that happens to invoke it (which right now includes special handling of RegisterSelfMessage, etc.).

I am fine with switching to `JSONParserStringifier` instead, but I assumed that others might now be. There are definitely trade-offs to making such a change. E.g., JSON is more verbose (so if bandwidth is at a premium, maybe FON is better); also, I don't know whether FON or JSON may have different relative performance for large vs. small payloads.

How do you feel about switching to `JSONParserStringifier`-equivalent logic @adamvy and @kgrgreer?

mdittmer: Not sure this was clear from my comment above. The knob is not "correct" or "fast". the `JSONParserStringifier` is correct (and fast). It's just using `parse(JSON.parse(str))` / `JSON.stringify(objectify(o))` rather than `parse(parseString(str))` / `stringify(objectify(o))` [1].

[1] `foam.json`'s parse/stringify is "effectively" `parse(parseString(str))` / `stringify(objectify(o))`, but is not implemented with separation between the two layers in both directions.

src/foam/core/JSON.js
```diff
@@ -459,6 +497,57 @@ foam.CLASS({
...
+  properties: [
+    {
+      documentation: `Map of {<class id>: true} for allowed classes`,
+      name: 'allowClassMap',
+      value: null
+    },
+    {
+      name: 'creationContext',
+      expression: function(allowClassMap) {
+        var baseCtx = this.__subContext__;
```
adamvy: Could you just reuse the ClassWhitelistContext I did for FON?

src/foam/core/JSON.js
```diff
@@ -523,6 +612,30 @@ foam.LIB({
...
+
+    // JSON (not FON): Short, but exclude network-transient properties.
+    NetworkJSON: foam.json.JSONParserStringifier.create({
+      pretty: false,
+      formatDatesAsNumbers: true,
+      outputDefaultValues: false,
+      // TODO: No deserialization support for shortnames yet.
+      //      useShortNames: true,
+      useShortNames: false,
+      strict: false,
```
adamvy: So in the base JSON Outputer.  strict:true was meant to output proper JSON.  Did you test if that works for you?

If we take all the flags of the standard Network config but also turn on strict: true, can we then parse it with JSON.parse?

src/foam/box/HTTPBox.js
adamvy: > I would have to subclass all Box/Service classes to do that.

No you would only have to subclass the two you want to replace (MessagePortService and MessagePortBox) and then use the context to replace which ones get created throughout the rest of the box system.

> All I want to do is customize serialization,

I know it may seem like you want a simple thing, but every knob we add the framework adds complexity and overhead for all other organizations that want to use it.  It's much better if the framework only has one canonical way to solve particular problems.  If every organization added knobs and customization hooks for their use cases it would be unmaintainable.

> is correct (and fast).

If it's both correct and fast then we could consider moving to it.  However your serialization approach right now creates unnecessary garabage by doing objectify and then JSON.stringify.  If you turn strict:true on in the existing serializers it is supposed to output proper json that you could use JSON.parse on, and this wouldn't require the intermediate objectify() step, which will hoepfully create less garbage.

> [1] foam.json's parse/stringify is "effectively" parse(parseString(str)) / stringify(objectify(o))

stringify doesn't go through an intermediate objectify step, it goes straight from the source object to the string.  objectify() was added for browser APIs that need plain old javascript objects rather than strings of JSON.

mdittmer: > No you would only have to subclass the two you want to replace (MessagePortService and MessagePortBox) and then use the context to replace which ones get created throughout the rest of the box system.

I don't see how I can use the context to make `(Raw)Socket(Box|Service)` and `(Raw)WebSocket(Box|Service)` use this serialization strategy. I would need to subclass them too if the strategy isn't customizable.

> every knob we add the framework adds complexity and overhead for all other organizations that want to use it.

In my view, adding `JSONParserStringifier` does add a second way of doing serialization (and, in that sense, complexity and/or cognitive overhead). However, making serialization customizable only adds complexity to those who care to customize it. Any users of the framework happy with the default `parser` and `serializer` don't even need to know that they can be customized. I think a case can be made for *not* including `JSONParserStringifier` in the framework more so than *not* allow users to choose a different strategy.

> If you turn strict:true on in the existing serializers it is supposed to output proper json that you could use JSON.parse on, and this wouldn't require the intermediate objectify() step, which will hoepfully create less garbage.

Good point. Perhaps `strict: true` on the stringifier side with a native `JSON.parse` on the parser side is the sweet spot here. The use case that uncovered performance issues showed them on the parsing side (though I haven't measured the stringifier side). Do you think this approach would work?

src/foam/core/JSON.js
mdittmer: (Somewhat duplicated from long comment above). Assuming no bugs in `strict: true` mode, that should work. Do you think that's an acceptable strategy? Perhaps we could then use a subclass to override `foam.json.Strict.parseString()` rather than introducing a new parser/stringifier?

mdittmer:
@adamvy @kgrgreer **nudge**. Should we be sorting out a VC or face-to-face to finalize JSON design modifications?

mdittmer:
I gathered more data that convinced me that we should switch to `foam.json`'s `parse`/`objectify` + `JSON`'s `parse`/`stringify`. I haven't tested other browsers yet, but here are timing results for parse/stringify 10k records on Chrome:

**Stringify**
`foam.json.Strict` ~ 750ms
`foam.json.Network` ~ 825ms
`JSON.stringify + foam.json.objectify` ~ 300ms

**Parse**
`foam.parsers.FON` ~ 4000ms
`foam.json.parseString` ~ 500ms
`foam.json.parse + JSON.parse` ~ 375ms

Each test run was done on its own page in Chrome. Code is here: https://github.com/foam-framework/foam2/commit/a63e530f9acdb72e5432252bf85d0d95a0897bab

For this data set, using non-strict FON saves ~15% space. (Switching to `JSON`wholesale would mean we no longer support non-strict `FON`).

Measuring GC time precisely is tedious (adding up little bits of sampled stack trace). However, my general impression is that using `JSON` (as I propose) does tend toward fewer *slightly* longer GC pauses. It looks like total time spent on GC is lower with `JSON` than with current methods, though.

I feel that this in combination with my observations from `ScrollDAOView` (which maintains a better frame rate with native  a `JSON` strategy) captures parsing small and large chunks.

mdittmer:
... All that said, this PR was *supposed* to be about the customization of box serialization (not an overhaul of serialization in general). @adamvy how would you feel about me modifying this to strip out `JSONParserStringifier`, and simply support custom `parser` and `stringifier` for boxes? That way I can implement my own parse/stringify strategy downstream until framework-level decisions are made.

adamvy:
I still have an issue with putting stringifier in the context.  In particular because there two key different means of serializing that are not distringuished between in your setup.

There is network serialization and storage serialization, and the lower level boxes need to be sure they're picking up the right one.

Just putting "stringifier" in the context risks DAOs being written against that as well and starting to serialize objects for storage using "network" serialization.

I think we need to put networkSerializer and storageSerializer in the context.  Or we should apply the factory pattern, and take advantage of the fact that we can already override .create(), and update the boxes to something like this

```
requires: [ 'foam.json.Outputter', 'foam.json.Parser' ],
this.Outputter.create({ network: true }) // or storage: true
this.Parser.create({ ... })
```

Then you can override Outputter in your context to build your own particular Outputter.

I favour the second approach because it just feels to me like the right thing to do.  With this setup there's no risk to multiple serializations sharing state by accident.  If a particular box wants to change some flags on the outputter it won't affect all other outputters being used by other boxes.  If you share a single outputter/parser then all client code must be careful not to mutate the outputter, or be extra careful to mutate and restore its state in a single synchronous action.

Downside of the second approach is we'll create a few extra objects for every object that needs to do serialization, but I don't think this will be a large amount of objects, and its something that can be optimized later using singletons and .freeze() or copy-on-write objects.

adamvy:
It should probably even be named foam.serialization.Outputter/Parser since its not necessarily JSON.  The JSON name is a bit more historical accident.

mdittmer:
Excellent point about mutating shared Outputter/Parser. I'll take a shot at implementing the (second) pattern you mentioned. I agree that the number of boxes performing serialization is probably small enough that the memory impact shouldn't be prohibitive.

I think I'll either open an issue or start a -discuss thread on switching to native `JSON`.

src/foam/core/JSON.js
adamvy: ping

mdittmer:
@adamvy PTAL. I switched to using factories in boxes that do parsing. The part that is still shared between boxes is a parser `creationContext` which `foam.box.Context` exports. That way a common whitelist can be used in each `Context`, but boxes/services select their own (de)serialization strategy.

src/foam/core/JSON.js
mdittmer: New implementation is passed a `creationContext`, usually provided by a `foam.box.Context`(which will use `ClassWhitelistContext` if it has a whitelist).

src/foam/core/JSON.js
```diff
@@ -459,6 +459,33 @@ foam.CLASS({
...
+    }
+  ],
+
+  methods: [
+    function parseString(str, opt_ctx) {
+      return this.strict ?
+          // JSON.parse() is faster; use it when data format allows.
+          foam.json.parse(JSON.parse(str), null,
+                          opt_ctx || this.creationContext) :
+          foam.json.parseString(str, null, opt_ctx || this.creationContext);
```
adamvy: This case should use FON.  parseString is unsafe

mdittmer:
@adamvy switched to safe `FON` parser in non-strict case of `foam.json.Parser.parseString()`. I used an expression to attempt to re-use a parser when no new creation context is injected at call-time.

mdittmer:
Last call on this PR. I'll probably merge tomorrow if no one objects.

PR #590, opened 2017-08-08
ScrollDAOView refactor

mdittmer:
Highlights:

- Decouple data fetch and row recycling. This allows for occassionally fetching
  large batches while aggressively updating scroll row anchor.

- Use data => htmlMarkupString function (a `RowFormatter`) to replace row
  contents immediately (no U2 Elements). This means that rows cannot respond
  to U2 events (at least, not via U2), but it significantly improves scroll
  performance.

### DIFF
--- a/src/foam/u2/view/ScrollDAOView.js
+++ b/src/foam/u2/view/ScrollDAOView.js
@@ -47,104 +47,25 @@ foam.CLASS({
     'foam.dao.QuickSink',
     'foam.u2.ViewSpec'
   ],
+  // Provide most state to inner controller and views.
   exports: [
     'anchorDAOIdx_',
     'anchorRowIdx_',
+    'batchSize',
+    'count_',
+    'data',
     'modNumRows_',
     'negativeRunway',
     'numRows',
-    'rows_'
-  ],
-
-  classes: [
-    {
-      name: 'Row',
-      extends: 'foam.u2.Element',
-
-      axioms: [
-        foam.u2.CSS.create({
-          code: function CSS() {/*
-            ^ {
-              display: block;
-              contain: layout;
-              will-change: transform;
-              padding: 5px;
-              box-sizing: border-box;
-            }
-            ^ * {
-              background-color: #dddddd;
-            }
-        */}
-        })
-      ],
-
-      properties: [
-        [ 'nodeName', 'li' ],
-        {
-          class: 'FObjectProperty',
-          of: 'foam.u2.Element',
-          name: 'view'
-        }
-      ],
-
-      methods: [
-        function initE() { this.addClass(this.myClass()).add(this.view$); }
-      ]
-    },
-    {
-      name: 'FetchSink',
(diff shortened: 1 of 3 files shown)


src/foam/u2/view/ScrollDAOView.js
```diff
@@ -341,9 +245,273 @@ foam.CLASS({
...
+        })
+      ],
+
+      properties: [
+        [ 'nodeName', 'li' ],
+        {
+          name: 'data',
+          postSet: function(old, nu) {
+            if ( this.state !== this.LOADED ) return;
+            this.el().innerHTML = this.rowFormatter.format(nu);
```
adamvy, line 284: So this is dangerous as hell.  One of the reasons we don't do anything like this in U2 was for security approval, because developers cannot be trusted to properly escape HTML every time. Just like they can't be trusted to properly manage memory.  (See endless security bugs in any C++ project, even Chrome).

I really _really_ don't think you should be side-stepping U2.  If you must, please do it in your own project and not the core framework.

adamvy: It may be possible to use U2 in a "formatter" like setup if we can't find a way to optimize U2 for your use case.  U2 elements have an output method, so you can use them to build an HTML string by first constructing the virtual DOM using U2 and then outputting where you need.  But that final output should be handled internally by the system like it currently is so there's no way for an external user to just provide their own HTML strings.

Although you should be getting the same performance as if you just added a slot to your U2 elements, and update that slot to a new element everytime you need to render a large chunk of HTML.  That should be roughly the same speed as just replace innerHTML, and if its not yet it should be able to be optimized to be nearly the same speed.

mdittmer: Re: formatter using U2. That was actually what I implemented first, but it was too slow. I'm also quite unhappy that this is what I had to resort to for good performance (with U2 as-is today).

I would be thrilled to see the U2 output cycle (for U2 formatters) or slot bindings (for "plain U2 views") optimized enough to get rid of this. I think it's a useful view to expose through the framework. Unfortunately, I don't have the resources to spend on U2 optimization right now. (Not that I'm the best person for the job anyway; I was never "the U2 expert".)

mdittmer: I'll dig up the demo code I was using to test performance so that anyone interested in experimenting with other strategies has the same benchmark I was using.

mdittmer: @adamvy FYI: I filed https://github.com/foam-framework/foam2/issues/598 with a link to demo code that performs as fast as I was able to make it. U2-based row formatters and/or U2 views with slot bindings did not keep the demo above 30fps.

adamvy: I still object to this being included in core framework code.  Some unsuspecting user is going to use it improperly, get a XSS vulnerability, and harm the reputation of the framework.

Please move ScrollDAOView to your own project.

PR #591, opened 2017-08-08
Many-to-many grid DAO

mdittmer:


jacksonic (commented):
My main question is whether this is a general purpose DAO, since it's always forcing the 'of' type. Should it be located in U2 or under a ./relationships dir?

mdittmer:
Yeah. I think it should probably be moved to a different package. I've also been contemplating generating code for its rows so that each `Boolean` is its own indexable property, but that introduces a fair bit of complexity. Probably do a follow-up PR to restructure around that.

mdittmer:
@jacksonic I moved them to a `grid` sub-package. Anything else on this change? I'm going to muck with code generation of `Boolean` properties in a separate PR.

PR #595, opened 2017-08-14
Customizable Radio Button

kennyqiyenkan:
Radio View’s radio buttons can now be customized using CSS using the after and before selector.

[Tutorial](https://www.templatemonster.com/blog/style-checkboxes-radio-buttons-css/)


src/foam/u2/view/RadioView.js
```diff
@@ -20,10 +20,27 @@ foam.CLASS({
   name: 'RadioView',
   extends: 'foam.u2.view.ChoiceView',
 
+  axioms: [
+    foam.u2.CSS.create({
+      code: function CSS() {/*
+        ^{ }
```
adamvy: I don't think you need to define an empty CSS class.  You can still apply this.myClass() to your elements and just have

src/foam/u2/view/RadioView.js
```diff
@@ -20,10 +20,27 @@ foam.CLASS({
...
+
+        ^ label {
+          position: relative;
+        }
+
+        .foam-u2-RadioContainer { }
```
adamvy: These are really generic names why are we defining them here?  What if someone adds a foam-u2-RadioContainer class or a foam.u2.Label ?

adamvy: Also again we don't need to define empty classes

src/foam/u2/view/RadioView.js
```diff
@@ -37,23 +54,34 @@ foam.CLASS({
   listeners: [
     function onChoicesUpdate() {
       var self = this;
+      var id;
 
       this.removeAllChildren();
 
       this.add(this.choices.map(function(c) {
         return this.E('div').
+          addClass('foam-u2-RadioContainer').
```
adamvy: I think it would be better to use the this.myClass() pattern here still, it will still be a name we can customize from outside, but then there' no risk of it accidentally conflicting with another class, especially for the Label class below.

The <label> element can already be addressed using  "^ label"

src/foam/u2/view/RadioView.js
kgrgreer: What's the reason for the three empty CSS rules?

src/foam/u2/view/RadioView.js
```diff
@@ -37,23 +54,34 @@ foam.CLASS({
...
               self.data = evt.srcElement.value;
             }).
           end().
-          add(c[1]);
+          start('label').
+            attrs({
+              for: id
+            }).
+            start('span').
+            addClass('foam-u2-Label').
```
kgrgreer: Indent this line and the next since they're inside a start().

PR #596, opened 2017-08-14
Changed "type" to "javaType" in foam.java.Constant for consistencies sake

kirkeaton:


adamvy:
This is more inconsistent.  This is already a java constant, foam.java.Constant.

We don't prefix everything else in foam.java, with java.  Like foam.java.Method just has "type" not javaType.

foam.java.Constant represents a Java constant.  Same as foam.java.Method represents a java method which has a type and a name etc etc.

kirkeaton:
@adamvy That's why I closed and deleted the branch

adamvy:
Ahaha oops sorry.  I thought it was closed and merged, not closed and deleted.

Carry on :)

PR #599, opened 2017-08-15
Composite Logger

Akenne:


Akenne:
@kgrgreer Here is the new pr

src/foam/nanos/logger/StdoutLogger.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Override
+    public StringBuilder get() {
+      StringBuilder b = super.get();
+      b.setLength(0);
+      return b;
+    }
+  };
+
+  public void start() {
```
kgrgreer, line 42: Is this actually being called?

src/foam/nanos/logger/StdoutLogger.java
```diff
@@ -0,0 +1,111 @@
...
+import java.sql.Timestamp;
+import java.text.SimpleDateFormat;
+import java.io.IOException;
+
+public class StdoutLogger
+  implements Logger
+{
+  protected java.util.logging.Logger logger;
+
+  private static final ThreadLocal<SimpleDateFormat> sdf = new ThreadLocal<SimpleDateFormat>() {
```
kgrgreer: Make all private's be protected since it makes testing and debugging easier.

src/foam/nanos/logger/CompositeLogger.java
```diff
@@ -0,0 +1,60 @@
...
+import foam.nanos.logger.FileLogger;
+import foam.nanos.logger.StdoutLogger;
+import foam.nanos.logger.Logger;
+import java.util.ArrayList; 
+import java.util.List; 
+import java.util.logging.*;
+import java.io.IOException;
+
+public class CompositeLogger
+implements foam.nanos.logger.Logger
```
kgrgreer: indent 2 spaces

src/foam/nanos/logger/FileLogger.java
```diff
@@ -41,10 +40,9 @@ public StringBuilder get() {
   };
 
   public void start() {
```
kgrgreer, line 42: Is this ever called?

src/foam/nanos/logger/CompositeLogger.java
kgrgreer: Make also implement NanoService so that when start()'ed it can also start its delegates which are NanoServices.

src/foam/nanos/logger/StdoutLogger.java
Akenne: yes in the services file ( and fileoutlogger )

StdoutLogger stdOut = new StdoutLogger(); 
stdOut.start(); 

FileLogger fileOut = new FileLogger(); 
fileOut.start(); 

return log.add(stdOut).add(fileOut);

kgrgreer: If I used one of these loggers directly by specifying the serviceClass I would expect them to be automatically started, like the other NanoServices. If the implemented the NanoService interface, which only has the start() method, then when instantiated directly as services, they would be automatically started. The CompositeLogger could check if each of its delegates implements NanoService and start them when it is itself started.

src/foam/nanos/logger/CompositeLogger.java
Akenne: The delegates aren't nanoservices anymore. they just have a start function. You told me that logger does not need to be a nanoservice and all of them implement logger

nothing is a nanoservice in this pull request

src/foam/nanos/logger/StdoutLogger.java
Akenne: So should Logger extend nanoservice? And then also stdout/fileoutLoggers extend implement nanoservice? 

Or is it enough for stdout/fileoutLoggers to implement logger and then logger extends nanoservice

And for compositelogger, the delegates are added after composite logger is already created, so wouldn't start have already been called? What I mean is that when start is being run for composite logger, it is not possible for it to have any delegates at that time? Or do I understand the start() wrong

Akenne:
@kgrgreer I pushed the new changes

PR #605, opened 2017-08-22
Fixed bug with FOAM_FILES and java generation

kirkeaton:
Implemented a fix that allows for the generation of java code as well as loading FOAM files in browser.


TanayParikh:
@kgrgreer

PR #606, opened 2017-08-22
AuthenticatedUserDAO, WebAuth updated

MarcRoopchand:


src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -0,0 +1,60 @@
...
+public class AuthenticatedUserDAO
+  extends ProxyDAO
+{
+  public AuthenticatedUserDAO(DAO delegate) {
+    setDelegate(delegate);
+  }
+
+  @Override
+  public FObject put_(X x, FObject fObject) {
+    if ( fObject instanceof User ) {
```
kgrgreer: If isn't necessary, just cast in DAO's assuming correct type. Will throw ClassCastExcpetion if it isn't, which is correct behaviour.

src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -0,0 +1,60 @@
...
+      User user = (User) fObject;
+
+      if ( super.find(user.getId()) != null ) {
```
kgrgreer: Better to call super.find_(x, user.getID() so context is preserved.

PR #607, opened 2017-08-23
Sets a default 'of' property of FObject if missing

waseem-chamaa:


src/foam/nanos/http/ClassesWebAgent.java
```diff
@@ -0,0 +1,34 @@
...
+package foam.nanos.http;
+
+import foam.core.*;
+import foam.dao.DAO;
+import java.net.URL;
+import java.net.URLClassLoader;
+import foam.nanos.boot.NSpec;
+import java.io.PrintWriter;
+
+public class ClassesWebAgent
```
kgrgreer: This doesn't match the description of the PR.

waseem-chamaa: that was my mistake i was in the wrong branch when putting it will update the description

waseem-chamaa: fixed

PR #609, opened 2017-08-24
Some UI fixes on prop views

mcarcaso:
Makes CheckBox, EnumView, and IntView respect the visibility of the property that they're views of. Also change the default value of CheckBox's showLabel to be true if a label
is present instead of always true.

### DIFF
--- a/src/foam/u2/CheckBox.js
+++ b/src/foam/u2/CheckBox.js
@@ -30,7 +30,7 @@ foam.CLASS({
     {
       class: 'Boolean',
       name: 'showLabel',
-      value: true
+      expression: function(label) { return !!label },
     },
     { 
       class: 'String', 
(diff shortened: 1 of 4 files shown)


src/foam/u2/CheckBox.js
```diff
@@ -30,7 +30,7 @@ foam.CLASS({
     {
       class: 'Boolean',
       name: 'showLabel',
-      value: true
+      expression: function(label) { return !!label },
```
kgrgreer: Couldn't we just check that the label is defined below when we do if ( this.showLabel ), since it doesn't help that this is an expression, since we're not going to go back and re-eval the 'if' anyway?

mcarcaso: Good point. Alternatively, I could add a slot on that property to initE so it handles the property changing. Do you have a preference?

kgrgreer: Just change this to a factory. Exact same result, but less expensive. Then the user can set showLabel explicitly if they want.

mcarcaso: Done

PR #612, opened 2017-08-24
DIFFBASE(#611): Serialize ArraySink according to "of" class

mdittmer:


src/foam/core/types.js
```diff
@@ -306,8 +306,8 @@ foam.CLASS({
         var c = this.instance_[prop.name];
 
         // Implement value and factory support.
-        if ( ! c ) {
-          if ( prop.value ) {
+        if ( foam.typeOf(c) === foam.Undefined ) {
```
kgrgreer: foam.Undefined.isInstance(c) is much much faster.

src/foam/dao/ArrayDAO.js
```diff
@@ -32,8 +32,8 @@ foam.CLASS({
       class: 'Class',
       name: 'of',
       factory: function() {
-        if (this.array.length === 0) return this.lookup('foam.core.FObject');
-        return this.array[0].cls_;
+        if ( this.array.length === 0 ) return this.lookup('foam.core.FObject');
+        return undefined;
```
kgrgreer: Factories aren't supposed to return undefined.

src/foam/dao/daoUtils.js
```diff
@@ -113,10 +113,40 @@ foam.CLASS({
   name: 'ArraySink',
   extends: 'foam.dao.AbstractSink',
 
+  constants: {
+    // Dual to outputJSON method.
+    PARSE_JSON: function(json, opt_cls, opt_ctx) {
```
kgrgreer: A better name would be parseJSON, since it's really a static method rather than an actual constant.

kgrgreer: How is defining PARSE_JSON for a class different than just implementing a custom initArgs(), which is also responsible for controlling how the object is built from JSON?

src/foam/dao/ArrayDAO.js
mdittmer: Is there a safe alternative in this case? Should I use `null`?

src/foam/dao/daoUtils.js
mdittmer: Re 1st comment: I would generally agree, but last I checked (which was, admittedly, a long time ago) constant definitions get CONST_CASE'd regardless of how you name them in the `constants` object. I used this naming convention for consistency between definition and invocation.

Re 2nd comment: It may be no different. I haven't used the strategy of implementing a custom `initArgs()` before. The advantage of having `parseJSON()` is the dual naming `outputJSON`/`parseJSON` in the algorithms in `JSON.js` that invoke them.

mdittmer: Also, I think there is a subtle difference between `parseJSON` and `initArgs`. The former is intended to co-operate with `foam.json.parse` to *deserialize* an instance of the class. The latter is intended to *initialize* an instance of the class.

I'm adding a `TODO` to turn `PARSE_JSON` into `parseJSON` once #613 is fixed.

mdittmer:
Plz comment on #619 (continued from this PR, but with `diffbase=master`)

PR #614, opened 2017-08-25
Pop Up Titled View

TanayParikh:
- Created a titled PopUpView
- Requires a specified `messageView` for content as well as `title`

### DIFF
--- a/src/foam/u2/view/PopUpTitledView.js
+++ b/src/foam/u2/view/PopUpTitledView.js
@@ -1,0 +1,235 @@
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'PopUpTitledView',
+  extends: 'foam.u2.View',
+
+  imports: [
+    'stack',
+    'closeDialog'
+  ],
+
+  documentation: 'Pop Up View',
+
+  properties: [ 
+     {
+       class: 'String',
+       name: 'title'
+     },
+     'messageView'
+   ],
+
+  axioms: [
+    foam.u2.CSS.create({
+      code: function CSS() {/*
+      ^ {
+        width: 448px;
+        margin: auto;
+      }
+
+      ^ .Message-Container {
+        width: 448px;
+        border-radius: 2px;
+        background-color: #ffffff;
+      }
+
+      ^ .Change-Container {
+        width: 448px;
+        height: 40.5px;
+        background-color: #14375d;
+      }
+
+      ^ .Change-Text {
+        width: 100;
+        height: 40px;
+        font-family: Roboto;
+        font-size: 14px;
+        line-height: 2.86;
+        letter-spacing: 0.2px;
+        text-align: left;
+        color: #ffffff;
+        margin-left: 19px;
+        display: inline-block;
+      }
+
+      ^ .mainMessage-Text {
+        height: 16px;
+        font-family: Roboto;
+        font-size: 14px;
+        font-weight: 300;
+        letter-spacing: 0.2px;
+        text-algin: left;
+        color: #093649;
+        margin-left: 20px;
+        margin-top: 19.5px;
+        margin-right: 64px;
+        margin-bottom: 10px;
+      }
+
+      ^ .close-Button {
+        width: 24px;
+        height: 24px;
+        margin-top: 8.5px;
+        margin-right: 16px;
+        float: right;
+        cursor: pointer;
+      }
+
+      ^ .input-Box, .foam-u2-tag-Input, .foam-u2-tag-Select {
+        height: 40px;
+        width: 408px;
+        background-color: #ffffff;
+        border: solid 1px rgba(!64, 179, 184, 0.5);
+        margin-left: 20px;
+        margin-right: 20px;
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/PopUpTitledView.js
```diff
@@ -0,0 +1,235 @@
...
+          }))
+        .end()
+      .end()
+    }
+  ],
+
+  actions: [
+    {
+      name: 'closeButton',
+      icon: 'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij4gICAgPGcgZmlsbD0ibm9uZSIgZmlsbC1ydWxlPSJldmVub2RkIiBzdHJva2U9IiNGRkYiIHN0cm9rZS1saW5lY2FwPSJyb3VuZCIgc3Ryb2tlLXdpZHRoPSIyIj4gICAgICAgIDxwYXRoIGQ9Ik01IDVsMTQuMTQyIDE0LjE0Mk0xOSA1TDUgMTkiLz4gICAgPC9nPjwvc3ZnPg==',
```
kirkeaton, line 235: Why base64 instead of just a path to an image?

TanayParikh: There aren't any other `svg`'s committed to FOAM, other images have been `base64` from what I recall.

src/foam/u2/view/PopUpTitledView.js
```diff
@@ -0,0 +1,241 @@
+/** 
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'PopUpTitledView',
```
kgrgreer: Rename PopUp to Popup to be consistent with other uses of the word.

src/foam/u2/view/PopUpTitledView.js
```diff
@@ -0,0 +1,241 @@
...
+      .end()
+    }
+  ],
+
+  actions: [
+    {
+      name: 'closeButton',
+      icon: 'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij4gICAgPGcgZmlsbD0ibm9uZSIgZmlsbC1ydWxlPSJldmVub2RkIiBzdHJva2U9IiNGRkYiIHN0cm9rZS1saW5lY2FwPSJyb3VuZCIgc3Ryb2tlLXdpZHRoPSIyIj4gICAgICAgIDxwYXRoIGQ9Ik01IDVsMTQuMTQyIDE0LjE0Mk0xOSA1TDUgMTkiLz4gICAgPC9nPjwvc3ZnPg==',
+      code: function (X) {
+        X.closeDialog();
```
kgrgreer, line 237: Is this needed? Can't it just close itself?

src/foam/u2/view/PopUpTitledView.js
TanayParikh: Updated

src/foam/u2/view/PopUpTitledView.js
TanayParikh: From the docs, it seems like clicking the background closes itself. 

https://github.com/foam-framework/foam2/blob/ca06f9e4873d5bd4ee6a81d773f95114e97f5506/src/foam/u2/dialog/Popup.js#L23-L26

However this if for the `X` close button on the top right. Please let me know if there's another way.

TanayParikh:
@kgrgreer Is this good to merge?

src/foam/u2/view/PopUpTitledView.js
```diff
@@ -0,0 +1,241 @@
...
+
+    this
+      .addClass(this.myClass())
+      .start()
+        .start().addClass('Message-Container')
+          .start().addClass('Change-Container')
+            .start().addClass('Change-Text').add(self.title).end()
+            .start()
+              .add(self.CLOSE_BUTTON)
+              .addClass('close-Button')
```
kgrgreer, line 221: Is adding a class here required? Aren't actions automatically assigned meaningful classes that you could use instead?

src/foam/u2/view/PopUpTitledView.js
kgrgreer: Who is exporting closeDialog and why do they need to? Can't the window just close itself without relying on some parent to do it? I think that this was meant to be more of a notification that the window was closed in case you're interested, like 'dialogClosed', but the dialog should just remove itself.

kgrgreer:
Added two comments.

PR #619, opened 2017-08-29
Serialize ArraySink according to "of" class

mdittmer:
Same as #612 with `diffbase=master`


src/foam/core/JSON.js
```diff
@@ -586,6 +586,9 @@ foam.LIB({
 
           if ( cls ) {
             var c = typeof cls === 'string' ? ( opt_ctx || foam ).lookup(cls) : cls;
+            // TODO(markdittmer): Turn into static method: "parseJSON" once
+            // https://github.com/foam-framework/foam2/issues/613 is fixed.
+            if ( c.PARSE_JSON ) return c.PARSE_JSON(json, opt_class, opt_ctx);
```
kgrgreer, line 591: How is PARSE_JSON different that just redefining initArgs?

mdittmer: https://github.com/foam-framework/foam2/pull/612#discussion_r135806161

mdittmer:
@kgrgreer **ping**

adamvy:
Why is this necessary?  This is a lot of code specific to ArraySink that I will now have to replicate into java.

Shouldn't default serialization of Arrays just work out of box and not require custom code?

PR #623, opened 2017-08-29
Drop getCaller in favour of explicit FOAM_FLAGS.src configuration

mdittmer:


kgrgreer:
This is breaking travis.

PR #625, opened 2017-08-30
Auth user dao

MarcRoopchand:
Restricted find()
Added regex for email, password validation


src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -60,4 +56,34 @@ public FObject put_(X x, FObject fObject) {
...
   }
+
+  @Override
+  /**
+   * Restrict find method to only return the logged in user object
+   * */
+  public FObject find_(X x, Object id) {
+    User user = (User) x.get("user");
+
+    if ( user == null ) {
```
kgrgreer: I think this should check if user == null or id != user.getId(). Otherwise someone could try to lookup someone else and think they've succeeded.

src/foam/nanos/boot/Boot.java
```diff
@@ -38,6 +38,9 @@ public void put(FObject obj, Detachable sub) {
     // Export the ServiceDAO
     ((ProxyDAO) root_.get("nSpecDAO")).setDelegate(serviceDAO_);
 
+    DAO localUserDAO = (DAO) root_.get("localUserDAO");
```
kgrgreer: I don't think this should be here.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -225,6 +234,122 @@ public static String generateRandomSalt() {
     return saltBuilder.toString();
   }
 
+  //Used regex recommended here http://www.ex-parrot.com/pdw/Mail-RFC822-Address.html
+  public static boolean validateEmail(String email) {
```
kgrgreer: I suspect that this is very slow to compile every time. Add a TODO to benchmark and then reuse the pattern if benchmark warrants it.

PR #628, opened 2017-08-30
Xml modifications

kirkeaton:
1. Removed outer <object class=""> tags
2. Removed <value> tags for non FObject arrays
3. Simplified setting of properties.

### DIFF
--- a/src/foam/core/XML.js
+++ b/src/foam/core/XML.js
@@ -210,10 +210,12 @@ foam.CLASS({
         Array:     function(v, p) {
           for ( var i = 0; i < v.length; i++ ) {
             if ( foam.core.FObjectArray.isInstance(p) ) {
+              // output FObject array
               this.start('<' + this.propertyName(p) + '>');
               this.output(p.toXML(v[i], this));
               this.end('</' + this.propertyName(p) + '>');
             } else {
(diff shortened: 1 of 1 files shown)


src/foam/core/XML.js
```diff
@@ -337,89 +335,52 @@ foam.CLASS({
...
+      // create DOM
+      var parser = new DOMParser();
+      var doc = parser.parseFromString(str, 'text/xml');
+      var root = doc.firstChild;
+
+      var rootClass = root.getAttribute('class');
+      if ( rootClass ) {
+        // look up root class from XML tags
+        return this.objectify(root, foam.lookup(rootClass));
+      } else if ( opt_class ) {
```
kgrgreer: No need for the 'else' if you've already returned in the above if.

src/foam/core/XML.js
```diff
@@ -337,89 +335,52 @@ foam.CLASS({
...
+        // lookup class if given a string
+        if ( typeof(opt_class) === 'string' )
+          opt_class = foam.lookup(opt_class);
+        return this.objectify(root, opt_class);
+      } else {
```
kgrgreer: No need for 'else'.

Issue #613, opened 2017-08-24
Add static/class methods to Model


kgrgreer:
Rather than having to use constants: to add class methods. Should support java.

kgrgreer:
@hchoura

kgrgreer:
Implemented in Static.js

PR #629, opened 2017-08-31
foam.box.node.ForkBox: Support app module loading and secure parsing

mdittmer:
UPDATE: Switch to a smaller change regarding app loading. Allow app to provide its own top-level script as before, but provide a static method (currently, a constant) for connecting to parent process.

OLD DESIGN (https://github.com/foam-framework/foam2/pull/629/commits/4f58d5fb6e123e7a40fa4d5688ef0fd4cd573db6):

This change deprecates `ForkBox.childScriptPath` (but does not yet remove
it). Instead, `ForkBox.appModule` names a script to require() to load app
code. The `appModule` is expected to export a `foam.box.Context` to be used
for IPC with the forked process. The `Context`'s `creationContext` is used on
the parser that reads bootstrap data from the forked process's `stdin`.


mdittmer:
@kgrgreer here's another use of constants-as-static-methods. I can switch them once #613 is closed.

mdittmer:
@arobins FYI: This API should enable safer behaviour in GoogleChrome/confluence

mdittmer:
FYI: Updated, less invasive design pushed (see updated PR description)

mdittmer:
Landing TBR @kgrgreer or delegate.

PR #630, opened 2017-08-31
Xml attributes

kirkeaton:
1. Added output / parsing of XML attributes.

### DIFF
--- a/src/foam/core/XML.js
+++ b/src/foam/core/XML.js
@@ -234,16 +234,16 @@ foam.CLASS({
           }
         },
         FObject: function(v, p) {
-          if ( v.value ) {
-            // if v.value exists then we have attributes
+          if ( v.xmlValue ) {
+            // if v.xmlValue exists then we have attributes
             // check if the value is an FObject and structure XML accordingly
-            if ( foam.core.FObject.isInstance(v.value) ) {
+            if ( foam.core.FObject.isInstance(v.xmlValue) ) {
               this.start('<' + this.propertyName(p) + this.outputAttributes(v) + '>');
               this.output(p.toXML(v, this));
               this.end('</' +  this.propertyName(p) + '>');
             } else {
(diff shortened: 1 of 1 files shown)


src/foam/core/XML.js
```diff
@@ -181,6 +186,16 @@ foam.CLASS({
       return this;
     },
 
+    function outputAttributes(v) {
+      var attributes = v.cls_.getAxiomsByClass(foam.core.Property).filter(function (p) { return p.xmlAttribute });
+      if ( attributes.length === 0 ) return this;
+
+      for ( var i = 0; i < attributes.length; i++ ) {
```
kgrgreer: spaces before ;'s

src/foam/core/XML.js
```diff
@@ -224,9 +239,24 @@ foam.CLASS({
           }
         },
         FObject: function(v, p) {
-          this.start('<' + this.propertyName(p) + '>');
-          this.output(p.toXML(v, this));
-          this.end('</' +  this.propertyName(p) + '>');
+          if ( v.xmlValue ) {
+            // if v.xmlValue exists then we have attributes
```
kgrgreer, line 243: Is this true? Aren't the use of attributes and a default xmlValue independent? Couldn't you have attributes with tags or a default xmlValue without attributes?

src/foam/core/XML.js
kirkeaton: Done

src/foam/core/XML.js
kirkeaton: I suppose you could have an xmlValue property without any attributes but can you think of the use case for that? You would be needlessly wrapping the true value in an extra class.

kgrgreer: <font size="+1">blah blah blah</font>, but I guess we're not going to be modelling HTML (actually, I suppose we do already model XML with U2), but we can leave it until we ever need it.

PR #632, opened 2017-08-31
Added outputDefinedValues flag

kirkeaton:


src/foam/core/XML.js
```diff
@@ -444,21 +453,24 @@ foam.LIB({
...
+      outputDefaultValues: false,
+      outputDefinedValues: true
     }),
 
     // Compact output (not pretty)
     Compact: foam.xml.Outputter.create({
       pretty: false,
       formatDatesAsNumbers: true,
       outputDefaultValues: false,
+      outputDefinedValues: true
```
kgrgreer: I think this should be false. Compact is meant to save space, so it doesn't make sense to output values which are set the the default.

src/foam/core/XML.js
```diff
@@ -444,21 +453,24 @@ foam.LIB({
...
     }),
 
     // Shorter than Compact (uses short-names if available)
     Short: foam.xml.Outputter.create({
       pretty: false,
       formatDatesAsNumbers: true,
       outputDefaultValues: false,
+      outputDefinedValues: true,
```
kgrgreer: Should also be false.

src/foam/core/XML.js
```diff
@@ -202,7 +207,11 @@ foam.CLASS({
 
     function outputProperty_(o, p) {
       if ( ! this.propertyPredicate(o, p ) ) return;
-      if ( ! this.outputDefaultValues && p.isDefaultValue(o[p.name]) ) return;
+      // don't output default values unless value is defined and outputDefinedValues set to true
+      if ( ( ! this.outputDefaultValues && p.isDefaultValue(o[p.name]) ) &&
+        ( this.outputDefinedValues && ! o.hasOwnProperty(p.name) ) ) {
+        return;
```
kgrgreer: I think this is easier to understand (and also more efficient):

if ( this.outputDefinedValues ) {
  if ( ! o.hasOwnProperty(p.name) ) return;
} else if ( this.outputDefaultValues ) {
  if ( p.isDefaultValue(o[p.name]) ) return;
}

PR #634, opened 2017-09-01
Stub: Support modeled 'typeName' in return type

mdittmer:


adamvy:
Where is typeName coming from?

mdittmer:
Re: `typeName`: It is used [here, for example](https://github.com/GoogleChrome/confluence/blob/cb584144f792d527aa29d3e8bb9ac184a64380b6/lib/web_apis/version_history.es6.js#L36) the same way as [FOAM2 uses it here](https://github.com/foam-framework/foam2/blob/a48d289f865fc73f7babb70b2dd2cb8c1fea8f87/src/foam/core/Argument.js#L81).

kgrgreer:
All FOAM2 usages: https://github.com/foam-framework/foam2/search?utf8=%E2%9C%93&q=typeName&type=

adamvy:
We really aught to model types at some point.

It looks like this is a fairly new addition you've added to have returns: {} be a map instead of a string.  It's the right direction, at some point I'd like returns to be a modeled Type object, but I'm worried about any other code making assumptions that returns is a string.

On the other hand I guess that doesn't apply to this PR as you're only updating Stub and not adding more uses of returns: {}.

mdittmer:
FWIW, I agree with the sentiment about modeling types; that would be great.
Care will be required for is-a-String assumptions.

PR #635, opened 2017-09-01
Fix enum (default) value serialization (was: Multiton: Key be default (or factory) value)

mdittmer:


src/foam/pattern/Multiton.js
```diff
@@ -57,7 +57,19 @@ foam.CLASS({
       cls.create = function(args) {
         var instances = this.private_.instances ||
             ( this.private_.instances = {} );
+
         var key = args[property];
+        if ( foam.Undefined.isInstance(key) ) {
+          var prop = cls[foam.String.constantize(property)];
```
adamvy: getAxiomByName would be more appropriate and obvious I think.

adamvy:
This feels like a lot of special casing, what is your use case?

mdittmer:
Context: This fixes a bug on Enums wherein the first value has the default `ordinal` value. When the enum value is marshaled to a different address space, it is not correctly Multiton'd to the existing Enum value, which messes up `foam.util.equals()` checks (which use object identity on FObjects by default).

adamvy:
So you don't actually need the factory support then?

Can we just fix this in enums?

adamvy:
Maybe just set the default value of ordinal to -1 so that it always gets serialized as it starts counting from 0.

mdittmer:
IMHO that addresses the symptom without addressing the cause of the problem. Default values and Multitons should not be incompatible. Neither should derivative default values (factories, expressions, etc.).

kgrgreer:
> IMHO that addresses the symptom without addressing the cause of the problem. Default values and Multitons should not be incompatible. Neither should derivative default values (factories, expressions, etc.).

I don't see how factories and expressions can be compatible with Multitons. A Multiton is a creational pattern, it happens before the object is created, whereas factories and expressions happen to existing objects. I can't create to object to know the factory/expression value before I've created the object. Can't we just set "value: -1" for ordinal?

mdittmer:
We could. Could we instead support default value, but not
factory/expression? Hacking ordinal leaves this as a footgun for any class
using a default value on a Multiton's property.

adamvy:
Using and/or relying on default value for a multiton feels like a bad idea IMO

mdittmer:
Switched to enum default value = -1 instead. @adamvy @kgrgreer PTAL.

mdittmer:
Filed https://github.com/foam-framework/foam2/issues/637

adamvy:
LGTM

PR #639, opened 2017-09-05
foam.web.URLState: Application state tracking via URL hash value

mdittmer:


src/foam/web/URLState.js
```diff
@@ -0,0 +1,236 @@
...
+      var first = true;
+      var value;
+      for ( var key in bindingsMap ) {
+        if ( ! bindingsMap.hasOwnProperty(key) ) continue;
+
+        if ( ! first ) hash += '&';
+        first = false;
+
+        value = this.serializer.stringify(bindingsMap[key].get());
+        hash += key + '=' + this.window.encodeURIComponent(value);
```
arobins: |key| should also be URI-encoded.

src/foam/web/URLState.js
```diff
@@ -0,0 +1,236 @@
...
+      }
+
+      var unboundMap = this.unboundMap_;
+      for ( var key in unboundMap ) {
+        if ( ! unboundMap.hasOwnProperty(key) ) continue;
+        value = this.serializer.stringify(unboundMap[key]);
+        hash += '&' + key + '=' + this.window.encodeURIComponent(value);
```
arobins: This breaks if there is nothing in bindingsMap.

src/foam/web/URLState.js
```diff
@@ -0,0 +1,236 @@
...
+        if ( ! first ) hash += '&';
+        first = false;
+
+        value = this.serializer.stringify(bindingsMap[key].get());
+        hash += key + '=' + this.window.encodeURIComponent(value);
+      }
+
+      var unboundMap = this.unboundMap_;
+      for ( var key in unboundMap ) {
+        if ( ! unboundMap.hasOwnProperty(key) ) continue;
```
arobins: This and the loop above are identical other than the input map. Maybe make a method for doing the conversion?

mdittmer: In order for `first` handling to be general, it would need to be input/output, so the helper method would return two things (the new string and the `first` state).

IMHO it's simpler like this (though repetitive). [If there were 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) or more loops, I would tend to agree: a method would be better.

arobins: Or just return a string, and let the caller deal with mashing them together. The logic for joining two potentially-empty strings is pretty trivial.

mdittmer: On second thought, I can use an output function to avoid requiring two return values. Updated logic with helper function is in `DetachedURLState.js`

src/foam/web/URLState.js
mdittmer: Indeed! Fixed.

src/foam/web/URLState.js
mdittmer: Indeed! Fixed.

src/foam/web/DetachedURLState.js
```diff
@@ -0,0 +1,212 @@
...
+      this.onStateChange();
+    },
+    function stateToHash_() {
+      var hash = '#' + this.path_ + '?';
+      var bindingsMap = this.bindingsMap_;
+      var first = true;
+      function out(str) { hash += str; }
+
+      // Output bindings, then unbound values.
+      this.outputMapToHashStr_(
```
arobins: I feel like this would be easier to follow if the helper accepted a current string and appended key/value pairs to it. You can replace the first flag with an empty string check, and return the string at the end instead of the flag.

mdittmer: Done. @arobins PTAL

arobins (commented):
lgtm, but I'd suggest waiting for someone else from FOAM to review.

PR #640, opened 2017-09-05
added parameters to User 

waseem-chamaa:


src/foam/nanos/auth/User.js
```diff
@@ -86,7 +86,7 @@
        name: 'profilePicture'
      },
      {
-      class: 'FObjectArray',
+      class: 'FObjectProperty',
```
kgrgreer: This should be
  class: 'FObjectArray',
  of: 'FObject'

kgrgreer:
@waseem-chamaa I accidentally merged this, but then reverted it. See my comment.

PR #655, opened 2017-09-11
Auth updates

MarcRoopchand:
Updated DAO and services to throw exceptions properly


kgrgreer:
RuntimeExceptions don't need to be caught. Better to throw a non-Runtime Exception like: https://docs.oracle.com/javase/7/docs/api/javax/naming/AuthenticationException.html, or make up our own.

PR #657, opened 2017-09-12
VersionNoDAO: Fix remove_() implementation

mdittmer:
To store "deleted" record, use existing.clone() instead of
create({ id: existing.id }) because some id implementations (such as
MultiPartID) have getters.


mdittmer:
TBR @jacksonic

jacksonic (approved):

PR #662, opened 2017-09-15
Nanos submenu

kennyqiyenkan:
Modified submenu to show under pressed element.


src/foam/nanos/menu/Menu.js
```diff
@@ -326,30 +359,35 @@ foam.CLASS({
...
       this.content = content;
     },
 
     function open() {
-      this.document.body.insertAdjacentHTML('beforeend', this.outerHTML);
-      this.load();
+      if ( this.parent ) {
+        this.parent.add(this);
+      } else {
+        this.document.body.insertAdjacentHTML('beforeend', this.outerHTML);
```
kgrgreer, line 388: Did you test if this case still works? Is it even possible actually? Do we still need it?

kgrgreer:
Check that you can add a s SubMenu to a SubMenu.

PR #673, opened 2017-09-25
PostgresDAO

kirkeaton:
Implemented select_, find_, remove_, and put_ functions for PostgresDAO. All functions are working as expected. Predicate support has not been implemented. If a table does not exist for a given model, then it will be created based on the properties.


src/foam/dao/SQLType.js
```diff
@@ -0,0 +1,64 @@
+foam.ENUM({
```
kgrgreer: I wouldn't make this an Enum because different databases have different type names and support a long-tail of specialized types, which can still be used by model authors if this is just a String value.

src/foam/core/AbstractPropertyInfo.java
```diff
@@ -79,4 +90,9 @@ public void toXML(FObject obj, Document doc, Element objElement) {
       objElement.appendChild(prop);
     }
   }
+
+  @Override
+  public SQLType getSqlType() {
+    return SQLType.VARCHAR;
```
kgrgreer: We should include the 'width' for String properties, otherwise becomes very inefficient.

src/foam/dao/SQLType.js
kirkeaton: Yeah good point. I have already run into issues related to this. PostgreSQL has different type names than say MySQL & SQL Server for things like byte arrays and date times.

kirkeaton: What would you suggest as an alternative?

src/foam/dao/EnabledAware.java
```diff
@@ -7,12 +7,15 @@
...
 import foam.mlang.predicate.Predicate;
+
+import java.sql.PreparedStatement;
 import java.util.Date;
 
 public interface EnabledAware {
   public final static Predicate ENABLED = new Predicate() {
     public boolean f(foam.core.FObject obj) { return true; }
-
+    public String createStatement(String table) { return ""; }
```
kgrgreer: Shouldn't this be "enabled = 't'"?

kirkeaton: I don't know. This was copied over from Carlos' PR

src/foam/dao/pg/ConnectionPool.java
```diff
@@ -0,0 +1,32 @@
+package foam.dao.pg;
+
+import org.apache.commons.dbcp2.BasicDataSource;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public class ConnectionPool {
```
kgrgreer: Make this a modelled object so it can be configured through the GUI. Then make it a service so it can be accessed from the Context and reused by DAO's.

src/foam/core/AbstractPropertyInfo.java
kirkeaton: Yeah good point. We would need some way to handle changes to the width property. Not sure how frequent that may be though.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,326 @@
...
+      } else {
+        stmt = c.prepareStatement(predicate.createStatement(table));
+      }
+
+      ResultSet resultSet = stmt.executeQuery();
+      while ( resultSet.next() ) {
+        sink.put(createFObject(resultSet), null);
+      }
+
+      resultSet.close();
```
kgrgreer: More care needed for exception handling and closing resources. If you get an exception you won't close your resultSet, stmt and connection, causing your connections to be drained from the thread-pool and blocking your whole server.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,326 @@
...
+  public FObject remove_(X x, FObject o) {
+    try {
+      Connection c = connectionPool.getConnection();
+      StringBuilder builder = sb.get()
+          .append("delete from ")
+          .append(table)
+          .append(" where id = ?");
+
+      PreparedStatement stmt = c.prepareStatement(builder.toString());
+      stmt.setLong(1, ((Number) o.getProperty("id")).longValue());
```
kgrgreer: Might not be a Long. Okay to handle it future PR, but add a TODO.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,326 @@
...
+  public FObject find_(X x, Object o) {
+    try {
+      Connection c = connectionPool.getConnection();
+      StringBuilder builder = sb.get()
+          .append("select * from ")
+          .append(table)
+          .append(" where id = ?");
+
+      PreparedStatement stmt = c.prepareStatement(builder.toString());
+      stmt.setLong(1, ((Number) o).longValue());
```
kgrgreer: Add TODO to support non-Numbers.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,326 @@
...
+
+      int index = 1;
+      PreparedStatement stmt = c.prepareStatement(builder.toString(),
+          Statement.RETURN_GENERATED_KEYS);
+      // set statement values twice: once for the insert and once for the update on conflict
+      index = setStatementValues(index, stmt, obj);
+      index = setStatementValues(index, stmt, obj);
+
+      int inserted = stmt.executeUpdate();
+      if (inserted == 0) {
```
kgrgreer: Fix spaces.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,326 @@
...
+   * @param resultSet
+   * @return FObject
+   * @throws SQLException
+   */
+  private FObject createFObject(ResultSet resultSet) throws Exception {
+    if ( getOf() == null ) {
+      throw new Exception("`Of` is not set");
+    }
+
+    List<PropertyInfo> props = getOf().getAxiomsByClass(PropertyInfo.class);
```
kgrgreer: You should cache a list of non-transient properties as an instance variable to avoid creating list every time.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,326 @@
...
+
+    int index = 1;
+    Iterator i = props.iterator();
+    while ( i.hasNext() ) {
+      // prevent reading out of bounds of result set
+      if ( index > metaData.getColumnCount() )
+        break;
+      // get the property and set the value
+      PropertyInfo prop = (PropertyInfo) i.next();
+      if ( prop.getStorageTransient() )
```
kgrgreer: Not needed if you use a pre-filtered cached list.

src/foam/dao/pg/PostgresDAO.java
kirkeaton: Done

src/foam/dao/pg/PostgresDAO.java
kirkeaton: Thanks. I modified all the closing of resources.

src/foam/dao/pg/PostgresDAO.java
kirkeaton: Done

src/foam/dao/pg/PostgresDAO.java
kirkeaton: Added the pre-filtered cached list.

src/foam/dao/pg/ConnectionPool.java
kirkeaton: What exactly am I modelling here?

src/foam/core/AbstractPropertyInfo.java
kgrgreer: We created a separate DAO in the system to record the version #'s of the models for each table in the database. If it changed we would check for a "Migration Strategy" which would be applied to migrate the data. Typically we would just stream the data to a file, rebuilt the table and then reload, but you could have other tables. Some database give bad performance after you've 'alter'-ed them. But I don't think we need to worry about this for now.

src/foam/dao/EnabledAware.java
kgrgreer: Yes, it should be ;-).

src/foam/dao/pg/ConnectionPool.java
kirkeaton: Okay I've modelled ConnectionPool and made it a NanoService. I'm fetching form the context in PostgresDAO now too.

src/foam/dao/SQLType.js
kgrgreer: Just store sqlTypes as Strings. Then you can also make the expressions if they include something like width.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,368 @@
...
+        sink.put(createFObject(resultSet), null);
+      }
+
+      return sink;
+    } catch (Exception e) {
+      e.printStackTrace();
+      return null;
+    } finally {
+      if ( resultSet != null )
+        try { resultSet.close(); } catch (SQLException ignored) {}
```
kgrgreer: Safer to catch Throwable.

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,368 @@
...
+        stmt = c.prepareStatement(predicate.createStatement(table));
+      }
+
+      resultSet = stmt.executeQuery();
+      while ( resultSet.next() ) {
+        sink.put(createFObject(resultSet), null);
+      }
+
+      return sink;
+    } catch (Exception e) {
```
kgrgreer: Catch Throwable.

kirkeaton: Modified all catch blocks to catch Throwables instead.

src/foam/dao/EnabledAware.java
kirkeaton: Okay. Added that in

src/foam/dao/SQLType.js
kirkeaton: Where am I storing those sqlTypes?

src/foam/dao/pg/PostgresDAO.java
```diff
@@ -0,0 +1,368 @@
...
+        // no rows
+        return null;
+      }
+
+      return createFObject(resultSet);
+    } catch (Exception e) {
+      e.printStackTrace();
+      return null;
+    } finally {
+      if ( resultSet != null )
```
kgrgreer: Since this pattern repeats, it could be a standard method called closeAll(resultSet, stmt, c).

src/foam/dao/pg/ConnectionPool.js
```diff
@@ -0,0 +1,91 @@
...
+    {
+      class: 'Object',
+      name: 'pool',
+      javaType: 'BasicDataSource',
+      documentation: 'Connection pool',
+    },
+    {
+      class: 'Int',
+      name: 'poolSize',
+      value: 4,
```
kgrgreer: 4 seems too small. A few long-running queries could starve the rest of the app. I don't know what the best value would be, but maybe 20 would be better.

src/foam/dao/pg/ConnectionPool.js
```diff
@@ -0,0 +1,91 @@
...
+    },
+    {
+      class: 'String',
+      name: 'password',
+      documentation: 'Database password for authentication'
+    },
+    {
+      class: 'String',
+      name: 'connectionString',
+      javaFactory: `return getPrefix() +
```
kgrgreer: What is the 'return' for?

kgrgreer: transient: true,

kirkeaton: Added

src/foam/dao/pg/ConnectionPool.js
kirkeaton: Changed to 20

src/foam/dao/pg/PostgresDAO.java
kirkeaton: Added a helper function

PR #674, opened 2017-09-25
Web Fetch API: Deal with missing headers API on Edge

mdittmer:


jacksonic (changes requested):
Could this be polyfilled on the Headers class itself?

mdittmer:
It could, but that would be more code, unless we have a spec-compliant Iterator implementation in FOAM already.

jacksonic (approved):

jacksonic:
That's a pain. Tell them to stop adding new collection interfaces; this isn't Python.

mdittmer:
I agree. This is `# Temporary I hope, hope, hope`. Presumably, Edge will fix the bug and we can remove this rather than investing in our own polyfill implementation.

PR #675, opened 2017-09-25
Swift support

mcarcaso:
It's a large one but anything in the swift dir doesn't need too close an eye because it's still likely to change. Just need to make sure the stuff I'm adding to core doesn't break anything.


kgrgreer:
@adamvy please pre-review before I do. Thanks.

demos/swift/SwiftTests/SwiftTests/AppDelegate.swift
```diff
@@ -0,0 +1,31 @@
+//
+//  AppDelegate.swift
+//  SwiftTests
+//
+//  Created by Michael Carcasole on 2017-02-22.
+//  Copyright © 2017 FOAM. All rights reserved.
```
kgrgreer: /**
 * @license
 * Copyright 2017 The FOAM Authors. All Rights Reserved.
 */

mcarcaso: Gone

demos/swift/gen_swift.js
```diff
@@ -0,0 +1,58 @@
...
+  'node': true,
+  'swift': true,
+  'debug': true,
+};
+
+var execSync = require('child_process').execSync
+
+var dir = __dirname;
+var root = dir + '/../..';
+var genDir = dir + '/gen';
```
adamvy: I think we should pass the output directory in as a command line parameter like we do for genjava.js

Also would be useful to take a classes.js file like we do for genjava.js

src/com/google/foam/demos/tabata/index.html
```diff
@@ -12,8 +13,8 @@
       var d = Tabata.create();
       foam.u2.DetailView.create({ data: d, showActions: true }).write();
 //      TabataBarCView.create({ data: d }).write();
-      TabataCView.create({ data: d }).write();
```
adamvy: Do we need to modify this demo?

src/files.js
```diff
@@ -92,6 +118,7 @@ FOAM_FILES([
   { name: "foam/java/Skeleton", flags: ['java'] },
   { name: "foam/java/JavaImport", flags: ['java'] },
   { name: "foam/java/refinements", flags: ['java'] },
+  { name: "foam/java/refinements", flags: ['swift'] }, // For createChildMethod_.
```
adamvy: Maybe add TODO: Move createChildMethod_ to core

src/foam/box/LocalBoxRegistry.js
```diff
@@ -63,14 +65,19 @@ foam.CLASS({
   methods: [
     {
       name: 'doLookup',
-      returns: 'foam.box.Box',
```
adamvy, line 66: Pretty sure we need this for stub generation, although maybe it doesn't matter cause its inherited from the interface?

src/foam/core/Listener.js
```diff
@@ -57,6 +57,9 @@ foam.CLASS({
 
   methods: [
     function installInProto(proto, superAxiom) {
+      // This can happen when there's no js implementation of a listener.
+      if (!this.code) return;
```
adamvy: FOAM style is if ( ! this.code )

src/foam/core/Method.js
```diff
@@ -137,6 +138,7 @@ foam.CLASS({
 
     function installInClass(cls, superMethod, existingMethod) {
       var method = this;
+      method.parentCls = cls;
```
adamvy: We should do this in createChildMethod_ which should clone the axiom if it's going to mutate it.

installInClass/installInProto shouldn't mutate the axiom IMO.

demos/swift/gen_swift.js
mcarcaso: Sounds good. Created a tool similar to genjava.js that this script now makes use of.

src/com/google/foam/demos/tabata/index.html
mcarcaso: Just the TabataState.js should be required. I'll undo the other changes.

src/files.js
mcarcaso: Moved that function into its own file that both swift and java make use of.

src/foam/box/LocalBoxRegistry.js
mcarcaso: That's right. It figures it out from the interface.

src/foam/core/Listener.js
mcarcaso: Done

src/foam/core/Method.js
mcarcaso: I agree with not wanting to mutate the axiom here but I need this applied to all methods, not just child methods. Do you know a better place for this?

demos/swift/js/somepackage/RequiredClass.js
```diff
@@ -0,0 +1,30 @@
+/**
```
kgrgreer, line 1: This is the wrong license header. Same for other files.

src/foam/core/Method.js
kgrgreer: Reusing axioms can happen with implements: but implemens already clones its axioms to avoid them being shared. What is the parentCls needed for?

mcarcaso: It's needed for figuring out if a method overrides a parent method because swift needs the 'override' keyword and it's also used for providing a default implementation to methods that aren't implemented when a class implements an interface. Swift doesn't have abstract classes so we fake it by adding the methods that are missing with a 'fatalError()' implementation.

It's used in this file: https://github.com/foam-framework/foam2/blob/35dc0bd85e197cec26c06ad6c9da1d4d6ea8107b/src/foam/swift/refines/Method.js

I realize there are alternatives here, just not sure which route to take. Any suggestions?

mcarcaso: I removed the parentCls property from Method and exposed some methods instead:
https://github.com/foam-framework/foam2/pull/675/commits/0cf057b7826996d3a2a68fb3627a269d31e2180c

I'm not crazy about having to use helper methods for accessing these properties but I'm not sure if there's an alternative if I can't store this extra bit of state on the model. I was trying to think of an approach that uses the context instead but I'm not sure that's any better than my previous solution. Thoughts?

demos/swift/js/somepackage/RequiredClass.js
mcarcaso: Updated here and a bunch of other files.

PR #676, opened 2017-09-25
SQL types

kirkeaton:
Added SQL types to property refinements.

TODO: figure out what to do with FObjectPropery, FObjectArray, Enum, add support for inserting StringArrays

### DIFF
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -70,7 +70,8 @@ foam.CLASS({
         propRequired: this.required,
         jsonParser:   this.javaJSONParser,
         extends:      this.javaInfoType,
-        transient:    this.transient
+        transient:    this.transient,
+        sqlType:      this.sqlType
       })
     },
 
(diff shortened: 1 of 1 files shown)


src/foam/java/refinements.js
```diff
@@ -70,7 +70,8 @@ foam.CLASS({
         propRequired: this.required,
         jsonParser:   this.javaJSONParser,
         extends:      this.javaInfoType,
-        transient:    this.transient
+        transient:    this.transient,
```
kgrgreer: I think we need both networkTransient and storageTransient, but don't actually need 'transient'.

src/foam/java/refinements.js
```diff
@@ -834,6 +842,9 @@ foam.CLASS({
       var c = info.getMethod('comparePropertyToObject');
       c.body = 'return compareValues((String) key, get_(o));';
 
+      var sql = info.getMethod('getSQLType');
```
kgrgreer: This could be done with a property:
{
  name: 'sqlType',
  expression: function(width) {
    return 'VARCHAR(' + this.width + ')';
  }
}

src/foam/java/refinements.js
kirkeaton: I've added networkTransient and storageTransient. But why would I remove just transient?

src/foam/java/refinements.js
kirkeaton: `this.width` or just `width`?

PR #678, opened 2017-09-25
PStream modifications

kirkeaton:
1. Modelled PStream as a FOAM interface
2. Added auto generation of ProxyPStream
3. Modelled Parser as a FOAM interface
4. Added auto generation of ProxyParser
5. Added apply method to PStream interface
6. Modified Parser implementations to make use of the apply function instead of the parser function.
7. Renamed StringPS to StringPStream


mdittmer:
@eatonk @kgrgreer This is a breaking change. Consumers of `StringPS` in JavaScript are broken without warning. Could we please revert and ease into deprecation in JavaScript?

kirkeaton:
@mdittmer I've submitted a PR that reverts the name back to `StringPS`

PR #679, opened 2017-09-25
Cronjob service and cronDAO fixes

diaswrd:
Cronjobs were not being stored in the cronDAO in the backend.


src/foam/nanos/client/Client.js
```diff
@@ -276,10 +276,12 @@ foam.CLASS({
         {
           name: 'cronDAO',
           factory: function() {
-            return this.EasyDAO.create({
-              daoType: 'CLIENT',
+            return this.RequestResponseClientDAO.create({
```
kgrgreer: Why is this change needed?

src/foam/nanos/cron/CronScheduler.java
```diff
@@ -32,13 +32,20 @@
...
    */
   private Date getMinScheduledTime() {
-    Min min = (Min) MLang.MIN(Cron.SCHEDULED_TIME);
-    cronDAO_.select(min);
-    return ((Cron) min.getValue()).getScheduledTime();
+    Min min = (Min) cronDAO_.select(
+      MLang.MIN(Cron.SCHEDULED_TIME)
+    );
+
+    if ((int) min.getValue() == 0) {
```
kgrgreer: Here you're casting to an (int) but below you're casting to a (Date).

src/foam/nanos/cron/CronScheduler.java
```diff
@@ -65,7 +72,10 @@ public void put(FObject obj, Detachable sub) {
         });
 
         Date minScheduledTime = getMinScheduledTime();
-        Thread.sleep(minScheduledTime.getTime() - now.getTime());
+
+        if (minScheduledTime != null) {
```
kgrgreer: Spacing is:
 if ( minScheduledTime != null ) {

src/foam/nanos/cron/CronSchedulerService.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.INTERFACE({
+  package: 'foam.nanos.cron',
+  name: 'CronSchedulerService',
```
kgrgreer: What is this needed for?

src/foam/nanos/client/Client.js
diaswrd: I thought `RequestResponseClientDAO` was the correct DAO to connect the front-end `cronDAO` with the `cronDAO` nano in the back-end. Any suggestions?

src/foam/nanos/cron/CronSchedulerService.js
diaswrd: In my understanding, the `CronSchedulerService` was necessary because of its `run` method that contains a loop which verifies any cronjobs that must run at a given time. 
But I could be wrong, perhaps it doesn't need its own service? 

The original class `CronSchedulerService` author could clarify that.

src/foam/nanos/client/Client.js
kgrgreer: The old code looks correct.

src/foam/nanos/cron/CronScheduler.java
diaswrd: Thanks. Pushing a fix.

src/foam/nanos/client/Client.js
diaswrd: Does EasyDAO support http delegates? Like:

```
delegate: this.HTTPBox.create({
  method: 'POST',
  url: 'http://localhost:8080/cronDAO'
})
```

Accepting suggestions here.

kirkeaton: @diaswrd All you should be required to do is:
```
return this.EasyDAO.create({
  daoType: 'CLIENT',
  of: this.Cron,
  serviceName: 'cronDAO'
});
```

Was that not working?

src/foam/nanos/cron/CronSchedulerService.js
kirkeaton: @diaswrd Yeah I don't see why this service is necessary to be honest. CronScheduler.java implements NanoService. The run method is there because it implements Runnable.

kirkeaton: @diaswrd Was this done so you could auto generate the Skeleton?

kirkeaton: In any case, I would remove the run method and add
`javaExtends: [ 'java.lang.Runnable' ]` instead

src/foam/nanos/client/Client.js
diaswrd: @eatonk it was working, but the changes on the GUI were not being persisted in the cronDAO service running on the server. A page reload would wipe any new cronjobs created.

diaswrd:
@kgrgreer Client.js file removed :+1:

PR #680, opened 2017-09-25
StringPStream modifications

kirkeaton:
1. Added apply function to StringPStream
2. Updated calls to Parser parse function to use apply function in StringPStream

### DIFF
--- a/src/foam/parse/parse.js
+++ b/src/foam/parse/parse.js
@@ -785,7 +785,7 @@ foam.CLASS({
         console.error('No symbol found for', this.name);
         return undefined;
       }
-      return p.parse(ps, grammar);
+      return ps.apply(p, grammar);
     },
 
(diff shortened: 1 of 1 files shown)


src/foam/parse/parse.js
```diff
@@ -123,6 +123,10 @@ foam.CLASS({
                   'Cannot make substring: end PStream is not a tail of this.');
 
       return this.str[0].substring(this.pos, end.pos);
+    },
+
+    function apply(p, obj) {
```
kgrgreer, line 128: apply() is a method of functions in JS, so this might make the code (more) confusing for people because it will look the ps's are functions (I know, I'm the one who suggested the bad name). So a better name would be better.

kgrgreer: But can be done in a separate PR.

PR #688, opened 2017-09-26
TracingPStream

kirkeaton:
Added a TracingPStream which prints out the character currently being parsed and whether or not it was successful or not.

### DIFF
--- a/src/foam/lib/parse/TracingPStream.java
+++ b/src/foam/lib/parse/TracingPStream.java
@@ -9,6 +9,8 @@ package foam.lib.parse;
 import foam.nanos.logger.Logger;
 import foam.nanos.logger.StdoutLogger;
 
+import java.io.OutputStream;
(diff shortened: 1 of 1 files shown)


src/foam/lib/parse/TracingPStream.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Override
+  public PStream setValue(Object value) {
+    return new TracingPStream(super.setValue(value), writer, pos, depth + 1);
+  }
+
+  @Override
+  public PStream apply(Parser ps, ParserContext x) {
+    String indentation = IntStream.range(1, ( depth * 2 ) + 1).mapToObj(i -> " ")
```
kgrgreer, line 53: You could also do:
String indentation = String.format("%" + (depth*2) + "s", "");

PR #691, opened 2017-09-27
ErrorReportingPStream

kirkeaton:
Added ErrorReportingPStream which is able to listen for any errors that the parser throws and can return a message containing the invalid character, position, and a list of characters that would have been accepted. There is a root ErrorReportingPStream which creates ErrorReportingNodePStream's that report to it whenever an error occurs.

Added a TrapPStream which can be used to see whether or not a character would be accepted by a parser. It also reports the valid character to the ErrorReportingPStream.

Added InvalidPStream which is a Singleton PStream in which the valid() function always returns false.

**Example printout of error:**
```
Invalid character '"' found at 0
Valid characters include: '{'
```

### DIFF
--- a/src/foam/lib/parse/ErrorReportingNodePStream.java
+++ b/src/foam/lib/parse/ErrorReportingNodePStream.java
@@ -10,26 +10,28 @@ public class ErrorReportingNodePStream
     extends ProxyPStream
 {
   protected int pos;
+  protected ErrorReportingPStream root;
   protected ErrorReportingNodePStream tail_ = null;
 
-  public ErrorReportingNodePStream(PStream delegate) {
-    this(delegate, 0);
+  public ErrorReportingNodePStream(ErrorReportingPStream root, PStream delegate) {
+    this(root, delegate, 0);
   }
 
-  public ErrorReportingNodePStream(PStream delegate, int pos) {
+  public ErrorReportingNodePStream(ErrorReportingPStream root, PStream delegate, int pos) {
     setDelegate(delegate);
+    this.root = root;
     this.pos = pos;
   }
 
   @Override
   public PStream tail() {
-    if ( tail_ == null ) tail_ = new ErrorReportingNodePStream(super.tail(), pos + 1);
+    if ( tail_ == null ) tail_ = new ErrorReportingNodePStream(root, super.tail(), pos + 1);
     return tail_;
   }
 
   @Override
   public PStream setValue(Object value) {
-    return new ErrorReportingNodePStream(super.setValue(value), pos);
+    return new ErrorReportingNodePStream(root, super.setValue(value), pos);
   }
 
   @Override
--- a/src/foam/lib/parse/ErrorReportingNodePStream.java
+++ b/src/foam/lib/parse/ErrorReportingNodePStream.java
@@ -13,7 +13,7 @@ package foam.lib.parse;
 public class ErrorReportingNodePStream
     extends ProxyPStream
 {
-  protected int pos;
+  protected int pos_;
   protected ErrorReportingPStream root_;
   protected ErrorReportingNodePStream tail_ = null;
 
@@ -24,18 +24,18 @@ public class ErrorReportingNodePStream
   public ErrorReportingNodePStream(ErrorReportingPStream root, PStream delegate, int pos) {
     setDelegate(delegate);
     this.root_ = root;
-    this.pos = pos;
+    this.pos_ = pos;
   }
 
   @Override
   public PStream tail() {
-    if ( tail_ == null ) tail_ = new ErrorReportingNodePStream(root_, super.tail(), pos + 1);
+    if ( tail_ == null ) tail_ = new ErrorReportingNodePStream(root_, super.tail(), pos_ + 1);
     return tail_;
   }
 
   @Override
   public PStream setValue(Object value) {
-    return new ErrorReportingNodePStream(root_, super.setValue(value), pos);
+    return new ErrorReportingNodePStream(root_, super.setValue(value), pos_);
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


src/foam/lib/parse/ErrorReportingNodePStream.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * This is a PStream decorator that listens for errors when parsing and reports them
+ * to the root ErrorReportingPStream
+ */
+public class ErrorReportingNodePStream
+    extends ProxyPStream
+{
+  protected int pos;
+  protected ErrorReportingPStream root;
```
kgrgreer: Should be consistent with instance variable naming. The FOAM convention is to have instance vars end with '_', but I don't mind if people don't, but in this case you mix both styles. So pick one and use it throughout a file (I prefer using _'s.)

src/foam/lib/parse/ErrorReportingPStream.java
```diff
@@ -0,0 +1,117 @@
...
+      b.setLength(0);
+      return b;
+    }
+  };
+
+  protected Parser errParser = null;
+  protected ParserContext errContext = null;
+  protected ErrorReportingNodePStream errStream = null;
+
+  protected int pos;
```
kgrgreer: Is this variable needed? Won't 'pos' always be zero?

src/foam/lib/parse/ErrorReportingPStream.java
```diff
@@ -0,0 +1,117 @@
...
+    // get the report with the furthest position
+    if ( errStream == null || errStream.pos < ernps.pos ) {
+      errStream = ernps;
+      errParser = parser;
+      errContext = x;
+    }
+  }
+
+  public void reportValidCharacter(Character character) {
+    System.out.println("report valid char = " + character);
```
kgrgreer: Remove this line.

src/foam/lib/parse/ErrorReportingPStream.java
```diff
@@ -0,0 +1,117 @@
...
+    if ( result == null ) {
+      // if result is null then self report
+      this.report(new ErrorReportingNodePStream(this, getDelegate(), pos), parser, x);
+    }
+    return result;
+  }
+
+  public void report(ErrorReportingNodePStream ernps, Parser parser, ParserContext x) {
+    // get the report with the furthest position
+    if ( errStream == null || errStream.pos < ernps.pos ) {
```
kgrgreer: I think < should be <=. This will cause you to get the most top-level parser to fail at a particular position, which will contain all possible characters, not just some.

src/foam/lib/parse/ErrorReportingPStream.java
kirkeaton: Thought I had removed that already. Gone now.

src/foam/lib/parse/ErrorReportingPStream.java
```diff
@@ -0,0 +1,117 @@
...
+    validCharacters.add(character);
+  }
+
+  public String getMessage() {
+    // check if err is valid and print the char, if not print EOF
+    String invalid = ( errStream.valid() ) ? String.valueOf(errStream.head()) : "EOF";
+
+    // get a list of valid characters
+    TrapPStream trap = new TrapPStream(this);
+    Iterator i = ASCII_CHARS.iterator();
```
kgrgreer: I'm okay if you keep it, but I think that the ASCII_CHARS collection is a bit overkill, since you could have just looped from 0 to 255 here instead.

src/foam/lib/parse/ErrorReportingPStream.java
kirkeaton: Modified.

src/foam/lib/parse/ErrorReportingPStream.java
```diff
@@ -0,0 +1,117 @@
...
+    while ( i.hasNext() ) {
+      Character character = (Character) i.next();
+      trap.setHead(character);
+      trap.apply(errParser, errContext);
+    }
+
+    StringBuilder builder = sb.get()
```
kgrgreer: StringBuilder is more efficient than doing += with Strings, but if you build the whole string at once using +'s, it is compiled to exactly what you've typed here, but is less code to type.

src/foam/lib/parse/ErrorReportingPStream.java
kirkeaton: Good point. I'm getting a bit carried away with using Streams

src/foam/lib/parse/ErrorReportingPStream.java
kirkeaton: Yeah I think so. Removing now.

src/foam/lib/parse/ErrorReportingNodePStream.java
kirkeaton: Added the underscore suffix.

src/foam/lib/parse/ErrorReportingPStream.java
kirkeaton: Removed StringBuilder

src/foam/lib/parse/ErrorReportingPStream.java
kirkeaton: Okay I removed the ASCII_CARS variable

src/foam/lib/parse/ErrorReportingNodePStream.java
```diff
@@ -0,0 +1,49 @@
...
+package foam.lib.parse;
+
+/**
+ * This is a PStream decorator that listens for errors when parsing and reports them
+ * to the root ErrorReportingPStream
+ */
+public class ErrorReportingNodePStream
+    extends ProxyPStream
+{
+  protected int pos;
```
kgrgreer: pos_

kirkeaton: Done

PR #693, opened 2017-09-28
Missing Predicates

kirkeaton:
Added support for StartsWith and Keyword Predicates in Java

### DIFF
--- a/src/foam/mlang/mlangJava.js
+++ b/src/foam/mlang/mlangJava.js
@@ -983,8 +983,8 @@ foam.CLASS({
 
   javaImports: [
     'foam.core.PropertyInfo',
-     'java.util.Iterator',
-     'java.util.List'
+    'java.util.Iterator',
+    'java.util.List'
   ],
 
   methods: [
--- a/src/foam/mlang/mlangJava.js
+++ b/src/foam/mlang/mlangJava.js
@@ -832,6 +832,7 @@ foam.CLASS({
   ]
 });
 
+
 foam.CLASS({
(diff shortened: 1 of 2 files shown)


src/foam/mlang/mlangJava.js
```diff
@@ -441,24 +441,67 @@ return ( s1 != null && s2 != null && s1.toUpperCase().contains(s2.toUpperCase())
...
+return false;`
+    },
+    {
+      name: 'createStatement',
+      javaReturns: 'String',
+      args: [{
+        name: 'table',
+        javaType : 'String'
+      }],
+      javaCode: 'return "";'
```
kgrgreer: The SQL for starts with is:

LIKE 'startString%'

src/foam/mlang/mlangJava.js
```diff
@@ -441,24 +441,67 @@ return ( s1 != null && s2 != null && s1.toUpperCase().contains(s2.toUpperCase())
...
+        javaType: 'java.sql.PreparedStatement'
+      }],
+      javaCode: " return; "
+    }
+  ]
+});
+
+
+foam.CLASS({
   refines: 'foam.mlang.predicate.StartsWithIC',
```
kgrgreer, line 486: SQL is UPPER(property) like 'UPPER_STRING%'

src/foam/mlang/mlangJava.js
kirkeaton: I don't really know how these functions work to be honest.

PR #696, opened 2017-09-28
ServiceWebAgent error reporting

kirkeaton:
Reports JSON parsing errors to console and http response

### DIFF
--- a/src/foam/nanos/http/ServiceWebAgent.java
+++ b/src/foam/nanos/http/ServiceWebAgent.java
@@ -69,7 +69,7 @@ public class ServiceWebAgent
       FObject result = requestContext.create(JSONParser.class).parseString(buffer_.toString());
       if ( result == null ) {
         resp.setStatus(resp.SC_BAD_REQUEST);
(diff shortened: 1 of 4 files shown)


src/foam/nanos/http/ServiceWebAgent.java
```diff
@@ -64,11 +67,11 @@ public void execute(X x) {
       buffer_.rewind();
 
       FObject result = requestContext.create(JSONParser.class).parseString(buffer_.toString());
-
       if ( result == null ) {
         resp.setStatus(resp.SC_BAD_REQUEST);
-        System.err.println("Failed to parse request");
-        out.print("Failed to parse request: " + buffer_.toString());
+        String message = getParsingError(x, buffer_.toString());
+        System.err.println(message);
```
kgrgreer: Log this so that it's also saved to the log file, rather than printing to System.err. Also, do the same in the JDAO if I missed it there.

kirkeaton: Okay thanks. Changed it

PR #698, opened 2017-09-29
CSV support for Java

kirkeaton:


src/foam/lib/csv/Outputter.java
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   * Gets a filtered list of properties. Removes network and storage transient variables
+   * if necessary, removes unsupported types and removes null values / empty strings
+   * @param obj the object to get the property list from
+   * @return the filtered list of properties
+   */
+  public List<PropertyInfo> getFilteredPropertyInfoList(FObject obj) {
+    List<PropertyInfo> props = obj.getClassInfo().getAxiomsByClass(PropertyInfo.class);
```
kgrgreer: Cache this value so you don't need to build collection every time.

kirkeaton: That would prevent reusing an instance of this class to output different classes of FObjects.

PR #703, opened 2017-09-30
GroupBy Java

kirkeaton:
Added the GroupBy Sink to Java.

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -1268,12 +1268,16 @@ foam.CLASS({
       name: 'arg2'
     },
     {
+      class: 'Map',
       name: 'groups',
-      factory: function() { return {}; }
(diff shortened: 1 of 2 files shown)


src/foam/mlang/mlang.js
```diff
@@ -1268,13 +1268,16 @@ foam.CLASS({
...
+      factory: function() { return {}; },
+      javaFactory: `return new java.util.HashMap<Object, foam.dao.Sink>();`
     },
     {
-      class: 'StringArray',
+      class: 'List',
       name: 'groupKeys',
-      factory: function() { return []; }
+      factory: function() { return []; },
+      javaFactory: `return new java.util.ArrayList();`
```
kgrgreer, line 1280: The javaFactory should just be set in the List Model definition. I just did that and will remove this line.

PR #720, opened 2017-10-06
PostgreSQL Predicate support

kirkeaton:
1. Added IndexedPreparedStatement class which maintains an index which is used to set the values of a prepared statements
2. Modified PostgresDAO to use IndexPreparedStatement
3. Added SQLStatement interface
4. Made PropertyInfo, Comparator, and Predicate implement SQLStatement
5. Added createStatement and prepareStatement functions for Mlangs

TODO: test the creation of the SQL queries to validate their correctness


src/foam/core/AbstractPropertyInfo.java
```diff
@@ -48,12 +49,11 @@ public void toCSV(foam.lib.csv.Outputter outputter, Object value) {
 
   @Override
   public String createStatement() {
-    return getName();
+    return getName().toLowerCase();
```
kgrgreer, line 52: Why is this needed? Isn't Postgres case-insensitive?

src/foam/mlang/MLang.java
```diff
@@ -21,7 +21,9 @@
   public static final Predicate FALSE = new False();
 
   public static Expr prepare(Object o) {
-    return o instanceof Expr ? (Expr) o : new Constant(o);
+    return o instanceof Expr ? (Expr) o :
```
kgrgreer, line 24: Good.

src/foam/core/AbstractPropertyInfo.java
kirkeaton: PostgreSQL converts identifiers to lowercase if they are not double-quoted. So instead of double quoting all table names & column names I figured it would be easier to just convert to lowercase instead.

https://stackoverflow.com/a/20880247

kgrgreer: Okay, then can we not include this method in the AbtractPropertyInfo and have a custom implementation generated for each one specificiall? This will let us convert to lowercase at compile time and so we'll be faster and generate less garbage. Can be done in future PR.

PR #725, opened 2017-10-12
Authy 2FA SMS support

kirkeaton:
1. Added Authy as a dependency
2. Created AuthyService
3. Added authy to services


src/foam/nanos/auth/twofactor/authy/AuthyService.js
```diff
@@ -0,0 +1,86 @@
...
+          javaType: 'foam.nanos.auth.User'
+        }
+      ],
+      javaCode:
+`AuthyApiClient client = getClient();
+if ( client == null )
+  return false;
+
+PhoneVerification phoneVerification = client.getPhoneVerification();
+Verification verification = phoneVerification.start(user.getPhone(), "1", "sms", new Params());
```
kirkeaton: @kgrgreer How would you suggest getting the user's country code for their phone number?

PR #728, opened 2017-10-13
Email template

kirkeaton:
1. Added sendEmailFromTemplate function to EmailService
2. Added EmailTemplate object.
3. Added emailTemplateDAO to services
4. Added org.apache.commons.commons-text dependency
5. Wrote sendEmailFromTemplate implementation for SMTPEmailService
6. Modified functions to not throw exceptions but simply return true if successful and false otherwise.


src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -45,70 +57,118 @@ foam.CLASS({
...
+      javaReturns: 'boolean',
+      javaCode: 'return sendMessage(emailMessage);'
+    },
+    {
+      name: 'sendEmailFromTemplate',
+      javaReturns: 'boolean',
+      javaCode:
+`DAO emailTemplateDAO = (DAO) getX().get("emailTemplateDAO");
+Sink sink = new ListSink();
+sink = emailTemplateDAO.where(MLang.EQ(EmailTemplate.NAME, name)).select(sink);
```
kgrgreer: The EmailTemplate's primary key is name, so it's much simpler to just do emailTemplateDAO.find(name)

PR #741, opened 2017-10-17
HTTPBox outputter fix

kirkeaton:
1. Setting the context of the Outputter before using

### DIFF
--- a/src/foam/box/HTTPBox.js
+++ b/src/foam/box/HTTPBox.js
@@ -157,10 +157,10 @@ try {
   java.io.OutputStreamWriter output = new java.io.OutputStreamWriter(conn.getOutputStream(),
                                                                      java.nio.charset.StandardCharsets.UTF_8);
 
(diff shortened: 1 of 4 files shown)


src/foam/box/HTTPBox.js
```diff
@@ -157,11 +157,11 @@ try {
   java.io.OutputStreamWriter output = new java.io.OutputStreamWriter(conn.getOutputStream(),
                                                                      java.nio.charset.StandardCharsets.UTF_8);
 
-
-  output.write(new Outputter().stringify(message));
+  Outputter outputter = new Outputter();
```
kgrgreer, line 160: Would be better if Outputter had a constructor that took a Context then you could follow the structure of the original code.

kirkeaton: @kgrgreer I'll add it later. Can this be merged?

PR #746, opened 2017-10-19
Moved Phone into nanos

kennyqiyenkan:


src/foam/nanos/auth/User.js
```diff
@@ -63,12 +61,13 @@ foam.CLASS({
       name: 'email'
     },
     {
-      class: 'PhoneNumber',
+      class: 'FObjectProperty',
```
kgrgreer, line 64: Since type is a property of Phone, shouldn't this just be an FObjectArray of Phone's?

src/foam/nanos/auth/Phone.js
```diff
@@ -0,0 +1,27 @@
...
+      name: 'verified'
+    },
+    {
+      class: 'PhoneNumber',
+      name: 'number',
+      required: true
+    },
+    {
+      class: 'String',
+      name: 'type'
```
kgrgreer: Define a ChoiceView for this to specify the common types. Or a combobox.

kirkeaton (approved):

src/foam/nanos/auth/Phone.js
kgrgreer: Or remove the type if we're going to keep the phone/mobile structure from User.

src/foam/nanos/auth/User.js
kennyqiyenkan: I think for now we will keep the structure of user and get rid of type in Phone.

src/foam/nanos/auth/Phone.js
kennyqiyenkan: Will do the latter

mcarcaso:
LGTM

PR #747, opened 2017-10-19
Modified SMTPEmailService to use Jtwig

kirkeaton:
1. Modified SMTPEmailService to use Jtwig templating library
2. Added DAOResourceLoader to handle when templates are including in other templates
3. Removed code that was being used with old template implementation


src/foam/nanos/notification/email/DAOResourceLoader.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  public InputStream load(String s) {
+    EmailTemplate template = (EmailTemplate) dao_.find(s);
+    return new ByteArrayInputStream(template.getBody().getBytes(StandardCharsets.UTF_8));
+  }
+
+  @Override
+  public boolean exists(String s) {
+    return ( dao_.find(s) != null );
```
kirkeaton, line 34: @kgrgreer Should I be caching the templates into a Map in this class?

src/foam/nanos/notification/email/DAOResourceLoader.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  public Optional<Charset> getCharset(String s) {
+    return Optional.absent();
+  }
+
+  @Override
+  public InputStream load(String s) {
+    EmailTemplate template = (EmailTemplate) dao_.find(s);
+    return new ByteArrayInputStream(template.getBody().getBytes(StandardCharsets.UTF_8));
```
kgrgreer: Wouldn't it be more direct to just use StringBufferInputStream?

src/foam/nanos/notification/email/DAOResourceLoader.java
kgrgreer: If StringBufferInputStream works, then the overhead is negligible, so I wouldn't bother.

src/foam/nanos/notification/email/DAOResourceLoader.java
kirkeaton: StringBufferInputStream is deprecated and the Java docs say this in regards to it:
>This class does not properly convert characters into bytes.

PR #749, opened 2017-10-20
Fixed bug with enum getters returning the first value of the enum

kirkeaton:
Getter functions for Enum properties were returning the first value defined in the Enum. This was causing those default values to be marshalled even if not explicitly set. This PR makes the getters return null if not set.

### DIFF
--- a/src/foam/core/Enum.js
+++ b/src/foam/core/Enum.js
@@ -320,12 +320,6 @@ foam.CLASS({
         return of && of.VALUES[0];
       },
     },
-    {
-      name: 'javaValue',
-      expression: function(of, value) {
-        return of.id + '.' + value;
-      },
-    },
     [
       'adapt',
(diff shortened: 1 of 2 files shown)


mcarcaso (changes requested):
Instead of applying the fix where you've put it, could you put a ['javaValue', 'null'] on the foam.core.Enum refinement? Then you shouldn't need to care about enum's specifically here.

kirkeaton:
@mcarcaso Thanks. That also worked

mcarcaso (commented):
Spoke with Kirk about this on Slack. I mislead him with this fix and he's going to take a different approach to the problem he's encountering. The real issue is that the outputter outputs all properties instead of only outputting properties that have been set. This requires adding hasOwnProperty to java's FObjects and checking that in the Outputter.

PR #752, opened 2017-10-20
Added isDefaultValue function to FObject & PropertyInfo classes

kirkeaton:


src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -19,6 +19,15 @@
 public abstract class AbstractArrayPropertyInfo
   extends AbstractPropertyInfo {
 
+  public int compareValues(Object[] b1, Object[] b2) {
+    if (b1 == b2) return 0;
```
kgrgreer: Spacing should be:
  if ( b1 == b2 ) return 0;

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -19,6 +19,15 @@
...
+    if (b2 == null) return 1;
+    if (b1 == null) return -1;
+    int h1 = b1.hashCode();
+    int h2 = b2.hashCode();
+    return h1 == h2 ? 0 : h1 > h2 ? 1 : -1;
```
kgrgreer: It isn't sufficient to compare hashCodes because different values can hash to the same value.

src/foam/core/AbstractArrayPropertyInfo.java
kirkeaton: Fixed

src/foam/core/AbstractArrayPropertyInfo.java
kirkeaton: Okay. I just copied this implementation over from AbstractFObjectArrayProperty. Not sure who originally wrote it. What would you suggest instead?

kirkeaton: @kgrgreer I ended up just removing the compareValues function because it wasn't being used anywhere to begin with

src/foam/java/PropertyInfo.js
```diff
@@ -162,6 +164,13 @@ foam.CLASS({
             type: 'boolean',
             args: [ { name: 'o', type: 'Object' } ],
             body: `return ((${this.sourceCls.name}) o).${this.propName}IsSet_;`
+          },
+          {
+            name: 'isDefaultValue',
+            visibility: 'public',
+            type: 'boolean',
+            args: [ { name: 'o', type: 'Object' } ],
+            body: `return compareValues(get_(o), ${this.propValue}) == 0;`
```
adamvy: Would be worth adding a TODO that this needs to be revisiting when/if expression support is added to Java.

src/foam/core/AbstractFObject.java
```diff
@@ -82,4 +82,11 @@ public boolean hasOwnProperty(String prop) {
     PropertyInfo property = (PropertyInfo) getClassInfo().getAxiomByName(prop);
     return property != null && property.hasOwnProperty(this);
   }
+
+  public boolean hasDefaultValue(String prop) {
+    if ( ! this.hasOwnProperty(prop) )
```
kgrgreer: BTW: hasOwnProperty() is a Javascript thing. Seems odd having it for Java. Would be better if called isSet() or isPropertySet() or something like that.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -18,7 +18,7 @@
 
 public abstract class AbstractArrayPropertyInfo
   extends AbstractPropertyInfo {
-
+
```
kgrgreer: Remove trailing whitespace.

kirkeaton: Updated

src/foam/core/AbstractFObject.java
kirkeaton: Changed the function names

src/foam/java/PropertyInfo.js
kirkeaton: Thanks, added TODO

Issue #737, opened 2017-10-16
Have GenJava not overwrite files that don't change


kgrgreer:
Builds are currently slower than necessary because genjava overwrites files even if they didn't change, causing every java file to be recompiled for every build.

Modify genjava so that it only overwrites .java files which have actually changed.

@hchoura

kgrgreer:
Done by @hchoura

PR #753, opened 2017-10-23
Genjava Optimise #737

hchoura:
we propose to generate all of the java files, but then don't actually write them if they have the same contents as the file that's already there. This will prevent the timestamp of the file from updating despite
its contents not changing, which will then prevent the code from being recompiled.

### DIFF
--- a/tools/genjava.js
+++ b/tools/genjava.js
@@ -159,11 +159,7 @@ function copyJavaClassesToBuildFolder(startPath) {
 
       result = result.concat(copyJavaClassesToBuildFolder(filePath));
     } else if (f.search('.js') < 0) {
-      var readFilePath = fs_.readFileSync(filePath);
-        if (! ( fs_.existsSync(outputPath) && (fs_.readFileSync(outputPath).toString() == readFilePath.toString()))) {
-          fs_.writeFileSync(outputPath, readFilePath);
-          result.push(outputPath);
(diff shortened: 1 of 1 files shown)


tools/genjava.js
```diff
@@ -79,12 +79,15 @@ function generateClass(cls) {
 
   ensurePath(outfile);
 
-  fs_.writeFileSync(outfile, cls.buildJavaClass().toJavaSource());
+  var buildJavaSource = cls.buildJavaClass().toJavaSource();
+  if (! ( fs_.existsSync(outfile) && (fs_.readFileSync(outfile).toString() == buildJavaSource))) {
```
kgrgreer: This pattern is repeated several times, so I think it would be better to have a helper method. Maybe something like:
writeFileIfUpdated(outfile, cls.buildJavaClass().toJavaSource());

tools/genjava.js
```diff
@@ -168,7 +159,7 @@ function copyJavaClassesToBuildFolder(startPath) {
 
       result = result.concat(copyJavaClassesToBuildFolder(filePath));
     } else if (f.search('.js') < 0) {
-      var readFilePath = fs_.readFileSync(filePath)
+      var readFilePath = fs_.readFileSync(filePath);
```
kgrgreer: Couldn't this use your writeFileIfUpdated() function also?

PR #757, opened 2017-10-24
Blob java

kirkeaton:
1. Added java generation of Blob, BlobService, AbstractBlob, ProxyBlob
2. Added IdentifiedBlob


src/foam/blob/BlobJava.js
```diff
@@ -0,0 +1,103 @@
...
+        },
+        {
+          name: 'id',
+          javaType: 'String'
+        }
+      ]
+    },
+    {
+      name: 'urlFor',
+      javaReturns: 'String[]',
```
adamvy: Not String[], just String

kirkeaton: The JS version has `returns: 'Strings'` is that a typo then?

src/foam/blob/IdentifiedBlob.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package foam.blob;
+
+import foam.core.X;
+
+import java.nio.Buffer;
+
+public class IdentifiedBlob
```
adamvy: We should generate this from the foam.CLASS so that serialization works.

kirkeaton: IdentifiedBlob has a property called delegate with a factory that looks up the delegate based on the ID. How would I handle that when auto generating?

src/foam/blob/BlobJava.js
adamvy: Yeah that's a typo

src/foam/blob/IdentifiedBlob.java
adamvy: Add a javaFactory to the property to do the same thing

PR #758, opened 2017-10-24
Additional Blob java changes

kirkeaton:
1. Fixed typo
2. Added javaFactory for delegate property in IdentifiedBlob

### DIFF
--- a/src/foam/blob/BlobJava.js
+++ b/src/foam/blob/BlobJava.js
@@ -101,14 +101,3 @@ foam.INTERFACE({
     }
   ]
 });
-
-foam.CLASS({
-  refines: 'foam.blob.IdentifiedBlob',
-
-  properties: [
-    {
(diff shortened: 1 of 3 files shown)


src/foam/blob/BlobJava.js
```diff
@@ -100,4 +100,15 @@ foam.INTERFACE({
...
   ]
+});
+
+foam.CLASS({
+  refines: 'foam.blob.IdentifiedBlob',
+
+  properties: [
+    {
+      name: 'delegate',
+      javaFactory: 'return ((BlobService) getBlobService()).find(getId());'
```
adamvy: We may have to inline this right into the main class.  I'm not sure refining javaFactory works properly.  It may clobber the regular JS factory.

kirkeaton: Okay. Moved

PR #759, opened 2017-10-25
Blob Java

kirkeaton:
1. Added foam.blob.Buffer java implementation
2. Added getSize and setSize to Blob interface
3. Added AbstractBlobService in JS and Java
4. Merged Blob_web into Blob.js
5. Added SubBlob Java implementation


src/foam/blob/Blob.js
```diff
@@ -29,8 +55,22 @@ foam.INTERFACE({
...
         {
-          name: 'offset',
-          of: 'Long'
+          class: 'Int',
+          name: 'offset'
+        }
+      ]
+    },
+    {
+      name: 'getSize',
```
adamvy: Oh I didn't notice you removed the size property.  If we're going to keep it removed, which is fine, we only need getSize.

Blobs are meant to be immutable, so no need for setSize.

src/foam/blob/Blob.js
```diff
@@ -47,60 +87,32 @@ foam.INTERFACE({
     {
       name: 'put',
       returns: 'Promise',
-      args: [
-        {
-          name: 'blob',
-          of: 'foam.blob.Blob'
-        }
-      ]
+      args: [ 'blob' ]
```
adamvy: I'd rather keep the additional information unless it's causing problems.  In the future we will use it to remove the need to specify javaType in the args

src/foam/blob/Blob.js
```diff
@@ -154,6 +159,28 @@ foam.CLASS({
...
 foam.CLASS({
   package: 'foam.blob',
+  name: 'AbstractBlobService',
+  abstract: true,
+
+  implements: ['foam.blob.BlobService'],
+
+  methods: [
+    function put(blob) {
+      return this.put_(this.__context__, blob);
```
adamvy: Add an .in(x) like we do for DAO to return a new ProxyDAO to this in the specified context.

adamvy: PRoxyBlobService I mean

src/foam/blob/Blob.js
```diff
@@ -175,12 +202,15 @@ foam.CLASS({
   extends: 'foam.blob.AbstractBlob',
   properties: [
     {
+      class: 'foam.core.Blob',
```
adamvy: Just clasS: 'Blob' is fine.  foam.core is the default

src/foam/blob/Blob.js
```diff
@@ -250,7 +283,7 @@ foam.CLASS({
   ],
   properties: [
     {
-      class: 'String',
+      class: 'Int',
```
adamvy: Id should be a string not an int, the real BlobServices use a hex representation of the SHA256 hash of the blob.  Only in TestBlobService did we use an int, but that was just for convenient, we can convert the ints to strings in TestBlobService.

src/foam/blob/Blob.js
```diff
@@ -311,7 +344,7 @@ foam.CLASS({
       name: 'fd'
     },
     {
-      class: 'Long',
+      class: 'Int',
```
adamvy: Why this change?  Blob's can be quite large.

src/foam/blob/Blob.js
kirkeaton: @adamvy The reason for removing it was that the getSize & setSize functions weren't being properly generated for ProxyBlob

adamvy: Sure that's fine, let's keep it this way, but let's remove setSize() entirely, size is meant to be read only as blobs are immutable.

kirkeaton: @adamvy Removed the setSize function

src/foam/blob/Blob.js
kirkeaton: Removed the prefix

src/foam/blob/Blob.js
kirkeaton: @adamvy Done.

src/foam/blob/Blob.js
kirkeaton: Okay, I've added it back

PR #761, opened 2017-10-25
Fix Stub/ReplyBox interaction for RPCs

mdittmer:
RPCs over forks/sockets were broken by eliminating `ReplyBox` wrapping in `Stub`.

### DIFF
--- a/src/lib/Stub.js
+++ b/src/lib/Stub.js
@@ -48,7 +48,7 @@ foam.CLASS({
             })
           });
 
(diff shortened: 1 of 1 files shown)


adamvy:
We always set the replyBox incase a "void" method throws exceptions.  Without setting the replyBox you don't get exceptions.

Where was this causing issues for you?

adamvy:
Disregard previous comment if you received it via email.  I'm cathcing up on unread emails and didn't realize how old this PR was.

PR #766, opened 2017-10-25
Profile picture blob

kirkeaton:
1. Made profile picture a Blob instead of a String
2. Added profilePicture to tableColumns
3. Added tableCellFormatter to profilePicture property


src/foam/nanos/auth/User.js
```diff
@@ -79,8 +79,11 @@ foam.CLASS({
       name: 'birthday'
     },
     {
-      class: 'String',
-      name: 'profilePicture'
+      class: 'Blob',
+      name: 'profilePicture',
+      tableCellFormatter: function (value) {
+        this.tag({ class: 'foam.u2.view.BlobView' });
```
adamvy: You can switch this to ImageBlobView to render the actual image instead of a Download link if you want.

src/services
```diff
@@ -18,7 +18,7 @@ p({"class":"foam.nanos.boot.NSpec", "name":"pmInfoDAO",                   "lazy"
...
 p({"class":"foam.nanos.boot.NSpec", "name":"localUserDAO",                "lazy":true,  "serve":false, "serviceScript":"return new foam.dao.PMDAO(new foam.dao.SequenceNumberDAO (new foam.dao.JDAO(foam.nanos.auth.User.getOwnClassInfo(), \"users\")));"})
-p({"class":"foam.nanos.boot.NSpec", "name":"userDAO",                     "lazy":true,  "serve":true,  "serviceScript":"return /* new foam.nanos.auth.AuthenticatedUserDAO(*/x.get(\"localUserDAO\")/*)*/;","client":"{\"of\":\"foam.nanos.auth.User\", \"cache\":true}"})
+p({"class":"foam.nanos.boot.NSpec", "name":"userDAO",                     "lazy":true,  "serve":true,  "serviceScript":"return new net.nanopay.auth.email.EmailVerificationUserDAO(x.get(\"localUserDAO\"));","client":"{\"of\":\"foam.nanos.auth.User\",\"cache\":true,\"decorators\": [{\"class\":\"foam.blob.BlobServiceDecorator\",\"of\":\"foam.nanos.auth.User\"}]}"})
```
adamvy: "of" is no longer required on the BlobServiceDecorator decorator.

adamvy:
LGTM two small changes

src/services
kirkeaton: Removed

src/foam/nanos/auth/User.js
kirkeaton: Changed

src/foam/nanos/auth/User.js
```diff
@@ -16,7 +16,7 @@ foam.CLASS({
   documentation: '',
 
   tableColumns: [
-    'id', 'enabled', 'firstName', 'lastName', 'organization', 'lastModified'
+    'id', 'enabled', 'firstName', 'lastName', 'organization', 'lastModified', 'profilePicture'
```
kgrgreer, line 19: I'm not too keen on having the provilePicture be a column in the table view, but people can hide that column if they want, so I guess it's okay.

kirkeaton: It was just for testing purposes. It can be removed.

PR #769, opened 2017-10-26
Array utils

marcogbarcellos:
Creating ArrayUtils class as a helper and adding first method `append` to append existing arrays to one or more items.


src/foam/util/ArrayUtils.java
```diff
@@ -0,0 +1,39 @@
...
+      String [] merged3 = ArrayUtils.append(s1, "Ball", "Soccer"); // Output : [{"a", "b", "c", "Ball", "Soccer"}]
+  *
+  * @param <T> the component type of the array
+  * @param array  the initial array
+  * @param items  new items to be added to original array
+  * @return The new array
+  */
+  public static <T> T[] append(T[] array, T... items) {
+
+    if( array == null && items == null ) {
```
kgrgreer: Follow the FOAM coding standard for spacing.

kgrgreer: If items == null but array != null then you should just return array, since it's already the correct value. Not adding new values shouldn't remove old ones.

kgrgreer: This check is redundant.

src/foam/util/ArrayUtils.java
```diff
@@ -0,0 +1,39 @@
+package foam.util;
+
+import java.lang.reflect.Array;
+
+public class ArrayUtils {
```
kgrgreer: Rename to just 'Arrays'.

PR #771, opened 2017-10-26
BlobStore java implementation

kirkeaton:
1. Implemented BlobStore in Java
2. Implemented FdBlob in Java
3. Added blobStore to services
4. Modified blobService to use HttpBlobService and RestBlobService


src/foam/blob/BlobStore.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.FileOutputStream;
+import java.nio.ByteBuffer;
+import java.security.MessageDigest;
+
+public class BlobStore
+    extends AbstractBlobService
+{
+  protected static final int BUFFER_SIZE = 8192;
+
+  protected String root_ = "/Users/kirk/backend";
```
kirkeaton: TODO: remove hardcoded path

PR #774, opened 2017-10-27
HttpBlobService doGet

kirkeaton:
1. Modelled BlobStore to allow setting of variables via the GUI
2. Made BlobStore extend AbstractBlobService
3. Fixed issues with FdBlob read function
4. Wrote doGet method of HttpBlobService


src/foam/blob/BlobJava.js
```diff
@@ -210,3 +210,155 @@ return getParent().read(buffer, offset + getOffset());`
...
+try {
+  MessageDigest hash = MessageDigest.getInstance("SHA-256");
+  Buffer buffer = new Buffer(BUFFER_SIZE, ByteBuffer.allocate(BUFFER_SIZE));
+
+  long chunk = 0;
+  long size = blob.getSize();
+  long chunks = (long) Math.ceil((double) size / (double) BUFFER_SIZE);
+
+  File tmp = allocateTmp(1);
+  while (chunk < chunks) {
```
kgrgreer: foam coding standard is:
while ( chunk < chunks ) { (extra spaces)

kirkeaton: Thanks. Modified

PR #776, opened 2017-10-30
End-to-end stub-over-socket (NodeJS) test + related fixes

mdittmer:
This obsoletes #763 and #761. @adamvy PTAL.


mdittmer:
N.B. https://github.com/foam-framework/foam2/pull/729 can be re-landed once this fix lands.

PR #780, opened 2017-10-30
Added java implementations for SessionReplyBox and SessionClientBox

kirkeaton:


src/lib/boxJava.js
```diff
@@ -101,3 +101,55 @@ try {
...
+  ],
+
+  methods: [
+    {
+      name: 'send',
+      javaCode:
+`Object object = message.getObject();
+if ( object instanceof RPCErrorMessage && ((RPCErrorMessage) object).getData() instanceof AccessControlException ) {
+  // TODO: should this be wrapped in new Thread() ?
+  ((Runnable) getX().get("requestLogin")).run();
```
kirkeaton, line 132: @kgrgreer Should this be wrapped in a new thread?

kgrgreer: I don't think so, since you want to know when it's done and the easiest way to do that is to just block this thread until run() returns.

kirkeaton: Okay thanks.

PR #782, opened 2017-10-31
Safe compare for Mlang predicates (EQ,NEQ,GTE,GT,LTE,LT)

waseem-chamaa:


src/foam/util/SafetyUtil.java
```diff
@@ -0,0 +1,23 @@
+package foam.util;
+
+
+import java.math.BigDecimal;
+
+public class SafetyUtil{
+
+  public static int compare(Object o1 , Object o2) {
+    
+    if (o1 == null && o2 == null) {
```
kgrgreer: Replace this if statement with these three:

if ( o1 == null && o2 == null ) Return 0;
if ( o2 == null ) return 1;
if ( o1 == null ) return -1;

See why?

src/foam/util/SafetyUtil.java
```diff
@@ -0,0 +1,23 @@
...
+      return 0;
+    }
+    if (o1 instanceof Number && o2 instanceof Number) {
+      if (o1.toString() != "NaN" || o2.toString() != "NaN") {
+        return new BigDecimal(o1.toString()).compareTo(new BigDecimal(o2.toString()));
```
kgrgreer: This is super-duper way-too expensive.

src/foam/util/SafetyUtil.java
```diff
@@ -0,0 +1,23 @@
...
+      }
+    }
+    if (o1 instanceof String && o2 instanceof String) {
```
kgrgreer: This if statement doesn't do anything because if you just remove it the last statement will do the same thing anyway.

src/foam/util/SafetyUtil.java
```diff
@@ -7,16 +7,14 @@
...
-      }
-    }
-    if (o1 instanceof String && o2 instanceof String) {
-      return ((String) o1).compareTo((String) o2);
+    if ( o1 == null && o2 == null) return 0;
+    if ( o2 == null ) return 1;
+    if ( o1 == null ) return -1;
+
+    if (o1 instanceof Number && o2 instanceof Number) {      
+      if( ((Number) o1).doubleValue() == ((Number) o2).doubleValue()) return 0;
```
kgrgreer: Store the double values in d1 and d2 rather than recomputing ever time.

src/foam/util/SafetyUtil.java
```diff
@@ -12,9 +12,13 @@ public static int compare(Object o1 , Object o2) {
     if ( o1 == null ) return -1;
 
     if (o1 instanceof Number && o2 instanceof Number) {      
-      if( ((Number) o1).doubleValue() == ((Number) o2).doubleValue()) return 0;
-      if( ((Number) o1).doubleValue() >  ((Number) o2).doubleValue()) return 1;
-      if( ((Number) o1).doubleValue() <  ((Number) o2).doubleValue()) return -1;
+      Double d1 =((Number) o1).doubleValue();
```
kgrgreer: double not Double, Double is a slow fat Object whereas double is a small fast native type

src/foam/util/SafetyUtil.java
```diff
@@ -12,9 +12,13 @@ public static int compare(Object o1 , Object o2) {
...
+      Double d2 =((Number) o2).doubleValue();
+     
+      if( d1 == d2) return 0;
```
kgrgreer: if ( d1 == d2 ) return 0;  // fixed spacing

src/foam/mlang/mlangJava.js
```diff
@@ -529,7 +529,7 @@ foam.CLASS({
     {
       name: 'f',
       // TODO(adamvy): Is there a better option than all the Comparable casts?
-      javaCode: 'return ((Comparable)getArg1().f(obj)).compareTo((Comparable)getArg2().f(obj)) == 0;'
+      javaCode: `return  foam.util.SafetyUtil.compare(getArg1().f(obj),getArg2().f(obj))==0;`
```
kgrgreer: Use ' instead of ` for single-line strings since it's faster.

PR #783, opened 2017-11-01
Rework foam.net.node.Server and friends

mdittmer:
New design discussed here: https://groups.google.com/forum/#!topic/foam-framework-discuss/kaFVJVeW9Ek

This is a breaking change for the `Server`, and `*Handler` interfaces.

A `Server` now has one main `Handler` (and a secondary for errors). Common `pathname`, or "route"-based handlers implement `Router` (which extends existing `Handler` interface). Routers also provide inversion-of-control to Handlers being bound to a route. This allows Handlers to decorate themselves, change state, and/or recontextualize when being bound to a specific route in a specific context.


src/foam/net/node/RestDAOHandler.js
```diff
@@ -18,7 +18,7 @@
 foam.CLASS({
   package: 'foam.net.node',
   name: 'RestDAOHandler',
-  extends: 'foam.net.node.Handler',
+  extends: 'foam.net.node.BaseHandler',
```
jacksonic: Docs would be handy, especially an overview describing its use relative to a RestDAO.

jacksonic (changes requested):

src/foam/net/node/RestDAOHandler.js
mdittmer: Done.

kgrgreer:
LGTM

jacksonic (approved):

PR #785, opened 2017-11-02
Swift warning suppression

waseem-chamaa:


src/foam/swift/SwiftJava.js
```diff
@@ -0,0 +1,103 @@
+foam.CLASS({
+  refines: 'foam.core.Argument',
+  properties: [
+  {
```
kgrgreer: Use the more compact version of properties for these, since we don't plan on adding extra details. ie.
properties: [ 'swiftType, 'swiftDefaultValue' ]
And for all similar uses as well.

src/foam/swift/SwiftJava.js
```diff
@@ -1,103 +1,61 @@
...
-    name: 'swiftExpression',
-  },
-  {
-    name: 'swiftExpressionArgs',
-  },
-],
+    'swiftType',
+    'swiftFactory',
+    'swiftExpression',
+    'swiftExpressionArgs',
```
kgrgreer: No trailing commas.

PR #789, opened 2017-11-02
Node server caching

mdittmer:
Use decorators over handler and response interfaces to achieve caching. Includes some unit tests (could probably use more for caching non-200s, requests that throw underneath decorator, etc.). An end-to-end test with a server would be nice too.


mdittmer:
Note: Current diffbase is `node-server` branch. Will keep merging against it until it lands, then merge `master` and change diffbase.

kgrgreer:
LGTM

src/foam/net/node/CachingResponse.js
```diff
@@ -0,0 +1,119 @@
...
+      value: null
+    }
+  ],
+
+  methods: [
+    function getStatusCode() {
+      return this.res.getStatusCode();
+    },
+    function setStatusCode(statusCode) {
+      if ( this.responseError ) return;
```
jacksonic, line 73: Should these more aggressively throw errors? Are they expected to ever get called more than once?

The more-work-but-future-proof solution could be to copy-on-write if the cached response does get altered after retrieval.

jacksonic (approved):

PR #791, opened 2017-11-02
AuthenticatedDAO

kirkeaton:
Added AuthenticatedDAO which checks whether or not has sufficient permissions to execute a DAO function.

### DIFF
--- a/src/foam/dao/AuthenticatedDAO.java
+++ b/src/foam/dao/AuthenticatedDAO.java
@@ -69,9 +69,7 @@ public class AuthenticatedDAO
 
   @Override
   public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
-    if ( ! ( sink instanceof AuthenticatedSink ) ) {
-      sink = new AuthenticatedSink(name_, "read", sink);
-    }
+    sink = new AuthenticatedSink(name_, "read", sink);
     return super.select_(x, sink, skip, limit, order, predicate);
   }
 
@@ -83,17 +81,13 @@ public class AuthenticatedDAO
 
   @Override
   public void listen_(X x, Sink sink, Predicate predicate) {
-    if ( ! ( sink instanceof AuthenticatedSink ) ) {
-      sink = new AuthenticatedSink(name_, "listen", sink);
-    }
+    sink = new AuthenticatedSink(name_, "listen", sink);
     super.listen_(x, sink, predicate);
   }
 
   @Override
   public void pipe_(X x, Sink sink) {
-    if ( ! ( sink instanceof AuthenticatedSink ) ) {
-      sink = new AuthenticatedSink(name_, "pipe", sink);
-    }
+    sink = new AuthenticatedSink(name_, "pipe", sink);
     super.pipe_(x, sink);
   }
 }
--- a/src/foam/dao/AuthenticatedDAO.java
+++ b/src/foam/dao/AuthenticatedDAO.java
@@ -1,0 +1,113 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.dao;
+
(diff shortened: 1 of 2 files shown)


src/foam/dao/AuthenticatedDAO.java
```diff
@@ -0,0 +1,113 @@
...
+      throw new RuntimeException("Insufficient permissions");
+    }
+
+    return super.find_(x, id);
+  }
+
+  @Override
+  public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+    AuthService authService = (AuthService) x.get("auth");
+    Permission permission = new AuthPermission(name_ + ".read");
```
kgrgreer: Should decorate the sink and filter each row individually. May have permission for some rows but not others.

src/foam/dao/AuthenticatedDAO.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    return super.select_(x, sink, skip, limit, order, predicate);
+  }
+
+  @Override
+  public void removeAll_(X x, long skip, long limit, Comparator order, Predicate predicate) {
+    AuthService authService = (AuthService) x.get("auth");
+    Permission permission = new AuthPermission(name_ + ".delete");
+    if ( ! authService.check(x, permission) ) {
```
kgrgreer: Decorate sink. Same as select().

src/foam/dao/AuthenticatedDAO.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    super.removeAll_(x, skip, limit, order, predicate);
+  }
+
+  @Override
+  public void listen_(X x, Sink sink, Predicate predicate) {
+    AuthService authService = (AuthService) x.get("auth");
+    Permission permission = new AuthPermission(name_ + ".listen");
+    if ( ! authService.check(x, permission) ) {
```
kgrgreer: Let listen but then decorate sink to check eac hrow.

src/foam/dao/AuthenticatedDAO.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    super.listen_(x, sink, predicate);
+  }
+
+  @Override
+  public void pipe_(X x, Sink sink) {
+    AuthService authService = (AuthService) x.get("auth");
+    Permission permission = new AuthPermission(name_ + ".pipe");
+    if ( ! authService.check(x, permission) ) {
```
kgrgreer: Let listen but then decorate sink to check each row.

src/foam/dao/AuthenticatedDAO.java
kirkeaton: There's no sink property in this method.

src/foam/dao/AuthenticatedSink.java
```diff
@@ -0,0 +1,45 @@
...
+import foam.core.FObject;
+import foam.nanos.auth.AuthService;
+
+import javax.security.auth.AuthPermission;
+import java.security.Permission;
+
+public class AuthenticatedSink
+    extends ProxySink
+{
+  protected String name_;
```
kgrgreer: These two properties could be combined into just 'prefix'. Not too concerned either way.

src/foam/dao/AuthenticatedSink.java
```diff
@@ -0,0 +1,45 @@
...
+  protected String method_;
+
+  public AuthenticatedSink(String name, String method, Sink delegate) {
+    this.name_ = name;
```
kgrgreer: Should take X as arg and set, otherwise how will it find the AuthService?

src/foam/dao/AuthenticatedSink.java
kirkeaton: Done

src/foam/dao/AuthenticatedSink.java
kirkeaton: Done

PR #792, opened 2017-11-03
Added getCurrentUser function to AuthService

kirkeaton:
Added getCurrentUser function to AuthService which returns the user currently stored in the context.


src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -155,6 +155,10 @@ public void start() {
     challengeMap  = new LRULinkedHashMap<Long, Challenge>(20000);
   }
 
+  public User getCurrentUser() {
```
kgrgreer: First argument of getCurrentUser() will need to be X, so that it's set by the SessionServerBox so that you actually know the current User.

kirkeaton: Thanks, added

PR #794, opened 2017-11-03
Added a cronjobs (Session Expiry)

minsunny:
For Session Expiry for checking expired sessions and remove them.
It runs the first minute of every hour.


minsunny:
I signed it!

PR #796, opened 2017-11-03
Node JS Server: FileHandler as PathnameHandler

mdittmer:
This change updates pathname-based handling to incorporate exact match for
files, which is different from prefix matching.


kgrgreer:
Has conflicts.

mdittmer:
@kgrgreer @jacksonic conflicts fixed.

PR #798, opened 2017-11-04
Added registration service

kirkeaton:
1. Removed AuthenticatedUserDAO
2. Made login, challengedLogin, and loginByEmail return the User
3. Created Email util class to verify email address
3. Created Password util class to hash & verify passwords
 * Passwords are hashed using PBKDF with a 64 bit salt, 512 key size, and 1000 iterations
 * BouncyCastle is used to perform the PBKDF
4. Modified UserAndGroupAuthService to use Password and Email Utils
5. Added RegistrationService, ClientRegistrationService and UserRegistrationService


src/foam/util/Password.java
```diff
@@ -0,0 +1,108 @@
...
+import org.bouncycastle.crypto.params.KeyParameter;
+import org.bouncycastle.util.encoders.Base64;
+
+import java.security.SecureRandom;
+import java.util.regex.Pattern;
+
+public class Password {
+
+  // Min 8 characters, at least one uppercase, one lowercase, one number
+  public static Pattern PASSWORD_PATTERN = Pattern.compile("^(?=.*[a-z])(?=.*[A-Z])(?=.*\\d)[a-zA-Z\\d]{8,}$");
```
kgrgreer, line 20: This is okay for now, but this should be handled by the AuthService which has a method for validating users, including their passwords. Then we can make this configurable. I'll show/explain to you how XHome did this.

kirkeaton: Thanks

PR #800, opened 2017-11-05
foam.mlang.In: Empty arrays may appear as undefined Constant values; …

mdittmer:
…make sure partialEval() and TreeIndex can handle this.


mdittmer:
Added @jacksonic since this change ultimately required TreeIndex changes

PR #806, opened 2017-11-06
got rid of grabbing cookies in trace

waseem-chamaa:


kgrgreer:
I would rather keep the Cookies section for webservers that support it, and instead, just wrap in a try/catch so that that section isn't displayed if unavailable.

PR #811, opened 2017-11-07
Xpostgres

xuerongCode:
Add setStatementValue() and setFromResultSet() in PropertyInfo;
change StringArray SQLtype to TEXT;
fix initial bug and comment out SQLException in PostgresDAO constructor;
replace fix 'id' string literal by getPrimaryKey().createStatement() in PostgresDAO.java;
add DateConstant in mlangJava.js;
fix createStatement() and preparedStatement in ArrayConstant;
fix prepare() method in MLang class;
override setFromString() method in AbstractArrayPropertyInfo;
add setObject(Object, int) method in IndexedPreparedStatement class;
fix select_() in PostgresDAO;


jacksonic:
Looks like the last batch of commits was with a different email address. Can you sign the CLA with that one too?

xuerongCode:
I only sign up one CLA. The problem happened because I git clone the code that I pushed from my other computer? I did not set the email and username at this computer. I do not know how to solve it

jacksonic:
Looks like you can edit and rebase the earlier commits: https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

Effectively you're editing the old commit email addresses, then rebuilding the chain of commits off of the new versions.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -67,4 +73,36 @@ public void toXML (FObject obj, Document doc, Element objElement) {
...
+      return;
+    }
+    Object[] os = (Object[]) obj;
+    java.lang.StringBuilder sb = new java.lang.StringBuilder();
+    int length = os.length;
+    if ( length == 0 ) {
+      stmt.setObject(null);
+      return;
+    }
+    for ( int i=0; i < length; i++) {
```
kgrgreer: FOAM Coding standard requires this line be:
for ( int i = 0 ; i < length ; i++ ) {

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -22,12 +22,43 @@
...
-   String[] s  = value.split(",", -1);
-   this.set(obj, s);
+   }
+   java.util.List<String> list = new java.util.LinkedList<String>();
+   StringBuilder sb = new StringBuilder(); 
+   char prev = '$';
+   int length = value.length();
+   char[] cs = value.toCharArray();
+   for ( int i = 0 ; i < cs.length ; i++ ) {
+       if ( cs[i] == '\\' ) {
```
kgrgreer: Indents should be two spaces.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -22,12 +22,43 @@
 {
   @Override
   public void setFromString(Object obj, String value) {
+    System.out.println("return String: " + value);
```
kgrgreer: Remove.

src/foam/mlang/mlangJava.js
```diff
@@ -567,7 +567,7 @@ for ( int i = 0; i < length; i++ ) {
   if ( obj[i] == null )
     builder.append("");
   else 
-    builder.append(obj[i]);
+    builder.append(obj[i].toString().replace("\\\\","\\\\\\\\").replace(",","\\\\,"));
```
kgrgreer: In-line regexes like this are difficult to understand. Instead, make a helper method which describes what it does and then move the regex to that function. Something like: escapeCommas().

Even better, do something like escapeCommasAndAppend(builder, obj[i]) which does the escaping and appending in one step to avoid generating garbage, which will slow us down. Not such a big deal in this case anyway since Postgres is already slow, but generally that would be a better design.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -21,7 +22,42 @@
...
+        }
+        prev = '$';
+      } else {
+        sb.append(cs[i]);
+        prev = cs[i];
+      }
+    }
+    list.add(sb.toString());
+    int resultSize = list.size();
+    System.out.println("list size: " + list.size());
```
kgrgreer: remove.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -67,4 +103,37 @@ public void toXML (FObject obj, Document doc, Element objElement) {
...
+    int length = os.length;
+    if ( length == 0 ) {
+      stmt.setObject(null);
+      return;
+    }
+    for ( int i = 0 ; i < length ; i++ ) {
+      if ( os[i] == null )
+        sb.append("");
+      else {
+        sb.append(os[i].toString().replace("\\","\\\\").replace(",","\\,"));
```
kgrgreer: Move to helper function with descriptive name.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -21,7 +22,42 @@
 {
   @Override
   public void setFromString(Object obj, String value) {
-    // TODO
+    if ( value == null ) {
+      this.set(obj, null);
+      return;
+    }
+    java.util.List<String> list = new java.util.LinkedList<String>();
```
kgrgreer: Import so you can just refer to java.util.List as List. Same for LinkedList. Add a TODO: to reuse this list with a ThreadLocal for better performance.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -21,7 +22,42 @@
...
+        sb.append(cs[i]);
+        prev = cs[i];
+      }
+    }
+    list.add(sb.toString());
+    int resultSize = list.size();
+    System.out.println("list size: " + list.size());
+    String[] result = new String[resultSize];
+    //add support for other array types
+    this.set(obj, list.subList(0, resultSize).toArray(result));
```
kgrgreer: What is the subList() call need for?

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -21,7 +23,42 @@
...
-    // TODO
+    if ( value == null ) {
+      this.set(obj, null);
+      return;
+    }
+    // TODO: TO REUSE THIS LIST WITH A THREADLOCAL FOR BETTER PERFORMANCE
+    List<String> list = new LinkedList<String>();
+    StringBuilder sb = new StringBuilder(); 
+    char prev = '$';
+    int length = value.length();
```
kgrgreer: Where is 'length' used? Did you mean to use this in for the for loop?

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -21,7 +23,42 @@
...
+          sb.setLength(0);
+        }
+        prev = '$';
+      } else {
+        sb.append(cs[i]);
+        prev = cs[i];
+      }
+    }
+    list.add(sb.toString());
+    int resultSize = list.size();
```
kgrgreer: Is it worth making this a var if you only use it once?

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -67,4 +104,46 @@ public void toXML (FObject obj, Document doc, Element objElement) {
...
+    }
+    Object[] os = (Object[]) obj;
+    StringBuilder sb = new StringBuilder();
+    int length = os.length;
+    if ( length == 0 ) {
+      stmt.setObject(null);
+      return;
+    }
+    for ( int i = 0 ; i < length ; i++ ) {
+      if ( os[i] == null )
```
kgrgreer: Add {}'s to if part.

src/foam/mlang/mlangJava.js
```diff
@@ -492,6 +492,24 @@ foam.CLASS({
...
+  package: 'foam.mlang',
+  name: 'DateConstant',
+  extends: 'foam.mlang.Constant',
+
+  methods: [
+    {
+      name: 'prepareStatement',
+      javaReturns: 'void',
+      javaCode: 
+`if ( getValue() instanceof java.util.Date ){
```
kgrgreer: If we're checking here, why do we need a special DateConstant model?

src/foam/mlang/mlangJava.js
```diff
@@ -492,6 +492,19 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
```
kgrgreer: Weren't we going to get rid of the DateConstant and just make the wrapped PreparedStatement smarter?

xuerongCode: You mean: add a if statement inside the IndexPreparedStatement->setObject(). If it is Date type, use delegate.setObject(obj, java.sql.types.TimeStamp)?

kgrgreer: Yes

xuerongCode: That is make sense. Since only Postgres has problem on the timeStamp, it dose not mean that other databases have same issues.

PR #818, opened 2017-11-08
Thread Web Agent

zahnno:
PR includes:

- Thread Web Agent.
- Provides a list of current active threads.
- Threads are clickable links that send params to browser which then redirects user to page with stack trace of the clicked thread.


src/foam/nanos/http/ThreadsWebAgent.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.PrintWriter;
+import javax.servlet.http.HttpServletRequest;
+
+public class ThreadsWebAgent
+  implements WebAgent
+{
+  public ThreadsWebAgent() {}
+
+  public void execute(X x) {
+    final PrintWriter out  = (PrintWriter) x.get(PrintWriter.class);
```
kgrgreer, line 23: Why is this final?

zahnno: PrintWriter out won't change.

src/foam/nanos/http/ThreadsWebAgent.java
```diff
@@ -0,0 +1,57 @@
...
+    HttpServletRequest req = (HttpServletRequest) x.get(HttpServletRequest.class);
+
+    Set<Thread> threadSet = Thread.getAllStackTraces().keySet();
+
+    Thread[] threadArray = threadSet.toArray(new Thread[threadSet.size()]);
+    out.println("<HEAD><TITLE>Threads</TITLE></HEAD>\n");
+    out.println("<pre>");
+    out.println("<H1>Threads</H1>\n");
+
+    for (Thread thread : threadArray){
```
kgrgreer: Follow https://github.com/foam-framework/foam2/blob/master/doc/guides/StyleGuide.md for FOAM code.

src/foam/nanos/http/ThreadsWebAgent.java
```diff
@@ -0,0 +1,57 @@
...
+  public ThreadsWebAgent() {}
+
+  public void execute(X x) {
+    final PrintWriter out  = (PrintWriter) x.get(PrintWriter.class);
+    HttpServletRequest req = (HttpServletRequest) x.get(HttpServletRequest.class);
+
+    Set<Thread> threadSet = Thread.getAllStackTraces().keySet();
+
+    Thread[] threadArray = threadSet.toArray(new Thread[threadSet.size()]);
+    out.println("<HEAD><TITLE>Threads</TITLE></HEAD>\n");
```
kgrgreer: Add BODY tags.

src/foam/nanos/http/ThreadsWebAgent.java
kgrgreer: Then why isn't 'req' on the line below final also?

zahnno: true true

PR #820, opened 2017-11-09
Added PasswordHashingDAO

kirkeaton:
Added PasswordHashingDAO decorator that hashes password property if found

### DIFF
--- a/src/foam/nanos/auth/PasswordHashingDAO.java
+++ b/src/foam/nanos/auth/PasswordHashingDAO.java
@@ -1,0 +1,32 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.auth;
+
+import foam.core.FObject;
+import foam.core.PropertyInfo;
+import foam.core.X;
+import foam.dao.DAO;
+import foam.dao.ProxyDAO;
+import foam.util.Password;
+
+public class PasswordHashingDAO
+    extends ProxyDAO
+{
+  public PasswordHashingDAO(X x, DAO delegate) {
+    setX(x);
+    setDelegate(delegate);
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    PropertyInfo prop = (PropertyInfo) obj.getClassInfo().getAxiomByName("password");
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/PasswordHashingDAO.java
```diff
@@ -0,0 +1,32 @@
...
+  public PasswordHashingDAO(X x, DAO delegate) {
+    setX(x);
+    setDelegate(delegate);
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    PropertyInfo prop = (PropertyInfo) obj.getClassInfo().getAxiomByName("password");
+    if ( prop != null && prop.get(obj) instanceof String ) {
+      prop.set(obj, Password.hash((String) prop.get(obj)));
```
kgrgreer, line 28: Hashing without salting isn't good because someone can just hash a dictionary of common passwords to get a dictionary of common hashes.

kirkeaton: Please refer to the Password utility class. A salt is randomly generated and used when hashing.

kgrgreer: Great!

TW80000: @eatonk @kgrgreer If we're trying to defend against this kind of attack by using a salt but sending the salt to the client, doesn't that defeat the purpose?

See https://github.com/foam-framework/foam2/blob/master/src/foam/util/Password.java#L87 for the return value of `Password.hash()` and https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/auth/UserPasswordHashingDAO.java#L32 for where we set the `password` property on the User model to the result of a call to `Password.hash()`. That means we're sending the salt and the hash of the password to the user on the client, which would be just as (in)secure as sending just the password hash and not using a salt at all, right?

Shouldn't we be storing the salt somewhere but never sending it to the client?

**Edit**: Answered through Slack. The `password` property on the User model is network transient, so it's not accessible on the client. Although since you can (at the time of writing) execute arbitrary scripts on the server from the client (through the admin panel "Scripts" and "Tests"), you could still do such an attack since any script you write would have access to the salt on the server. Once you're able to execute arbitrary scripts on the server though, this is the least of your security concerns.

PR #822, opened 2017-11-09
Prevent duplicate email dao

kirkeaton:
Added DAO decorator to prevent the inserting of a new user with a duplicate email address

### DIFF
--- a/src/foam/nanos/auth/PreventDuplicateEmailDAO.java
+++ b/src/foam/nanos/auth/PreventDuplicateEmailDAO.java
@@ -1,0 +1,53 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.auth;
+
+import foam.core.FObject;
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/PreventDuplicateEmailDAO.java
```diff
@@ -0,0 +1,53 @@
...
+    setDelegate(delegate);
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    User user = (User) obj;
+    boolean newUser = ( getDelegate().find(user.getId()) == null );
+
+    if ( newUser ) {
+      if ( user.getEmail() == null || user.getEmail().isEmpty() ) {
```
kgrgreer, line 34: SafetyUtil.isEmpty(user.getMail()) would be nicer.

kirkeaton: Does that exist?

kgrgreer: SafetyUtil does, but isEmpty() doesn't yet, but you could add.

PR #825, opened 2017-11-09
Placeholder functionality for input TextFields

blakegreen32:
Added functionality to set the placeholder element of a DOM when adding a TextField. All you have to do is pass in the field placeholder: '' when creating a TextField view.


src/foam/u2/TextField.js
```diff
@@ -40,6 +41,7 @@ foam.CLASS({
 
       if ( this.type         ) this.setAttribute('type', this.type);
       if ( this.displayWidth ) this.setAttribute('size', this.displayWidth);
+      if ( this.placeholder  )  this.setAttribute('placeholder', this.placeholder);
```
kgrgreer: Doesn't TextField extend Input, so isn't also adding this here redundant?

kgrgreer:
But now you're no longer adding placeholder to Input.

PR #826, opened 2017-11-10
Made SMTPEmailService asynchronous

kirkeaton:
1. Made SMTPEmailService asynchronous. 
2. Fixed NPE in FixedThreadPool

### DIFF
--- a/src/foam/nanos/notification/email/SMTPEmailService.js
+++ b/src/foam/nanos/notification/email/SMTPEmailService.js
@@ -16,7 +16,10 @@ foam.CLASS({
   ],
 
   javaImports: [
+    'foam.core.ContextAgent',
+    'foam.core.X',
     'foam.dao.DAO',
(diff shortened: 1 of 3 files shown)


src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -64,6 +67,18 @@ foam.CLASS({
     }
   ],
 
+  axioms: [
+    {
+      name: 'javaExtras',
+      buildJavaClass: function (cls) {
+        cls.extras.push(foam.java.Code.create({
+          data:
+  `FixedThreadPool pool = new FixedThreadPool();`
```
kgrgreer: There's already a ThreadPool in the context which you should use instead of creating your own.

kirkeaton: Thanks, fixed

PR #829, opened 2017-11-10
PasswordHashingDAO fix

kirkeaton:
1. Checking to see if incoming object has password property and that the property is set. 
2. Checks to see if the object already exists in the DAO, if it doesn't or doesn't have the password property set, then hash the incoming password. 
3. Additional check to make sure the incoming password matches the stored password to prevent illegal password changes.


src/foam/nanos/auth/PasswordHashingDAO.java
```diff
@@ -22,11 +26,53 @@ public PasswordHashingDAO(X x, DAO delegate) {
   }
 
   @Override
+  public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+    if ( sink == null ) sink = new ListSink();
```
kgrgreer, line 32: Weren't you going to handle this in AbstractDAO.

src/foam/nanos/auth/PasswordHashingDAO.java
```diff
@@ -22,11 +26,53 @@ public PasswordHashingDAO(X x, DAO delegate) {
...
+    ProxySink proxy = new ProxySink(x, sink) {
```
kgrgreer: Make own top-level class rather than in-line sub-class so that it's serializable for network DAO's.

src/foam/nanos/auth/PasswordHashingDAO.java
kirkeaton: It isn't working because the DAOSkeleton calls select_ instead of select

src/foam/nanos/auth/PasswordHashingDAO.java
kirkeaton: Okay done.

src/foam/nanos/auth/PasswordRemovalSink.java
```diff
@@ -0,0 +1,31 @@
...
+{
+  public PasswordRemovalSink(X x, Sink delegate) {
+    setX(x);
+    setDelegate(delegate);
+  }
+
+  @Override
+  public void put(FObject obj, Detachable sub) {
+    FObject clone = obj.fclone();
+    clone.setProperty("password", null);
```
kgrgreer: Next three lines could be in a static method which could also be called from UserAndGroupAuthService to avoid duplicating the same logic. Or vise-versa.

kirkeaton: Done

kgrgreer: You should call that static method from here.

kirkeaton: Done for real this time.

PR #833, opened 2017-11-14
add ReaderPStream

hchoura:


src/foam/lib/parse/ReaderPStream.java
```diff
@@ -0,0 +1,86 @@
...
+		return false;
+	}
+
+	public PStream tail() {
+		if (tail_ == null)
+			tail_ = new ReaderPStream(str, pos + 1);
+		return tail_;
+	}
+
+	private Object value_ = null;
```
kgrgreer, line 60: Declare all vars first.

src/foam/lib/parse/ReaderPStream.java
```diff
@@ -0,0 +1,86 @@
+package foam.lib.parse;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.io.StringReader;
+import java.nio.CharBuffer;
+
+public class ReaderPStream implements PStream {
+
+	protected Reference<StringReader> str;
```
kgrgreer, line 10: Indent two spaces.

src/foam/lib/parse/ReaderPStream.java
```diff
@@ -0,0 +1,86 @@
...
+		this(s, p, null);
+	}
+
+	public ReaderPStream(Reference<StringReader> s, int p, Object value) {
+		str = s;
+		pos = p;
+		value_ = value;
+	}
+
+	public void setString(StringReader data) {
```
kgrgreer, line 32: This was so that you could reset the underlying String and reuse the StringPStreams, which wouldn't work the same way for a Reader.

src/foam/lib/parse/ReaderPStream.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.CharBuffer;
+
+public class ReaderPStream implements PStream {
+
+	protected Reference<StringReader> str;
+	protected int pos;
+	protected ReaderPStream tail_ = null;
+
+	public ReaderPStream() {
+		this(new Reference<StringReader>());
```
kgrgreer, line 15: The Reference was used so that you could replace the underlying String for a whole chain of StringPStreams to reuse them by just swapping out the reference, but that doesn't make sens for a Reader, so we should just get rid of the extra Reference indirection.

PR #836, opened 2017-11-14
GoogleMapsGeocodingDAO

kirkeaton:
1. Added latitude and longitude to Address model
2. Modelled GoogleMaps geocoding response
3. Wrote GoogleMapsGeocodingDAO which sets the latitude and longitude of a User's address property
4. Modified FObjectParser to use the defaultClass passed in the constructor
5. Modified JSONParser to not set the defaultClass in the parser context
6. Modified FObjectArrayParser to accept defaultClass in the constructor
7. Modified auto generated jsonParser function for FObjectArrayPropertyInfo classes to set the defaultClass
8. Fixed bug with cast method for StringArray properties
9. Modified FObjectArrayParser to not require a trailing comma at the end of the array

### DIFF
--- a/src/foam/lib/json/JSONParser.java
+++ b/src/foam/lib/json/JSONParser.java
@@ -25,7 +25,9 @@ public class JSONParser
     ps.setString(data);
     ParserContext x = new ParserContextImpl();
     x.set("X", getX());
-    x.set("defaultClass", defaultClass);
+    if ( defaultClass != null ) {
+      parser = new ExprParser(defaultClass);
+    }
     ps = (StringPStream) ps.apply(parser, x);
 
     return ps == null ? null : (FObject) ps.value();
--- a/src/foam/lib/json/JSONParser.java
+++ b/src/foam/lib/json/JSONParser.java
@@ -29,7 +29,6 @@ public class JSONParser
       parser = new ExprParser(defaultClass);
     }
     ps = (StringPStream) ps.apply(parser, x);
-
     return ps == null ? null : (FObject) ps.value();
   }
 }
--- a/src/foam/lib/json/JSONParser.java
+++ b/src/foam/lib/json/JSONParser.java
@@ -25,10 +25,7 @@ public class JSONParser
     ps.setString(data);
     ParserContext x = new ParserContextImpl();
     x.set("X", getX());
-    if ( defaultClass != null ) {
-      parser = new ExprParser(defaultClass);
-    }
-    ps = (StringPStream) ps.apply(parser, x);
+    ps = (StringPStream) ps.apply(defaultClass == null ? parser : new ExprParser(defaultClass), x);
     return ps == null ? null : (FObject) ps.value();
   }
 }
--- a/src/foam/nanos/geocode/GoogleMapsAddressComponent.js
+++ b/src/foam/nanos/geocode/GoogleMapsAddressComponent.js
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
 foam.CLASS({
   package: 'foam.nanos.geocode',
   name: 'GoogleMapsAddressComponent',
--- a/src/foam/nanos/geocode/GoogleMapsAddressComponent.js
+++ b/src/foam/nanos/geocode/GoogleMapsAddressComponent.js
@@ -8,18 +8,23 @@ foam.CLASS({
   package: 'foam.nanos.geocode',
   name: 'GoogleMapsAddressComponent',
 
+  documentation: 'Component of an address (i.e. street number, or country)',
+
   properties: [
     {
       class: 'String',
-      name: 'long_name'
+      name: 'long_name',
+      documentation: 'Full text description or name of the address component as returned by the Geocoder'
     },
     {
       class: 'String',
-      name: 'short_name'
+      name: 'short_name',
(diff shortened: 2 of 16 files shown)


kirkeaton:
Closing because I need to add logic to handle when the address is changed.

kgrgreer:
Don't break up, but this should have been several unrelated PR's.

src/foam/lib/json/JSONParser.java
```diff
@@ -25,9 +25,10 @@ public FObject parseString(String data, Class defaultClass) {
     ps.setString(data);
     ParserContext x = new ParserContextImpl();
     x.set("X", getX());
-    x.set("defaultClass", defaultClass);
+    if ( defaultClass != null ) {
+      parser = new ExprParser(defaultClass);
```
kgrgreer: This makes this class be non reusable since calling parseString() has side-effects (ie. rewriting 'parser').
A safer approach would be to change the line below to:
 ps = (StringPStream) ps.apply(defaultClass == null ? parser : new ExprParser(defaultClass), x);

src/foam/nanos/geocode/GoogleMapsAddressComponent.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.geocode',
+  name: 'GoogleMapsAddressComponent',
+
```
kgrgreer, line 10: Add documentation:

src/foam/nanos/geocode/GoogleMapsBoundary.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.geocode',
+  name: 'GoogleMapsBoundary',
+
```
kgrgreer, line 10: Add documentation:

src/foam/nanos/geocode/GoogleMapsGeocodingDAO.java
```diff
@@ -0,0 +1,168 @@
...
+import foam.nanos.auth.Address;
+import foam.nanos.auth.User;
+import foam.util.SafetyUtil;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+public class GoogleMapsGeocodingDAO
```
kgrgreer, line 23: This is pretty nifty.

src/foam/nanos/geocode/GoogleMapsGeocodingDAO.java
```diff
@@ -0,0 +1,168 @@
...
+      b.setLength(0);
+      return b;
+    }
+  };
+
+  public GoogleMapsGeocodingDAO(X x, String apiKey, DAO delegate) {
+    setX(x);
+    setDelegate(delegate);
+    this.apiKey_ = apiKey;
+    this.parser_ = getX().create(JSONParser.class);
```
kgrgreer: Is the JSONParser thread-safe?

src/foam/nanos/geocode/GoogleMapsGeocodingDAO.java
kgrgreer, line 23: If this only works for Users I would put that in the name. Or better, provide the PropertyInfo for the Address field so it can be used for any Model with an Address.

src/foam/nanos/geocode/GoogleMapsGeocodingDAO.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    // append api key
+    builder.append("&key=").append(apiKey_);
+
+    String line = null;
+    HttpURLConnection conn = null;
+    BufferedReader reader = null;
+
+    try {
```
kgrgreer, line 108: Would it make sense for this to be a separate GeoService in the Context rather than being in-lined here? In case you wanted to GeoLocate in some other context?

src/foam/nanos/geocode/GoogleMapsGeocodingDAO.java
kirkeaton: I don't imagine it is so I will just make this be a local variable.

src/foam/lib/json/JSONParser.java
kirkeaton: Thanks, done.

src/foam/nanos/geocode/GoogleMapsBoundary.js
kirkeaton: Added documentation to all of the classes

src/foam/nanos/geocode/GoogleMapsGeocodingDAO.java
kirkeaton: Modified it to accept PropertyInfo and modified the put_ method to use that instead of checking if it's a user or not.

src/foam/nanos/geocode/GoogleMapsGeocodingDAO.java
kirkeaton: Yeah it would make sense. I can do that in a following PR.

PR #837, opened 2017-11-15
Added merchant user type

kirkeaton:
Added merchant as a choice to user type choice view


kgrgreer:
FYI: @diaswrd @marcogbarcellos

PR #840, opened 2017-11-15
Address hours

kirkeaton:
1. Added address hours
2. added day of week enum
3. added relationship between address and hours

### DIFF
--- a/src/foam/nanos/auth/Hours.js
+++ b/src/foam/nanos/auth/Hours.js
@@ -8,7 +8,7 @@ foam.CLASS({
   package: 'foam.nanos.auth',
   name: 'Hours',
 
-  documentation: 'Represents open and closing time'
+  documentation: 'Represents open and closing time',
 
   properties: [
     {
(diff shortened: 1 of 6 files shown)


src/foam/nanos/auth/Relationships.js
```diff
@@ -3,49 +3,58 @@
...
+  inverseName: 'group',
+  sourceProperty: {
+    hidden: true
+  },
+  targetProperty: {
+    hidden: true
+  }
+});
+
+foam.RELATIONSHIP({
```
kgrgreer: This isn't going to work because Address doesn't have a primaryKey or a DAO. Instead it should be an FObjectArray. The factory could return an array of 7 Hours already setup for each day.

kirkeaton: Thanks. Done.

src/foam/nanos/auth/Hours.js
```diff
@@ -0,0 +1,39 @@
...
+
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'Hours',
+
+  documentation: 'Represents open and closing time',
+
+  constants: [
+    {
+      name: 'SUNDAY',
```
kgrgreer: Remove trailing , and empty line.

kirkeaton: The constants should be removed altogether. Removed.

PR #841, opened 2017-11-17
AddressDetailView

minsunny:
1. Created AddressDetailView.js (new)
- added new field "Structure(boolean), It is checked, shown Street Number, Street Name, Suite.
   But It is unchecked, shown Address1, Address2. (Cause an address structure depends on countries)
- Modified Label
  as-is : CountryId, RegionId
  to-be : Country, Province/State

2. Modified Address.js
- added new filed address2, structure

3. Modified User.js
- applied AddressDetailView


src/files.js
```diff
@@ -355,6 +355,7 @@ FOAM_FILES([
   { name: "foam/u2/DateTimeView", flags: ['web'] },
   { name: "foam/u2/RangeView", flags: ['web'] },
   { name: "foam/u2/ReadWriteView", flags: ['web'] },
+  { name: "foam/u2/AddressDetailView", flags: ['web'] },
```
kgrgreer, line 358: Should be in the foam.nanos.auth package, which is where Address is.

src/foam/nanos/auth/Address.js
```diff
@@ -23,15 +23,36 @@ foam.CLASS({
       class: 'Boolean',
       name: 'deleted'
     },
+    {
+      class: 'Boolean',
+      name: 'structure',
```
kgrgreer, line 28: Rename to 'structured' (add the 'd').

src/foam/nanos/auth/Address.js
```diff
@@ -23,15 +23,36 @@ foam.CLASS({
...
+      class: 'Int',
+      name: 'streetNumber',
+    },
+    {
+      class: 'String',
+      name: 'streetName',
+    },
+    {
+      class: 'String',
+      name: 'address',
```
kgrgreer, line 51: For consistency with address2, let's name this 'address1'.

src/foam/u2/AddressDetailView.js
```diff
@@ -0,0 +1,143 @@
...
+
+  methods: [
+    function initE() {
+      this.SUPER();
+
+      this.
+        addClass(this.myClass()).
+        start().
+          start().
+            start('p').style({'display': 'inline', 'margin-top': '10px'}).add('Country').end().
```
kgrgreer, line 98: If you're going to keep repeating this style, better to add a class and then style in the CSS block above.

src/foam/u2/AddressDetailView.js
```diff
@@ -0,0 +1,143 @@
...
+            start('p').style({'display': 'inline', 'margin-top': '10px', 'margin-left': '170px'}).add(this.data.TYPE.label).end().
+          end().
+          start().
+            add(this.data.COUNTRY_ID).add(this.data.TYPE).
+          end().
+          start().
+            start('p').style({'display': 'inline', 'margin-top': '10px'}).add(this.data.VERIFIED.label).end().
```
kgrgreer, line 105: Instead of using a 'p' and adding display: inline, you could just use a 'span' and get the same result, I think.

src/foam/nanos/auth/User.js
```diff
@@ -106,9 +106,8 @@ emailIsSet_ = true;`
     },
     {
       class: 'FObjectProperty',
-      of: 'foam.nanos.auth.Address',
```
zahnno, line 109: it would be a good idea to keep this of property set to 'foam.nanos.auth.Address'

src/foam/u2/AddressDetailView.js
zahnno: we shouldn't be using inline styling (ex. .style({ })). It creates hard coded styles which are hard to overwrite in other places in the app & makes it hard for develops to find the css (considering that almost all css is contained in the css axiom.

src/foam/u2/AddressDetailView.js
```diff
@@ -0,0 +1,143 @@
...
+      margin-bottom: 10px;
+    }
+    ^ .property-suite {
+      width: 90px;
+      margin-left: 60px;
+      margin-bottom: 10px;
+    }
+    ^ .property-city {
+      width: 110px;
+      margin-bottom: 10px;
```
zahnno, line 69: I see a lot of repeated styles here, maybe you can combine a few of them together. A good tip as well is to use the html element as a css identifier, for example to place a style on all .start('p') you can do ^ p{ width: 110px }. This will provide all 'p's the width of 110px.

src/foam/u2/AddressDetailView.js
blakegreen32: Might not be necessary to have two .start() after the addClass(this.myClass()) tags at the beginning here. See if it works with only one. Also if you have a class you are reusing a lot add it to the AppStyles.js file and you can call it as many times as you want without adding a new class everytime.

src/foam/u2/AddressDetailView.js
kgrgreer: You can also set the "displayWidth: numberOfChars" property of the Properties in Address and then let the fields size themselves to the proper size. Involves less code, no CSS, and is transferable to Java and Swift clients.

src/foam/nanos/auth/User.js
kgrgreer: Agreed.

waseem-chamaa:
All changes were implemented and added to PR#888

kgrgreer:
Closed at @waseem-chamaa's request.

PR #842, opened 2017-11-17
Static method

hchoura:


src/foam/core/Static.js
```diff
@@ -0,0 +1,61 @@
+/**
```
kgrgreer: Year is 2017 and we're not Google, so the header should be:

/**
 * @license
 * Copyright 2017 The FOAM Authors. All Rights Reserved.
 */

src/foam/core/Static.js
```diff
@@ -0,0 +1,61 @@
...
+foam.CLASS({
+  package: 'foam.core',
+  name: 'Static',
+  extends: 'foam.core.AbstractMethod',
+
+  documentation: 'An Axiom for defining static methods.',
+
+  properties: [
+    //get prop from foam.core.AbstractMethod
+    { name: 'code', required: true },//TODO javaCode
```
kgrgreer: You inherit this from AbstractMethod so no need to duplicate.

src/foam/core/Static.js
```diff
@@ -0,0 +1,61 @@
...
+    { name: 'static', value: true }
+  ],
+
+  methods: [
+    function exportAs(obj) {
+      /*var m = obj[this.name];
```
kgrgreer: Remove comment.

kgrgreer: You also need to implement installInClass() and installInProto() to add the static method to the JS class/proto.

src/foam/java/refinements.js
```diff
@@ -288,6 +288,7 @@ foam.CLASS({
         name: this.name,
         type: this.javaReturns || 'void',
         visibility: 'public',
+        static : this.static ? 'static ' : '',
```
kgrgreer: This can be simplified to just:
static: this.static

since both statics are booleans.
Better, add a method to AbstractMethod called isStatic() which always returns false and add same method to Static that returns true. Then we don't need to store a static property in Static's, which doesn't really makes sense since since we don't want to let users make static Methods or non-static Statics.

src/foam/core/Static.js
kgrgreer: Use shorter header shown above.

src/foam/core/Static.js
```diff
@@ -0,0 +1,57 @@
...
+  extends: 'foam.core.AbstractMethod',
+
+  documentation: 'An Axiom for defining static methods.',
+
+  properties: [
+    //get prop from foam.core.AbstractMethod
+  ],
+
+  methods: [
+    function exportAs(obj) {
```
kgrgreer: Needs to installInClass() and installInProto() (like in Constant.js) so that these are added to the JS class.

src/foam/core/Static.js
```diff
@@ -0,0 +1,58 @@
...
+    },
+    function isStatic(){ return true;},
+    function installInClass(cls) {
+      Object.defineProperty(
+        cls,
+        foam.String.constantize(this.name),
```
kgrgreer: I don't think we should call constantize() in this case. Just use the name the developer provides.

src/foam/core/Static.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+foam.CLASS({
+  package: 'foam.core',
+  name: 'Static',
+  extends: 'foam.core.AbstractMethod',
+
+  documentation: 'An Axiom for defining static methods.',
+
+  properties: [
```
kgrgreer: Delete properties: if not used.

PR #844, opened 2017-11-17
restBolbService

xuerongCode:
implementing RestBolbService


kgrgreer:
@eatonk Please review.

src/foam/blob/RestBlobService.java
```diff
@@ -0,0 +1,164 @@
...
+
+public class RestBlobService
+    extends AbstractBlobService
+{
+  public static final int BUFFER_SIZE = 8192;
+
+  protected String address_;
+
+  public RestBlobService(String address) {
+    this.address_ = address + "/httpBlobService";
```
kirkeaton: Could replace this line with ```this(null, address);```

src/foam/blob/RestBlobService.java
```diff
@@ -0,0 +1,164 @@
...
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.nio.ByteBuffer;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.InputStreamReader;
+
+public class RestBlobService
```
kgrgreer, line 21: Add documentation.

src/foam/blob/RestBlobService.java
```diff
@@ -0,0 +1,164 @@
...
+        buffer.getData().clear();
+        chunk++;
+      }
+
+      os.flush();
+
+      if ( connection.getResponseCode() == HttpURLConnection.HTTP_OK ) {
+        is = connection.getInputStream();
+        BufferedReader reader = new BufferedReader(new InputStreamReader(is));
+        String json = "";
```
adamvy: I would parse this the same way we do in src/foam/nanos/http/ServiceWebAgent.java

CharBuffer buffer = CharBuffer.allocate(65535);
reader.read(buffer);
buffer.rewind();
parserString(buffer.toString());

Later we're going to let the parser read directly from an Reader, but in the meantime we might as well be consistent so anybody grepping the code will find both instances.

src/foam/blob/RestBlobService.java
```diff
@@ -0,0 +1,164 @@
...
+
+        result = (IdentifiedBlob) (new foam.lib.json.JSONParser()).parseString(json, IdentifiedBlob.class);
+        result.setX(getX());
+      } else {
+        throw new RuntimeException("upload fail");
+      }
+    } catch ( MalformedURLException e ) {
+      e.printStackTrace();
+    } catch ( IOException e ) {
+      e.printStackTrace();
```
adamvy: Might want to re-throw the exception as well, or re-throw it in a RuntimeException.

src/foam/blob/RestBlobService.java
```diff
@@ -0,0 +1,164 @@
...
+      if ( connection.getResponseCode() == HttpURLConnection.HTTP_OK ) {
+        is = connection.getInputStream();
+        blob = new InputStreamBlob(is);
+      } else {
+        throw new RuntimeException("download fail");
+      }
+    } catch ( MalformedURLException e ) {
+      e.printStackTrace();
+    } catch ( IOException e ) {
+      e.printStackTrace();
```
adamvy: Again might want to rethrow this exception.

kirkeaton:
@kgrgreer I am reviewing

src/foam/blob/RestBlobService.java
```diff
@@ -0,0 +1,164 @@
...
+      connection.setRequestMethod("PUT");
+
+      //configure http header
+      connection.setRequestProperty("Accept", "*/*");
+      connection.setRequestProperty("Connection", "keep-alive");
+      connection.setRequestProperty("Content-Type", "application/octet-stream");
+
+      //output blob into connection
+      long chunk = 0;
+      long size = blob.getSize();
```
kirkeaton, line 70: This is fine if the size of the blob is known but there are some cases where it is not. For example if you were to use an InputStreamBlob then this would not work.

src/foam/blob/RestBlobService.java
```diff
@@ -0,0 +1,164 @@
...
+        result.setX(getX());
+      } else {
+        throw new RuntimeException("upload fail");
+      }
+    } catch ( MalformedURLException e ) {
+      e.printStackTrace();
+    } catch ( IOException e ) {
+      e.printStackTrace();
+    } finally {
+      closeSource(is, os);
```
kirkeaton: You should be calling disconnect on your connection parameter in your close function as well

src/foam/blob/RestBlobService.java
xuerongCode: You mean there are some blob that we do not know the size. When we call blob.getSize(), it returns nothing? I check InputStreamBlob. We can get the size of a InputStreamBlob.

kirkeaton: Yes, but the InputStreamBlob implementation needs to be reworked because it crashes when you pass in an InputStream that contains too much data that can be stored in a byte array.

xuerongCode: If we do not know the size of a blob, how can we know how many iterations we need in order to get all bytes from a blob. Even though we can pass an empty buffer to the blob.read(), then check the return length of the buffer and put it in the while loop condition. It throws a IndexOutOfBoundsException, when we try to read the last chunk of the data.

PR #847, opened 2017-11-20
HttpServletRequestBlob

kirkeaton:
1. Created HttpServletRequest blob which adapts a servlet request into a blob
2. Removed InputStreamBlob
3. Replaced usage of InputStreamBlob in HttpBlobService with HttpServletRequest


src/foam/blob/HttpServletRequestBlob.java
```diff
@@ -0,0 +1,62 @@
...
+  public HttpServletRequestBlob(HttpServletRequest request) {
+    this.request_ = request;
+    this.size_ = request.getContentLengthLong();
+  }
+
+  @Override
+  public Buffer read(Buffer buffer, long offset) {
+    try {
+      if ( offset != pos_ ) {
+        throw new Exception("Offset does not match stream position");
```
adamvy, line 30: Should be RuntimeException?

src/foam/blob/HttpServletRequestBlob.java
```diff
@@ -0,0 +1,62 @@
...
+      }
+
+      int outOffset = 0;
+      long length = Math.min(buffer.getLength(), getSize() - offset);
+      if (length < buffer.getLength()) {
+        buffer = buffer.slice(0, length);
+      }
+
+      ByteBuffer bb = buffer.getData();
+      BufferedInputStream bufferedInputStream = new BufferedInputStream(request_.getInputStream());
```
adamvy, line 40: I don't think we should get the input stream every time, let's store it in a private/protected member variable and read from it.

We shouldn't do this decoration with BufferedInputStream every time .read() is called.  We should only do it once and store it in a member variable.

In fact we should fix this in FileReader as well.

src/foam/blob/HttpServletRequestBlob.java
kirkeaton: Shouldn't matter cause it should be caught by the catch block regardless.

src/foam/blob/HttpServletRequestBlob.java
kirkeaton: Thanks. I'll change that. So then there's no point in storing the whole request as a member variable then. Just the input stream and the content length.

kirkeaton:
Closing this to fix CLA issues

PR #850, opened 2017-11-20
Modified InputStreamBlob to accept a size as the 2nd parameter

kirkeaton:
1. Modified InputStream blob to accept the size as the 2nd parameter
2. Modified InputStreamBlob read function

### DIFF
--- a/src/foam/blob/InputStreamBlob.java
+++ b/src/foam/blob/InputStreamBlob.java
@@ -6,28 +6,50 @@
 
 package foam.blob;
 
-import com.google.common.io.ByteStreams;
-
+import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 
 public class InputStreamBlob
     extends foam.blob.AbstractBlob
 {
-  protected byte[] buffer_;
+  protected long size_;
+  protected BufferedInputStream reader_;
 
-  public InputStreamBlob(java.io.InputStream in) throws IOException {
-    buffer_ = ByteStreams.toByteArray(in);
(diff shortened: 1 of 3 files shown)


src/foam/blob/RestBlobService.java
```diff
@@ -117,14 +117,12 @@ public Blob find_(X x, Object id) {
 
       if ( connection.getResponseCode() == HttpURLConnection.HTTP_OK ) {
         is = connection.getInputStream();
-        blob = new InputStreamBlob(is);
+        blob = new InputStreamBlob(is, connection.getContentLengthLong());
```
adamvy: We should check getContentLenghtLong() != -1 and return an HTTP 400 perhaps with a reasonable error message if so.

kirkeaton: Thanks. Done.

src/foam/blob/InputStreamBlob.java
```diff
@@ -6,28 +6,50 @@
...
+    this.size_ = size;
+    this.reader_ = new BufferedInputStream(in);
   }
 
   @Override
   public Buffer read(Buffer buffer, long offset) {
-    int length = (int) (getSize() - offset);
-    return new Buffer(length, ByteBuffer.wrap(buffer_, (int) offset, length));
+    try {
+      int outOffset = 0;
```
adamvy: Since InputStream does not support seeking we should probably assert that offset is the next sequential offset after the previous read like we do for servletrequestblob.

kirkeaton: Good idea

PR #857, opened 2017-11-21
More swift features and fixes

mcarcaso:
Includes:
Socket support
Promised axiom support
Swift implementations for various boxes, sinks, and other models
Topic support
Proxy events (might not fully work yet)
A fix to node's SocketService to respect ReturnBox
toJSON support for enums
Remote axiom support


src/foam/box/SocketConnectBox.js
```diff
@@ -37,7 +42,20 @@ foam.CLASS({
             connectTo(this.address).then(function(s) {
               return this.RawSocketBox.create({ socket: s });
             }.bind(this));
-      }
+      },
+      swiftFactory: `
+let fut = Future<FObject>()
```
adamvy: I think for proper multithreading we should move this logic into socketservice, and add synchronizatino there so that multiple socket boxes to the same address receive the same socket.

src/foam/swift/net/RawSocketBox.js
```diff
@@ -0,0 +1,75 @@
...
+
+  methods: [
+    {
+      name: 'send',
+      swiftCode: `
+var replyBox = msg.attributes["replyBox"] as? Box
+if replyBox != nil {
+  let export = registry.register(nil, nil, replyBox!) as! SubBox
+  msg.attributes["replyBox"] = export
+  replyBox = ReplyBox_create([
```
adamvy: I don't think you did this right.  Why are you creating a new ReplyBox? What we want to do is update the message's replyBox to the export box, and the restore it after we've sent the message.

src/foam/box/SocketConnectBox.js
mcarcaso: Moved this into a getSocketBoxFuture method on SocketService that will handle not connecting to the same address twice.

src/foam/swift/net/RawSocketBox.js
mcarcaso: removed line 65 and the similar bug that exists on foam.box.RawSocketBox

PR #870, opened 2017-11-23
Menu Bar Fix

blakegreen32:
Fixed UI issue with Menu Bar, added table columns into the User model for the UserView page.


src/foam/nanos/auth/User.js
```diff
@@ -16,7 +16,7 @@ foam.CLASS({
   documentation: '',
 
   tableColumns: [
-    'id', 'enabled', 'firstName', 'lastName', 'organization', 'lastModified'
+    'id', 'enabled', 'firstName', 'lastName', 'organization', 'lastModified', 'email', 'type'
```
zahnno, line 19: tryn'a squeeze something in here bud?

PR #872, opened 2017-11-24
menuDAO

xuerongCode:
add UnknownFObject.java, UnknownFObjectParser.java.
fix DAOMenu
fix json in menus
add menuDao2 in servers


src/foam/lib/json/UnknownFObjectParser.java
```diff
@@ -0,0 +1,39 @@
...
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+package foam.lib.json;
+
+import foam.lib.parse.*;
+import foam.core.X;
+
+public class UnknownFObjectParser implements Parser {
```
kgrgreer, line 11: This doesn't work for all valid JSON. Ex.:
{key:"}"}
Open ticket to fix and assign to yourself. I'll explain how to fix.

src/services
```diff
@@ -31,6 +31,7 @@ p({"class":"foam.nanos.boot.NSpec", "name":"exportDriverRegistryDAO",     "lazy"
...
+p({"class":"foam.nanos.boot.NSpec", "name":"menuDAO2",                  "lazy":true,  "serve":true,  "serviceScript":"return new foam.dao.PMDAO(new foam.dao.JDAO(foam.nanos.menu.Menu.getOwnClassInfo(), \"menus\"));","client":"{\"of\":\"foam.nanos.menu.Menu\"}"})
```
kgrgreer: This should be named menuDAO so that the client will read it and the old one should be removed.

PR #878, opened 2017-11-28
MDAO find using primary key

amymeng5277:
- [x] 1. Add MDAO testcase 
- [x] 2. Add find_() in AltIndex 
- [x] 3. Refine compare() in PropertyInfo.js


amymeng5277:
I signed it!

src/foam/dao/index/AltIndex.java
```diff
@@ -55,22 +57,31 @@ public Object remove(Object state, FObject value) {
...
+    }
+    return plan.find(s[0], key);
+  }
+
   public FindPlan planFind(Object state, Object key) {
     Object[] s = toObjectArray(state);
     Plan bestPlan = NoPlan.instance();
 
-    for ( int i = 0 ; i < delegates_.size() ; i++ ) {
+    for (int i = 0; i < delegates_.size(); i++) {
```
kgrgreer: Here and on other lines you've broken the FOAM spacing guideline:

src/foam/java/PropertyInfo.js
```diff
@@ -97,7 +97,7 @@ foam.CLASS({
             type: 'int',
             visibility: 'public',
             args: [ { name: 'o1', type: 'Object' }, { name: 'o2', type: 'Object' } ],
-            body: 'return compareValues(get_(o1), get_(o2));'
+            body: 'return (o1 instanceof ' + this.sourceCls.name + ' ? compareValues(get_(o1), get_(o2)) : compareValues(cast(o1), cast(o2)));'
```
kgrgreer: In JS, we have both compare() and comparePropertyValues().
See: https://github.com/foam-framework/foam2/blob/master/src/foam/core/Property.js#L179

I don't mind also having a compareToPropertyValue or comparePropertyToObj which compares a property value to an object if needed.

src/foam/nanos/auth/Country.js
```diff
@@ -21,5 +21,14 @@ foam.CLASS({
       class: 'String',
       name: 'name'
     }
-  ]
+  ],
+  methods: [{
+    name: 'toString',
+    returns: 'String',
+    javaReturns: 'String',
```
kgrgreer, line 28: Does this mess up the JS toString()?

amymeng5277: I don’t think it will mess up js toString, since this is javaCode, right?

amymeng5277:
@kgrgreer  I have fixed all your suggestion please check again.

src/foam/java/PropertyInfo.js
```diff
@@ -103,8 +103,8 @@ foam.CLASS({
             name: 'comparePropertyToObject',
             type: 'int',
             visibility: 'public',
-            args: [ { name: 'key', type: 'Object' }, { name: 'o', type: 'foam.core.FObject' } ],
-            body: 'return compare(cast(key), get_(o));'
+            args: [ { name: 'key', type: 'Object' }, { name: 'o', type: 'Object' } ],
+            body: 'return foam.util.SafetyUtil.compare(cast(key), get_(o));'
```
kgrgreer, line 107: I think this could be better. Let's discuss tomorrow.

kgrgreer: I think this line should be: return compareValues(cast(key), get_(o));

PR #882, opened 2017-11-28
Disable background scrolling when pop up is open

blakegreen32:
When pop ups are open background scrolling is disabled and then re-enabled upon closing the pop up


kgrgreer:
I think this approach is fragile and limits the reusability of the Popup view since it can only be used in apps where the document body has overflow=scroll.

PR #889, opened 2017-12-04
Java multiton

mcarcaso:
Mostly untested but it compiles and nanos starts happily so I don't think it's doing any harm. Just want to see if you're ok with the approach I'm taking with this.

### DIFF
--- a/src/foam/core/ClassInfoImpl.java
+++ b/src/foam/core/ClassInfoImpl.java
@@ -61,10 +61,14 @@ public class ClassInfoImpl
     return parent_;
   }
 
-  public ClassInfo addProperty(PropertyInfo p) {
-    p.setClassInfo(this);
-    axioms.add(p);
-    axiomsByName_.put(p.getName(), p);
+  public ClassInfo addAxiom(Axiom a) {
+    // TODO: Should all axioms have setClassInfo? If not, create an interface
+    // that has setClassInfo and make PropertyInfo implement it.
+    if (a instanceof PropertyInfo) {
+      ((PropertyInfo)a).setClassInfo(this);
+    }
+    axioms.add(a);
+    axiomsByName_.put(a.getName(), a);
     return this;
   }
 
--- a/src/foam/core/ClassInfoImpl.java
+++ b/src/foam/core/ClassInfoImpl.java
@@ -64,7 +64,7 @@ public class ClassInfoImpl
   public ClassInfo addAxiom(Axiom a) {
     // TODO: Should all axioms have setClassInfo? If not, create an interface
     // that has setClassInfo and make PropertyInfo implement it.
-    if (a instanceof PropertyInfo) {
+    if ( a instanceof PropertyInfo ) {
       ((PropertyInfo)a).setClassInfo(this);
     }
     axioms.add(a);
--- a/src/foam/core/ContextFactory.java
+++ b/src/foam/core/ContextFactory.java
@@ -6,6 +6,7 @@
 
 package foam.core;
 
-public interface Factory<T> {
+public interface ContextFactory<T> {
   public T getInstance(java.util.Map<String, Object> args, X x);
 }
+
--- a/src/foam/core/MultitonInfo.java
+++ b/src/foam/core/MultitonInfo.java
@@ -10,7 +10,7 @@ import java.util.Map;
 import java.util.HashMap;
 
 public class MultitonInfo<T>
-  implements Axiom, Factory<T>
+  implements Axiom, ContextFactory<T>
 {
   Map<Object, T> instanceMap = new HashMap<Object, T>();
 
--- a/src/foam/core/MultitonInfo.java
+++ b/src/foam/core/MultitonInfo.java
@@ -26,14 +26,14 @@ public class MultitonInfo<T>
     return name;
   }
 
-  public T getInstance(Map<String, Object> args, X x) {
+  public synchronized T getInstance(Map<String, Object> args, X x) {
     Object key = args.get(p.getName());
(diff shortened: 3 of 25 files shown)


adamvy:
Seems reasonable.  I'm wondering if there's alternatives to using a Map for arguments, but I haven't come up with anything.

mcarcaso:
Yeah I’m torn too. Maps are kind of annoying to work with in java but I
think they’re the most intuitive option for what I’m trying to accomplish
here. Im definitely open to changing this as we see how it unfolds but at
the moment I can’t think of anything that’s any better than a map.

mcarcaso:
Renamed Factory to ContextFactory. I'll merge this in the morning if nobody has any objections.

src/foam/core/ClassInfoImpl.java
```diff
@@ -61,10 +61,14 @@ public ClassInfo getParent() {
...
   }
 
-  public ClassInfo addProperty(PropertyInfo p) {
-    p.setClassInfo(this);
-    axioms.add(p);
-    axiomsByName_.put(p.getName(), p);
+  public ClassInfo addAxiom(Axiom a) {
+    // TODO: Should all axioms have setClassInfo? If not, create an interface
+    // that has setClassInfo and make PropertyInfo implement it.
+    if (a instanceof PropertyInfo) {
```
kgrgreer: Follow FOAM style guideline for spacing.

src/foam/core/ContextFactory.java
```diff
@@ -0,0 +1,12 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.core;
+
+public interface ContextFactory<T> {
```
kgrgreer: We also have XFactory: https://github.com/foam-framework/foam2/blob/f383a842c33d46772cdc81ddd13b89eb0490221a/src/foam/core/XFactory.java

Context is called X everywhere else.

src/foam/core/MultitonInfo.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  public T getInstance(Map<String, Object> args, X x) {
+    Object key = args.get(p.getName());
+    if ( ! instanceMap.containsKey(key) ) {
+      try {
+        Class<T> type = (Class<T>)p.getClassInfo().getObjClass();
+        T obj = type.newInstance();
+        ((ContextAware)obj).setX(x);
+        for (Map.Entry<String, Object> entry : args.entrySet()) {
```
kgrgreer: spacing

src/foam/core/MultitonInfo.java
```diff
@@ -0,0 +1,47 @@
...
+  public MultitonInfo(String name, PropertyInfo p) {
+    this.name = name;
+    this.p = p;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public T getInstance(Map<String, Object> args, X x) {
```
kgrgreer: Isn't thread safe so could cause duplicate creates.

src/foam/core/SimpleFacetManager.java
```diff
@@ -15,17 +18,44 @@ public Object getInstanceOf(Object value, Class type, X x) {
...
+        ClassInfo classInfo = (ClassInfo)method.invoke(null);
+
+        Object f = null;
+        if ( x.get(classInfo.getId() + "_Factory") != null ) {
+          f = x.get(classInfo.getId() + "_Factory");
+        } else if ( classInfo.getAxiomsByClass(ContextFactory.class).size() == 1 ) {
+          f = classInfo.getAxiomsByClass(ContextFactory.class).get(0);
+        }
+
+        if (f != null) {
```
kgrgreer: spacing

src/foam/core/SimpleFacetManager.java
```diff
@@ -15,17 +18,44 @@ public Object getInstanceOf(Object value, Class type, X x) {
...
       if ( java.lang.reflect.Modifier.isAbstract(type.getModifiers()) ) {
         type = (Class<T>) Class.forName(type.getName() + "Impl");
       }
 
+      try {
+        java.lang.reflect.Method method = type.getMethod("getOwnClassInfo");
+        ClassInfo classInfo = (ClassInfo)method.invoke(null);
+
+        Object f = null;
+        if ( x.get(classInfo.getId() + "_Factory") != null ) {
```
kgrgreer: Add comments to explain what's going on.

src/foam/core/ContextFactory.java
mcarcaso: Yeah this is a little different than XFactory and I'm not sure what else to call this. Do you have any suggestions? XFObjectFactory?

src/foam/core/ClassInfoImpl.java
mcarcaso: Done

src/foam/core/MultitonInfo.java
mcarcaso: Done

src/foam/core/MultitonInfo.java
mcarcaso: Good eye. Made this method synchronized which should do the trick.

src/foam/core/SimpleFacetManager.java
mcarcaso: Done

src/foam/core/SimpleFacetManager.java
mcarcaso: Done. Let me know if it's still confusing.

PR #893, opened 2017-12-06
Fix Select Function in MDAO of AltIndex

amymeng5277:
1. Fixed select() function in MDAO. 
2. Add getAllValues() function in TreeNode to help add element from MDAO to Sink. 
3. Add AltFindPlan class and AltSelectPlan to make the plan have paired state.


src/foam/dao/index/AltFindPlan.java
```diff
@@ -0,0 +1,38 @@
...
+import foam.core.FObject;
+
+import foam.dao.Sink;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Not;
+import foam.mlang.predicate.Predicate;
+
+public class AltFindPlan implements FindPlan {
+  protected Object state_;
+  protected Plan bestPlan_ = new NotFoundPlan();
```
kgrgreer: Make this a FindPlan instead of a Plan, then you don't need to cast below.

src/foam/dao/index/AltSelectPlan.java
```diff
@@ -0,0 +1,40 @@
...
+import foam.core.FObject;
+
+import foam.dao.Sink;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Not;
+import foam.mlang.predicate.Predicate;
+
+public class AltSelectPlan implements SelectPlan {
+  protected Object state_;
+  protected Plan bestPlan_ = new NotFoundPlan();
```
kgrgreer: Make a SelectPlan to avoid casting below.

src/foam/dao/index/AltSelectPlan.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public AltSelectPlan(Object state, Plan bestPlan){
+    state_ = state;
+    bestPlan_ = bestPlan;
+  }
+
+  public Plan getPlan(){
```
kgrgreer: Why are these two getters needed?

src/foam/dao/index/TreeNode.java
```diff
@@ -345,4 +357,24 @@ public TreeNode lte(TreeNode s, Object key, PropertyInfo prop) {
       s.level, s.left, null);
   }
 
+  public Object[] getAllValues(Object s){
```
kgrgreer: Getting all values uses up memory for the Object[]. Passing in a sink lets each row be processed in sequence without accumulating any extra memory. So a better method would be:
  public void select(Object state, Sink sink)

src/foam/dao/index/TreeNode.java
```diff
@@ -345,4 +357,24 @@ public TreeNode lte(TreeNode s, Object key, PropertyInfo prop) {
...
+    Object[] nodeList = new Object[(int)((TreeNode)s).size];
+    getValuesHelper((TreeNode)s,nodeList,0);
+    return nodeList;
+  }
+  public int getValuesHelper(TreeNode currentNode, Object[] nodeList, int i){
+    if(currentNode.getValue()!=null)
+      nodeList[i]=(currentNode.getValue());
+    i++;
+    TreeNode left = currentNode.getLeft();
```
kgrgreer: Process left branch before processing getValue() so you get an in-order traversal (ie. sorted correctly).

src/foam/dao/index/TreePlan.java
```diff
@@ -0,0 +1,33 @@
...
+  protected TreePlan() {}
+
+  public long cost() { return 1; }
+
+  public FObject find(Object state, Object key) {
+    return null;
+  }
+
+  public void select(Object state, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+    Object[] nodeList = ((TreeNode)state).getAllValues(state);
```
kgrgreer: Better if TreeNode just had a select() method you could delegate to.

src/foam/dao/index/AltFindPlan.java
amymeng5277: Ok, I will change that.

src/foam/dao/index/AltSelectPlan.java
amymeng5277: Ok

src/foam/dao/index/AltSelectPlan.java
amymeng5277: Ok, it looks useless.

amymeng5277:
@kgrgreer  1. add select() function in TreeNode. 2. Change AltFindPlan and AltSelectPlan 3. use in-order traversal.

kgrgreer:
Much nicer.

PR #894, opened 2017-12-07
CSV parser

xuerongCode:
add CSVNormalStringParser.java, CSVEscapeStringParser.java, CSVStringParser.java, CSVsupport.java
change csvParser for String in the refinements.js


src/foam/lib/csv/CSVSupport.java
```diff
@@ -0,0 +1,83 @@
...
+
+  public void inputCSV(InputStream is, Sink sink, ClassInfo classInfo) {
+
+    try {
+      BufferedReader reader = new BufferedReader(new InputStreamReader(is, "UTF-8"));
+      String line = reader.readLine();
+      StringPStream ps = new StringPStream();
+      ps.setString(line);
+      ParserContext x = new ParserContextImpl();
+      x.set("X", getX());
```
kgrgreer, line 33: What is this for?

kgrgreer:
Please review: @adamvy

src/foam/lib/csv/CSVSupport.java
xuerongCode: Actually, PaserContext is useless in this case. I will delete it and pass null as ParserContext.

src/foam/lib/csv/CSVSupport.java
```diff
@@ -0,0 +1,83 @@
...
+import foam.core.PropertyInfo;
+import foam.core.FObject;
+import foam.lib.parse.*;
+import foam.dao.Sink;
+
+
+public class CSVSupport 
+  extends foam.core.ContextAwareSupport
+{
+  protected Parser headParser = new Parser() {
```
adamvy: You can just do headParser = new Repeat(new CSVSTtringParser(), new Literal(","));

No need for the anonymous class if all it does is call it's delegate.

src/foam/lib/csv/CSVEscapeStringParser.java
```diff
@@ -0,0 +1,74 @@
...
+
+    int delimCount = 0;
+    StringBuilder sb = builders.get();
+    ps = ps.tail();
+
+    while ( ps.valid() ) {
+      head = ps.head();
+      if ( head == ESCAPE ) {
+        if ( delimCount == 1) {
+          sb.append(ESCAPE);
```
adamvy, line 42: Do you really want to append the ESCAPE character to the string buffer?

It looks like a string `"foo"` will parse to `"foo"` and `foo` will parse to `foo`

But I think we'd rather have `foo` and `"foo"` both parse to `foo`

src/foam/lib/csv/CSVEscapeStringParser.java
```diff
@@ -0,0 +1,74 @@
...
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.lib.csv;
+
+import foam.lib.parse.*;
+
+
+public class CSVEscapeStringParser implements Parser {
```
adamvy: Can we document what makes this different from the JSON string parser?

src/foam/lib/csv/CSVEscapeStringParser.java
xuerongCode: "foo" will parse to foo
the first " will not be appended, same as the last one

src/foam/lib/csv/CSVEscapeStringParser.java
xuerongCode: Yes, I will do that

src/foam/lib/csv/CSVEscapeStringParser.java
```diff
@@ -0,0 +1,74 @@
...
+  public final static char ESCAPE = '"';
```
adamvy: Is double quote the correct character for ESCAPE?

src/foam/lib/csv/CSVSupport.java
adamvy: I would keep it in case we add support for parsing FObjectProperties from a CSV file.

src/foam/lib/csv/CSVEscapeStringParser.java
adamvy: I see, nvm

src/foam/lib/csv/CSVEscapeStringParser.java
adamvy: Should \r \n and others be interpreted to real carriage returns and line feeds?

src/foam/lib/csv/CSVNormalStringParser.java
```diff
@@ -8,6 +8,15 @@
...
 
+/**
+ * The class parse the CSV normal string base on the CSV string rule
+ * The string can not contain '"', '\r', '\n' and ','
+ * eg:
+ *  abc123    : legal
+ *  ab"e123   : illegal
+ *  ab\ne123  : illegal
+ *  ab\re123  : illegal
+ */
```
adamvy, line 19: So NormalStringParser is a strict subset of CSVEscapeStringParser.

What is the value is keeping NormalStringParser?  Do we ever need to be this strict when parsin?

src/foam/lib/csv/CSVEscapeStringParser.java
xuerongCode: No, \r\n will be parsed inside the CSVEscapeStringParser. I have modified the CSVNormalStringParser to avoid \r\n.

PR #902, opened 2017-12-14
 Modified sendEmailFromTemplate to accept user as the first parameter

kirkeaton:
1. Modified sendEmailFromTemplate to accept user as the first parameter
2. Made some modifications to SMTPEmailService to check if the user is null before accessing the group property
3. Modified DAOResourceLoader to search by group only if it is provided

### DIFF
--- a/src/foam/nanos/notification/email/DAOResourceLoader.java
+++ b/src/foam/nanos/notification/email/DAOResourceLoader.java
@@ -19,28 +19,27 @@ public class DAOResourceLoader
     implements ResourceLoader
 {
   public static EmailTemplate findTemplate(DAO dao, String templateName, String groupName) {
-    Sink list;
-    List data;
+    Sink list = new ListSink();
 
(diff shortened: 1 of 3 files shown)


src/foam/nanos/notification/email/DAOResourceLoader.java
```diff
@@ -15,20 +16,32 @@
...
 
-    List data = ((ListSink) list).getData();
-
-    if ( data.size() == 0 ) {
+    // if group is provided, query based on that
+    if ( ! SafetyUtil.isEmpty(groupName) ) {
+      list = dao.where(AND(
+          EQ(EmailTemplate.NAME, templateName),
+          EQ(EmailTemplate.GROUP, groupName))).limit(1).select(null);
+    } else {
```
kgrgreer: Don't do in an else, but instead, like before do:
if ( data.size() == 0 ) {

PR #905, opened 2017-12-18
graph view feature

hchoura:


src/files.js
```diff
@@ -409,6 +409,11 @@ FOAM_FILES([
   { name: "foam/u2/FoamTagLoader", flags: ['web'] },
   { name: "foam/graphics/CView", flags: ['web'] },
   { name: "foam/graphics/ScrollCView", flags: ['web'] },
+  { name: "foam/graphics/CViewGraph", flags: ['web'] },
```
kgrgreer: I would rather not add these files to files.js since most people won't use them. I would rather they get their own graph.js (like nanos.js) which people can add to their index.html when they need them.

PR #907, opened 2017-12-19
Checkbox properties feature

hchoura:


src/foam/doc/DocBrowser.js
```diff
@@ -223,7 +223,8 @@ foam.CLASS({
 
   imports: [
     'selectedAxiom',
-    'showInherited'
+    'showInherited',
+	  'checkBox'
```
kgrgreer: A more meaningful name like 'showOnlyProperties' would be better.

kgrgreer: Use spaces not tabs otherwise things are misaligned for others.

zahnno:
lgtm!

PR #911, opened 2017-12-19
Wrapped Geocoding in thread pool call

kirkeaton:


kgrgreer:
Don't pool the agent until after you delegate the put(). This will ensure the proper id is set and make sure your agent doesn't put() even if the original put() failed.

kirkeaton:
@kgrgreer Thanks made the modifications

PR #913, opened 2017-12-20
Auth Web Agent

zahnno:
PR includes:

Auth Web agent, which when decorated on a web agent, throws a login screen & permits authentication when credentials are correct.


src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -0,0 +1,106 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.http;
+
+import foam.core.*;
```
kgrgreer, line 9: sort imports

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -0,0 +1,106 @@
...
+  public void execute(X x) {
+    HttpServletRequest  req  = (HttpServletRequest) x.get(HttpServletRequest.class);
+    HttpServletResponse resp = (HttpServletResponse) x.get(HttpServletResponse.class);
+    PrintWriter         out  = (PrintWriter) x.get(PrintWriter.class);
+
+    DAO                 sessionDAO = (DAO) x.get("sessionDAO");
+
+    Cookie[] cookies = req.getCookies();
+    String sessionId = null;
+    if( cookies != null ){
```
kgrgreer: Follow FOAM coding standard for spacing.

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -0,0 +1,106 @@
...
+      for ( Cookie cookie : cookies ) {
+        if ( cookie.getName().toString().equals("sessionId" )){
```
kgrgreer: getName() is already a String.

kgrgreer: Make "sessionId" be a constant since it's repeated in two spots.

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -0,0 +1,106 @@
...
+          sessionId = cookie.getValue().toString();
+          Session session = (Session) sessionDAO.find(sessionId);
+          Date expiry = session.getExpiry();
+          Date dateNow = new Date();
```
kgrgreer: You don't need to worry about expiry because there's a cron job to remove expired sessions.

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -0,0 +1,106 @@
...
+          if ( session != null && expiry.before(dateNow) ) {
+            getDelegate().execute(x);
```
kgrgreer: Should be:
getDelegate().execute(session.getContext());

So that it runs in the User's context (which is more restricted / authenticated).

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -0,0 +1,106 @@
...
+      }
+    } catch (Throwable t) {
+      t.printStackTrace();
+    }
+  }
+
+  public void createCookie(Session session, X x){
+    HttpServletResponse resp  = (HttpServletResponse) x.get(HttpServletResponse.class);
+    PrintWriter        out  = (PrintWriter) x.get(PrintWriter.class);
+    Cookie cookie = new Cookie("sessionId", session.getId());
```
kgrgreer: Use constant.

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -0,0 +1,108 @@
...
+    Cookie[] cookies = req.getCookies();
+    String sessionId = null;
+    if( cookies != null ){
+      for ( Cookie cookie : cookies ) {
+        if ( cookie.getName().equals(SESSION_ID) ){
+          sessionId = cookie.getValue().toString();
+          Session session = (Session) sessionDAO.find(sessionId);
+
+          if ( session != null ) {
+            getDelegate().execute(session.getContext());
```
kgrgreer: We still need to call auth.check() to make sure that the user has permission to run this web agent.

PR #917, opened 2017-12-21
Replaced FileServlet with FileWebAgent

kirkeaton:
Files are now served at http://localhost:8080/service/static

1. Replaced FileServlet with FileWebAgent. 
2. Added setContentLengthLong to ResponseWrapper class


src/foam/nanos/http/FileWebAgent.java
```diff
@@ -0,0 +1,125 @@
...
+        resp.setContentType(EXTS.get("html"));
+        pw.write(
+            "<!DOCTYPE html>\n" +
+                "<html>\n" +
+                "<body>\n" +
+                "<ul style=\"list-style-type:disc\">");
+
+        File[] files = src.listFiles();
+        for ( File file : files ) {
+          pw.write("<li>" + "<a href=\"/service/" + nspec_.getName() + "/" + path
```
adamvy: We should make sure to escape special characters here otherwise we risk html/script injection.  Even though the filenames should be controlled by us.

src/foam/nanos/http/FileWebAgent.java
```diff
@@ -0,0 +1,125 @@
...
+        return;
+      }
+
+      // handle reading of files
+      if ( src.isFile() && src.canRead() && pathStartsWithCwd ) {
+        String ext = EXTS.get(FilenameUtils.getExtension(src.getName()));
+        is = new BufferedInputStream(new FileInputStream(src));
+
+        resp.setContentType(! SafetyUtil.isEmpty(ext) ? ext : DEFAULT_EXT);
+        resp.setHeader("Content-Disposition", "filename=\"" + src.getName() + "\"");
```
adamvy: Esacpe special characters again.

src/foam/nanos/http/FileWebAgent.java
```diff
@@ -0,0 +1,125 @@
...
+        return;
+      }
+
+      // file not found
+      throw new FileNotFoundException("File not found: " + path);
+    } catch (Throwable t) {
+      t.printStackTrace();
+      resp.setStatus(HttpServletResponse.SC_NOT_FOUND);
+      resp.setContentType(EXTS.get("json"));
+      pw.write("{\"error\": \"File not found\"," + "\"filename\": \"" + path + "\"}");
```
adamvy: Escape characters

src/foam/nanos/http/FileWebAgent.java
```diff
@@ -0,0 +1,125 @@
...
+    String path = null;
+    BufferedInputStream is = null;
+
+    try {
+      path = req.getRequestURI().replaceFirst("/?service/" + nspec_.getName() + "/?", "") + path_;
+      File src = new File(SafetyUtil.isEmpty(path) ? "./" : path);
+      boolean pathStartsWithCwd = src.getAbsolutePath().startsWith(cwd_);
+
+      // handle reading of directories
+      if ( src.isDirectory() && src.canRead() && pathStartsWithCwd ) {
```
adamvy: Not a critical change, but I think it's a safer style to early exit if pathStartsWithCd is false, rather than including it in every serving condition.

Otherwise if someone adds a case later that forgets to include && pathStartsWithCwd then we could have a security problem.

PR #918, opened 2017-12-21
Added inline auditing to JDAO

kirkeaton:
Added auditing to JDAO so that whenever a line is modified it prints a comment in the following format: /* Modified by [first name] [last name] ([user id]) at yyyy-MM-dd'T'HH:mm:ss'Z' */

### DIFF
--- a/src/foam/dao/JDAO.java
+++ b/src/foam/dao/JDAO.java
@@ -26,7 +26,7 @@ import java.util.regex.Pattern;
 public class JDAO
     extends ProxyDAO
 {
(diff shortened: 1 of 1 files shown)


src/foam/dao/JDAO.java
```diff
@@ -101,6 +121,16 @@ protected String getParsingErrorMessage(String line) {
     return eps.getMessage();
   }
 
+  protected void writeComment(User user) throws IOException {
+    out_.write("/* Modified by ");
```
kgrgreer: I think a // comment would be neater.

kirkeaton: But would it be spiffy keen?

kirkeaton: Done thanks.

PR #921, opened 2017-12-21
Added ProxyIndex

kirkeaton:


src/foam/dao/index/ProxyIndex.java
```diff
@@ -0,0 +1,70 @@
...
+import foam.core.FObject;
+import foam.dao.Sink;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+
+public class ProxyIndex
+    implements Index
+{
+  protected Index delegate_;
+  private boolean delegateIsSet_ = false;
```
kgrgreer: This is done in modelled objects to support default values, but isn't needed here. Just remove all of the delegateIsSet_ support.

kirkeaton: Done

PR #927, opened 2017-12-27
modified for DIG - JSON/J

minsunny:
Modified JDAO for the out in DIG.


kgrgreer:
I don't think you should modify the JDAO, just DIG. DIG should output data in the same format as the JDAO, but these modifications will make the JDAO slower even when not being used with DIG.

PR #928, opened 2017-12-28
Added LRUCachingIndex

kirkeaton:
1. Added LRUCachingIndex 
2. Added LRUCachingState which stores the key, value, and next and previous node
3. Added wrap and unwrap functions to Index interface
4. Created an AbstractIndex class that implements the default functionality for wrap & unwrap

### DIFF
--- a/src/foam/dao/index/LRUCachingIndex.java
+++ b/src/foam/dao/index/LRUCachingIndex.java
@@ -10,7 +10,7 @@ public class LRUCachingIndex
     extends ProxyIndex
 {
   protected int size_ = 0;
-  protected final int maxSize;
+  protected final int maxSize_;
 
   protected LRUCachingState head_ = new LRUCachingState();
   protected LRUCachingState tail_ = new LRUCachingState();
@@ -19,7 +19,7 @@ public class LRUCachingIndex
     setDelegate(index);
     this.head_.setNext(tail_);
     this.tail_.setPrev(head_);
-    this.maxSize = maxSize;
+    this.maxSize_ = maxSize;
   }
 
   @Override
@@ -31,9 +31,9 @@ public class LRUCachingIndex
 
     size_ += 1;
     // remove oldest entry
-    if ( size_ >= maxSize ) {
-      LRUCachingIndex toRemove = tail_.getPrev();
-      LRUCachingIndex prev = toRemove.getPrev();
+    if ( size_ >= maxSize_ ) {
+      LRUCachingState toRemove = tail_.getPrev();
+      LRUCachingState prev = toRemove.getPrev();
       prev.setNext(tail_);
       tail_.setPrev(prev);
       toRemove.setValue(null);
--- a/src/foam/dao/index/LRUCachingIndex.java
+++ b/src/foam/dao/index/LRUCachingIndex.java
@@ -31,11 +31,7 @@ public class LRUCachingIndex
 
     // remove oldest entry
     if ( size_ >= maxSize_ ) {
-      LRUCachingState toRemove = tail_.getPrev();
-      LRUCachingState prev = toRemove.getPrev();
-      prev.setNext(tail_);
-      tail_.setPrev(prev);
-      toRemove.setValue(null);
+      tail_.getPrev().remove();
     } else {
       size_++;
     }
--- a/src/foam/dao/index/LRUCachingIndex.java
+++ b/src/foam/dao/index/LRUCachingIndex.java
@@ -1,0 +1,15 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.dao.index;
+
+public class LRUCachingIndex
+    extends ProxyIndex
+{
+  public LRUCachingIndex(Index index) {
+    setDelegate(index);
+  }
+}
--- a/src/foam/dao/index/LRUCachingIndex.java
+++ b/src/foam/dao/index/LRUCachingIndex.java
@@ -46,19 +46,9 @@ public class LRUCachingIndex
       cache.setValue(getDelegate().unwrap(cache.getKey()));
     }
 
-    // get previous node of cached node
-    // set the next node of the previous node
-    // to the next node of the cached node
-    LRUCachingState prev = cache.getPrev();
-    prev.setNext(cache.getNext());
-
-    // get next node of cached node
-    // set the prev node of the next node
-    // to the prev node of the cached node
-    LRUCachingState next = cache.getNext();
-    next.setPrev(cache.getPrev());
-
-    // set cached node to be head
+    // remove cached node from location
+    // and set to be head
+    cache.remove();
     cache.setPrev(head_);
     cache.setNext(head_.getNext());
     head_.setNext(cache);
--- a/src/foam/dao/index/LRUCachingIndex.java
+++ b/src/foam/dao/index/LRUCachingIndex.java
@@ -9,13 +9,19 @@ package foam.dao.index;
 public class LRUCachingIndex
     extends ProxyIndex
 {
+  public static long
+
+  protected int size_ = 0;
+  protected final int maxSize;
+
   protected LRUCachingState head_ = new LRUCachingState();
   protected LRUCachingState tail_ = new LRUCachingState();
 
-  public LRUCachingIndex(Index index) {
+  public LRUCachingIndex(int maxSize, Index index) {
     setDelegate(index);
-    head_.setNext(tail_);
-    tail_.setPrev(head_);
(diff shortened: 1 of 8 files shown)


src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,75 @@
...
+
+package foam.dao.index;
+
+public class LRUCachingIndex
+    extends ProxyIndex
+{
+  public static long
+
+  protected int size_ = 0;
+  protected final int maxSize;
```
kgrgreer: Rename maxSize_. We follow the convention of have protected instance variables end in _.

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,75 @@
...
+  public LRUCachingIndex(int maxSize, Index index) {
+    setDelegate(index);
+    this.head_.setNext(tail_);
+    this.tail_.setPrev(head_);
+    this.maxSize = maxSize;
+  }
+
+  @Override
+  public Object wrap(Object state) {
+    LRUCachingState cache = new LRUCachingState(state, getDelegate().wrap(state));
```
kgrgreer: I think this is wrong. wrap(state) should be the key (ie. the file position) and value should be the original state.

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,75 @@
...
+      size_ -= 1;
+    }
+
+    return cache;
+  }
+
+  public Object unwrap(Object state) {
+    LRUCachingState cache = (LRUCachingState) state;
+    if ( cache.getValue() == null ) {
+      cache.setValue(getDelegate().unwrap(state));
```
kgrgreer: I think this is wrong. I think it should be unwrap(cache.key).

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    // get previous node of cached node
+    // set the next node of the previous node
+    // to the next node of the cached node
+    LRUCachingState prev = cache.getPrev();
+    if ( prev != null ) {
```
kgrgreer: I don't think this check is needed.

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,75 @@
...
+      prev.setNext(cache.getNext());
+    }
+
+    // get next node of cached node
+    // set the prev node of the next node
+    // to the prev node of the cached node
+    LRUCachingState next = cache.getNext();
+    if ( next != null ) {
```
kgrgreer: I don't think this check is needed.

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,75 @@
...
+      next.setPrev(cache.getPrev());
+    }
+
+    // set cached node to be head
+    cache.setPrev(head_);
```
kgrgreer: Also set cache.next

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,70 @@
...
+
+    size_ += 1;
+    // remove oldest entry
+    if ( size_ >= maxSize_ ) {
+      LRUCachingState toRemove = tail_.getPrev();
+      LRUCachingState prev = toRemove.getPrev();
+      prev.setNext(tail_);
+      tail_.setPrev(prev);
+      toRemove.setValue(null);
+      size_ -= 1;
```
kgrgreer: You could remove this line and just add the size++ above to the else.

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,70 @@
...
+    LRUCachingState cache = new LRUCachingState(getDelegate().wrap(state), state);
+    cache.setNext(head_.getNext());
+    cache.setPrev(head_);
+    head_.setNext(cache);
+
+    size_ += 1;
+    // remove oldest entry
+    if ( size_ >= maxSize_ ) {
+      LRUCachingState toRemove = tail_.getPrev();
+      LRUCachingState prev = toRemove.getPrev();
```
kgrgreer: Adding a remove() method to LRUCachingState would avoid some code duplication. That method would just connect next to prev and prev to next. Could be used here and where moving node to front of list.

src/foam/dao/index/LRUCachingIndexTest.java
```diff
@@ -0,0 +1,46 @@
...
+    for ( int i = 0; i < SAMPLE_SIZE; i++ ) {
+      samples[i] = generator.generate(50);
+    }
+
+    LRUCachingIndex index = new LRUCachingIndex(1000000,
+        new PersistedIndex("test", new TreeIndex(Country.CODE)));
+    long start = System.nanoTime();
+
+    for ( int i = 0; i < SAMPLE_SIZE; i++ ) {
+      index.wrap(samples[i]);
```
kgrgreer: Add some unwrap()'s also and then benchmark with and without the cache.

src/foam/dao/index/LRUCachingIndex.java
kirkeaton: Thanks. Done

src/foam/dao/index/LRUCachingIndex.java
kirkeaton: Added the remove function

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  public Object wrap(Object state) {
+    LRUCachingState cache = new LRUCachingState(getDelegate().wrap(state), state);
+    cache.setNext(head_.getNext());
+    cache.setPrev(head_);
+    head_.setNext(cache);
+
+    // remove oldest entry
+    if ( size_ >= maxSize_ ) {
+      tail_.getPrev().remove();
```
kgrgreer: Still need to null the value.

kirkeaton: The value is nulled in remove()

src/foam/dao/index/LRUCachingIndexTest.java
kirkeaton: I have a separate test which just sets the performance of the PersistedIndex. I'll add a benchmark for the unwraps too though.

src/foam/dao/index/LRUCachingState.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public void setPrev(LRUCachingState prev) {
+    this.prev_ = prev;
+  }
+
+  public void remove() {
+    if ( this.prev_ != null ) this.prev_.setNext(this.next_);
+    if ( this.next_ != null ) this.next_.setPrev(this.prev_);
+    this.value_ = null;
```
kgrgreer: Shouldn't null the value here because this could also be called when a state is moved to the front of the list.

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,66 @@
...
+  public Object unwrap(Object state) {
+    LRUCachingState cache = (LRUCachingState) state;
+    if ( cache.getValue() == null ) {
+      cache.setValue(getDelegate().unwrap(cache.getKey()));
+    }
+
+    // get previous node of cached node
+    // set the next node of the previous node
+    // to the next node of the cached node
+    LRUCachingState prev = cache.getPrev();
```
kgrgreer: Next four lines could be replaced by remove() if remove() didn't also null value.

kirkeaton: Done

src/foam/dao/index/LRUCachingState.java
kirkeaton: Modified to not null the value

src/foam/dao/index/LRUCachingIndex.java
```diff
@@ -0,0 +1,57 @@
...
+  public Object wrap(Object state) {
+    LRUCachingState cache = new LRUCachingState(getDelegate().wrap(state), state);
+    cache.setNext(head_.getNext());
+    cache.setPrev(head_);
+    head_.setNext(cache);
+
+    // remove oldest entry
+    if ( size_ >= maxSize_ ) {
+      tail_.getPrev().remove();
+      tail_.getPrev().setValue(null);
```
kgrgreer: This line needs to move above the previous one because after its been removed, tail_.getPrev() is a different object.

kirkeaton: Thanks. Done.

PR #933, opened 2018-01-03
Add GroupBy logic and Count logic

amymeng5277:
1. GroupBy(PropertyInfo,GroupBy(PropertyInfo,count)) still need deepCopy.
2. GroupBy with no Index worked well.


src/foam/dao/index/GroupByPlan.java
```diff
@@ -0,0 +1,79 @@
...
+  protected boolean needGroupBy = true;
+
+  public GroupByPlan(Object state, Sink sink, long skip, long limit, Comparator order, Predicate predicate, PropertyInfo propertyInfo, Index tail) {
+    state_ = state;
+    skip_ = skip;
+    limit_ = limit;
+    order_ = order;
+    cost_ = calculateCost(propertyInfo);
+    tail_ = tail;
+    if ( sink instanceof GroupBy ) {
```
kgrgreer: Why would we be using a GroupByPlan if the sink weren't a GroupBy?

src/foam/mlang/MLang.java
```diff
@@ -58,6 +58,17 @@ public static Predicate AND(Predicate... args) {
     return new And(args);
   }
 
+  public static Sink GROUPBY(Object o1, Object o2) {
```
kgrgreer: Rename to GROUP_BY to be compatible with the JS version.

src/foam/dao/index/GroupByPlan.java
amymeng5277: Ok, I will fix that.

src/foam/mlang/MLang.java
amymeng5277: Ok.

src/foam/mlang/mlang.js
```diff
@@ -1765,8 +1765,12 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'Object',
+      class: 'Double',
```
kgrgreer: Doesn't this prevent it from working with dates and also convert the type from int/long fields to double?

src/foam/mlang/MLang.java
```diff
@@ -82,7 +82,7 @@ public static Sink MAX(Object o1) {
   }
 
   public static Sink MIN(Object o1) {
-    return new Min(0, MLang.prepare(o1));
+    return new Min(Double.MAX_VALUE, null, MLang.prepare(o1));
```
kgrgreer: I think it would be better to make the initial value be null and then have the compare take the first put() value as the min value if the current value is null.

src/foam/dao/index/GroupByPlan.java
```diff
@@ -0,0 +1,79 @@
...
+      if ( ( (GroupBy) sink ).getArg1().toString().equals(propertyInfo.toString()) ) {
```
kgrgreer: I would do this check in the TreeIndex before creating the GroupByPlan, since if it's on the wrong property, then there's no use in using a GroupByPlan.

src/foam/dao/index/TreeIndex.java
```diff
@@ -136,6 +136,8 @@ public SelectPlan planSelect(Object state, Sink sink, long skip, long limit, Com
     Object[] statePredicate = simplifyPredicate(state, predicate);
     state = statePredicate[0];
     predicate = (Predicate) statePredicate[1];
+    if ( predicate == null && sink instanceof GroupBy )
```
kgrgreer: Also check that it's the correct property in the GroupBy. Also, simplify/eliminate the comparator before creating the GroupByPlan.

src/foam/dao/index/TreeNode.java
```diff
@@ -383,6 +383,40 @@ protected void select_(TreeNode currentNode, Sink sink, long skip, long limit, l
     }
   }
 
+  protected long[] groupBy(TreeNode currentNode, Sink sink, long skip, long limit, long size, Index tail) {
```
kgrgreer: I don't think skip and limit work well with GroupBy, or getting it to work properly would be extremely complicated. Which is fine, because I've never seen them used together anyway. I would make the GroupByPlan only be used if skip and limit aren't set. Otherwise, use a SkanPlan.

src/foam/mlang/MLang.java
```diff
@@ -58,6 +58,17 @@ public static Predicate AND(Predicate... args) {
     return new And(args);
   }
 
+  public static Sink GROUP_BY(Object o1, Object o2) {
```
kgrgreer: o2 should be a Sink?

kgrgreer: o1 should be an Expr

src/foam/mlang/mlangJava.js
```diff
@@ -954,8 +954,10 @@ foam.CLASS({
         }
       ],
       javaCode: function() {
-/*if (obj.compareTo(this.getValue()) > 0) {
-  this.setValue(obj);
+/*
```
kgrgreer: if ( getValue() == null || getArg1().f(obj).compareTo(getValue()) < 0 ) setValue(getArg1().f(obj));

src/foam/dao/index/TreeIndex.java
```diff
@@ -136,7 +137,9 @@ public SelectPlan planSelect(Object state, Sink sink, long skip, long limit, Com
     Object[] statePredicate = simplifyPredicate(state, predicate);
     state = statePredicate[0];
     predicate = (Predicate) statePredicate[1];
-    if ( predicate == null && sink instanceof GroupBy )
+    if ( predicate == null && sink instanceof GroupBy
+        && ( (GroupBy) sink ).getArg1().toString().equals(prop_.toString())
+        && order == null && skip == 0 && limit == AbstractDAO.MAX_SAFE_INTEGER )
       return new GroupByPlan(state, sink, skip, limit, order, predicate, prop_, tail_);
```
kgrgreer: Don't pass skip and limit to GroupByPlan since it doesn't use it.

amymeng5277: Ok, I will fix that.

PR #934, opened 2018-01-03
Xtree

xuerongCode:
functional tree context, and Context toString() method


src/foam/core/EmptyX.java
```diff
@@ -62,27 +78,70 @@ public Object getInstanceOf(Object value, Class type) {
...
+  protected void setLeftParent(X leftParent) { leftParent_ = leftParent; }
+  protected void setRightParent(X rightParent) { rightParent_ = rightParent; }
 
   public Object get(X x, Object key) {
-    return key.equals(key_) ? value_ : parent().get(key);
+    String okey = key_.toString();
+    String nkey = key.toString();
+    if ( nkey.compareTo(okey) == 0 ) {
+      return value_;
+    } else if ( nkey.compareTo(okey) < 0 ) {
```
kgrgreer: 'else' not needed since you've already returned the line above.

src/foam/core/EmptyX.java
```diff
@@ -62,27 +78,70 @@ public Object getInstanceOf(Object value, Class type) {
...
+      return getLeftParent().get(x, key);
+    } else {
```
kgrgreer: 'else' not needed since you've already returned the line above.

src/foam/core/EmptyX.java
```diff
@@ -91,24 +150,38 @@ public Object get(X x, Object key) {
 class FactoryXI
   extends AbstractX
```
kgrgreer: Easier if this extends XI or AbstractXI since they share common code.

src/foam/core/EmptyX.java
```diff
@@ -62,27 +78,70 @@ public Object getInstanceOf(Object value, Class type) {
...
+    AbstractX x = null;
+    try {
+      x = (AbstractX) super.clone();
+    } catch ( CloneNotSupportedException e ) {
+      e.printStackTrace();
+    }
+    return x;
+  }
+
+  abstract protected Object getKey();
```
kgrgreer: These methods shouldn't be here since not all X's are binary-tree like.

src/foam/core/EmptyX.java
```diff
@@ -62,27 +78,70 @@ public Object getInstanceOf(Object value, Class type) {
...
+  abstract protected X getLeftParent();
```
kgrgreer: The naming is confusing since binary trees usually have children not parents.

src/foam/core/EmptyX.java
```diff
@@ -124,4 +197,21 @@ private EmptyX() {}
   public static X instance() { return x_; }
 
   public Object get(X x, Object key) { return null; }
+
+  public X put(Object key, Object value) {
+    return new XI(this, this, key, value);
```
kgrgreer: This doesn't look correct.

src/foam/core/EmptyX.java
```diff
@@ -124,4 +197,21 @@ private EmptyX() {}
...
+  }
+
+  public X putFactory(Object key, XFactory factory) {
+    return new FactoryXI(this, this, key, factory);
```
kgrgreer: This doesn't look correct.

src/foam/core/EmptyX.java
```diff
@@ -23,7 +23,7 @@
 // TODO: make this a functional tree rather than a linked list. (for performance)
 
 abstract class AbstractX
-  implements X
+  implements X, Cloneable
```
kgrgreer: Why do you need to implement Cloneable?

src/foam/core/EmptyX.java
```diff
@@ -64,51 +56,128 @@ public Object getInstanceOf(Object value, Class type) {
...
+  implements Cloneable
 {
-  final X      parent_;
-  final Object key_;
-  final Object value_;
+  protected X   leftChild_;
+  protected X   rightChild_;
+  
+  protected abstract Object getKey();
+  protected X getLeftChild() { return leftChild_; }
```
kgrgreer, line 67: Why do these methods need to be in AbstractX? Couldn't we have an AbstractTreeX or AbstractXI which XI and FactoryXI extend? This way people can make other X implementations which don't follow this structure.

src/foam/core/EmptyX.java
```diff
@@ -124,4 +193,15 @@ private EmptyX() {}
   public static X instance() { return x_; }
 
   public Object get(X x, Object key) { return null; }
+
+  public X put(Object key, Object value) {
+    return new XI(this, this, key, value);
```
kgrgreer, line 198: Is this correct?

src/foam/core/EmptyX.java
```diff
@@ -124,4 +193,15 @@ private EmptyX() {}
...
+  }
+
+  public X putFactory(Object key, XFactory factory) {
+    return new FactoryXI(this, this, key, factory);
```
kgrgreer, line 202: Is this correct?

src/foam/core/EmptyX.java
xuerongCode: ??, It is in AbstractXI already

src/foam/core/EmptyX.java
xuerongCode: When context itself is EmptyX, or context can not find key until reaches the EmptyX, the EmptyX will directly create a XI or FactoryXI and make all children pointer point EmptyX, and return the reference to its parent. The cloned parent will hold the reference to access this child. In this case, all children pointers of the leafs will point to the EmptyX.

PR #935, opened 2018-01-04
Added highlighting on the view

minsunny:
Added highlighting on this view, 
if permissions which has asterisk (*) are checked, highlighted all the permissions of them.


src/foam/nanos/auth/Group.js
```diff
@@ -66,19 +66,24 @@ foam.CLASS({
...
-      `
+          }
+        }
+        return false;`
+      ,
+      code: function(permissionId, permission, group) {
+        if ( this.permissions == null ) return false;
+
+        for ( var i = 0 ; i < this.permissions.length ; i++ ) {
+          if ( group.permissions[i].id.endsWith('*') ) {
```
kgrgreer: What is the endsWith() check needed for?

src/foam/nanos/auth/Permission.js
```diff
@@ -23,17 +23,10 @@
...
+         var prefix = gPermissionId.substring(0, gPermissionId.length-1);
 
-         // Check that is a valid permission string (should be done as a property validator)
-         if ( prefix.length && ! prefix.endsWith('.') ) return false;
-
-         return permission.startsWith(prefix);
-       }
-
-       return permission == this.id;
+         return permissionId.startsWith(prefix);
```
kgrgreer: This check is too simple. Suppose you own account #1 so I give you permission account.read.1 so you can read your account balance. But consider if I have account 10. Just doing startsWith() would give your permission to read my account also, since account.read.10 startsWith account.read.1.

src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2018

src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -44,27 +37,30 @@ foam.CLASS({
...
-            .start('th').add(p.id).end()
-            .select(self.groupDAO, function(g) {
-              var cb = foam.u2.CheckBox.create({data: self.checkPermissionForGroup(p.id, g)});
+            .start('td').style({'text-align': 'left', 'width': '480'}).add(p.id).end()
+            .select(self.groupDAO.orderBy(self.Group.ID), function(g) {
+              var cb = foam.u2.md.CheckBox.create({data: self.checkPermissionForGroup(p.id, g)});
               cb.data$.sub(function() { self.updateGroup(p, g, cb.data); });
-              this.start('td').style({'text-align': 'center', 'width': '130', 'margin-bottom': '30'}).tag(cb).end();
+              this.start('td').style({'text-align': 'center', 'width': '100'}).tag(cb).call(function() {
+              if ( g.implies(p.id, p, g)  ) { cb.style({'border-color': '#40C75B'}) };
```
kgrgreer: indent this line

src/foam/nanos/auth/Group.js
```diff
@@ -66,19 +66,22 @@ foam.CLASS({
...
-      `
+          }
+        }
+        return false;`
+      ,
+      code: function(permissionId, permission, group) {
+        if ( this.permissions == null ) return false;
+
+        for ( var i = 0 ; i < this.permissions.length ; i++ ) {
+          return group.permissions[i].implies(group.permissions[i].id, permissionId);
```
kgrgreer: Why are you passing group.permission[i].id to group.permission[i].implies()?, since it's the same object, it already knows its own .id, so doesn't need to be told.

src/foam/nanos/auth/Permission.js
```diff
@@ -23,17 +23,17 @@
...
 
          // Check that is a valid permission string (should be done as a property validator)
          if ( prefix.length && ! prefix.endsWith('.') ) return false;
 
-         return permission.startsWith(prefix);
+         return permissionId.startsWith(prefix);
        }
 
-       return permission == this.id;
+       return false;
```
kgrgreer: This line needs to be return permissionId == this.id; in case the permission is an exact match.

PR #944, opened 2018-01-09
File Relocator/Formatter

mcarcaso:
Note that the FileRelocator model was put through itself so that's what a formatted model looks like.

I plan to put Element.js through this soon but it has some FObject instances in the model which will cause things to break so I need to fix those.


demos/filerelocator/src/foam/tools/FileRelocator.js
```diff
@@ -0,0 +1,140 @@
...
+            }
+            m.name = name;
+            m.package = package;
+            files[path] = {
+              path: path,
+              dir: dir,
+              contents: `
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
adamvy, line 105: I don't think we can just stamp all files with a newer copyright date.

adamvy: Actually do we even need to include copyright in the output?

mcarcaso: Good question that I don't know the answer to. Happy to remove it but I'm currently using this on files we end up checking in (I used it on itself for funsies)

src/foam/core/JSON.js
```diff
@@ -164,6 +170,11 @@ foam.CLASS({
       name: 'propertyPredicate',
       value: function(o, p) { return ! p.transient; }
     },
+    {
+      class: 'Function',
+      name: 'objectKeyValuePredicate',
```
kgrgreer, line 175: Document

PR #945, opened 2018-01-09
Blob refactor

kirkeaton:
1. Changed Long variables to Int
2. Removed Java Buffer refinement
3. Modified Java Blob interface to replace Buffer with OutputStream and add a length parameter
4. Re-added InputStreamBlob
5. Modified FileBlob and ByteArrayBlob to be ProxyBlob's that delegate everything to InputStreamBlob
6. Modified RestBlobService & HttpBlobService to work with new interface.

### DIFF
--- a/src/foam/blob/Blob.js
+++ b/src/foam/blob/Blob.js
@@ -22,13 +22,11 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'Long',
+      class: 'Int',
       name: 'length'
     },
     {
-      class: 'Object',
       name: 'data',
-      javaType: 'java.nio.ByteBuffer',
       factory: function() {
         return new ArrayBuffer(this.length);
       }
@@ -55,7 +53,7 @@ foam.INTERFACE({
           name: 'buffer',
         },
         {
-          class: 'Long',
+          class: 'Int',
           swiftType: 'Int',
           name: 'offset'
         }
@@ -63,7 +61,7 @@ foam.INTERFACE({
     },
     {
       name: 'getSize',
-      returns: 'Long',
+      returns: 'Int',
       swiftReturns: 'Int',
     }
   ]
@@ -275,11 +273,11 @@ foam.CLASS({
       name: 'parent',
     },
     {
-      class: 'Long',
+      class: 'Int',
       name: 'offset'
     },
     {
-      class: 'Long',
+      class: 'Int',
       name: 'size',
       assertValue: function(value) {
         foam.assert(this.offset + value <= this.parent.size, 'Cannot create sub blob beyond end of parent.');
@@ -420,7 +418,7 @@ foam.CLASS({
       name: 'fd'
     },
     {
-      class: 'Long',
+      class: 'Int',
       name: 'size',
       expression: function(fd) {
         return require('fs').fstatSync(fd).size;
--- a/src/foam/blob/HashingOutputStream.java
+++ b/src/foam/blob/HashingOutputStream.java
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
 package foam.blob;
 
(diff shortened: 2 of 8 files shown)


src/foam/blob/HashingOutputStream.java
```diff
@@ -0,0 +1,47 @@
+package foam.blob;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+public class HashingOutputStream
```
kgrgreer: Add some docs.

src/foam/nanos/blob/HttpBlobService.java
```diff
@@ -55,22 +55,19 @@ public void execute(X x) {
   }
 
   protected void download(HttpServletRequest req, HttpServletResponse resp) {
+    OutputStream os = null;
+
     try {
       String path = req.getRequestURI();
-      String id = path.substring(path.lastIndexOf("/") + 1);
+      String id = path.replaceFirst("/service/blobService/", "");
```
kgrgreer: Might someone ever want to mount this somewhere else? If so, you can find out where you're mounted by implementing NSpecAware so that you're provided your NSpec so that you can know where you're mounted. If not, then this is fine.

src/foam/blob/HashingOutputStream.java
kirkeaton: Done. Also added missing license header to HashingOutputStream and InputStreamBlob

src/foam/nanos/blob/HttpBlobService.java
kirkeaton: Good point. Made the fix.

src/foam/blob/Blob.js
```diff
@@ -22,13 +22,11 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'Long',
+      class: 'Int',
```
adamvy, line 25: Switching these all to Ints will limit us to Blobs under 4GB will it not?  Why the change?

src/foam/blob/RestBlobService.java
```diff
@@ -59,23 +59,19 @@ public Blob put_(X x, Blob blob) {
...
+      int chunk = 0;
+      int size = blob.getSize();
+      int chunks = (int) Math.ceil((double) size / (double) BUFFER_SIZE);
+
       while ( chunk < chunks ) {
-        buffer = blob.read(buffer, chunkOffset(chunk));
-        byte[] buf = buffer.getData().array();
-        os.write(buf, 0, (int) buffer.getLength());
-        buffer.getData().clear();
+        blob.read(os, chunk * BUFFER_SIZE, size);
```
adamvy, line 71: "size" is the total size of the blob, but we're trying to read it in chunks of size BUFFER_SIZE.

The third parameter should be ( size % BUFFER_SIZE == 0 ? BUFFER_SIZE : size % BUFFER_SIZE ) or similar.

If size % BUFFER_SIZE is 0, then all chunks are the same size, if non-zero then it should give you the size of the final chunk.

src/foam/blob/Blob.js
kirkeaton: I was noticing in a lot of cases previously that I was casting down to an integer anyway because the read & write functions of InputStream / OutputStream respectively don't take long's as the offset and length parameters. But beyond that do we ever envision having to upload a file larger than 4GB? If that's the case then we can modify this.

src/foam/blob/RestBlobService.java
kirkeaton: I think this can actually just be simplified to ```blob.read(os, 0, size);```. The read function can perform it's own internal chunking.

adamvy: I agree

src/foam/blob/Blob.js
adamvy: I have use cases in other work and at home where I'll want to deal with larger files :)

PR #946, opened 2018-01-10
Add factory to Constant and use it in Element.js

mcarcaso:
This is in preparation to splitting up the Element.js file using the FileRelocator tool ( https://github.com/foam-framework/foam2/pull/944 )


src/foam/u2/Element.js
```diff
@@ -528,90 +528,140 @@ foam.CLASS({
...
+      `,
+      name: 'PSEDO_ATTRIBUTES',
+      value: {
+        value: true,
+        checked: true
+      },
     },
 
-    DEFAULT_VALIDATOR: foam.u2.DefaultValidator.create(),
+    {
```
kgrgreer, line 573: Is this so it can be serialized?

mcarcaso: Yeah otherwise it gets serialized as:
DEFAULT_VALIDATOR: { class: 'foam.u2.DefaultValidator' }
And with that, there's no way of knowing for sure whether this should be converted to an FObject or left as a regular Object so making it use a factory makes it clear what the value should be (and it can be lazier this way).

PR #948, opened 2018-01-10
Scroll fix

kennyqiyenkan:
- Handle size was fixed before which made it seem that the drag and scroll was not working.
  - only occurs when you have really small data sizes (2 pages or so)
- added missing click to scroll


kgrgreer:
FYI: @jlhughes

PR #952, opened 2018-01-11
fixed bug for selecting EmailTemplateDAO 

minsunny:
https://github.com/nanoPayinc/NANOPAY/issues/531

changed OutputterMode to be NETWORK 
added try catch when parse XML array


src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -210,13 +210,15 @@ public void put(FObject o, Detachable d) {
...
 
         if ( "json".equals(format) ) {
-          foam.lib.json.Outputter outputterJson = new foam.lib.json.Outputter();
+          foam.lib.json.Outputter outputterJson = new foam.lib.json.Outputter(OutputterMode.NETWORK);
           outputterJson.output(sink.getArray().toArray());
 
           if ( email.length != 0 && !email[0].equals("") && email[0] != null ) {
             output(x, outputterJson.toString());
           } else {
+            out.println("<textarea style=\"width:700;height:400;\" rows=10 cols=120>");
```
kgrgreer: Don't wrap in a textarea because, while this makes the output look better in the GUI, it will make it difficult for programs to parse, which is the primary use-case for DIG. Instead, set the HTTResponse content type to get it to display properly without the textarea wrapper.

Ex.: response.setContentType("application/json");

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -264,22 +266,24 @@ public void put(FObject o, Detachable d) {
...
+          String dataArray[] = outputterJson.toString().split("\"}");
           for ( int k = 0 ; k < dataArray.length; k++ ) {
-            dataToString += "p(" + dataArray[k] + "})\n";
+            dataToString += "p(" + dataArray[k] + "\"})\n";
           }
 
           if ( email.length != 0 && !email[0].equals("") && email[0] != null ) {
             output(x, dataToString);
           } else {
+            out.println("<textarea style=\"width:700;height:400;\" rows=10 cols=120>");
```
kgrgreer: Set content type like above.

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -244,7 +246,7 @@ public void put(FObject o, Detachable d) {
             out.println("</textarea>");
```
kgrgreer: Set contentType to "text/plain" instead of using textarea.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -63,10 +63,17 @@ public Object fromXML(X x, XMLStreamReader reader) {
   public void toXML (FObject obj, Document doc, Element objElement) {
     if ( this.f(obj) == null ) return;
 
+    try {
```
kgrgreer: Add a comment to explain what this is for.

minsunny: @kgrgreer When select EmailTemplateDAO It throws error caused by failing to cast from FObject to Object[], even the FObject parameter is not NULL, or invalid. I guess inside EmailTemplateDAO It has some a bunch of special characters not to cast. So I added above.

PR #960, opened 2018-01-15
AppConfig Mode

kennyqiyenkan:
Adds an enum property to AppConfig to determine what mode the app is running on.


src/foam/nanos/app/Mode.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kirkeaton: Should be Copyright 2018

kirkeaton (approved):

PR #962, opened 2018-01-16
Added ConcurrentTestRunner

kirkeaton:
1. Added ConcurrentTestRunner class
2. Added ConcurrentTest model


src/foam/nanos/bench/BenchmarkRunner.java
```diff
@@ -0,0 +1,145 @@
...
+          latch.countDown();
+        }
+      });
+      // start the thread
+      thread.start();
+    }
+
+    try {
+      // wait until latch reaches 0, or until timeout is reached
+      if (timeout_ != 0) {
```
kgrgreer, line 131: spacing

src/foam/nanos/bench/BenchmarkRunner.java
```diff
@@ -0,0 +1,145 @@
...
+
+    // set up the test
+    test_.setup(x);
+
+    // get start time
+    long startTime = System.currentTimeMillis();
+
+    // execute all the threads
+    for ( Thread thread : threads ) {
+      thread = new Thread(new Runnable() {
```
kgrgreer, line 115: Create a ThreadGroup for the test and create all threads in that group.
Provide a toString() for the threads.
Both of these will make it easier when looking at thread dumps.

src/foam/nanos/bench/BenchmarkRunner.java
```diff
@@ -0,0 +1,145 @@
...
+        latch.await(timeout_, TimeUnit.MILLISECONDS);
+      } else {
+        latch.await();
+      }
+    } catch (Throwable ignored) {}
+
+    // calculate length taken
+    long endTime = System.currentTimeMillis();
+    long duration = endTime - startTime;
+    System.out.println(threadCount_ +
```
kgrgreer, line 141: I would like to have a compound TPS and transactions per thread.

src/foam/nanos/bench/BenchmarkRunner.java
kirkeaton: What should the toString for the threads return?

PR #963, opened 2018-01-16
BenchmarkRunner modifications

kirkeaton:
1. Made use of ThreadGroup
2. Custom toString for each thread
3. Added name to BenchmarkRunner
4. Printing transaction per second and transaction per second per thread
5. Modified TransactionBenchmark


src/foam/nanos/bench/BenchmarkRunner.java
```diff
@@ -111,17 +128,24 @@ public void execute(final X x) {
...
-      thread = new Thread(new Runnable() {
+    for ( int i = 0 ; i < threadCount_ ; i++ ) {
+      final int tno = i;
+      Thread thread = new Thread(group, new Runnable() {
         @Override
         public void run() {
-          for ( int i = 0 ; i < invocationCount_ ; i ++ ) {
+          for ( int j = 0 ; j < invocationCount_ ; j++ ) {
             test_.execute(x);
+            count.incrementAndGet();
```
adamvy: This seems like it will add unnecessary contention between the threads.  Now they all have to synchronize on a single integer just so we can count how many were completed.  Can't we just calculate that upfront?  We know its going to be invocationCount_ * threadCount_ unless one of them throws an error and doesn't complete all their work.

adamvy: On the thought of throwing errors, maybe we should try/catch test_.execute and abort the whole benchmark if something goes wrong.

kgrgreer: Yes, a heavily contested lock like this is really going to throw off the numbers.

PR #966, opened 2018-01-17
Fix Menu Item Selection

kennyqiyenkan:
- Adds currentMenu to Application Controller
- Adds menuListener to Application Controller
- Adds use of menuListener in Menu.
- Adds logic and use of currentMenu in MenuBar to determine which
menuItem should have the focus

UX has been reviewed (but can still be changed in the future)


kirkeaton (approved):
LGTM

PR #967, opened 2018-01-17
Fix the bug of predicate will change after select

amymeng5277:


src/foam/dao/MDAO.java
```diff
@@ -69,19 +69,29 @@ public FObject find_(X x, Object o) {
 
   public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
     SelectPlan plan;
-    if ( predicate != null ) predicate = predicate.partialEval();
-    if ( predicate instanceof Or ) {
-      int length = ( (Or) predicate ).getArgs().length;
+    Predicate copyPredicate = null;
+    if ( predicate != null ) {
+      copyPredicate = (Predicate) ( (FObject) predicate ).deepClone();
```
kgrgreer: Why do you need to clone?

amymeng5277: Because the predicate will change after we deal with it or did partialEval()

kgrgreer: partialEval should create a new predicate, not mutate itself

amymeng5277: So you mean we don't need use predicate = predicate.partialEval(), we use it directly? Actually we return a new predicate.

PR #971, opened 2018-01-17
Took out create and edit buttons from all invoices page

blakegreen32:
Added callIf statement in DAOController to take out create and edit buttons from all invoices page


src/menus
```diff
@@ -7,6 +7,7 @@ p({"class":"foam.nanos.menu.Menu","id":"export","label":"Export Drivers","handle
 p({"class":"foam.nanos.menu.Menu","id":"groups","label":"Groups","handler":{"class":"foam.nanos.menu.DAOMenu","daoKey":"groupDAO"},"order":1000,"parent":"admin"})
 p({"class":"foam.nanos.menu.Menu","id":"lang","label":"Languages","handler":{"class":"foam.nanos.menu.DAOMenu","daoKey":"languageDAO"},"order":1000,"parent":"admin"})
 p({"class":"foam.nanos.menu.Menu","id":"log","label":"View Logs","order":1000,"parent":"admin"})
+p({"class":"foam.nanos.menu.Menu","id":"alterna-eft","label":"Alterna-EFT","handler":{"class":"foam.nanos.menu.ViewMenu", "view":{"class":"net.nanopay.cico.ui.AlternaEFTDownload"}},"order":1000,"parent":"admin"})
```
kgrgreer: Is AlternaEFTDownload checked in?

src/foam/comics/DAOControllerView.js
```diff
@@ -80,12 +83,14 @@ foam.CLASS({
...
               start('span').
                 style({background: 'rgba(0,0,0,0)'}).
-                show(this.mode$.map(function(m) { return m == foam.u2.DisplayMode.RW; })).
-                start().
-                  style({padding: '4px 4px 4px 1px'}).
-                  add(this.cls.getAxiomsByClass(foam.core.Action)).
-                end().
-              end().
+                show(self.mode$.map(function(m) { return m == foam.u2.DisplayMode.RW; })).
+                callIf(this.window.location.hash != '#allInvoices', function() {
```
kgrgreer: This is not a sustainable way to customize. Doing this will eventually litter this otherwise generic code with many application specific exceptions. Better to make a 'showActions' property and then just configure the allInvoices one differently, or else just add a class to this section and then use CSS to hide it.

src/menus
kirkeaton: This references a class that's not in foam2

src/foam/comics/DAOControllerView.js
kirkeaton: This references work done in our private repo so I don't think it should be added here.

kirkeaton (changes requested):
References work & functionality in nanopay so it shouldn't be merged into foam2

src/menus
blakegreen32: yea it works fine

blakegreen32: it was a request from laurence and varun, I could make the file be in foam2 but i didnt think it was necessary

PR #972, opened 2018-01-18
added xml escape on DIG

minsunny:
added xml escape which converts some special characters when select the format xml


src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -220,8 +221,7 @@ public void put(FObject o, Detachable d) {
 
             output(x, xmlData);
           } else {
-            response.setContentType("application/xml");
```
kgrgreer, line 223: Why did you remove setting the content type?

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -29,6 +29,7 @@
 import javax.servlet.ServletException;
 import javax.xml.stream.XMLInputFactory;
 import javax.xml.stream.XMLStreamReader;
+import org.apache.commons.text.StringEscapeUtils;
```
kgrgreer: Not used.

src/foam/nanos/fs/File.js
```diff
@@ -12,8 +12,9 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'Long',
-      name: 'id'
+      class: 'String',
```
kgrgreer, line 15: This file shouldn't be part of this PR.

src/services
```diff
@@ -40,4 +40,4 @@ p({"class":"foam.nanos.boot.NSpec", "name":"blobService",                 "lazy"
 p({"class":"foam.nanos.boot.NSpec", "name":"resetPasswordToken",          "lazy":true,  "serve":true,  "serviceClass":"foam.nanos.auth.resetPassword.ResetPasswordTokenService","boxClass":"foam.nanos.auth.token.TokenServiceSkeleton","client":"{\"class\":\"foam.nanos.auth.token.ClientTokenService\", \"delegate\": { \"class\":\"foam.box.HTTPBox\", \"url\":\"service/resetPasswordToken\" } }"})
```
kgrgreer, line 40: This file shouldn't be part of this PR.

PR #979, opened 2018-01-22
Test & fix: Broken MDAO behaviour on query: AND(*, CONTAINS_IC(ID, *))

mdittmer:
Currently, this query is not serviced properly. @jacksonic says he understands the issue; something to do with `ValueIndex` not handling predicates.


mdittmer:
This tests #980

mdittmer (commented):
Re:fix. This LGTM. @adamvy or @kgrgreer want to chime in before this lands?

PR #984, opened 2018-01-25
Made modifications to SMTPEmailService

kirkeaton:
1. Unmodelled SMTPEmailService
2. Added builder class to SMTPEmailService
3. Added modifications with how the Session is set up


adamvy:
Why?

A modeled class generates a builder for you now, so it looks like it would have been easier to just keep it modeled?

kirkeaton:
Oh I wasn't aware that builder classes were being generated. Whoops

adamvy:
Its a relatively new feature, it _would_ have saved you a lot of typing haha.

PR #985, opened 2018-01-25
Improvements to SMTPEmailService

kirkeaton:
1. Added more configurable properties like starttls and authenticate
2. Replaced service class with service script.
3. Added extra properties to service class for emails service


src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -200,12 +234,15 @@ sendEmail(emailMessage);`
...
-props.put("mail.smtp.starttls.enable", "true");
-props.put("mail.smtp.host", getHost());
-props.put("mail.smtp.port", getPort());
-
-setSession(Session.getInstance(props, null));`
+props.setProperty("mail.smtp.auth", authenticate_ ? "true" : "false");
+props.setProperty("mail.smtp.starttls.enable", starttls_ ? "true" : "false");
+props.setProperty("mail.smtp.host", host_);
+props.setProperty("mail.smtp.port", port_);
+if ( authenticate_ ) {
```
adamvy: Where is authenticate_ coming from?

You could also have left session as a property and defined a javaFactory for it which does this work, and it will be run the first get getSession() is accessed.  That would be the more FOAMy way of doing it, but this works too.

adamvy: Oh nvm, I see it now as a property.

Should probably call getAuthenticate() instead.

kirkeaton: Yeah. I've replaced all the direct property access calls with their getter methods now.

adamvy (approved):

PR #986, opened 2018-01-25
SMTP modifications

kirkeaton:
1. Added from, replyTo, and displayName properties to SMTPEmailService
2. Removed from, replyTo from EmailMessage
3. Added displayName property

### DIFF
--- a/src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js
+++ b/src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js
@@ -79,8 +79,6 @@ token = (Token) tokenDAO.put(token);
 
 EmailService email = (EmailService) getEmail();
 EmailMessage message = new EmailMessage();
-message.setFrom("info@nanopay.net");
-message.setReplyTo("noreply@nanopay.net");
 message.setTo(new String[] { user.getEmail() });
 
 HashMap<String, Object> args = new HashMap<>();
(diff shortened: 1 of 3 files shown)


mcarcaso:
Sorry for the late drive-by here but this PR/design doesn't feel right and I'd like to know a little more about it. Why are 'from' and 'replyTo' not properties on EmailMessage? From a modelling standpoint, I don't see why they shouldn't be.

I understand that the SMTPEmailService might not have permissions to send from any email address but I don't think the SMTPEmailService needs to care about that. Telling it to send an email from an email address it can't send from is more an issue with the caller than the SMTPEmailService.

There's a couple of issues with the sendEmailFromTemplate API method. According to the API, first argument is the user sending the email but the SMTPEmailService isn't respecting that because it fills in its own from address. If EmailMessage had a 'from' address then this first argument would be unnecessary and could be removed to simplify the API.

This stuff is hiding a bug in the ResetPasswordTokenService:
https://github.com/eatonk/foam2/blob/cee3f16077a625b684c675234817f72e85745630/src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js#L88
The first argument of the call to sendEmailFromTemplate is the user that initiated the password reset which is not who we'd want to be sending the email from.

Am I missing something here or should we file a bunch of cleanup bugs to address the issue? The bugs I'd suggest are:
1. Add from/replyTo to EmailMessage and remove them from SMTPEmailService
2. Remove the user argument from sendEmailFromTemplate
3. Fix up any calls to sendEmailFromTemplate that are relying on the incorrect behaviour.

PR #992, opened 2018-01-29
Added FileService

kirkeaton:
1. Added FileService class to service Blobs from the fileDAO and provide authentication
2. Replaced HttpBlobService with FileService in services file
3. Modified upload and download functions of HttpBlobService to take in the context
4. Added ownerId to File model


src/foam/nanos/blob/FileService.java
```diff
@@ -0,0 +1,80 @@
...
+
+      // check to see if current user has access to file owner
+      if ( userDAO_.find_(x, file.getOwnerId()) == null ) {
+        resp.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+        return;
+      }
+
+      // get blob and blob size
+      String blobId = ((IdentifiedBlob) file.getData()).getId();
+      Blob blob = getDelegate().find_(x, blobId);
```
adamvy: I believe you can just call file.getData().read().

IdentifiedBlob has is a ProxyBlob with a factory on `delegate` that does the find internally.

You may have to call setX() on file.getData() because it might not be coming out of the DAO in the right context.  Although that's something we should consider fixing as well if its true.

src/foam/nanos/blob/FileService.java
```diff
@@ -0,0 +1,80 @@
...
+    HttpServletRequest  req  = x.get(HttpServletRequest.class);
+    HttpServletResponse resp = x.get(HttpServletResponse.class);
+
+    try {
+      String path = req.getRequestURI();
+      String id = path.replaceFirst("/service/" + nspec_.getName() + "/", "");
+
+      // find file from file dao
+      File file = (File) fileDAO_.find_(x, id);
+      if ( file == null || !(file.getData() instanceof IdentifiedBlob) ) {
```
adamvy: Let's log a different error if file.getData() is not a blob.

That sounds more like an internal service error.

src/services
```diff
@@ -35,7 +35,7 @@ p({"class":"foam.nanos.boot.NSpec", "name":"email",                       "lazy"
 p({"class":"foam.nanos.boot.NSpec", "name":"emailTemplateDAO",            "lazy":true,  "serve":true,  "serviceScript":"return new foam.dao.PMDAO(new foam.dao.SequenceNumberDAO(new foam.dao.JDAO(x, foam.nanos.notification.email.EmailTemplate.getOwnClassInfo(), \"emailTemplates\")));","client":"{\"of\":\"foam.nanos.notification.email.EmailTemplate\"}" })
 p({"class":"foam.nanos.boot.NSpec", "name":"authy",                       "lazy":true,  "serve":false, "serviceClass":"foam.nanos.auth.twofactor.authy.AuthyService","service":{"class":"foam.nanos.auth.twofactor.authy.AuthyService","debug":true,"apiHost":"https://api.authy.com","apiKey":""}})
 p({"class":"foam.nanos.boot.NSpec", "name":"blobStore",                   "lazy":true,  "serve":false, "serviceClass":"foam.blob.BlobStore"})
-p({"class":"foam.nanos.boot.NSpec", "name":"httpBlobService",             "lazy":true,  "serve":false, "authenticate": false, "serviceScript":"return new foam.nanos.blob.HttpBlobService(x, x.get(\"blobStore\"));"})
```
adamvy: I think fileService should be its own service rather than repurposing httpBlobService.

src/foam/nanos/blob/FileService.java
eatonk: You're right. The Blob doesn't have it's context set

src/foam/nanos/blob/FileService.java
kirkeaton: I'm modifying this if statement to just check if file.getData == null. In this case the 404 status still makes sense.

src/foam/nanos/blob/FileService.java
kirkeaton: The delegate factory doesn't seem to be working. I'm getting a null pointer exception when trying to get the size of the blob. And it's because the delegate is not set (I think)

src/services
kirkeaton: Okay sure

src/foam/nanos/blob/FileService.java
adamvy: Interesting.  Leave a TODO here and I'll check it out later.

src/services
kirkeaton: Done.

PR #995, opened 2018-01-31
Added signing support to FObjects

kirkeaton:
1. Added Signer interface
2. Added Signable interface
3. Made FObject and PropertyInfo implement Signable and Signer respectively
4. Added Signing Benchmark test


src/foam/core/AbstractFObject.java
```diff
@@ -144,4 +144,29 @@ public boolean hasDefaultValue(String prop) {
...
     }
   }
+
+  public byte[] sign(PrivateKey key) {
+    return this.sign("SHA256withRSA", key);
+  }
+
+  public byte[] sign(String algorithm, PrivateKey key) {
+    try {
+      Signature signature = Signature.getInstance(algorithm);
```
kgrgreer, line 154: How expensive are these two lines? Can we reuse these by just passing in the Signature?

kirkeaton: It didn't make an effect in terms of performance.

src/foam/nanos/bench/SigningBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.bench;
+
+import foam.core.X;
+import foam.dao.DAO;
+import foam.dao.ListSink;
+import foam.dao.Sink;
+import net.nanopay.tx.model.Transaction;
```
adamvy, line 13: This adds nanopay dependencies to foam2.  Can we add this class to the NANOPAY repo instead?  Standalone nanos is broken as a result.

kirkeaton: Yeah, my mistake.

PR #996, opened 2018-01-31
Time View

blakegreen32:
Made a TimeView class that is used for setting Time properties.


src/foam/u2/TimeView.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc. All Rights Reserved.
```
kirkeaton: Should use this header instead:
```
/**
 * @license
 */
```

src/foam/core/types.js
```diff
@@ -115,6 +115,15 @@ foam.CLASS({
   label: 'Date and time'
 });
 
+foam.CLASS({
+  package: 'foam.core',
+  name: 'Time',
+  extends: 'String',
```
kirkeaton, line 121: Why does it extend String instead of extending Date?

kgrgreer: A time is not a date. A date is a moment in time but a time is a moment in a day. There are an infinite number of dates with the same time.

PR #997, opened 2018-02-01
Added Verify method to Signable interface

kirkeaton:
1. Added verify function to Signable interface
2. Renamed Hasher and Signer method from hash and sign to update to reflect what it's actually doing

### DIFF
--- a/src/foam/crypto/hash/Hasher.js
+++ b/src/foam/crypto/hash/Hasher.js
@@ -12,7 +12,7 @@ foam.INTERFACE({
 
   methods: [
     {
-      name: 'hash',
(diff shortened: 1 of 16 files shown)


src/foam/crypto/hash/Hasher.js
```diff
@@ -12,7 +12,7 @@ foam.INTERFACE({
 
   methods: [
     {
-      name: 'hash',
+      name: 'update',
```
kgrgreer: Is this too generic a name? Maybe updateDigest or just digest?

kirkeaton:
I’ll go with updateDigest

PR #999, opened 2018-02-01
Added SigningDAO

kirkeaton:
Added SigningDAO which signs incoming FObjects and stores them in the signatureDAO


src/foam/crypto/sign/SigningDAO.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class SigningDAO
+    extends ProxyDAO
+{
+  protected PrivateKey key_;
+  protected String algorithm_;
+  protected DAO signatureDAO_;
+
+  // TODO: come up with a better way to set the algorithm and key than this
+  public SigningDAO(X x, DAO delegate, String algorithm, PrivateKey key) {
```
kgrgreer: Can we get a default constructor which a reasonable default algorithm already set?

kirkeaton: Yeah, just added one

PR #1001, opened 2018-02-01
JSON Import tool

kirkeaton:
Rough draft of JSON import tool. No support for handling nested Objects


src/foam/tools/jsonimport/JSONImport.js
```diff
@@ -0,0 +1,110 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2018

kirkeaton: Done

PR #1002, opened 2018-02-01
Added rolling hashes to JSON Outputter

kirkeaton:
1. Added Rolling hashes to JSON Outputter
2. Renamed JDAO into AbstractJDAO
3. Created HashingJDAO
4. Created RollingHashingJDAO

### DIFF
--- a/src/foam/crypto/hash/HashingJDAO.java
+++ b/src/foam/crypto/hash/HashingJDAO.java
@@ -1,0 +1,45 @@
+/**
+ * @license
+ * Copyright 2017,2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.crypto.hash;
+
+import foam.core.ClassInfo;
+import foam.core.X;
+import foam.dao.AbstractJDAO;
+import foam.dao.DAO;
+import foam.dao.MapDAO;
+import foam.lib.json.Outputter;
+import foam.lib.json.OutputterMode;
+
+public class HashingJDAO
+    extends AbstractJDAO
+{
+  protected String algorithm_ = null;
+  protected Outputter outputter_ = null;
+
+  public HashingJDAO(X x, ClassInfo classInfo, String filename) {
+    this(x, classInfo, filename, "SHA-256");
+  }
+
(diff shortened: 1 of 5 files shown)


src/foam/crypto/hash/HashingJDAO.java
```diff
@@ -0,0 +1,45 @@
+/**
+ * @license
+ * Copyright 2017,2018 The FOAM Authors. All Rights Reserved.
```
kgrgreer: Copyright should just be the first year of creation (according to Google lawyers) and doesn't need to ever be updated after that.

src/foam/crypto/hash/HashingJDAO.java
```diff
@@ -0,0 +1,45 @@
...
+
+import foam.core.ClassInfo;
+import foam.core.X;
+import foam.dao.AbstractJDAO;
+import foam.dao.DAO;
+import foam.dao.MapDAO;
+import foam.lib.json.Outputter;
+import foam.lib.json.OutputterMode;
+
+public class HashingJDAO
```
kgrgreer: Is this cascading/chained?

src/foam/crypto/hash/HashingJDAO.java
kirkeaton: Fixed

src/foam/dao/AbstractJDAO.java
```diff
@@ -0,0 +1,204 @@
...
+import java.io.*;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.regex.Pattern;
+import java.util.TimeZone;
+
+public abstract class AbstractJDAO
+    extends ProxyDAO
+{
+  protected Pattern COMMENT = Pattern.compile("(/\\*([^*]|[\\r\\n]|(\\*+([^*/]|[\\r\\n])))*\\*+/)|(//.*)");
```
kgrgreer, line 28: This seems like overkill.  Isn't it enough to just check if the first two characters are //?
Not really related to this PR though.

kirkeaton: Kind of is yeah but I wanted it to be exhaustive

PR #1010, opened 2018-02-07
MenuBar settings selection fix

blakegreen32:
Fixed MenuBar.js isSelected function to take into account settings menu selections

### DIFF
--- a/src/foam/nanos/menu/MenuBar.js
+++ b/src/foam/nanos/menu/MenuBar.js
@@ -59,7 +59,7 @@ foam.CLASS({
                   if ( ! self.selected ) self.selected = menu;
                   this.start()
                     .addClass('menuItem')
-                    .enableClass('selected', self.currentMenu$.map(function (value) { return self.isSelected(value, menu) || self.selected === menu.id; }))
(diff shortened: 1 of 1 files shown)


src/foam/nanos/menu/MenuBar.js
```diff
@@ -84,6 +84,10 @@ foam.CLASS({
         return false;
       }
 
+      if ( this.selected === menu.id && this.selected.parent != 'settings' ) {
```
kgrgreer: Why is this required? Add comment to explain.

blakegreen32: I have actually changed it because this wasnt persisting in every case, I used window.location.hash.includes('#set') so that if the user navigates to any settings pages the previous menu items wont stay underlined. With the latest commit the menu items work as they should and only show as selected if the user is on that page and has not navigated to a settings page.

blakegreen32: I think this makes sense as if any new menu items are added to the settings menu they will include the hash of "#set" in them

kirkeaton (approved):

PR #1012, opened 2018-02-08
Added Data Update Gateway (DUG)

kirkeaton:
1. Added DUG class with implementation
2. Added HTTPSink which PUTS incoming data to an external URL
3. Created Outputter interface and made JSON, CSV, and HTML outfitters implement it
4. Added DUGDAO which executes a DUG on put
5. Added DUGDAO to services file


src/foam/lib/Outputter.java
```diff
@@ -0,0 +1,15 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2018

src/foam/lib/Outputter.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+package foam.lib;
+
+import foam.core.FObject;
+
+// not modelled because method with
+// name stringify is not generated
+public interface Outputter {
+  String stringify(FObject obj);
```
kgrgreer, line 14: Couldn't we add the output(Object) method here as well?

src/foam/nanos/dig/DUG.js
```diff
@@ -4,41 +4,57 @@
...
+foam.ENUM({
+  package: 'foam.nanos.dig',
+  name: 'Format',
 
-   values: [
-     { name: 'JSON', label: 'JSON' },
-     { name: 'XML',  label: 'XML'  }
-   ]
- });
+  documentation: 'CRUD controller modes: CREATE/VIEW/EDIT.',
```
kgrgreer: Cut & Paste documentation error. My bad.

src/foam/nanos/dig/DUGDAO.java
```diff
@@ -0,0 +1,34 @@
...
+import foam.dao.JDAO;
+import foam.dao.ProxyDAO;
+import foam.nanos.auth.User;
+
+public class DUGDAO
+    extends ProxyDAO
+{
+  public DUGDAO(X x) {
+    setX(x);
+    setDelegate(new JDAO(x, DUG.getOwnClassInfo(), "dugs"));
```
kgrgreer: This design doesn't re-execute() DUG's on startup. It would have to do a select() in the constructor to find all existing DUG's to start or don't setDelegate() and insert between the JDAO and the MapDAO so that all values automatically get re-put_() on startup. Second approach isn't as general though, because it wouldn't work with a Postgres DAO for example.

src/foam/lib/Outputter.java
kirkeaton: Yes done.

src/foam/lib/Outputter.java
kirkeaton: Done

src/foam/nanos/dig/DUG.js
kirkeaton: Fixed

src/foam/nanos/dig/DUGDAO.java
kirkeaton: Modified to select all DUGS on start

src/foam/nanos/dig/DUGDAO.java
```diff
@@ -0,0 +1,44 @@
...
+    this(x, new JDAO(x, DUG.getOwnClassInfo(), "dugs"));
+  }
+
+  public DUGDAO(X x, DAO delegate) {
+    super(x, delegate);
+    delegate.select(new AbstractSink() {
+      @Override
+      public void put(FObject obj, Detachable sub) {
+        User user = (User) x.get("user");
+        ((DUG) obj).setOwner(user.getId());
```
kgrgreer: I think this line is a mistake. The user will be System at this point and we don't want to change the user to that. The user was already set in put_() when the object was originally created.

kgrgreer: The script needs to recreate the owner's context so that it doesn't run the DUG with extra permissions. To do this, see how the sudo() method is written in Script.js. Maybe even move that code somewhere reusable under auth.

src/foam/dao/HTTPSink.java
```diff
@@ -0,0 +1,70 @@
...
+      }
+
+      // create new writer if null
+      if ( writer_ == null ) {
+        writer_ = new BufferedWriter(
+            new OutputStreamWriter(conn_.getOutputStream()));
+      }
+
+      writer_.write(outputter_.stringify(obj));
+      super.put(obj, sub);
```
kgrgreer: This line isn't needed since AbstractSink doesn't do anything in put().

kirkeaton: Removed

src/foam/dao/HTTPSink.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.*;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+
+public class HTTPSink
+    extends AbstractSink
+{
+  protected String url_;
+  protected Outputter outputter_;
```
kgrgreer: If two threads hit the same DUG at the same point their output will be interlaced. Just create a new Outputter for each put().

src/foam/nanos/script/Script.js
```diff
@@ -127,7 +106,7 @@ foam.CLASS({
           shell.set("currentScript", this);
           shell.set("x", x);
           shell.eval("runScript(String name) { script = x.get(\\"scriptDAO\\").find(name); if ( script != null ) eval(script.code); }");
-          shell.eval("sudo(String user) { currentScript.sudo(user); }");
+          shell.eval("sudo(String user) { foam.util.Auth.sudo(x, user); }");
```
kgrgreer, line 109: Nice!

PR #1015, opened 2018-02-09
JS: 'In' mLang: Adapt arg2 for IN(enum, [enumValue1, enumValue1]) case

mdittmer:


src/foam/mlang/mlang.js
```diff
@@ -866,6 +866,24 @@ foam.CLASS({
...
+      preSet: function(old, nu) {
+        if ( ! ( this.Constant.isInstance(nu) &&
+                 Array.isArray(nu.value) ) ) {
+          return nu;
+        }
+
+        // Convert enum value arg2 values to strings for uppercase comparison.
+        var array = nu.value;
+        for ( var i = 0; i < array.length; i++ ) {
+          if ( foam.core.AbstractEnum.isInstance(array[i]) )
```
kgrgreer, line 880: Why do we do this?

mdittmer: Lines 899-907: Constant value on right will be treated as array of strings and used as keys for a value set. This detects when an enum value, rather than its string name, is used in the array on the right.

We need actual strings for the toUpperCase() case. I guess the alternative would be to use toString().toUpperCase() below. 

@kgrgreer  WDTY?

mdittmer:
*ping* @kgrgreer

adamvy:
@mdittmer I re-wrote the IN expression a few days ago to use foam.util.compare(), does this fix your issue without this PR?

mdittmer:
@adamvy ack. I'll give it a try.

mdittmer:
https://github.com/foam-framework/foam2/commit/e5c20561ef92b58b9d52ec8c993800 does indeed fix this. Thanks @adamvy !

PR #1017, opened 2018-02-09
Make U2/CSS axiom management more resilient in models that are loaded…

mdittmer:
… in both Node JS and the browser

### DIFF
--- a/src/foam/core/Window.js
+++ b/src/foam/core/Window.js
@@ -200,7 +200,7 @@ foam.CLASS({
     },
     function installCSS(text, id) {
       /* Create a new <style> tag containing the given CSS code. */
-      this.document.head.insertAdjacentHTML('beforeend',
+      this.document && this.document.head.insertAdjacentHTML('beforeend',
           '<style owner="' + id + '">' + text + '</style>');
     }
   ]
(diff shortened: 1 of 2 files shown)


src/foam/core/Window.js
```diff
@@ -200,7 +200,7 @@ foam.CLASS({
     },
     function installCSS(text, id) {
       /* Create a new <style> tag containing the given CSS code. */
-      this.document.head.insertAdjacentHTML('beforeend',
+      this.document && this.document.head.insertAdjacentHTML('beforeend',
```
kgrgreer, line 203: When does this happen? If we don't have a document/window we shouldn't be installing a Window in the context, correct?

mdittmer: Here is an example:

https://github.com/mdittmer/mdn-confluence/blob/7e0e97cd10c330c57d5cb138a25a8f8ff872c0be/lib/org/mozilla/mdn/GridProperty.es6.js#L16

This property should be usable in NodeJS, but the rawRowFormatter and the CSS should belong to the property too.

mdittmer:
*ping* @kgrgreer

jacksonic (approved):

PR #1028, opened 2018-02-13
AuthenticatedUserDAO

kirkeaton:
1. Added ServiceProvider class and relationship to User class
2. Implemented AuthenticatedUserDAO


jlhughes (approved):

src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -10,7 +16,12 @@
 import foam.nanos.auth.AuthService;
 import foam.nanos.auth.User;
 import java.security.AccessControlException;
+
+import foam.util.SafetyUtil;
 import net.nanopay.model.Account;
+import org.apache.http.util.TextUtils;
```
adamvy: Are you actually using this import anywhere?

src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -10,7 +16,12 @@
 import foam.nanos.auth.AuthService;
 import foam.nanos.auth.User;
 import java.security.AccessControlException;
+
+import foam.util.SafetyUtil;
 import net.nanopay.model.Account;
```
adamvy: We need to something about the fact that this class in FOAM2 is referencing a class in nanopay.  Doesn't have to be done in the PR but we need a plan for it soon, right now building nanos outside of nanopay is broken.

kirkeaton: It's not actually being used in this class so I can just remove it.

adamvy: Did you reply to the right comment?  It looks like Account is used in many places in this class.

kirkeaton: I did reply to the right comment. It's not being used anywhere here.

src/foam/nanos/auth/AuthenticatedUserDAO.java
kirkeaton: No I just cleaned up the imports.

src/foam/nanos/auth/AuthenticatedUserDAO.java
adamvy: Oh I see, didn't read your changes closely enough, you removed all the references to Account.

adamvy (approved):

PR #1030, opened 2018-02-14
Modified JDAO to continue parsing lines through parse errors

kirkeaton:
1. Modified JDAO to not stop on parse failure. 
2. Added different parsing of lines depending on the operation

### DIFF
--- a/src/foam/dao/AbstractJDAO.java
+++ b/src/foam/dao/AbstractJDAO.java
@@ -94,6 +94,10 @@ public abstract class AbstractJDAO
             }
             break;
         }
+
+        if ( object == null ) {
+          System.err.println(getParsingErrorMessage(line) + ", source: " + line);
+        }
       } catch (Throwable t) {
         System.err.println("Error reading journal line: " + line);
         t.printStackTrace();
--- a/src/foam/dao/AbstractJDAO.java
+++ b/src/foam/dao/AbstractJDAO.java
@@ -79,25 +79,19 @@ public abstract class AbstractJDAO
         // remove first two characters and last character
         line = line.trim().substring(2, line.trim().length() - 1);
 
-        FObject object;
-        switch ( operation ) {
-          case 'p':
-            object = parser.parseString(line);
-            if ( object != null ) {
-              getDelegate().put(object);
-            }
-            break;
-          case 'r':
-            Object id = parser.parseStringForId(line);
-            if ( id != null && (object = getDelegate().find(id)) != null ) {
-              getDelegate().remove(object);
-            }
-            break;
-        }
-
-        // log when object is not parsed properly
+        FObject object = parser.parseString(line);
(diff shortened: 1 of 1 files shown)


src/foam/dao/AbstractJDAO.java
```diff
@@ -79,18 +79,19 @@ protected void loadJournal()
         // remove first two characters and last character
         line = line.trim().substring(2, line.trim().length() - 1);
 
-        FObject object = parser.parseString(line);
-        if ( object == null ) {
-          System.err.println(getParsingErrorMessage(line) + ", source: " + line);
-          break;
```
kirkeaton: @kgrgreer @jlhughes This break statement was causing a journal file to stop being parsed

jlhughes (commented):
Are parsing failures caught in the 'catch'? Otherwise can you log them when object is nil?

kirkeaton:
@jlhughes I don't think they will be caught in the catch but I will log them.

jlhughes (approved):

src/foam/dao/AbstractJDAO.java
```diff
@@ -79,20 +79,26 @@ protected void loadJournal()
...
           case 'p':
-            getDelegate().put(object);
+            object = parser.parseString(line);
+            if ( object != null ) {
+              getDelegate().put(object);
+            }
             break;
           case 'r':
-            getDelegate().remove(object);
+            Object id = parser.parseStringForId(line);
```
adamvy: There's a reason I removed the parseStringForId line and used parseObject in both cases.

parseStringForId fails for multipart ID objects.

kirkeaton: Okay then how do you propose we solve this. The current remove lines that are printed out look like this: ```r({"id":1})``` and parseString does not parse this either

adamvy: I changed the remove lines to output the full object as well when I made this change.  I propose leaving it as is.

adamvy: https://github.com/foam-framework/foam2/commit/a501a9c7b6544154aa9aadae7aa93578c43a495a

kirkeaton: Actually I see that you modified the remove_ function to print out the whole object just like the put_ method. But as it stands the old remove lines are broken.

adamvy: Do we have existing data in the old format we need to support?

I guess we could try to parse it both ways.  If the current method fails to parse, then try to parse it as just an id.

kirkeaton: @adamvy Okay leaving it as is but instead of stopping on parse failure Ill change it to just continue.

adamvy (approved):

PR #1031, opened 2018-02-14
Fix cron scheduler service

diaswrd:
Two minor things required for cronjobs to run successfully:

1. CronScheduler service wasn't defined/initialized in the services file.

2. The CronScheduler loop to run new tasks was set to sleep until the next scheduled task was due, which could be in one minute or in any other given time in the future. 
Meaning that any new task registered after the service start would be executed only at the next thread loop, which could be in minutes, hours or days. 
Updating the watcher to run every minute enables it to pickup any new registered jobs and run them as soon as they need to run, independent of other jobs' scheduled date.


diaswrd:
My assumption to start the cronScheduler here is that it needs the cronDAO in order to get the list of tasks that should be scheduled.

PR #1032, opened 2018-02-15
Resend verification email

blakegreen32:
Added check to ApplicationController to push ResendVerificationEmail view to the stack if the users email is not verified.


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -168,6 +168,10 @@ foam.CLASS({
         self.loginSuccess = !! result;
         if ( result ) {
           self.user.copyFrom(result);
+          if ( ! self.user.emailVerified ) {
+            self.stack.push({ class: 'net.nanopay.auth.email.ResendVerificationEmail' });
```
kirkeaton: Hmm so you're referencing a class in our repo. So this view may actually need to be moved here. If you move the view. I can make the modifications to move the email verification service to foam2.

PR #1037, opened 2018-02-22
Build tool

mcarcaso:
I've been favouring this branch over master so I think it's good enough to merge now before it grows more and I'll fix anything that breaks as a result.

There's still some work to be done so the build tool isn't ready for prime time yet but it's getting there.


kgrgreer:
FYI: @jlhughes

mcarcaso:
foam2's make and nanopay's run-tomcat.sh seem to be happy with these changes so I'm going to merge them today unless anybody objects.

PR #1040, opened 2018-02-23
Added sessionId to FileService

kirkeaton:


src/foam/nanos/blob/FileService.java
```diff
@@ -49,8 +56,33 @@ protected void download(X x) {
         return;
       }
 
+      // parse query parameters
```
kgrgreer: I think it would be better to put session support in the AuthenticatedWebAgent, that way it would be reusable.

src/foam/nanos/auth/ProfilePictureView.js
```diff
@@ -92,11 +92,16 @@ foam.CLASS({
           return this.E('img').addClass('shopperImage')
             .attrs({
               src: this.data$.map(function (data) {
+                console.log(data);
                 if ( data && data.data ) {
                   var blob = data.data;
-                  return self.BlobBlob.isInstance(blob) ?
-                    URL.createObjectURL(blob.blob) :
-                    ( "/service/httpFileService/" + data.id );
+                  var sessionId = localStorage['defaultSession'];
```
kgrgreer, line 97: This looks good.

src/foam/nanos/blob/FileService.java
kirkeaton: Yeah you're right. I've moved it to AuthWebAgent

src/foam/nanos/auth/ProfilePictureView.js
kirkeaton: Thanks

PR #1044, opened 2018-02-26
Implement FirestoreDAO and an authentication decorator

mdittmer:
The DAO works with both NodeJS `@google-cloud/firestore` module and the Firestore web API (v4.9.1). `AuthenticatedDAO` gates API calls to its delegate on an authentication ID. This does not enforce access, of course, but does allow DAO operations to queue until authentication is complete.

The `Default*` classes are *not* inner classes so that they are serializable.


src/com/google/firebase/AuthenticatedDAO.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'com.google.firebase',
+  name: 'AuthenticatedDAO',
```
bshepherdson: This should probably be named something like `AwaitAuthenticationDAO`; it's not an `AuthenticatedDAO` in the server sense, which is misleading.

src/com/google/firebase/FirestoreDAO.js
```diff
@@ -0,0 +1,221 @@
...
+      this.putBacklog_ = [];
+    },
+    function remove_(x, obj) {
+      return this.getDoc_(obj).delete();
+    },
+    function find_(x, idOrObj) {
+      return (foam.core.FObject.isInstance(idOrObj) ?
+              this.getDoc_(idOrObj) : this.collection.doc(idOrObj.toString()))
+          .get()
+          .then(function(docSnapshot) { return docSnapshot.data() || null; });
```
bshepherdson: Not running this through `this.fobject.f`?

bshepherdson (commented):
Overall looks good. Firestore is a better fit for DAOs in general than the old Realtime Database, and faster too.

Subcollections are a good fit for relationships, in particular.

src/com/google/firebase/AuthenticatedDAO.js
mdittmer: Done.

src/com/google/firebase/FirestoreDAO.js
mdittmer: That's a bug. Good catch. Fix pushed.

mdittmer:
Got offline LGTM from @shepheb

PR #1045, opened 2018-02-26
Parser/outputter up to Context; *Sockets deal in Strings

mdittmer:
This change facilitates applications that wish to configure different (de)serialization schemes at a high level without having other schemes introduced by lower-level components.

As much as possible parsers and outputters are provided by the enclosing `foam.box.Context`. Whenever possible, lower level `*Socket` components deal with strings, rather than yet-to-be-serialized objects. There are some exceptions required by, for example, the asymmetry of how WebSockets are managed on client and server.

### DIFF
--- a/src/foam/box/BoxJsonOutputter.js
+++ b/src/foam/box/BoxJsonOutputter.js
@@ -1,0 +1,23 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.box',
+  name: 'BoxJsonOutputter',
+  extends: 'foam.json.Outputter',
+
+  requires: [ 'foam.box.ReturnBox' ],
+  imports: [ 'me' ],
+
+  methods: [
+    function output(o) {
+      if ( o === this.me ) {
+        return this.SUPER(this.ReturnBox.create());
+      }
+      return this.SUPER(o);
+    }
+  ]
+});
--- a/src/files.js
+++ b/src/files.js
@@ -309,6 +309,7 @@ FOAM_FILES([
   { name: "foam/box/ReturnBox" },
   { name: "foam/box/RawMessagePortBox" },
   { name: "foam/box/WebSocketBox" },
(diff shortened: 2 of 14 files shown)


adamvy:
Can you outline what you need this support for?

The future plan for boxes is to make serialization just one the boxes in the chain of boxes you might receive when you look up a particular service.

Globally configuring serialization isn't appropriate because talking to different services and servers may require different serialization strategies, configuring this at a global level even in a single application is not sufficient.

mdittmer:
I need this to support the setup at https://github.com/mdittmer/mdn-confluence where I have:

    ModelStorage <-- FON --> Models
    DataStorage <-- JSON --> Data(of:oneOf(Models))

The previous setup had parsers and outputters auto-initialized at too many levels to sanely create this setup. My solution was to create two separate box contexts that declare their (de)serialization schemes at the top level, and have lower level components receive their parser/outputters from their context.

A `(De)SerializerBox` scheme would also work, but there are two potential issues.

1. Configuration/wiring is (still) a pain. The top-level context that just wants to, say "register a DAO that sends its data to a (Web)SocketBox" (or similar high-level description) needs to concern itself with every layer until it gets down to (de)serialization, and instantiate the right strategy for the system.
2. `(De)SerializerBox`es are kind of square-peg-round-hole. (De)serialization is `message/string <=> string/message`, and not `message <=> message`. Implementing the `Box` interface can be done (maybe make the serialization of the message to be sent over the wire into `msg.object`?), but that seems like a workaround to coerce the (de)serialization interface into the `Box` interface.

src/foam/box/BoxJsonOutputter.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
```
kgrgreer, line 7: Add docs to explain what this is for.

kgrgreer:
Couldn't you use Context.register(cls) to replace the regular serialization classes with substitutes?

kgrgreer:
Dead by agreement

PR #1051, opened 2018-02-26
Printable Model browser

zahnno:
PR includes:
- Model browser which displays UML diagram & properties of Models available through nSpecDAO.


src/foam/doc/DocBrowser.js
```diff
@@ -76,6 +76,20 @@ foam.CLASS({
       name: 'axiom',
       hidden: true
     },
+    {
+      name: 'documentation',
+      label: 'Documentation',
+      tableCellFormatter: function(value, obj, axiom) {
```
kgrgreer: How is this different than what happens by default?

src/foam/doc/ModelBrowser.js
```diff
@@ -0,0 +1,238 @@
+foam.CLASS({
+  package: 'foam.doc',
+  name: 'ModelBrowser',
+  extends: 'foam.u2.Element',
+  documentation: 'Show UML & properties for passed in models',
+
+  requires: [
+    'foam.doc.DocBorder',
```
kgrgreer: Sort lists.

src/foam/doc/ModelBrowser.js
```diff
@@ -0,0 +1,238 @@
...
+    }
+  `,
+
+  methods: [
+    function initE(){
+      this.SUPER();
+      var self = this;
+
+      this.nSpecDAO.select().then(function(d){
+        d.array.forEach(function(n){
```
kgrgreer: You could just pass this function to select() directly and skip the extra step/nesting.

src/foam/doc/ModelBrowser.js
```diff
@@ -0,0 +1,238 @@
...
+          if ( cls.class ) {
+            self.models.push(cls.class)
+            self.models = Array.from(self.models)
+          }
+        });
+      });
+      this.start().addClass(this.myClass())
+        .start('h2').add("Model Browser").end()
+        .start().add(this.PRINT_PAGE).end()
+        .add(this.slot(function(models) {
```
kgrgreer: Do you need to use slot()? Is models going to change?

zahnno: How can I grab my models before initE method?

PR #1053, opened 2018-02-27
Add bwd compat & toString() support for multi-part IDs

mdittmer:
Also add tests


mdittmer:
Amendment: Added backwards compatibility for existing data in old-style array representation.

PR #1058, opened 2018-02-28
Version JDAO

xuerongCode:
create VersionJDAO to get data from filename.0,filename.1,filename.2 .......
only write to filename.1


jlhughes (approved):
Ok, can you also update the find.sh script and include it with this pr.

src/foam/crypto/hash/RollingHashingJDAO.java
```diff
@@ -8,14 +8,14 @@
...
-import foam.dao.AbstractJDAO;
+import foam.dao.JDAO;
 import foam.dao.DAO;
 import foam.dao.MapDAO;
 import foam.lib.json.Outputter;
 import foam.lib.json.OutputterMode;
 
 public class RollingHashingJDAO
-    extends AbstractJDAO
+    extends JDAO
```
kirkeaton: Why the change here?

src/foam/crypto/sign/SigningJDAO.java
```diff
@@ -17,7 +17,7 @@
 import java.security.PrivateKey;
 
 public class SigningJDAO
-    extends AbstractJDAO
+    extends JDAO
```
kirkeaton: Why the change here?

src/foam/crypto/hash/RollingHashingJDAO.java
xuerongCode: AbstractJDAO can not load entries to MapDAO now. That is why I changed to extend JDAO

src/foam/crypto/sign/SigningJDAO.java
xuerongCode: Same as above

src/foam/dao/VersionJDAO.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package foam.dao;
+
+import foam.core.ClassInfo;
+import foam.lib.json.Outputter;
+import foam.lib.json.OutputterMode;
+import java.io.*;
+
+public class VersionJDAO
```
kgrgreer: I don't think we want to do this as a sub-class because if we then want a HashingJDAO or a SigningJDAO or whatever to have this feature, we'll then need a VersionHashingJDAO and a VersionSigningJDAO also. This will lead to an exponential explosion of JDAO sub-classes.

kgrgreer: I would rather build this into the base AbstractJDAO functionality so that all JDAO's can use this functionality.
Maybe, just have it look for filename.0 and load it if available, then use filename, just like normal. This way it works just like normal if the filename.0 or filename.base file doesn't exist.

src/foam/dao/AbstractJDAO.java
```diff
@@ -58,16 +58,24 @@ public AbstractJDAO(foam.core.X x, DAO delegate, String filename) {
...
-      file_ = getX().get(foam.nanos.fs.Storage.class).get(filename);
-
-      if ( ! file_.exists() ) file_.createNewFile();
-
-      setDelegate(delegate);
-      loadJournal();
-
-      out_ = new BufferedWriter(new FileWriter(file_, true));
+      //get repo entries in filename.0 journal first
+      File infile = getX().get(foam.nanos.fs.Storage.class).get(filename + ".0");
```
kgrgreer: Inconsistent naming. You have "infile" and "writeFile", which mix camel-case and non-camel-case with in/out vs read/write.

PR #1062, opened 2018-03-01
Change default SMTP server to mail.nanopay.net, port 25

jlhughes:


kirkeaton (changes requested):
You're referencing mail.nanopay.net but this is the foam2 repo. Shouldn't add that here.

jlhughes:
This should be in nanopay proper, not on the foam side as it's nanopay specific.

PR #1063, opened 2018-03-02
SessionWebAgent

kirkeaton:
1. Added SessionWebAgent class which checks if the sessionId query parameter is set in the URL
2. Removed sessionId support from AuthWebAgent
3. Updated httpFileService and httpBlobService to decorate with SessionWebAgent
4. Modified FileService and HttpBlobService to accept a name parameter since NSpec was only being decorated by parent class and not any of the delegates.


src/foam/nanos/http/SessionWebAgent.java
```diff
@@ -0,0 +1,79 @@
...
+      // check permissions
+      session.setContext(session.getContext().put("user", user));
+      if ( ! auth.check(session.getContext(), permission_) ) {
+        throw new AccessControlException("Access denied");
+      }
+
+      // execute delegate
+      getDelegate().execute(x.put(Session.class, session).put("user", user));
+    } catch ( Throwable t ) {
+      logger.error(t);
```
kgrgreer: Add a first argument to error() to explain the context of the error. Like error("Unexpected exception in SessionWebAgent", t);

kirkeaton: Thanks, done

PR #1064, opened 2018-03-02
OverlayDropdown CSS Fix

kennyqiyenkan:
Fixes the positioning of the dropdown overlay.


mdittmer:
Can I get a screenshot and/or explanation of how this is broken? I've recently had to tweak this CSS, and I think this undoes some of my work.

kennyqiyenkan:
Basically, the way the CSS worked previously, it would not position the dropdown according to it's container and place its position based on the window instead.

kgrgreer:
But this was working last week and the OverlayDropdown hasn't changed in a month, so why did it just stop breaking now?

kennyqiyenkan:
Closed until further update

mdittmer:
@kgrgreer @kennyqiyenkan the overlay dropdown sometimes needs to layout somewhere other than "top-right corner of the whole document". For that reason it uses CSS+JS to layout relative to the top-right corner of its parent (well, actually, the top-right corner of its closest ancestor with non-default `position`, but that's a CSS rant for another time).

I've attached screenshots of the layout + devtools on broken layout on `/com/google/foam/demos/u2/Table.html` demo, and a project that uses overlay dropdown somewhere other than the top-right corner of the whole document.

This doesn't explain the exact spacing of the broken demo, but I hope it articulates the properties/layout strategies that should continue to layout correctly if we change the CSS.

foam-demo_overlay_dropdown_layout

The above demo should be fixable by correctly styling a common parent to the "vertical . . ." button and the overlay dropdown itself.

mdn-confluence_overlay_dropdown_encapsulation

PR #1071, opened 2018-03-06
Disabled blob cloning for ProxyBlob classes

kirkeaton:
Added AbstractBlobPropertyInfo which prevents cloning of Blob properties. This aims to fix a bug that occurs when Blob files are read from a journal. When loading the blobs from the journal file the MapDAO would clone the IdentifiedBlob's delegate which resulted in unnecessary network calls.


adamvy:
I would suggest this instead https://github.com/foam-framework/foam2/pull/1072

This makes identified blobs clone property by not cloning their delegate.  This should fix your issue since we should only be storing identified blobs in the journal.

PR #1078, opened 2018-03-06
Added conversion of String arguments from ISO-8859-1 to UTF-8

kirkeaton:
Fixes issue where certain characters were being rendered incorrectly.

### DIFF
--- a/src/foam/nanos/notification/email/SMTPEmailService.js
+++ b/src/foam/nanos/notification/email/SMTPEmailService.js
@@ -34,6 +34,7 @@ foam.CLASS({
     'javax.mail.*',
     'javax.mail.internet.InternetAddress',
     'javax.mail.internet.MimeMessage',
+    'java.nio.charset.StandardCharsets',
     'java.util.Date',
     'java.util.Properties'
   ],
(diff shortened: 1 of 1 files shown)


src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -252,6 +254,14 @@ EmailTemplate emailTemplate = DAOResourceLoader.findTemplate(getX(), name, group
 if ( emailMessage == null )
   return;
 
+for ( String key : templateArgs.keySet() ) {
+  Object value = templateArgs.get(key);
+  if ( value instanceof String ) {
+    String s = (String) value;
+    templateArgs.put(key, new String(s.getBytes(StandardCharsets.ISO_8859_1), StandardCharsets.UTF_8));
```
adamvy, line 261: I don't think this is the right way to convert from ISO_8859_1 to UTF_8.

This just reinterprets the ISO-8859-1 bytes as UTF-8 bytes.  I guess most (maybe all?) character sequences in iso-8859-1 will be the same in utf-8, but that's a conincidence IMO, not something we should rely on.

What bug is this fixing?  My understanding of Java is that you shouldn't need to know or care what character encoding a String is using internally, and I believe this code here doesn't actually change what the string is encoded as internally.  It looks like this would be a no-op to me.

adamvy: Okay I see how this would work if a multi byte utf-8 character was saved as multiple iso-8859-1 characters and read back in as iso-8859-1 characters we could convert it back to utf-8.

But this feels like a deeper bug, is the Journal not serializing/deserializing in utf-8?

PR #1080, opened 2018-03-07
Validations

blakegreen32:
Validations added to User model, Address model and Phone model.


src/foam/nanos/auth/Address.js
```diff
@@ -45,29 +45,63 @@ foam.CLASS({
...
+        }
+      }
     },
     {
       class: 'String',
       name: 'postalCode',
-      required: true
+      required: true,
+      validateObj: function (postalCode) {
+        var postalCodeRegex = /^[A-Za-z]\d[A-Za-z][ -]?\d[A-Za-z]\d$/;
```
kirkeaton: This regex isn't correct for Canadian postal codes. From wikipedia:
>Postal codes do not include the letters D, F, I, O, Q or U, and the first position also does not make use of the letters W or Z

src/foam/nanos/auth/Phone.js
```diff
@@ -17,7 +17,14 @@ foam.CLASS({
     {
       class: 'PhoneNumber',
       name: 'number',
-      required: true
+      required: true,
+      validateObj: function (number) {
+        var numberRegex = /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/;
```
kirkeaton: This doesn't look like it would capture if someone were to add their country code to the beginning. i.e. +14169000000

src/foam/nanos/auth/User.js
```diff
@@ -44,7 +44,14 @@ foam.CLASS({
     {
       class: 'String',
       name: 'firstName',
-      tableWidth: 160
+      tableWidth: 160,
+      validateObj: function(firstName) {
+        var firstNameRegex = /^[a-zA-Z- ]{1,70}$/;
```
kirkeaton: Might be good to escape the hyphen character just in case.
```[a-zA-Z\- ]```

File: src/foam/nanos/auth/User.js [hunk 4]
```diff
@@ -53,14 +60,26 @@ foam.CLASS({
     {
       class: 'String',
       name: 'lastName',
-      tableWidth: 160
+      tableWidth: 160,
+      validateObj: function (lastName) {
+        var lastNameRegex = /^[a-zA-Z- ]{1,70}$/;
```
kirkeaton: Likewise here, escape the hyphen character just in case.

kirkeaton:
May not be a good idea to limit first and last names to only use characters a-zA-Z. If someone has a _é_ in their name then the regex would not allow that. If someone were to enter their name in another language's then that would fail as well.

PR #1081, opened 2018-03-07
Fix NavigationBar Hover

kennyqiyenkan:
- dismissing the dropdown removes the CSS
- Adds `lastSelectedMenu` to `ApplicationController`
- Adds the listeners: `lastMenuListener` & `removeLastMenuListener` to
`ApplicationController`
- used in `Menu` and `SubMenuView`


kgrgreer:
Looks good, but please add more docs.

PR #1089, opened 2018-03-09
Message model

vaishnavi7096:
1. Modified TicketMessage Model which necessary changes recommended.
2. Moved reference files from files.js to support.js
3. Created ticketMessages DAO which necessary changes in files such as classes.js, services, support.js
4. Also deleted unnecessary lines in users and cronjobs files.


tools/classes.js
```diff
@@ -208,7 +208,8 @@ var classes = [
 
   'foam.nanos.dig.Format',
   'foam.nanos.dig.DUG',
-  'foam.nanos.dig.DigPostParameters',
+  'foam.support.model.TicketMessage',
```
zahnno: please place under DigPostParameters like so 

> 'foam.nanos.dig.DigPostParameters',
> 
> //Support
> 'foam.support.model.TicketMessage'

PR #1092, opened 2018-03-10
Model what's given to GENMODEL

mcarcaso:
This is needed to make this all play nicely with the build tool.


mcarcaso:
Actually, hold off on this for now. I'm not crazy about the way it's turning out.

PR #1099, opened 2018-03-13
Password Expiry

kirkeaton:
1. Added passwordExpiry property to User model. 
2. Added PasswordExpiryAuthService decorator. 
3. Decorated auth service with PasswordExpiryAuthService


src/foam/nanos/auth/PasswordExpiryAuthService.js
```diff
@@ -0,0 +1,64 @@
...
+`User user = getDelegate().login(x, userId, password);
+ if ( isPasswordExpired(user) ) {
+   throw new AuthenticationException("Password expired");
+ }
+ return user;`
+    },
+    {
+      name: 'loginByEmail',
+      javaCode:
+`User user = getDelegate().loginByEmail(x, email, password);
```
adamvy: This will end up creating a session and putting into the session dao even though we're rejecting the login due to stale password. Are we sure this is the behaviour we want?

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -35,6 +35,13 @@
   // pattern used to check if password has only alphanumeric characters
   java.util.regex.Pattern alphanumeric = java.util.regex.Pattern.compile("[^a-zA-Z0-9]");
 
+  public UserAndGroupAuthService(X x) {
+    setX(x);
```
adamvy, line 39: Do we need this?  Couldn't we just proxy the start() method in our PasswordExpireAuthService?

src/foam/nanos/auth/PasswordExpiryAuthService.js
```diff
@@ -0,0 +1,64 @@
...
+        {
+          name: 'user',
+          javaType: 'foam.nanos.auth.User'
+        }
+      ],
+      javaCode:
+`if ( user == null ) return true;
+if ( user.getPasswordExpiry() == null ) return false;
+
+// if we are after the expiry date then prevent login
```
kgrgreer, line 69: Does this do the same thing:?
return user.getPasswordExpiry().getTime() < System.currentTimeMillis();

adamvy: Or what about user.getPasswordExpiry().compareTo(new Date()) ?

kgrgreer: Second version creates a new object whereas the first version doesn't.

src/foam/nanos/auth/PasswordExpiryAuthService.js
kgrgreer: Good point.

src/foam/nanos/auth/PasswordExpiryAuthService.js
kirkeaton: I went with the former

adamvy (approved):

PR #1100, opened 2018-03-13
Add toString to FObject

hchoura:


src/foam/core/AbstractFObject.java
```diff
@@ -197,4 +197,26 @@ public boolean verify(byte[] signature, String algorithm, PublicKey key) {
...
+
+  public void append(StringBuilder sb)  {
+    List props = getClassInfo().getAxiomsByClass(PropertyInfo.class);
+    Iterator i = props.iterator();
+
+    try {
+      while (i.hasNext()) {
+        PropertyInfo prop = (PropertyInfo) i.next();
+        if (!(prop.get(this) instanceof Appendable)) {
+          sb.append(prop.getName() + " " + prop.get(this) );
```
kgrgreer: Replace this if statement with:

sb.append(prop.getName());
sb.append(" ");
Object value = prop.get(this);
if ( prop instanceof Appendable ) {
  ((Appendable) value).append(sb);
} else {
  sb.append(String.valueOf(value));
}

PR #1102, opened 2018-03-14
Return object on dig put request.

zahnno:
DIG put requests now outputs objects in json format.


jlhughes (changes requested):
The same 'out' processing is required around line 220/226.  This class presently has duplicate command processing for web and api requests.

jlhughes (approved):

PR #1105, opened 2018-03-15
Java query parser

hchoura:


src/foam/java/PropertyInfo.js
```diff
@@ -64,6 +77,18 @@ foam.CLASS({
...
+            name: 'getShortName',
+            visibility: 'public',
+            type: 'String',
+            body:  this.propShortName ? 'return "' +this.propShortName+'";' : 'return null;'
+          },
+          {
+            name: 'getAliases',
+            visibility: 'public',
+            type: 'String[]',
+            body: 'return ' + this.buildBody
```
adamvy: "buildBody" is a pretty generic name.  We only appear to use buildBody in this spot.  Can we just inline the buildBody expression here?

src/foam/java/refinements.js
```diff
@@ -34,6 +34,11 @@ foam.CLASS({
       name: 'javaJSONParser',
       value: 'foam.lib.json.AnyParser.instance()'
     },
+    {
+      class: 'String',
+      name: 'javaQueryParser',
+      value: 'foam.parse.AnyParser.instance()'
```
adamvy: Instead of value: 'foam.parse.AnyParser.instance()' let's use expression: function(javaJSONParser) { return javaJSONParser; }

This way the default javaQueryParser is the same as the JSON parser, so we don't have to specify it if it's the same.

src/foam/java/refinements.js
```diff
@@ -552,6 +568,7 @@ foam.CLASS({
     ['javaType', 'int'],
     ['javaInfoType', 'foam.core.AbstractIntPropertyInfo'],
     ['javaJSONParser', 'new foam.lib.json.IntParser()'],
+    ['javaQueryParser', 'new foam.lib.json.IntParser()'],//'new foam.parse.IntParser()'
```
adamvy: After you add the expression to javaJSONParser then we can remove all these javaQueryParser statements when they are the same as the javaJSONParser.

src/foam/lib/parse/Literal.java
```diff
@@ -31,6 +31,7 @@ public Literal(String s, Object v) {
   }
 
   public PStream parse(PStream ps, ParserContext x) {
+    if ( string ==null) return null;
```
adamvy: You should have a space between "==" and "null" as well as between "null" and ")" to match the FOAM style guide.

src/foam/parse/DateParser.java
```diff
@@ -0,0 +1,229 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.parse;
```
adamvy, line 7: Can we move this to a sub package like foam.parse.query, as this is pretty specific to the QueryParser

src/foam/parse/DateParser.java
```diff
@@ -0,0 +1,229 @@
...
+//today-7
+public class DateParser
+  extends ProxyParser
+{
+  public DateParser() {
+    super(
+        new Alt(
+
+          //YYYY-MM-DD..YYYY-MM-DD
+          new Seq(
```
adamvy, line 39: You have set up a bit Alt() of all the different date formats.  Let's move each potential date format such as "YYYY-MM-DD..YYYY-MM-DD" or "today-7" to their own Parser objects.  Then this class can just become

new Alt(
  new YYYMMDDParser(),
  new YYYYMMDDRangeParser(),
  ...);

That way the body of this parser doesn't have to do a bunch of conditionals to detect what the format of the query was.

You should try to come up with reasonable names for all the sub cases.

src/foam/parse/EnumParserFactory.java
```diff
@@ -0,0 +1,68 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.parse;
```
adamvy, line 7: Does this class differ at all from the EnumParserFactory in foam/lib/json/EnumParserFactory.java

src/foam/parse/ExprParser.java
```diff
@@ -0,0 +1,19 @@
...
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.parse;
+
+import foam.lib.parse.Alt;
+
+public class ExprParser extends foam.lib.parse.ProxyParser {
```
adamvy, line 11: Why can't we use the ExprParser in foam/lib/json/ExprParser.java ?

src/foam/parse/FObjectArrayParser.java
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.parse;
+
+import foam.lib.json.Whitespace;
+import foam.lib.parse.*;
+
+public class FObjectArrayParser extends ProxyParser {
```
adamvy, line 12: Again this looks the same as the parser in foam/lib/json/FObjectArrayParser.java let's just use that if we can.

src/foam/parse/FObjectParser.java
```diff
@@ -0,0 +1,75 @@
...
+
+import foam.core.X;
+import foam.lib.parse.Alt;
+import foam.lib.parse.PStream;
+import foam.lib.parse.Parser;
+import foam.lib.parse.ParserContext;
+import foam.lib.parse.ProxyParser;
+import foam.mlang.predicate.Predicate;
+
+public class FObjectParser extends ProxyParser {
```
adamvy, line 20: Again, can we use the one in foam/lib/json/FObjectParser.java ?  Are there any differences here that we need?

src/foam/parse/IdentifierStartParser.java
```diff
@@ -0,0 +1,15 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.parse;
```
adamvy, line 7: Again, use the one in foam/lib/json if we can, or explain why we cant.

src/foam/parse/PropertyParser.java
```diff
@@ -0,0 +1,145 @@
...
+import foam.lib.parse.Alt;
+import foam.lib.parse.Literal;
+import foam.lib.parse.PStream;
+import foam.lib.parse.ParserContext;
+import foam.lib.parse.ProxyParser;
+import foam.lib.parse.Seq3;
+import foam.mlang.MLang;
+import foam.mlang.predicate.Predicate;
+
+public class PropertyParser extends ProxyParser {
```
adamvy, line 23: I think this would be better named PropertyPredicateParser or PropertyConditionParser as its not parsing a Property value like the foam/lib/json/PropertyParser is, but rather it's parsing a property predicate/conditional.

src/foam/lib/parse/InvalidPStream.java
```diff
@@ -38,11 +43,21 @@ public PStream tail() {
   public Object value() {
     return null;
   }
+    
+  @Override
+  public Object operator() {
```
adamvy, line 48: "operator" is very specific to query parsing, but PStream is supposed to be used for all parsers.  We shouldn't add "operator" support here, but rather just implement the behaviour we need in the appropriate query parser.

src/foam/lib/parse/Seq3.java
```diff
@@ -0,0 +1,40 @@
...
+
+import foam.lib.json.KeyParser;
+
+public class Seq3 implements Parser {
+  private Parser[] parsers;
+  private int      index;
+  private int      opIndex;
+  private Object   operator;
+
+  public Seq3(int i, int j, Parser... args) {
```
adamvy, line 17: This class is hard coded to pass something called "operator" through.  I would expect Seq3 to take 3 index values, just like Seq2 takes 2 values, and Seq1 takes 1 value, and Seq0 takes 0 values.

If we take int i, int j and int k, and pass all three values via ps.setValue(), then we don't need to modify PStream to support operator()/setOperator()

src/foam/parse/DateParser.java
```diff
@@ -0,0 +1,229 @@
...
+
+//YYYY
+//YYYY-MM               YYYY/MM
+//YYYY-MM-DD            YYYY/MM/DD
+//YYYY-MM-DDTHH         YYYY/MM/DDTHH
+//YYYY-MM-DDTHH:MM      YYYY/MM/DDTHH:MM
+//YYYY-MM-DD..YYYY-MM-DDYYYY-MM-DD..YYYY-MM-DD
+//today
+//today-7
+public class DateParser
```
adamvy, line 31: Maybe DateRangeParser or DateQueryParser would be a more appropriate name.

src/foam/parse/EnumParserFactory.java
```diff
@@ -0,0 +1,68 @@
...
+      return parsers_.get(c);
+    }
+
+    Parser parser = buildInstance_(c);
+    parsers_.put(c, parser);
+    return parser;
+  }
+
+  public static Parser buildInstance_(Class c) {
+    return new Repeat0(new Seq0(
```
adamvy, line 35: I think Enum values for queries will need a very different parser.  If we have a situation for

foam.ENUM({
  name: 'FooEnum',
  values: [
    'FOO',
    'BAR'
  ]
});

foam.CLASS({
  name: 'Foo',
  properties: [
    { class: 'Enum', of: 'FooEnum' name: 'foo' }
  ]
});
```

We'll want to parse queries like

foo:BAR or -foo:FOO

This looks like it's trying to parser Enum values as JSON.

File: src/foam/parse/ModelParserFactory.java [hunk 17]
```diff
@@ -0,0 +1,76 @@
...
+import foam.core.PropertyInfo;
+import foam.lib.json.Whitespace;
+import foam.lib.parse.Alt;
+import foam.lib.parse.Optional;
+import foam.lib.parse.Parser;
+import foam.lib.parse.Repeat0;
+import foam.lib.parse.Seq;
+import foam.lib.parse.Seq1;
+
+public class ModelParserFactory {
```
adamvy, line 24: I think we should call this ModelQueryParserFactory since it's building a QueryParser for a specific model, not a value parser like the ModelParserFactory in foam/lib/json does.

Also let's move this and related classes to a subpackage.

src/foam/parse/NotExpression.java
```diff
@@ -0,0 +1,41 @@
...
+            new Literal("-")));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse(ps, x);
+    if (ps == null)
+      return null;
+
+    if (ps.value().toString().toLowerCase().equals("not") || ps.value().toString().equals("-")) {
```
adamvy, line 36: You shouldn't need to check ps.value(), super.parse() will return null if the value wasn't one of your specified values of NOT/Not/not/-

Also instead of adding all the different capitalizations of "not"  Let's add a LiteralIC class under foam/lib/parse which is similar to Literal() but will ignore the case of the argument.  Then we can simplify this to just new Alt(new LiteralIC("not"), new Literal("-"));

src/foam/parse/QueryParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+        "birthday:2010-9",
+        "birthday:2010/9",
+        "birthday:2010-9-10",
+        "birthday>=2010-09-10",
+        "id>5",
+        "lastLogin:today",
+        "lastLogin:today-2",//that birthday data is not included
+        "lastLogin:2010-9-10..2020-9-10",
+        "f=Simon",
+        "fname=Simon",
```
adamvy, line 141: Let's add some tests for Enum values when you fix the Enum parser.

src/foam/parse/StringParser.java
```diff
@@ -0,0 +1,71 @@
...
+import foam.lib.parse.Parser;
+import foam.lib.parse.ParserContext;
+
+public class StringParser implements Parser {
+  public final static char ESCAPE = '\\';
+
+  public StringParser() {
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
```
adamvy, line 21: Does this differ from the JSON one?

adamvy (changes requested):
I think we can make the QueryParser more tree like.  We're currently passing around this List of predicates and mutating it in different spots.  Instead we should set up the parser in a way that sub-parsers match a property name, a predicate, and a value, and then a parser above them combines them into an MLang predicate.  And a parser above them matches predicates with AND and OR logic and combines them that way.

We should be able to remove these large switch statements by setting up parsers for each Mlang.

Look at the grammar in QueryParser.js structured

A query is an OR value, which is a series of AND values combined via the literal "or" or "|"

Structuring the grammar this way also defines the precendence of AND vs OR based upon how the parser are structured.

src/foam/lib/query/DateQueryParser.java
```diff
@@ -0,0 +1,59 @@
...
+            //today-7
+            //today
+            new RelativeDateParser()));
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse(ps, x);
+
+    if ( ps == null || ps.value() == null ) return null;
+    if ( ps.value() instanceof java.lang.Long ) { return ps.setValue(new java.util.Date((java.lang.Long) ps.value())); }
```
adamvy, line 49: No need for curly braces if its only one statement after the if.

src/foam/parse/PropertyParser.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  protected Predicate createDatePredicate(String operator, Object dates) {
+    Predicate p = null;
+    Object date1 = ((Object[]) dates)[0];
+    Object date2 = ((Object[]) dates)[1];
+
+    switch (operator) {
+    case ":":
+    case "=":
```
adamvy, line 87: See how in my example we don't need the addition of "operator()" to the PStream?

Instead the operator() is defined by the particular parser that matches, the LtExpression parser just returns an actual foam.mlang.predicate.Lt(), so we don't need to add these switch statements.

Can we move this towards my style?

src/foam/parse/PropertyParser.java
```diff
@@ -0,0 +1,145 @@
...
+import foam.lib.parse.ProxyParser;
+import foam.lib.parse.Seq3;
+import foam.mlang.MLang;
+import foam.mlang.predicate.Predicate;
+
+public class PropertyParser extends ProxyParser {
+  private PropertyInfo property;
+
+  public PropertyParser(PropertyInfo p) {
+    super(new Seq3(4, 3,
```
adamvy, line 27: In my latest example, you can see how we actually construct a unique expression parser for each property.  This means we don't need a generic PropertyParser like you have here.  This parser is rather large and complicated because it has to handle every possible type of property and value combination.  In my example, we don't have any PropertyParser anymore because for any given property there are only a few valid expression for it.  Numeric properties can only be less-than, greather-than, less-than-or-equal, greater-than-or-equal, or equal to a value.  String properties have different possibilties, date properties have their own set of possibilities.

By constructing a different expression parser for each property type, we can avoid a mega parser like this and having more smaller parsers that are easier to understand and maintain.  It also means the system is more extensible, as new properties types don't require updating this one class.

adamvy (changes requested):
Two main comments.

First of all this diff is a bit confusing, it looks like you left some dead code around.  Is src/foam/parse/DateParser.java still being used?  IT looks like you split it up into a several parsers under the "query" package, but left the old code in place as well.

Let's remove any code that is no longer needed so the diff is easier to follow.


Secondly take a look at my updated example here: https://github.com/foam-framework/foam2/tree/master/src/foam/lib/query

In particular, look at QueryParser.java, and you will see how it actually constructs a query parser specific to the class info that it is passed.

The advantage of this style means that for Date properties we actually go down different parse trees for each possible property.

This lets us remove conditionals in the parse() method of parser that check if the value is a Long but we wanted a Date, or similar.  It also means we won't parse EqExpression for a Date field but will only parse Lt, Gt or During.

For the next iteration, let's try to remove more of these conditionals from the parser bodies, and instead construct a parser specific to the ClassInfo that we are parsing.

PR #1109, opened 2018-03-16
Remove GENMODEL and make the app controller deal with the ClientBuilder and using the right context

mcarcaso:
The motivation behind this is that GENMODEL was only every being used here and was causing special casing in the classloader/modeldao. To keep class loading simple, I'm getting rid of GENMODEL and making the ApplicationController deal with creating the Client via the ClientBuilder.

Note that this will break anybody depending on GENMODEL but there's only one user of it that I know of and I'll have a fix for them that.


src/foam/nanos/client/ClientBuilder.js
```diff
@@ -9,18 +9,19 @@ foam.CLASS({
...
     'foam.box.HTTPBox',
     'foam.box.RetryBox',
-    'foam.dao.ClientDAO',
+    'foam.dao.EasyDAO',
     'foam.dao.RequestResponseClientDAO',
-    'foam.nanos.boot.NSpec'
+    'foam.nanos.boot.NSpec',
   ],
 
+  axioms: [ foam.pattern.Singleton.create() ],
```
adamvy: Why singleton?

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -28,18 +28,19 @@ foam.CLASS({
   documentation: 'FOAM Application Controller.',
 
   implements: [
-    'foam.nanos.client.Client',
-    'foam.nanos.controller.AppStyles'
+    'foam.box.Context',
+    'foam.nanos.controller.AppStyles',
```
adamvy: FOAM style guide says no trailing commas.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -129,37 +145,42 @@ foam.CLASS({
   methods: [
     function init() {
       this.SUPER();
-      var self = this;
 
-      this.getCurrentUser();
+      var self = this;
+      self.clientPromise.then(function() {
+        self.getCurrentUser();
```
adamvy: I kindof think you shoudl set the subSubContext here rather than in the postset you did earlier, I think it's more clear why you're setting it if you do this, and any race conditions between creating child element and the setting of subSubContext will be more apparent.

src/foam/nanos/client/ClientBuilder.js
```diff
@@ -38,98 +39,65 @@ foam.CLASS({
...
-        }
-      });
+    },
+    {
+      name: 'promise',
+      factory: function() {
+        var self = this;
+        return new Promise(function(resolve) {
+          var client = {
+            package: 'foam.nanos.client',
```
adamvy: I still think that ClientBuilder should be building just a Context and not a Class, otherwise we can't talk to more than one nanos server at time.

But I guess we can leave that as a TODO for now.

adamvy: But add a TODO for it.

src/foam/nanos/client/ClientBuilder.js
mcarcaso: Because it seems like something that can only really be used once per Context. Not sure why you'd ever want to instantiate more than one. It's no trouble to make it not a singleton but that was my rational here. What do you think? Is that boxing us into something that I'm not seeing?

src/foam/nanos/controller/ApplicationController.js
mcarcaso: Removed here but I'm sure I'm doing this in other places. This feels like something a tool should do for us :p

src/foam/nanos/controller/ApplicationController.js
mcarcaso: Makes sense. Done.

src/foam/nanos/client/ClientBuilder.js
mcarcaso: Will do. The advantage to building a class here is that it makes use of properties with factories that are lazy. Contexts should have some sort of equivalent but I dont think that's a thing in JS. I'll add a TODO as suggested.

src/foam/nanos/client/ClientBuilder.js
mcarcaso: Hmm I guess it does prevent us from being able to build clients from different servers right? This class doesn't really allow for that because the nspecDAO url is hardcoded but that's easy to make configurable if we want to do that.  I'll make it not a singleton.

adamvy (approved):

mcarcaso:
Last call on this. Will submit it tomorrow.

PR #1116, opened 2018-03-20
 Added extra message when there is no data on select

minsunny:
when there is no selected data, it should print out saying 'no data'


minsunny:
https://github.com/nanoPayinc/NANOPAY/issues/1013

PR #1117, opened 2018-03-21
Added javaAssertValue

kirkeaton:
Added javaAssertValue which, if set, asserts the value before setting the variable

### DIFF
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -56,13 +56,15 @@ foam.CLASS({
     },
     {
       class: 'String',
-      name: 'javaCloneProperty',
-      value: null
+      name: 'javaCloneProperty'
     },
     {
       class: 'String',
-      name: 'javaDiffProperty',
-      value: null
+      name: 'javaDiffProperty'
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

src/foam/java/refinements.js
```diff
@@ -112,6 +117,7 @@ foam.CLASS({
       var constantize = foam.String.constantize(this.name);
       var isSet       = this.name + 'IsSet_';
       var factoryName = capitalized + 'Factory_';
+      var assertName  = capitalized + 'Assert_';
```
kgrgreer, line 120: I think the name should be more friendly in this case. So getFoo(), setFoo(), and assertFoo(). Then below, instead of checking if javaSetter is set or not, we always make the assertFoo() call. This is friendlier for sub-classing and for those marking their models as abstract and putting larger code-blocks in a Java file. In the case where the implementation is empty, Java will just eliminate the call anyway.

kirkeaton: Done

PR #1120, opened 2018-03-22
Fix print X to string very slow

amymeng5277:
Fix print X to string very slow


amymeng5277:
@kgrgreer could you merge my code?

PR #1124, opened 2018-03-29
Output changed field

xuerongCode:
output only changed property to journal

Ticket:
https://github.com/nanoPayinc/NANOPAY/issues/1346

Test scenarios for the testers :
1) modify a property value of a model. 
For example: modify user firstname for a user. 
The changes will print out into the journals/user
The JSON in the "journals/user" should only contain fields of class, id, and firstname
* You may see that other properties are also printed out. This is because that the property is AbstractObjectPropertyInfo or AbstractArrayPropertyInfo, or it nests AbstractObjectPropertyInfo or AbstractArrayPropertyInfo. I currently did not implement the array compare. So the array should be printed out also. If you find a model contain a array, but it does not be printed out, then that will be a bug.
2) modify a nested property value. For example  user->phone
3) modify a Enum type property. (I have not tested)
4) modify a array property. see if array be printed out to journal
5) reboot the system(./run-tomcat), to see if all changes still remain
6) delete a entry, see if only id property be outputted
7) reboot the system, to see if entry is deleted
    * NOTE: the most recent build system - journals will not be deleted.


kgrgreer:
@xuerongCode Do you plan on implementing compare for arrays?

xuerongCode:
@kgrgreer Yes, I plan to do that. Just think how to do it for now, then will also implement it.

jlhughes:
I'm good with this to get testing started.

PR #1125, opened 2018-04-02
 Added filtering and searching features on PermissionTableView

minsunny:
on the left side of the top of the view, filtering and searching features are added.
They allow to search Permissions by Text, and filter Groups by checkbox, so the table is supposed to be shown according to these options.

https://github.com/nanoPayinc/NANOPAY/issues/1434

### DIFF
--- a/src/foam/nanos/auth/PermissionTableView.js
+++ b/src/foam/nanos/auth/PermissionTableView.js
@@ -17,8 +17,7 @@ foam.CLASS({
 
   requires: [
     'foam.nanos.auth.Group',
-    'foam.nanos.auth.Permission',
-    'foam.u2.search.SearchManager'
+    'foam.nanos.auth.Permission'
   ],
 
   css: `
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -26,32 +26,88 @@ foam.CLASS({
...
   `,
 
+  properties: [
+    {
+      class: 'String',
+      name: 'query',
+      view: {
+        class: 'foam.u2.TextField',
+        type: 'search',
+        placeholder: 'Search',
```
kgrgreer: Maybe something more descriptive like "Permission" or "Permission Search".

src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -26,32 +26,88 @@ foam.CLASS({
...
+        onKey: true
+      }
+    },
+    {
+      name: 'selected'
```
kgrgreer: What is this property used for?

PR #1127, opened 2018-04-04
Abstract predicate

kirkeaton:
Consolidating Predicate and AbstractPredicate f function to resolve issue where the f function in the swift version of AbstractPredicate was throwing a fatalError instead of returning false.


src/foam/mlang/mlang.js
```diff
@@ -150,8 +150,13 @@ foam.INTERFACE({
     {
       name: 'f',
       swiftReturns: 'Bool',
+      javaReturns: 'boolean',
```
kgrgreer: This is refined in in mlangJava.js.

kirkeaton: Okay undid the change here. Thanks

PR #1128, opened 2018-04-05
ManyToManyRelationship Swift

kirkeaton:
Added support for ManyToManyRelationships in Swift.


src/foam/dao/Relationship.js
```diff
@@ -412,14 +428,19 @@ foam.CLASS({
       javaCode: `foam.core.FObject junction = (foam.core.FObject)getX().create(getJunction().getObjClass());
 getTargetProperty().set(junction, targetId);
 getSourceProperty().set(junction, getSourceId());
-return junction;
-`
+return junction;`,
+
+      swiftCode: `let junction: FObject = self.junction.create(x: __context__) as! FObject
+(targetProperty as? PropertyInfo)?.set(junction, value: targetId)
```
adamvy: Can we add swiftType to the properties so we don't have to cast here?

kirkeaton: Yeah no problem. I've just made the change

PR #1130, opened 2018-04-06
Java Query Parser

adamvy:
Making a pull request so I can make comments easily.


src/foam/lib/parse/Literal.java
```diff
@@ -31,6 +31,7 @@ public Literal(String s, Object v) {
   }
 
   public PStream parse(PStream ps, ParserContext x) {
+    if ( string == null ) return null;
```
adamvy: Is this necessary? Where are we ever creating a Literal where string is null, I feel like we should just throw in the constructor if the passed string is null.

src/foam/lib/parse/LiteralIC.java
```diff
@@ -0,0 +1,39 @@
...
+  public LiteralIC(String s, Object v) {
+    string = s;
+    value = v;
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
+    if ( string ==null) return null;
+    for ( int i = 0 ; i < string.length() ; i++ ) {
+      if ( ! ps.valid() ||
+          Character.toUpperCase(ps.head()) != string.toUpperCase().charAt(i) ) {
```
adamvy: Do string = s.toUpperCase() in the constructor so we don't have to upper case it every time parse is called.

src/foam/lib/query/AfterGtParser.java
```diff
@@ -0,0 +1,40 @@
...
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
+    if ( ps == null ) return ps;
+
+    foam.mlang.predicate.Binary predicate = new foam.mlang.predicate.Gt ();
+    predicate.setArg1 ( ( foam.mlang.Expr ) x.get ( "arg1" ) );
+
+    if ( ps.value () instanceof Object[] )
+      predicate.setArg2 ( ( foam.mlang.Expr ) ( ( Object[] ) ps.value () )[0] );
```
adamvy: Can you comment what you're doing here?  In what cases in ps.value() an Object[] and why do we only take the first element in that situation?

src/foam/lib/query/AndParser.java
```diff
@@ -1,31 +1,42 @@
...
+import foam.lib.parse.Seq0;
 
 public class AndParser
   extends foam.lib.parse.ProxyParser {
   public AndParser(Parser exprParser) {
     setDelegate(new Repeat(exprParser,
                            new Seq0(new foam.lib.json.Whitespace(),
-                                    new Not(new Literal("OR")))));
+                                    new Alt(new LiteralIC("AND "),
+                                            new Literal(" "),
```
adamvy: You shouldn't need New Literal(" ") as that should be captured by the whitespace parser before.

src/foam/lib/query/CompoundValue.java
```diff
@@ -0,0 +1,28 @@
...
+                        new OrValue(),
+                        new AndValue()));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
+    if ( ps == null || ps.value () == null ) return null;
+
+    return ps.setValue ( ps.value () );
```
adamvy: You can just return ps;  ps.setValue(ps.value()) is basically a no-op that consumes memory.

src/foam/lib/query/charParser.java
```diff
@@ -0,0 +1,32 @@
...
+
+package foam.lib.query;
+
+import foam.lib.parse.Alt;
+import foam.lib.parse.Literal;
+import foam.lib.parse.PStream;
+import foam.lib.parse.ParserContext;
+import foam.lib.parse.Range;
+
+public class charParser extends foam.lib.parse.ProxyParser {
```
adamvy: Class names should start with a capital letter "CharParser"

src/foam/lib/parse/StringPStream.java
```diff
@@ -77,4 +78,12 @@ public String substring(PStream end) {
   public PStream apply(Parser ps, ParserContext x) {
     return ps.parse(this, x);
   }
+
+  public int decrement() {
+    return this.pos = this.pos == 0 ? 0 : this.pos-1;
```
adamvy: PStream is designed to be a purely functional object, its state shouldn't be mutated like this.

Instead of decrementing, you can backtrack by just remembering the previous pstream object you had a reference too.

src/foam/lib/parse/StringPStream.java
```diff
@@ -77,4 +78,12 @@ public String substring(PStream end) {
...
+  }
+
+  public char beforeHead() {
+    return pos == 0 ?  head() :str.get().charAt(pos-1);
```
adamvy: Same thing here, you can backtrack by remembering the previous pstream object, I don't think we should need to add this.

src/foam/lib/query/StringParser.java
```diff
@@ -0,0 +1,83 @@
...
+
+public class StringParser implements Parser {
+  public final static char ESCAPE = '\\';
+
+  public StringParser() {
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
+
+    if ( !ps.valid() )  return null;
```
adamvy, line 23: I think our purposes can be accomplished by just parsing any character that isn't whitespace.

We're only using this string parser to parse values now, not to parse property named.

I don't see why beforeHead() would be a userful delimiter, if beforeHead was a ":" that doesn't mean the string is going to end with a colon.

Just capturing characters until the next whitespace or close parenthesis should be sufficient.

src/foam/lib/query/StringParser.java
```diff
@@ -0,0 +1,83 @@
...
+
+    char lastc = delim;
+
+    // TODO: use thread-local SB instead to avoid generating garbage
+    StringBuilder sb = new StringBuilder();
+
+    while ( ps.valid() ) {
+      char c = ps.head();
+      if ( c == ')' ) {
+        ps.decrement();
```
adamvy: Instead of decrementing, you can just hold on to the previous pstream in a different variable.

Look at how foam.lib.json.StringParser stores the pstream in a separate variable named tail, and updates ps = tail at the end of its inner loop.

src/foam/lib/query/AfterGtParser.java
hchoura: The only case that we will get Object[] is when we parse date. For each date parsed, we generate two date.(beginning and the end date, since the date contain milli seconds ).
The first date is the concrete date and the second date is the complement of the first.(Example d=2000/1; d1=2000/1/1 :00:00:00:000; d2=2000/1/31:24:59:59:999.
The second date in used only in the equal or in a partial date (like year).

src/foam/lib/parse/StringPStream.java
```diff
@@ -55,7 +55,8 @@ public boolean valid() {
   }
 
   public PStream tail() {
-    if ( tail_ == null ) tail_ = new StringPStream(str, pos + 1);
+    //if ( tail_ == null )
+    tail_ = new StringPStream(str, pos + 1);
```
adamvy: Can we undo this change here?  It shouldn't be necessary now that you've restored the immutable aspect of pstreams.

src/foam/java/PropertyInfo.js
```diff
@@ -75,6 +88,18 @@ foam.CLASS({
             type: 'String',
             body: 'return "' + this.propName + '";'
           },
+          {
+            name: 'getShortName',
+            visibility: 'public',
+            type: 'String',
+            body:  this.propShortName ? 'return "' +this.propShortName+'";' : 'return null;'
```
adamvy, line 95: FOAM style i believe would put a space on both sides of the +

src/foam/lib/parse/LiteralIC.java
```diff
@@ -0,0 +1,39 @@
...
+    this(s, s);
+  }
+
+  public LiteralIC(String s, Object v) {
+    string = s.toUpperCase();
+    value = v;
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
+    if ( string ==null) return null;
```
adamvy: FOAM style is `if ( string == null ) return null;`  you're missing spaces around the first null

src/foam/lib/query/AfterGtParser.java
```diff
@@ -0,0 +1,40 @@
...
+  public AfterGtParser(Parser valueParser) {
+    setDelegate(new Seq1(1,
+                         new Literal(">"),
+                         valueParser
+                        ));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
```
adamvy: FOAM style is super.parse(ps, x); no extra spaces for method calls, only for control contstructs like "if" "while" etc.

src/foam/lib/query/AfterGtParser.java
adamvy: But this is the AfterGtParser, so it doesn't need to take an array of values.

 If you look at how the example I shared was structured, there's no way for the AfterParser to parse a date range, as it only includes a single date as its value.  This results in less fixing up of values in the parse() method, because we can only parse a single date for before/after situations.

src/foam/lib/query/AfterGteParser.java
```diff
@@ -0,0 +1,43 @@
...
+    setDelegate(new Seq1(1,
+                        new Alt(new Literal(">="),
+                                new LiteralIC("-after:")),
+                        valueParser
+                        ));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
```
adamvy: FOAM style is `ps = super.parse(ps, x);`

src/foam/lib/query/AfterGteParser.java
```diff
@@ -0,0 +1,43 @@
...
+    if ( ps == null ) return ps;
+
+    foam.mlang.predicate.Binary predicate = new foam.mlang.predicate.Gte ();
+    predicate.setArg1 ( ( foam.mlang.Expr ) x.get ( "arg1" ) );
+
+    if ( ps.value () instanceof Object[] )
+      predicate.setArg2 ( ( foam.mlang.Expr ) ( ( Object[] ) ps.value () )[0] );
+    else predicate
+        .setArg2 ( ( ps.value () instanceof foam.mlang.Expr ) ? ( foam.mlang.Expr ) ps
+            .value () : new foam.mlang.Constant ( ps.value () ) );
```
adamvy: FOAM style would be `.value() : new foam.mlang.Constant(ps.value()));`

src/foam/lib/query/AfterGteParser.java
```diff
@@ -0,0 +1,43 @@
...
+                        ));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
+    if ( ps == null ) return ps;
+
+    foam.mlang.predicate.Binary predicate = new foam.mlang.predicate.Gte ();
+    predicate.setArg1 ( ( foam.mlang.Expr ) x.get ( "arg1" ) );
```
adamvy: FOAM style would be `.setArg1((foam.mlang.expr)x.get("arg1"));`

src/foam/lib/query/AfterGteParser.java
```diff
@@ -0,0 +1,43 @@
...
+
+    if ( ps.value () instanceof Object[] )
```
adamvy: FOAM style would be `if ( ps.value() instanceof Object[] )`

src/foam/lib/query/AndParser.java
```diff
@@ -1,31 +1,40 @@
...
+    foam.mlang.predicate.And and = new foam.mlang.predicate.And ();
 
-    Object[] values = (Object[])ps.value();
+    Object[] values = ( Object[] ) ps.value ();
 
     foam.mlang.predicate.Predicate[] args = new foam.mlang.predicate.Predicate[values.length];
 
-    for ( int i = 0 ; i < values.length ; i++ ) {
-      args[i] = (foam.mlang.predicate.Predicate)values[i];
+    for ( int i = 0; i < values.length; i++ ) {
```
adamvy: FOAM style would keep the spaces on both sides of the semicolon `for ( int i = 0 ; i < values...`

src/foam/lib/parse/LiteralIC.java
```diff
@@ -0,0 +1,39 @@
...
+    this(s, s);
+  }
+
+  public LiteralIC(String s, Object v) {
+    string = s.toUpperCase();
+    value = v;
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
+    if ( string ==null ) return null;
```
adamvy: We shouldn't need to check this, throw an error in the constructor if String is null, nobody should build a Literal or LiteralIC that has string as null

src/foam/lib/parse/StringPStream.java
adamvy: ping

src/foam/lib/query/CharParser.java
```diff
@@ -0,0 +1,32 @@
...
+    setDelegate(
+        new Alt(new Range('A', 'z'), new Range('0', '9'), new Literal("-"), new Literal("^"),
+            new Literal("_"), new Literal("@"), new Literal("%"), new Literal("."))
+        );
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
+    if ( ps == null || ps.value () == null ) return null;
```
adamvy: Why return null if ps.value() == null?  That case shouldnt' come up here.

src/foam/lib/query/CompoundValue.java
```diff
@@ -0,0 +1,28 @@
...
+  public CompoundValue() {
+    setDelegate(new Alt(new NegateValue(),
+                        new OrValue(),
+                        new AndValue()));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
+    if ( ps == null || ps.value () == null ) return null;
```
adamvy: Again why return null if ps.value() == null.  If super.parse() returns not null that means it believes it matches something, we should just pase it upstream.  In fact I don't think you need a parse() method implementation here at all, just pick up the default implementation from ProxyParser.

src/foam/lib/query/MeParser.java
```diff
@@ -0,0 +1,44 @@
...
+  public MeParser() {
+    setDelegate(
+        new Seq(new LiteralIC("me"),
+                new Not(new charParser())));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
+    if ( ps == null || ps.value () == null ) return null;
```
adamvy: Again don't check ps.value() == null

src/foam/lib/query/MeParser.java
```diff
@@ -0,0 +1,44 @@
...
+
+    if ( ( ( Object[] ) ps.value () )[0].equals ( "me" ) ) {
+      User user = ( User ) x.get ( "user" );
+      try {
+        if ( user == null )
+          throw new AccessControlException ( "User is not logged in" );
```
adamvy: This is overly complex to throw an exception and then catch it the next line, why not just if ( user == null ) { System.warn.println(); return null; }

src/foam/lib/query/NegateParser.java
```diff
@@ -0,0 +1,39 @@
...
+                        new Seq(new LiteralIC("NOT "),
+                                exprParser)));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    ps = super.parse ( ps, x );
+    if ( ps == null || ps.value () == null ) return null;
+
+    Object[] values = ( Object[] ) ps.value ();
```
adamvy: Instead of using Object[], if you use a Seq1 parser above for the two cases, they will just return the inner value.

src/foam/lib/query/NegateParser.java
```diff
@@ -0,0 +1,39 @@
...
+import foam.lib.parse.PStream;
+import foam.lib.parse.Parser;
+import foam.lib.parse.ParserContext;
+import foam.lib.parse.Seq;
+import foam.mlang.predicate.Not;
+
+public class NegateParser extends foam.lib.parse.ProxyParser  {
+
+  public NegateParser(Parser exprParser) {
+    setDelegate(new Alt(new Seq(new Literal("-"),
```
adamvy: Simpler to put the Alt in the sequence, and use a Seq1 so you don't need the Object[] dance inside.

```
setDelegate(new Seq1(1,
  new Alt(new Literal("-"), new LiteralIC("not "),
  exprParser));
```

Then in parse do

```
setArg1((foam.mlang.Expr)ps.value());
```

src/foam/lib/query/NegateValue.java
```diff
@@ -0,0 +1,32 @@
...
+            new Literal("("),
+            new Alt(new Literal("-"), new LiteralIC("not ")),
+            new ValueParser(),//("value"),
+            new Literal(")")
+        ));
+  }
+
+  @Override
+  public PStream parse(PStream ps, ParserContext x) {
+    return null;
```
adamvy: Is this just incomplete?  Can you leave a TODO saying what this parser is supposed to do when this is implemented?

PR #1132, opened 2018-04-11
Html doc serivce

waseem-chamaa:


src/foam/nanos/auth/email/EmailDocService.js
```diff
@@ -0,0 +1,70 @@
...
+  EmailMessage message = new EmailMessage();
+  message.setTo(new String[] { user.getEmail() });
+
+  HashMap<String, Object> args = new HashMap<>();
+  args.put("doc", doc.getBody());
+
+  email.sendEmailFromTemplate(user, message, "docEmail", args);
+  return true;
+}catch(Exception e){
+  e.printStackTrace();
```
jlhughes: Do we not have access to the Logger?

waseem-chamaa: Logger was added

PR #1135, opened 2018-04-12
Checking if a Thread is running before printing its stacktrace.

daudich:


blakegreen32 (commented):
I would reword the out.println message in the else statement, seems a little confusing to me.

zahnno (approved):

blakegreen32 (approved):

kirkeaton (approved):

PR #1136, opened 2018-04-12
LRUDAO, FixedSizeDAO, LoggerService

xuerongCode:
implement LRUDAO,
implement FixedSizedDAO
add Logger service
ticket: https://github.com/nanoPayinc/NANOPAY/issues/1388


xuerongCode:
@jlhughes @kgrgreer Please review

jlhughes (approved):

src/files.js
```diff
@@ -506,5 +506,7 @@ FOAM_FILES([
   { name: "foam/crypto/hash/Hashable" },
   { name: "foam/crypto/sign/Signer" },
   { name: "foam/crypto/sign/Signable" },
-  { name: "foam/crypto/sign/SignedFObject" }
+  { name: "foam/crypto/sign/SignedFObject" },
+  { name: 'foam/nanos/logger/Log', flags: ['web']},
```
adamvy: nanos files should be added to src/foam/nanos/nanos.js

src/foam/dao/FixedSizeDAO.java
```diff
@@ -0,0 +1,72 @@
...
+    if ( key == null ) {
+      throw new RuntimeException("Missing Primary Key in " + this.getOf().getId() + ".put()");
+    }
+    list.enQueue(obj);
+    onPut(obj.fclone());
+    return obj;
+  }
+
+  public FObject remove_(X x, FObject obj) {
+    //TODO
```
adamvy: What is the TODO here? At first glace this looks implemented, leave a more detailed comment if there's work to be done.

src/foam/dao/FixedSizeDAO.java
```diff
@@ -0,0 +1,72 @@
...
+package foam.dao;
+
+import java.util.Iterator;
+import foam.core.*;
+import foam.util.ConcurrentList;
+import foam.util.List;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+
+public class FixedSizeDAO
```
adamvy, line 15: I wonder if its possible to implement this as a decorator, one which just tracks the number of puts and removes going through it and throws an error if you exceed the maxSize number of puts.

Then we can use any suitable underlying storage DAO such as JDAO or MDAO or even ClientDAO rather than having to implemented another ArrayDAO here.

src/foam/nanos/logger/GUIDLogger.java
```diff
@@ -0,0 +1,55 @@
...
+public class GUIDLogger
+  extends AbstractLogger
+{
+  private DAO logDAO_ = null;
+  private Random r = new Random();
+  public GUIDLogger(X x) {
+    logDAO_ = (DAO) x.get("logDAO");
+  }
+
+  private FObject generateLogModel(String type, Object... args) {
```
adamvy, line 23: I may be confused but what makes this a GUIDLogger?  Where is a GUID coming in to play?

src/foam/nanos/logger/GUIDLogger.js
```diff
@@ -0,0 +1,116 @@
...
+  name: 'GUIDLogger',
+
+  imports: [
+    'logDAO',
+    'user'
+  ],
+
+  propertes: [
+    {
+      name: 'self',
```
adamvy: I'm not sure this is doing what you think its doing, and it doesn't appear to be used anywhere, delete it?

src/foam/nanos/logger/GUIDLogger.js
```diff
@@ -0,0 +1,116 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+foam.CLASS({
```
adamvy, line 6: If we have both a Java and JS implementation, what about generating the Java one from the model rather than writing it by hand?

src/foam/nanos/logger/GUIDLogger.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+  ],
+
+  requires: [
+    'foam.nanos.logger.Log'
+  ],
+
+  methods: [
+    function log() {
+      if (arguments.length > 1) {
```
adamvy: These method bodies look all the same except for the first parameter to addToDAO, make a helper method and call that rather than copying the code 5 times.

src/foam/nanos/logger/Log.js
```diff
@@ -0,0 +1,47 @@
...
+  name: 'Log',
+
+  tableColumns: ['time', 'from', 'belong', 'type', 'description', 'detail'],
+
+  properties: [
+    {
+      class: 'String',
+      name: 'id',
+      factory: function() {
+        return '' + new Date().getTime() + '-' + Math.floor(Math.random() * 100000);;
```
adamvy, line 17: You can add javaFactory too with the java equivalent code for this, then you don't have to put this logic in GUIDLogger.java but can keep it right beside the JS factory.  This makes it more readable as we can see the behaviour on all platforms side by side.

src/foam/nanos/logger/GUIDLogger.java
```diff
@@ -0,0 +1,55 @@
...
+    Date d = new Date();
+    Log log = new Log();
+    log.setId(Long.toString(d.getTime()) + "-" + Integer.toString(r.nextInt(100000), 5));
+    log.setTime(new Date());
+    log.setFrom("BackEnd");
+    log.setBelong(System.getProperty("user.name"));
```
adamvy: Do we really want to log user.name?  That's the operating system level user that nanos is running as.  I guess it could be useful, but I would have expected instead to log the user for whom we're completing this DAO request.

src/foam/nanos/logger/Log.js
```diff
@@ -0,0 +1,47 @@
...
+      class: 'DateTime',
+      name: 'time'
+    },
+    {
+      class: 'String',
+      name: 'from'
+    },
+    {
+      class: 'String',
+      name: 'belong',
```
adamvy: I find belong a confusing name, maybe just call this user?  Also add a documentation: '' field describing what we're going to store here.

src/foam/dao/LRUDAO.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+package foam.dao;
+
+import java.util.Map;
+import foam.core.*;
+import foam.util.ConcurrentLRU;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+
+public class LRUDAO
```
adamvy, line 14: You seem to have implemented LRU and ConcurrentLRU as java level collections, and then implemented LRUDAO as a DAO on top of these collection classes.

What if instead we implemented LRUDAO as a DAO Decorator, then we can apply it to any underlying DAO storage mechanism like an MDAO, and get high performance queries for free.

src/foam/util/List.java
```diff
@@ -0,0 +1,325 @@
...
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+package foam.util;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+public class List<O> {
```
adamvy, line 11: Yeah I think this would be better done at the DAO level so we can Decorate DAOs with a max size rather than Lists.

DAOs are more useful than Lists.

adamvy (changes requested):

src/foam/nanos/logger/GUIDLogger.java
kgrgreer: FOAM already inserts a prefix logger which logs the current user when there is one.

src/foam/nanos/logger/GUIDLogger.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package foam.nanos.logger;
+
+import java.util.Date;
+import java.util.Random;
+
+import foam.core.*;
+import foam.dao.DAO;
+
+public class GUIDLogger
```
kgrgreer, line 14: I think DAOLogger would be a better name.

src/foam/dao/FixedSizeDAO.java
```diff
@@ -0,0 +1,71 @@
...
+  extends AbstractDAO
+{
+  private List<FObject> list;
```
kgrgreer, line 18: Make instance variables be protected instead of private and end names with underscores. ie. rename list to list_.

src/foam/dao/FixedSizeDAO.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public FObject remove_(X x, FObject obj) {
+    FObject ret = find_(x, obj);
+    if ( ret == null ) return null;
+    return list.remove(ret, 0);
+  }
+
+  public FObject find_(X x, Object o) {
+    FObject ret = null;
+    Iterator<FObject> i = list.iterator();
```
kgrgreer, line 46: Easier to use Java's built-in iterator looping support:
for ( FObject i : list ) {

}

src/foam/dao/FixedSizeDAO.java
```diff
@@ -0,0 +1,71 @@
...
+    while( i.hasNext() ) {
```
kgrgreer, line 47: Add space between while and (. See FOAM coding style guide:

src/foam/nanos/logger/GUIDLogger.js
```diff
@@ -0,0 +1,83 @@
...
+      if ( foam.core.Exception.isInstance(e) ) {
+        return '' + e + '\n' + '[name]: ' + e.name + '\n' + '[message]: ' + e.message + '\n';
+      } else if ( e instanceof Error ) {
+        return '' + e + '\n' + '[name]: ' + e.name + '\n' + '[message]: ' + e.message + '\n';
+      } else {
+        return '' + e + '\n';
+      }
+    },
+    function normalizeDescription(e) {
+      return (!! e.name) ? '' + e.name : undefined;
```
kgrgreer, line 66: Why is the !! needed?
Why is the '' + needed?
Is this just the same as 'e.name'?

src/foam/nanos/logger/Log.js
kgrgreer: Add documentation to explain what you're doing and why.

src/foam/util/ConcurrentLRU.java
```diff
@@ -0,0 +1,83 @@
...
+package foam.util;
+
+import java.util.*;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+public class ConcurrentLRU<K, V>
+  extends LRU<K, V>
+{
+  private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock();
```
kgrgreer, line 15: Here and elsewhere, make private vars protected and add underscore suffix.

src/foam/nanos/logger/Log.js
kgrgreer: Why is this a random number rather than just a sequence number?

kgrgreer:
I think this whole thing could be greatly simplified and optimized but just having a CircularLogBuffer which uses a fixed size circular buffer to store new log entries. It also adapts this internal buffer into a partial-DAO. This DAO would only need to implement select() and maybe find().

PR #1138, opened 2018-04-13
Merge HttpParametersWebAgent into NanoRouter

jlhughes:
HttpParametersWebAgent provided support for processing parameter data from our UI Dig forms, as well as conventional POST, GET parameters, such that the consuming WebAgent didn't have to be concerned where they were coming from. 
Specifically 'data' could be either a URL parameter or built from the request.reader. 
Common parameters such as 'dao', 'cmd', 'format' are mapped to their Enum values. 

Previously the HttpParametersWebAgent was a WebAgentProxy which wrapped DigWebAgent for example.  This functionality has been merged into NanoRouter.  

NANOPAY-1648


jlhughes:
This implementation does not work as the HttpParameters logic consumes authentication credentials.  Better as a Proxy installed via nspec flags.

PR #1144, opened 2018-04-20
changing all PMInfo properties to camelCase

daudich:


src/foam/nanos/pm/PMInfo.js
```diff
@@ -4,54 +4,53 @@
...
     {
       class: 'String',
-      name: 'pmname',
+      name: 'pmName',
       label: 'Name'
     },
     {
       class: 'Int',
-      name: 'numoccurrences',
+      name: 'numOccurrences',
```
kgrgreer: I think just 'count' would be better. Who can remember how to spell occurrences.

kirkeaton (approved):
LGTM

src/foam/nanos/pm/PMInfo.js
daudich: Agree, changing it now.

PR #1151, opened 2018-04-25
Discard logger

deejes:
Adding a DiscardLogger to prevent null logs from being passed to ProxyLogger


src/foam/nanos/logger/DiscardLogger.java
```diff
@@ -0,0 +1,10 @@
+package foam.nanos.logger;
+
+public class DiscardLogger
+  extends ProxyLogger
+
+info(m) {
```
jlhughes: So the java code is implemented in the js file, see foam/nanos/http/DefaultHttpParameters.js for an example.  Even though, at the moment, this is used entirely on the java side, we'll still add the   info(... method implementation for java in the js.

src/foam/nanos/logger/DiscardLogger.js
```diff
@@ -0,0 +1,57 @@
...
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.logger',
+  name: 'DiscardLogger',
+  implements: [ 'foam.nanos.logger.Logger' ],
+  extends: [ 'ProxyLogger' ],
```
jlhughes: so this is just 
extends: foam.nanos.logger.ProxyLogger,
place if after name:

src/foam/nanos/logger/DiscardLogger.js
```diff
@@ -0,0 +1,57 @@
...
+
+  properties: [
+    {
+      name: 'discardInfo',
+      class: 'Boolean',
+      value: 'false'
```
jlhughes: boolean values are unquoted.

src/foam/nanos/logger/DiscardLogger.js
```diff
@@ -0,0 +1,57 @@
...
+  methods: [
+    {
+    name: 'info',
+    args: [
+      {
+      name: 'message',
+      javaType: 'String'  
+      }
+    ],
+    javaCode: `
```
jlhughes: the javaCode is just the method implementation so between
public void info(Object... args) {
// and 
}

so just the 
  if ( ! getDiscardInfo() ) {
    getDelegate().info(args);
  }

src/foam/nanos/logger/DiscardLogger.js
```diff
@@ -0,0 +1,57 @@
...
+      value: 'false'
+    }
+  ],
+    
+  methods: [
+    {
+    name: 'info',
+    args: [
+      {
+      name: 'message',
```
jlhughes: matching the interface, 
name: 'args'
with javaType: 'Object...'  (including the tree dots - for varargs)

jlhughes (commented):
These would be hard to notice when looking for examples.

src/foam/nanos/logger/DiscardLogger.js
deejes: Not sure what you mean by - "place if after name:"
Assuming it was a typo for 'it'

src/foam/nanos/logger/DiscardLogger.js
```diff
@@ -0,0 +1,99 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.logger',
+  name: 'DiscardLogger',
```
kgrgreer: Why is this called DiscardLogger instead of FilteredLogger and the boolean values not reversed to be enable instead of discard? This seems to be the reverse of normal?

src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+ 
+foam.CLASS({
+  package: 'foam.nanos.logger',
+  name: 'DiscardLogger',
```
kgrgreer: Don't we already have a DiscardLogger above?

deejes: This was a temporary arrangement because I was having a hard time with the script file.

src/foam/nanos/logger/DiscardLogger.js
deejes: Makes sense, will amend accordingly.

deejes: So there is another logger decorator in the pipeline, that will filter against regexs, and in the schema that was initially drawn out, that decorator was named FilteredLogger and the one implemented here DiscardLogger, hence the naming.
Flipped the boolean values.

src/foam/nanos/logger/DiscardLogger.js
```diff
@@ -0,0 +1,99 @@
...
+  extends: 'foam.nanos.logger.ProxyLogger',
+ 
+  properties: [
+    {
+      name: 'logError',
+      class: 'Boolean',
+      value: true
```
kgrgreer: Do you need to set the javaFactory: also?

src/foam/nanos/logger/DiscardLogger.js
kgrgreer: Remove DiscardLogger if it's now replaced with FilteredLogger.

src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -0,0 +1,229 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+
+foam.CLASS({
+  package: 'foam.nanos.logger',
+  name: 'DiscardLogger',
```
kgrgreer: Should be called RepeatLogger, not DiscardLogger.

src/foam/nanos/logger/DiscardLogger.js
deejes: No, don't. Have removed it.

src/services
```diff
@@ -4,7 +4,7 @@ p({"class":"foam.nanos.boot.NSpec", "name":"httprouter",                       "
...
-p({"class":"foam.nanos.boot.NSpec", "name":"logger",                           "lazy":true,  "serve":false, "serviceScript":"import foam.nanos.logger.CompositeLogger; import foam.nanos.logger.StdoutLogger; import foam.nanos.logger.FileLogger; CompositeLogger log = new CompositeLogger(); StdoutLogger stdOut = new StdoutLogger(); FileLogger fileOut = new FileLogger(); fileOut.start(); return log.add(stdOut).add(fileOut);"})
+p({"class":"foam.nanos.boot.NSpec", "name":"logger",                           "lazy":true,  "serve":false, "serviceScript":"import foam.nanos.logger.CompositeLogger; import foam.nanos.logger.StdoutLogger; import foam.nanos.logger.FileLogger; import foam.nanos.logger.ProxyLogger; import foam.nanos.logger.DiscardLogger; CompositeLogger log = new CompositeLogger(); StdoutLogger stdOut = new StdoutLogger(); FileLogger fileOut = new FileLogger(); fileOut.start(); DiscardLogger discard = new DiscardLogger(x); discard.setDelegate(fileOut);  return log.add(stdOut).add(discard);"})
```
kgrgreer: Rename DiscardLogger here to LogLevelFilterLogger.

tools/classes.js
```diff
@@ -150,7 +150,9 @@ var classes = [
   'foam.nanos.script.Language',
   'foam.nanos.auth.Language',
   'foam.nanos.auth.Region',
+  'foam.nanos.logger.DiscardLogger',
```
kgrgreer: Rename DiscardLogger here to LogLevelFilterLogger.

deejes: Did that, added RepeatLogger as well. 

The services file needs to be updated as well, its on my radar.

deejes:
services file has been updated, loggers working per expectation. The LastLogLevelDiscard logger sits before the RepeatLogger.

src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -0,0 +1,229 @@
...
+    }        
+    case ERROR: {
+      getDelegate().error("The last log was repeated " + getRepeatCount() + " times" );
+      break;
+    }
+  }     
+`
+    },
+    {
+      name: 'repeatLogConstructor',
```
kgrgreer: This is an odd name, since it isn't an actually a constructor. Perhaps a different name would be better.

src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -0,0 +1,224 @@
...
+        },
+      ],
+      javaReturns: 'void',
+      javaCode: `
+  switch ( LogLevel.values()[logLevelName.ordinal()] ) {
+    case DEBUG: {
+      setRepeatCount(1);
+      setLastUniqueObject(args);
+      getDelegate().debug(args);  
+      if (setLastLogLevel){
```
jlhughes: I'm not clear on setLastLogLevel and setLastLogLevel(enum)?

src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -0,0 +1,224 @@
...
+          javaType: 'Boolean'
+        },
+        {
+          name: 'args',
+          javaType: 'Object...'
+        },
+      ],
+      javaReturns: 'void',
+      javaCode: `
+  switch ( LogLevel.values()[logLevelName.ordinal()] ) {
```
jlhughes: Can this be reduced to 
setRepeatCount(1)
setLastUniqueObject(orgs);
switch ( .. ) {
}
if (setLastLogLevel) {
setLastLogLevel(logLevelName)
}

src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -0,0 +1,224 @@
...
+      break;
+    }
+  }
+`  
+    },
+    {
+      name: 'logLastLogRepeats',
+      javaReturns: 'void',
+      javaCode: 
+      `
```
jlhughes, line 88: create a String with '... last log repeated .. ', 
then call getDelegate().debug(message), etc

jlhughes (changes requested):

src/foam/nanos/logger/RepeatLogger.js
deejes: Yes it can be reduced in that way.

src/foam/nanos/logger/RepeatLogger.js
deejes: done.

src/foam/nanos/logger/RepeatLogger.js
deejes: One is a boolean value that decides whether the lastLogLevel should be updated, the other is the method that does it. Changed the former to updateLastLogLevel.

deejes:
@googlebot I signed it

deejes:
I signed it

src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -0,0 +1,219 @@
...
+  String logCountMessage = ("The last log was repeated " + getRepeatCount() + " times" );
```
kgrgreer: parenthesis not needed

PR #1156, opened 2018-05-08
Add .freeze() to java FObjects

adamvy:
Add a test to demonstrate functionality.


mcarcaso (approved):
LGTM. Will we ever need to unfreeze an object you think? Or serialize the fact that an object is frozen?

adamvy:
PTAL added freezing to MDAO and MapDAO.

Object unfreeze when you .fclone() them.  I don't think we need to serialize the frozen state of an object.

mcarcaso (approved):

PR #1167, opened 2018-05-10
Fixed issue with diff in AbstractFObjectPropertyInfo

kirkeaton:
1. Resolves issue where changes weren't being persisted when previous value was null and new value was being set
2. Removed empty AbstractFilePropertyInfo class


src/foam/core/AbstractFObjectPropertyInfo.java
```diff
@@ -105,7 +106,7 @@ public boolean hardDiff(FObject o1, FObject o2, FObject diff) {
      *  X.get("DAO").put(obj);
      * In this case: the before value inside the model is loss, so can not find difference
      */
-    if ( (this.get(o1) == this.get(o2)) ) {
+    if ( SafetyUtil.compare(o1, o2) == 0 ) {
```
adamvy, line 109: If the object is the same shouldn't we not want to set it on the diff since there is no difference?

adamvy: Also SafetyUtil.compare() is pretty different than == since .compare does a deep comparison between the object rather than just checking if they're the same instance like the comment indicates.

adamvy: Can we fix the grammar of the comment while we're here?

kgrgreer: == wasn't correct because it wouldn't compare Strings correctly

kgrgreer: This should be SafetyUtil.compare(this.get(o1), this.get(o2))  == 0 or SafetyUtil.equals(this.get(o1), this.get(o2)) or just compare(o1, o2), since compare is a method of PropertyInfo and the implementation is probably something like:
      public int compare(Object o1, Object o2) {
        return foam.util.SafetyUtil.compare(get_(o1), get_(o2));
      }

adamvy: @kgrgreer This is in FObjectPropertyInfo, the value of an FObjectProperty should never be a String it should be an FObject or null.

The comment above suggests that this is for the case where the property points to the same instance, there's no point during a deep comparison here because just after this conditional we use a recursive hardDiff on the values.

src/foam/core/AbstractFObjectPropertyInfo.java
```diff
@@ -105,7 +105,7 @@ public boolean hardDiff(FObject o1, FObject o2, FObject diff) {
      *  X.get("DAO").put(obj);
      * In this case: the before value inside the model is loss, so can not find difference
      */
-    if ( (this.get(o1) == this.get(o2)) ) {
+    if ( ( this.get(o1) == null && this.get(o2) != null ) || ( this.get(o1) == this.get(o2) ) ) {
```
kgrgreer: I've changed this to the following code which is more general because it also checks the case where o1 != null and o2 does. Have merged and will close this PR.

    Object fo1 = this.get(o1), fo2 = this.get(o2);
    if ( fo1 == fo2 || ( ( fo1 == null ) != ( fo2 == null ) ) ) {

kgrgreer:
Fixed in another PR.

PR #1177, opened 2018-05-24
Modified Java Blob Store to make use of Storage class

kirkeaton:
Resolves https://github.com/nanoPayinc/NANOPAY/issues/1992

1. Modified Java Blob Store to make use of Storage class
2. Removed root property from Java BlobStore

### DIFF
--- a/src/foam/blob/Blob.js
+++ b/src/foam/blob/Blob.js
@@ -477,7 +477,8 @@ foam.CLASS({
   properties: [
     {
       class: 'String',
-      name: 'root'
+      name: 'root',
+      generateJava: false
     },
     {
       class: 'String',
--- a/src/foam/blob/Blob.js
+++ b/src/foam/blob/Blob.js
@@ -485,7 +485,7 @@ foam.CLASS({
       class: 'String',
       name: 'tmp',
       transient: true,
-      documentation: 'Temp directory of where files are stored before hashing'
+      documentation: 'Temp directory of where files are stored before hashing',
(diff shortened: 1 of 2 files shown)


src/foam/blob/BlobJava.js
```diff
@@ -190,21 +190,21 @@ foam.CLASS({
...
-    'java.io.FileOutputStream'
+    'java.io.FileOutputStream',
+    'foam.nanos.fs.Storage'
   ],
 
   constants: [
     {
-      name: 'BUFFER_SIZE',
       value: 4096,
+      name: 'BUFFER_SIZE',
```
kgrgreer: Why switch the order? I think it's more conventional to have the name before the value.

jlhughes (approved):

src/foam/blob/Blob.js
```diff
@@ -477,7 +477,8 @@ foam.CLASS({
   properties: [
     {
       class: 'String',
-      name: 'root'
+      name: 'root',
```
kgrgreer, line 480: Is this property still used? Can you document what it's for?

src/foam/blob/BlobJava.js
kirkeaton: i don't recall changing this

src/foam/blob/Blob.js
kirkeaton: Yes it is, yes I can

jlhughes (approved):

PR #1178, opened 2018-05-24
Using provided path from parameters (when given) for nanos log

daudich:
Instead of `user.dir`.

### DIFF
--- a/src/foam/nanos/logger/FileLogger.java
+++ b/src/foam/nanos/logger/FileLogger.java
@@ -24,7 +24,15 @@ public class FileLogger
     logger.setLevel(Level.ALL);
 
     try {
-      Handler handler = new FileHandler("nano.log");
+      String logPath = System.getProperty("LOG_HOME");
+      Handler handler = null;
+
+      if ( logPath == null ) {
+        handler = new FileHandler("nano.log");
+      } else {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/logger/FileLogger.java
```diff
@@ -24,7 +24,15 @@ public void start() {
     logger.setLevel(Level.ALL);
 
     try {
-      Handler handler = new FileHandler("nano.log");
+      String logPath = System.getProperty("LOG_HOME");
```
kgrgreer, line 27: Don't use System.getProperty() this way because it is a global variable that can't be set on a context-by-context basis. You can make it be a property and then have this be the default if not set, but the way the code is you couldn't have two FileLoggers logging to different directories.

jlhughes (approved):

src/foam/nanos/logger/FileLogger.java
kgrgreer: Better to Model FileLogger so it can be configured through the GUI.

jlhughes: so we'll push this through today and create another ticket to model FileLogger with property 'path' which has a factory defaulting to System.getProperty("LOG_HOME")

PR #1182, opened 2018-05-24
Add initRelationship method to foam.dao.Relationship

mcarcaso:
Give relationships an initRelationship method and replace the init method with that so relationships don't immediately initialize because you might not want it to get initialized when querying relationships from a dao.

### DIFF
--- a/src/foam/apploader/ClassLoader.js
+++ b/src/foam/apploader/ClassLoader.js
@@ -23,7 +23,8 @@ foam.CLASS({
 have multiple classloaders running alongside eachother`
 ],*/
   requires: [
-    'foam.classloader.OrDAO'
+    'foam.classloader.OrDAO',
+    'foam.dao.Relationship',
   ],
   properties: [
     {
@@ -134,7 +135,10 @@ have multiple classloaders running alongside eachother`
           if ( foam.lookup(id, true) ) return Promise.resolve(foam.lookup(id));
 
           return this.pending[id] = this.modelDAO.find(id).then(function(m) {
-            if ( ! m ) return Promise.reject(new Error('Class Not Found: ' + id));
+            if ( ! m ) return Promise.reject(new Error('Model Not Found: ' + id));
+            if ( self.Relationship.isInstance(m) ) {
+              return m.initRelationship();
+            }
 
             return this.buildClass_(m, path);
           }.bind(this), function() {
--- a/src/foam/apploader/ModelFileDAO.js
+++ b/src/foam/apploader/ModelFileDAO.js
@@ -38,10 +38,11 @@ foam.CLASS({
 
             if ( jsonId !== id ) {
               self.cache[jsonId] = m;
-              return jsonId;
+            } else {
+              json = m;
             }
 
-            json = m;
+            return jsonId;
           };
 
           context.foam.INTERFACE = function(json) {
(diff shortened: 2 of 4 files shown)


src/foam/apploader/ClassLoader.js
```diff
@@ -134,7 +135,10 @@ have multiple classloaders running alongside eachother`
           if ( foam.lookup(id, true) ) return Promise.resolve(foam.lookup(id));
 
           return this.pending[id] = this.modelDAO.find(id).then(function(m) {
-            if ( ! m ) return Promise.reject(new Error('Class Not Found: ' + id));
+            if ( ! m ) return Promise.reject(new Error('Model Not Found: ' + id));
+            if ( self.Relationship.isInstance(m) ) {
```
mcarcaso, line 139: This is to support using the classloader to load relationships. Should we support this or should I remove these changes? I was using this in another branch but I could go without it if this doesn't make sense.

src/foam/apploader/ModelFileDAO.js
```diff
@@ -56,21 +56,26 @@ foam.CLASS({
...
           context.foam.RELATIONSHIP = function(r) {
-            var references = foam.json.references(x, r);
-
-            Promise.all(references.concat([
-              foam.package.waitForClass(r.sourceModel),
-              foam.package.waitForClass(r.targetModel)
-            ])).then(function() {
-              var obj = foam.dao.Relationship.create(r, x);
-
+            relationship = Promise.all(foam.json.references(x, r)).then(function() {
```
adamvy: Can we clean this up further by making context.foam.RELATIONSHIP match context.foam.ENUM and context.foam.INTERFACE in behaviour?

We might need to be careful about how we compute the id of the json, and maybe special case the logic of self.cache so that we don't cache a relationship but init it eagerly like you do here.

src/foam/apploader/ModelFileDAO.js
```diff
@@ -56,21 +56,26 @@ foam.CLASS({
...
+               var obj = foam.dao.Relationship.create(r, x);
               obj.validate && obj.validate();
-              foam.package.registerClass(obj);
```
adamvy: You appear to have removed all of the foam.package.registerClass calls.  Do relationships get registered at all now in the global context?  There may be code that still references them that way, so you may want to still call registerClass in initRelationship()

mcarcaso: Yeah this is something I wasn't quite sure about when I saw it so I took it out to raise the discussion. An instance of a relationship isn't exactly a class so calling registerClass didn't feel right but I can definitely put it back. Should relationships get registered as classes or should they be put into the context in some other way?

src/foam/apploader/ModelFileDAO.js
mcarcaso: Yeah relationships are a little trickier than ENUM and INTERFACE. The sourceModel and targetModel are needed for computing the ID so I'll need to wait for those before doing anything here. context.foam.CLASS pulls the package and name out of the json, which typically isn't set on a relationship's json so I'd have to have those set before handing the json over to context.foam.CLASS. The result would be something like:
```
foam.RELATIONSHIP = function(json) {
  json.class = json.class || 'foam.dao.Relationship'
  json.package = json.package || json.sourceModel.substring(0, json.sourceModel.lastIndexOf('.'))
  json.name = json.name || json.sourceModel.substring(json.sourceModel.lastIndexOf('.')) + json.targetModel.substring(json.targetModel.lastIndexOf('.')) + 'Relationship';
  ...
```
Which copies a bunch of logic from foam.dao.Relationship (though that logic looks up the classes which might not exist yet).

Feels a little awkward but I can make it work if you think this is the way to go.

src/foam/apploader/ClassLoader.js
mcarcaso: Going to leave this as supporting loading relationships from the classloader seems like a useful feature.

mcarcaso (commented):
Made a bunch of updates and I think this is good to commit now.

PR #1186, opened 2018-05-28
Session Expiry

kirkeaton:
Added isExpired method to Session model. Added checking of expiry to SessionServerBox

Fixes: https://github.com/nanoPayinc/NANOPAY/issues/2064


kgrgreer:
The cron job runs every minute to remove expired sessions, so this isn't necessary.

PR #1195, opened 2018-05-29
Modified password hashing

kirkeaton:
1. Created a property called "desiredPassword" which is never stored but when set is hashed into the Password used for authentication. 
2. Made password and previousPassword networkTransient.
3. Removed Password.sanitize function
4. Removed PasswordRemovalSink
5. Renamed PasswordHashingDAO to UserPasswordHashingDAO


src/foam/nanos/auth/User.js
```diff
@@ -203,9 +203,11 @@ foam.CLASS({
...
     {
       class: 'Password',
-      name: 'password',
+      name: 'desiredPassword',
+      label: 'Password',
       displayWidth: 30,
       width: 100,
+      storageTransient: true,
       validateObj: function (password) {
         var re = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9]).{7,32}$/;
```
jlhughes, line 212: Not a blocking comment. Curious if during password entry (or after failure) we tell them what the requirements are.

jlhughes (approved):

src/foam/nanos/auth/User.js
kirkeaton: Not sure if we do or not

PR #1197, opened 2018-05-30
default script status set to Unscheduled

shreymahendru:
setting the default value of the script status to unscheduled.

### DIFF
--- a/src/foam/nanos/script/Script.js
+++ b/src/foam/nanos/script/Script.js
@@ -30,7 +30,8 @@ foam.CLASS({
     'java.io.PrintStream',
     'java.util.Date',
     'java.util.List',
-    'static foam.mlang.MLang.*'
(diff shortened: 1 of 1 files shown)


src/foam/nanos/script/Script.js
```diff
@@ -30,7 +30,8 @@ foam.CLASS({
     'java.io.PrintStream',
     'java.util.Date',
     'java.util.List',
-    'static foam.mlang.MLang.*'
+    'static foam.mlang.MLang.*',
+    'foam.nanos.script.ScriptStatus'
```
kgrgreer: Is this needed if they are in the same package?

shreymahendru: It works desirably without it, but I saw "cannot resolve symbol 'ScriptStatus'" error in the generated java file, so I wanted to be explicit and actually import the class.

PR #1200, opened 2018-05-30
Update history view

TW80000:
Update the history view to use `Element.select()` so that it updates when the underlying DAO changes.

### DIFF
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -1546,6 +1546,17 @@ foam.CLASS({
       return slot;
     },
 
+    /**
+     * Given a DAO and a function that maps from a record in that DAO to an
+     * Element, call the function with each record as an argument and add the
+     * returned elements to the view.
+     * Will update the view whenever the contents of the DAO change.
+     * @param {DAO<T>} dao The DAO to use as a data source
+     * @param {T -> Element} f A function to be called on each record in the DAO. Should
+     * return an Element that represents the view of the record passed to it.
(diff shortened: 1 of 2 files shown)


src/foam/u2/history/HistoryView.js
```diff
@@ -8,7 +8,10 @@ foam.CLASS({
   package: 'foam.u2.history',
   name: 'HistoryView',
   extends: 'foam.u2.View',
-  requires: [ 'foam.u2.history.HistoryItemView' ],
+  requires: [
+    'foam.u2.history.HistoryItemView',
```
kgrgreer: sort requires

src/foam/u2/Element.js
```diff
@@ -1546,6 +1546,17 @@ foam.CLASS({
       return slot;
     },
 
+    /**
```
kgrgreer, line 1549: Great!

src/foam/u2/history/HistoryView.js
```diff
@@ -68,24 +73,15 @@ foam.CLASS({
...
+          return this.E().start('div')
+            .addClass('timelineRecord')
+            .start('div').addClass('timeline').end()
+            .call(function() {
+              view.historyItemView.outputRecord(this, record)
             })
-          })
-        });
+          .end();
+        }, false);
```
kgrgreer: Can remove: ", false"

TW80000:
I signed it!

PR #1212, opened 2018-06-04
b/d/fixing persist bug

xuerongCode:
add outputFObjectDelta in the JSON outputter.
add Nullparser to the FObjectParser, FObjectArrayParser, DateParser, StringArrayParser.
fix bug on loading diff entries.


src/foam/dao/AbstractJDAO.java
```diff
@@ -283,49 +275,31 @@ public void removeAll_(final X x, long skip, final long limit, Comparator order,
...
-    // get the PropertyInfo for the id
-    PropertyInfo idInfo = (PropertyInfo) getOf().getAxiomByName("id");
-    // set id property to new instance
-    idInfo.set(diff, idInfo.get(o));
-    return diff;
-  }
-
-  protected FObject mergeChange(FObject o, FObject c) {
-    //if no change to merge, return FObject;
+  protected FObject mergeFObject(FObject o, FObject c) {
```
kgrgreer: How about more meaningful argument names than 'o' and 'c', and maybe a comment to explain what this method does.

src/foam/dao/AbstractJDAO.java
```diff
@@ -283,49 +275,31 @@ public void removeAll_(final X x, long skip, final long limit, Comparator order,
...
-      }
+      mergeProperty(o, c, prop);
     }
-
     return o;
   }
 
+  protected void mergeProperty(FObject o, FObject c, PropertyInfo prop) {
+    if ( ! prop.isSet(c) ) return;
+    if ( prop instanceof AbstractFObjectPropertyInfo ) {
```
kgrgreer: Is this block needed? We don't merge sub-objects do we? Don't we just send the whole object if it changed?

src/foam/lib/json/ObjectNullParser.java
```diff
@@ -0,0 +1,23 @@
...
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.lib.json;
+
+import foam.lib.parse.*;
+
+public abstract class ObjectNullParser
```
kgrgreer: What is this for? Document.

src/foam/lib/json/Outputter.java
```diff
@@ -84,6 +85,17 @@ public String stringify(FObject obj) {
     return this.toString();
   }
 
+  public String stringifyDiff(FObject o, FObject n) {
+    if ( stringWriter_ == null ) {
+      stringWriter_ = new StringWriter();
+      writer_ = new PrintWriter(stringWriter_);
+    }
```
kgrgreer: This block should be moved to a helper method since it is duplicated elsewhere.

src/foam/lib/json/Outputter.java
```diff
@@ -232,11 +244,90 @@ protected Boolean maybeOutputProperty(FObject fo, PropertyInfo prop, boolean inc
...
+
     if ( includeComma ) writer_.append(",");
     outputProperty(fo, prop);
     return true;
   }
 
+  protected void outputFObjectDelta(FObject o, FObject n, boolean outerObject) {
+    ClassInfo info = o.getClassInfo();
+    boolean outputComma = true;
+    if ( ! o.equals(n) ) {
```
kgrgreer: Transient fields could be the only fields that are different which would cause this to output.

src/foam/lib/json/StringArrayParser.java
```diff
@@ -12,22 +12,28 @@
...
-    super(new Seq1(3,
-      new Whitespace(),
-      new Literal("["),
-      new Whitespace(),
-      new Repeat(
-        new StringParser(),
-        new Seq0(new Whitespace(), new Literal(","), new Whitespace())),
-      new Whitespace(),
-      new Literal("]")));
+    super(
```
kgrgreer, line 15: Why is this needed?

xuerongCode: if the StringArrayParser can not parse null, the change will not be merged to memory, when system reboot

src/foam/dao/AbstractJDAO.java
```diff
@@ -219,26 +217,22 @@ public synchronized FObject put_(X x, FObject obj) {
...
     try {
       // TODO: supress class name from output
-      writeComment((User) x.get("user"));
-      out_.write("p(");
-      // TODO: output string directly here rather than converting to 'record'
-      // String above.
-      out_.write(record);
-      out_.write(")");
-      out_.newLine();
+      if ( ! "".equals(record.trim())) {
```
kgrgreer: Why is this needed? Why is the trim needed?

src/foam/dao/AbstractJDAO.java
```diff
@@ -283,47 +277,24 @@ public void removeAll_(final X x, long skip, final long limit, Comparator order,
...
-    if ( c == null ) return o;
-    //merge change
-    return maybeMerge(o, c);
-  }
-
-  protected FObject maybeMerge(FObject o, FObject c) {
-    if ( o == null ) return o = c;
+  //add diff property to old property
+  protected FObject mergeFObject(FObject oldFObject, FObject diffFObject) {
+    if ( diffFObject == null ) return oldFObject;
```
kgrgreer: When is diffFObject null? If it's null couldn't I just not call mergeFObject() and not call put()?

src/foam/dao/AbstractJDAO.java
```diff
@@ -283,47 +277,24 @@ public void removeAll_(final X x, long skip, final long limit, Comparator order,
...
 
     //get PropertyInfos
-    List list = o.getClassInfo().getAxiomsByClass(PropertyInfo.class);
+    List list = oldFObject.getClassInfo().getAxiomsByClass(PropertyInfo.class);
```
kgrgreer: Why is this called oldFObject when it's a List of Properties?

src/foam/lib/json/Outputter.java
```diff
@@ -74,14 +75,23 @@ public Outputter(PrintWriter writer, OutputterMode mode) {
   }
 
   public String stringify(FObject obj) {
+    initWriter();
+    outputFObject(obj);
+    return this.toString();
+  }
+
+  public String stringifyDiff(FObject oldFObject, FObject newFObject) {
```
kgrgreer: Rename stringifyDelta to be consistent with outputFObjectDelta

PR #1219, opened 2018-06-05
Fix cannot choose the same profile image again after removing the existing one 

Henry0422:
For https://github.com/nanoPayinc/NANOPAY/issues/2236, I believe we need to set the `this.document.querySelector('.attachment-input').value ` to null if we remove the profile images.


src/foam/nanos/auth/ProfilePictureView.js
```diff
@@ -200,6 +200,7 @@ foam.CLASS({
     function onRemoveClicked (e) {
       this.dragActive = false;
       this.ProfilePictureImage= null;
+      this.document.querySelector('.attachment-input').value = null;
```
kgrgreer: What if there's more than one attachment-input on the screen at once? I'll show you a better way to do this.

Henry0422: What would be the better solution for this issue? Thanks.

PR #1228, opened 2018-06-07
Implemented Journal interface

kirkeaton:
1. Implemented the Journal interface which extends the Sink interface and contains one method: replay
2. Created FileJournal implementation that reads data from a file
3. Modified AbstractJDAO to use FileJournal


kirkeaton:
Too many changes to AbstractJDAO and JSON Outputter in the time that I started this. Not comfortable with merging.

src/foam/dao/JDAO.java
```diff
@@ -7,24 +7,117 @@
...
+   *
+   * @param user user who modified entry
+   */
+  protected void writeComment(User user) {
+    StringBuilder builder = sb.get()
+      .append("// Modified by ")
+      .append(user.getFirstName());
+
+    // append last name if present
+    if (!SafetyUtil.isEmpty(user.getLastName())) {
```
kgrgreer: if ( ! SafetyUtil.isEmpty(user.getLastName()) ) {

kgrgreer: User has a label() method that we could use for this.

src/foam/dao/JDAO.java
```diff
@@ -7,24 +7,117 @@
...
+      builder.append(" ")
+        .append(user.getLastName());
+    }
+
+    builder.append(" (")
+      .append(String.valueOf(user.getId()))
```
kgrgreer: We don't need the String.valueOf() since StringBuilder has append() methods for primitive types.

src/foam/dao/JDAO.java
```diff
@@ -7,24 +7,117 @@
...
+  public FObject remove_(X x, FObject obj) {
+    writeComment((User) x.get("user"));
+    journal_.remove(obj, null);
+    return getDelegate().remove_(x, obj);
+  }
+
+  @Override
+  public void removeAll_(final X x, long skip, final long limit, Comparator order, Predicate predicate) {
+    getDelegate().select_(x, new RemoveSink(x, this), skip, limit, order, predicate);
+    getDelegate().removeAll_(x, skip, limit, order, predicate);
```
adamvy: This doesn't look right, select_ into a RemoveSink and then delegate.removeAll_()?  Only one of these should be required.

src/foam/dao/JDAO.java
```diff
@@ -7,24 +7,117 @@
...
+        new FileJournal.Builder(getX())
+          .setFilename(filename)
+          .build()
+      })
+      .build().replay(delegate);
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    writeComment((User) x.get("user"));
```
mcarcaso: Could this writeComment stuff be a journal decorator too?

src/foam/dao/JDAO.java
```diff
@@ -7,24 +7,117 @@
...
-  public JDAO(foam.core.X x, ClassInfo classInfo, String filename) {
+  protected FileJournal journal_;
+
+  public JDAO(X x, ClassInfo classInfo, String filename) {
     this(x, new MapDAO(classInfo), filename);
   }
 
-  public JDAO(foam.core.X x, DAO delegate, String filename) {
-    super(x, delegate, filename);
+  public JDAO(X x, DAO delegate, String filename) {
```
mcarcaso: The constructor for this should just take a journal. These convenience constructors should be removed and done using the facade pattern.

src/foam/dao/JDAO.java
```diff
@@ -7,24 +7,117 @@
...
+import foam.util.SafetyUtil;
+
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.TimeZone;
 
 public class JDAO
-  extends AbstractJDAO
+  extends ProxyDAO
```
mcarcaso: Why are we not modelling this in JS and generating this class rather than writing it by hand?

src/foam/dao/JDAO.java
kirkeaton: I would have liked it to have been but the Sink interface doesn’t provide the context anywhere for me to get the current user.

src/foam/dao/JDAO.java
kirkeaton: This was code that was inherited from the previous implementation of the JDAO. I’ll investigate this further.

src/foam/dao/JDAO.java
kirkeaton: We should be. I’ll make it so.

mcarcaso:
Should we add context oriented methods to sink to solve this issue?

mdittmer:
@mcarcaso sinks can be deserialized from remotes. Is there a security risk associated with trusting a Sink or its context to faithfully report info like who the user is?

mcarcaso:
Something to keep in mind for sure but the context on the server at this
point should be trustworthy. The server should construct the context as the
message it’s sent passes through the various boxes on its way to this sink.

src/foam/dao/JDAO.java
kirkeaton: Replaced first name & last name with the label method.

src/foam/dao/JDAO.java
kirkeaton: Fixed

src/foam/dao/JDAO.java
kirkeaton: Only the first line is necessary here. The RemoveSink will create the remove entry in the journal and also remove the object from the delegate.

src/foam/dao/JDAO.java
kirkeaton: Finished modelling JDAO

src/foam/dao/JDAO.java
kirkeaton: Once it's modelled it will have all the constructors and the builders needed. I'm going to leave the convenience constructors since a lot of our services our defined using this constructor.

adamvy:
@mdittmer We don't serialize contexts, so even though the client can provide the sink, they don't provide the context of that sink, that is created on the server when processing the request.

adamvy:
LGTM

PR #1231, opened 2018-06-07
compare map property

xuerongCode:
fix castException on compare Map

### DIFF
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -932,6 +932,45 @@ foam.CLASS({
     ['javaJSONParser', 'new foam.lib.json.MapParser()'],
     ['javaInfoType', 'foam.core.AbstractObjectPropertyInfo'],
     ['javaFactory', 'return new java.util.HashMap();']
+  ],
+
+  methods: [
+    function createJavaPropertyInfo_(cls) {
+      var info = this.SUPER(cls);
+      var compare = info.getMethod('compare');
+      compare.body = this.compareTemplate();
+      return info;
+    }
+  ],
+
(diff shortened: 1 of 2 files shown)


src/foam/java/refinements.js
```diff
@@ -932,6 +932,45 @@ foam.CLASS({
     ['javaJSONParser', 'new foam.lib.json.MapParser()'],
     ['javaInfoType', 'foam.core.AbstractObjectPropertyInfo'],
     ['javaFactory', 'return new java.util.HashMap();']
+  ],
+
+  methods: [
```
kgrgreer: Wouldn't it be less code to create AbstractMapPropertyInfo which extends AbstractObjectPropertyInfo and just has this method defined once so that we don't have multiple generated copies?

src/foam/java/refinements.js
```diff
@@ -930,8 +930,17 @@ foam.CLASS({
   properties: [
     ['javaType', 'java.util.Map'],
     ['javaJSONParser', 'new foam.lib.json.MapParser()'],
-    ['javaInfoType', 'foam.core.AbstractObjectPropertyInfo'],
+    ['javaInfoType', 'foam.core.AbstractMapPropertyInfo'],
     ['javaFactory', 'return new java.util.HashMap();']
+  ],
+
+  methods: [
```
kgrgreer, line 937: Could we remove this whole block and in properties: above just do:

 [ 'compare', 'return super.compare(o1, o2);' ],

?

xuerongCode: did not work

PR #1232, opened 2018-06-07
fix check if Id is set in put_  method SequenceNumberDAO

smirnovaka:
@kgrgreer


adamvy:
This makes 0 no longer a valid id, is that really the behaviour we want?  What was the issue this commit was attempting to fix?

PR #1234, opened 2018-06-07
Output the property name when a Reference 'of' is invalid/null.

jlhughes:
This helps troubleshooting during genjava.sh when an 'of' path is invalid.

### DIFF
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -491,7 +491,12 @@ foam.CLASS({
     {
       class: 'String',
       name: 'targetDAOKey',
-      expression: function(of) { return foam.String.daoize(of.name); }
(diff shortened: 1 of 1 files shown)


src/foam/core/types.js
```diff
@@ -491,7 +491,12 @@ foam.CLASS({
     {
       class: 'String',
       name: 'targetDAOKey',
-      expression: function(of) { return foam.String.daoize(of.name); }
+      expression: function(of) {
+        if ( ! of ) {
+          console.log('ERROR: invalid \'of\' for property with targetDAOKey', this.name);
```
kgrgreer: Wouldn't console.error() be better? If you use " instead of ' then you won't need to escape the 's.

PR #1242, opened 2018-06-11
Escape Commas

ZacZhang:
https://github.com/nanoPayinc/NANOPAY/issues/2174

### DIFF
--- a/src/foam/lib/csv/Outputter.java
+++ b/src/foam/lib/csv/Outputter.java
@@ -144,7 +144,13 @@ public class Outputter
     //   remove commas:   , ->
     //   quote strings:   Acme, Inc. -> "Acme, Inc."
 
-    return s.replace("\n","\\n").replace("\"", "\\\"");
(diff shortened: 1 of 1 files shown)


src/foam/lib/csv/Outputter.java
```diff
@@ -144,7 +144,13 @@ public String escape(String s) {
     //   remove commas:   , ->
     //   quote strings:   Acme, Inc. -> "Acme, Inc."
 
-    return s.replace("\n","\\n").replace("\"", "\\\"");
+    s = s.replace("\n","\\n").replace("\"", "\\\"");
+
+    // quote strings:   Acme, Inc. -> "Acme, Inc."
+    if ( s.contains(",") )
+      return "\"" + s + "\"";
```
jlhughes, line 151: Do these " need to be escaped?  Perhaps this should go above the \\\" escaping.

jlhughes (changes requested):
See code comments.

jlhughes (approved):
Order is correct.

PR #1243, opened 2018-06-11
Fixed bug with DateParser

kirkeaton:
There were some precision issues in parsing milliseconds which this PR aims to fix. Also there were some issues with parsing milliseconds that were less than 3 digits.

### DIFF
--- a/src/foam/lib/json/DateParser.java
+++ b/src/foam/lib/json/DateParser.java
@@ -56,16 +56,16 @@ public class DateParser
   public PStream parse(PStream ps, ParserContext x) {
     ps = super.parse(ps, x);
 
-    if (ps == null) {
+    if ( ps == null ) {
       return null;
     }
 
-    if (ps.value() == null) {
+    if ( ps.value() == null ) {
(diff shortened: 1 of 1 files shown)


src/foam/lib/json/DateParser.java
```diff
@@ -7,72 +7,102 @@
...
+        new Repeat(new Chars("0123456789")),
+        new Literal("Z"),
+        new Literal("\""))));
   }
 
   public PStream parse(PStream ps, ParserContext x) {
     ps = super.parse(ps, x);
 
-    if ( ps == null ) return null;
+    if (ps == null) {
```
kgrgreer: space after ( and before )
if ( ps == null )

src/foam/lib/json/DateParser.java
```diff
@@ -7,72 +7,102 @@
...
+      return null;
+    }
 
-    if ( ps.value() == null ) 
+    if (ps.value() == null) {
```
kgrgreer: space after ( and before )

PR #1249, opened 2018-06-13
Make Enums FObjects in swift instead of enums

mcarcaso:
This allows us to remove some special casing on enums and use the regular serialization and deserialization that we get for free with fobjects.


src/foam/swift/refines/FObject.js
```diff
@@ -42,25 +42,12 @@ foam.LIB({
...
-          .filter(function(p) {
-            return !!p.getSwiftSupport(this);
-          }.bind(this));
-      var actions = this.getOwnAxiomsByClass(foam.core.Action).filter(axiomFilter)
-          .filter(function(p) {
-            return !this.getSuperAxiomByName(p.name);
-          }.bind(this));
-
-      var multiton = this.getOwnAxiomsByClass(foam.pattern.Multiton);
+      var multiton = this.getAxiomsByClass(foam.pattern.Multiton);
```
kgrgreer, line 45: Is there anyway to move this logic to the Multiton so that it's self contained?

src/foam/core/Enum.js
```diff
@@ -268,6 +268,7 @@ foam.CLASS({
     {
       class: 'String',
       name: 'documentation',
+      transient: true,
       adapt: function(_, d) {
```
kgrgreer: Why is this adapt needed? Doesn't String already do this?

src/foam/swift/refines/AbstractEnum.js
```diff
@@ -6,95 +6,77 @@
...
+                }
+              `,
+              returnType: self.model_.swiftName,
+            }));
+          }
+          return cls
         };
-      }
-    }
+      },
```
kgrgreer: No trailing commas.

src/foam/core/Enum.js
mcarcaso: Good question. Looks like this existed before adding class: 'String' so I guess it's no longer necessary. I'll delete it.

src/foam/swift/refines/AbstractEnum.js
mcarcaso: Removed here and in another spot or two.

src/foam/swift/refines/FObject.js
mcarcaso: It definitely should be. I have plans to refactor a lot of this stuff and
that's one thing I already have in mind to do.

PR #1250, opened 2018-06-13
Serialize and deserialize enums like FObjects rather than just using ordinal

mcarcaso:
This cuts down on the ambiguity that happens when using enums in the Constant expression.

### DIFF
--- a/src/foam/core/AbstractEnumPropertyInfo.java
+++ b/src/foam/core/AbstractEnumPropertyInfo.java
@@ -36,7 +36,6 @@ public abstract class AbstractEnumPropertyInfo
 
   public abstract int getOrdinal(Object o);
   public abstract java.lang.Enum forOrdinal(int ordinal);
-  public abstract void toJSON(foam.lib.json.Outputter outputter, Object value);
   public abstract void toCSV(foam.lib.csv.Outputter outputter, Object value);
 
   @Override
--- a/src/foam/core/Enum.js
+++ b/src/foam/core/Enum.js
@@ -268,6 +268,7 @@ foam.CLASS({
     {
       class: 'String',
       name: 'documentation',
+      transient: true,
       adapt: function(_, d) {
         return typeof d === 'function' ? foam.String.multiline(d).trim() : d;
       }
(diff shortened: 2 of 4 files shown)


mdittmer:
IIRC using the ordinal was deliberate for when serializing a large
collection where the object contains one or more enum properties. I'm not
sure that switching to a more inflated representation is the best move
overall.

mcarcaso:
The ambiguity that happens when we just send the ordinal is a problem that we can't get around without adding a bunch of special casing. If we want to cut down on the size of a payload that contains a bunch of repeated text, we should be able to compress the payload in a more generic way like using boxes.

adamvy:
Can you fix the test in test/src/core/Enum.js that's failing due to this change.

Also does this bring all our implementations in sync now with how they serialize/deserialize enums?

mcarcaso:
Test fixed. I also made some fields transient while I was at it.

This should make everything work for JS and Java but not swift. I have a separate PR for swift.

mdittmer:
Can you give an example of where this ambiguity is a problem? Usually
constant enum values look like SomeOp{arg1: SomeEnumProperty, arg2:
Constant(<ordinal value>)}. It's pretty clear how that case should be
interpreted.

adamvy:
I agree in your case it's not ambiguous but that ties the deserialization of arg2 to the current value of arg1.  Now we have to be careful to do it in the right order, or what if someone has the reverse of EQ(Status.OPEN, Issue.STATUS), that's still a valid expression but now we have to deserialize arg2 before arg1.

I guess it could be made to work by just keeping it as an integer and implementing Enum.compareTo(int) to work as expected, but that means information can be lost after serialization.  I'd be much more comfortable if deserialize(serialize(a)) == a rather than if deserialize(serialize(a)) ~= a.

Better to be correct than fast.  If size becomes a concern someone should implement a serialization layer that shortens property and class names, maybe even consider a binary encoding.  That would be more much effective and have global benefits rather than micro-optimizing enums.

adamvy (approved):

mcarcaso:
Going to hold off on this for now. Once I have a more concrete plan in place to cut down on the size of our payloads or journals, we'll go forward with this.

PR #1254, opened 2018-06-18
changed links to mintchip in AppConfig

waseem-chamaa:
Remove AppLink and PlayLink from foam and add them via refinement from application if required.


jlhughes:
not required, refining in nanopay

PR #1259, opened 2018-06-19
Create PublicUserInfoDAO

TW80000:
This commit introduces a new generic DAO decorator that decorates the
find and select methods of a DAO. It takes two property names, one that
holds a user id and one that will hold a subset of the properties of
said user. It looks up the user associated with the id and sets the
other property to an object with a few public properties of the user
that was looked up.

See the documentation in the code for example usage and more details.

### DIFF
--- a/src/foam/dao/PublicUserInfoDAO.java
+++ b/src/foam/dao/PublicUserInfoDAO.java
@@ -1,0 +1,112 @@
+package foam.dao;
+
+import foam.dao.*;
+import foam.core.FObject;
+import foam.core.PropertyInfo;
+import foam.core.X;
+import foam.dao.DAO;
+import foam.dao.ProxyDAO;
+import foam.dao.Sink;
+import foam.nanos.auth.PublicUserEntity;
+import foam.nanos.auth.User;
+import net.nanopay.tx.model.Transaction;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+import foam.nanos.logger.Logger;
+
+/**
+ * Populate a model returned from find or select with a small subset of public 
+ * user information.
+ *
+ * Example: A call like
+ *
+ *   new foam.dao.PublicUserInfoDAO(x, "payerId", "payer", ...)
+ *
+ * will look for a property named "payerId" on the object being selected and
+ * look up the user associated with that id. It will then set the given property
+ * ("payer") to the object being returned that contains some public properties
+ * of the associated user.
+ *
+ * Requires both properties to already be defined on the model. For example, on
+ * Transaction.js:
+ *
+ *     ...
+ *   },
+ *   {
+ *    class: 'FObjectProperty',
+ *    of: 'net.nanopay.tx.model.TransactionEntity',
+ *    name: 'payee',
+ *    storageTransient: true
+ *   },
+ *   {
+ *    class: 'Long',
+ *    name: 'payeeId'
+ *   },
+ *   {
+ *     ...
+ */
+public class PublicUserInfoDAO
+  extends ProxyDAO
+{
+  private String idPropertyName_;
+  private String propertyNameToSet_;
+
+  public PublicUserInfoDAO(
+      X x,
+      String idPropertyName,
+      String propertyNameToSet,
+      DAO delegate
+  ) {
+    super(x, delegate);
+    idPropertyName_ = idPropertyName;
+    propertyNameToSet_ = propertyNameToSet;
+    userDAO_ = (DAO) x.get("localUserDAO");
+    logger_ = (Logger) x.get("logger");
+  }
+
+  protected DAO userDAO_;
+  protected Logger logger_;
+
+  private class DecoratedSink extends foam.dao.ProxySink {
+    public DecoratedSink(X x, Sink delegate) {
+      super(x, delegate);
+    }
+
+    @Override
+    public void put(Object obj, foam.core.Detachable sub) {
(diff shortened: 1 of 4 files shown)


TW80000:
@kgrgreer @shreymahendru @zahnno Here's my abstracted version of what Shrey wrote for transactions. This should be usable anywhere now, I already have it working for the partners feature. If this looks good to you guys, I'll delete Shrey's existing decorator and replace the call to it with two calls to this more generic one.

src/foam/dao/PublicUserInfoDAO.java
```diff
@@ -0,0 +1,112 @@
...
+ *    class: 'Long',
+ *    name: 'payeeId'
+ *   },
+ *   {
+ *     ...
+ */
+public class PublicUserInfoDAO
+  extends ProxyDAO
+{
+  private String idPropertyName_;
```
kgrgreer, line 51: Use protected instead of private to make easier to debug.
Document instance variables.

src/foam/dao/PublicUserInfoDAO.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @Override
+  public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+    Sink decoratedSink = new DecoratedSink(x, sink);
+    getDelegate().select_(x, decoratedSink, skip, limit, order, predicate);
+    return sink;
+  }
+
+  private FObject fillEntitiesInfo(FObject obj) {
```
kgrgreer, line 98: I think fillPublicInfo or fill PublicUserInfo would be a better name.
Make protected or package protection. There's nothing to protect here and being private makes unit testing more difficult.

src/foam/nanos/auth/PublicUserEntity.js
```diff
@@ -0,0 +1,73 @@
+foam.CLASS({
```
kgrgreer, line 1: Is this really a generic nanos feature? Would it be better to move to NP?

src/foam/nanos/auth/PublicUserEntity.js
```diff
@@ -0,0 +1,73 @@
...
+    {
+      class: 'foam.nanos.fs.FileProperty',
+      name: 'profilePicture',
+      view: { class: 'foam.nanos.auth.ProfilePictureView' }
+    }
+  ],
+  axioms: [
+    {
+      writeToSwiftClass: function(cls) {
+        cls.method(foam.swift.Method.create({
```
kgrgreer, line 37: Why do we use objects for swift and strings for java?

src/foam/dao/PublicUserInfoDAO.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public class PublicUserInfoDAO
+  extends ProxyDAO
+{
+  private String idPropertyName_;
+  private String propertyNameToSet_;
+
+  public PublicUserInfoDAO(
+      X x,
+      String idPropertyName,
```
kgrgreer, line 56: Rather than passing in property names which won't be checked by the compiler, you can pass in the PropertyInfo's directly. Ex.: User.ID, or Transaction.PAYEE_INFO. Make the type be foam.mlang.F, which propertyInfo's implement then you can call idProperty_.f(clone) instead of clone.getProperty(idPropertyName_), with the first option being both shorter and faster.

src/foam/nanos/auth/PublicUserEntity.js
TW80000: I realized after opening this PR that it's only generic to foam so long as the subset of User properties comes from the User model defined in foam and doesn't include any refinements NANOPAY makes to User. Right now all of the properties come from the foam definition, so it could work as a generic nanos feature as is. Let me know if you'd like me to move it anyway though.

src/foam/nanos/auth/PublicUserEntity.js
TW80000: @shreymahendru can answer this, all I did was rename this file.

src/foam/dao/PublicUserInfoDAO.java
TW80000: That was my first thought, but I didn't know if it could be done. Right now it would look like this in the services file:

`new foam.dao.PublicUserInfoDAO(x, "payerId", "payer", ...)`

Would the new one look like the following? Will those properties be defined by the time that code is executed?

`new foam.dao.PublicUserInfoDAO(x, net.nanopay.tx.model.PAYER_ID, net.nanopay.tx.model.PAYER, ...)`

My other question was how to write to the second property. I know you can read a value with the `.f()` method, but how would I write to net.nanopay.tx.model.PAYER, for example?

src/foam/nanos/auth/PublicUserEntity.js
shreymahendru: `writeToSwiftClass` should be removed altogether. Cause there is no need for this method on the client-side as an Entity object is not supposed to be initialized there. I figured we would need it for the deserialization of the DTO.

src/foam/nanos/auth/PublicUserEntity.js
TW80000: Update: I have to move it because I'm going to have to add some properties to it that come from NANOPAY's refinement of User, namely `businessAddress` and `businessPhone`.

TW80000:
Closing in favour of: https://github.com/nanoPayinc/NANOPAY/pull/2467

PR #1260, opened 2018-06-20
DAOLogger

jlhughes:
Using a DAO will simplify some decoration such as Notification logging,
and will simplify future requirements such as centralized network logging.

NP-2425

re: NSpec.js - nspec loading generates log messages and which causes infinite recursion.  A problem initializing the DAOLogger and also NotificationLogger.  A quick solution is to output using a StdoutLogger, but a proper solution needs to be discussed.  One possibility is to expose the logger that nspec is using such that the logging services themselves can change it once they are fully started. 

Updates: 29 June 2018 
Added a WriteOnlyJDAO which extends JDAO as the setup is greatly simplified. It uses a WriteOnlyFileJournal which dispenses with id lookup and delta/diff.   This initial configuration is tailored to Log output. 
Example output:
p({"created":"2018-06-29T04:41:12.742Z","severity":1,"createdBy":1,"message":",[JDAO],scripts,Loading file: scripts"})


src/foam/nanos/boot/NSpec.js
```diff
@@ -173,7 +173,8 @@ foam.CLASS({
 
         Interpreter shell = new Interpreter();
         try {
-          shell.set("x", x);
+          foam.core.X y = x.put("logger", new foam.nanos.logger.StdoutLogger());
```
adamvy: Perhaps in the root context that we build in Boot.java, we can put a StdoutLogger as "logger" in the context.  Then any logging that happens before the real logger service is loaded will just use that logger.

src/foam/nanos/logger/LogMessage.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.logger',
+  name: 'LogMessage',
+
+  properties: [
+    {
+      name: 'date',
+      class: 'String'
```
adamvy: I feel like "date" shouldn't be a formatted string but rather just an epoch timestamp that we can format at display time when needed.

If I recall formatting timestamps can take a surprising amount of processing, also a formatted timestamp in your current format is 23 characters long, whereas an epoch timestamp to milisecond precision is only 13 characters (would be even smaller if we used binary encoding)

jlhughes: This is a replacement for our traditional logging, which of course formats the time, and I think from a developer/troubleshooting standpoint, formatted dates are very useful.  Regarding processing time - we'll be replacing like with like.  We could provide flags for toggling formatting  so in the dev/demo environments we get formatted dates in production we get epoch.

src/foam/nanos/boot/NSpec.js
jlhughes: We've got another branch/PR were we are implementing the real solution for this, which involves two Boot.java service setups, the first dealing with logging, and once complete, the regular services setup.  We are going to leave this PR and that for NotificationLogging until it's complete.

src/foam/dao/WriteOnlyFileJournal.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+  ],
+
+  axioms: [
+    {
+      name: 'javaExtras',
+      buildJavaClass: function (cls) {
+        cls.extras.push(`
+          protected Outputter outputter_ = new Outputter(OutputterMode.STORAGE);
+          // FIXME: the following line fails compilation
```
kirkeaton, line 26: Is this fixed? Or does compilation still fail?

src/foam/dao/WriteOnlyFileJournal.js
```diff
@@ -0,0 +1,51 @@
...
+      }
+    }
+  ],
+
+  methods: [
+    {
+      name: 'put',
+      synchronized: true,
+      javaCode: `
+        // FIXME: can't get this line to compile in cls.extras.push
```
kirkeaton, line 38: Same question here.

src/foam/dao/WriteOnlyFileJournal.js
jlhughes: Still fails and I haven't figured it out yet.

kirkeaton (changes requested):
Rejecting until it compiles

PR #1264, opened 2018-06-20
Make the var name in swift overridable

mcarcaso:
This was kind of possible before but would also mess with serialization. Now, it should be isolated to just the variable name.

### DIFF
--- a/src/foam/swift/refines/Property.js
+++ b/src/foam/swift/refines/Property.js
@@ -15,8 +15,16 @@ foam.CLASS({
   properties: [
     {
       class: 'String',
-      name: 'swiftName',
-      expression: function(name) { return name; },
+      name: 'swiftVarName',
+      expression: function(name) {
+        var reserved = [
+          'description',
+        ];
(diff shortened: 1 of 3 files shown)


src/foam/swift/refines/Property.js
```diff
@@ -15,8 +15,16 @@ foam.CLASS({
...
     {
       class: 'String',
-      name: 'swiftName',
-      expression: function(name) { return name; },
+      name: 'swiftVarName',
+      expression: function(name) {
+        var reserved = [
+          'description',
+        ];
+        if ( reserved.indexOf(name) != -1 ) {
```
kgrgreer, line 23: If reserved where a map this could just be reserved[name] and with be N times faster, though N is 1 in this case.

mcarcaso: For sure but I went for an approach that's slightly easier to add items to instead of a more performant approach because it doesn't really matter here. This stuff only runs when we generate code so shouldn't slow down users at all.

PR #1270, opened 2018-06-25
Use package names in swift class names

mcarcaso:
Currently, the package is essentially dropped in the swift world and that can cause some collisions when models have the same name but a different package. This is a rather large PR to go through and use package name everywhere.


kennyqiyenkan (approved):
Its a lot more typing but it LGTM

PR #1271, opened 2018-06-25
Notification view fix

TW80000:
Only show notification message if there is one.

Needed for https://github.com/nanoPayinc/NANOPAY/pull/2467

### DIFF
--- a/src/foam/nanos/notification/NotificationView.js
+++ b/src/foam/nanos/notification/NotificationView.js
@@ -14,7 +14,17 @@ foam.CLASS({
   ],
 
   properties: [
-    'of'
+    'of',
+    {
+      class: 'Boolean',
+      name: 'fullyVisible',
+      documentation: `If a notification's body content is too long, it will be
+          truncated and an ellipsis will be added to the cutoff point. Clicking
+          on a notification toggles between truncated form and being fully
+          visible. This property tracks whether the notification is being
+          truncated or displayed in full.`,
+      value: false
+    }
   ],
 
   methods: [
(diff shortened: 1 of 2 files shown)


TW80000:
@penzital @zahnno @kgrgreer please review

src/foam/nanos/notification/NotificationView.js
```diff
@@ -8,24 +8,36 @@ foam.CLASS({
...
+          .show(this.data.body !== '')
+        .end();
     }
   ],
+
   listeners: [
-    function onClick () {
-      if ( this.childNodes[0].css.display == "block" ) {
-        this.childNodes[0].style({display:'-webkit-box'})
+    function onClick() {
```
kgrgreer: Referencing childNodes[0] is fragile. Better to toggle a boolean property and then bind the style to the slot with a map at the creation side so that this code doesn't need to be changed if the child changes its location.

TW80000:
@kgrgreer Updated, please review again, thanks.

src/foam/nanos/notification/NotificationView.js
```diff
@@ -8,24 +8,43 @@ foam.CLASS({
...
+    }
+  ],
+
   methods: [
     function initE() {
       this.SUPER();
       this
         .addClass(this.myClass())
-        .start('div').on('click', this.onClick).addClass('msg').add(this.data.body).end();
+        .start('div')
```
zahnno: `.start()` defaults to div. Passing in 'div' as an argument is somewhat redundant... Just a side note.

TW80000: Thanks, fixed.

PR #1273, opened 2018-06-26
Notification logger config

deejes:
Increases the initial errorLevelThreshhold amount so that the logger is not executed by default.
Adds documentation to the class and the errorLevelThreshhold field.

### DIFF
--- a/src/foam/nanos/logger/NotificationLogger.js
+++ b/src/foam/nanos/logger/NotificationLogger.js
@@ -37,7 +37,12 @@ foam.CLASS({
     {
       class: 'Int',
       name: 'errorLevelThreshhold',
-      value: 5,
+      value: 4,
+      /** This value needs to be changed for the logger to execute. 
+       *  Here, DEBUG -> 0, INFO -> 1, WARNING -> 2, ERROR -> 3.
+       *  The equality condition for execution is greater than or equal to.
+      */
+
     },
     {
       class: 'Enum',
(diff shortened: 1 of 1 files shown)


src/foam/nanos/logger/NotificationLogger.js
```diff
@@ -33,7 +37,12 @@ foam.CLASS({
     {
       class: 'Int',
       name: 'errorLevelThreshhold',
-      value: 1,
+      value: 4,
```
jlhughes, line 40: Is this exposed in the GUI? If so, would any save enable it as it would select something other than '4'.  I'd rather have an 'enabled' flag, and the the severity selection list will always be valid.

jlhughes: We'll install a script to enable it for now until we move the config to Notification Center profiles.

deejes: p({"class":"foam.nanos.script.Script","id":"notificationLogger - setErrorLevelThreshhold","enabled":true,"lastRun":"2018-06-26T20:09:00.994Z","lastDuration":2409,"status":2,"code":"logger = x.get(\"logger\");\nprint(logger);\n\nnotificationLogger = logger.getDelegate().getDelegate().getDelegate();\n\nnotificationLogger.setErrorLevelThreshhold(3);\n\no = new Object[3];\no[0] = 'd';\no[1] = 'a';\no[2] = '3';\n\nlogger.error(o);\n\n"})

PR #1274, opened 2018-06-26
LastModifiedByAwareDAO

kirkeaton:
- Made some concurrency changes to JDAO and FileJournal
 - Wrote LastModifiedByAwareDAO


jlhughes (approved):

src/foam/nanos/auth/LastModifiedByAwareDAO.js
```diff
@@ -0,0 +1,40 @@
...
+  name: 'LastModifiedByAwareDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  methods: [
+    {
+      name: 'put_',
+      javaCode: `
+        if ( obj instanceof LastModifiedByAware ) {
+          User user = (User) x.get("user");
+          ((LastModifiedByAware) obj).setLastModifiedBy(user.getId());
```
jlhughes, line 18: Should we be cloning here? (fclone)

eatonk:
Why do we need to clone here?

src/foam/nanos/auth/LastModifiedByAwareDAO.js
kirkeaton: Why do we need to clone here?

jlhughes: ah yes, it's already been cloned.

jlhughes (approved):

PR #1277, opened 2018-06-27
 Modified LastModifiedByAware to be a reference to a user instead of a Long

kirkeaton:


jlhughes (approved):

jlhughes (approved):
This is great, thanks.

PR #1288, opened 2018-07-05
Remove false statement from documentation

TW80000:


adamvy:
I don't believe this statement is false.  It will re-run the renderers whenever the contents of the DAO change, _if_ it receives notifications from the DAO.

In nanos most of the DAOs are currently just HTTP Client DAOs which do not issue notifications when the contents change, so the view is unable to update automatically.  This is a bug/limitation in those DAO implementations, but other DAOs like websocket client daos or client side MDAOs or ArrayDAOs will issue notifications and the Element.select() will work as expected.

If you want you can add a disclaimer to this comment saying that it relies on a DAO implementation to issue update events and certain client DAOs do not do that reliably, but don't remove the comment entirely.

TW80000:
> It will re-run the renderers whenever the contents of the DAO change, *if* it receives notifications from the DAO

How would one send such a notification?

As an aside, if it means anything to you, I wrote this documentation comment a few weeks after starting work with FOAM and I wasn't actually sure if `.select()` updated automatically when the underlying DAO changed. I wrote that because of a conversation I had with Kevin about it but I might have misunderstood what he had said. So now that I'm using `.select()` and expecting it to update but it's not, I assumed I had just written the documentation wrong. I could be using it wrong though. Do you know of a working example somewhere in the code I can look at?

adamvy:
You and Kevin may have been talking about the two different .select()s.  On the DAO there is a .select() which just reads the contents of the DAO once and is not updated when the DAO updates.  On a U2 Element there is .select() which is this one here, which creates a listener on the DAO and _does_ re-render when the DAO issues notifications.

You can see it in action in the sevenguis CRUD demo. Under src/com/google/foam/demos/sevenguis/index.html (you can open this with just file:///path/to/foam2/src/com/google/foam/demos/sevenguis/index.html).  Notice when you create objects in the CRUD demo they are automatically added to the tableview?  This is handled in foam/u2/view/TableView.js which essentially does this.select(this.dao, function() { <format row> }).

Didn't notice that you initially wrote this documentation, its great!  I think your initial understanding was correct, it should update when the contents of the DAO change.

In JS the notifications are issued from the DAO by doing this.pub('on', 'put', ...) as seen here in MDAO.js
https://github.com/foam-framework/foam2/blob/master/src/foam/dao/MDAO.js#L155

In Java they are done by calling onPut()/onRemove() as seen in MapDAO here:
https://github.com/foam-framework/foam2/blob/master/src/foam/dao/MapDAO.java#L58

However they can't be easily added for the Http client dao because HTTP is a request response protocol, there's no easy mechanism for the server to push events to the client that the client has not specifically asked for.  We could hook something up using long polling and a server side event queue, but I haven't had a chance to make that ;).

TW80000:
> You and Kevin may have been talking about the two different .select()s. On the DAO there is a .select() which just reads the contents of the DAO once and is not updated when the DAO updates. On a U2 Element there is .select() which is this one here, which creates a listener on the DAO and does re-render when the DAO issues notifications.

We were talking about `Element.select()`, I was clear on that at the time. Thanks for making sure that was clear though.

> You can see it in action in the sevenguis CRUD demo. Under src/com/google/foam/demos/sevenguis/index.html (you can open this with just file:///path/to/foam2/src/com/google/foam/demos/sevenguis/index.html). Notice when you create objects in the CRUD demo they are automatically added to the tableview? This is handled in foam/u2/view/TableView.js which essentially does this.select(this.dao, function() { }).

Thanks, I just checked the demo out, still looking through the code now. I noticed it uses `foam.u2.TableView`, which is deprecated, and not `foam.u2.view.TableView` which is its replacement. The old (deprecated) TableView the demo is using uses an `onDAOUpdate` approach that explicitly subs to changes on the DAO and explicity replaces the view content in the `onDAOUpdate` code. This isn't what the new TableView does. It uses `Element.select()` instead.

> In JS the notifications are issued from the DAO by doing this.pub('on', 'put', ...) as seen here in MDAO.js
https://github.com/foam-framework/foam2/blob/master/src/foam/dao/MDAO.js#L155

> In Java they are done by calling onPut()/onRemove() as seen in MapDAO here:
https://github.com/foam-framework/foam2/blob/master/src/foam/dao/MapDAO.java#L58

Hmm. That seems to be an event when you put to a dao. I want the view to update when a filter is applied to the dao. Ie, if I pass `someDAO$proxy` as the first argument to `Element.select()`, I want it to update when the delegate DAO that the proxy is pointing to changes. Should that be happening? Or am I taking the wrong approach to accomplish that? The delegate is the same DAO, but for example I'm switching between that DAO and a filtered version of it (ie: with some `.where()` conditions) based on what the user is filtering for.

> However they can't be easily added for the Http client dao because HTTP is a request response protocol, there's no easy mechanism for the server to push events to the client that the client has not specifically asked for. We could hook something up using long polling and a server side event queue, but I haven't had a chance to make that ;).

I understand what you're talking about here, but my DAO's are only changing on the frontend (my use case mentioned above). My use case isn't to listen for puts or anything, which I understand would be happening on the backend and wouldn't be easily detectable without either a polling or interrupt-based scheme (or otherwise). I think that's a different issue that I'm dealing with here.

kgrgreer:
Perhaps better text would be: 

Will update the view whenever the DAO fires change events.

PR #1290, opened 2018-07-06
Polyfill local storage with in memory version if local storage is in a broken state.

### LINKED ISSUE #1289: nanos fails to load if localStorage is broken
It's possible to get localStorage into a corrupted state where all operations throw an exception as seen here: https://stackoverflow.com/questions/18877643/error-in-local-storage-ns-error-file-corrupted-firefox

Currently nanos requires localStorage to not throw in order to operate successfully, otherwise things like the SessionClientBox and the ClientBuilder will throw exceptions and never finish their execution.

We could fix all these call sites to guard against localStorage exceptions, however those exceptions are a very rare occurence and not expected by most developers, especially as they are not even documented on the Mozilla Developer Network (https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem)

Instead I plan to polyfil localStorage (and possibly sessionStorage) with just an in memory storage that will persist for the lifetime of the page when localStorage is detected to be broken.

adamvy:
Fixes #1289 

I'm open to other suggestions too.  Long term it would be good to display a notification to the user that they should restart their browser or something.

It would also be better to stop using local storage directly and use a modeled foam service provided via context for local storage things, but this fixes a real problem we have now finding and re-writing all localstorage usages.

### DIFF
--- a/src/foam/core/poly.js
+++ b/src/foam/core/poly.js
@@ -113,3 +113,42 @@ if( ! Object.is ) {
     return x !== x && y !== y;
   };
 }
+
+if ( typeof localStorage != "undefined" ) {
+  try {
+    localStorage.getItem('asdf')
+  } catch(e) {
+    console.warn("Local storage error", e);
+    console.warn("Substituting in memory local storage.");
+
+    Object.defineProperty(this, 'localStorage', {
+      value: (function() {
+        var prototype = {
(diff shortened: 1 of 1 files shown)


kgrgreer (approved):

kgrgreer:
How does localStorage get broken?

PR #1294, opened 2018-07-09
Document RenderSink

TW80000:
~~Currently, the part of the view created by a call to `.select()` will
only update when the dao that it uses is `put` to or `remove`d from. This
commit changes that behaviour so that the view is updated on any change
to the DAO, not just put and remove specifically.~~

~~Necessary for https://github.com/nanoPayinc/NANOPAY/pull/2624~~

~~Makes https://github.com/foam-framework/foam2/pull/1288 unnecessary~~

Adds documentation for RenderSink.


src/foam/u2/Element.js
```diff
@@ -511,7 +526,7 @@ foam.CLASS({
 
           var objs = a.array;
           self.cleanup();
-          for ( var i = 0 ; i < objs.length ; i++ ) {
+          for ( var i = 0; i < objs.length; i++ ) {
```
kgrgreer: Add spaces before ;'s.

src/foam/u2/Element.js
```diff
@@ -471,20 +478,28 @@ foam.CLASS({
...
-      name: 'batch'
+      name: 'batch',
+      documentation: `Used to check whether a paint should be performed or not.`
     }
   ],
 
   methods: [
+    function init() {
+      // Repaint on every change to the DAO.
+      this.dao.sub('on', 'reset', this.paint);
```
kgrgreer: I don't think this should be done here. I don't think it's the RenderSink's responsibility and that the listen should be done in the Element.select() method.

src/foam/u2/Element.js
```diff
@@ -1611,6 +1622,8 @@ foam.CLASS({
       this.onDetach(dao.listen(listener));
       listener.delegate.paint();
 
+      dao.sub('on', 'reset', listener.delegate.paint);
```
kgrgreer: I think this should be wrapped in this.onDetach() so that when the GUI is removed it doesn't keep getting events. Why is this calling listener.delegate.paint? Wouldn't listener be sufficient?

kgrgreer: Also, this seems like a problem with the DAO design. We are already doing a dao.listen() above. We shouldn't need to do both a listen() and sub(). @adamvy, why is this?

TW80000: Replacing

```JavaScript
dao.sub('on', 'reset', listener.delegate.paint);
```
with
```JavaScript
dao.sub('on', 'reset', listener);
```
doesn't work.

src/foam/u2/Element.js
```diff
@@ -455,11 +455,19 @@ foam.CLASS({
...
-  implements: [ 'foam.dao.Sink' ],
+  implements: [
+    'foam.dao.Sink'
+  ],
+
+  documentation: `
+    Given a DAO and a function that maps from an object in that DAO to an
+    Element, apply the function to each object in the DAO and add the returned
+    elements to the view. Will render again every time the DAO is put to or
+    removed from.
```
zahnno: This is worded oddly.

TW80000: What about it isn't clear so I can improve it?

zahnno: Maybe you can try rewording "Given a DAO and a function that maps from an object in that DAO to an Element"

src/foam/u2/Element.js
TW80000: @adamvy @kgrgreer Any update on this?

adamvy: Can you make a repro test case/demo that this fixes?  I have not had any problem with views not re-rendering when reset is called.

adamvy: Or if you can give me repro steps for nanoPayinc/NANOPAY#2537 I can try that as well.

TW80000: Repro steps for NANOPAY:

1. Go to "Invoices > Payable"
2. Click on a summary card at the top to filter invoices. It should work. Clicking again will deselect it and the table should update as expected.
3. Click on any invoice in the table.
4. Click the "Back" button in the upper left of the page.
5. Try filtering the table using the summary cards again, now it doesn't work.

adamvy: Look at https://github.com/nanoPayinc/NANOPAY/compare/b/NP-2537/fix-table-dao-update?expand=1 for a proper fix for that bug.

zahnno:
Any updates or concerns stopping the merge on this PR? @kgr @TW80000 @adamvy

TW80000:
@adamvy Thinks this shouldn't be necessary and there should be a different way to fix the issue I was facing. He was helping me look into it but I've been working on other things for a bit now and haven't gone back to this yet.

adamvy:
There's an issue with the dao listener being detached and never recreated I believe.  I didn't finish investigating.  I believe this could be stack related.  Will hopefully investigate later this week.

adamvy:
Also my concerns for the code as it stands is it will add a memory leak as the listener is hooked subscribed but never unsubscribed, and that .on.sub is not meant to be a part of the dao interface, its am implementation detail of how abstract dao implements .listen

adamvy:
I have determined the root bug.  There is a flaw in the design of ProxyDAO/ProxyListener.  I am working on a fix.  Follow #1381 for updates on the root bug.

adamvy:
I have fixed the root bug! https://github.com/nanoPayinc/NANOPAY/pull/2624 now works out of box on foam2 master as of https://github.com/foam-framework/foam2/commit/435dcd78b76ffb16f4cc3d99173015aa99007c3b (I have verified this).

Thanks for finding the bug and bringing it to my attention, and for your patience while I investigated.

Can we take the documentation portions of this PR out into their own PR and drop this one as the behaviour changes are not unnecessary.  Thanks!

TW80000:
@adamvy Thanks for finding the root cause and making the fix. I've updated this PR to remove the behaviour change and left the documentation. I also updated the title of this PR to reflect that.

PR #1295, opened 2018-07-09
Modelled JournalEntry class

kirkeaton:


src/foam/dao/journal/CommentJournalEntry.js
```diff
@@ -0,0 +1,17 @@
+/**
```
kgrgreer, line 1: I don't think we need this since what we used to store in comments is now stored in the JournalEntry itself.

src/foam/dao/journal/EasyJournalEntryDAO.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao.journal',
+  name: 'EasyJournalEntryDAO',
```
kgrgreer, line 9: What is this for?

src/foam/dao/journal/FileAppendDAO.js
```diff
@@ -0,0 +1,100 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao.journal',
+  name: 'FileAppendDAO',
+  extends: 'foam.dao.AbstractDAO',
```
kgrgreer, line 10: I think it's confusing to make this a DAO when it only supports two methods. Better to make a separate Journal interface.

src/foam/dao/journal/PutJournalEntry.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao.journal',
+  name: 'PutJournalEntry',
```
kgrgreer: Rather than having subclasses of JournalEntry, I would rather have 'cmd' be a property of JournalEntry. This lets us disable 'outputDefaultClasses' and decrease journal sizes.

src/foam/dao/journal/PutRemoveJournalDAODecorator.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao.journal',
+  name: 'PutRemoveJournalDAODecorator',
```
kgrgreer, line 9: What is this for?

src/foam/dao/journal/UserCommentJournalEntryDAODecorator.js
```diff
@@ -0,0 +1,86 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao.journal',
+  name: 'UserCommentJournalEntryDAODecorator',
```
kgrgreer, line 9: Better to put this information as properties of JournalEntry.

src/foam/dao/journal/PutJournalEntry.js
kirkeaton: Should command just be a character? Or an enum?

src/foam/dao/journal/CommentJournalEntry.js
mcarcaso: We should think about whether we want the comment in the journal entry or in the object. We recently moved it into the object with the LastModifiedByAware stuff but I'm not sure if that's the best place for it because now it requires us to add these traits to each of those objects. If this info is just for audit purposes, I think we should do it this way because it'll apply to everything that's put into a dao and the object doesn't need to care when it was last modified. It makes for slightly less developer overhead because we won't need to worry about adding the various implements on each object.

src/foam/dao/journal/EasyJournalEntryDAO.js
mcarcaso: Convenience for when we want to add more features to these journal entries. Helps ensure we use the right order when putting the decorators together.

src/foam/dao/journal/FileAppendDAO.js
mcarcaso: It's nice for this to be a DAO because we can swap it out with any other DAO for testing purposes or we can decorate it with any of the many DAO decorators we already have. It's already being decorated in the EasyJournalEntryDAO to get sequence numbers on the journal entries.

src/foam/dao/journal/PutJournalEntry.js
mcarcaso: I've thought about this approach and ultimately went with this because we can now add features independent of any list of commands that already exist and nothing needs to know about them. This lets projects that don't care about certain features never need to know about this. For example, if project 'foo' needs encryption and project 'bar' doesn't, project 'bar' will never get any code that suggests there's an encryption feature on journal entries. If journal size is a concern, we need a better outputter or a CompressedJournalEntry decorator.

src/foam/dao/journal/PutRemoveJournalDAODecorator.js
mcarcaso: This is the most basic DAO decorator for all of this. It's what deals with telling the JournalEntryDAO about a put or a remove by putting a JournalEntry in the context under "journalEntry". It doesn't care about comments, diffing, hasing, etc. It's super simple and using this alone should achieve a working journal.

src/foam/dao/journal/UserCommentJournalEntryDAODecorator.js
mcarcaso: The problem with that approach is that it limits you to a single comment on a journal entry. Using decoration allows us to add any number of comments if we want to.

src/foam/dao/journal/PutJournalEntry.js
kirkeaton: @kgrgreer Modelling the JournalEntry is going to increase journal sizes no matter how you spin it. Even _if_ we disable outputDefaultClasses.

src/foam/dao/journal/CommentJournalEntry.js
jlhughes: This seems reasonable for two reasons:
1. retains current auditing information without immediately requiring us to update all objects with Created/LastModified.
2. This along with the UserComment decorator can be disable where it's not needed/wanted.

jlhughes (commented):
Mike, you should add your 'answers' as documentation. 

@kgrgreer overall, I'm ok with this. It's working, gives us parity with our existing output, and with regard to 'cmd', could occur later.

mcarcaso:
I do not like this 'cmd' approach for reasons I stated above that were never addressed. Please explain the benefits for this approach. This is not the place to be solving the journal size issue. That's a problem for the outputter to solve, not the JournalEntry. Now the CommentJournalEntry has properties on it that don't make much sense and this has introduced more complexity than the previous approach.

PR #1297, opened 2018-07-09
FixedSizeDAO

deejes:
merge FixedSizeDAO to store a specified number of objects


src/foam/dao/FixedSizeDAO.js
```diff
@@ -0,0 +1,75 @@
...
+          name: 'x',
+          of: 'foam.core.X'
+        },
+        {
+          name: 'obj',    
+          of: 'foam.core.FObject',
+        }
+      ],
+      javaCode: `
+getFixedSizeArray().add(obj);
```
jlhughes: put via the delegate first and place the result in the array.

src/foam/dao/FixedSizeDAO.js
```diff
@@ -0,0 +1,75 @@
...
+return getDelegate().put_(x, obj);
+  `
+    },
+    {
+      name: 'select_',
+      javaReturns: 'foam.dao.Sink',
+      javaCode: `
+if ( sink.equals(null) ) {
```
jlhughes: This will fail if sink is null. 
Change to .   sink == null

src/foam/dao/FixedSizeDAO.js
```diff
@@ -0,0 +1,75 @@
...
+    'org.apache.commons.collections.buffer.CircularFifoBuffer',
+    'foam.dao.Sink',
+  ],
+  
+  properties: [
+    {
+      class: 'Object',
+      name: 'FixedSizeArray',
+      javaType: 'org.apache.commons.collections.buffer.CircularFifoBuffer',
+      javaFactory: `return new org.apache.commons.collections.buffer.CircularFifoBuffer(10000); `
```
jlhughes: move the buffer size to a property, then we can set it via the builder. 
also, on buffer size change, recreate the fixedSizeArray, then we can change it at runtime via a script.

jlhughes (changes requested):

src/foam/dao/FixedSizeDAO.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * @license
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
kgrgreer: Not copyright Google.

/**
 * @license
 */

src/foam/dao/FixedSizeDAO.js
deejes: Do you know of an example for the second issue ( rerun factory on property update? ).
Also, will we be wanting to copy over however many preexisting logs will fit into the newly created array?

src/foam/dao/FixedSizeDAO.js
```diff
@@ -0,0 +1,69 @@
...
+    {
+      class: 'Int',
+      name: 'fixedArraySize',
+      value: 10000
+    },
+    {
+      class: 'Object',
+      name: 'FixedSizeArray',
+      javaType: 'org.apache.commons.collections.buffer.CircularFifoBuffer',
+      javaFactory: `return new org.apache.commons.collections.buffer.CircularFifoBuffer(getFixedArraySize()); `
```
jlhughes: This buffer is itself not thread safe. 
Two options: 
1. Buffer fifo = BufferUtils.synchronizedBuffer(new CircularFifoBuffer());
2. Use a regular Array and add a synchronized 'nextIndex' method. 

Option 1 is easier as the CircularFifoBuffer handles correct iteration order during toArray on select().

jlhughes (changes requested):

jlhughes (approved):

src/foam/dao/FixedSizeDAO.js
```diff
@@ -0,0 +1,71 @@
...
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'FixedSizeDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  documentation: function () {/*
```
kgrgreer, line 12: Just use a String rather than a function comment or a multi-line string (quoted with back-quotes "`") if it spans more than one line.

src/foam/dao/FixedSizeDAO.js
```diff
@@ -0,0 +1,71 @@
...
+    DAO that stores a fixed number of objects.
+  */},
+
+  javaImports: [
+    'org.apache.commons.collections.buffer.CircularFifoBuffer',
```
kgrgreer, line 17: sort imports

src/foam/dao/FixedSizeDAO.js
```diff
@@ -0,0 +1,71 @@
...
+  `
+    },
+    {
+      name: 'select_',
+      javaReturns: 'foam.dao.Sink',
+      javaCode: `
+if ( sink == null ) {
+  sink = new foam.dao.ArraySink();
+}
+Object[] tester = getFixedSizeArray().toArray();
```
kgrgreer, line 63: Is there a way to walk the data without making a copy?

jlhughes: @kgrgreer Alternative is to use/manage an Array ourselves and not use the Apache CircularFifoBuffer.

PR #1299, opened 2018-07-10
Fix MDAO test

TW80000:
The MDAO test was getting a beanshell interpreter error because in the
test a call was made to `MDAO.addIndex()`, but the definition of that
method takes varargs, which aren't supported by the version of beanshell
we're using (I think). According to

  https://github.com/beanshell/beanshell/issues/23

though, varargs are now supported in beanshell, so we should update in
order to avoid problems like this in the future.

This commit fixes the problem with a workaround by calling a different
method on MDAO to add the index that doesn't take varargs.


TW80000:
@jlhughes Please review.

src/tests
```diff
@@ -1,5 +1,5 @@
...
+p({"class":"foam.nanos.test.Test","id":"MDAO test","description":"MDAO test","server":true,"scheduled":false,"code":"mdao = new foam.dao.MDAO(foam.nanos.auth.Country.getOwnClassInfo());\nmdao.addUniqueIndex(foam.nanos.auth.Country.CODE); \ncountry = new foam.nanos.auth.Country();\ncountry.setCode(\"CN\");\ncountry.setName(\"CHINA\");\nmdao.put(country);\ncountry = country.fclone();\ncountry.setCode(\"CA\");\ncountry.setName(\"CANADA\");\nmdao.put(country);\ncountry = country.fclone();\ncountry.setCode(\"USA\");\ncountry.setName(\"United States of America\");\nmdao.put(country);\nresult = mdao.find(\"CA\");\ntest(result.getName().equals(\"CANADA\"),\"test if country name of CA is CANADA?\");\nresult = mdao.find(\"CAD\");\ntest(result==null,\"test if country name of CAD not found?\");","output":"SUCCESS: test if country name of CA is CANADA?\nSUCCESS: test if country name of CAD not found?\n","notes":""})
```
kgrgreer: Calling addUniqueIndex() isn't required because 'code' is the primaryKey, so is already automatically added.

PR #1305, opened 2018-07-11
InnerEnum Fix

mcarcaso:


adamvy:
This looks reasonable, but when is o.class ever not InnerEnum?

mcarcaso:
Not sure but I guess it's possible for somebody to put in a class that could extend InnerEnum or something? You think I should force it to be an InnerEnum rather than using class?

And btw, this is needed because the build tool I'm working on spits out inner enums like this:

"enums":[{"class":"foam.core.InnerEnum","model":{"class":"foam.core.EnumModel","values":[{"class":"foam.core.internal.EnumValueAxiom","definition":{"name":"NO_CONFIRM","ordinal":0}},{"class":"foam.core.internal.EnumValueAxiom","definition":{"name":"CONFIRM","ordinal":1}},{"class":"foam.core.internal.EnumValueAxiom","definition":{"name":"DEBOUNCE","ordinal":2}},{"class":"foam.core.internal.EnumValueAxiom","definition":{"name":"ARMED","ordinal":3}}],"name":"ButtonState"}}]

Might want to change the outputter to write out something smaller in the future but this output should be handled properly.

adamvy:
I still don't understand why this is needed in your case.  You have enums: [ array of InnerEnum ].  Your new adapt will adapt do x.lookup('foam.core.InnerEnum').create({ "class": "InnerEnum" "model" { " class" ... }})

whereas before it would have done foam.core.InnerEnum.create({ "class": InnerEnum", "model": { "class" ...

It's the same thing is it not?  Only that one does foam.lookup?

mcarcaso:
It does:
foam.core.InnerEnum.create({model: o})
So expanded, it'd be:
foam.core.InnerEnum.create({model: { class: 'InnerEnum', model: {...}})

adamvy:
Oh gotcha, that's kinda subtle and annoying lol.

I wonder if there's a more obvious way to write the condiional.

PR #1308, opened 2018-07-12
Log parameter key/value pairs so TraceWebAgent can also serve as Logg…

jlhughes:
…ingWebAgent


src/foam/nanos/http/TraceWebAgent.java
```diff
@@ -74,13 +87,19 @@ public void execute(X x) {
...
+        kv.put("_WebAgentParameters_", params.toString());
       }
       out.println("</TABLE>");
 
       out.println("</BODY></HTML>");
 
+      StringWriter stringWriter = new StringWriter();
+      Outputter outputter = new Outputter(new PrintWriter(stringWriter), OutputterMode.FULL);
+      outputter.output(kv);
+      logger.info("TraceWebAgent", stringWriter.toString());
```
kgrgreer: Minor point: is to .toString() required?

PR #1312, opened 2018-07-12
Fix notification message

TW80000:
Changed notification popup menu item text from "Not show like this" to "Hide similar notifications".

Also fixes the CSS for the popup menu.

Old: 
screen shot 2018-07-12 at 11 42 40 am

New: 
screen shot 2018-07-12 at 11 42 14 am

### DIFF
--- a/src/foam/nanos/notification/NotificationRowView.js
+++ b/src/foam/nanos/notification/NotificationRowView.js
@@ -28,24 +28,6 @@
     ],
 
     css: `
-      /*^ {
-        padding-right: 8px;
-        margin: 8px;
-        display: flex;
-        //background: gray;
-        width: 220px;
-        border-radius: 5px;
-      }
-      *//*^:hover {
-        background: #DDD;
-      }*//*
-      ^ .id {
-        padding: 8px;
-        border-radius: 4px 0 0 4px;
-        //color: white;
-        //background: #607D8B;
-      }*/
-      
       ^ i {
         margin-top: 5px;
       }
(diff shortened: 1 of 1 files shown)


TW80000:
@jingyilai0809 @tzchnn Please review.

tzchnn:
I think we should change the list to something like this:

screen shot 2018-07-12 at 2 04 24 pm

Note the text is left aligned.

alternatives:
- Remove --> red text, since this actually sounds pretty destructive?
- "Don't show notifications like this"
- "Hide notifications like this"

**Context:**
> If you say that you don't want to see a certain type of notification (by clicking "Hide similar notifications"), you will still get those notifications, but they won't show up in your notifications screen until you go to your settings and say that you want to see them again. Then you'll see them.
Clicking remove will actually delete the notification from storage, so you'll never be able to see it again.

screen shot 2018-07-12 at 1 17 03 pm

TW80000:
Update: 
screen shot 2018-07-13 at 9 08 39 am

Hovered: 
screen shot 2018-07-13 at 9 08 49 am

jingyilai0809:
I agree with @tzchnn on making it a red color as it is a destructive action. Also, it will be great if we can keep consistent with the hover effect.

TW80000:
Note to self from conversation with @jingyilai0809: When "Remove" option is hovered, text should be white.

tzchnn:
looks good to me. 

the reason I put red text as an alternative is b/c it depends how destructive it is.
Less severe = if the notification isn't that important and wouldn't be something users come back to reference, use white text, and "Dismiss" works too
Severe = red text, possibly include modal or overlay to notify user of consequences

jingyilai0809:
We have other actions like Delete payee, but since there is a confirmation screen for that, we didn't make it red. But I don't think it is a good choice to add confirmation for the notification as it is not so important ( they also have an email delivered to their account) so a red text looks good to me for an alert. 

But I want to know what "Remove" means...it is to delete this notification?

kgrgreer:
@TW80000 Ready to merge?

TW80000:
@kgrgreer Not yet, will work on it now.

TW80000:
@jingyilai0809 Yes, it will delete the notification. You won't be able to see it anywhere else after it's deleted.

TW80000:
@kgrgreer PR is ready for merge if @tzchnn and @jingyilai0809 are OK with it.

tzchnn:
looks good to me

PR #1313, opened 2018-07-12
Suppress outputting password value

jlhughes:


kgrgreer:
Do the same for the "sessionId".

PR #1315, opened 2018-07-12
Add 'of' property to Constant and use it in Element

mcarcaso:
This allows dependencies to be identified easier.

### DIFF
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -560,7 +560,6 @@ foam.CLASS({
 
   requires: [
     'foam.u2.AttrSlot',
-    'foam.u2.DefaultValidator',
     'foam.u2.Entity',
     'foam.u2.ViewSpec'
   ],
@@ -596,6 +595,7 @@ foam.CLASS({
 
     {
       name: 'DEFAULT_VALIDATOR',
+      of: 'foam.u2.DefaultValidator',
       factory: function() { return foam.u2.DefaultValidator.create(); }
     },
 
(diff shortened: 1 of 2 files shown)


src/foam/u2/Element.js
```diff
@@ -625,6 +627,7 @@ foam.CLASS({
         An unloaded Element can be readded to the DOM.
       `,
       name: 'UNLOADED',
+      of: 'foam.u2.UnloadedElementState',
```
adamvy, line 630: LOL this reads very redundant with `of: 'UnloadedState', factory: new UnloadedState()` but I see why you did it. :)

mcarcaso: haha yeah I was thinking we could get fancy and maybe generate a factory if an 'of' is given but not sure that's a good idea yet.

adamvy: Yeah me either, more often I think of would just be ElementState not the specific instance but in this case you need to caprue all the deps.

What context is the factory run in?  Would this.UnloadedState work if UnloadedState was required?

Furthermore I actually tried to model state machines like this in the past so it was much more high level and explicit rather than being constants and stuff, but I never managed to get it to a state that I felt it was good enough :)

mcarcaso: Good question. Currently, `this` is just the window so this.UnloadedState won't be a thing. Running these factories from a context that has these classes in them is something worth considering but not sure I really care to go down that road just yet. I'll need to get annoyed a little more first haha.

PR #1316, opened 2018-07-12
From put_(x call find_(x instead of find(

jlhughes:
Also, add javascript implementations of put() to set Created/LastModified)

### DIFF
--- a/src/foam/nanos/auth/CreatedAwareDAO.js
+++ b/src/foam/nanos/auth/CreatedAwareDAO.js
@@ -19,10 +19,16 @@ foam.CLASS({
   methods: [
     {
       name: 'put_',
+      code: function(value) {
+        if ( foam.util.SafetyUtil.isEmpty(obj.id) ) {
+          value.created = new Date();
+        }
+        return this.SUPER(value);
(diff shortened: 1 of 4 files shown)


kirkeaton (approved):

src/foam/nanos/auth/CreatedAwareDAO.js
```diff
@@ -19,10 +19,16 @@ foam.CLASS({
   methods: [
     {
       name: 'put_',
+      code: function(value) {
+        if ( foam.util.SafetyUtil.isEmpty(obj.id) ) {
```
kgrgreer, line 23: Do we have foam.util.SafetyUtil in JS?

adamvy: I am fairly certain we don't, also this is the wrong method signature for JS it should be put_(x, obj).

Also you called it value in the arguments but obj in the body.

Can we not commit JS implementations that have obviously never been run?  I would much rather leave the implemtnation empty or throw an exception than submit code that give the impression that its implemented but actually isn't.

PR #1322, opened 2018-07-13
Complete and test javascript implementations of put on CreatedAware a…

jlhughes:
…nd LastModifiedAware


src/files.js
```diff
@@ -242,11 +242,29 @@ FOAM_FILES([
   { name: "foam/blob/BlobJava", flags: ['java'] },
   { name: "lib/node/json_dao", flags: ['node'] },
   { name: "lib/utf8" },
+  { name: "foam/nanos/auth/EnabledAware" },
```
adamvy: Most of the nanos files are already listed in src/foam/nanos/nanos.js Can we add require('../../src/foam/nanos/nanos.js') to test/helpers/assert.js similar to how we load foam.js there?

src/foam/nanos/auth/CreatedAwareDAO.js
```diff
@@ -19,11 +19,11 @@ foam.CLASS({
   methods: [
     {
       name: 'put_',
-      code: function(value) {
-        if ( foam.util.SafetyUtil.isEmpty(obj.id) ) {
-          value.created = new Date();
+      code: function(x, obj) {
+        if ( ! obj.id ) {
+          obj.created = new Date();
```
adamvy: I feel like the JS should also check that obj implements CreatedAware same as Java.

adamvy: In java we check if delegate.find_(x, id) == null, can we do the same for JS?  It would be best to keep these implementations consistent.

I also don't like checks for if ( obj.id ), because that makes values like 0 not a valid id but sometimes people try to use that, let's just check delegate.find_() or just check obj.hasOwnProperty('id') like SequenceNumberDAO does.

src/foam/nanos/auth/LastModifiedAwareDAO.js
```diff
@@ -20,9 +20,9 @@ foam.CLASS({
   methods: [
     {
       name: 'put_',
-      code: function(value) {
-        value.lastModified = new Date();
-        return this.SUPER(value);
+      code: function(x, obj) {
+        obj.lastModified = new Date();
```
adamvy: Again I feel like we should be checking if obj implements LastModifiedAware

src/foam/nanos/auth/LastModifiedByAwareDAO.js
```diff
@@ -12,9 +12,9 @@ foam.CLASS({
   methods: [
     {
       name: 'put_',
-      code: function(value) {
-        value.lastModifiedBy = x.get('user').id;
-        return this.SUPER(value);
+      code: function(x, obj) {
+        obj.lastModifiedBy = x.user.id;
```
adamvy: ditto

PR #1325, opened 2018-07-15
Gui view for FixedSizeDAO

deejes:


src/foam/nanos/logger/AbstractLogger.js
```diff
@@ -84,7 +84,6 @@ return String.valueOf(obj);`
       javaCode:
       `StringBuilder str = sb.get();
 for ( Object n : args ) {
-  str.append(',');
```
kgrgreer: Why remove?

src/foam/nanos/logger/LogsView.js
```diff
@@ -0,0 +1,231 @@
...
+      name: 'data',
+      factory: function() {
+        data = this.logMessageDAO.orderBy(this.DESC(this.LogMessage.CREATED));
+        return data;
+      }
+    },
+    {
+      name: 'filteredLogMessageDAO',
+      expression: function(data, filter ) {
+        if (filter){
```
kgrgreer: spacing:
  if ( filter ) {

src/foam/nanos/logger/LogsView.js
```diff
@@ -0,0 +1,231 @@
...
+        ]
+      }
+    }
+  ],
+
+
+  methods: [
+    function initE() {
+      this.SUPER();
+      var self = this;
```
kgrgreer: This isn't used.

kgrgreer: By which, I mean 'self' isn't used.

src/foam/nanos/logger/AbstractLogger.js
jlhughes: @deejes this should either be moved to after str.append(formatArg(n)) and/or use a 'for index' loop and only append it if there are more args.

kgrgreer: @jlhughes I think what you're saying didn't used to be required because the timestamp was already output so a comma was needed before each argument no matter the position.

PR #1331, opened 2018-07-17
of property info on propertyInfo

zahnno:
Display of class on Reference, FObjectProperty & ENUM properties on the propertyInfo Table.


src/foam/doc/ModelBrowser.js
```diff
@@ -186,11 +186,12 @@ foam.CLASS({
 foam.CLASS({
   package: 'foam.doc',
   name: 'PropertyInfo',
-  documentation: "Table view model to display model in printable model browser's",
+  documentation: 'Table view model to display' +
```
kgrgreer, line 189: I thought the original version was easier to read.

zahnno: Are we following eslint coding standards? Regardless I agree.

PR #1334, opened 2018-07-17
Make countryDAO read-only

TW80000:
Also adds a test to make sure the countryDAO is read-only.


TW80000:
@zahnno @kgrgreer Please review.

kgrgreer:
Make authenticated not read-only.

PR #1336, opened 2018-07-17
Documentation on model properties.

zahnno:
Added documentation to the following models:
- History record
- Country
- Group
- User
- Menu
- Notification
- Address


src/foam/nanos/auth/User.js
```diff
@@ -289,14 +319,14 @@ foam.CLASS({
...
     {
-      class: 'DateTime',
+      class: 'Date',
       name: 'created',
-      documentation: 'Creation date'
+      documentation: 'Creation date.'
     },
     {
-      class: 'DateTime',
+      class: 'Date',
```
jlhughes: Why did you change from DateTime to Date?  These are more operational and I've been using DateTime for all CreatedAware, LastModifiedAware so when the properties are displayed we do see the time properties.

jlhughes (approved):
Just one question on User, but otherwise looks good.

src/foam/nanos/notification/Notification.js
```diff
@@ -20,33 +20,33 @@ foam.CLASS({
...
       name: 'read',
+      documentation: 'Determines if notification has been read.',
       visibility: foam.u2.Visibility.RO
     },
     {
       class: 'Long',
       name: 'id',
       visibility: foam.u2.Visibility.RO
     },
-    {
```
kgrgreer: Why remove template property?

src/foam/nanos/auth/User.js
kgrgreer: Should be DateTime.

kgrgreer:
Thanks Patrick. Try to avoid adding code changes in documentation PR's. Knowing that a PR is only documentation or formatting changes makes it easier to review and reduces the changes of unwanted code changes slipping through. Better to split into two PR's: one containing documentation and one containing code changes (at least when the PR is mostly about documentation).

src/foam/nanos/notification/Notification.js
jlhughes: I suspect this was a merge conflict resolution - do you remember?

zahnno:
@kgr makes sense. I'll look out for those changes.

jlhughes (approved):

PR #1338, opened 2018-07-17
Fixing sudo for scripts

shreymahendru:
Added `x` as an arg for `runTest`, for using `sudo` on the modelled java tests.
Explicitly specifying the return type and the `String user` for the bean shell wrapper of `sudo`, because it does not recognize the correct overridden `sudo` method in `foam.util.Auth`.

### DIFF
--- a/src/foam/nanos/script/Script.js
+++ b/src/foam/nanos/script/Script.js
@@ -52,7 +52,10 @@ foam.CLASS({
       class: 'Boolean',
       name: 'enabled',
       tableCellFormatter: function(value) {
-        this.start().style({color: value ? 'green' : 'gray'}).add(value ? 'Y' : 'N').end();
+        this.start()
+          .style({ color: value ? 'green' : 'gray' })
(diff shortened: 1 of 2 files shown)


src/foam/nanos/script/Script.js
```diff
@@ -126,7 +137,7 @@ foam.CLASS({
           shell.set("currentScript", this);
           shell.set("x", x);
           shell.eval("runScript(String name) { script = x.get(\\"scriptDAO\\").find(name); if ( script != null ) eval(script.code); }");
-          shell.eval("sudo(String user) { foam.util.Auth.sudo(x, user); }");
+          shell.eval("foam.core.X sudo(String user) { foam.util.Auth.sudo(x, (String) user); }");
```
kgrgreer, line 140: Don't you need a return statement here?

shreymahendru: No that is implied. It automatically returns the value of the the last function call. Which would be the return value of `foam.util.Auth.sudo`

PR #1339, opened 2018-07-18
Convert RepeatLogger to DAODecorator and decorate logMessageDAO

deejes:


src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -8,17 +8,16 @@
 foam.CLASS({
   package: 'foam.nanos.logger',
   name: 'RepeatLogger',
```
jlhughes: please rename to RepeatLogMessageDAO

src/foam/nanos/logger/RepeatLogger.js
```diff
@@ -34,178 +33,37 @@ foam.CLASS({
...
+      if ( log.getSeverity().equals(getLastLogLevel()) ) {        
+        if ( log.getMessage().equals( getLastLogMessage()) ) {
+          setRepeatCount( getRepeatCount() + 1 );       
+          return null;
+        } else {
+          if ( getRepeatCount() > 1 ) {    
+            LogMessage repeatLog = new foam.nanos.logger.LogMessage();
+            repeatLog.setSeverity( getLastLogLevel() );
+            repeatLog.setMessage("The last log was repeated " + getRepeatCount() + " times");
+            super.put_(x,repeatLog);
```
jlhughes: does this fall through and call the super.put below?

jlhughes (approved):

deejes:
So as it stands right now, stdOutLogger and FileLogger are upstream of the RepeatLogMessageDAO, which means that they will continue to write out duplicates. Is that behaviour acceptable? If not we might want to move them lower down in the chain...

Also I've noticed a bug in the gui view of the logs, where the repeat message is not showing up directly after the repeated log message. Would it make sense to include a snippet of the repeated message in the log that states how many times it was repeated? Will investigate the ordering issue regardless.

Issue #1340, opened 2018-07-19
Add authentication to groupDAO


TW80000:
`groupDAO` must be authenticated so users are not allowed to give themselves new permissions.

It will need custom authentication so that users will be allowed to create new groups, but only give those groups permissions that they already have.

TW80000:
Closed by https://github.com/foam-framework/foam2/pull/1615

PR #1341, opened 2018-07-19
Add basic authentication to groupDAO

TW80000:
Custom authentication is needed in the future, but this will be better
than no authentication in the meantime.

Related to https://github.com/foam-framework/foam2/issues/1340 but does not close it.


TW80000:
@kgrgreer Please review.

PR #1344, opened 2018-07-19
Add tests for AuthenticatedDAO

TW80000:
* Adds tests for AuthenticatedDAO
* Adds a `TestUtils` class with testing helper methods
* Adds a new context that throws errors when it's used. The point of creating this and using it in AuthenticatedDAO is to prevent users from using the non-context-oriented methods, which they shouldn't be allowed to use for a DAO decorated by AuthenticatedDAO. This way, if a user tries to use a non-context-oriented method, an exception with a helpful message will be thrown.


TW80000:
@kgrgreer Please review.

src/foam/core/InvalidX.java
```diff
@@ -0,0 +1,46 @@
...
+  public X putFactory(Object key, XFactory factory) {
+    throw getException();
+  }
+
+  @Override
+  public String toString() {
+    throw getException();
+  }
+
+  private AccessControlException getException() {
```
kgrgreer: I would make the message an argument which would make this class more general/useful.

src/foam/dao/AuthenticatedDAO.java
```diff
@@ -1,17 +1,20 @@
 /**
  * @license
- * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
kgrgreer: According to Google lawyers, you just leave the copyright as the year the file was first created.

src/foam/test/TestUtils.java
```diff
@@ -0,0 +1,107 @@
...
+public class TestUtils {
+
+  /**
+   * Mock out a DAO in the given context by replacing it with an undecorated, empty MDAO.
+   * Will keep the same 'of' type.
+   * @param x The context you're replacing the DAO in.
+   * @param daoName The name of the DAO in the context you'd like to mock out.
+   * @return A subcontext with the given DAO replaced with an MDAO of the same type.
+   */
+  public static X mockDAO(X x, String daoName) {
```
kgrgreer, line 35: Interesting.

src/foam/test/TestUtils.java
```diff
@@ -0,0 +1,107 @@
...
+    user.setLastName("Smith");
+    user.setBusinessName("Widgets International");
+    user.setEmail("john@example.com");
+    File profilePicFile = new File();
+    profilePicFile.setFilename("Profile picture");
+    user.setProfilePicture(profilePicFile);
+    Address businessAddress = new Address();
+    businessAddress.setCity("Toronto");
+    user.setBusinessAddress(businessAddress);
+    Phone businessPhone = new Phone();
```
kgrgreer: Business Phone (and possibly other properties as well) isn't part of the standard User model, it's refined in in net.nanopay.auth.Model. This means that this code wouldn't compile for anyone other than nanopay.

TW80000:
Comments addressed. Please review again.

kgrgreer (approved):

PR #1345, opened 2018-07-19
Test enhancements

shreymahendru:
New SmartTest.
Sudo available to all java scripts/cron/test now as a function.
Script/cron/test are searchable by id, also tableCellFormatters for max-width to cell columns.

### DIFF
--- a/src/foam/nanos/script/Script.js
+++ b/src/foam/nanos/script/Script.js
@@ -125,6 +125,19 @@ foam.CLASS({
 
   methods: [
     {
+      name: 'sudo',
+      args: [
+        {
+          name: 'x', javaType: 'foam.core.X',
+        },
+        {
+          name: 'emailAddress', javaType: 'String'
+        }
+      ],
+      javaReturns: 'foam.core.X',
+      javaCode: 'return foam.util.Auth.sudo(x, emailAddress);'
+    },
+    {
       name: 'createInterpreter',
       args: [
         { name: 'x', javaType: 'foam.core.X' }
--- a/src/foam/nanos/script/Script.js
+++ b/src/foam/nanos/script/Script.js
@@ -8,7 +8,7 @@ foam.CLASS({
   package: 'foam.nanos.script',
   name: 'Script',
 
-  implements: [ 'foam.nanos.auth.EnabledAware' ],
+  implements: ['foam.nanos.auth.EnabledAware'],
 
   requires: [
     'foam.nanos.script.ScriptStatus',
@@ -41,12 +41,22 @@ foam.CLASS({
     'id', 'server', 'description', 'lastDuration', 'status', 'run'
   ],
 
-  searchColumns: [],
+  searchColumns: [ 'id' ],
 
   properties: [
     {
       class: 'String',
-      name: 'id'
+      name: 'id',
+      tableCellFormatter: function(value) {
+        this.start()
+          .style({
+            'overflow': 'hidden',
+            'max-width': '35ch',
+            'min-width': '35ch',
+            'text-overflow': 'ellipsis'
+          }).add(value)
+          .end();
+      }
     },
     {
       class: 'Boolean',
@@ -62,7 +72,16 @@ foam.CLASS({
     {
       class: 'String',
       name: 'description',
-      displayWidth: 80
+      tableCellFormatter: function(value) {
+        this.start()
+          .style({
+            'overflow': 'hidden',
+            'max-width': '55ch',
+            'min-width': '55ch',
+            'text-overflow': 'ellipsis'
+          }).add(value)
+        .end();
+      }
     },
     {
       class: 'DateTime',
@@ -72,7 +91,17 @@ foam.CLASS({
     {
       class: 'Long',
       name: 'lastDuration',
-      visibility: 'RO'
+      visibility: 'RO',
+      tableCellFormatter: function(value) {
+        this.start()
+          .style({
+            'overflow': 'hidden',
+            'max-width': '5ch',
+            'min-width': '5ch',
+            'text-overflow': 'ellipsis'
+          }).add(value)
+          .end();
+      }
     },
     /*
     {
--- a/src/foam/nanos/test/SmartTest.java
+++ b/src/foam/nanos/test/SmartTest.java
@@ -1,0 +1,112 @@
+package foam.nanos.test;
+
+import foam.core.X;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+
+import static java.lang.System.err;
+
+public class SmartTest extends Test {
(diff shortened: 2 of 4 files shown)


src/foam/nanos/test/Test.js
```diff
@@ -51,7 +51,35 @@ foam.CLASS({
...
+      name: 'description',
+      tableCellFormatter: function(value) {
+        this.start()
+          .style({
+            'overflow': 'hidden',
+            'max-width': '25ch',
+            'min-width': '25ch',
+            'text-overflow': 'ellipsis'
+          }).add(value)
+          .end();
```
kgrgreer, line 80: end should line up with start

kgrgreer:
Include a test case with this PR.

kgrgreer:
Normally in a context-oriented system the setup will involve creating a special sub-context which may mock or decorate certain services for the purposes of the test. For this to work, the setup method would need to return a context that would then subsequently be used by the test and cleanup methods.

src/foam/nanos/test/SmartTest.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Map;
+
+import static java.lang.System.err;
+
+public class SmartTest extends Test {
+  public static final String TEST_METHOD_SUFFIX = "Test";
+  public static final String CLEANUP_METHOD_SUFFIX = "Cleanup";
+  public static final String SETUP_METHOD_SUFFIX = "Setup";
+
+  private Method globalSetupMethod;
```
kgrgreer, line 18: Generally, I prefer to use protected over private as it makes things easier to test and debug.

src/foam/nanos/test/SmartTest.java
```diff
@@ -0,0 +1,145 @@
...
+  private Method globalCleanupMethod;
+  private Method preTestMethod;
+  private Method postTestMethod;
+  private final Map<String, TestChain> testChains = new HashMap<>();
+
+  protected X x_ = EmptyX.instance();
+
+  public X getX() {
```
kgrgreer, line 26: What is getX() and setX() used for? I believe that Test already implements ContextAware and has these methods, but why is it needed in the first place? Having two contexts: one passed as an argument and one from getX() opens the possibility of using the wrong one.

src/foam/nanos/test/SmartTest.java
```diff
@@ -0,0 +1,145 @@
...
+    setX(x);
+    prepareTestChains();
+    runTestChains();
+  }
+
+  private void runTestChains() {
+    if ( invokeMethod(globalSetupMethod) ) {
+      for ( String test: testChains.keySet() ) {
+        TestChain testChain = testChains.get(test);
+        runTestChain(testChain);
```
kgrgreer, line 45: Your choice, but I think the code would be a little cleaner if runTestChain() were just a run() method on the test Chain itself.

src/foam/nanos/test/SmartTest.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  private void runTestChain(TestChain testChain) {
+
+    // don't run when the test Chain if doesn't have a test method
+    if ( testChain.test == null ) {
+      return;
+    }
+    if ( invokeMethod(preTestMethod) ) {
+      if ( invokeMethod(testChain.setup) )  {
```
kgrgreer, line 58: I think the setup method should return a Context for use by test and cleanup.

src/foam/nanos/test/SmartTest.java
```diff
@@ -0,0 +1,145 @@
...
+    if ( invokeMethod(globalSetupMethod) ) {
+      for ( String test: testChains.keySet() ) {
+        TestChain testChain = testChains.get(test);
+        runTestChain(testChain);
+      }
+      invokeMethod(globalCleanupMethod);
+    }
+  }
+
+  private void runTestChain(TestChain testChain) {
```
kgrgreer, line 51: It would be nice if the current test name was recorded and used to prefix test(boolean,message) output so that you could easily test what test was producing test messages.

kgrgreer:
Nice job. Added some suggestions. Also, a few more comments would be useful.

src/foam/nanos/script/Script.js
```diff
@@ -41,12 +41,22 @@ foam.CLASS({
     'id', 'server', 'description', 'lastDuration', 'status', 'run'
   ],
 
-  searchColumns: [],
+  searchColumns: [ 'id' ],
```
jlhughes, line 44: add **description** to the search columns please.

src/foam/nanos/test/SmartTest.java
jlhughes: yes, if the map was moved into TestChain.

jlhughes (commented):
re: Kevin's comment about documentation, perhaps add a section or example on the use of the PREFIXes and such.

src/foam/nanos/test/SmartTest.java
shreymahendru: ok I'll user protected.

src/foam/nanos/test/SmartTest.java
shreymahendru: I used getX() and setX() to expose the X SmartTest gets from the runTest method. The X that I get from the parent's getter and setter (ContextAware) was giving problems when using sudo. This X returns every service as well as user as a null. Eg: userDAO. 

Also all the other ways to test eg GUI/beanshell test, Modelled test use the x that is passed to the runTest which is different from the X returned from the parents' getX() method.

src/foam/nanos/test/SmartTest.java
shreymahendru: ~~True, that would be cleaner~~

It won't be a good choice cause the ``invokeMethod`` would be used by both, SmartTest to run global ``cleanup`` and ``setup`` as well as ``run`` if I decide to have that in my TestChain Class, Unless I have a util class that has that method or make ``invokeMethod`` static in SmartTest.

src/foam/nanos/test/SmartTest.java
shreymahendru: shouldn't this be the responsibility of the person writing the tests? Like saving the context created in the setup as a class variable, and using that in the cleanup after the test is ran? That would be more intuitive for a person who is writing the test, in my opinion.

kgrgreer: The setup for each test would be responsible for building a sub-context. If this were stored in an instance variable then you have a lot of negative potential for tests not resetting the context or inadvertently running in subcontexts of other tests. When the setup and the test are defined in the same method subcontexts can just be local variables which can't escape the test but now that setup and test are different methods we need a method for passing along the sub-context in an organized way which isn't going to mess things up. Also, it occurs to me that many tests may want to create multiple sub-contexts, in which case having a separate setup method is going to be problematic.

jlhughes:
Closing - abandoned.

PR #1347, opened 2018-07-23
Map equal casting error

xuerongCode:
https://github.com/nanoPayinc/NANOPAY/issues/2757
throw exception when value of map is not comparable

### DIFF
--- a/src/foam/core/AbstractMapPropertyInfo.java
+++ b/src/foam/core/AbstractMapPropertyInfo.java
@@ -28,12 +28,15 @@ public abstract class AbstractMapPropertyInfo
     Object v1;
     Object v2;
     for ( Object key : m1.keySet() ) {
-      v1 = m2.get(key);
+      v1 = m1.get(key);
       v2 = m2.get(key);
       if ( v1 == null && v2 == null ) continue;
       if ( v2 == null ) return 1;
       if ( v1 == null ) return -1;
-      result = ((Comparable) v1).compareTo(v2);
+      if ( v1 instanceof Comparable) 
+        result = ((Comparable) v1).compareTo(v2);
+      else 
+        result = v1.equals(v2) ? 0 : -1;
       if ( result != 0 ) return result;
     }
     return 0;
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

src/foam/core/AbstractMapPropertyInfo.java
```diff
@@ -28,12 +28,17 @@ public int compare(Object o1, Object o2) {
...
       if ( v1 == null && v2 == null ) continue;
       if ( v2 == null ) return 1;
       if ( v1 == null ) return -1;
-      result = ((Comparable) v1).compareTo(v2);
+      if ( v1 instanceof Comparable) 
+        result = ((Comparable) v1).compareTo(v2);
+      else
+        //compare if refer to the same object
+        //will help when value is a static value
+        result = v1.equals(v2) ? 0 : -1;
```
kgrgreer, line 41: What class implements equals but is not comparable? Could all of this code just be replaced with a call to SafetyUtil.compare()?

xuerongCode: {
      name: 'props_',
      class: 'Map',
      javaType: 'java.util.Map',
      hidden: true,
      transient: true,
      factory: function() {
        var map = {};
        var ps = this.cls_.getAxiomsByClass(foam.core.Property);
        for ( var i = 0; i < ps.length; i++ ) {
          map[ps[i].name] = ps;
        }
        return map;
      },
      javaType: 'java.util.Map',
      javaFactory: `
  java.util.List<foam.core.PropertyInfo> properties = (java.util.List<foam.core.PropertyInfo>) this.getClassInfo().getAxiomsByClass(foam.core.PropertyInfo.class);
  java.util.Map<String, foam.core.PropertyInfo> map = new java.util.HashMap<String, foam.core.PropertyInfo>();
  for ( foam.core.PropertyInfo prop : properties ) {
    if ( ! "props_".equals(prop.getName()) ) {
      map.put(prop.getName(), prop);
    }
  }
  return map;
`
    }
In DefaultHttpParameters
SafetyUtil.compare() can not compare Map. It will also throw Exception if it is not comparable

PR #1348, opened 2018-07-23
Random fixes in preparation for build tool.

mcarcaso:
- Naming some refinements
- Adding some requires
- Splitting up the net stuff into a node and web script
- Wrapping some more code with foam.SCRIPT


src/lib/netNode.js
```diff
@@ -0,0 +1,44 @@
+/**
```
kgrgreer, line 1: Wrong license.

mcarcaso: Fixed

src/foam/core/stdlib.js
```diff
@@ -756,7 +756,7 @@ foam.LIB({
     function is(a, b) { return a === b; },
     function isInstance(o) {
       return typeof o === 'object' && ! Array.isArray(o) &&
-          ! foam.core.FObject.isInstance(o);
+          ! foam.core.FObject.isInstance(o) && ! foam.Null.isInstance(o);
```
adamvy: Let's do this in a different PR :)

mcarcaso: Done

PR #1351, opened 2018-07-23
check email duplication for existing users

deejes:


src/foam/nanos/auth/PreventDuplicateEmailDAO.js
```diff
@@ -38,6 +38,7 @@ if ( newUser ) {
   if ( ! Email.isValid(user.getEmail()) ) {
     throw new RuntimeException("Invalid Email");
   }
+}
 
   Count count = new Count();
```
kgrgreer: Fix the indentation of this block. (move left)

PR #1355, opened 2018-07-23
Override mmethod's toString

mcarcaso:
This is for the build tool because it needs to write out models that use this method. argsString also needed to be updated to not choke when args can't be figured out.


src/foam/core/stdlib.js
```diff
@@ -276,8 +276,8 @@ foam.LIB({
           match = str.match(/^(async )?function(\s+[_$\w]+|\s*)\((.*?)\)/);
 
       if ( ! match ) {
-        /* istanbul ignore next */
-        throw new TypeError("foam.Function.argsStr could not parse input function:\n" + ( f ? f.toString() : 'undefined' ) );
+        console.warn("foam.Function.argsStr could not parse input function:\n" + ( f ? f.toString() : 'undefined' ) );
```
adamvy: I think it would be better to preserve the function and catch+log the warning at usage spots like AbstractMethod.args.factory

src/foam/core/stdlib.js
```diff
@@ -863,6 +863,13 @@ foam.LIB({
         }
         return ( type[uid] || opt_defaultMethod ).apply(this, arguments);
       };
+      var toString = f.toString.bind(f);
```
kgrgreer: Add some comments to explain why this is being done.

src/foam/core/stdlib.js
mcarcaso: Good call. Done.

src/foam/core/stdlib.js
mcarcaso: Done. Does the comment make sense?

Also, what do you think of the fallback for foam.json.stringify not existing? I'm torn on whether to send the original toString's result or just returning something like 'function() {}'. What do you think?

src/foam/core/stdlib.js
mcarcaso: Ah missed a spot where this will break in debug type checking. Fixed there too.

src/foam/core/Method.js
```diff
@@ -64,7 +64,11 @@ foam.CLASS({
     'returns',
     {
       name: 'args',
-      factory: function() { return this.code ? foam.Function.args(this.code) : []; }
+      factory: function() {
+        if ( this.code )
+          try { return foam.Function.args(this.code) } catch(_) {}
```
adamvy: Should probably log a warning in the catch() block.

src/foam/core/debug.js
```diff
@@ -399,7 +399,12 @@ foam.LIB({
       if ( fn.isTypeChecker__ ) return fn;
 
       // parse out the arguments and their types
-      var args = foam.Function.args(fn);
+      try {
+        var args = foam.Function.args(fn);
+      } catch (_) {
+        // Could not parse args so don't bother decorating.
+        return fn;
```
adamvy: Ditto, should probably log a warning

PR #1357, opened 2018-07-23
Fixed size dao - UI improvement

deejes:
Limit text displayed in summaryView, and create a detailedView for LogMessages


deejes:
Might it make sense to separate the view files into a UI folder within logger at this point?

PR #1360, opened 2018-07-24
Improve JavaScript test support

TW80000:
* JS implementation of `test` method
* JS implementation of `print` method
* Reset number of passed/failed tests on each run
* Track duration of test execution
* Track last executed time

### DIFF
--- a/src/foam/nanos/script/Script.js
+++ b/src/foam/nanos/script/Script.js
@@ -227,7 +227,7 @@ foam.CLASS({
 
           with ( { log: log, print: log, x: self.__context__ } ) {
             this.status = this.ScriptStatus.RUNNING;
-            var ret = eval(this.code);
+            this.runScript(x);
             var self = this;
             Promise.resolve(ret).then(function() {
               self.status = self.ScriptStatus.UNSCHEDULED;
--- a/src/foam/nanos/script/Script.js
+++ b/src/foam/nanos/script/Script.js
@@ -145,6 +145,13 @@ foam.CLASS({
     },
     {
       name: 'runScript',
+      code: function() {
+        var log = () => {
+          this.output += Array.from(arguments).join('') + '\n';
+        };
+        with ( { log: log, print: log, x: self.__context__ } )
+          return Promise.resolve(eval(this.code));
+      },
       args: [
         {
           name: 'x', javaType: 'foam.core.X'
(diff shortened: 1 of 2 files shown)


TW80000:
@kgrgreer @shreymahendru Please review.

shreymahendru:
@TW80000 The runScript/runTest method should be wrapped in a promise, or it should return a promise where ``eval`` is wrapped in a ``Promise.resolve``. This is because when a user is writing a script/test, it could have a Promise. Eg: a select/find on a DAO, your scripts' output should be the response to that promise and also once the promise is resolved, update the script's status and output.
And wrapping synchronous code in a promise has no effect, so it would work the same for when the user doesn't have a promise in their script.

TW80000:
> This is because when a user is writing a script/test, it could have a Promise. Eg: a select/find on a DAO, your scripts' output should be the response to that promise and also once the promise is resolved, update the script's status and output.

~~It already works that way as is. I can include asynchronous code in a test that calls the `test()` method in a callback and it will update `passed` and `failed` when the test completes and show the output as well.~~

I see what you mean now. Will fix.

src/foam/nanos/script/Script.js
```diff
@@ -227,7 +227,7 @@ foam.CLASS({
 
           with ( { log: log, print: log, x: self.__context__ } ) {
             this.status = this.ScriptStatus.RUNNING;
-            var ret = eval(this.code);
+            this.runScript(x);
```
shreymahendru: If you are calling ``this.runScript`` here you need to have a js ``runScript`` function in Script.js, which would have an ``eval``

src/foam/nanos/test/Test.js
```diff
@@ -88,6 +99,9 @@ foam.CLASS({
     },
     {
       name: 'print',
+      code: function(message) {
+        this.output += message + '\n';
```
shreymahendru: This ``print`` method should be in Script.js

src/foam/nanos/test/Test.js
```diff
@@ -100,6 +114,25 @@ foam.CLASS({
...
+          this.runTest(x);
+        } catch (err) {
+          this.failed += 1;
+          this.output += err;
+        }
+
+        var endTime = Date.now();
+        var duration = endTime - startTime; // Unit: milliseconds
+
+        this.lastRun = new Date();
```
shreymahendru: You need to do a ``put`` to the dao once the script has finished running.

PR #1364, opened 2018-07-24
Refactor emails to use a DAO

jlhughes:
Presently just duplicates existing services. The SMTPEmailMessageDAO calls the smtpEmailService.
NOTE: the smtp configuration is on the nanopay side.


kirkeaton (approved):

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -0,0 +1,123 @@
...
+  ],
+
+  axioms: [
+    {
+      name: 'javaExtras',
+      buildJavaClass: function(cls) {
+        cls.extras.push(foam.java.Code.create({
+          data:
+          `
+            protected ThreadLocal<Boolean> initializing = new ThreadLocal<Boolean>() {
```
kgrgreer, line 54: What is this used for?

jlhughes: Now that it's all complete, it's not required. It was modelled after DAOLogMessage, but was replacing the logger which we only wanted to do once.

jlhughes: From an admin point of view it's nice having only one initialization sequence in the logs or output, and once it's run it's replaced so there is no overhead, so I'm going to leave it.

PR #1367, opened 2018-07-25
Fix JavaScript test and script context

TW80000:
My mistake from a PR yesterday.


TW80000:
@kgrgreer please review.

PR #1369, opened 2018-07-25
Split up the createChildMethod_ refinement.

mcarcaso:
Java and swift need refinements on createChildMethod_ to merge method info from
parent. There's a subtle difference between swift and java when checking
hasOwnProperty. It's up for debate which platform's behaviour is more
correct and once we've decided on that, we should try to get the two
platforms to match.


src/foam/swift/refines/Method.js
```diff
@@ -293,3 +293,31 @@ return MInfo(classInfo())
...
+  package: 'foam.swift.refines',
+  name: 'CreateChildRefines',
+  flags: ['swift'],
+  methods: [
+    function createChildMethod_(child) {
+      var result = child.clone();
+      var props = child.cls_.getAxiomsByClass(foam.core.Property);
+      for ( var i = 0 ; i < props.length ; i++ ) {
+        var prop = props[i];
+        if ( this.hasOwnProperty(prop.name) && ! child.hasOwnProperty(prop.name) ) {
```
mcarcaso, line 308: I think this is the more correct behaviour because, without the this.hasOwnProperty check, we'll be setting javaCode on children whose parents don't have it set. Removing this check breaks java because there's a few models out there that don't implement all methods on the interface and if javaCode isn't set, the method doesn't get generated and the compiler complains.

kgrgreer:
Will review.

mcarcaso:
Ping?

src/foam/swift/refines/Method.js
mcarcaso: The reason java breaks is because there were some models without javaCode on all methods of the interface they implement. An alternative to this PR is here: https://github.com/foam-framework/foam2/pull/1407

mcarcaso:
Gonna close this in favour of https://github.com/foam-framework/foam2/pull/1407

Issue #1352, opened 2018-07-23
Highlight inherited permissions in permissions matrix


TW80000:
Implied permissions are currently given a green border, but permissions inherited from a parent group aren't given any visual indication that they are inherited.

For example: If my group inherits from basicUser, the permissions I inherit from basicUser should be visually highlighted in the GUI in my group's column.

kgrgreer:
@minsunny

kgrgreer:
Design: show inherited permissions with a green border and then add a tooltip to where implied permissions are derived from. ie. basicUser:service.*

PR #1377, opened 2018-07-26
Added Highlight inherited permissions

minsunny:
https://github.com/foam-framework/foam2/issues/1352#event-1748292327

1. changed parent type string to reference in Group model
2. added highlights inherited permission and made it updated along parent


src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -94,7 +97,17 @@ foam.CLASS({
                               cb.data$.sub(function() { self.updateGroup(p, g, cb.data); });
                               this.start('td').show(groups[g.id].data$.map(function() {return groups[g.id].data;}))
                                     .style({'text-align': 'center', 'width': '100'}).tag(cb).call(function() {
-                                      if ( g.implies(p.id) ) { cb.style({'border-color': '#40C75B'}) };
+                                      self.groupDAO.find(g.parent).then(function(a){
```
kgrgreer: Space between ) and {.

kgrgreer: You can do g.parent$find(function(a) { , which is shorter.

src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -94,7 +97,17 @@ foam.CLASS({
...
+                                        if ( a != undefined && a.implies(p.id) ) {
+                                          cb.setAttribute('title', g.parent + ': ' + p.id); //'Permission inherited from ' + g.parent
+                                          cb.style({'border-color': '#40C75B'})
+                                        }
+                                      });
+
+                                      if ( g.implies(p.id) ) {
+                                        cb.setAttribute('title', g.id + ': ' + p.id);  //'Permissions granted from ' + g.id
+                                        cb.style({'border-color': '#40C75B'});
+                                      }
```
kgrgreer: Move the parent group code above into the else block of this 'if' to avoid extra work and avoid a race condition.

src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -116,17 +129,33 @@ foam.CLASS({
...
+      var e = foam.mlang.Expressions.create();
+
       dao.find(group).then(function(group) {
         // Remove permission if found
         var permissions = group.permissions.filter(function(p) {
           return p.id != permission.id;
         });
 
+        //parents' permissions
+        dao.find(group.parent).then(function(groupParent) {
```
kgrgreer: group.parent$find(function(groupParent) {

src/foam/nanos/notification/email/DAOResourceLoader.java
```diff
@@ -43,7 +43,7 @@ public static EmailTemplate findTemplate(X x, String name, String groupId) {
       }
 
       Group group = (Group) groupDAO.find(groupId);
-      groupId = ( group != null && ! SafetyUtil.isEmpty(group.getParent()) ) ? group.getParent() : "*";
+      groupId = ( group != null && ! SafetyUtil.isEmpty(group.getParent().toString()) ) ? group.getParent().toString() : "*";
```
kgrgreer: Why is this change needed?

src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -92,9 +95,19 @@ foam.CLASS({
...
                                     .style({'text-align': 'center', 'width': '100'}).tag(cb).call(function() {
-                                      if ( g.implies(p.id) ) { cb.style({'border-color': '#40C75B'}) };
+                                      self.groupDAO.find(g.parent).then(function(a) {
+                                        if ( a != undefined && a.implies(p.id) ) {
+                                          cb.setAttribute('title', g.parent + ': ' + p.id);
+                                          cb.style({'border-color': '#40C75B'})
+                                        }
+                                      });
+
+                                      if ( g.implies(p.id) ) {
```
kgrgreer: If g.implies(p.id) is true then you don't need to do the self.groupDAO.find() above, so move that code into an 'else' block of this if statement.

src/foam/nanos/notification/email/DAOResourceLoader.java
minsunny: getParent() already returns a String so the toString() isn't needed.

PR #1383, opened 2018-07-26
Fix ProxyListener in JS and Swift

### LINKED ISSUE #1381: Chains of ProxyDAOs do not receive reset events as expected.
If you have ProxyDAO1 -> ProxyDAO2 -> SomeDAO and call .listen() on ProxyDAO1, you will not receive notifications when the delegate of ProxyDAO2 is modified.

This is due to how ProxyDAO.listen/ProxyListener were designed.  Fix incoming.

adamvy:
In ProxyDAO always decorate the listener in listen_ rather than listen() as we had before.

The previous behaviour would break if you had several ProxyDAOs in a chain and changed the delegate of one half way in the chain, listeners against the top proxydao would not get notified.

We need to install a proxy listener for every proxydao in the chain, not just the top one.

Fixes #1381

### DIFF
--- a/src/foam/dao/daoUtils.js
+++ b/src/foam/dao/daoUtils.js
@@ -58,16 +58,13 @@ if let oldValue = oldValue as? foam_dao_AbstractDAO {
 
   methods: [
     {
-      name: 'listen',
-      code: function listen(sink) {
-        if ( ! foam.core.FObject.isInstance(sink) ) {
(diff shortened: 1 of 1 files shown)


adamvy:
@mcarcaso No idea if the swift code even compiles.

@kgrgreer in case you're curious.

mcarcaso (approved):

PR #1389, opened 2018-07-27
Fixed size array

deejes:
implement data structure synchronized on index counter to replace CircularFifoBuffer


src/foam/dao/FixedSizeDAO.js
```diff
@@ -46,24 +62,57 @@ foam.CLASS({
...
-Object[] arrayObject = getFixedSizeArray().toArray();
-for ( int i = 0; i < arrayObject.length; i++ ) {
-  decorated.put( arrayObject[i], null );
+
+Integer backCounter;
+
+if ( getNextIndex() == 0 ) {
+  backCounter = ( getInternalArraySize() - 1 ); 
+} else {
+  backCounter = ( getNextIndex() - 1 );
```
jlhughes, line 97: Check that backCounter is not -1 after this.

src/foam/dao/FixedSizeDAO.js
```diff
@@ -13,22 +13,38 @@ foam.CLASS({
...
+      name: 'nextIndex'
+    },
     {
       class: 'Int',
       name: 'fixedArraySize',
       value: 10000
     },
+    {
+      class: 'Int',
+      name: 'internalArraySize',
```
jlhughes: document the internal/external array and sizes and a bit about why the internal is created a bit larger.

jlhughes (changes requested):

src/foam/dao/FixedSizeDAO.js
```diff
@@ -13,22 +13,38 @@ foam.CLASS({
 
   javaImports: [
     'foam.dao.Sink',
-    'org.apache.commons.collections.buffer.CircularFifoBuffer',
-    'org.apache.commons.collections.Buffer',
-    'org.apache.commons.collections.BufferUtils'
+    // why am I using a ReentractLock ?
```
jlhughes: Remove. A reentrant lock is fine.

src/foam/dao/FixedSizeDAO.js
deejes: This section to prevent precisely that. 
It could only be -1 if nextIndex was zero, which I'm checking for. 
I don't think it should be necessary to do another explicit check...

jlhughes: This is outside the lock so it could be reset between the two calls.

deejes: Makes sense, made the change.

jlhughes (approved):

PR #1394, opened 2018-07-30
Fix NPE in SendEmailNotificationDAO

TW80000:
My first thought was that it might be better if the generated Java code returned an empty `List` by default instead of `null` for `StringArray` foam properties, but I wasn't sure how to do that.

This code is just a simple null check, but I'd happily change it to a more generally applicable fix if anyone else can confirm that would be a better fix and can point me where to make the change.

### DIFF
--- a/src/foam/nanos/notification/SendEmailNotificationDAO.java
+++ b/src/foam/nanos/notification/SendEmailNotificationDAO.java
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
 package foam.nanos.notification;
 
 import foam.core.FObject;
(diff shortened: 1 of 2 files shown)


TW80000:
@penzital @kgrgreer @eatonk please review.

jlhughes:
@TW80000 The properties in question in foam/nanos/auth/User have a default value of null. This could be changed to a function returning a new empty array.

src/foam/nanos/notification/SendEmailNotificationDAO.java
```diff
@@ -1,4 +1,5 @@
 package foam.nanos.notification;
```
kirkeaton, line 7: Add missing license header to this file please.

TW80000:
@jlhughes Yeah, that's what I'm not sure how to do. Do I add a `javaCode` to the `disabledTopicsEmail` property on User? Is it `javaFactory`? Is it `value`? Is it `default`? Should it go in `types.js` instead so it applies to all properties of type `StringArray`? I get the idea of the better fix, I'm just not sure how to actually write the code for it.

kirkeaton:
@TW80000 In refinements.js look for the StringArray refinement and add a javaFactory to the list of properties.
```
['javaFactory', 'return new String[0];']
```

TW80000:
@jlhughes @eatonk Updated.

kirkeaton (approved):

PR #1397, opened 2018-07-30
Rename to PipelineManagerNonRPC

mcarcaso:
I'm not sure if anything is actually using this but it having the same id as PipelineManager is an issue when crawling the dir for the build tool.


mdittmer (dismissed):
Doesn't look like [all references](https://github.com/foam-framework/foam2/search?q=PipelineManager&unscoped_q=PipelineManager) were updated. The only project that uses `foam.box.pipeline` that I'm aware of is [this one](https://github.com/GoogleChromeLabs/webidl-diff) which doesn't instantiate `PipelineManager` directly, so once all references are updated, this should be fine.

*Aside*: Why is this an issue? Shouldn't another class called `PipelineManager` that needs to reasonabout things by ID be using the package-qualified ID of the class you're renaming (and therefore not need a rename)?

mcarcaso:
The issue is that we have two models with the same ID. One lives in src/foam/box/pipeline/PipelineManager.js and the other lives in src/foam/box/pipeline/PipelineManagerNonRPC.js so I'm not sure what model is to be used when something requires foam.box.pipeline.PipelineManager. Since files.js has no reference to foam/box/pipeline/PipelineManagerNonRPC I'm assuming it's not actually being used anywhere.

mcarcaso:
Spoke with Mark offline about this and simply removing this model seems to be the way to go.

PR #1401, opened 2018-07-30
ClassReferenceParser

kirkeaton:
Added a parser to parse the following string:
```
{"class":"__Class__","forClass_":"foam.nanos.auth.User"}
```


src/foam/lib/json/ClassReferenceParser.java
```diff
@@ -42,38 +41,16 @@ public PStream parse(PStream ps, ParserContext x) {
...
-      Object axiom = info.getAxiomByName(propName);
-
-      if ( axiom == null ) {
-        System.err.println("Unknown Property Reference: " + classId + "." + propName);
-      }
-
-      return ps.setValue(axiom);
+      String classId = (String) x.get("forClass_");
+      Class cls = Class.forName(classId);
+      return cls != null ? ps.setValue(cls) : null;
```
mcarcaso, line 51: Should we be returning a ClassInfo here or just a class?

kgrgreer: ClassInfo I think. But it happens to work for our current use because INSTANCE_OF will take both.

kgrgreer: Which can be looked up like this:
    try {
      return INSTANCE_OF((ClassInfo) cls.getMethod("getOwnClassInfo").invoke(null));
    } catch(NoSuchMethodException|IllegalAccessException|InvocationTargetException e) {
      throw new RuntimeException("Attempt to call INSTANCE_OF on non Modelled class." + cls);
    }

mcarcaso: Yeah I imagine things might act up if we're trying to set a Class property to an actual java class rather than a ClassInfo because the javaType on the Class property is ClassInfo.

PR #1407, opened 2018-07-31
Move createChildMethod_ logic to core.

mcarcaso:
This is an alternative to https://github.com/foam-framework/foam2/pull/1369
This resolves the issue causing java to choke when checking hasOwnProperty in createChildMethod_ by flagging some models that don't have java implementations and implementing some methods in java as NOOPs.


src/foam/core/Method.js
```diff
@@ -177,3 +174,30 @@ foam.CLASS({
 
 
 foam.boot.phase2();
+
+foam.CLASS({
+  refines: 'foam.core.AbstractMethod',
+  package: 'foam.core',
+  name: 'CreateChildRefines',
```
adamvy: Why is this a refines here and not just a method when AbstractMethod is defined?

mcarcaso: I didn't get a chance to dig too deep into why, but before phase2, the child.clone() call is failing here: https://github.com/foam-framework/foam2/blob/a6039879d5f3f5c55070cb75ce47b3ffeda7bc7c/src/foam/core/FObject.js#L910

I'll look a little deeper into it when I get a chance.

mcarcaso: Ok looked into this a little further and the short story is that the FObjects aren't mature enough to do a clone yet before phase2 and that's needed by createChildMethod_.

While debugging, I noticed that https://github.com/foam-framework/foam2/blob/a6039879d5f3f5c55070cb75ce47b3ffeda7bc7c/src/foam/core/FObject.js#L910 was failing because cloneProperty didn't exist yet. During phase2, we upgrade all of the models that have already been registered and during foam.core.Property's upgrade, it tries to install methods on itself and it triggers createChildMethod_ on toString. Since foam.core.Property hasn't been upgraded, foam.core.Property.CLONE_PROPERTY isn't ready yet and we can't clone.

src/foam/core/Argument.js
```diff
@@ -318,3 +318,30 @@ foam.CLASS({
     }
   ]
 });
+
+foam.CLASS({
+  refines: 'foam.core.AbstractMethod',
+  package: 'foam.core',
+  name: 'CreateChildRefines',
```
kgrgreer, line 325: Can you add documentation: to explain what this does/is-for?

mcarcaso: Can do. Done.

PR #1408, opened 2018-08-01
Modify chained hashing

kirkeaton:
Currently we are chaining hashes by updating the hash function with the previous hash and the current message. This PR updates the functionality to hash the current message before chaining the hashes.

### DIFF
--- a/src/foam/core/AbstractFObject.java
+++ b/src/foam/core/AbstractFObject.java
@@ -171,6 +171,15 @@ public abstract class AbstractFObject
         prop.updateDigest(this, md);
       }
 
+      // no chaining so return digest
+      if ( hash == null || hash.length == 0 ) {
+        return md.digest();
+      }
+
+      // calculate digest, update with previous hash and current hash
+      byte[] digest = md.digest();
+      md.update(hash);
+      md.update(digest);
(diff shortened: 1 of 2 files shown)


src/foam/core/AbstractFObject.java
```diff
@@ -171,6 +166,15 @@ public boolean hasDefaultValue(String prop) {
...
       }
 
+      // no chaining so return digest
+      if ( hash == null || hash.length == 0 ) {
+        return md.digest();
+      }
+
+      // calculate digest, update with previous hash and current hash
+      byte[] digest = md.digest();
+      md.update(hash);
```
nickovs, line 176: This looks like it is using the same hash object for the hashing together of the old hash and new message digest. To compute H(H(old_tail), H(new_object)) you probably need a new message digest instance.

kirkeaton: According to the JavaDocs, calling `digest` resets the MessageDigest instance allowing it to be reused again.

nickovs: Fair enough. That's different to most of the hash libraries I see but if it works that way then you're good.

PR #1409, opened 2018-08-01
Fix Literal data parser

hchoura:


jlhughes (approved):
Looks sane.

src/foam/lib/query/YYYYMMDDLiteralDateParser.java
```diff
@@ -1,127 +1,129 @@
...
+//YYYY-MM
+//YYYY
+public class YYYYMMDDLiteralDateParser extends ProxyParser {
+
+  public YYYYMMDDLiteralDateParser() {
+    super(
+        new Alt(
+
+            //YYYY-MM-DDTHH:MM
+            new Seq(
```
jlhughes: can you fix the spacing, these don't line up with the Seq's below.

PR #1410, opened 2018-08-01
ReciprocalSearch improvements

TW80000:
## Changes
* Update UI
* Property filters are now hidden behind a checkbox by default
* Fix serialization for Object properties that get set to Date values (fixes dates being used in binary mlangs like Gt and Lt, which are used in DateSearchView)
* Clear button still clears each search view, but now also collapses all property search views

## New features
* Add a search view for Date properties
* Add a search view for Currency properties

## New UI
[Original designs](https://app.zeplin.io/project/58ed17e322a7ba2e76e55cde/screen/5a0f3cb00c6b67ef13d2f8b8)

screen shot 2018-08-01 at 1 54 49 pm


TW80000:
Closing because it's not quite ready yet.

TW80000:
@zahnno Please review. I will be applying this to ExpensesView and SalesView in a NANOPAY PR later as per https://github.com/nanoPayinc/NANOPAY/issues/2401.

@kgrgreer Please review, particularly changes to Date (de)serialization.

@jingyilai0809 Please review, particularly the UI.

jingyilai0809:
My notes
1. will need to think about the order of the filter items
2. can we change the background color to match the design? also the checkbox @TW80000 
image
3. Font family should match the design (for next time as it is a lot of work to change according to Troy)

TW80000:
To expand on @jingyilai0809's third point, currently a monospace font is being used in GroupBySearchView because we're currently using a bunch of non-breaking spaces to separate the name of an entry in the list from its count. Monospace fonts have fixed-width characters so they all line up. If we were to switch to a non-monospace font in that view - like the design calls for - the alignment would break. The way we do the formatting should be improved.

kgrgreer:
Can be done in a future PR, but the formatting under 'Status' doesn't match the rest of the views.

src/foam/lib/json/ObjectDateParser.java
```diff
@@ -0,0 +1,101 @@
...
+ * Now that we've proven the need for an unambiguous date parser, why wouldn't
+ * we just replace the original date parser with this one? Why have two? We do
+ * so in the name of performance. The original date parser uses less space,
+ * meaning less bandwidth is used and less time is spent serializing and
+ * deserializing.
+ */
+public class ObjectDateParser
+  extends ProxyParser
+{
+  protected static ThreadLocal<StringBuilder> sb = new ThreadLocal<StringBuilder>() {
```
kgrgreer: Is this StringBuilder used?

TW80000:
## Update

* General MQL search was added
* Input outline style changed from default to blue border

screen shot 2018-08-03 at 1 48 15 pm

## TODO

* Fix font used in GroupBy search views
* Fix status styling - @kgrgreer What do you mean by this?
* ~~Fix background colour~~
* ~~Fix checkbox colour~~

TW80000:
## Update

* Fixed background colour

screen shot 2018-08-03 at 2 09 08 pm

src/foam/u2/FloatView.js
```diff
@@ -51,22 +64,22 @@ foam.CLASS({
       this.SUPER(p);
 
       this.precision = p.precision;
-      this.min       = p.min;
-      this.max       = p.max;
+      this.min = p.min;
```
kgrgreer, line 67: Why messing up alignment?

src/foam/u2/FloatView.js
```diff
@@ -51,22 +64,22 @@ foam.CLASS({
...
+      this.max = p.max;
     },
 
     function formatNumber(val) {
       if ( ! val ) return '0';
       val = val.toFixed(this.precision);
       var i = val.length - 1;
-      for ( ; i > 0 && val.charAt(i) === '0' ; i-- ) {}
+      for ( ; i > 0 && val.charAt(i) === '0'; i -- ) {}
```
kgrgreer: The space between i and -- doesn't comply with the google/foam style guideline.

src/foam/u2/FloatView.js
TW80000: [Horizontal alignment is discouraged in the Google JavaScript Styleguide](https://google.github.io/styleguide/jsguide.html#formatting-horizontal-alignment). Our linter settings also flag it as an error.

src/foam/u2/search/CurrencySearchView.js
```diff
@@ -0,0 +1,152 @@
...
+      }
+    },
+    {
+      name: 'predicate',
+      documentation: `All SearchViews must have a predicate as required by the
+          SearchManager. The SearchManager will read this predicate and use it
+          to filter the dao being displayed in the view.`,
+      expression: function(qualifier, amount) {
+        amount = typeof amount === 'number' ?
+            Math.floor(amount * 100) :
```
kgrgreer: If this is the only thing that distinguishes this from a regular number searchView, couldn't this whole class be made reusable and have just the line abstracted out to a template method in a very small sub-clas?

src/foam/u2/FloatView.js
TW80000: You're right. Since we enforce a space between `!` and its argument as part of our rule exceptions, our linter flags all unary operators without a space between them and their arguments. I'll revert these spots back and have a quick look to see if the linter settings can be more fine-grained.

src/foam/u2/search/DateSearchView.js
```diff
@@ -0,0 +1,127 @@
...
+      documentation: `A date selectable by the user. Used as the second argument
+          to the predicate generated by this view.`,
+    },
+    {
+      name: 'predicate',
+      documentation: `All SearchViews must have a predicate as required by the
+          SearchManager. The SearchManager will read this predicate and use it
+          to filter the dao being displayed in the view.`,
+      expression: function(qualifier, date) {
+        if ( qualifier.length > 0 && date && ! isNaN(date.valueOf()) ) {
```
kgrgreer, line 47: If Gt it should Gt end-of-day. If Lt it should be Lt end-of-day. 
As mentioned in demos, Between would also be good.
Other options might include: 'Yesterday', 'Today', 'Tomorrow', 'This Week', 'This Month', 'Next Month', etc.

src/foam/u2/search/DateSearchView.js
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Clears the fields to their default values.
+     * Required on all SearchViews. Called by ReciprocalSearch.
+     */
+    function clear() {
+      this.qualifier = 'True';
+    }
+  ],
+
+  css: `
```
kgrgreer, line 82: I prefer to put css near the top with constants and messages since it's more of a class/static kind of thing. But your choice.

src/foam/u2/view/SearchViewWrapper.js
```diff
@@ -0,0 +1,120 @@
...
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'SearchViewWrapper',
+  extends: 'foam.u2.View',
+
+  documentation: `Wraps a SearchView with a checkbox to enable or disable its
+    predicate.`,
+
+  requires: [
+    'foam.mlang.predicate.True',
```
kgrgreer: No trailing ,

src/foam/u2/view/SearchViewWrapper.js
```diff
@@ -0,0 +1,120 @@
...
+      name: 'checkbox',
+      documentation: `A named reference to the checkbox in this view so that we
+        can subscribe to events that it publishes.`,
+    },
+    {
+      class: 'Boolean',
+      name: 'active',
+      documentation: `Tracks whether the property is being used as part of the
+        filter criteria or not.`,
+      value: false
```
kgrgreer: value: false is the default, so this line is redundant

src/foam/u2/search/DateSearchView.js
TW80000: I chose to put it at the bottom because CSS usually takes up a lot of space and I rarely need to look at it after it's written the first time. It's at the bottom to avoid having to scroll past it every time I want to add a require or check a property and then go back to the initE function. I'll just leave it at the bottom if it's up to me.

src/foam/u2/view/SearchViewWrapper.js
```diff
@@ -0,0 +1,120 @@
...
+          .addClass('section')
+          .addClass('form-element-container')
+          .show(this.active$)
+          .add(this.searchView)
+        .end();
+
+      this.checkbox.data$.sub(this.checkboxChanged);
+    },
+
+    function toggleActive() {
```
kgrgreer: This could just be put in the listener.

kgrgreer:
Good job. Added some comments.

src/foam/u2/search/CurrencySearchView.js
TW80000: Yeah, that makes sense. I'll split this into NumberSearchView and CurrencySearchView.

TW80000:
## Update

* Changed checkbox colour

screen shot 2018-08-03 at 3 52 43 pm

jingyilai0809:
@TW80000 
I really appreciate the to-do list, a title and a screenshot to review the design.
It makes things a lot easier to understand and provide lots of clarity.
Thank you so much for doing that!!!! 
@kgrgreer 
It will be great if we can adopt the same approach for every PR which requires a designer's review.

**Feedback:**
1. Checkbox is good
2. Background is good
3. The search field styling is matched

Good job! Thank you!

src/foam/u2/FloatView.js
kgrgreer: I've just updated the FOAM guide to officially reflect what has been the de-facto standard all along.

src/foam/u2/search/DateSearchView.js
kgrgreer: ok

src/foam/u2/search/FilterController.js
```diff
@@ -505,3 +505,123 @@ foam.CLASS({
...
+      name: 'searchView',
+      value: { class: 'foam.u2.search.FloatSearchView' }
+    }
+  ]
+});
+
+
+foam.CLASS({
+  package: 'foam.u2.search',
+  name: 'DoubleRefinement',
```
kgrgreer: Double extends Float so just refining Float is sufficient.

src/foam/u2/search/IntegerSearchView.js
```diff
@@ -0,0 +1,152 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.search',
+  name: 'IntegerSearchView',
+  extends: 'foam.u2.View',
```
kgrgreer: If you extend Controller instead of View then you will get the exports: 'as data' for free. Generally, a View is like an MVC View which is bound to some data, but in this case the GUI is its own data, which makes it more of a Controller.

src/foam/u2/search/DateSearchView.js
TW80000: This is more or less blocked by the big date/time refactoring that @mcarcaso proposed. I could implement this before the refactor, but then I'd need to go back and change it again afterwards.

kgrgreer: Okay, let's just do in a followup PR.

kgrgreer (approved):

PR #1411, opened 2018-08-01
Added Two-Factor authentication sign in view

kirkeaton:
1. Added Two-Factor authentication sign in view
2. Added logic to show two-factor sign in screen if user has two factor enabled
3. Added logic to throw an Exception if user has not passed two factor sign in


src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -69,6 +70,12 @@ public User getCurrentUser(X x) throws AuthenticationException {
       throw new AuthenticationException("User group disabled");
     }
 
+    // check for two-factor authentication
+    Object twoFactorSuccess = session.getContext().get("twoFactorSuccess");
```
kgrgreer: This is ugly, we should add a getBoolean() to X.

src/foam/nanos/auth/twofactor/TwoFactorSignInView.js
```diff
@@ -0,0 +1,136 @@
...
+      this
+        .addClass(this.myClass())
+        .start()
+          .start('h1').add('Two-Factor Authentication').end()
+          .start('form').addClass('tfa-container')
+            .start().addClass('label').add('Token').end()
+            .start(this.TWO_FACTOR_TOKEN).addClass('full-width-input').end()
+            .start(this.VERIFY).addClass('full-width-button').end()
+          .end()
+        .end()
```
kgrgreer: minor: missing ;

src/foam/nanos/auth/twofactor/TwoFactorSignInView.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+  ],
+
+  actions: [
+    {
+      name: 'verify',
+      label: 'Verify',
```
kgrgreer: label is redundant here since the default is to just capitalize the name.

PR #1412, opened 2018-08-02
Fix double menu

TW80000:


TW80000:
@kgrgreer Please review.

src/foam/u2/Element.js
```diff
@@ -731,6 +731,13 @@ foam.CLASS({
     }
   `,
 
+  messages: [
+    {
+      name: 'SelectBadUsage',
```
kgrgreer: SELECT_BAD_USAGE

PR #1413, opened 2018-08-02
Added SecurityUtil class with GetSecureRandom function

kirkeaton:
Added SecurityUtil class to collect common crypto functions
Added GetSecureRandom function which has the ability to reseed itself after too many usages.
Replaced SecureRandom.getInstance(String) method calls with SecurityUtil.GetSecureRandom calls

https://github.com/nanoPayinc/NANOPAY/issues/2994


daudich (approved):
LGTM

PR #1416, opened 2018-08-02
Give javaReturns a default value

mcarcaso:
Look at the value of `returns` to fill in a default for the `javaReturns`. This allows us to shorten something like:
```
returns: 'foam.foo.Bar',
javaReturns: 'foam.foo.Bar',
```
into just
```
returns: 'foam.foo.Bar',
```

### DIFF
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -443,7 +443,11 @@ foam.CLASS({
     },
     {
       class: 'String',
-      name: 'javaReturns'
+      name: 'javaReturns',
+      expression: function(returns) {
+        var of = foam.lookup(returns, true);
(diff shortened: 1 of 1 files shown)


adamvy:
Why the lookup step?  Why not just expression: function(returns) { return returns; }.  Presumable for core things that need to be changed to foam.core.Foo ?

I'd rather fix that in the generators or somewhere else, this to me risks not working if the type it returns isn't loaded yet.

mcarcaso:
Yeah figured I'd be able to catch anything that needs foam.core in front by making use of lookup simply returning `returns || ''` works too. Need the `|| ''` for the expression not to complain about returning undefined.

PR #1418, opened 2018-08-02
Fix up the way many to many relationships are added to models.

mcarcaso:
Now, install an axiom on the model which installs a property (js/swift only) and a method that takes the context.


adamvy:
Verify this with the many to many relationship test in the nanos test dao.

mcarcaso:
Oh jeez missed the OneToMany part of this. Closing for now...

PR #1419, opened 2018-08-02
Added FileServlet

kirkeaton:
Added FileServlet which outputs Files given a directory. Useful for outputting static images

### DIFF
--- a/src/foam/nanos/servlet/FileServlet.java
+++ b/src/foam/nanos/servlet/FileServlet.java
@@ -1,0 +1,74 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.servlet;
+
+import foam.util.SafetyUtil;
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.text.StringEscapeUtils;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FileServlet
+  extends HttpServlet
+{
+  protected static final int                 BUFFER_SIZE = 4096;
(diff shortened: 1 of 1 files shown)


src/foam/nanos/servlet/FileServlet.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Override
+  protected void service(HttpServletRequest req, HttpServletResponse resp)
+    throws ServletException, IOException
+  {
+    // get path
+    String cwd = System.getProperty("user.dir");
+    String path = getServletConfig().getInitParameter("path");
+    String reqPath = req.getRequestURI().replaceFirst("/?images/?", "/");
```
adamvy, line 54: If this is specific for images would be good to have a slightly less general name.

src/foam/nanos/servlet/FileServlet.java
```diff
@@ -0,0 +1,74 @@
...
+
+    // get file
+    File src = new File(cwd + "/" + path + reqPath);
+    if ( src.isFile() && src.canRead() ) {
```
adamvy, line 58: Let's add a security check to prevent someone using .. to escape out of the images directory.


I believe that src.getCanonicalPath().startsWith(new File(path).getCanonicalPath()) should be sufficient, or something similar. Bascially ensure that we only serve files from a sub directory of the path init parameter.

getCanonicalPath() will resolve symlinks so if someone symlinks out of the images directory it won't be served.  If we need to do that we could use getAbsolutePath() but I think it's probably safer to disallow links outside of the "path" configuration parameter.

src/foam/nanos/servlet/FileServlet.java
```diff
@@ -0,0 +1,74 @@
...
+      String ext = EXTS.get(FilenameUtils.getExtension(src.getName()));
+      try ( BufferedInputStream is = new BufferedInputStream(new FileInputStream(src)) ) {
+        resp.setContentType(!SafetyUtil.isEmpty(ext) ? ext : DEFAULT_EXT);
+        resp.setHeader("Content-Disposition", "filename=\"" + StringEscapeUtils.escapeHtml4(src.getName()) + "\"");
+        resp.setContentLengthLong(src.length());
+
+        int read = 0;
+        byte[] buffer = new byte[BUFFER_SIZE];
```
adamvy, line 66: Can also use IOUtils.copy(is, resp.getOuptutStream()) since I think we're pulling in commons-io as a dependency already.  Saves you a bit of code, but I'm also okay with what you already have since it avoids a dependency ;).

PR #1422, opened 2018-08-03
Change relationships properties to axioms

mcarcaso:
Change `OneToManyRelationshipProperty` to `OneToManyRelationshipAxiom` and `ManyToManyRelationshipProperty` to `ManyToManyRelationshipAxiom`. Also implement the `installInClass` methods on these axioms to install a method and property axioms to the class. The method takes a context and returns the predicated DAO and the property is just a getter that calls this method with the object's own context.

Note that this breaks any existing java code using the relationships because the methods now have a `get` in front of them. I've updated the relationship test to reflect this and it passes.

Also implement the swift pieces of `RelationshipDAO`. Without this change swift fails to compile. With this change, it compiles but this stuff is mostly untested. I'll fix anything that comes up in that realm but I'm looking to land this asap to unblock anybody working with swift.


src/foam/dao/Relationship.js
```diff
@@ -589,76 +628,68 @@ foam.CLASS({
...
-
-  methods: [
-    function buildJavaClass(cls) {
-      this.SUPER(cls);
-      cls.method({
-        name: this.name,
-        visibility: 'public',
-        type: 'foam.dao.ManyToManyRelationship',
-        args: [{ name: 'x', type: 'foam.core.X' }],
-        body: `return new foam.dao.ManyToManyRelationshipImpl.Builder(x)
```
adamvy, line 623: Could we just add this as javaCode to the method we add in JS?

I guess then we would be generating this code when running in web mode so maybe that's not the best idea.

adamvy: Nevermind my idea appears to be what you did, I just didn't read the diff very well.  github needs a colourblind mode because red/green diffs are not fun when your display brightness isn't stellar.

PR #1434, opened 2018-08-07
Some swift generation features around constants

mcarcaso:
Give `foam.core.Constant` a `swiftFactory` property, add some extra handling to `foam.swift.stringify`, and use `foam.swift.stringify` in the `expression` for `foam.core.Constant`'s `swiftValue`

### DIFF
--- a/src/foam/core/Constant.js
+++ b/src/foam/core/Constant.js
@@ -38,7 +38,7 @@ foam.CLASS({
     {
       name: 'value',
       expression: function(factory) {
-        return factory();
+        return factory ? factory() : null;
       },
     },
     'factory',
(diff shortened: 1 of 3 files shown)


src/foam/core/Constant.js
```diff
@@ -38,7 +38,7 @@ foam.CLASS({
     {
       name: 'value',
       expression: function(factory) {
-        return factory();
+        return factory ? factory() : null;
```
adamvy, line 41: I'd be curious to know when this happens, are people setting factory to null?

mcarcaso: I triggered it when I had a constant that only had a swiftFactory and no js values.

PR #1439, opened 2018-08-08
Swift - Implement generic compare method

mcarcaso:
This requires a similar approach that stdlib provides where each primitive has a set of functions for determining isInstance and compare and the generic compare method determines the type before comparing.


mcarcaso:
Woops merged this one a little prematurely but it only touches swift stuff so should be ok. Feel free to retro review it.

PR #1441, opened 2018-08-08
Swift relationship support and more

mcarcaso:
Bunch of swift stuff:
- Implement mlang.sink.Max
- Implement inX
- Implement ContextualizingDAO
- Implement SequenceNumberDAO
- Fix bug in swift around synchronized methods not using `try` if the method throws and issue with overriding by making synchronized method private.
- Support a swiftSetter
- Support MultiPartID, IDAlias, and Reference properties.
- Write a swift relationship test that mostly mimics java's.


src/foam/dao/Relationship.js
```diff
@@ -685,10 +685,11 @@ foam.CLASS({
         name: this.propertyName,
         of: 'foam.dao.ManyToManyRelationship',
         flags: ['swift', 'js'],
-        getter: function() {
+        transient: true,
+        factory: function() {
           return this[methodName](this.__context__);
```
adamvy: If we switch this to a factory we also need to make sure we clear the property when id properties are changed, or make it an expression on id.

mcarcaso: Good call. Done.

PR #1454, opened 2018-08-11
DO_NOT_MERGE: Add instance: (instanceOf) filtering in MQL

jlhughes:
@hchoura I need your assistance with adding a feature to MQL. 

IsInstanceOfParser is intended to add an 'instance:' query which will add an INSTANCE_OF predicate. 

- Experiencing ClassNotFoundException on Class.forName on the instance: value. 
- Presently inserted in QueryParser, but believe it should be added to the parser stack so it can be used from within AND/OR... 

To Test: 

http://localhost:8080/service/dig?dao=userDAO&cmd=select&format=json&q=instance:foam.nanos.auth.User

Thanks for your help.


src/foam/lib/query/IsInstanceOfParser.java
```diff
@@ -0,0 +1,58 @@
...
+      // TODO(adamvy): Better handle errors.
+
+      ClassInfo info;
+      try {
+        info = (ClassInfo) cls.getMethod("getOwnClassInfo").invoke(null);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+
+      ps.setValue( new IsInstanceOf(info) );
```
hchoura: In setValue, we need to pass a Predicate. In this case, foam.mlang.predicate.IsInstanceOf exp = new foam.mlang.predicate.IsInstanceOf();

PR #1459, opened 2018-08-14
Add test for Unauthorized API request

TW80000:
Sends an unauthorized request to the API, expecting to get a "401 Unauthorized" response.

It's failing right now because our API currently responds with a "200 OK" with a login form in the response body for unauthorized requests. I think this is incorrect behaviour as far as our customers consuming the API would be concerned.

Related to https://github.com/nanoPayinc/NANOPAY/issues/3149

## Code

```JavaScript
UnauthorizedRequest();

/** Send an unauthorized request to the API */
async function UnauthorizedRequest() {
  var uri = `http://localhost:8080/service/dig?dao=userDAO`;
  var body = JSON.stringify({
    email: 'email@example.com',
    password: 'somePassword123',
    firstName: 'John',
    lastName: 'Doe'
  });
  var headers = new Headers({
    'Accept': 'application/json',
    'Cache-Control': 'no-cache',
    'Content-Type': 'application/json'
    // Exclude 'Authorization' header.
  });
  var requestInit = {
    method: 'POST',
    headers: headers,
    body: body,
    credentials: 'omit' // Don't send cookies.
  }
  var request = new Request(uri, requestInit);
  try {
    var response = await fetch(request);
    test(response.status === 401, 'Unauthorized requests get a 401 status response.');
    if ( response.status !== 401 ) print(`Expected status code to be 401 but was ${response.status}.`);
    test(response.statusText === 'Unauthorized', `Unauthorized requests get an 'Unauthorized' status response.`);
    if ( response.statusText !== 'Unauthorized' ) print(`Expected status code to be 'Unauthorized' but was '${response.statusText}'.`);
    try { print(await response.text()); } catch (err) {}
  } catch (err) {
    print(err.message);
    test(false, `Unauthorized requests shouldn't be causing an error here.`);
  }
}
```


kgrgreer:
We need to distinguish between unauthorized and authorization-required exceptions so we know when a login request is really required.

PR #1461, opened 2018-08-14
Move foam.SCRIPT initialization to earlier in the boot process.

mcarcaso:


kgrgreer:
Do we need to do this? Can we realistically go back this far in the boot process and be loading code from a DAO anyway?

mcarcaso:
For this, it's not so much about loading code from a DAO and more about putting code into a DAO that we can later select from and write out to a file tree of individual files. Doing it this way makes foam.js and files.js the only special scripts and everything else is a file containing a single foam.CLASS, foam.SCRIPT, foam.RELATIONSHIP, foam.INTERFACE, etc. It also allows us to add some metadata to a script if need be like requires or documentation.

src/foam.js
```diff
@@ -100,5 +100,14 @@
   //  delete this.FOAM_FILES;
   };
 
+  this.foam = {
```
adamvy: I think we should move this to lib.js so that it's available regardless of the method used to boot foam. If we concat files.js this file won't make the list.

mcarcaso: Ah yes sorry I missed the fact that we won't need foam.js when we build by concat-ing files.js. Will update and reopen.

mcarcaso:
PTAL. Moved foam.SCRIPT init into lib.js and I'll have to special case poly.js and lib.js in build tool to know now to wrap those in foam.SCRIPT. Everything else in the system can be wrapped.

src/foam/core/Method.js
```diff
@@ -173,5 +173,10 @@ foam.CLASS({
   ]
 });
 
-
-foam.boot.phase2();
+foam.SCRIPT({
+  package: 'foam.core',
+  name: 'MethodScript',
```
adamvy: I think there can be a better name for this, like "Phase2" or "BootPhase2"

PR #1464, opened 2018-08-14
Query parser test

hchoura:


src/foam/parse/QueryParserUser.java
```diff
@@ -0,0 +1,38 @@
...
+
+    StringPStream sps = new StringPStream();
+    sps.setString(query);
+    PStream ps = sps;
+    ParserContext x = new ParserContextImpl();
+    ps = parser.parse(ps, x);
+    if (ps == null)
+      return false;
+
+    foam.mlang.predicate.Nary result = (foam.mlang.predicate.Nary) ps.value();
```
kgrgreer, line 34: Why not just import Nary rather than use full path?

src/foam/parse/QueryParserUser.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class QueryParserUser {
+
+  /**
+   * Determines if the query is valid or not
+   *
+   * @param query the query to validate
+   * @return true if it is a valid query, otherwise false
+   */
+  public static boolean isValid(String query, String statement) {
```
kgrgreer, line 23: Why not just move this method to QueryParserUserTest?

src/foam/parse/QueryParserUserTest.js
```diff
@@ -0,0 +1,62 @@
+foam.CLASS({
+  package: 'foam.parse',
+  name: 'QueryParserUserTest',
```
kgrgreer, line 3: Why is this called UserTest instead of just Test? What does User refer to?

src/foam/parse/QueryParserUserTest.js
```diff
@@ -0,0 +1,62 @@
+foam.CLASS({
```
kgrgreer, line 1: It's okay that you modelled it, but you don't need to. You could just create a Java class which extends Test and overrides the runTest method.

src/foam/parse/QueryParserUserTest.js
hchoura: For the QueryParser, we need to define a ClassInfo. I use User as a ClassInfo to be able to parse the query dynamically.

kgrgreer: ok

PR #1469, opened 2018-08-15
Allow edit button to redirect to customized view in DAOControllerView

Henry0422:
- Added the updateView property in `DAOMenu.js`
- Added `updateView` in `exports` in `BrowserView.js` to let import `'updateView? as importedUpdateView'`work
- Updated `onEdit` event listener


src/foam/comics/DAOControllerView.js
```diff
@@ -122,10 +131,17 @@ foam.CLASS({
...
-      }, this);
+      if ( this.updateView.class === 'foam.comics.DAOUpdateControllerView' ) {
+        this.stack.push({
+          class: 'foam.comics.DAOUpdateControllerView',
+          key: id
+        }, this);
+      } else {
+        this.stack.push({
+          class: 'net.nanopay.auth.ui.UserDetailView',
+          userId: id
```
zahnno: We shouldn't be pushing any nanopay specific code into foam. I think the idea here would be to replace the class being pushed into the stack and have a requirement of a key property on the detail view being passed in. Allowing us to remove this condition.

kgrgreer: Yes, like @zahnno says, we can't have nanopay specific code in a generic FOAM feature. If you've modified the updateView then why do you need this change?

PR #1474, opened 2018-08-15
DAOControllerView Appended Export Feature

zahnno:
Appended an action to the DAOController which displays an Export modal within the DAOControllerView. The modal provides the ability to convert objects within a passed in DAO or a single FObject to the formats listed in the driverRegistryDAO (XML, CSV, JSON). Exports filtered objects.

Relates to: https://github.com/nanoPayinc/NANOPAY/issues/3189

### DIFF
--- a/src/foam/u2/ExportModal.js
+++ b/src/foam/u2/ExportModal.js
@@ -5,32 +5,15 @@ foam.CLASS({
 
   documentation: 'Export Modal',
 
+  imports: [
+    'exportDriverRegistryDAO'
+  ],
+
   requires: [
-    'foam.u2.ModalHeader',
-    'foam.nanos.export.JSONDriver',
-    'foam.nanos.export.XMLDriver',
-    'foam.nanos.export.CSVDriver'
+    'foam.u2.ModalHeader'
   ],
 
   properties: [
-    {
-      name: 'jsonDriver',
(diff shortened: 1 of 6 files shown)


src/foam/u2/ExportModal.js
```diff
@@ -0,0 +1,145 @@
+foam.CLASS({
+  package: 'foam.u2',
+  name: 'ExportModal',
+  extends: 'foam.u2.View',
+
+  documentation: 'Export Modal',
+
+  requires: [
+    'foam.u2.ModalHeader',
+    'foam.nanos.export.JSONDriver',
```
kgrgreer: Isn't the list of export drivers stored in a DAO, which would make it more extensible than hard-coding it here?

src/foam/u2/ExportModal.js
```diff
@@ -0,0 +1,145 @@
...
+  methods: [
+    function initE() {
+      this.SUPER();
+
+      this
+      .tag(this.ModalHeader.create({
+        title: 'Export'
+      }))
+      .addClass(this.myClass())
+        .startContext({ data: this })
```
kgrgreer: Why is this line indented (or why aren't the previous two). In either case, it should be the same as the previous two.

PR #1479, opened 2018-08-16
Fixed duplicate issue and updating permissionView

minsunny:
https://github.com/nanoPayinc/NANOPAY/issues/3135

- Fixed duplication of group header
- Updating to show only user's own group's permission first

### DIFF
--- a/src/foam/nanos/auth/PermissionTableView.js
+++ b/src/foam/nanos/auth/PermissionTableView.js
@@ -12,7 +12,8 @@ foam.CLASS({
   imports: [
     'auth',
     'groupDAO',
-    'permissionDAO'
+    'permissionDAO',
+    'user'
   ],
 
   requires: [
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/PermissionTableView.js
```diff
@@ -73,7 +74,12 @@ foam.CLASS({
             .add('Groups: ').tag('br')
 
             .start().select(self.groupDAO.orderBy(self.Group.ID), function(g) {
-              groups[g.id] = foam.u2.md.CheckBox.create({label: g.id, data: true});
+              if ( self.user.group == g.id ) {
```
kgrgreer: This would be shorter as:
groups[g.id] = foam.u2.md.CheckBox.create({label: g.id, data: self.user.group == g.id});

src/foam/u2/Element.js
```diff
@@ -1650,7 +1650,7 @@ foam.CLASS({
       }, this);
 
       this.onDetach(dao.listen(listener));
-      listener.delegate.paint();
+      //listener.delegate.paint();
```
kgrgreer: Are you sure this doesn't break other things?

PR #1480, opened 2018-08-16
Add generic search on String properties to ReciprocalSearch

TW80000:
Lets you type free-form text into the general search box at the top of ReciprocalSearch. Will still use MQL, but will also do a CONTAINS_IC on all columns of type String.

This lets you type in 'account' instead of 'name:account', for example.

### DIFF
--- a/src/foam/u2/search/TextSearchView.js
+++ b/src/foam/u2/search/TextSearchView.js
@@ -21,10 +21,18 @@ foam.CLASS({
   extends: 'foam.u2.View',
 
   requires: [
-    'foam.mlang.predicate.ContainsIC',
     'foam.mlang.predicate.True',
+    'foam.mlang.predicate.False',
     'foam.parse.QueryParser',
     'foam.u2.tag.Input'
+  ],
+
+  imports: [
+    'filterController'
+  ],
+
+  implements: [
+    'foam.mlang.Expressions'
   ],
 
   properties: [
@@ -100,11 +108,27 @@ foam.CLASS({
       name: 'updateValue',
       code: function() {
         var value = this.view.data;
-        this.predicate = ! value ?
-            this.True.create() :
-            this.richSearch ?
-                this.queryParser.parseString(value) :
-                this.ContainsIC(this.property, value);
+
+        if ( ! value ) {
+          this.predicate = this.True.create();
+          return;
+        }
+
+        if ( this.richSearch ) {
+          this.predicate = this.filterController.filters
+            .map((name) => this.of.getAxiomByName(name))
+            .filter((property) => foam.core.String.isInstance(property))
+            .reduce(
+              (acc, property) => this.OR(
+                this.CONTAINS_IC(property, value),
+                acc || this.False.create()
+              ),
+              this.queryParser.parseString(value)
+            );
+          return;
+        }
+
+        this.predicate = this.CONTAINS_IC(this.property, value);
       }
     }
   ]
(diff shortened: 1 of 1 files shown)


src/foam/u2/search/TextSearchView.js
```diff
@@ -100,11 +108,27 @@ foam.CLASS({
...
-                this.ContainsIC(this.property, value);
+
+        if ( ! value ) {
+          this.predicate = this.True.create();
+          return;
+        }
+
+        if ( this.richSearch ) {
+          this.predicate = this.filterController.filters
+            .map((name) => this.of.getAxiomByName(name))
```
kgrgreer: Better to do getAxiomByClass(foam.core.String)

src/foam/u2/search/TextSearchView.js
```diff
@@ -100,11 +108,27 @@ foam.CLASS({
...
+            .filter((property) => foam.core.String.isInstance(property))
+            .reduce(
```
kgrgreer: OR is NARY, so you can use .apply() to only create one OR instead of reducing into a tree of OR's.

kgrgreer: Although, it won't make any difference in the end since FOAM will partialEval it into one OR anyway. Just letting you know.

src/foam/u2/search/TextSearchView.js
```diff
@@ -100,11 +108,27 @@ foam.CLASS({
...
+              (acc, property) => this.OR(
+                this.CONTAINS_IC(property, value),
+                acc || this.False.create()
+              ),
+              this.queryParser.parseString(value)
+            );
+          return;
+        }
+
+        this.predicate = this.CONTAINS_IC(this.property, value);
```
kgrgreer: There should be an option for turning CONTAINS_IC on or off since it requires doing a full-table scan which is okay for small reference tables but could be very expensive for tables like Transactions. There's a way to fix this by creating a new type of keyword search MDAO Index (like an inverted-index or FMIndex) but we don't have this yet. Actually, we have the start of an FMIndex.

src/foam/u2/search/TextSearchView.js
TW80000: If we did that, then it could return properties that aren't set as search columns on the model. The current way avoids that problem. We could do getAxiomByClass and filter afterwards to make sure the properties are search columns, but I don't see the advantage of doing it that way over the current approach.

kgrgreer: Where/how is it filtering search columns? Search columns are those that appear in the reciprocal search panel but it doesn't mean we can't search them in the query parser or in your full-text search.

TW80000: > Where/how is it filtering search columns? 

`this.filterController.filters` is an array of the search column names. It gets them from ReciprocalSearch, which exports itself as filterController.

> Search columns are those that appear in the reciprocal search panel but it doesn't mean we can't search them in the query parser or in your full-text search.

That's true, but I think that would seem like strange behaviour from the user's perspective. If I got rows that didn't visibly contain the text I searched for I'd be confused.

TW80000:
@kgrgreer We could do this as well:

```JavaScript
        var value = this.view.data;
        this.predicate = ! value ?
          this.True.create() :
          this.richSearch ?
            this.OR(
              this.queryParser.parseString(value) || this.False.create(),
              this.KEYWORD(value)
            ) :
            this.checkStrictEquality ?
              this.EQ(this.property, value) :
              this.CONTAINS_IC(this.property, value);
```

The Keyword mlang does pretty much what I did here but isn't specific to search columns.

kgrgreer:
Using KEYWORD seems better. Higher-level / more-declarative will make easier to optimize.

TW80000:
@kgrgreer Updated.

PR #1483, opened 2018-08-17
Create the DirCrawlModelDAO

mcarcaso:
This DAO will crawl the given directory and fill itself with any models, libs, relationships, interfaces, etc. that it finds. This will be an integral part to the build tool.

### DIFF
--- a/src/foam/build/DirCrawlModelDAO.js
+++ b/src/foam/build/DirCrawlModelDAO.js
@@ -151,7 +151,9 @@ foam.CLASS({
           return scriptFilesExp.exec(o) && !blacklistExp.exec(o)
         })
 
-      // TODO: Do something with java files?
+      // TODO: Add a JavaClass model for manually written java classes.
+      // Parse the package/name from the .java and then store them in the DAO as
+      // well. Also do the same for swift files.
       var javaFiles = files
         .filter(function(o) {
           return o.endsWith('.java');
--- a/src/foam/build/DirCrawlModelDAO.js
+++ b/src/foam/build/DirCrawlModelDAO.js
@@ -1,0 +1,201 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.build',
+  name: 'DirCrawlModelDAO',
+  extends: 'foam.dao.PromisedDAO',
+  requires: [
+    'foam.build.Lib',
+    'foam.core.Script',
+    'foam.dao.EasyDAO',
+    'foam.dao.Relationship',
+  ],
+  imports: [
+    'classloader',
+  ],
+  properties: [
+    {
+      name: 'unwrappedScripts',
+      value: [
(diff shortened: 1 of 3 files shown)


src/foam/build/DirCrawlModelDAO.js
```diff
@@ -0,0 +1,192 @@
...
+        promises.push(Promise.all(foam.json.references(self.__context__, m)).then(function(p) {
+          var cls = self.lookup(m.class || 'foam.core.Model')
+          return dao.put(cls.create(m));
+        }))
+      };
+
+      context.foam.RELATIONSHIP = function(m) {
+        var r = self.Relationship.create(m);
+        promises.push(Promise.all([
+          self.classloader.load(r.sourceModel),
```
adamvy: Why is this necessary?  I don't think the Relationship model will try to resolve source/targetModel anymore until initRelationship() is called.

src/foam/build/DirCrawlModelDAO.js
```diff
@@ -0,0 +1,192 @@
...
+        ]).then(function() {
+          return dao.put(r);
+        }))
+      };
+
+      var self = this;
+      var scriptFilesExp = new RegExp(self.unwrappedScripts.join('|'));
+      var blacklistExp = new RegExp(self.blacklist.join('|'));
+      var files = require('child_process')
+        .execSync(`find ${self.srcDir}`)
```
adamvy: Oh god.  Let's implement a file tree walk in JS instead.  Something like the posix ftw() function https://linux.die.net/man/3/ftw

src/foam/build/DirCrawlModelDAO.js
mcarcaso: It's used when figuring out the package which is needed for the ID:
https://github.com/foam-framework/foam2/blob/c7ddf2d90bce6d81247c229f493b4349b3988241/src/foam/dao/Relationship.js#L51

src/foam/build/DirCrawlModelDAO.js
```diff
@@ -0,0 +1,192 @@
...
+        .filter(function(o) {
+          return !scriptFilesExp.exec(o) && !blacklistExp.exec(o)
+        })
+
+      var scriptFiles = jsFiles
+        .filter(function(o) {
+          return scriptFilesExp.exec(o) && !blacklistExp.exec(o)
+        })
+
+      // TODO: Do something with java files?
```
adamvy: TODO: Add a JavaClass model for manually written java classes.  Parse the package/name from the .java and then store them in the DAO as well.

Also TODO: Do the same for swift.

mcarcaso: Done

src/foam/build/DirCrawlModelDAO.js
mcarcaso: Done

src/foam/build/DirCrawlModelDAO.js
mcarcaso: Updated the foam.dao.Relationship code to figure out the package/name without having to do a lookup so I was able to remove this.

PR #1485, opened 2018-08-19
Modified WebSocketBox

kirkeaton:
Modified WebSocketBox to modify url to be /socket/{service} instead of /service/{service}.

The reason behind this change was that the NanoRouter's pathSpec of of "/service/*" was capturing the incoming WebSocket requests instead of the NanoWebSocketServer.


kirkeaton:
Didn't fix the issue. Closing

PR #1487, opened 2018-08-20
Fix button spacing

TW80000:
Fixes https://github.com/nanoPayinc/NANOPAY/issues/3263

Old:

screen shot 2018-08-20 at 1 33 16 pm

New:

screen shot 2018-08-20 at 1 48 01 pm

### DIFF
--- a/src/foam/comics/DAOControllerView.js
+++ b/src/foam/comics/DAOControllerView.js
@@ -52,13 +52,12 @@ foam.CLASS({
 
     .middle-row .actions {
       display: inline-block;
-      margin-bottom: 10px;
     }
 
     .middle-row .net-nanopay-ui-ActionView {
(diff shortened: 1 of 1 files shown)


kgrgreer:
Could we move the Export button into the same row as the other buttons? Also, why is the Edit button grey? I think that all of the buttons should be the same colour (except for delete buttons which are red) or all buttons should be the same except for creation buttons which should be a different colour, but the danger there is that the UI's look too much like christmas trees if you have too many colours.

TW80000:
It is on the same row, the wrapping only happens when you make the screen narrow enough for the buttons to start wrapping.

The edit button is grey because it's disabled until you click on a row in the table, then it turns blue and you can click it to edit the row you selected.

PR #1489, opened 2018-08-20
Fix DAO Controller border support

TW80000:
Closes https://github.com/nanoPayinc/NANOPAY/issues/3186


src/foam/u2/borders/NullBorder.js
```diff
@@ -0,0 +1,27 @@
...
+  name: 'NullBorder',
+  extends: 'foam.u2.Element',
+
+  documentation: `
+    An unstyled border. Intended for use as a default value for
+    border properties.
+  `,
+
+  methods: [
+    function initE() {
```
kgrgreer: If you don't provide this initE() method does the standard Element just work?

TW80000: I just tried and it doesn't.

Issue #1488, opened 2018-08-20
DAO Controller View does not show ReciprocalSearch sidebar if searchColumns is an empty array


TW80000:
It should still be visible with the general search and clear button.

TW80000:
This is incorrect. See https://github.com/foam-framework/foam2/issues/1544

PR #1490, opened 2018-08-20
Show ReciprocalSearch when searchColumns is empty

TW80000:
- The general search field can still be used.

Closes https://github.com/foam-framework/foam2/issues/1488

### DIFF
--- a/src/foam/u2/view/ReciprocalSearch.js
+++ b/src/foam/u2/view/ReciprocalSearch.js
@@ -134,8 +134,6 @@ foam.CLASS({
       this.
         addClass(self.myClass()).
         add(this.slot(function(filters) {
-          self.show(filters.length);
-
           var searchManager = self.SearchManager.create({
             dao$: self.dao$,
             predicate$: self.data$


kgrgreer:
How do I specify I don't want any search border from the model then?

You can tell if the user set searchColumns to [] or it's just blank like this:

foam.nanos.menu.Menu.model_.searchColumns;
undefined
foam.nanos.cron.Cron.model_.searchColumns;
(2) ["id", "description"]

TW80000:
I added the ability to do that through a flag on DAOController last week.

https://github.com/foam-framework/foam2/blob/master/src/foam/comics/DAOController.js#L90

Either solution could solve the problem. One advantage of having a flag on the view makes it possible to include the filters on one page but hide them in another if we wanted to. Having the preference set on the model means it would have to be the same everywhere, wouldn't it? I'm open to whatever solution we think is best though.

kgrgreer:
I think that setting the searchColumns should be an option on the controller as well in case you don't want the defaults and we should treat searchColumns: [] an an explicit request for now searchColumns.

TW80000:
> I think that setting the searchColumns should be an option on the controller as well...

Did you mean model instead of controller? Assuming you did, should an empty array for searchColumns mean the general search field is still available, or should the entire sidebar be hidden?

kgrgreer:
I meant controller. It's already an option to set it on the model. I should be able to set it on the controller if I don't want to use what the model specifies.

TW80000:
Thanks, I understand now. Sorry for the confusion. Will update soon.

TW80000:
Opened a ticket for the proposed behaviour here: https://github.com/foam-framework/foam2/issues/1544

Closing this since it doesn't make sense anymore. I'll open a new PR to implement https://github.com/foam-framework/foam2/issues/1544 when I get around to it.

Issue #1023, opened 2018-02-11
foam.core.FObject.isSubClass() is broken.


adamvy:
foam.core.FObject.isSubClass({ id: 'foo' }) === true

Because we try to be clever and short circuit it if this === foam.core.FObject.

PR #1493, opened 2018-08-21
Create a tool for writing stripped files to a directory

mcarcaso:


src/foam/build/JsCodeOutputter.js
```diff
@@ -0,0 +1,132 @@
...
+          } else if ( type == foam.Array ) {
+            out.array();
+            for ( var i = 0 ; i < v.length ; i++ ) {
+              this.output(x, v[i])
+            }
+            out.end()
+          } else if ( type == foam.Date ) {
+            debugger;
+          } else if ( type == foam.Object ) {
+            if ( v['getAxioms'] ) { // Is an actual class
```
adamvy: Use this logic from foam/core/JSON.js for detecting if its a class

```
( o && o.prototype && foam.core.FObject.prototype.isPrototypeOf(o.prototype)
```

adamvy: And put a reference to issue #1023 so people know why you had to do this :)

PR #1496, opened 2018-08-21
files.js generator

mcarcaso:
A tool for generating files.js that'll be used with the build tool.

### DIFF
--- a/src/foam/build/FilesJsGen.js
+++ b/src/foam/build/FilesJsGen.js
@@ -176,7 +176,8 @@ foam.CLASS({
                 order.unshift(k)
               });
             }
-            // Remove anyting not in the DAO.
+            // Remove anyting not in the DAO. This can happen for inner classes
+            // and enums. TODO: Would be nice if we didn't have to do this.
             return Promise.all(order.map(function(id) {
               return self.modelDAO.find(id);
             })).then(function(models) {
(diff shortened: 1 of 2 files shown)


src/foam/build/FilesJsGen.js
```diff
@@ -0,0 +1,290 @@
...
+    {
+      name: 'modelDAO',
+      expression: function(srcDir) {
+        return this.DirCrawlModelDAO.create({srcDir: srcDir})
+      },
+    },
+  ],
+  methods: [
+    function execute() {
+      this.getFilesJs().then(console.log);
```
adamvy: I think you should console.log.bind(console), its not pre-bound on all systems.

src/foam/build/FilesJsGen.js
```diff
@@ -0,0 +1,290 @@
...
+      if ( foam.Array.isInstance(o) ) {
+        return Promise.all(o.map(function(i) {
+          return self.getDepsTree(i, seen, head);
+        })).then(function() { return head });
+      } else if ( foam.Object.isInstance(o) ) {
+        // Check if it's an actual class. foam.core.FObject.isSubClass
+        // should work but doesn't:
+        // https://github.com/foam-framework/foam2/issues/1023
+        if ( o && o.prototype && (
+            foam.core.FObject.prototype === o.prototype ||
```
adamvy: Indentation here is confusing, move the open parenthesis to this line so it read

```
if ( o && o.prototype &&
     ( foam.core.FObject... ||
       foam.core.FObject... ) ) {
```

mcarcaso: Ah yeah that's nicer. Fixed here and in the JsCodeOutputter.

src/foam/build/FilesJsGen.js
mcarcaso: Done

adamvy:
Merging as is because I haven't come up with a good TREE mlang yet.  But if I think of one I'll let you know.

PR #1498, opened 2018-08-21
Create the tool that brings all of these build tools together

mcarcaso:
This tool takes srcDirs, a destDir, and required models and does the following:
1. Strips and writes all models from the srcDirs into the destDir.
2. Copies all non-js files from the srcDirs to the destDir (for things like images and html files).
3. Creates a files.js based on the required models that are given.
4. Concats the files in files.js into a foam-bin.js that can be used to boot foam and all required models.


src/foam/build/DirCrawlModelDAO.js
```diff
@@ -84,7 +84,7 @@ foam.CLASS({
 
       context.foam.LIB = function(m) {
         if ( ! m.id ) {
-          m.id = 'lib.Lib' + Date.now()
+          m.id = 'lib.Lib' + foam.next$UID()
```
adamvy: Can we name libs of the first method/axiom or something?  Or could we use a local incrementing id in order to make the generated ids a bit more stable?

mcarcaso: Yup how does the name + first method name sound? I could add all method names to be ultra safe but that'll cause really long IDs.

PR #1499, opened 2018-08-21
Custom Top Navigation & Footer View

zahnno:
Append custom Top Navigation & Footer View to groups allowing for custom top Nav & footers on user login.

I was a little hesitant on storing the views as strings but my other approach consisted appending viewSpecs onto the group model. I ran into the issue of making viewSpecs configurable.

### DIFF
--- a/src/foam/nanos/controller/ApplicationController.js
+++ b/src/foam/nanos/controller/ApplicationController.js
@@ -39,6 +39,7 @@ foam.CLASS({
     'foam.nanos.auth.User',
     'foam.nanos.auth.resetPassword.ResetView',
     'foam.nanos.u2.navigation.TopNavigation',
+    'foam.nanos.u2.navigation.FooterView',
     'foam.u2.stack.Stack',
     'foam.u2.stack.StackView'
   ],
@@ -141,7 +142,9 @@ foam.CLASS({
     'secondaryHoverColor',
     'tableColor',
     'tableHoverColor',
-    'accentColor'
+    'accentColor',
+    'topNavigation_',
+    'footerView_'
   ],
 
   methods: [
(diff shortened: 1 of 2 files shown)


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -172,30 +175,48 @@ foam.CLASS({
...
-          .tag({class: 'foam.nanos.u2.navigation.TopNavigation'})
+          .start('div', null, self.topNavigation_$).end()
           .start('div').addClass('stack-wrapper')
             .tag({class: 'foam.u2.stack.StackView', data: self.stack, showActions: false})
-          .end();
+          .end()
+          .start('div', null, self.footerView_$).end();
+
+          // Sets up application view
+          self.topNavigation_.add(self.TopNavigation.create());
```
kgrgreer, line 185: Why is this done here rather than just done directly above?

kgrgreer: So you can replace them in setPortalView?

zahnno: topNavigation_ is undefined at that point.

TW80000:
This approach seems very specific to the current Controller we're using. Here we're coupling groups to our current UI fairly tightly. What if a group in the future needs a UI with elements other than these two to look different? If we kept adding elements to configure to the group model, the coupling will get tighter and tighter.

What if instead, each group provided a more generic view, like an entire controller or a border? Then in NANOPAY we could just make something like `SelfServeBorder` or `SelfServeApplicationController` and the group only ever gets one extra property, for a top-level container. I haven't thought about it enough to know if that would actually be possible though, because you'd have to make a decision on where to draw the line between the border and the content if you were to go with the border approach. With an entire controller you don't have that issue as far as I've thought about it though.

Mainly just thinking out loud here though because of the feeling of tight coupling I get from this solution. Maybe I'm worried about nothing though.

zahnno:
@TW80000 The controller is currently responsible for dictating what view gets pushed to the stack after login, I don’t see why extending the feature to include changing the topNavigation & footer is that much of a concern. The recurring design that persists through all of our app iterations is the navigation bar & footer are always persistent when traversing through the app. I can't really think of other UI elements that might need to persist based off any sort of user or group value.

PR #1503, opened 2018-08-22
Fixed removing error and added success response message

minsunny:
- Added a Modelled success message according to a format
- Fixing some bugs regarding removing, putting format CSV error message

https://github.com/nanoPayinc/NANOPAY/issues/2632
https://github.com/nanoPayinc/NANOPAY/issues/3145
https://github.com/nanoPayinc/NANOPAY/issues/3150


src/foam/core/AbstractObjectPropertyInfo.java
```diff
@@ -22,7 +22,7 @@ public void setFromString(Object obj, String value) { }
 
   public Object fromString(String value) {
     //TODO
```
jlhughes: Is the //TODO still required?

PR #1508, opened 2018-08-23
DO_NOT_MERGE: Modify element lifecycle so that views can be .load()ed after .unload().

adamvy:
This makes views re-usable rather than having to construct an entire new view even time they arrive at .unload().

This add .whileLoaded(f) to Element, which can be used to attach a function that will fire every time a view is loaded.  This is useful for things like DAOListeners that we want to unload when the view is inactive but want to be able to re-attach the next time the view is rendered.

Element.select() has been updated to use whileLoaded() in order to detach/reattach the DAO listener when the element goes in and out of its lifecycle.

### DIFF
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -205,6 +205,8 @@ foam.CLASS({
     function output(out) {},
     function load() {},
     function unload() {},
+    function onLoad() {},
+    function onUnload() {},
     function onRemove() {},
     function onSetClass() {},
     function onFocus() {},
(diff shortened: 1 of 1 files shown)


adamvy:
Not ready to be merged yet.  This is code complete, but I have to test this against nanopay to make sure nothing breaks and fix anything that does.  So it's ready for review, but don't merge until I've had a chance to test.

src/foam/u2/Element.js
```diff
@@ -1466,7 +1496,12 @@ foam.CLASS({
       return c;
     },
 
-    function end() {
+    function end(opt_name) {
```
adamvy, line 1499: This is an unrelated feature but nice addition.  Helps to make balanced trees.

PR #1510, opened 2018-08-24
Fix handling of loading models in classloader.

mcarcaso:


src/foam/apploader/ClassLoader.js
```diff
@@ -164,7 +164,7 @@ have multiple classloaders running alongside eachother`
         }
 
         if ( foam.core.Model.isInstance(id) ) {
-          return this.pending[id] = this.buildClass_(id, path);
+          return this.pending[id.id] = this.buildClass_(id, path);
```
adamvy, line 167: lol

PR #1511, opened 2018-08-24
Remove copyFrom warning

TW80000:
We get this warning hundreds of times in the console. Having so many
warnings makes all warnings useless, because no developer is going to
look through hundreds of warnings to see if the code they changed
generated a new one or not, they're just going to start ignoring all
warnings.

The intent of the warning seems to be to inform the developer that when
they call `copyFrom`, it won't copy anything other than a
`foam.core.Property`. Please correct me if I'm wrong.

I think the intent of the warning is fine, but the problem is that the
vast majority of the warnings are generated by foam internal code, not
developers *using* foam, who would actually benefit from the warning.
Therefore, the warning is doing more harm than good.

### DIFF
--- a/src/foam/core/EndBoot.js
+++ b/src/foam/core/EndBoot.js
@@ -149,7 +149,7 @@ foam.CLASS({
     */
     function initArgs(args, ctx) {
       if ( ctx  ) this.__context__ = ctx;
-      if ( args ) this.copyFrom(args);
+      if ( args ) this.copyFrom(args, true);
     },
 
     /**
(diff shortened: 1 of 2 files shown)


kgrgreer:
The problem is with setting java and swift properties when we don't have java or swift support loaded. In the future when @mcarcaso finishes his build tool this won't be an issue because all of these properties will be stripped from the JS versions of the models. So we can turn it off or now, but in the future we should turn it back on because this reports things like when you do: Person.create({firstname: 'Troy'}) but the actual property is called firstName and it ignores your value. Or as a short-term solution we could just skip warning about properties that contain "swift" or "java" or to Property models.

TW80000:
@kgrgreer Updated to temporarily blacklist warnings related to Java code generation.

PR #1513, opened 2018-08-24
Application Controller Stack Style

zahnno:
- Append a style slot to the main application stack class.

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -74,10 +74,10 @@ foam.CLASS({
       name: 'footerView',
       value: 'foam.nanos.u2.navigation.FooterView'
     },
-    { 
+    {
       class: 'String',
-      name: 'stackStyle',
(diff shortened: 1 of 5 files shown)


src/foam/nanos/auth/Group.js
```diff
@@ -74,6 +74,7 @@ foam.CLASS({
       name: 'footerView',
       value: 'foam.nanos.u2.navigation.FooterView'
     },
+    { class: 'String', name: 'stackStyle' },
```
kgrgreer: Add documentation: to explain what this is for.

src/foam/u2/stack/StackView.js
```diff
@@ -38,6 +38,8 @@ foam.CLASS({
     }
   ],
 
+  css: ` %GROUPCSS% `,
```
kgrgreer: No need for back-quotes when just a single-line string.

PR #1517, opened 2018-08-24
More build related fixes and flags

mcarcaso:


src/foam/build/DirCrawlModelDAO.js
```diff
@@ -84,7 +84,7 @@ foam.CLASS({
 
       context.foam.LIB = function(m) {
         if ( ! m.id ) {
-          m.id = 'lib.Lib' + Date.now()
+          m.id = m.name + '.' + m.methods[0].name;
```
adamvy, line 87: What libs have name but not id?  We should fix that.

mcarcaso: Just about all of them lol. We have stuff like:
```
foam.LIB({
  name: 'foam.Function',
  methods: [
    function argsStr() { ... },
  ]
})
foam.LIB({
  name: 'foam.Function',
  methods: [
    function args() { ... },
  ]
})
```

adamvy: Ohhh, I follow.

PR #1518, opened 2018-08-24
Model UserUserJunction so it can be extended

TW80000:
This exists as a dependency for https://github.com/nanoPayinc/NANOPAY/pull/3358

### DIFF
--- a/src/foam/nanos/auth/UserUserJunction.js
+++ b/src/foam/nanos/auth/UserUserJunction.js
@@ -1,0 +1,23 @@
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'UserUserJunction',
+
+  documentation: `A junction between two User models. Used in relationships.`,
+
+  ids: ['sourceId', 'targetId'],
+
+  properties: [
+    {
+      class: 'Reference',
+      name: 'sourceId',
+      shortName: 's',
+      of: foam.nanos.auth.User,
+    },
+    {
+      class: 'Reference',
+      name: 'targetId',
+      shortName: 't',
+      of: foam.nanos.auth.User
+    }
+  ]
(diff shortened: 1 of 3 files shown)


kgrgreer:
Doesn't the relationship already do this internally? Can't you already extend or refine the internally generated junction model as long as you do it after you've loaded the relationship?

TW80000:
@kgrgreer I think you're right. All I should need is the line in `classes.js`, and since the relationships are both defined in nanopay, I can just put the line there instead of in foam. I'll try that, thanks.

adamvy:
@TW80000 If you specify junctionModel: 'MyUserUserJunction' in your RELATIONSHIP, the junction model it generates will be named MyUserUserJunction instead of the default UserUserJunction.

Then you can refine MyUserUserJunction with your custom properties.  We could also consider adding support to RELATIONSHIP to let you add extra information about the junction model.

TW80000:
@adamvy That's good to know, thanks. In this case, I'm not planning to do that because if I use the same base junction model for both, then I can reuse the authentication decorator I wrote for both junction DAOs. I'm open to other ways of accomplishing the same thing though if there's a good reason to do it differently.

adamvy:
Is it possible to parameterize your authentication decorator so it's not hard coded to the specific junction class?

TW80000:
~~Probably, but it would be more work and I don't see any advantage to doing it that way as opposed to refining `UserUserJunction`.~~

Clarification: I guess the advantage would be that the custom properties are more limited in scope. At the moment the refinements don't clash anyway though. Doing custom junction models is probably a bit cleaner and more future-proof though.

adamvy:
Will continue this on your NANOPAY PR

PR #1519, opened 2018-08-27
Added final method support for Java

kirkeaton:


src/foam/java/Class.js
```diff
@@ -47,8 +47,20 @@ foam.CLASS({
       name: 'static',
       value: false
     },
-    'abstract',
-    'extends',
+    {
+      class: 'Boolean',
+      name: 'final',
+      value: false
```
kgrgreer: Default value for value: is false so this line is redundant here and for abstract.

src/foam/java/refinements.js
```diff
@@ -459,6 +459,11 @@ foam.CLASS({
       name: 'javaReturns',
       expression: function(returns) { return returns || '' },
     },
+    {
+      class: 'Boolean',
+      name: 'final',
+      value: false
```
kgrgreer: redundant

PR #1520, opened 2018-08-27
Security hardening Jetty

daudich:
- Preventing Jetty server from serving version number
- Adding a shutdown hook in case of improper application shutdown

### DIFF
--- a/src/foam/nanos/jetty/HttpServer.js
+++ b/src/foam/nanos/jetty/HttpServer.js
@@ -63,6 +63,10 @@ foam.CLASS({
         org.eclipse.jetty.server.Server server =
           new org.eclipse.jetty.server.Server(getPort());
 
+        /*
+          Prevent Jetty server from broadcasting its version number in the HTTP
+          response headers.
+        */
         for ( org.eclipse.jetty.server.Connector conn : server.getConnectors() ) {
           for ( org.eclipse.jetty.server.ConnectionFactory f : conn.getConnectionFactories() ) {
             if ( f instanceof org.eclipse.jetty.server.HttpConnectionFactory ) {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/jetty/HttpServer.js
```diff
@@ -59,10 +59,18 @@ foam.CLASS({
       name: 'start',
       javaCode: `
       try {
-        System.out.println("Starting jetty http server.");
+        System.out.println("Starting Jetty http server.");
         org.eclipse.jetty.server.Server server =
           new org.eclipse.jetty.server.Server(getPort());
 
+        for ( org.eclipse.jetty.server.Connector conn : server.getConnectors() ) {
```
kgrgreer: Add comment to explain what this does/is-for.

kirkeaton (changes requested):
Make use of javaImports so you don't have to write out the full class name please.

adamvy:
Personally I prefer to type out the complete class names, which is what I did when I wrote this class.

The advantage to me is that it's easier to refactor later as the code can be copy pasted as is rather than having to both copy the imports and copy the code.  Additionally you never end up with unused imports littering the top of your class.

It is more typing, but hey we shouldn't be typing so much code anyways ;)

src/foam/nanos/jetty/HttpServer.js
```diff
@@ -125,12 +133,39 @@ foam.CLASS({
...
+        addJettyShutdownHook(server);
         server.setHandler(handler);
         server.start();
       } catch(Exception e) {
         e.printStackTrace();
       }
       `
+    },
+    {
+      name: 'addJettyShutdownHook',
```
adamvy, line 149: I feel like this could have just been inlined, but not a big deal, do what feels right for you in this case.

src/foam/nanos/jetty/HttpServer.js
daudich: Added a comment.

kirkeaton (approved):

PR #1528, opened 2018-08-28
Create human name trait

TW80000:
This looks pointless on its own, but it this got merged then in the nanopay repo we could use this for contacts as originally suggested here: https://github.com/nanoPayinc/NANOPAY/pull/3358/files/7ca73a988330a11afd0316e173e1a408fba3aa33#r213310409

### DIFF
--- a/src/foam/nanos/auth/HumanNameTrait.js
+++ b/src/foam/nanos/auth/HumanNameTrait.js
@@ -1,0 +1,71 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.INTERFACE({
+  package: 'foam.nanos.auth',
+  name: 'HumanNameTrait',
+
+  properties: [
+    {
+      class: 'String',
+      name: 'firstName',
+      tableWidth: 160,
+      documentation: 'First name of user.',
+      validateObj: function(firstName) {
+        if ( firstName.length > 70 ) {
+          return 'First name cannot exceed 70 characters.';
+        }
+
+        if ( /\d/.test(firstName) ) {
+          return 'First name cannot contain numbers.';
+        }
+      }
+    },
+    {
+      class: 'String',
+      name: 'middleName',
+      documentation: 'Middle name of user.',
+      validateObj: function(middleName) {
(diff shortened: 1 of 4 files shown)


src/foam/nanos/auth/HumanNameTrait.js
```diff
@@ -0,0 +1,71 @@
...
+    {
+      class: 'String',
+      name: 'middleName',
+      documentation: 'Middle name of user.',
+      validateObj: function(middleName) {
+        if ( middleName.length > 70 ) {
+          return 'Middle name cannot exceed 70 characters.';
+        }
+
+        if ( /\d/.test(middleName) ) {
```
kgrgreer, line 36: Reminds me: we should add a 'pattern' property to Property.

adamvy: Agree, although I disagree with the decision to disallow numbers in people's names.  Who are we to say that's valid or invalid.  I'm sure somebody somewhere has a digit in their name and I really don't see the value this extra "validation" is providing us.

src/foam/nanos/auth/User.js
```diff
@@ -46,64 +47,10 @@ foam.CLASS({
...
-      expression: function(firstName, middleName, lastName) {
-        return middleName != '' ? firstName + ' ' + middleName + ' ' + lastName : firstName + ' ' + lastName;
-      },
-      javaGetter: `
-        return ! getMiddleName().equals("")
-          ? getFirstName() + " " + getMiddleName() + " " + getLastName()
-          : getFirstName() + " " + getLastName();
-      `,
-    },
+    'firstName',
```
adamvy, line 59: Why are we defining the properties again?

adamvy: Oh presumably for the java generation issue that Kirk was talking about.  Can we file an issue about that and then reference that issue from a //TODO comment here?

TW80000: Yes, it's because of the java generation problem. I'll do that now.

TW80000: Issue: https://github.com/foam-framework/foam2/issues/1529

TW80000: TODO comment added in 6c69df9

PR #1532, opened 2018-08-28
Prevented the generation of ClassInfo for AbstractFObject

kirkeaton:
In some rare cases: the `getOwnClassInfo` static method in AbstractFObject was being called instead of the subclasses.


kgrgreer:
Should we not generate for all Abstract classes, not just AbstractFObject?

kirkeaton:
@kgrgreer I don't think so. It could happen with non-abstract classes being subclassed too.

mcarcaso:
I'm confused why this is needed. When and why does this get called when you
don't expect it to?

kirkeaton:
@mcarcaso Kenny was the only one experiencing this issue and I have not been able to reproduce this at all. This fixes it for him.

mcarcaso:
I worry this could be masking a deeper issue is all. This should be
investigated because it's not clear why it's needed.

mcarcaso:
Could it be from calling the method on a hand written Java class that
extends abstractfobject?

PR #1534, opened 2018-08-29
LogMessage stdout added space between severity and message

KateMengjunXia:


src/foam/nanos/logger/StdoutLogger.java
```diff
@@ -52,6 +52,7 @@ public String format(LogRecord record) {
         str.append(record.getLevel());
       }
 
+      str.append(" ");
```
kgrgreer: Why is it a space instead of a comma?

KateMengjunXia: I can change it to a comma (or would it be better to change it to a comma followed by a space?)

kgrgreer: To be consistent with the comma after the timestamp I think it should just be a comma and no space.

PR #1535, opened 2018-08-29
Fix ProxyDAO listen support

kirkeaton:


src/foam/dao/MDAO.java
```diff
@@ -82,6 +82,7 @@ public FObject put_(X x, FObject obj) {
       }
 
       state_ = index_.put(state_, obj);
+      onPut(obj);
```
adamvy: Can we move this outside the writeLock?  Don't need to block other writers while listeners are firing.

src/foam/dao/MDAO.java
```diff
@@ -95,6 +96,7 @@ public FObject remove_(X x, FObject obj) {
 
       if ( found != null ) {
         state_ = index_.remove(state_, found);
+        onRemove(found);
```
adamvy: ditto

src/foam/dao/MDAO.java
kirkeaton: That would require moving the return statement outside of the lock as well. I'm not too dialled into the functionality of the MDAO so I don't know if moving the return values outside of the lock will have an unintended side effect

kgrgreer: onPut() and return could both be moved outside the lock.

kirkeaton: Done

src/foam/dao/MDAO.java
kirkeaton: Done

PR #1536, opened 2018-08-29
Added Java pre and post set functions

kirkeaton:
1. Added the ability to define pre and post set java functions
2. Modified Java property setters to set via PropertyInfo
3. Modified set function for PropertyInfo to call pre and post set functions
4. Significantly reduced the size of generated Enums. I removed the classInfo and setters from Enum's as you are unable to set Enum's in that fashion in Java anyway.
5. Added classInfo to generated Skeleton models.

Modifying the setters to use the PropertyInfo to set the values makes Java more closely resemble Swift as a side effect.

### DIFF
--- a/src/foam/java/PropertyInfo.js
+++ b/src/foam/java/PropertyInfo.js
@@ -141,7 +141,14 @@ foam.CLASS({
             type: 'void',
             visibility: 'public',
             args: [{ name: 'o', type: 'Object' }, { name: 'value', type: 'Object' }],
-            body: '((' + this.sourceCls.name + ') o).' + this.setterName + '(cast(value));'
+            body: `
+              ${this.sourceCls.name} obj = (${this.sourceCls.name}) o;
+              obj.assert${foam.String.capitalize(this.propName)}(cast(value));
+              Object oldValue = obj.${this.propName}IsSet_ ? obj.${this.propName}_ : null;
+              obj.${this.propName}_ = obj.${this.propName}PreSet_(oldValue, cast(value));
+              obj.${this.propName}IsSet_ = true;
+              obj.${this.propName}PostSet_(oldValue, obj.${this.propName}_);
+            `
           },
           {
             name: 'cast',
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -252,7 +252,7 @@ foam.CLASS({
             }
           ],
           body: this.javaPostSet
-        })
+        });
 
 
       if ( this.javaFactory ) {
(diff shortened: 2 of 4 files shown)


src/foam/java/refinements.js
```diff
@@ -156,14 +165,13 @@ foam.CLASS({
     function generateSetter_() {
       return this.javaSetter ? this.javaSetter : `
```
adamvy, line 166: The intended goal of this section was to generate different setters (and getters) depending on whether you actually have a pre/postSet (or factory/expression) so we don't pay the costs for it.

Although the compiler may be smart enough to just eliminate the calls to pre/postset if they are no-ops so maybe its not necessary.  Would be nice if we could verify this some way.

src/foam/java/refinements.js
```diff
@@ -156,14 +165,13 @@ foam.CLASS({
...
-        assert${foam.String.capitalize(this.name)}(val);
-        ${this.name}_ = val;
-        ${this.name}IsSet_ = true;
+        setProperty(\"${this.name}\", val);
       `;
     },
 
     function buildJavaClass(cls) {
       if ( ! this.generateJava ) return;
+      if ( foam.java.Enum.isInstance(cls) ) return;
```
adamvy, line 174: You haven't just removed setters for Enums in Java, you've completely disabled properties for enums in java.  This is now inconsistent with JS where we allow properties because often an Enum contains data you want to read.

I think it's better to continue to allow properties for Enums and just make them final by default, either by adding final support as necessary or initializing enums as frozen.

src/foam/java/PropertyInfo.js
```diff
@@ -141,7 +141,14 @@ foam.CLASS({
             type: 'void',
             visibility: 'public',
             args: [{ name: 'o', type: 'Object' }, { name: 'value', type: 'Object' }],
-            body: '((' + this.sourceCls.name + ') o).' + this.setterName + '(cast(value));'
+            body: `
+              ${this.sourceCls.name} obj = (${this.sourceCls.name}) o;
+              obj.assert${foam.String.capitalize(this.propName)}(cast(value));
+              Object oldValue = obj.${this.propName}IsSet_ ? obj.${this.propName}_ : null;
+              obj.${this.propName}_ = obj.${this.propName}PreSet_(oldValue, cast(value));
```
kgrgreer, line 148: This approach is expensive because it involves created two new objects for every primitive set. In XHome we just inserted the code-snippet from the preset and postSet in the generated call rather than calling out to the method stored in the PropertyInfo. This avoided casting to objects and avoided any call overhead when pre/postset weren't defiined. Same for the assert above, so this code is now actually creating 3 objects.

PR #1537, opened 2018-08-29
Rework "Class" property so expressions work.

### LINKED ISSUE #241: Class property ignores expression
And doesn't warn either.

This is most evident with 

foam.dao.ProxyDAO.create({ delegate: foam.dao.ArrayDAO.create({ of: SomeClass }) }).of;

which results in undefined, even though ProxyDAO.of is supposed to have an expression.

adamvy:
Instead of poorly replicating all the work that installInProto does, just let installInProto do its thing and then decorate the getter of the resulting prototype.

Fixes #241


mcarcaso (approved):
This looks fine to me but is there a bit of a deeper problem happening where having getters trump expressions (happens here: https://github.com/foam-framework/foam2/blob/master/src/foam/core/Property.js#L414 )?

Actually in this case I guess we'd still want expressions to go through the adapt and that wouldn't happen for normal expressions on properties but would happen for the Class property as you did it here.

Anyway, LGTM....was just thinking out loud.

adamvy:
I think we want getters to trump expressions. Getters trump all other property get logic, so you should only use them of you really know what you're doing or really don't need any other features.

adamvy:
Ping @kgrgreer

PR #1538, opened 2018-08-29
Changed severity in logs from id to label

KateMengjunXia:
Added a flag "javaJSONOutput" to refinements.js, and set it in LogMessage.js, so that the "severity" in logs are labels instead of id's.

### DIFF
--- a/src/foam/nanos/logger/LogMessage.js
+++ b/src/foam/nanos/logger/LogMessage.js
@@ -30,6 +30,7 @@ Implement LastModifiedByAware to suppress 'modified by' comment in journal outpu
       name: 'severity',
       class: 'Enum',
       of: 'foam.nanos.logger.LogLevel',
+      toJSON: function(value) { return value && value.label; },
       javaJSONOutput: `((foam.nanos.logger.LogLevel) value).getLabel()`
     },
     {
--- a/src/foam/nanos/logger/LogMessage.js
+++ b/src/foam/nanos/logger/LogMessage.js
@@ -30,7 +30,7 @@ Implement LastModifiedByAware to suppress 'modified by' comment in journal outpu
       name: 'severity',
       class: 'Enum',
       of: 'foam.nanos.logger.LogLevel',
-      toJSON: function(value) { return value && value.label; }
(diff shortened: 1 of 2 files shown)


src/foam/nanos/logger/LogMessage.js
```diff
@@ -30,7 +30,8 @@ Implement LastModifiedByAware to suppress 'modified by' comment in journal outpu
       name: 'severity',
       class: 'Enum',
       of: 'foam.nanos.logger.LogLevel',
-      toJSON: function(value) { return value && value.label; }
+      toJSON: function(value) { return value && value.label; },
+      javaJSONOutput: `((foam.nanos.logger.LogLevel) value).getLabel()`
```
adamvy: Will this break parsing of the journal entries back into the system?  I don't believe we parse labels into enums in java currently.

What's the motivation for this change?

kgrgreer: Implementing fromJSON as well would fix parsing.

jlhughes: The motivation in this particular case to have have human readable log messages.  So severity shows as 'debug', 'error' rather than 0, or 5 for example. 
In this case we are explicitly using a WriteOnlyJDAO with no intent, in the short term, for reading it back.  But yes, should we need/want to fromJSON would be required.   Makes sense to add it now so it's complete if we later want an external log viewer, which would be the case if log messages are all sent to a NOC. 
@KateMengjunXia please implement fromJSON.

KateMengjunXia:
I generated the following code in LogMessage.java, let me know what you think:
```
      public foam.nanos.logger.LogLevel forLabel(String label) {
        return foam.nanos.logger.LogLevel.forLabel(label);
      }
```

```
      public Object fromJSON(String value) {
        return forLabel(value);
      }
```

jlhughes:
I would just simplify it with:
```
      public Object fromJSON(String value) {
        return foam.nanos.logger.LogLevel.forLabel(label);
      }

```
If in the future there are other calls to forLabel you could pull it out then.

jlhughes (approved):

PR #1541, opened 2018-08-30
Added java pre and post set functions

kirkeaton:


src/foam/java/refinements.js
```diff
@@ -154,12 +162,39 @@ foam.CLASS({
...
+        setter += `val = ${capitalized}PreSet_(oldValue, val);\n`;
+      };
+
+      // set value
+      setter += `${this.name}_ = val;\n`;
+      setter += `${this.name}IsSet_ = true;\n`;
+
+      // add post-set function
+      if ( this.javaPostSet ) {
+        setter += `${capitalized}PostSet_(oldValue, ${this.name}_);\n`;
```
adamvy: I believe @kgrgreer was getting at literally dumping the javaPostSet code into the setter code `setter += this.javaPostSet` so we don't even have method call overhead.

kgrgreer: Yes, exactly.

kirkeaton: Done.

PR #1542, opened 2018-09-05
Update copyright of footer

Henry0422:
To show the year in the copyright at runtime.


src/foam/nanos/u2/navigation/FooterView.js
```diff
@@ -99,11 +99,14 @@ foam.CLASS({
           .start().addClass('mode').add(this.appConfig.mode$.map(function(m) { return m.label; }), ' version: ', this.appConfig.version$).end()
         .end()
         .start('div').addClass('col').addClass('copyright-label')
-          .start('p').add(this.appConfig.copyright$).end()
-        .end()
-
-        
-
+          .start('p').add(this.appConfig.copyright$.map(function(str) {
+            str = str.replace(/@\{(\w+)\}/g, function() {
```
kgrgreer, line 103: Document regex's to make easier for others to understand what you're matching. Also document in the AppConfig copyright property so that people know that this feature exists. Actually, it would make more sense for this to be a method on AppConfig itself, so that it's a self contained feature. Something like a this.appConfig.formattedCopyright$ as an expression. Also, are you sure that we need to update the copyright year? According to Google lawyers the copyright is the first year of copyright, not the last/current year.

Henry0422: I put this function in the `FooterView.js` since we have another appConfig in the service file which will overwrite the value of copyright property in `AppConfig.js`.

I will ask the legal professionals about the year of copyright.

PR #1543, opened 2018-09-05
Modified ManyToManyRelationshipImpl to use context oriented put and remove methods

kirkeaton:


adamvy:
What's the motivation for this change?  The context oriented methods are intended to be "private" hence the underscore suffix.  They used internally as part of the collective DAO system implementation, but the client API is meant to be the un-underscored version.

You can also use dao().inX(x).put() to achieve the same result but I'm curious why this is necessary.

kirkeaton:
@adamvy.
@shreymahendru Was encountering an issue where the user in the context was the system user and I discovered that it was because of the use of the non context oriented method here.

PR #1545, opened 2018-09-07
Use HTTP Request URL rather than AppConfig URL for email links

ZacZhang:
https://github.com/nanoPayinc/NANOPAY/issues/3286

- handle client requests in the nanorouter and put the URL in the AppConfig context

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -111,6 +111,7 @@ foam.CLASS({
     'foam.dao.DAO',
     'foam.nanos.app.AppConfig',
     'foam.nanos.session.Session',
+    'foam.util.SafetyUtil',
     'org.eclipse.jetty.server.Request',
     'javax.servlet.http.HttpServletRequest'
   ],
@@ -160,7 +161,7 @@ Session session = x.get(Session.class);
 if ( session != null ) {
   User user = (User) userDAO.find(session.getUserId());
   if ( user != null ) {
-    if ( this.getUrl() != null ) {
+    if ( ! SafetyUtil.isEmpty(this.getUrl()) ) {
       //populate AppConfig url with group url
       config.setUrl(this.getUrl());
     } else {
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -155,19 +155,23 @@ foam.CLASS({
       ],
       javaCode: `
 DAO userDAO         = (DAO) x.get("localUserDAO");
+DAO groupDAO        = (DAO) x.get("groupDAO");
 AppConfig config    = (AppConfig) ((AppConfig) x.get("appConfig")).fclone();
 
 Session session = x.get(Session.class);
 if ( session != null ) {
   User user = (User) userDAO.find(session.getUserId());
   if ( user != null ) {
-    if ( ! SafetyUtil.isEmpty(this.getUrl()) ) {
+    Group group    = (Group) groupDAO.find(user.getGroup());
+    if ( ! SafetyUtil.isEmpty(group.getUrl()) ) {
       //populate AppConfig url with group url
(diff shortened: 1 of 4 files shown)


src/foam/nanos/http/NanoRouter.java
```diff
@@ -64,6 +67,9 @@ protected void service(final HttpServletRequest req, final HttpServletResponse r
 
     resp.setContentType("text/html");
 
+    AppConfig appConfig = (AppConfig) getX().get("appConfig");
+    appConfig.setUrl(((Request) req).getRootURL().toString());
```
jlhughes: So when does this occur?
On first url request or on system start?  
At system start the calling url will not be known.

ZacZhang: For every request received to the servlet, the service() method is called. So each time a new request comes in, a new URL is set.

jlhughes: OK, additionally a copy of AppConfig with the calling url must be placed into the subcontext created for each service request. See line 72.  Presently one AppConfig is shared, and callers from different URLs will get last callers URL. 
72: X y = getX().put(HttpServletRequest.class, req)

If AppConfig is a singleton, then two options:
1. make it not a singleton. It may also be possible to make it a multiton using the url as part of key. 
2. change property url to method getUrl(x)

kgrgreer: This isn't thread-safe since the AppConfig is shared between all threads. AppConfig is a service installed by the NANOS boot process. If you're going to mutate it you should clone() it first.

kgrgreer: However, I would rather we don't mutate the appConfig but instead don't use the AppConfig.url for this purpose and instead just use the Request.getRootURL() if that's what we want to use.

jlhughes: Be we don't have access to Request everywhere we need the url, so perhaps, calculate it, then place it in the context.

kgrgreer: Request is in the Context so is available anywhere X is.

jlhughes: Ok, after a lengthy debate, please do the following:
1. Group - add property url - default to null.
2. Group - add method to create a cloned AppConfig "getAppConfig(x)" populated with the groups url if provided.   Also, use Request.getRootURL if group url not provided and logged in.  The cron generation needs to use the group url or the default AppConfig.url value.
3. on Login place the cloned AppConfig in the context
4. on Email generation, decorate to add the cloned AppConfig to the context.

src/foam/nanos/auth/Group.js
```diff
@@ -128,6 +143,41 @@ foam.CLASS({
...
+          javaType: 'X'
+        }
+      ],
+      javaCode: `
+DAO userDAO         = (DAO) x.get("localUserDAO");
+DAO groupDAO        = (DAO) x.get("groupDAO");
+AppConfig config    = (AppConfig) ((AppConfig) x.get("appConfig")).fclone();
+
+Session session = x.get(Session.class);
+if ( session != null ) {
```
jlhughes, line 162: So how does this work for emails initiated by cron, when the session is null?  
I don't see any other relevant source in the PR?

ZacZhang: For emails initiated by cron, the session is not null, it's a session with userId 1. So for cron, it will check the system group url first,  if it's empty, then use the default appconfig url. I changed some code for ScheduledEmail.java for cron, but it's not in foam folder. I think I should create PR for foam first.

jlhughes (approved):

src/foam/nanos/auth/Group.js
jlhughes: re: ScheduledEmail.java - sure create a seperate PR if you like.

PR #1546, opened 2018-09-07
Auth Service: User permission check

zahnno:
Added a new method to the auth service which checks if the user passed in has the passed in permission attributed to its group.

### DIFF
--- a/src/foam/nanos/auth/AuthService.js
+++ b/src/foam/nanos/auth/AuthService.js
@@ -111,7 +111,7 @@ foam.INTERFACE({
     },
     {
       name: 'checkUser',
-      javaReturns: 'Boolean',
+      javaReturns: 'boolean',
       swiftReturns: 'Bool',
       swiftThrows: true,
       args: [
@@ -132,7 +132,7 @@ foam.INTERFACE({
     },
     {
       name: 'checkUserPermission',
-      javaReturns: 'Boolean',
+      javaReturns: 'boolean',
       swiftReturns: 'Bool',
       swiftThrows: true,
       args: [
@@ -152,25 +152,8 @@ foam.INTERFACE({
       ]
     },
     {
-      name: 'check',
-      javaReturns: 'Boolean',
-      swiftReturns: 'Bool',
-      swiftThrows: true,
-      args: [
-        {
-          name: 'x',
-          javaType: 'foam.core.X',
-          swiftType: 'Context'
-        },
-        {
-          name: 'permission',
-          javaType: 'String',
-        }
-      ]
-    },
-    {
       name: 'checkPermission',
-      javaReturns: 'Boolean',
+      javaReturns: 'boolean',
       swiftReturns: 'Bool',
       swiftThrows: true,
       args: [
@@ -187,7 +170,7 @@ foam.INTERFACE({
     },
     {
       name: 'check',
-      javaReturns: 'Boolean',
+      javaReturns: 'boolean',
       swiftReturns: 'Bool',
       swiftThrows: true,
       args: [
--- a/src/foam/nanos/auth/AuthService.js
+++ b/src/foam/nanos/auth/AuthService.js
@@ -110,6 +110,27 @@ foam.INTERFACE({
       ]
     },
     {
+      name: 'checkUserPermission',
+      javaReturns: 'Boolean',
+      swiftReturns: 'Bool',
+      swiftThrows: true,
+      args: [
+        {
+          name: 'x',
+          javaType: 'foam.core.X',
+          swiftType: 'Context'
+        },
+        {
+          name: 'user',
+          javaType: 'foam.nanos.auth.User'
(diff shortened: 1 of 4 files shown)


src/foam/nanos/auth/AuthService.js
```diff
@@ -109,6 +109,27 @@ foam.INTERFACE({
         }
       ]
     },
+    {
+      name: 'checkUserPermission',
```
kgrgreer: The regular checkPermission() method takes a Permission, which is why it's called checkPermission. The version that takes a String is just called check(). This method just takes a String, so it should be called checkUser(). Or maybe, checkUser() (or maybe userCheck()). Then, for consistency, you might want a second version called userCheckPermission() which takes a Permission. Or name them checkForUser() and checkPermissionForUser().

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -167,6 +167,46 @@ public User loginByEmail(X x, String email, String password) throws Authenticati
...
   }
 
+  /**
+    Checks if the user passed into the method has the passed
+    in permission attributed to it by checking their group.
+    No check on User and group enabled flags.
+  */
+  public Boolean checkUserPermission(foam.core.X x, User user, String permission) {
+    // check whether user has permission to check user permissions
+    if ( ! check(x, permissionCheckPrivilege_) ) {
```
kgrgreer: Where is permissionCheckPrivilege defined? If it's a constant, it should be capitalized.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -167,6 +167,46 @@ public User loginByEmail(X x, String email, String password) throws Authenticati
...
+      return false;
```
kgrgreer: You should throw an authorization exception rather than return false.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -167,6 +167,46 @@ public User loginByEmail(X x, String email, String password) throws Authenticati
     return user;
   }
 
+  /**
+    Checks if the user passed into the method has the passed
+    in permission attributed to it by checking their group.
+    No check on User and group enabled flags.
+  */
+  public Boolean checkUserPermission(foam.core.X x, User user, String permission) {
```
kgrgreer: return type should be 'boolean' not Boolean.
Fix the other two check() methods while you're at it.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -31,6 +31,8 @@
   protected DAO groupDAO_;
   protected DAO sessionDAO_;
 
+  private final String PermissionPrivilege_ = "permission.privilege.*";
```
kgrgreer: Make public and rename to CHECK_USER_PERMISSION.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -167,11 +169,51 @@ public User loginByEmail(X x, String email, String password) throws Authenticati
...
 
+  /**
+    Checks if the user passed into the method has the passed
+    in permission attributed to it by checking their group.
+    No check on User and group enabled flags.
+  */
+  public boolean checkUserPermission(foam.core.X x, User user, Permission permission) {
+    // check whether user has permission to check user permissions
+    if ( ! check(x, PermissionPrivilege_) ) {
+      throw new AuthenticationException();
```
kgrgreer: Sorry, my mistake. Should be an AuthorizationException.

import foam.nanos.auth.AuthorizationException;

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -31,6 +31,8 @@
   protected DAO groupDAO_;
   protected DAO sessionDAO_;
 
+  public final static String CHECK_USER_PERMISSION = "permission.privilege.*";
```
kgrgreer: Actually, permission.privilege.* isn't a very good permission name. Maybe something like service.auth.checkUser.

src/permissions
```diff
@@ -2,3 +2,4 @@ p({"class":"foam.nanos.auth.Permission", "id":"*",         "description":"Do any
 p({"class":"foam.nanos.auth.Permission", "id":"menu.auth", "description":"Perform authentication related configuration" })
 p({"class":"foam.nanos.auth.Permission", "id":"service.*", "description":"Global permission for reading of services"})
 p({"class":"foam.nanos.auth.Permission", "id":"service.read.*", "description":"Global permission for reading of services from a DAO"})
+p({"class":"foam.nanos.auth.Permission","id":"permission.privilege.*"})
```
kgrgreer: Add a 'description'.

PR #1547, opened 2018-09-07
Validation improvements

TW80000:
- If a property on a model has a `validateObj` function, then use that to validate the UI input when that property is used as a viewspec.
- If the value in the ui element is invalid, show a popup with a message returned by the `validateObj` function. Also change the border colour to red to indicate that the value is invalid.

Before entering any text:
screen shot 2018-09-07 at 1 48 00 pm

After entering invalid text and changing focus away from the input:
screen shot 2018-09-07 at 1 48 11 pm

After clicking anywhere on the screen to hide the popup, border stays red to show invalid state:
screen shot 2018-09-07 at 1 48 20 pm

Part of the solution to https://github.com/nanoPayinc/NANOPAY/issues/3451

### DIFF
--- a/src/foam/u2/tag/Input.js
+++ b/src/foam/u2/tag/Input.js
@@ -20,7 +20,20 @@ foam.CLASS({
   name: 'Input',
   extends: 'foam.u2.View',
 
-  css: '^:read-only { border: none; background: rgba(0,0,0,0); }',
+  css: `
+    ^:read-only {
+      border: none;
+      background: rgba(0,0,0,0);
+    }
+
+    input.invalid {
+      border-color: red;
+    }
+  `,
+
+  requires: [
+    'foam.u2.PopupView'
+  ],
 
   properties: [
     [ 'nodeName', 'input' ],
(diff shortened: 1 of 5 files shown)


kgrgreer:
Will review.

src/foam/u2/tag/Input.js
```diff
@@ -56,7 +82,22 @@ foam.CLASS({
...
+        the value is invalid, the function should return a string containing a
+        message telling the user about the invalid value. If the argument is
+        valid, the function should not return anything, which is the same as
+        returning undefined.
+
+        Usually instead of setting this directly, you will define a
+        'validateObj' function on a property. Doing that will automatically set
+        this property to the validateObj function you defined on the property.
+      `,
+      value: function() {}
```
zahnno: Add a console error if the value being set isn't a function.

src/foam/u2/tag/Input.js
```diff
@@ -32,6 +45,19 @@ foam.CLASS({
           this.warn('Set Input data to non-primitive:' + d);
           return o;
         }
+
+        if ( typeof this.validateValue === 'function' && o != undefined ) {
```
zahnno: validateValue can always equate a function if you make it a requirement on the prop.

TW80000:
All comments addressed. Ready for review again.

TW80000:
Closing because I don't think we want this anymore.

PR #1548, opened 2018-09-10
Don't render top nav until logged in

TW80000:
Currently the top nav loads even when the user isn't logged in. This
means that the initE method for child views of the top navigation run
before user is set in the context. This makes it so that the views in
the top nav don't load until the user is properly set in the context.

Required by https://github.com/nanoPayinc/NANOPAY/pull/3492

### DIFF
--- a/src/foam/nanos/u2/navigation/TopNavigation.js
+++ b/src/foam/nanos/u2/navigation/TopNavigation.js
@@ -99,7 +99,7 @@ foam.CLASS({
 
   messages: [
     {
-      name: 'Greeting',
+      name: 'GREETING',
       message: 'Welcome'
     }
   ],
@@ -118,7 +118,7 @@ foam.CLASS({
         .start()
           .addClass('welcome-label')
           .hide(this.loginSuccess$)
-          .add(this.Greeting)
+          .add(this.GREETING)
         .end();
 
       this.user.id$.sub(this.userLoggedIn);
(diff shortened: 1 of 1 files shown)


kgrgreer:
reviewing...

src/foam/nanos/u2/navigation/TopNavigation.js
```diff
@@ -101,24 +101,33 @@ foam.CLASS({
...
+    },
+    'welcome'
   ],
 
   methods: [
     function initE() {
       this
         .addClass(this.myClass())
+        .start('div', null, this.welcome$)
+          .add('Welcome').addClass('welcome-label').hide(this.loginSuccess$)
```
zahnno: Let's make this 'Welcome' string into a messages prop.

TW80000: Done.

TW80000:
All comments addressed. Waiting on approval or new comments.

src/foam/nanos/u2/navigation/TopNavigation.js
```diff
@@ -97,6 +97,13 @@ foam.CLASS({
     }
   `,
 
+  messages: [
+    {
+      name: 'Greeting',
```
kgrgreer: Should this be GREETING (or WELCOME). But don't we name messages like constants?

zahnno: @kgrgreer If we're following suit to our previous messages then no, we use PascalCase.

TW80000: I looked through the project for how we use messages and we do it both ways in many different places.

zahnno:
@kgrgreer Please address

PR #1555, opened 2018-09-12
Added AbstractListPropertyInfo

kirkeaton:
Need to add this to override the compare method.

### DIFF
--- a/src/foam/core/AbstractListPropertyInfo.java
+++ b/src/foam/core/AbstractListPropertyInfo.java
@@ -1,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.core;
+
+import java.util.List;
+
+public abstract class AbstractListPropertyInfo
+  extends AbstractObjectPropertyInfo
+{
+
+  @Override
+  public int compare(Object o1, Object o2) {
+    Object v1 = this.get(o1);
+    Object v2 = this.get(o2);
+
+    if ( v1 == null & v2 == null ) return 0;
+    if ( v2 == null ) return 1;
+    if ( v1 == null ) return -1;
+
+    List l1 = (List) v1;
+    List l2 = (List) v2;
+
+    // check list sizes
+    if ( l1.size() > l2.size() ) return 1;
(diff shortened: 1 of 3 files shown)


src/foam/core/AbstractListPropertyInfo.java
```diff
@@ -0,0 +1,34 @@
...
+
+    List l1 = (List) v1;
+    List l2 = (List) v2;
+
+    // check list sizes
+    if ( l1.size() > l2.size() ) return 1;
+    if ( l1.size() < l2.size() ) return -1;
+
+    // check if lists are equal, if not return -1 suggesting the 2nd list is newer
+    return ( l1.containsAll(l2) && l2.containsAll(l1) ) ? 0 : -1;
```
kgrgreer: This isn't stable because if compare(a, b) == -1 then it isn't necessary that compare(b, a) == 1. If used for sorting this will lead to a random order. If nothing else, you can do:

if ( l1.containsAll(l2) && l2.containsAll(l1) ) return 0;
return Integer.compare(l1.hashCode(), l2.hashCode()); 

This will ensure that comparing l1 and l2 is stable.

src/foam/java/refinements.js
```diff
@@ -1084,14 +1084,18 @@ foam.CLASS({
...
     ['javaFactory', 'return new java.util.ArrayList();'],
-    ['javaJSONParser', 'new foam.lib.json.ListParser()']
   ],
 
   methods: [
     function createJavaPropertyInfo_(cls) {
       var info = this.SUPER(cls);
 
+      var compare = info.getMethod('compare');
+      compare.body = 'return super.compare(o1, o2);';
```
kgrgreer: Is this necessary?

kirkeaton: Yes otherwise it will use the SafetyUtil.compare method which fails for Lists

src/foam/core/AbstractListPropertyInfo.java
kirkeaton: Changed to compare the hash codes.

src/foam/java/refinements.js
kgrgreer: Okay, maybe add a comment to explain that, since it won't be obvious to future developers reading the code.

kirkeaton: Done

PR #1558, opened 2018-09-14
Hide all warnings related to Java code generation

TW80000:
I tried to do this before with a blacklist approach but that didn't end
up catching all of the errors. This commit changes the approach to one
that should catch all cases, so long as the naming convention is
followed.

### DIFF
--- a/src/foam/core/debug.js
+++ b/src/foam/core/debug.js
@@ -281,17 +281,9 @@ foam.CLASS({
 
   methods: [
     function unknownArg(key, value) {
-      if ( key == 'class' ) return;
-
       // Temporarily disable warnings related to generating Java code.
-      var blackList = [
-        'javaThrows',
-        'javaReturns',
-        'javaCode'
-      ];
-      if ( ! blackList.some((keyword) => key.includes(keyword)) ) {
-        this.warn('Unknown property ' + this.cls_.id + '.' + key + ': ' + value);
-      }
+      if ( key == 'class' || key.startsWith('java') ) return;
+      this.warn('Unknown property ' + this.cls_.id + '.' + key + ': ' + value);
     },
 
     function describe(opt_name) {


mcarcaso:
Why not add all of the java properties we know about here? Suppressing java* means I won't get a warning if I accidentally use something like `javaReturn` instead of `javaReturns` when generating java code.

Also, don't we already have a means to suppress these warnings by using the NoJava/NoSwift refinements? I think we're better off not suppressing warnings here and adding those refinements for now. @adamvy has the `pressure-wash` branch that, once merged, should remove the need for all this stuff but it's a pretty intrusive change so it needs to be merged with some care.

TW80000:
> Why not add all of the java properties we know about here? Suppressing java* means I won't get a warning if I accidentally use something like `javaReturn` instead of `javaReturns` when generating java code.

That makes sense, I didn't think of that.

> Also, don't we already have a means to suppress these warnings by using the NoJava/NoSwift refinements?

I wasn't aware of those refinements before. I just had a look at those files and I think I get it now. By not giving NoJava.js the 'web' flag in files.js, those models don't have the Java-related properties on the web client.

In light of that new information, I'll just close this PR and if I have time to back later and move things to NoJava I will. Thanks for the info, @mcarcaso.

mcarcaso:
> I wasn't aware of those refinements before. I just had a look at those files and I think I get it now. By not giving NoJava.js the 'web' flag in files.js, those models don't have the Java-related properties on the web client.

It has the `js` flag so it won't get included when we're doing java or swift generation but it will get included when we're loading foam for web/nodejs. What it does is essentially add dummy properties to the models so the values get stored but nobody really uses them. Whereas the equivalent refinements in foam/java/refinements.js are actually meaningful to the generation that gets provided in that file.

With `pressure-wash` we're going to remove these flags from files.js completely (and nuke these NoJava/NoSwift files) and we'll rely on model level flags to enable/disable features. Flags should be seen as controlling features rather than platforms. For example, the `java` flag's meaning would be seen as enabling the java code generation feature. The build tool will allow us to strip out features that we don't need and all associated properties.

Explaining this is kinda tough so feel free to ping me if you want me to elaborate on any of this.

PR #1562, opened 2018-09-18
Update PasswordView with customized password visibility icon

Henry0422:
To enable the passwordIcon feature when initializing password property:
```
{
   class: 'Password',
   name: 'password',
   view: { class: 'foam.u2.view.PasswordView', passwordIcon: true }
}
```

In a particular view, customize the location of the visibility icon using CSS:
```
^ .input-image {
   position: absolute;
   width: XXpx;
   height: XXpx;
   bottom: XXpx;
   right: XXpx;
}
```

If the view (such as `SignInView` & `ResetView`) needs `full-width-input`, add customized css class`^ .full-width-input-password`


Henry0422:
@kgrgreer I am not sure how to get the relative path of an icon in foam2.

After I add an icon file under `foam2/src/foam/u2` & put `{"class":"foam.nanos.servlet.ServletMapping","className":"foam.nanos.servlet.ImageServlet","pathSpec":"/images/*",initParameters:{"paths":"foam2/src/foam/u2/images"}}` at the end of the `http` in the  foam2 services file, the web page still cannot load the image based on the relative path (`/images/baseline-visibility_off-24px.svg`)

src/foam/nanos/auth/resetPassword/ResetView.js
```diff
@@ -167,9 +176,9 @@ foam.CLASS({
         .start().addClass('Reset-Password').add("Reset Password").end()
         .start().addClass('Message-Container')
           .start().addClass('newPassword-Text').add("New Password").end()
-          .start(this.NEW_PASSWORD).addClass('full-width-input').end()
+          .start(this.NEW_PASSWORD).end()
           .start().addClass('confirmPassword-Text').add("Confirm Password").end()
-          .start(this.CONFIRM_PASSWORD).addClass('full-width-input').end()
+          .start(this.CONFIRM_PASSWORD).end()
```
kgrgreer: You can just do .add(this.CONFIRM_PASSWORD), which is shorter than start() and end().

src/foam/u2/view/PasswordView.js
```diff
@@ -18,7 +18,85 @@
 foam.CLASS({
   package: 'foam.u2.view',
   name: 'PasswordView',
-  extends: 'foam.u2.tag.Input',
+  extends: 'foam.u2.View',
 
-  properties: [ [ 'type', 'password' ] ]
+  css: `
+    .input-field-container {
```
kgrgreer: These are global names. Try not to pollute the global namespace, so use ^ and myClass() instead.

src/foam/u2/view/PasswordView.js
```diff
@@ -18,7 +18,85 @@
...
+    },
+    {
+      class: 'Boolean',
+      name: 'passwordInvisible',
+      value: true
+    },
+    {
+      class: 'Boolean',
+      name: 'passwordIcon',
+      defaultValue: false
```
kgrgreer: defaultValue is from FOAM1, use value: instead. However, false is already the default so not needed.

src/foam/nanos/auth/SignInView.js
```diff
@@ -123,7 +132,7 @@ foam.CLASS({
           .start().addClass('label').add("Email Address").end()
           .start(this.EMAIL).addClass('full-width-input').end()
           .start().addClass('label').add("Password").end()
-          .start(this.PASSWORD).addClass('full-width-input').end()
+          .start(this.PASSWORD).end()
```
kgrgreer: Here you can also just use .add() instead of start().end().

src/foam/u2/view/PasswordView.js
```diff
@@ -18,7 +18,85 @@
...
+          type: this.type
+        }, this.inputElement$).
+          addClass('input-field').addClass('full-width-input-password').
+        end().
+        start('img').show(this.passwordIcon$).addClass('input-image').
+          attr('src', this.visibilityIcon$).on('click', this.visible).
+        end().
+      end();
+
+      this.inputElement.attrSlot(
```
kgrgreer: You don't need to do this. the start('input') is mapped to a foam.u2.tag.Input element which already supports this logic. Just set it's onKey property and then bind to its data$.

src/foam/u2/view/PasswordView.js
```diff
@@ -18,7 +18,77 @@
 foam.CLASS({
   package: 'foam.u2.view',
   name: 'PasswordView',
-  extends: 'foam.u2.tag.Input',
+  extends: 'foam.u2.View',
 
-  properties: [ [ 'type', 'password' ] ]
```
TW80000, line 23: We still need this for the `type` attribute on the HTML input element to be set correctly.

```JavaScript
{
  class: 'String',
  name: 'type',
  value: 'password'
}
```

TW80000 (changes requested):

TW80000 (approved):

PR #1568, opened 2018-09-22
Add an fclone method that takes a context

mcarcaso:
So we can call fclone similar to how we call it in js. If no context is given, the object being clone's context is used which is a change of behaviour from what currently happens but, imo is the more correct behavior and also matches JS.

### DIFF
--- a/src/foam/core/AbstractFObject.js
+++ b/src/foam/core/AbstractFObject.js
@@ -82,6 +82,10 @@ foam.CLASS({
             verifier.initVerify(key);
             return this.verify(signature, verifier);
           }
+
+          public foam.core.FObject fclone() {
+            return fclone(getX());
+          }
         `);
       }
     }
@@ -174,10 +178,12 @@ foam.CLASS({
     },
     {
       name: 'fclone',
+      args: [ { name: 'x', javaType: 'foam.core.X' } ],
       javaReturns: 'foam.core.FObject',
       javaCode: `
         try {
           FObject ret = getClass().newInstance();
+          ret.setX(x);
           List<PropertyInfo> props = getClassInfo().getAxiomsByClass(PropertyInfo.class);
           for( PropertyInfo prop : props ) {
             if ( ! prop.isSet(this) ) continue;
--- a/src/foam/core/FObject.java
+++ b/src/foam/core/FObject.java
@@ -17,6 +17,7 @@ public interface FObject extends
   ClassInfo getClassInfo();
   FObject copyFrom(FObject obj);
   FObject fclone();
+  FObject fclone(foam.core.X x);
   FObject deepClone();
   FObject shallowClone();
   Map diff(FObject obj);


kgrgreer:
We don't want objects stored in DAO's to have contexts since this causes both security bugs and memory leaks. I would rather FObject not be ContextAware by default. We added contexts to the Relationship methods so that they don't have to rely on getX().

mcarcaso:
Having FObjects not be context aware is a pretty drastic departure from the
way FObjects work in JS and it would limit our ability to easily import and
export properties and methods. I feel like that’d be a step back in terms
of developer efficiency.

If you don’t want a DAO to store any context wouldn’t we be able to
decorate it with a Contextualizing DAO?

jlhughes:
@mcarcaso @adamvy 
We would like to do the following instead:
1. Remove ContextAware from FObject
2. create a HoneyPotContext which is assigned on MDAO.put to log any post put uses of an objects Context. 

For ContextAwareSupport, add back traits and then add ContextAwareSupport as a trait, which a model would explicitly add if required.

mcarcaso:
If we're going to start down this road we should think how this is going to
affect android and consider the possibility of having separate generators
for server java and android java otherwise development on Android could get
really annoying when some of the features we have in JS are desired. I
don't like the idea of having two separate generators but that might be
inevitable if one platform requires FObjects to act so differently than
we're used to in JS.

Personally, I'd rather see what happens if we ditched these context
oriented methods all together and see if we can design things in such a way
that the context isn't as big of a concern as it is currently. That
approach could require decorating service DAOs on the fly with decorators
that are created in the context of the request and then thrown away when
the request is processed rather than having long lived decorators that have
context oriented methods.

kgrgreer:
I don't have an issue with GUI components or services being ContextAware, but having data-objects be CA has proven to be very dangerous (because contexts represent permissions), difficult to debug (because of different timing resulting in different contexts being trapped), and causing memory leaks. We're proposing using a trait to add a CA support when needed, but not having it by default. Also, if we model imports/exports, then it could be added automatically. But, in general, having ContextOriented methods is much less error prone as then you're assured of what context you're using and you don't have contexts leaking between sessions/clients. No object stored in a DAO should ever have a Context, which is why the DAO clones objects going in.

> That approach could require decorating service DAOs on the fly with decorators
that are created in the context of the request and then thrown away when
the request is processed rather than having long lived decorators that have
context oriented methods.

People try to do this with Sessions/Contexts stored in ThreadLocal holders but then this also causes issues when you start using thread-pools, multi-threading, async, or want to create sub-contexts. Just making methods explicitly CO is the easiest and safest in my experience.

mcarcaso:
Hmm yeah I guess there's a place for both context aware FObjects and
context oriented methods. I was trying to think what either extreme would
look like (i.e. everything context oriented and nothing context aware vs.
nothing context oriented and everything context aware) but neither extreme
is really sufficient. I'm convinced :)

PR #1573, opened 2018-09-26
Journal modifications, Routing Journal, RoutingJDAO

kirkeaton:
1) Modified Journal interface to not extend Sink interface
2) Modified Journal interface to make put, remove, and replay context oriented
3) Removed writeComment from JDAO and put it in FileJournal
4) Simplified & cleaned up code


jlhughes (approved):
Looks good.
Do we need to impose any naming restrictions on service 'name' now that it is more important?

kirkeaton:
@jlhughes The only restriction would be to make sure it doesn't include a `.` in the name as it will create issues with the routing.

kgrgreer (commented):
LGTM

PR #1581, opened 2018-10-01
SUGAR - non DAO service end point

minsunny:
#3282

Service Unified GAteway Relay called "SUGAR" is a webAgent which allows to call services not DAOs with its arguments.

Requirements :

Service List
Methods List in the service according to the step 1
Parameters for the method according to the step 2
Providing Parameters Info
Creating dynamic URL
Add modelled Exceptions


src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.dig',
+  name: 'SUGAR',
+
+  documentation: 'Service Unified GAteway Relay - Perform non-DAO operations against a web service',
```
deejes: Case error in gateway

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+      name: 'argumentInfo',
+      documentation: 'Set the arguments Info of the method',
+      postSet: function() {
+      var self = this;
+
+      for ( let j = 0 ; j < this.argumentInfo.length ; j++ ) {
+        this.argumentInfo[j].sub(function(argInfo) {
+          self.flag = !self.flag;
+        });
+      }
```
deejes: function logic needs indentation

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+          url += "interfaceName=" + interfaceName;
+        }
+        if ( hiddenMethod ) {
+          url += query ? "&" : "?";
+          query = true;
+          url += "method=" + hiddenMethod;
+        }
+
+        for ( let j = 0 ; j < argumentInfo.length ; j++ ) {
+        var paramUrl = "";
```
deejes: indentation for function body required

src/foam/nanos/dig/SUGAR.js
jlhughes: That's intentional as it is the GA in SUGAR.

jlhughes (commented):
Please add your tests to this PR.  
Also, if you don't have a test for 'flag', at least add a test with comments regarding what the flag format would be.

src/foam/nanos/boot/NSpec.js
```diff
@@ -113,6 +113,11 @@ foam.CLASS({
       name: 'serviceClass',
       displayWidth: 80
     },
+    {
+      class: 'String',
```
jlhughes: Please add documentation as to the purpose and use of this field.

src/foam/nanos/dig/Argument.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.dig',
```
jlhughes, line 8: Please document the class and properties as to purpose and use.

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,173 @@
...
+import javax.servlet.http.HttpServletResponse;
+import foam.nanos.http.HttpParameters;
+import java.nio.CharBuffer;
+import foam.util.SafetyUtil;
+import foam.lib.json.Outputter;
+import java.lang.Exception;
+import java.lang.reflect.*;
+import java.lang.Exception;
+
+public class SugarWebAgent
```
jlhughes, line 28: Please document.

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+        if ( ! of ) return;
+
+        this.interfaceName = of.id.toString();
+
+        var methods = of.getOwnAxiomsByClass(foam.core.Method);
+        //var methodName = methods.map(function(m) { return m.name; }).sort();
+
+        var filteredMethod =
+          methods.filter(function(fm) {
+            for ( var j = 0; j < fm.args.length; j++ ) {
```
kgrgreer: spaces before ;'s

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+               if ( fm.args[j].javaType.toString() == "foam.core.X" ) {
+                  return false;
+               }
+               return true;
+             }
+          }).map(function(m) { return m.name; }).sort();
+
+        if ( filteredMethod.length > 0) {
+          for ( var i = 0; i < methods.length; i++ ) {
```
kgrgreer: spaces before ;'s

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+          var of = this.lookup(service.cls_.getAxiomByName('delegate').of);
+
+          if ( ! of ) return;
+
+          var methods = of.getOwnAxiomsByClass(foam.core.Method);
+          //var methodNames = methods.map(function(m) { return m.name; }).sort();
+
+          var filteredMethod =
+            methods.filter(function(fm) {
+              for ( var j = 0; j < fm.args.length; j++ ) {
```
kgrgreer: spaces before ;'s

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+
+          if ( ! service.cls_.getAxiomByName('delegate') ) return;
+
+          var of = this.lookup(service.cls_.getAxiomByName('delegate').of);
+
+          if ( ! of ) return;
+
+          var methods = of.getOwnAxiomsByClass(foam.core.Method);
+
+          for ( var i = 0; i < methods.length; i++ ) {
```
kgrgreer: spaces before ;'s

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+    },
+    {
+      class: 'String',
+      name: 'hiddenMethod',
+      documentation: 'to use for a url',
+      hidden: true
+    },
+    {
+      class: 'URL',
+      // TODO: appears not to work if named 'url', find out why.
```
kgrgreer, line 203: Please open a ticket to track this issue.

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,257 @@
...
+
+            if ( k == Number(index) ) url += paramUrl;
+
+            if ( argumentInfo[k].value != "") {
+              url += query ? "&" : "?";
+              url += argumentInfo[k].name + "=" + argumentInfo[k].value;
+            }
+          }
+        } else { // use this flag to give a change event for argumentInfo (FObjectArray)
+          for ( let k = 0 ; k < argumentInfo.length ; k++ ) {
```
kgrgreer: Us 'var' instead of 'let' if not trapping value in a closure (function).

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,173 @@
+package foam.nanos.dig;
+
+import foam.lib.json.JSONParser;
```
kgrgreer, line 11: sort imports

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,173 @@
...
+      }
+
+      Class c = Class.forName(interfaceName);
+
+      Method m[] = c.getMethods();  // get Methods' List
+
+      Class[] paramTypes = null; // for picked Method's parameters' types
+      Object arglist[] = null; // to store each parameters' values
+
+      for ( int k = 0; k < m.length; k++ ) {
```
kgrgreer: spaces before ;'s, here and elsewhere

src/foam/nanos/dig/exception/EmptyParameterException.js
```diff
@@ -0,0 +1,33 @@
...
+  extends: 'foam.nanos.dig.exception.DigErrorMessage',
+
+  properties: [
+    {
+      class: 'String',
+      name: 'status',
+      value: '400'
+    },
+    {
+      class: 'String',
```
kgrgreer: Should this be an Int? Same as status?

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,256 @@
...
+
+        if ( ! service ) return;
+
+        if ( ! service.cls_.getAxiomByName('delegate') ) return;
+
+        var of = this.lookup(service.cls_.getAxiomByName('delegate').of);
+
+        if ( ! of ) return;
+
+        this.interfaceName = of.id.toString();
```
shreymahendru, line 67: You have the logic of these `if` conditions more than one place.... you should abstract that into a function of its own.

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,256 @@
...
+
+        var methods = of.getOwnAxiomsByClass(foam.core.Method);
+        //var methodName = methods.map(function(m) { return m.name; }).sort();
+
+        var filteredMethod =
+          methods.filter(function(fm) {
```
shreymahendru, line 72: Have a function that abstracts the filtering of methods, that should return a single method since you are only concerned with the 0th element of the filteredMethod. (Correct me if I am wrong)

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,256 @@
...
+               if ( fm.args[j].javaType.toString() == "foam.core.X" ) {
+                  return false;
+               }
+               return true;
+             }
+          }).map(function(m) { return m.name; }).sort();
+
+        if ( filteredMethod.length > 0 ) {
+          for ( var i = 0 ; i < methods.length ; i++ ) {
+            if ( methods[i].name == filteredMethod[0] ) {
```
shreymahendru: Not sure if we support overriding on services' method, if we do, filtering would be a problem just by name. 
If we don't, why don't you just break the loop once you find the correct method with the same name as filteredmethod?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,177 @@
...
+          .setMessage("Empty Method Name")
+          .build();
+        outputException(x, resp, "JSON", out, error);
+
+        return;
+      }
+
+      Class c = Class.forName(interfaceName);
+
+      Method m[] = c.getMethods();  // get Methods' List from the class
```
shreymahendru: you might want to use more descriptive names for your variables

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,177 @@
...
+
+              return;
+            }
+
+            paramTypes[j] = pArray[j].getType();
+            arglist[j] = p.getParameter(pArray[j].getName());
+
+            logger.debug(pArray[j].getName() + " :   " + p.getParameter(pArray[j].getName()));
+
+            // casting and setting according to parameters type
```
shreymahendru: Switch would be perfect and more efficient/readable for this

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,177 @@
...
+import java.lang.Exception;
+import java.lang.reflect.*;
+import javax.servlet.http.HttpServletResponse;
+
+public class SugarWebAgent
+  implements WebAgent
+{
+  public SugarWebAgent() {}
+
+  public void execute(X x) {
```
shreymahendru, line 33: Please divide the execute function into smaller functions with descriptive names (something like steps), this makes it more readable/ enjoyable for the second dev.

src/foam/nanos/dig/SugarWebAgent.java
minsunny: conditional expressions are different depending on primitive types or not though..?

src/foam/nanos/dig/SUGAR.js
shreymahendru: `array.find` would work here if you are trying to search a **single** method by name.

src/foam/nanos/dig/SugarWebAgent.java
shreymahendru: Sorry. I don't understand... But you can have 
```
String typeName = pArray[j].getType().getCanonicalName() 
switch(typeName) {
case "double": 
arglist[j] = Double.parseDouble(p.getParameter(pArray[j].getName()));
case "boolean":
arglist[j] = Boolean.parseBoolean(p.getParameter(pArray[j].getName()));
....
default:
DigErrorMessage
}
```

kgrgreer: In the future, we're going to switch method arguments to FOAM modelled types, the same as for properties. At that point we'll be able to just use the parser provided by the PropertyInfo, which is better than either switch or many if/elses. Until then, a switch would be better than ifelseifelse's. Another method would be to define a static hashmap of String to Parser mappings and then just lookup the parser to use.

src/foam/nanos/dig/SUGAR.js
```diff
@@ -0,0 +1,259 @@
...
+      },
+      postSet: function() {
+        var service = this.__context__[this.serviceKey];
+
+        if ( ! service ) return;
+
+        if ( ! service.cls_.getAxiomByName('delegate') ) {
+          this.interfaceName = "";
+          return;
+        } else {
```
kgrgreer: No need for 'else' block since you've already returned in the 'if'. Just remove the 'else' and move the code in that block two spaces to the left.

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,194 @@
...
+      }
+
+      Class[] paramTypes = null; // for picked Method's parameters' types
+      Object arglist[] = null; // to store each parameters' values
+
+      if ( class_ != null ) {
+        Method method_[] = class_.getMethods();  // get Methods' List from the class
+
+        for (int k = 0; k < method_.length; k++) {
+          if (method_[k].getName().equals(methodName)) { //found picked Method
```
kgrgreer: fix spacing

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,194 @@
...
+
+            logger.debug("service : " + serviceName);
+            logger.debug("methodName : " + method_[k].getName());
+
+            Parameter[] pArray = method_[k].getParameters();
+            paramTypes = new Class[pArray.length];
+            arglist = new Object[pArray.length];
+
+            for (int j = 0; j < pArray.length; j++) {  // checking the method's each parameter
```
kgrgreer: fix spacing

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,194 @@
...
+                  outputException(x, resp, "JSON", out, error);
+
+                  return;
+              }
+            }
+          }
+        }
+      }
+
+      executeMethod(x, resp, out, class_, serviceName, methodName, paramTypes, arglist);
```
kgrgreer: Do we want to call this if class_ is null?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,194 @@
...
+
+      outputterJson.output(declaredMethod_.invoke(x.get(serviceName), arglist));
+      out.println(outputterJson);
+    } catch (InvocationTargetException e) {
+      DigErrorMessage error = new GeneralException.Builder(x)
+        .setMessage("InvocationTargetException: " + e.getTargetException().getMessage())
+        .build();
+      outputException(x, resp, "JSON", out, error);
+
+      return;
```
kgrgreer: No need for return.

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -0,0 +1,194 @@
...
+    } catch (Exception e) {
+      DigErrorMessage error = new GeneralException.Builder(x)
+        .setMessage("Exception: " + e.getMessage())
+        .build();
+      outputException(x, resp, "JSON", out, error);
+
+      return;
```
kgrgreer: No need for return.

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -192,4 +178,81 @@ protected void outputException(X x, HttpServletResponse resp, String format, Pri
...
+              if ( prop.getValueClass().toString().equals("class [Ljava.lang.String;") ) {  //String[]
+                prop.set(clsObj, p.getParameterValues(prop.getName()));
+              } else {
+                prop.set(clsObj, p.getParameter(prop.getName()));
+              }
+
+              try {
+                fieldList[m].set(clsObj, prop);
+              } catch (IllegalAccessException e) {
+                e.printStackTrace();
```
jlhughes: Can this be logged?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -192,4 +178,81 @@ protected void outputException(X x, HttpServletResponse resp, String format, Pri
...
+              }
+            }
+          }
+        }
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
```
jlhughes: Can this be logged?

jlhughes (commented):
Changes look OK. 
Couple comments regarding logging.

src/foam/nanos/dig/Argument.js
```diff
@@ -0,0 +1,58 @@
...
+      name: 'value'
+    },
+    {
+      class: 'FObjectProperty',
+      of: this.javaType,
+      name: 'objectType',
+      label: 'Object Value',
+      documentation: 'Parameters input value (Object Type Parameters)',
+      factory: function() {
+        var model = this.lookup(this.javaType.toString(), true);
```
kgrgreer: javaType is already a String, so you don't need to call toString().

src/foam/nanos/dig/Argument.js
```diff
@@ -0,0 +1,58 @@
...
+     visibility: foam.u2.Visibility.RO
+    },
+    {
+      class: 'String',
+      documentation: 'Parameters input value',
+      name: 'value'
+    },
+    {
+      class: 'FObjectProperty',
+      of: this.javaType,
```
kgrgreer: This line can't work since 'this' isn't defined here.

src/foam/nanos/dig/Argument.js
```diff
@@ -0,0 +1,58 @@
...
+      name: 'objectType',
+      label: 'Object Value',
+      documentation: 'Parameters input value (Object Type Parameters)',
+      factory: function() {
+        var model = this.lookup(this.javaType.toString(), true);
+
+        if ( this.javaType != '' && ( this.javaType != 'String' && this.javaType != 'double' && this.javaType != 'boolean' && this.javaType != 'long' ) )
+          return model.create(null, this);
+        else return null;
```
kgrgreer: No need for 'else' since you've already returned.

src/foam/nanos/dig/SUGAR.js
minsunny: done

src/foam/nanos/dig/SUGAR.js
minsunny: done

jlhughes (approved):
Approved, minus Shrey's comments, which are valid, but will have to be dealt with at a later time.

PR #1584, opened 2018-10-02
Agent Auth Services

zahnno:
# Agent Auth Service

The agent auth service is responsible for allowing users to act as others if permitted and places the requesting user as an agent of the system within the context.

A junction between users are created to determine if a user is permitted to act as another user.

*An AuthenticatedDecorator on the junctionDAO will have to be developed to extended any additional authentication logic.*

Along with the target & source Id on the junction object (UserUserJunction), a group is appended and is used to determine the list of permissions for the current user agent access to the system. When acting as another user, the user requesting to act is an "agent" within the context. The user being logged into becomes the "user" within the context, and the group associated to the junction object becomes the "group" within the context.

Changes to the auth services were made to account for placing a group within the context. This group is used throughout the rest of the auth service when checking permissions.

**Summary of flow**

1. Initial login on user.
2. Call to agent auth service actAs.
3. Lookup junction object associated to the context user and the super user being passed into the service. 
4. Clone super user and set super user group to the group associated to the junction object.
5. Freeze super user.
6. Clone context user and freeze.
7. Set super user as the context "user" and the current context user as an "agent".
8. Put session into DAO


src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
TW80000: Shouldn't this be 2018?

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,25 @@
+foam.INTERFACE({
```
TW80000, line 7: Please add license header.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,25 @@
...
+  package: 'foam.nanos.auth',
+  name: 'AgentAuthService',
+
```
TW80000, line 10: Please add documentation.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,25 @@
...
+  methods: [
+    {
+      name: 'ActAs',
```
TW80000: Methods should start with a lowercase letter.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,25 @@
...
+      args: [
+        {
+          name: 'x',
+          javaType: 'foam.core.X',
+          swiftType: 'Context'
+        },
+        {
+          name: 'sudoUser',
+          javaType: 'foam.nanos.auth.User'
+        }
```
TW80000, line 41: Should `swiftType` be included here?

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+  public AgentUserAuthService(X x) {
+    setX(x);
+  }
+
+  @Override
+  public void start() {
+    userDAO_     = (DAO) getX().get("localUserDAO");
+    groupDAO_    = (DAO) getX().get("groupDAO");
+    sessionDAO_  = (DAO) getX().get("sessionDAO");
+    agentJunctionDAO_  = (DAO) getX().get("agentJunctionDAO");
```
TW80000, line 36: Why don't we get these DAOs in the actual method so they're dependent on the context passed to the method?

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /* 
+    Sets the currently logged in user as an "agent" within the context &
+    sets "user" in the context to the passed in user. This allows users to
+    act on behalf of others while retaining information on the user.
+  */
+  public User ActAs(X x, User sudoUser) throws AuthenticationException {
+    User agent = (User) x.get("user");
+    User user = (User) userDAO_.find(sudoUser.getId());
```
TW80000: Why aren't we using context-oriented methods here and below? Eg: `find_` instead of `find`.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+
+    if ( agent == null ) {
+      throw new AuthenticationException();
+    }
+
+    if ( user == null || ! user.getEnabled() ) {
```
TW80000: If we don't allow sudo-ing as disabled users then we won't be able to sudo as a Business because businesses will be disabled.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+      throw new AuthenticationException("Super user doesn't exist.");
```
TW80000, line 53: `AuthenticationException` is only to be used when a user is not logged in. Something else should be used here, maybe `IllegalArgumentException`.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    Group group = (Group) groupDAO_.find(user.getGroup());
+
+    if ( group == null ) {
+      throw new AuthenticationException("User must exist within a group.");
```
TW80000: Please change the exception thrown here.

TW80000: Message should be consistent with the other error messages. For example, "Super user must ..." would be more consistent with the error message above.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // Finds the AgentUserJunction object to see if user can log in as the passed in user.
+    // Source users are permissioned to act as target users, not visa versa.
```
TW80000: Should be "permitted", not "permissioned" and "vice versa", not "visa versa".

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+
+    UserUserJunction permissionJunction = (UserUserJunction) agentJunctionDAO_.find(AND(
+      EQ(UserUserJunction.SOURCE_ID, agent.getId()),
+      EQ(UserUserJunction.TARGET_ID, user.getId())
+    ));
+
+    if ( permissionJunction == null ) {
+      throw new AuthenticationException("You don't have access to act as the requested user.");
```
TW80000: Please change exception and make message consistent.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // Junction object contains a group which has a unique set of permissions specific to the relationship.
+    Group ActingWithinGroup = (Group) groupDAO_.find(permissionJunction.getGroup());
```
TW80000: Local variable names should start with a lowercase letter.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+
+    if ( group != null && ! group.getEnabled() ) {
+      throw new AuthenticationException("User group disabled");
```
TW80000: Please fix exception type and message.

src/foam/nanos/auth/AuthService.js
```diff
@@ -23,6 +23,20 @@ foam.INTERFACE({
         }
       ]
     },
+    {
+      name: 'getCurrentAgent',
```
TW80000, line 27: Should this also be part of the `AgentAuthService` interface?

src/foam/nanos/auth/ClientAgentAuthService.js
```diff
@@ -0,0 +1,37 @@
+foam.CLASS({
```
TW80000, line 8: Please add license header.

src/foam/nanos/auth/ClientAgentAuthService.js
```diff
@@ -0,0 +1,37 @@
...
+  package: 'foam.nanos.auth',
+  name: 'ClientAgentAuthService',
+
```
TW80000, line 11: Please document this class.

src/foam/nanos/auth/User.js
```diff
@@ -345,3 +345,24 @@ foam.RELATIONSHIP({
     tableWidth: 120
   }
 });
+
+foam.RELATIONSHIP({
+  cardinality: '*:*',
+  sourceModel: 'foam.nanos.auth.User',
+  targetModel: 'foam.nanos.auth.User',
+  forwardName: 'agents',
+  inverseName: 'sudoUsers',
```
TW80000: I think this should be `superUsers` to make the terminology consistent.

src/foam/nanos/auth/User.js
```diff
@@ -345,3 +345,24 @@ foam.RELATIONSHIP({
...
+});
+
+foam.CLASS({
+  refines: 'foam.nanos.auth.UserUserJunction',
+
+  properties: [
+    {
+      class: 'Reference',
+      of: 'foam.nanos.auth.Group',
+      name: 'group'
```
TW80000, line 367: Somewhere in this file seems like a good spot to document how the agent feature works.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -129,10 +148,26 @@ public User login(X x, long userId, String password) throws AuthenticationExcept
     Session session = x.get(Session.class);
     session.setUserId(user.getId());
     session.setContext(session.getContext().put("user", user));
+    session.setContext(session.getContext().put("group", group));
```
TW80000, line 151: I don't understand why we're putting a group in the context. Everywhere that looks at the group looks at the group on the user in the context.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+    
+    Session session = x.get(Session.class);
+    session.setUserId(user.getId());
+    session.setContext(session.getContext().put("user", user));
+    session.setContext(session.getContext().put("agent", agent));
+    session.setContext(session.getContext().put("group", ActingWithinGroup));
```
TW80000: I thought the point was to set the group on the user in the context to the group referenced by the junction, not put the group into the context directly. Everywhere in the code that we check a user's group we're first getting the user from the context and then looking up their group. All of those places would have to be updated to check the context directly instead if we were to do this.

That's my biggest question about this PR.

TW80000 (changes requested):
I thought the point was to set the group on the user in the context to the group referenced by the junction, not put the group into the context directly. Everywhere in the code that we check a user's group we're first getting the user from the context and then looking up their group. All of those places would have to be updated to check the context directly instead if we were to do this.

That's my biggest question about this PR. The rest of my comments are pretty minor.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: I'm not 100% sure if my answer is correct but we don't want the context from the client to somehow affect the context in our services and visa versa. @kgrgreer might be able to expand on that.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: I don't see why we'd limit a business user from logging in.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: I disagree, AuthenticationExceptions are thrown when a form of auth fails, this is a user attempting to log in as another user. I think the AuthenticationException is fair in this use case.

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: I think there were two reasons:

1. That's what sudo-ing is for. The administrator of the business sudos into the business as opposed to them logging into the business directly.
2. The administrator's email address isn't tied to the business itself. That's good because if that administrator leaves, their email isn't tied to the login for the business.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: The problem with that approach is if a user from the context is updated. The group appended to the context user would be updated to the user journal. This would cause users to be apart of groups they shouldn't be apart of. Regardless this implementation requires some sort of permitted update on super users in the future.

Also, there aren't many places that check groups other than the auth service and hopefully I've accounted for all the places where we have to set the group to the context.

src/foam/nanos/auth/UserAndGroupAuthService.java
kirkeaton: Yeah. Doing this is going to introduce a lot of confusion and improper code.

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: > The problem with that approach is if a user from the context is updated.

Right. One way we could solve this is to add a new decorator  to `userDAO` (or just modify AuthenticatedUserDAO) to prevent any updates to the current user if there's an agent in the context.

I think this is a better solution than changing the way we access a user's group everywhere in the system.

Example code:

```Java
User agent = (User) x.get("agent");
User existingUser = (User) super.find_(x, obj);
boolean isUpdate = existingUser != null;
boolean isUpdatingSelf = isUpdate && existingUser.getId() == user.getId();
if ( isUpdatingSelf && agent != null ) {
  throw new AuthorizationException("You cannot update a user you are sudoed into.");
}
```

src/foam/nanos/auth/UserAndGroupAuthService.java
TW80000: (Copying over our conversation from an outdated comment to this one for greater visibility)

@zahnno said:
> The problem with that approach is if a user from the context is updated. The group appended to the context user would be updated to the user journal. This would cause users to be apart of groups they shouldn't be apart of. Regardless this implementation requires some sort of permitted update on super users in the future.

> Also, there aren't many places that check groups other than the auth service and hopefully I've accounted for all the places where we have to set the group to the context.

Right. One way we could solve this is to add a new decorator  to `userDAO` (or just modify AuthenticatedUserDAO) to prevent any updates to the current user if there's an agent in the context.

I think this is a better solution than changing the way we access a user's group everywhere in the system.

Example code:

```Java
User agent = (User) x.get("agent");
User existingUser = (User) super.find_(x, obj);
boolean isUpdate = existingUser != null;
boolean isUpdatingSelf = isUpdate && existingUser.getId() == user.getId();
if ( isUpdatingSelf && agent != null ) {
  throw new AuthorizationException("You cannot update a user you are sudoed into.");
}
```

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: Resolving this conversation because it has been succeeded by https://github.com/foam-framework/foam2/pull/1584#discussion_r222286872

src/foam/nanos/auth/UserAndGroupAuthService.java
zahnno: That could work but I still think they should be separated considering the new agent user relationship group isn't part of the user. It's part of a relationship junction object. And I guess the user in the context isn't necessarily the user either but consider what we did with our bareUserDAO, where we striped all authentication on the DAO. All new DAOs storing users will have to account for the limiting condition. It sort of becomes a hidden property value in my opinion, I'm kinda iffy on it.

TW80000: > All new DAOs storing users will have to account for the limiting condition.

The same decorator could be reused on any new DAOs that store `User`s and are separate from the existing `userDAO` and `bareUserDAO`.

---

I think keeping group on the user only introduces fewer changes to the system and less complexity than adding a group to the context and is therefore the better solution. The issue of an agent updating a super user's group is important, but easily mitigated by a simple decorator that can be applied to any DAO that needs it, and right now there's only one DAO that needs it anyway, making things very simple.

src/foam/nanos/auth/ClientAgentAuthService.js
zahnno: The client agent auth service? Not sure if it makes much sense to document. I've documented the agent auth service interface which should be fine considering it contains the service methods and is displayed in the generated api documentation.

src/foam/nanos/auth/UserAndGroupAuthService.java
zahnno: > I think keeping group on the user only introduces fewer changes to the system and less complexity than adding a group to the context and is therefore the better solution.

I don't see how having a group exist within the context is complex. The changes required to make this work is minimal and accounted for within this PR (hopefully) The thing I see with this being a problem is other applications using FOAM accounting for this change but the same thing goes for what you're purposing.

I think that once the group context is introduced, it will lead to less problems in the future.

src/foam/nanos/auth/AgentAuthService.js
zahnno: @kennyqiyenkan care to expand on this?

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: > The administrator's email address isn't tied to the business itself. That's good because if that administrator leaves, their email isn't tied to the login for the business.

Okay... If the administrator leaves and no one is left to be held accountable for the business than we'd have to implement some sort of claiming system or something along those lines. Which seems like added effort to me.

Plus this implementation isn't solely being created to satisfy our needs on the SME platform. It's a generic sudo system and I do think a user should be enabled to be logged into, considering that's how we flag deleted users in the system.

TW80000: The crux of the matter here is whether or not you should be able to log in as a business. That's what we need to come to an agreement on.

Here are the reasons I think we should not allow businesses to log in:

1. See above
2. See above
3. In Ablii our plan was to always require you to act as (sudo as) a business when you sign in. If you sign in as a business, what `User` are you going to act as?
4. Auditability. If you sign in as a business and modify a bunch of permissions, how do we know who was actually signed into the account? If we require you to log into your User account and act as the business before changing permissions we can track who's actually making the changes.

We need to figure this out because if we agree that you shouldn't be able to log in as a business, then this line would prevent you from sudo-ing as a business and we wouldn't be able to use this in Ablii, which is the whole reason we wanted it in the first place. If you still disagree and think you should be able to log in as a business, you need to tell me the solutions you have in mind for the 4 things I've outlined above and we can discuss from there. My guess is even if we could solve those problems, it would still be simpler to solve them by just removing part of the condition in this line, and the simplest solution is often the best one.

TW80000: > I do think a user should be enabled to be logged into, considering that's how we flag deleted users in the system.

Enabled is used to determine whether a `User` can log in, not to mark them as deleted. Also, we're not "logging in" to another user, we're acting as them.

src/foam/nanos/auth/UserAndGroupAuthService.java
TW80000: > I don't see how having a group exist within the context is complex.

Here's how it's more complex. Currently the group is only in one place: on the user. What you're proposing is having the group be in two different places. That's more complex because any code that wants to look at a User's group has to know which group to look at. If the group were only in one place like it currently is it would be unambiguous.

> The changes required to make this work is minimal and accounted for within this PR (hopefully)

There are three methods in this file alone - `auth.userAndGroupContext(), auth.checkUserPermission()` and `auth.updatePassword()` - that weren't accounted for. Who knows how many other places were missed? Who knows how many places in the future are going to be missed? And even if you did change them all, why add the extra complexity when it's not necessary?

> The thing I see with this being a problem is other applications using FOAM accounting for this change but the same thing goes for what you're purposing.

You're correct that both cases require change. Here are some potential solutions:

1. Don't put this functionality in FOAM. We only need it in Ablii at the moment and there's no other use case for it that I've heard of. As a rule of thumb I disagree with making a general solution to a specific problem unless you have at least two or preferably three use cases that would justify the extra level of abstraction.
2. Keep it in FOAM, but add the decorator I proposed to the `userDAO` in FOAM by default.

> I think that once the group context is introduced, it will lead to less problems in the future.

Why do you think having the group in the context will lead to less problems in the future?

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: Addressing nanopay specific requirements for businesses: 
> 1. That's what sudo-ing is for. The administrator of the business sudos into the business as opposed to them logging into the business directly.

This is the argument topic.

> 2. The administrator's email address isn't tied to the business itself. That's good because if that administrator leaves, their email isn't tied to the login for the business.

Businesses tend to have a business specific email. Associating no email to a business will leave orphaned businesses if all users leave.

> 3. In Ablii our plan was to always require you to act as (sudo as) a business when you sign in. If you sign in as a business, what User are you going to act as?

The business will be the user. If required we can always have the business act as it's self and also be the agent.
Regardless we should be developing the nanopay system to support non "acting as" sessions.

>4.  Auditability. If you sign in as a business and modify a bunch of permissions, how do we know who was actually signed into the account? If we require you to log into your User account and act as the business before changing permissions we can track who's actually making the changes.

See above.

> Enabled is used to determine whether a User can log in, not to mark them as deleted. Also, we're not "logging in" to another user, we're acting as them.

True but when we "remove" a user, this is the property value we account for. Meaning we'd be able to act as users who have been "removed".

zahnno:
Replaced by: https://github.com/foam-framework/foam2/pull/1598

PR #1589, opened 2018-10-03
 Distinguish between http and https

ZacZhang:
Distinguish between http and https based on the default URL of AppConfig

### DIFF
--- a/src/foam/nanos/app/AppConfig.js
+++ b/src/foam/nanos/app/AppConfig.js
@@ -11,24 +11,29 @@ foam.CLASS({
   properties: [
     {
       class: 'String',
-      name: 'name'
+      name: 'name',
+      visibility: 'RO'
     },
     {
       class: 'String',
-      name: 'version'
+      name: 'version',
+      visibility: 'RO'
     },
     {
       class: 'String',
       name: 'privacy',
-      value: 'Privacy Policy'
+      value: 'Privacy Policy',
+      visibility: 'RO'
     },
     {
       class: 'String',
-      name: 'privacyUrl'
+      name: 'privacyUrl',
+      visibility: 'RO'
     },
     {
       class: 'String',
-      name: 'copyright'
+      name: 'copyright',
+      visibility: 'RO'
     },
     {
       class: 'String',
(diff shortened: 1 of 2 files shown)


ronin00 (changes requested):
This change assumes as default a scenario where the incoming req URL protocol identifier doesn't match what should be used on URLs in the outbound resp.  (e.g: A loadbalancer strips the HTTPS and forwards the HTTP req to the webserver.  But the webserver should make sure to write it's hrefs as HTTPS in the response).

The problem is that the above scenario is not the DEFAULT for most webapps.  The simple default scenario would be where the HTTPS or HTTP terminate directly on the webserver.  And thus the protocol identifier of the req will simply match the protocol identifiers in outbound resp. 

Thus the default logic in the framework should account for the default scenario -- which was handled by the removed code.  

And any special logic for unusual handling:
- should be optional logic impl in the framework but controlled by a deployment time flag
- or be directly impl by the downstream user app via overrides/refinements mechanisms

src/foam/nanos/app/AppConfig.js
```diff
@@ -63,6 +63,11 @@ foam.CLASS({
       class: 'String',
       name: 'playLink',
       value: 'https://play.google.com/store?hl=en'
+    },
+    {
+      class: 'Boolean',
+      name: 'enableForceHttps',
```
jlhughes, line 69: rename to forceHttps
But this can wait for the next PR, we need to get this out

ronin00 (approved):

src/foam/nanos/auth/Group.js
```diff
@@ -179,13 +179,27 @@ if ( session != null ) {
     } else {
       //populate AppConfig url with request's RootUrl
       HttpServletRequest req = x.get(HttpServletRequest.class);
-      if (req.getRequestURI() != null) {
-        config.setUrl(((Request) req).getRootURL().toString());
+
+      if ( config.getEnableForceHttps() ) {
+        if ( ! SafetyUtil.isEmpty(req.getRequestURI()) && ! SafetyUtil.isEmpty(config.getUrl())) {
```
jlhughes, line 184: Want to review this later.

jlhughes (approved):

PR #1590, opened 2018-10-04
Added side navigation bar

KateMengjunXia:
Related to https://github.com/nanoPayinc/NANOPAY/pull/3536

Added views to support side bar menu.

### DIFF
--- a/src/foam/nanos/u2/navigation/NavigationView.js
+++ b/src/foam/nanos/u2/navigation/NavigationView.js
@@ -1,31 +1,31 @@
 foam.CLASS({
-  package: 'foam.nanos.u2.navigation',
-  name: 'NavigationView',
-  extends: 'foam.u2.View',
-   documentation: 'Navigation bars',
-   css: `
-    ^ .top-nav {
-      width: calc(100% - 200px);
-      display: inline;
-    }
-    ^ .side-nav {
-      float:left;
-      display: inline-block;
-      width: 200px;
-      height: 100%
-    }
-  `,
-   methods: [
-    function initE() {
-      this.addClass(this.myClass())
-        .start()
-          .addClass(this.myClass('top-nav'))
-          .tag({ class: 'foam.nanos.u2.navigation.TopNavigation' })
-        .end()
-        .start()
-          .addClass(this.myClass('side-nav'))
-          .tag({ class: 'foam.nanos.u2.navigation.SideNavigation' })
-        .end();
-    }
-  ]
-});
+    package: 'foam.nanos.u2.navigation',
+    name: 'NavigationView',
+    extends: 'foam.u2.View',
+     documentation: 'Navigation bars',
+     css: `
+      ^top-nav {
+        width: calc(100% - 200px);
+        display: inline;
+      }
+      ^side-nav {
+        float:left;
+        display: inline-block;
+        width: 200px;
+        height: 100%
+      }
+    `,
+     methods: [
+      function initE() {
+        this.addClass(this.myClass())
+          .start()
+           .addClass(this.myClass('top-nav'))
+            .tag({ class: 'foam.nanos.u2.navigation.TopNavigation' })
+          .end()
+          .start()
+            .addClass(this.myClass('side-nav'))
+            .tag({ class: 'foam.nanos.u2.navigation.SideNavigation' })
+          .end();
+      }
+    ]
+  });
--- a/src/foam/nanos/u2/navigation/NavigationView.js
+++ b/src/foam/nanos/u2/navigation/NavigationView.js
@@ -1,31 +1,34 @@
 foam.CLASS({
-    package: 'foam.nanos.u2.navigation',
-    name: 'NavigationView',
-    extends: 'foam.u2.View',
-     documentation: 'Navigation bars',
-     css: `
-      ^top-nav {
-        width: calc(100% - 200px);
-        display: inline;
-      }
-      ^side-nav {
-        float:left;
-        display: inline-block;
-        width: 200px;
-        height: 100%
-      }
-    `,
-     methods: [
-      function initE() {
-        this.addClass(this.myClass())
-          .start()
-           .addClass(this.myClass('top-nav'))
-            .tag({ class: 'foam.nanos.u2.navigation.TopNavigation' })
-          .end()
-          .start()
-            .addClass(this.myClass('side-nav'))
-            .tag({ class: 'foam.nanos.u2.navigation.SideNavigation' })
-          .end();
-      }
-    ]
-  });
+  package: 'foam.nanos.u2.navigation',
+  name: 'NavigationView',
+  extends: 'foam.u2.View',
+
+  documentation: 'Navigation bars',
+
+  css: `
+    ^top-nav {
+      width: calc(100% - 200px);
+      display: inline;
+    }
+    ^side-nav {
+      float:left;
+      display: inline-block;
+      width: 200px;
+      height: 100%
+    }
+  `,
+
+  methods: [
+    function initE() {
+      this.addClass(this.myClass())
+        .start()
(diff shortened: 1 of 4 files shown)


adamvy:
It's more consistent with existing FOAM code if you do `.addClass(this.myClass('top-nav'))` and put `^top-nav { ... }` in your CSS.  This way the elements just have a single unique class name that the browser can look up to find all styles which may be faster.

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,143 @@
...
+       methods: [
+        function initE() {
+          var Menu = this.Menu;
+          var self = this;
+          var dao = this.menuDAO.orderBy(Menu.ORDER)
+              .where(this.EQ(Menu.PARENT, this.menuName));
+              this.addClass(this.myClass())
+              .start().addClass('side-nav')
+                .select(dao, function(menu) {
+                    self.accordionCardShowDict[menu.id] = true;
```
adamvy: This line should be indented the same as the return this.E() line below

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,143 @@
...
+                    .call(function() {
+                      var self2 = this;
+                      this
+                        .start('a').addClass('menuItem')
+                          .add(menu.label)
+                          .on('click', function() {
+                            menu.children.select().then(function(temp) {
+                              // Only display submenu if array length is longer than 0
+                              temp.array.length === 0 ?
+                                  menu.launch_(self2.__context__, self2) :
```
adamvy: You can just call launch(x, e) instead of launch_(x, e) i believe.  Underscore suffix implies we meant the method to be private.  There is an action named "launch" and actions also get added as method to the object.

src/foam/nanos/u2/navigation/NavigationView.js
```diff
@@ -0,0 +1,31 @@
+foam.CLASS({
+  package: 'foam.nanos.u2.navigation',
+  name: 'NavigationView',
+  extends: 'foam.u2.View',
+   documentation: 'Navigation bars',
+   css: `
+    ^ .top-nav {
```
adamvy: Update your CSS to `^top-nav` now, not `^ .top-nav` as we're using myClass('top-nav')

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,147 @@
...
+      ],
+       methods: [
+        function initE() {
+          var Menu = this.Menu;
+          var self = this;
+          var dao = this.menuDAO.orderBy(Menu.ORDER)
+              .where(this.EQ(Menu.PARENT, this.menuName));
+              this.addClass(this.myClass())
+              .start()
+                .addClass('side-nav')
```
adamvy: Use addClass(myClass('side-nav')) and update the CSS to ^side-nav, here and everywhere below.

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,147 @@
...
+    package: 'foam.nanos.u2.navigation',
+    name: 'SideNavigation',
+    extends: 'foam.u2.View',
+  
+    documentation: 'Side navigation bar',
+  
+    implements: [
+        'foam.mlang.Expressions'
+      ],
+       requires: [
```
adamvy: Indentation is weird.

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,147 @@
...
+          name: 'accordionCardShow',
+          value: true
+        },
+        {
+          class: 'String',
+          name: 'menuName',
+        }
+      ],
+       methods: [
+        function initE() {
```
adamvy: I think this view is trying to do too much, and is causing some limitations and complication as a result.

This view appears to be responsible for rendering the entire tree of menu items.  As a result, it only appears to support one level of nesting of menu items, and the click handler for those menu items is repeated twice.  Additionally, state is being maintained imperatively in a property (accordingShowCardDict) rather than declaratively all the way down.

A simpler design would be to write a recursive view that is responsible for rendering a single menu item, and expanding/collapsing the children for that menu item.  The children will be rendered using another instance of the same view.  That way if if the child menus have their own children, we can expand and collapse those for free as well.

Here's some example code to show you the structure I mean:

```
foam.CLASS({
  name: 'VerticalMenuItemView',
  extends: 'foam.u2.View',

  properties: [
    {
      class: 'Boolean',
      name: 'expanded',
      value: false
    }
  ]

  methods: [
    function initE() {
      var view = this;
      
      this.
        add(this.data.label).
        on('click', function() {
          var menu = this.data;
          menu.children.select().then(function(c) {
            if ( c.array.length == 0 )
              menu.launch(view.__context__, view);
            else
              this.expanded = ! this.expanded;
          });
        }).
        add(this.slot(function(expanded, data) {
          return ! expanded ?
            this.E() :
            this.E().
            select(data.children, function(child) {
              return view.cls_.create({ data: child }, view);
            })
        }));
    }
  ]
});


foam.CLASS({
  name: 'VerticalMenuView',
  requires: [ 'VerticalMenuItemView' ],
  methods: [
    function initE() {
      var view = this;
      
      this.select(this.menuDAO, function(menu) {
        return this.VerticalMenuItemView.create({ data: menu });
      });
    }
  ]
});
```

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,147 @@
...
+          var Menu = this.Menu;
```
kgrgreer: I would prefer you not do this since when I read the code below I wonder what Menu is. It looks like a global variable since it isn't scoped and it's capitalized. If I read it as this.Menu then I would know it was a required class and not be confused. It's only being accessed twice so this approach is also actually more characters.

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,109 @@
...
+    function initE() {
+      var self = this;
+      var dao = this.menuDAO
+        .orderBy(this.Menu.ORDER)
+        .where(this.EQ(this.Menu.PARENT, this.menuName));
+      this.addClass(this.myClass())
+        .start()
+          .addClass(this.myClass('side-nav'))
+          .select(dao, function(menu) {
+            return foam.nanos.u2.navigation.SideNavigationItemView.create({ data: menu });
```
adamvy: Pass "this" as the second parameter to create `.create({ data: menu }, this)`

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,109 @@
...
+          })
+        .end();
+    }
+  ]
+});
+
+foam.CLASS({
+  package: 'foam.nanos.u2.navigation',
```
adamvy: Let's move this to its own file and add to files.js

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -0,0 +1,109 @@
...
+            });
+          })
+        .end()
+        .start('div')
+          .add(this.slot(function(expanded, data) {
+            return ! expanded ?
+              this.E() :
+              this.E()
+                .select(data.children, function(child) {
+                  return view.cls_.create({ data: child }, view).style({'margin-left': '30px'});
```
adamvy: Does the styling still work if you move the 'margin-left' style to div we create on line 97?  I think that would be nicer as it would indent ALL children 30px at once rather than indent every child by 30px individually.

PR #1593, opened 2018-10-05
Updated Support Portal

kirkeaton:


src/com/google/foam/demos/u2/Border.js
```diff
@@ -20,9 +20,19 @@ foam.CLASS({
   name: 'Tab',
   extends: 'foam.u2.Element',
 
+  css: `
+    ^ { xxxposition: absolute; }
```
kgrgreer: This block doesn't do anything and could be removed. Actually all changes to demos should be removed from this PR.

src/foam/support/view/MessageCard.js
```diff
@@ -0,0 +1,149 @@
...
+    'ticketDAO',
+    'ticketMessageDAO'
+  ],  
+
+  javaImports: [
+    'java.util.Date'
+  ],
+
+  css: `
+  * {
```
kgrgreer: This looks like its scope is too large.

src/foam/support/view/ReplyView.js
```diff
@@ -0,0 +1,163 @@
...
+        this.viewData.message = newValue;
+      }
+    }
+  ],
+
+  imports: [
+    'viewData'
+  ],
+  
+  exports: [
```
kgrgreer: If it wants to do this is should just remove the 'exports:' and just extend 'Controller' instead of 'View'.

src/foam/support/view/SummaryCard.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+foam.CLASS({
+  package: 'foam.support.view',
+  name: 'SummaryCard',
+  extends: 'foam.u2.View',
+
+  documentation: 'Cards for summary views',
+
+	css: `
```
kgrgreer, line 14: indentation is off

src/foam/support/view/SupportEmailView.js
```diff
@@ -0,0 +1,216 @@
...
+      }
+    }
+  ],
+
+  classes: [
+    {
+      name: 'EmailSupportTableView',
+      extends: 'foam.u2.View',
+      
+      exports: [ 'as data' ],
```
kgrgreer, line 179: Is this used/needed?

src/foam/support/view/TicketSummaryView.js
```diff
@@ -0,0 +1,172 @@
...
+    ^ .ticketdiv{
+      margin: 30px 0px;
+    }
+    ^ .foam-support-view-SummaryCard {
+      width: 170px;
+    }
+  `,
+
+  messages: [
+    { name: 'title',          message: 'Tickets' },
```
kgrgreer: Align columns.

src/foam/support/view/TicketSummaryView.js
```diff
@@ -0,0 +1,172 @@
...
+
+  properties: [
+    {
+      name: 'dao',
+      factory: function() { return this.user.tickets; }
+    },
+    {
+      class: 'Int',
+      name: 'newCount',
+      value: "...",
```
kgrgreer: We can't set the value of an Int to '...'. Same problem below as well.

src/foam/support/view/TicketView.js
```diff
@@ -0,0 +1,122 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.support.view',
+  name: 'TicketView',
+  extends: 'foam.u2.View',
+
+  requires: [
+    'foam.u2.ListCreateController',
```
kgrgreer: sort requires

src/foam/support/view/SummaryCard.js
kirkeaton: no it's not.

PR #1596, opened 2018-10-09
Fix UML diagram layout

hchoura:
Fix layout (user case).
UML Adjust height based on content.

### DIFF
--- a/src/foam/doc/DocBrowser.js
+++ b/src/foam/doc/DocBrowser.js
@@ -750,11 +750,15 @@ foam.CLASS({
       }
     },
     {
+      name: 'height',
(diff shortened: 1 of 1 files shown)


src/foam/doc/DocBrowser.js
```diff
@@ -797,7 +801,27 @@ foam.CLASS({
...
       value: 200
     },
-    'data'
+    'data',
+    {
+      name: 'canvasHeightExtension',
+      value: 0
+    },
+    {
+      name: 'lastRequireY',
```
kgrgreer, line 811: document: what these properties are for

PR #1597, opened 2018-10-09
Add context menu to table rows

TW80000:
Implementation for https://github.com/nanoPayinc/NANOPAY/issues/3861. Please read that ticket for context.

This PR adds a context menu to table views. That mean that each row gets a button (three vertical dots) at the end of it that when clicked will show a popover that contains actions that can be taken on the object in that row. These actions are defined by the model of the object.

Tony wants this for our tables in Ablii. Here's a screenshot from the lo-fi wireframes:

screen shot 2018-10-09 at 3 39 14 pm

and here's a screenshot with this branch running:

screen shot 2018-10-09 at 3 40 12 pm

## Future work/will add to this PR if requested

* Ability to style some of the buttons differently
  * I'm thinking we could add a property to Action called `isDestructive`. If that property is set to true, we'll style that button with red text or something.

### DIFF
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -220,7 +220,14 @@ foam.CLASS({
                     });
                   }).
                   callIf( ! view.disableUserSelection, function() {
-                    this.on('click', function() {
+                    this.on('click', function(evt) {
+                      // If we're clicking somewhere to close the context menu,
+                      // don't do anything.
+                      if (
+                        evt.target.nodeName === 'DROPDOWN-OVERLAY' ||
+                        evt.target.classList.contains(view.myClass('vertDots'))
+                      ) return;
+
                       view.selection = obj;
                       if ( view.importSelection$ ) view.importSelection = obj;
                       if ( view.editRecord$ ) view.editRecord(obj);
@@ -275,15 +282,14 @@ foam.CLASS({
                     return this.start('td').
                       add(overlay).
                       style({ 'text-align': 'right' }).
-                      on('click', function(e) {
-                        e.stopImmediatePropagation();
-                        view.positionOverlayDropdown(overlay);
-                        overlay.open();
-                      }).
                       start('span').
                         addClass(view.myClass('vertDots')).
                         addClass(view.myClass('noselect')).
                         add(view.vertMenuIcon).
+                        on('click', function(evt) {
+                          view.positionOverlayDropdown(overlay);
+                          overlay.open();
+                        }).
                       end().
                     end();
                   });
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -257,13 +257,19 @@ foam.CLASS({
                     }
 
                     overlay.forEach(availableActions, function(action) {
-                      this.start()
+                      var elm = this.start()
                         .addClass(view.myClass('context-menu-item'))
-                        .add(action.label)
-                        .on('click', function(evt) {
-                          action.code.call(obj);
-                        })
-                      .end();
+                        .add(action.label);
+
+                      if ( action.isEnabledFor(obj) ) {
+                        elm = elm.on('click', function(evt) {
+                          action.maybeCall(view.__subContext__, obj);
+                        });
+                      } else {
+                        elm.addClass('disabled');
+                      }
+
+                      elm.end();
                     });
 
                     return this.start('td').
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -157,7 +157,7 @@ foam.CLASS({
       var boundingBox = origin.getBoundingClientRect();
       var dropdownMenu = current.getBoundingClientRect();
 
-      overlay.style({ top: boundingBox.top - dropdownMenu.top + 'px'});
+      overlay.style({ top: boundingBox.top - dropdownMenu.top + 'px' });
     },
 
     function initE() {
@@ -268,15 +268,17 @@ foam.CLASS({
 
                     return this.start('td').
                       add(overlay).
-                      addClass(view.myClass('th-editColumns')).
+                      style({ 'text-align': 'right' }).
                       on('click', function(e) {
                         e.stopImmediatePropagation();
                         view.positionOverlayDropdown(overlay);
                         overlay.open();
                       }).
-                      add(' ', view.vertMenuIcon).
-                      addClass(view.myClass('vertDots')).
-                      addClass(view.myClass('noselect')).
+                      start('span').
+                        addClass(view.myClass('vertDots')).
+                        addClass(view.myClass('noselect')).
+                        add(view.vertMenuIcon).
+                      end().
                     end();
                   });
               });
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -25,14 +25,6 @@ foam.CLASS({
   ],
 
   imports: [
-    // Each table row has a context menu that contains actions you can perform
-    // on the object in that row. The actions used to populate that menu come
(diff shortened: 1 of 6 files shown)


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -236,8 +240,47 @@ foam.CLASS({
...
+                    var availableActions = actions.filter(function(action) {
+                      return action.isAvailableFor(obj);
+                    });
+
+                    // Don't show the context menu if there are no available
+                    // actions defined on the model.
+                    if ( availableActions.length === 0 ) {
+                      if ( view.editColumnsEnabled ) {
+                        return this.tag('td');
+                      } else {
```
kgrgreer: No need for an 'else' if you've already returned in the 'if'.

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -236,8 +240,47 @@ foam.CLASS({
...
+                        return;
+                      }
+                    }
+
+                    availableActions.forEach((action) => {
```
kgrgreer: You could call overlap.forEach(availableActions, fn) and save some code.

kgrgreer: And not store 'overlay' in a variable and just embed in-line below.

TW80000: > And not store 'overlay' in a variable and just embed in-line below.

The problem with that is that I need it in a variable to pass it as an argument to `positionOverlayDropdown()` and call the `open()` method on below.

TW80000:
## TODO

* [x]  Fix position of context menu (8a860bb)
* [x]  Add check for `isEnabled` as well as `isAvailable` (bcbcc76)
* [x]  The call to `stopImmediatePropagation()` is preventing the click event on the context menu overlay from triggering, meaning you can't click anywhere to get rid of the context menu. However, we can't simply remove it because then the click event listener on table rows triggers when you click the vertical dots to open the menu, which we don't want to happen. (40c18fc)
* [x]  Pull actions from the controller as well as the model (4f77da6)
* [ ]  Make it so that when you right click anywhere on the row, the context menu appears where you clicked

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -257,13 +257,19 @@ foam.CLASS({
...
                         .addClass(view.myClass('context-menu-item'))
-                        .add(action.label)
-                        .on('click', function(evt) {
-                          action.code.call(obj);
-                        })
-                      .end();
+                        .add(action.label);
+
+                      if ( action.isEnabledFor(obj) ) {
+                        elm = elm.on('click', function(evt) {
```
kgrgreer: You don't need the 'elm =' since .on() just returns this;

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -257,13 +257,19 @@ foam.CLASS({
...
                     overlay.forEach(availableActions, function(action) {
-                      this.start()
+                      var elm = this.start()
                         .addClass(view.myClass('context-menu-item'))
-                        .add(action.label)
-                        .on('click', function(evt) {
-                          action.code.call(obj);
-                        })
-                      .end();
+                        .add(action.label);
```
kgrgreer: If you didn't want to break up the call and store it in the variable 'elm' you could do .call(function() { if ( action.isEnabledFor() { this.on() } else this.addClass('disabled'); })...

TW80000:
To test passing an action from a view to the context menu, you can make the changes from the diff below to ExpensesView.

My first thought was to make an FObjectArray of Actions property on UnstyledTableView and leave it up to the views who are instantiating it to set it. The problem with that is that views like ScrollTableView instantiate a TableView, but doesn't extend it, meaning ScrollTableView would have to get that array property as well so that it could pass it along to the TableView it's instantiating. Just importing and exporting from the context seemed like a much better solution because it avoids all of the intermediary passing. So that's why I'm getting the actions from the context instead of a property.

It's a little bit janky that I'm creating a property that's an array of Actions and exporting that instead of defining the actions directly on ExpensesView, but I didn't know how to export those in a way that's any cleaner.

Anyway, at this point everything is working except for right-clicking to bring up the context menu. I talked with @zahnno  and we think that's something we'd like to leave for later since it's a "nice to have" for Ablii but not a necessity. If it's alright with you @kgrgreer I'd prefer to open a ticket for the right-click context menu and come back to it later. This PR is required for a lot of the views we need for Ablii so the sooner we can get this merged the better.

```
diff --git a/nanopay/src/net/nanopay/invoice/ui/ExpensesView.js b/nanopay/src/net/nanopay/invoice/ui/ExpensesView.js
index 5f86d2ea9..578f786e0 100644
--- a/nanopay/src/net/nanopay/invoice/ui/ExpensesView.js
+++ b/nanopay/src/net/nanopay/invoice/ui/ExpensesView.js
@@ -17,10 +17,12 @@ foam.CLASS({
   ],
 
   imports: [
+    'stack',
     'user'
   ],
 
   exports: [
+    'contextMenuActions',
     'hideSummary'
   ],
 
@@ -48,6 +50,27 @@ foam.CLASS({
       factory: function() {
         return this.expensesDAO.orderBy(this.DESC(this.Invoice.ISSUE_DATE));
       }
+    },
+    {
+      class: 'FObjectArray',
+      of: 'foam.core.Action',
+      name: 'contextMenuActions',
+      factory: function() {
+        var view = this;
+        return [
+          foam.core.Action.create({
+            name: 'payNow',
+            label: 'Pay now',
+            code: function() {
+              view.stack.push({
+                class: 'net.nanopay.ui.transfer.TransferWizard',
+                type: 'regular',
+                invoice: this
+              });
+            }
+          })
+        ];
+      }
     }
   ],
```

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -23,6 +25,14 @@ foam.CLASS({
...
 
   imports: [
+    // Each table row has a context menu that contains actions you can perform
+    // on the object in that row. The actions used to populate that menu come
+    // from two different sources. The first source is the 'contextMenuActions'
+    // import. If you want a context menu action to do something in the view,
+    // then you should write the code for that action on the view and export it.
+    // The second source of actions is from the model of the object being shown
+    // in the table.
+    'contextMenuActions',
```
kgrgreer: Makes more sense to be a property, which makes tableviews easier to configure.

TW80000:
Diff of `ExpensesView` we can use to test:

```
   ],
 
   imports: [
+    'stack',
     'user'
   ],
 
@@ -98,6 +99,7 @@ foam.CLASS({
   methods: [
     function initE() {
       this.SUPER();
+      var view = this;
 
       this
         .addClass(this.myClass())
@@ -117,7 +119,23 @@ foam.CLASS({
               isBill: true
             },
             detailView: { class: 'net.nanopay.invoice.ui.ExpensesDetailView' },
-            summaryView: this.ExpensesTableView,
+            summaryView: function() {
+              return view.ExpensesTableView.create({
+                contextMenuActions: [
+                  foam.core.Action.create({
+                    name: 'payNow',
+                    label: 'Pay now',
+                    code: function() {
+                      view.stack.push({
+                        class: 'net.nanopay.ui.transfer.TransferWizard',
+                        type: 'regular',
+                        invoice: this
+                      });
+                    }
+                  })
+                ]
+              });
+            },
             showActions: false
           })
         .end()
@@ -177,6 +195,11 @@ foam.CLASS({
 
       properties: [
         'selection',
+        {
+          class: 'FObjectArray',
+          of: 'foam.core.Action',
+          name: 'contextMenuActions'
+        },
         {
           name: 'data',
           factory: function() {
@@ -228,6 +251,7 @@ foam.CLASS({
                 'invoiceNumber', 'purchaseOrder', 'payeeId', 'dueDate',
                 'amount', 'status'
               ],
+              contextMenuActions: this.contextMenuActions
             }).end();
         }
       ]
```

PR #1598, opened 2018-10-09
Agent Auth Service

zahnno:
# Agent Auth Service

The agent auth service is responsible for allowing users to act as others if permitted and places the requesting user as an agent of the system within the context.

A junction between users are created to determine if a user is permitted to act as another user.
*An AuthenticatedDecorator on the junctionDAO will have to be developed to extended any additional authentication logic.*

Along with the target & source Id on the junction object (UserUserJunction), a group is appended and is used to determine the list of permissions for the current user agent access to the system. When acting as another user, the user requesting to act is an "agent" within the context. The user being logged into becomes the "user" within the context, and the group associated to the junction object becomes associated to the "user" within the context.

**Summary of flow**

1. Initial login on user.
2. Call to agent auth service actAs.
3. Lookup junction object associated to the context user and the super user being passed into the service. 
4. Clone super user and set super user group to the group associated to the junction object.
5. Freeze super user.
6. Clone context user and freeze.
7. Set super user as the context "user" and the current context user as an "agent".
8. Put session into DAO

*https://github.com/foam-framework/foam2/pull/1584 Previous PR including discussion on various aspects of the implementation*


src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+  public User actAs(X x, User superUser) throws AuthenticationException {
+    User agent = (User) x.get("user");
+    User user = (User) userDAO_.find(superUser.getId());
+
+    if ( agent == null ) {
+      throw new AuthenticationException();
+    }
+
+    if ( user == null ) {
+      throw new AuthenticationException("Super user doesn't exist.");
```
TW80000: `AuthenticationException` was created solely for the purpose of being thrown when a user is trying to do something without being logged in. If the front end makes a remote procedure call that throws an `AuthenticationException`, it will show the user the login screen. That makes sense because the user should have to log in before doing anything else in the system. However, being shown the login screen when you _are_ logged in doesn't make sense. We should use other exception types for those cases.

That's why in the previous PR I asked if most of these exceptions could be changed. Currently if you logged in as yourself and then tried to act as someone who doesn't exist, you'll be shown the login screen, which I don't think makes sense. I think we should throw a RuntimeException here instead to avoid showing the login screen to the user.

* AuthenticatedException means you weren't logged in when you tried to do something
* AuthorizationException means you were logged in but you didn't have permission to do something
* RuntimeException means something went wrong but we don't have a more specific exception to throw that would be more appropriate

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    Group group = (Group) groupDAO_.find(user.getGroup());
+
+    if ( group == null && ! group.getEnabled() ) {
+      throw new AuthenticationException("Super user must exist within a group.");
```
TW80000: Please change to RuntimeException.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /*
+      Finds the AgentUserJunction object to see if user can log in as the passed in user.
```
TW80000: Is there such thing as an AgentUserJunction or should this say UserUserJunction? Also, we should try to keep the wording as consistent as possible. I think it should say "... if user can act as the ..." instead of "log in". In other comments I think we say "log in" instead of "act as", so those should be switched as well.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+      Source users are permitted to act as target users, not vice versa.
```
TW80000: It should probably be the other way around, right? The person creating the junction is always the source. If Alice wants to act as Bob, then Bob should have to create the junction between himself and Alice, which would give her permission to act as him. In that case, Bob is the source and Alice is the target, so the target is permitted to act as the source, which is the opposite of what's written in the comment. It wouldn't make sense for it to be the other way around from a security perspective because Alice could create the junction between herself and Bob, which would let her act as him. That would mean you could give yourself permission to act as anyone else, which we wouldn't want to allow.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+    */
+    UserUserJunction permissionJunction = (UserUserJunction) agentJunctionDAO_.find(AND(
+      EQ(UserUserJunction.SOURCE_ID, agent.getId()),
+      EQ(UserUserJunction.TARGET_ID, user.getId())
```
TW80000, line 73: If what I wrote above is true, these should be switched.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+    ));
+
+    if ( permissionJunction == null ) {
+      throw new AuthorizationException("You don't have access to act as the requested super user.");
+    }
+
+    // Junction object contains a group which has a unique set of permissions specific to the relationship.
+    Group actingWithinGroup = (Group) groupDAO_.find(permissionJunction.getGroup());
+
+    if ( actingWithinGroup == null && ! actingWithinGroup.getEnabled() ) {
```
TW80000: I think this should be an `||` instead.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /* 
+    Sets the currently logged in user as an "agent" within the context &
+    sets "user" in the context to the passed in user. This allows users to
+    act on behalf of others while retaining information on the user.
+  */
+  public User actAs(X x, User superUser) throws AuthenticationException {
+    User agent = (User) x.get("user");
+    User user = (User) userDAO_.find(superUser.getId());
```
TW80000: Since this method accepts a context as a parameter, I think we should be calling context-oriented methods where we can. Please use the context-oriented `find_` here and fix all places below as well.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+      throw new AuthenticationException();
+    }
+
+    if ( user == null ) {
+      throw new AuthenticationException("Super user doesn't exist.");
+    }
+
+    Group group = (Group) groupDAO_.find(user.getGroup());
+
+    if ( group == null && ! group.getEnabled() ) {
```
TW80000: I think this should be an `||` instead.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+
+    if ( permissionJunction == null ) {
+      throw new AuthorizationException("You don't have access to act as the requested super user.");
+    }
+
+    // Junction object contains a group which has a unique set of permissions specific to the relationship.
+    Group actingWithinGroup = (Group) groupDAO_.find(permissionJunction.getGroup());
+
+    if ( actingWithinGroup == null && ! actingWithinGroup.getEnabled() ) {
+      throw new AuthenticationException("No permissions are appended to the super user relationship.");
```
TW80000: Please change to RuntimeException.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+    
+    // Clone user and associate new junction group to user. Clone and freeze both user and agent. 
+    user = (User) user.fclone();
+    user.setGroup(actingWithinGroup.getId());
+    user.freeze();
+
+    agent = (User) user.fclone();
```
TW80000: Should be `agent = (User) agent.fclone();`.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+    if ( session == null ) {
+      throw new AuthenticationException("Not logged in");
+    }
+
+    X sessionContext = session.getContext();
+    // get agent from session context
+    User agent = (User) sessionContext.get("agent");
+
+    if ( agent == null ) {
+      throw new AuthenticationException("Agent not found.");
```
TW80000: Please change to RuntimeException.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // check if user enabled
+    if ( ! agent.getEnabled() ) {
+      throw new AuthenticationException("Agent disabled");
```
TW80000: Please change to RuntimeException.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -79,6 +79,32 @@ public User getCurrentUser(X x) throws AuthenticationException {
...
+    if ( session == null ) {
+      throw new AuthenticationException("Not logged in");
+    }
+
+    X sessionContext = session.getContext();
+    // get agent from session context
+    User agent = (User) sessionContext.get("agent");
+
+    if ( agent == null ) {
+      throw new AuthenticationException("Agent not found.");
```
TW80000: Please change to RuntimeException.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -79,6 +79,32 @@ public User getCurrentUser(X x) throws AuthenticationException {
...
+    }
+
+    // check if user enabled
+    if ( ! agent.getEnabled() ) {
+      throw new AuthenticationException("Agent disabled");
```
TW80000: Please change to RuntimeException.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -126,13 +145,32 @@ public User login(X x, long userId, String password) throws AuthenticationExcept
...
+  /**
+   * Login a user by the id provided, validate the password
+   * and return the user in the context.
+   */
+  public User login(X x, long userId, String password) throws AuthenticationException {
+    if ( userId < 1 || SafetyUtil.isEmpty(password) ) {
+      throw new AuthenticationException("Invalid Parameters");
+    }
+
+    User user = (User) userDAO_.find(userId);
```
TW80000, line 143: Please call context-oriented methods when you're in a context-oriented method.

src/services
```diff
@@ -74,3 +74,5 @@ p({"class":"foam.nanos.boot.NSpec","name":"appConfig",                         "
...
 p({"class":"foam.nanos.boot.NSpec", "name":"regulatoryNoticeAuditDAO",             "lazy":true,  "serve":true,  "serviceScript":"return new foam.dao.PMDAO(new foam.dao.SequenceNumberDAO (new foam.dao.java.JDAO(x, foam.demos.net.nap.web.model.RegulatoryNoticeAudit.getOwnClassInfo(), \"regulatoryNoticeAudits\")));","client":"{\"of\":\"foam.demos.net.nap.web.model.RegulatoryNoticeAudit\"}"})
 p({"class":"foam.nanos.boot.NSpec", "name":"documentDAO", "lazy": true, "serve": true, "serviceScript": "new foam.dao.ReadOnlyDAO.Builder(x).setDelegate(new foam.flow.DocumentationFolderDAO.Builder(x).setDir(\"src/documentation\").build()).build();","client":"{\"of\":\"foam.flow.Document\"\}"})
+p({"class":"foam.nanos.boot.NSpec", "name":"agentAuth",                        "lazy":false, "serve":true, "authenticate": false, "description":"Permits users to act as others if permissioned. Places user as an agent of the system.", "boxClass":"foam.nanos.auth.AgentAuthServiceSkeleton", "serviceScript":"return new foam.nanos.auth.AgentUserAuthService(x);","client":"{\"class\":\"foam.nanos.auth.ClientAgentAuthService\"}"})
```
TW80000: Please fix typo. "permissioned" should be "permitted".

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,57 @@
...
+    the group associated to the junction object is associated to the "user" within the context.
+  `,
+
+
+  methods: [
+    {
+      name: 'actAs',
+      javaReturns: 'foam.nanos.auth.User',
+      swiftReturns: 'foam_nanos_auth_User',
+      javaThrows: ['foam.nanos.auth.AuthenticationException'],
```
TW80000: Please add missing spaces.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,57 @@
...
+  name: 'AgentAuthService',
+
+  documentation: `
+    The agent auth service is responsible for allowing users to act as
+    others if permitted and places the requesting user as an agent of the system within the context.
+    A junction between users are created to determine if a user is permitted to act as another user.
+    Along with the target & source Id on the junction object (UserUserJunction), a group is
+    appended and is used to determine the list of permissions for the current user agent
+    access to the system. When acting as another user, the user requesting to act is an "agent"
+    within the context. The user being logged into becomes the "user" within the context, and
```
TW80000: Should say "acted as" instead of "logged into" to keep the wording consistent and avoid confusion.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,57 @@
...
+    the group associated to the junction object is associated to the "user" within the context.
```
TW80000, line 19: Should say "associated with", not "associated to".

TW80000 (changes requested):

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: > If the front end makes a remote procedure call that throws an AuthenticationException, it will show the user the login screen.

That's not true. I still stand by what I said in the previous PR, the agent auth service is a form of authentication and should throw authentication exceptions where necessary.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: I've accounted for this within the user user relationship. Source user is permitted to act as target user. I've updated the comment to clarify the term "source" and "target" in this context.

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: > That's not true.

I'm pretty sure it is unless I misunderstood the code here: https://github.com/foam-framework/foam2/blob/master/src/foam/box/SessionClientBox.js#L30 Isn't that doing what I said?

src/foam/nanos/auth/UserAndGroupAuthService.java
zahnno: I don't believe this is necessary considering this is an interface service not a decorator. I could be wrong. @kgrgreer

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: That's not the problem. What's stopping Alice from doing this?

```JavaScript
var bob = ctrl.__subContext.userDAO.find(123);

// When Alice adds bob to her list of superusers, the junction is
// created with `sourceId` set to Alice's id and `targetId`
// becomes Bob's id.
ctrl.user.superUsers.add(bob);

// Alice is able to act as Bob without him ever knowing, which is bad.
ctrl.agentAuth.actAs(bob);
```

The way it should be is the other way around. Bob should have to add Alice as an agent, in which case his id would be the `sourceId` and Alice's would be the `targetId`, and in that case, the target is acting as the source, not the source acting as the target.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: Whoops, you're right. I was testing the junction authorization exception when checking to see if the authentication exception redirected user to login screen. I'll make the changes where necessary.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: I don't think the source and targetId matter much in this case. This issue would have to be resolved by a custom authenticator on the agentJunctionDAO which I mentioned in the summary.

src/foam/nanos/auth/AgentUserAuthService.java
jlhughes: With regard to using the passed context, subsequent calls should be of the form:
userDAO_.inX(x).find(superUser.getId());

src/foam/nanos/auth/UserAndGroupAuthService.java
jlhughes: With regard to calling optional context-oriented methods from within a method taking the context, please use the form:
dao.inX(x).find(id)
rather than
dao.find(x, id);

TW80000: I think this line was missed. It just needs an `inX(x)` and then I think all of the methods have been fixed.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -76,7 +76,7 @@ public User actAs(X x, User superUser) throws AuthenticationException {
     // Junction object contains a group which has a unique set of permissions specific to the relationship.
     Group actingWithinGroup = (Group) groupDAO_.inX(x).find(permissionJunction.getGroup());
 
-    if ( actingWithinGroup != null && ! actingWithinGroup.getEnabled() ) {
+    if ( actingWithinGroup == null || ! actingWithinGroup.getEnabled() ) {
```
TW80000: This will throw a NullPointerException if actingWithinGroup is null.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,126 @@
...
+      throw new AuthenticationException();
+    }
+
+    if ( user == null ) {
+      throw new AuthorizationException("Super user doesn't exist.");
+    }
+
+    Group group = (Group) groupDAO_.inX(x).find(user.getGroup());
+
+    if ( group != null && ! group.getEnabled() ) {
```
TW80000: I don't think this will throw the exception if the group is null. I think it should be:

```Java
if ( group == null || ! group.getEnabled() ) {
```

which is the same as below on line 79.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,57 @@
...
+  `,
+
+
+  methods: [
+    {
+      name: 'actAs',
+      javaReturns: 'foam.nanos.auth.User',
+      swiftReturns: 'foam_nanos_auth_User',
+      javaThrows: [ 'foam.nanos.auth.AuthenticationException' ],
```
TW80000: This also throws 'foam.nanos.auth.AuthorizationException' now.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,57 @@
...
+          name: 'superUser',
+          javaType: 'foam.nanos.auth.User'
+        }
+      ]
+    },
+    {
+      name: 'getCurrentAgent',
+      javaReturns: 'foam.nanos.auth.User',
+      swiftReturns: 'foam_nanos_auth_User?',
+      javaThrows: [ 'foam.nanos.auth.AuthenticationException' ],
```
TW80000: This also throws 'foam.nanos.auth.AuthorizationException' now.

TW80000 (changes requested):
* I left a couple of comments that should only take a minute to fix.
* I also checked out the branch to test it but was getting null pointer exceptions on the backend. When I switched to the foam2 master branch logging in worked again so I'm assuming the issue is to do with this branch. Please check if it's working on your machine.
* There's a merge conflict that needs to be resolved.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -0,0 +1,57 @@
...
+    the group associated to the junction object is associated to the "user" within the context.
+  `,
+
+
+  methods: [
+    {
+      name: 'actAs',
+      javaReturns: 'foam.nanos.auth.User',
+      swiftReturns: 'foam_nanos_auth_User',
+      javaThrows: [ 'foam.nanos.auth.AuthorizationException' ],
```
TW80000: Sorry, maybe my previous comment wasn't clear. It throws both exceptions, so both should be in the array.

TW80000 (approved):

jlhughes (approved):
Looks good.

src/foam/nanos/auth/User.js
```diff
@@ -347,3 +347,25 @@ foam.RELATIONSHIP({
     tableWidth: 120
   }
 });
+
+// Relationship used in the agent auth service. Determines permission list when acting as a super user.
+foam.RELATIONSHIP({
+  cardinality: '*:*',
+  sourceModel: 'foam.nanos.auth.User',
+  targetModel: 'foam.nanos.auth.User',
+  forwardName: 'superUsers',
```
kgrgreer: Would 'agentcies' be a better word since the forward and inverse names should be the inverse of each other.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,126 @@
...
+    sessionDAO_  = (DAO) getX().get("sessionDAO");
+    agentJunctionDAO_  = (DAO) getX().get("agentJunctionDAO");
+  }
+
+  /* 
+    Sets the currently logged in user as an "agent" within the context &
+    sets "user" in the context to the passed in user. This allows users to
+    act on behalf of others while retaining information on the user.
+  */
+  public User actAs(X x, User superUser) throws AuthenticationException {
```
kgrgreer: I think superUser is a poor choice of argument name. Maybe just 'user'.

kgrgreer: Does this need to be a whole User object, or would just a userId be better?

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,126 @@
...
+    User agent = (User) x.get("user");
+    User user = (User) userDAO_.inX(x).find(superUser.getId());
```
kgrgreer: Do you need to call inX(x) here or below for the groupDAO?

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    if ( user == null ) {
+      throw new AuthorizationException("Super user doesn't exist.");
+    }
+
+    Group group = (Group) groupDAO_.inX(x).find(user.getGroup());
+
+    if ( group == null || ! group.getEnabled() ) {
+      throw new AuthorizationException("Super user must exist within a group.");
```
kgrgreer: I would make two checks. One saying the current message, the other saying that the user's group must be enabled.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /*
+      Finds the UserUserJunction object to see if user can act as the passed in user.
+      Source (agent) users are permitted to act as target (superUser) users, not vice versa.
+    */
+    UserUserJunction permissionJunction = (UserUserJunction) agentJunctionDAO_.inX(x).find(AND(
```
kgrgreer: Isn't the

kgrgreer: Isn't there a method on the relationship property to do this check for you?

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,126 @@
...
+      EQ(UserUserJunction.SOURCE_ID, agent.getId()),
+      EQ(UserUserJunction.TARGET_ID, user.getId())
+    ));
+
+    if ( permissionJunction == null ) {
+      throw new AuthorizationException("You don't have access to act as the requested super user.");
```
kgrgreer: "... act as an agent for the requested entity."

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    // Junction object contains a group which has a unique set of permissions specific to the relationship.
+    Group actingWithinGroup = (Group) groupDAO_.inX(x).find(permissionJunction.getGroup());
+
+    if ( actingWithinGroup == null || ! actingWithinGroup.getEnabled() ) {
+      throw new AuthorizationException("No permissions are appended to the super user relationship.");
```
kgrgreer: update message

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,126 @@
...
+    user = (User) user.fclone();
+    user.setGroup(actingWithinGroup.getId());
+    user.freeze();
+
+    agent = (User) agent.fclone();
+    agent.freeze();
+
+    // Set user and agent objects into the session context and place into sessionDAO.
+    Session session = x.get(Session.class);
+    session.setUserId(user.getId());
```
kgrgreer, line 97: Do we want to update the session user id?

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: Could be either, but the user object would sustain the idea that they have access to finding or selecting the user object.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: Sure, considering it's used as a reference to the user within the context. We'd still want it to relate to the new entity we put into the context.

kgrgreer: I don't think it is used. When the user is needed from the context we grab "user". This is just used to identity the user in the session object for things like only enabling one login per user.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,130 @@
...
+    session.setContext(session.getContext().put("user", user));
+    session.setContext(session.getContext().put("agent", agent));
+    sessionDAO_.put(session);
```
kgrgreer: Is this line required? Doesn't it still work without it?

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,130 @@
...
+    
+    return user;
+  }
+
+  /**
+    Retrieves the agent user from the current sessions context.
```
kgrgreer, line 105: Is the agent the original user or the user they're acting as. Document this.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,130 @@
...
+  */
+  public User getCurrentAgent(X x) throws AuthorizationException {
+    // fetch context and check if not null or user id is 0
```
kgrgreer: Couldn't this all just be:  return (User) x.get("agent");

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -79,6 +79,32 @@ public User getCurrentUser(X x) throws AuthenticationException {
     return user;
   }
 
+  /**
+    Retrieves the agent user from the current sessions context.
+  */
+  public User getCurrentAgent(X x) throws AuthenticationException {
```
kgrgreer: Is this needed?

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: It's used in https://github.com/foam-framework/foam2/blob/95b422f4b9928107f5ed810ec1190162a8216ba7/src/foam/nanos/auth/UserAndGroupAuthService.java#L59

If we didn't set the userId in the context it would return the wrong user.

src/foam/nanos/auth/UserAndGroupAuthService.java
zahnno: I think it would be beneficial to provide the client the ability to retrieve the current agent within the sessions context.

src/foam/nanos/auth/AgentUserAuthService.java
kgrgreer: Can we modify UserAndGroupAuthService to just pull the User out of the context instead?

src/foam/nanos/auth/UserAndGroupAuthService.java
kgrgreer: If we know that you have to have Agent support then the AgentAuthService and AuthService could be merged. If not, then the AuthService shouldn't have this method.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -0,0 +1,120 @@
...
+    session.setContext(session.getContext().put("agent", agent));
+    
+    return user;
+  }
+
+  /**
+    Retrieves the agent user from the current sessions context.
+  */
+  public User getCurrentAgent(X x) throws AuthorizationException {
+    // Fetch context and check if not null.
```
kgrgreer, line 108: Can you figure out why get accessing "user" from Context doesn't work?

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: There are other areas that use the userId to pull the user from the session. I don't think going around and changing all of those places woulds be a good idea, unless you can think of a good reason to do that.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: Will investigate and address in a later PR.

src/foam/nanos/auth/AgentUserAuthService.java
kgrgreer: Can you provide me with a list of places that do that?

zahnno: https://github.com/foam-framework/foam2/blob/0c9301e7a8e81937b67dc9955e569159977c397e/src/foam/box/SessionServerBox.java#L63

https://github.com/foam-framework/foam2/blob/0c9301e7a8e81937b67dc9955e569159977c397e/src/foam/nanos/auth/Group.js#L178

https://github.com/foam-framework/foam2/blob/0c9301e7a8e81937b67dc9955e569159977c397e/src/foam/nanos/auth/UserAndGroupAuthService.java#L54

https://github.com/foam-framework/foam2/blob/0c9301e7a8e81937b67dc9955e569159977c397e/src/foam/nanos/auth/UserAndGroupAuthService.java#L59

https://github.com/foam-framework/foam2/blob/0c9301e7a8e81937b67dc9955e569159977c397e/src/foam/nanos/auth/UserAndGroupAuthService.java#L223

https://github.com/foam-framework/foam2/blob/0c9301e7a8e81937b67dc9955e569159977c397e/src/foam/nanos/auth/UserAndGroupAuthService.java#L228

As far as I know.

PR #1602, opened 2018-10-11
New Permissions Matrix View

adamvy:
This is a new approach to the Permissions Matrix View.

It transposes the groupDAO into a PermissionMatrixDAO where the columns are groups and the rows are individual permissions.

Then we can just throw that in a normal TableView to edit them as a large matrix.

As this is a DAO we can also use any other DAO view we have like a DAOController if we want a more rich editing experience.


jlhughes:
Replaced by Kevin's recent work.

PR #1606, opened 2018-10-12
cascading decorator to remove children onDelete

smirnovaka:


src/foam/dao/OnDeleteRemoveChildrenDAO.js
```diff
@@ -0,0 +1,58 @@
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'OnDeleteRemoveChildrenDAO',
```
kgrgreer: How about 'RemoveChildrenOnRemoveDAO' instead. Since the DAO method is called 'remove' not 'delete'.

src/foam/dao/OnDeleteRemoveChildrenDAO.js
```diff
@@ -0,0 +1,58 @@
...
+  extends: 'foam.dao.ProxyDAO',
+
+  javaImports: [
+    'foam.mlang.MLang'
+  ],
+
+  properties: [
```
kgrgreer: These properties are all stored in the Relationship. Shouldn't/couldn't we just use the Relationship instead?

src/foam/dao/RemoveChildrenOnRemoveDAO.js
```diff
@@ -0,0 +1,58 @@
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'RemoveChildrenOnRemoveDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  javaImports: [
+    'foam.mlang.MLang'
```
kgrgreer: Could you change this to 'static foam.mlang.MLang.*' and then avoid having to prefix MLangs with MLang?

PR #1607, opened 2018-10-12
Fixed output JsonJ properly

minsunny:
https://github.com/nanoPayinc/NANOPAY/issues/3385

Fixed output Jsonj Properly by using the flag in json Outputter
Deleted the inner script UI in digWebAgent

### DIFF
--- a/src/foam/nanos/dig/DigWebAgent.java
+++ b/src/foam/nanos/dig/DigWebAgent.java
@@ -69,7 +69,7 @@ public class DigWebAgent
     try {
       if ( SafetyUtil.isEmpty(daoName) ) {
         resp.setContentType("text/html");
-        outputPage(x);
+
         // FIXME: Presently the dig UI doesn't have any way to submit/send a request.
         //   String url = "/#dig";
         //   try {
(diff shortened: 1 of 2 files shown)


src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -295,27 +295,16 @@ public void execute(X x) {
             foam.lib.json.Outputter outputterJson = new foam.lib.json.Outputter(OutputterMode.NETWORK);
             List a = sink.getArray();
             String dataToString = "";
+            outputterJson.setOutputJsonj_(true);
 
             //resp.setContentType("application/json");
-            for ( int i = 0; i < a.size(); i++ ) {
+            for ( int i = 0 ; i < a.size() ; i++ )
```
jlhughes, line 301: Does this compile? looks like the for loop is missing the opening {
And is the space before ; part of our style guide?

minsunny: Yes, It is compiled successfully, I just followed this our guide line below.

https://github.com/foam-framework/foam/wiki/FOAM2-Style-Guidelines
----------------------------------------------------------

One-statement if, while, and for statements that fit in less than 80 characters, do not need braces.
if ( ! found ) return false;

for ( var i = 0 ; i < a.length ; i++ ) a[i] = '';

jlhughes (approved):

PR #1612, opened 2018-10-15
Add missing SequenceNumberDAO constructor back

TW80000:
It was replaced in a9a5a9068d110df298bc18fdbccce0d450dae0c3 but doing so broke all the places that used the previous interface.


src/foam/dao/SequenceNumberDAO.js
```diff
@@ -141,6 +141,11 @@ foam.CLASS({
             setValue((long) ( ( (foam.mlang.sink.Max) sink ).getValue() == null ? 1 : ( (Number) ( (foam.mlang.sink.Max) sink ).getValue() ).longValue() + 1.0 ));
           }
 
+          public SequenceNumberDAO(foam.dao.DAO delegate) {
+            System.err.println("Direct constructor use is deprecated. Use Builder instead.");
+            setDelegate(delegate);
```
kirkeaton: Replace these lines with `this(1, delegate);`

PR #1613, opened 2018-10-15
User Group Validation

shreymahendru:


src/foam/box/SessionServerBox.java
```diff
@@ -68,22 +68,27 @@ public void send(Message msg) {
 
           if ( authenticate_ && ! auth.check(session.getContext(), "service." + spec.getName()) ) {
             Logger logger   = (Logger) x.get("logger");
-            logger.debug("missing permission", group.getId(), "service." + spec.getName());
+            logger.debug("missing permission", group != null ? group.getId() : "NO GROUP" , "service." + spec.getName());
```
jlhughes, line 72: Change this to a warning so we fix them, and potentially notify on this situation.

src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -51,21 +51,26 @@ public FObject put_(X x, FObject obj) {
...
       ! auth.check(x, GLOBAL_SPID_UPDATE) &&
       ! auth.check(x, "spid.update." + user.getSpid()) ) {
       throw new RuntimeException("Unable to update user");
     }
 
     // set spid if not set
-    if ( SafetyUtil.isEmpty((String) toPut.getSpid()) &&
-        ! SafetyUtil.isEmpty((String) user.getSpid()) ) {
+    if ( SafetyUtil.isEmpty(toPut.getSpid()) &&
+        ! SafetyUtil.isEmpty(user.getSpid()) ) {
```
jlhughes, line 73: Can we also log a warning when the users spid is empty, I don't see this as a valid scenario.

src/foam/box/SessionServerBox.java
```diff
@@ -68,22 +68,27 @@ public void send(Message msg) {
...
             // msg.replyWithException(new NoPermissionException("No permission"));
             // return;
           }
 
+          // padding this cause if group is null this can cause an NPE
+          // technically the user shouldn't be created without a group
+          if (group != null) {
```
kgrgreer, line 79: How can this happen or why is this being done?

src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -51,21 +51,26 @@ public FObject put_(X x, FObject obj) {
...
-      throw new RuntimeException("Unable to update user");
-    }
+     User toPut = (User) obj;
+
+     if ( toPut == null ) {
+       throw new RuntimeException("User can't be null");
+     }
 
-    if ( toPut != null && ! SafetyUtil.equals(toPut.getId(), user.getId()) &&
+     if ( isExistingUser(toPut) &&  toPut.getId() < 1000 ) {
```
kgrgreer: Replace two spaces after && with one.

src/foam/box/SessionServerBox.java
jlhughes: This is exactly the problem we are trying to address.

src/foam/box/SessionServerBox.java
shreymahendru: added the warning later, where the `group != null` is checked

jlhughes (approved):

PR #1458, opened 2018-08-13
Prevent users from adding permissions they don't have to a group

TW80000:
- Anyone can create or update a group if they have permission (normal AuthenticatedDAO already exists on groupDAO), but this PR adds an extra check to make sure you can't add any permissions that aren't implied by ones you already have to an existing group or a newly created group.

Marks https://github.com/nanoPayinc/NANOPAY/issues/3130 ready for testing.

PR #1615, opened 2018-10-16
Prevent setting own group

deejes:
Extends  PR #1458


src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -64,17 +63,36 @@ public FObject put_(X x, FObject obj) {
...
+      ! auth.check(x, "spid.update." + user.getSpid())
+    ) {
+      throw new AuthorizationException("You do not have permission to update this user.");
     }
 
     // set spid if not set
-    if ( SafetyUtil.isEmpty(toPut.getSpid()) &&
-        ! SafetyUtil.isEmpty(user.getSpid()) ) {
-      toPut.setSpid(user.getSpid());
+    if ( SafetyUtil.isEmpty((String) newUser.getSpid()) &&
```
shreymahendru, line 72: Where are you getting the `newUser` from?

src/foam/nanos/auth/AuthenticatedGroupDAO.java
```diff
@@ -0,0 +1,76 @@
...
+ * Features:
+ *   - If creating a group, you can only give it permissions implied by those
+ *     that you already have.
+ *   - If updating a group, you can't modify the permissions array to have a
+ *     permission that isn't implied by one you have.
+ *   - If updating a group, you can't modify the parent of the group such that
+ *     the group gains a permission that isn't implied by one you have.
+ */
+public class AuthenticatedGroupDAO extends ProxyDAO {
+  public AuthenticatedGroupDAO(X x, DAO delegate) {
```
kirkeaton, line 27: Constructor isn't necessary as you're only calling the super method.

src/foam/nanos/auth/AuthenticatedGroupDAO.java
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    // If a new group is being created, check that the user has permission to create the group
+    if ( getDelegate().find_(x, obj) == null ) {
+      AuthService auth = (AuthService) x.get("auth");
+      User user = (User) x.get("user");
```
TW80000: Is this used or can it be removed?

src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -64,17 +63,36 @@ public FObject put_(X x, FObject obj) {
...
+        ! SafetyUtil.isEmpty((String) user.getSpid()) ) {
+      newUser.setSpid(user.getSpid());
     }
 
-    return super.put_(x, toPut);
+    User oldUser = (User) super.find_(x, newUser.getId());
```
shreymahendru, line 77: you can modify the `isExistingUser` function to return you the user if it exists. You should probably change the name of the function to something like `getExistingUser`

src/foam/nanos/auth/AuthenticatedGroupDAO.java
```diff
@@ -0,0 +1,76 @@
...
+      if ( group == null ) {
+        throw new RuntimeException("The '" + toCheck.getId() + "' group has a null ancestor named '" + groupId + "'.");
+      }
+    }
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    // If a new group is being created, check that the user has permission to create the group
+    if ( getDelegate().find_(x, obj) == null ) {
```
kirkeaton: Does passing the FObject into the delegate's find method like this work? I thought it's meant to accept the object's primary key.

src/foam/nanos/auth/AuthenticatedUserDAO.java
```diff
@@ -64,17 +63,36 @@ public FObject put_(X x, FObject obj) {
...
+
+    if (
+      oldUser != null &&
+      ! SafetyUtil.equals(oldUser.getGroup(), newUser.getGroup())
+    ) {
+      boolean hasOldGroupUpdatePermission = auth.check(x, "group.update." + oldUser.getGroup());
+      boolean hasNewGroupUpdatePermission = auth.check(x, "group.update." + newUser.getGroup());
+      if ( updatingSelf ) {
```
shreymahendru, line 85: I think is `updatingSelf ` an undeclared variable.

shreymahendru (changes requested):

src/foam/nanos/auth/AuthenticatedUserDAO.java
TW80000: I originally changed `toPut` to `newUser` and `existingUser` to `oldUser`. Some of those changes must have gotten lost, probably in merge conflict fixes.

src/foam/nanos/auth/AuthenticatedGroupDAO.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Make sure that the user has all permissions in the given group.
+   */
+  public void checkUserHasAllPermissionsInGroup(X x, Group toCheck) {
+    AuthService auth = (AuthService) x.get("auth");
+    String groupId;
+    Group group = toCheck;
+
+    while ( true ) {
```
shreymahendru: You should have a better condition. EG: `while( getAncestor(group) != null )` where `getAncestor(Group group)` is a helper that returns a groups if it exists null otherwise.

shreymahendru (changes requested):

src/foam/nanos/auth/AuthenticatedUserDAO.java
deejes: yeah I think thats whats happened

src/foam/nanos/auth/AuthenticatedGroupDAO.java
kirkeaton: Agree. You only have one break condition in your while loop (excluding exceptions) so you could make that your while loop exit condition.

src/foam/nanos/auth/AuthenticatedGroupDAO.java
kirkeaton: Not being used. Please remove.

src/foam/nanos/auth/AuthenticatedGroupDAO.java
```diff
@@ -0,0 +1,70 @@
...
+    Permission[] permissions = group.getPermissions();
+    for ( Permission permission : permissions ) {
+      String id = permission.getId();
+      if ( ! auth.check(x, id) ) {
+        throw new AuthorizationException("Permission Denied. You do not have the '" + id + "' permission.");
+        }
+     }
+  }
+
+  // returns ancestor of a group if it exists, null otherwise null
```
TW80000, line 58: Please fix comment. "null otherwise null" doesn't make sense.

src/foam/nanos/auth/AuthenticatedGroupDAO.java
```diff
@@ -0,0 +1,70 @@
...
+  
+  // Make sure that the user has all permissions in the given group.
+  private void checkUserHasAllPermissionsInGroup(X x, Group group) {
+    AuthService auth = (AuthService) x.get("auth");
+    Permission[] permissions = group.getPermissions();
+    for ( Permission permission : permissions ) {
+      String id = permission.getId();
+      if ( ! auth.check(x, id) ) {
+        throw new AuthorizationException("Permission Denied. You do not have the '" + id + "' permission.");
+        }
```
TW80000, line 54: Please fix indentation.

src/foam/nanos/auth/AuthenticatedGroupDAO.java
```diff
@@ -0,0 +1,70 @@
...
+     }
+  }
+
+  // returns ancestor of a group if it exists, null otherwise null
+  private Group getAncestor(X x, Group group) {
```
TW80000, line 59: Please make all of these methods public. Making them private doesn't help us in any way, but making them public makes it easier to access them from Scripts, Tests, and Crons if needed.

src/foam/nanos/auth/AuthenticatedGroupDAO.java
```diff
@@ -0,0 +1,70 @@
...
+ *   - If updating a group, you can't modify the parent of the group such that
+ *     the group gains a permission that isn't implied by one you have.
+ */
+public class AuthenticatedGroupDAO extends ProxyDAO {
+  public AuthenticatedGroupDAO(X x, DAO delegate) {
+    super(x, delegate);
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
```
TW80000, line 32: Why was the check for the create permission removed?

deejes: Won't that be handled by AuthenticatedDAO? Removed that section per comments above.

TW80000: I'm not sure which comment you're referring to, but the plain `AuthenticatedDAO` check won't work as it's currently written. It only checks if you have the `group.create` permission, which isn't fine-grained enough. We want to check `"group.create." + ((Group) obj).getId()`.

TW80000: It might make more sense to update `AuthenticatedDAO` to check the id though rather than do a second check here.

PR #1622, opened 2018-10-17
Fix styling to conform to style guide

deejes:
No changes to functionality, only style changes.

### DIFF
--- a/src/foam/nanos/auth/Address.js
+++ b/src/foam/nanos/auth/Address.js
@@ -45,7 +45,7 @@ foam.CLASS({
     {
       class: 'String',
       name: 'address1',
-      //required: true
+      // required: true
       width: 70,
       displayWidth: 50,
       documentation: 'for an unstructured address, use this as a main address field.',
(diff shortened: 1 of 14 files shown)


shreymahendru (commented):
Remove package-lock.json

src/foam/nanos/auth/Address.js
```diff
@@ -45,7 +45,7 @@ foam.CLASS({
     {
       class: 'String',
       name: 'address1',
-      //required: true
+      // required: true
```
shreymahendru, line 48: remove the commented out code if it's no  `required`

src/foam/nanos/auth/Country.js
```diff
@@ -28,7 +28,9 @@ foam.CLASS({
     name: 'toString',
     returns: 'String',
     javaReturns: 'String',
-    code: function() { return "Country: " + this.code + ", " + this.name; },
+    code: function() {
+      return 'Country: ' + this.code + ', ' + this.name;
```
shreymahendru, line 32: User string interpolation here.

kgrgreer: Interpolation used to be about 100 times slower, but I just ran a benchmark and it seems that interpolation is as fast or faster now.

kgrgreer: But I wouldn't bother changing existing code to use interpolation.

PR #1624, opened 2018-10-18
Send a notification when cron job start or end

ZacZhang:
https://github.com/nanoPayinc/NANOPAY/issues/2371
- Send a notification when cron job start or end
- Other problems of this ticket have already been fixed

https://github.com/foam-framework/foam2/pull/1589
https://github.com/nanoPayinc/NANOPAY/issues/3890
- rename enableForceHttps to forceHttps in AppConfig model

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -185,7 +185,7 @@ if ( session != null ) {
       //populate AppConfig url with request's RootUrl
       HttpServletRequest req = x.get(HttpServletRequest.class);
 
-      if ( config.getEnableForceHttps() ) {
+      if ( config.getForceHttps() ) {
         if ( ! SafetyUtil.isEmpty(req.getRequestURI()) && ! SafetyUtil.isEmpty(config.getUrl())) {
           String curConfigUrl = config.getUrl();
           String request = ((Request) req).getRootURL().toString();
--- a/src/foam/nanos/cron/Cron.js
+++ b/src/foam/nanos/cron/Cron.js
@@ -87,25 +87,15 @@ foam.CLASS({
       javaReturns: 'void',
       javaCode:
 `DAO notification = (DAO) x.get("notificationDAO");
-            
+                
 Notification cronStartNotify = new Notification();
-StringBuilder startMsg = new StringBuilder();
-startMsg.append("Cron job: ").append(this.getId()).append(" started.");
-if ( ! SafetyUtil.isEmpty(this.getDescription()) ) {
-  startMsg.append(" Cron job description: ").append(this.getDescription());
-}
-cronStartNotify.setBody(startMsg.toString());
+cronStartNotify.setBody("Cron STARTED - " + this.getId() + " " + this.getDescription());
 notification.put(cronStartNotify);
 
 super.runScript(x);
 
 Notification cronEndNotify = new Notification();
-StringBuilder endMsg = new StringBuilder();
-endMsg.append("Cron job: ").append(this.getId()).append(" ended.");
-if ( ! SafetyUtil.isEmpty(this.getDescription()) ) {
-  endMsg.append(" Cron job description: ").append(this.getDescription());
-}
-cronEndNotify.setBody(endMsg.toString());
+cronEndNotify.setBody("Cron ENDED - " + this.getId() + " " + this.getDescription());
 notification.put(cronEndNotify);
 
 setScheduledTime(getNextScheduledTime());`
--- a/src/foam/nanos/cron/Cron.js
+++ b/src/foam/nanos/cron/Cron.js
@@ -12,6 +12,8 @@ foam.CLASS({
   imports: [ 'cronDAO as scriptDAO' ],
 
   javaImports: [
+    'foam.dao.DAO',
+    'foam.nanos.notification.Notification',
     'java.util.Date',
     'java.util.Calendar'
   ],
@@ -83,7 +85,18 @@ foam.CLASS({
       ],
       javaReturns: 'void',
       javaCode:
-`super.runScript(x);
+`DAO notification = (DAO) x.get("notificationDAO");
+    
+Notification cronStartNotify = new Notification();
+cronStartNotify.setBody(this.getId() + " started");
+notification.put(cronStartNotify);
+
+super.runScript(x);
+
+Notification cronEndNotify = new Notification();
+cronEndNotify.setBody(this.getId() + " ended");
+notification.put(cronEndNotify);
+
 setScheduledTime(getNextScheduledTime());`
     },
     {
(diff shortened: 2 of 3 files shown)


src/foam/nanos/cron/Cron.js
```diff
@@ -83,7 +85,18 @@ foam.CLASS({
       ],
       javaReturns: 'void',
       javaCode:
-`super.runScript(x);
+`DAO notification = (DAO) x.get("notificationDAO");
+    
+Notification cronStartNotify = new Notification();
+cronStartNotify.setBody("Cron job: " + this.getId() + " started");
```
jlhughes: Can you add the description as well, incase one uses the auto generated id which is just a number. - so use both.

ZacZhang: added cron job description

src/foam/nanos/auth/Group.js
```diff
@@ -185,7 +185,7 @@ if ( session != null ) {
       //populate AppConfig url with request's RootUrl
       HttpServletRequest req = x.get(HttpServletRequest.class);
 
-      if ( config.getEnableForceHttps() ) {
+      if ( config.getForceHttps() ) {
```
ronin00, line 188: Just a style suggestion.  Usually for boolean properties isWhatever() is oft recommended over getWhatever() for readbility purposes.  Assuming the is-verb methods are generated by foam2:
```suggestion
      if ( config.isForceHttps() ) {
```

src/foam/nanos/cron/Cron.js
```diff
@@ -86,15 +87,25 @@ foam.CLASS({
       javaReturns: 'void',
       javaCode:
 `DAO notification = (DAO) x.get("notificationDAO");
-    
+            
 Notification cronStartNotify = new Notification();
-cronStartNotify.setBody("Cron job: " + this.getId() + " started");
+StringBuilder startMsg = new StringBuilder();
+startMsg.append("Cron job: ").append(this.getId()).append(" started.");
+if ( ! SafetyUtil.isEmpty(this.getDescription()) ) {
```
jlhughes: Ok, this is messy.
how about: 
Cron STARTED - id description 
Cron ENDED - id description

src/foam/nanos/cron/Cron.js
```diff
@@ -83,7 +86,28 @@ foam.CLASS({
...
+`DAO notification = (DAO) x.get("notificationDAO");
+            
+Notification cronStartNotify = new Notification();
+StringBuilder startMsg = new StringBuilder();
+startMsg.append("Cron job: ").append(this.getId()).append(" started.");
+if ( ! SafetyUtil.isEmpty(this.getDescription()) ) {
+  startMsg.append(" Cron job description: ").append(this.getDescription());
+}
+cronStartNotify.setBody(startMsg.toString());
+notification.put(cronStartNotify);
```
ronin00, line 93: Wondering ...
1) Is Cron the terminology for scheduling used in the app/UI?  If not would it be better to use the same terminology in for user facing notifications?  e.g: Scheduled Job #123 started
2) Can we shorten the msg?  Maybe no need to repeat the phrase 'Cron job description' everytime.  Maybe better to compose the msg more tersely.  e.g:
```Scheduled job #<job-id> '<job-description>' started```
or perhaps
```Scheduled job `<job-description>` (id#<job-id>) ended```

**Review by ronin00 — COMMENTED** [review]
Few possible style points ..

File: src/foam/nanos/cron/Cron.js (hunk 3 above)
Comment (ronin00) [reply src/foam/nanos/cron/Cron.js]: Actually I vote for @jlhughes's suggestion above  than what I suggested below as well.

Above is even more simpler -- easier to do grepping on, etc.

File: src/foam/nanos/cron/Cron.js (hunk 4 above)
Comment (ronin00) [reply src/foam/nanos/cron/Cron.js:93]: For point 2  .. see @jlhughes suggestion earlier

File: src/foam/nanos/auth/Group.js (hunk 2 above)
Comment (ronin00) [inline src/foam/nanos/auth/Group.js:188]: And regarding lines 172 - 208 of Group.js:
Why is the whitespace formatting inconsistent with the rest of the file?  Usually newly inserted code should match the style set by the existing file (unless what was there before was really bad and inconsistent itself).

File: src/foam/nanos/cron/Cron.js (hunk 4 above)
Comment (jlhughes) [reply src/foam/nanos/cron/Cron.js:93]: Crons will be administered by 'admin's of the system, not regular users.

File: src/foam/nanos/auth/Group.js (hunk 2 above)
Comment (jlhughes) [reply src/foam/nanos/auth/Group.js:188]: re: styling, we have a style guide and we recommend to update a file touched to the new styling, but hopefully in a different PR.

Comment (jlhughes) [reply src/foam/nanos/auth/Group.js:188]: Java generation - we generate getFoo, on javascript you have access to the variable so just 'foo'
so for javascipt - 'foo' is fine, and I would expect getFoo from java.

**Review by ronin00 — CHANGES_REQUESTED** [review]
@ZacZhang 
Regarding your earlier change to Group.js via [PR#1589](https://github.com/foam-framework/foam2/pull/1589)
And @jlhughes additional review req via [NP Issue](https://github.com/nanoPayinc/NANOPAY/issues/3890)

@jlhughes was prob referring to the fact that the handling of the HTTPS enforcement could prob be simplified for a bit of clarity / commented on 2nd thought.

Maybe something like this for the entire block ...?
```
      String reqUrlStr = ((req != null) && !SafeUtil.isEmpty(req.getRequestURI())) ? ((Request)req).getRootURL().toString() : null;

      // Check to see if incoming URL is even valid ...
      if (! SafeUtil.isEmpty(reqUrlStr)) {
        // Just use whatever is present in base config.  Assumption is that it'll be appropriately configured including HTTP/HTTPS
        return config;
      }

      // If incoming URL is already HTTPS just use it as is.  This covers the 'normal' scenario where the client is hitting the
      // app directly w/ HTTPS.  i.e: With no load-balancer or anything else in between stripping the HTTPS and forwarding HTTP.
      if (reqUrlStr.startsWith("https")) {
        config.setUrl(reqUrlStr);
        return config;
      }

      // At this point incoming URL is not HTTPS leading to two possible scenarios:
      // a) this is just a simple HTTP deployment w/ client hitting app directly via HTTP
      // b) this is an HTTPS deployment but we have an entity (e.g: load-balancer) stripping HTTPS and forwarding HTTP.  But we then need to
      //   enforce HTTPS on any URLs we send back out in response.
      //
      // Whether this is a HTTP or HTTPS deployment requiring HTTPS enforcement is decided by the forceHttps flag set at deployment time
      if ( config.getForceHttps() ) {
            // HTTPS required ... so replace the HTTP scheme prefix
            config.setUrl("https" + reqUrlStr.substring(4));
          } else {
            // HTTPS not required ... so just pass through
            config.setUrl(request);
          }
        }
      } else {
        }
      }
```

**Comment by jlhughes** [comment]
Summary of requirements.
```
group.getAppConfig() {
  config = ctx get "appConfig"
  if requestUrl
    config.url = request url
  else if group.url 
    config.url = group.url
  
  if forceHttps
    set https on url

  return config
```

jlhughes (approved):

PR #1625, opened 2018-10-18
side menu style changes with layers

KateMengjunXia:
Changes on side menu:
- changed background colour for consistency
- added styling for hover, select, etc
- changed label styles as levels went down


src/foam/nanos/u2/navigation/BusinessLogoView.js
```diff
@@ -20,64 +20,71 @@ foam.CLASS({
...
         .addClass(this.myClass())
         .start()
-          .start({ class: 'foam.u2.tag.Image', data: this.logo }).end()
+          .tag({class: 'foam.u2.tag.Image', data$: logoSlot})
+          .on('click', self.goToDefault)
         .end();
     }
+  ],
+
+  listeners: [
```
kgrgreer: What is this for?

KateMengjunXia: I accidentally pushed my change for a different ticket to this PR just now. Will remove the BusinessLogoView.js changes.

src/foam/nanos/u2/navigation/SideNavigationItemView.js
```diff
@@ -1,45 +1,74 @@
...
+        .addClass(this.myClass())
+        .start('a')
+          .add(this.data.label)
+          .enableClass(this.myClass('selected'), view.currentMenu$.map(function (value) {
+            return window.location.hash.substring(1) === (view.data.id);
+          }))
+          .style({'padding-left': view.paddingLeft + 'px', 'font-size': view.fontSize + 'px', 'opacity': view.opacity})
+          .on('click', function() {
+            var menu = view.data;
+            menu.children.select().then(function(c) {
```
kgrgreer, line 55: I don't think this is required. Just always call launch() and set expanded.

src/foam/nanos/u2/navigation/SideNavigationItemView.js
```diff
@@ -1,45 +1,74 @@
...
+    }
+  `,
+
+  properties: [
+    {
+      class: 'Boolean',
+      name: 'expanded',
+      value: true
+    },
+    {
```
kgrgreer, line 27: Just have one property called 'level' and have Item figure out its own style.

src/foam/nanos/u2/navigation/SideNavigationItemView.js
KateMengjunXia: menu.launch() doesn't seem to work when c.array.length !== 0 (when menu is a submenu that has children). because the launch method in SubMenu.js calls the open method in PopupMenu.js, which doesn't work here. 
Should I fix launch, so that it works for a submenu in the side bar (instead of just the ones in top navigation), or should I just leave the if condition?

kgrgreer: Can't you check the type rather than looking at the length?

KateMengjunXia: Updated code is in PR: https://github.com/foam-framework/foam2/pull/1634

PR #1626, opened 2018-10-19
Discard empty group permissions

sirenChen:
Refs #4036
On save, filter out all the empty entries in the preset method.

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -49,7 +49,7 @@ foam.CLASS({
       preSet: function(oldVal, newVal) {
         // if old value is undefined, that means save action triggered
         return oldVal !== undefined ? newVal :
-          newVal.filter(permission => {return permission.id});
+          newVal.filter(permission => permission.id);
       }
     },
     {
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -49,7 +49,7 @@ foam.CLASS({
       preSet: function(oldVal, newVal) {
         // if old value is undefined, that means save action triggered
         return oldVal !== undefined ? newVal :
-          newVal.filter(permission => {return Object.keys(permission.instance_).length !== 0});
+          newVal.filter(permission => {return permission.id});
       }
     },
     {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/Group.js
```diff
@@ -45,7 +45,12 @@ foam.CLASS({
       class: 'FObjectArray',
       of: 'foam.nanos.auth.Permission',
       name: 'permissions',
-      documentation: 'Permissions set on group.'
+      documentation: 'Permissions set on group.',
+      preSet: function(oldVal, newVal) {
+        // if old value is undefined, that means save action triggered
+        return oldVal !== undefined ? newVal :
```
kgrgreer, line 51: What is this check for? Why not just always filter the newVal?

src/foam/nanos/auth/Group.js
```diff
@@ -45,7 +45,12 @@ foam.CLASS({
...
+          newVal.filter(permission => {return Object.keys(permission.instance_).length !== 0});
```
kgrgreer: As a programmer you shouldn't be using .instance_ directly as that's an internal structure of the objects. Also, it still doesn't ensure that the permission id is set. The id could be empty and the description could be set which would trick this check. Just do 'return permission.id;'

src/foam/nanos/auth/Group.js
sirenChen: When we click 'Add Item' button, it will tigger FobjectArrayView 'addItem' action. It will clone the object, and call preSet method eventually.
So if we always filter the newVal, you can never add a new item after you click 'Add Item' button.

src/foam/nanos/auth/Group.js
```diff
@@ -45,7 +45,12 @@ foam.CLASS({
       class: 'FObjectArray',
       of: 'foam.nanos.auth.Permission',
       name: 'permissions',
-      documentation: 'Permissions set on group.'
+      documentation: 'Permissions set on group.',
+      preSet: function(oldVal, newVal) {
+        // if old value is undefined, that means save action triggered
+        return oldVal !== undefined ? newVal :
+          newVal.filter(permission => {return permission.id});
```
TW80000: This can be simplified to 
```suggestion
          newVal.filter(permission => permission.id);
```

sirenChen:
Close this pull request, will submit a new one later.

PR #1627, opened 2018-10-19
Show disabled context menu button in TableView when there are no actions

### LINKED ISSUE #1614: Show disabled context menu button in TableView when there are no actions

TW80000:
Closes #1614 

screen shot 2018-10-19 at 10 41 07 am

Also made the disabled colour a little bit darker for better contrast.

### DIFF
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -268,49 +268,39 @@ foam.CLASS({
                     var actions = allActions.filter(function(action) {
                       return action.isAvailableFor(obj);
                     });
-
-                    // Show the context menu in a disabled state if there are
-                    // no available actions.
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -269,13 +269,18 @@ foam.CLASS({
...
-                      if ( view.editColumnsEnabled ) {
-                        return this.tag('td');
-                      }
-                      return;
+                      return this.start('td').
+                        style({ 'text-align': 'right' }).
+                        start('span').
+                          addClass(view.myClass('vertDots')).
+                          addClass(view.myClass('noselect')).
+                          addClass('disabled').
```
kgrgreer: Why is this always added? Seems to be duplicating a lot of the code from below.

PR #1628, opened 2018-10-19
Some tweaks around flow documents

mcarcaso:
Rely on the documents being classloaded instead of being in files.js.
Make css class of a document configurable.
Create a PromiseSlot and use it to classload foam tags in flow documents.


src/foam/flow/Document.js
```diff
@@ -327,33 +336,42 @@ foam.CLASS({
...
-              cssClass('foam-flow-Document-code').
-              add(code).
-              end();
+            this.start('code').add(code).end();
           };
         },
 
         'foam': function(attributes) {
           return function(x) {
-            // TODO: Reuse FoamTagLoader support, support classloading
```
adamvy, line 340: Keep the TODO about reusing FoamTagLoader support as that's about re-using the code in FoamTagLoader, but you can delete the part about classloading :)

src/foam/flow/PromiseSlot.js
```diff
@@ -0,0 +1,33 @@
...
+  implements: [ 'foam.core.Slot' ],
+  documentation: `
+    A slot that takes a promise and sets its value to its value when it
+    resolves.
+  `,
+  properties: [
+    {
+      name: 'promise',
+      postSet: function(_, n) {
+        n.then(function(v) {
```
adamvy: We could consider detaching everyone listening to us at this point, as a Promise can only resolve once and it could help prevent memory leaks if others forget to unsub, but maybe that's not necessary.

src/foam/flow/PromiseSlot.js
```diff
@@ -0,0 +1,33 @@
...
+          if ( n === this.promise ) this.value = v;
+        }.bind(this));
```
adamvy: We should also handle the error case of the promise I think.  Not sure what we should do, but I'm not sure we should just ignore it.

adamvy: Maybe its okay, since we ignore exceptions in most other places and a Promise rejection is basically an exception...  not sure

src/foam/flow/Document.js
```diff
@@ -327,33 +336,42 @@ foam.CLASS({
...
             var viewName = attributes.view;
             var className = attributes.class;
-            var cls = x.lookup(className, true);
-            var view = x.lookup(viewName, true);
 
-            if ( className && ! cls )
-              this.add('Unknown class', className);
-            if ( viewName && ! view )
-              this.add('Unknown view', viewName);
+            var slot = foam.flow.PromiseSlot.create({
```
adamvy: As discussed, let's move PromiseSlot to core/Slot.js as its probably useful, but let's also update Element.js to support just adding a Promise to the DOM

src/foam/flow/PromiseSlot.js
mcarcaso: We'll need to decide whether we want to support the ability to change the promise or not. If we don't care about supporting that we can detach listeners but if we want that, we can't. Currently, it supports the promise being able to change.

src/foam/flow/PromiseSlot.js
mcarcaso: Yeah not sure what to do either but I'm ok with ignoring this until we do think of a good way to handle it rather than rushing in behaviour that we might not end up keeping.

src/foam/flow/Document.js
mcarcaso: Moved to Slot.js and added a promise(promise, func) method to Element.

src/foam/flow/Document.js
mcarcaso: Done

PR #1630, opened 2018-10-19
Change logic of forceHttps

ZacZhang:
https://github.com/nanoPayinc/NANOPAY/issues/3890
https://github.com/foam-framework/foam2/pull/1589

Current Logic:
if (group.url exist) {
  config.url = group.url
}

else if (http request exist) {
  if (forceHttps && config.getUrl().startsWith("https")) {
    set https on request url
  }
  config.url = requestr url
}

else {
 use default config.url
}

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -176,7 +176,7 @@ AppConfig config = (AppConfig) ((AppConfig) x.get("appConfig")).fclone();
 String configUrl = config.getUrl();
 
 HttpServletRequest req = x.get(HttpServletRequest.class);
-if ( (req != null) && !SafetyUtil.isEmpty(req.getRequestURI()) ) {
+if ( (req != null) && ! SafetyUtil.isEmpty(req.getRequestURI()) ) {
   // populate AppConfig url with request's RootUrl
   configUrl = ((Request) req).getRootURL().toString();
 } else {
@@ -185,7 +185,7 @@ if ( (req != null) && !SafetyUtil.isEmpty(req.getRequestURI()) ) {
   User user = (User) userDAO.find(session.getUserId());
   if ( user != null ) {
     Group group = (Group) groupDAO.find(user.getGroup());
-    if ( SafetyUtil.isEmpty(group.getUrl()) ) {
+    if ( ! SafetyUtil.isEmpty(group.getUrl()) ) {
       configUrl = group.getUrl();
     }
   }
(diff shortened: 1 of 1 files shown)


ronin00:
**NOTE: Do not merge.  Doesn't fully satisfy the points discussed here:**
https://github.com/foam-framework/foam2/pull/1624#issuecomment-431195443

jlhughes:
This is what I want.

group.getAppConfig() {
  config = ctx get "appConfig"
  if requestUrl
    config.url = request url
  else if group.url 
    config.url = group.url
  
  if forceHttps
    set https on url

  return config

NOTE: test for leading http:// before stripping, we want to move to having URL just contain the domain name, without the protocol prefix.   Which will be the next PR, but it includes build/jenkins changes.

src/foam/nanos/auth/Group.js
```diff
@@ -185,18 +185,25 @@ if ( session != null ) {
       //populate AppConfig url with request's RootUrl
       HttpServletRequest req = x.get(HttpServletRequest.class);
 
-      if ( req != null && ! SafetyUtil.isEmpty(req.getRequestURI()) ) {
+      if ( (req != null) && !SafetyUtil.isEmpty(req.getRequestURI()) ) {
```
jlhughes: Why do we test for session above?  What happens when cron runs and there is no session?

ZacZhang: Actually, we can remove the session check. 
I think session cannot be null, even for cron job, the session is 
screen shot 2018-10-22 at 11 00 20 am

jlhughes: Great, one less thing.

src/foam/nanos/auth/Group.js
```diff
@@ -173,38 +173,30 @@ DAO userDAO      = (DAO) x.get("localUserDAO");
...
-          config.setUrl(((Request) req).getRootURL().toString());
-        }
-      }
+      configUrl = group.getUrl();
     }
   }
 }
+
+if ( config.getForceHttps() && configUrl.startsWith("http://") ) {
+  configUrl = "https" + configUrl.substring(4);
```
jlhughes: Need to handle the url being just the domain name like ‘portal.nanopay.net’ and with protocol: ‘http://portal.nanopay.net’.   If forceHttps is true, then both need to become ‘https://portal.nanopay.net’

ZacZhang: updated code

jlhughes (approved):

PR #1632, opened 2018-10-19
Break up the API doc components

mcarcaso:
Splits them up into multiple models/files. This will make them easier to add to flow documents.


src/foam/doc/AxiomListView.js
```diff
@@ -0,0 +1,81 @@
...
+    'foam.doc.dao.AxiomDAO',
+    'foam.mlang.sink.Count',
+  ],
+  implements: [
+    'foam.mlang.Expressions',
+  ],
+  properties: [
+    {
+      class: 'String',
+      name: 'modelID',
```
adamvy: I think naming convention is modelId not modelID

mcarcaso: done

PR #1635, opened 2018-10-22
Swift FObjectArray tests

mcarcaso:
Also fix the swift project to know about all classes/files so it can compile again.


kennyqiyenkan (approved):
lgtm

PR #1637, opened 2018-10-23
AuthenticatedAgentUserJunctionDAO

zahnno:
### Authenticated Agent User Junction DAO 
 * Restrict users from creating a junction object where they are the source user (Agent)
 * Restrict users from finding and selecting junction objects they are not referenced to. (Agent or Entity)
 * Restrict users from removing junction objects where they aren't the source user or target user (Agent and Entity)

Tests incl. as AuthenticatedJunctionDAOTest


src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,160 @@
...
+
+import static foam.mlang.MLang.EQ;
+import static foam.mlang.MLang.OR;
+import static foam.mlang.MLang.AND;
+
+/**
+ * Authenticated AgentJunctionDAO
+ * Restrict users from creating a junction object where they are the source user (Agent)
+ * Restrict users from finding and selecting junction objects they are not referenced to. (Agent or Entity)
+ * Restrict users from removing junction objects where they are the source user (Agent)
```
kgrgreer: Shouldn't this be from removing junction objects where they aren't the target user? This sounds like I could remove other people's junctions.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -43,7 +43,7 @@ public void start() {
   */
   public User actAs(X x, User entity) throws AuthenticationException {
     User agent = (User) x.get("user");
-    User user = (User) userDAO_.find(entity.getId());
+    User user = (User) userDAO_.find_(x, entity.getId());
```
deejes, line 47: I feel like entity might be a clearer name than user for this variable?

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,160 @@
...
+import foam.dao.DAO;
+import foam.dao.ProxyDAO;
+import foam.dao.Sink;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+import foam.util.SafetyUtil;
+
+import static foam.mlang.MLang.EQ;
+import static foam.mlang.MLang.OR;
+import static foam.mlang.MLang.AND;
```
deejes: Its nitpicky, but you could sort the static imports

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -43,7 +43,7 @@ public void start() {
...
```
TW80000, line 46: Shouldn't this be the following?

```suggestion
    User user = (User) userDAO_.inX(x).find(entity.getId());
```

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -54,7 +54,7 @@ public User actAs(X x, User entity) throws AuthenticationException {
       throw new AuthorizationException("Entity user doesn't exist.");
     }
 
-    Group group = (Group) groupDAO_.find(user.getGroup());
+    Group group = (Group) groupDAO_.find_(x, user.getGroup());
```
TW80000, line 57:
```suggestion
    Group group = (Group) groupDAO_.inX(x).find(user.getGroup());
```

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -78,7 +78,7 @@ public User actAs(X x, User entity) throws AuthenticationException {
     }
 
     // Junction object contains a group which has a unique set of permissions specific to the relationship.
-    Group actingWithinGroup = (Group) groupDAO_.find(permissionJunction.getGroup());
+    Group actingWithinGroup = (Group) groupDAO_.find_(x, permissionJunction.getGroup());
```
TW80000, line 81:
```suggestion
    Group actingWithinGroup = (Group) groupDAO_.inX(x).find(permissionJunction.getGroup());
```

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,160 @@
...
+  @Override
+  public FObject put_(X x, FObject obj) {
+    User user = (User) x.get("user");
+    User agent = (User) x.get("agent");
+    AuthService auth    = (AuthService) x.get("auth");
+
+    UserUserJunction junctionObj = (UserUserJunction) obj;
+
+    // Permit permission check if agent is not present within context.
+    if ( agent == null && auth.check(x, GLOBAL_AGENT_JUNCTION_UPDATE)) {
```
TW80000:
```suggestion
    if ( agent == null && auth.check(x, GLOBAL_AGENT_JUNCTION_UPDATE) ) {
```

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Allows put if user or agent is target of the junction object.
+   * If agent exists within context, condition checks on agent user.
+   */
+  @Override
+  public FObject put_(X x, FObject obj) {
+    User user = (User) x.get("user");
+    User agent = (User) x.get("agent");
+    AuthService auth    = (AuthService) x.get("auth");
```
kgrgreer: Remove extra spaces before =.

kgrgreer: Or line up all of the ='s.

TW80000 (changes requested):
I'm going to stop reviewing here because I think AuthenticatedAgentJunctionDAO could be simplified quite a bit. We already have [a generic AuthenticatedUserUserJunctionDAO](https://github.com/nanoPayinc/NANOPAY/blob/development/nanopay/src/net/nanopay/partners/AuthenticatedUserUserJunctionDAO.java) that can be used to satisfy requirements 2 and 3. Then we can make this into a much simpler decorator that only has to satisfy requirement 1.

On a different note, I think we need a 4th requirement, which is to prevent the group from being changed by the agent or the entity to one that has permissions the caller doesn't have. It's the same way we prevent users from updating their own group in AuthenticatedUserDAO. Similar to what we do here: https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/auth/AuthenticatedUserDAO.java#L79

So in summary, the requirements are:

1. Restrict users from creating a junction object where they are the source user (Agent)
2. Restrict users from finding and selecting junction objects they are not referenced to. (Agent or Entity)
3. Restrict users from removing junction objects where they aren't the target user or source user. (Agent or Entity)
4. Prevent privilege escalation by restricting what the group can be set to.

and my proposed solution is to use AuthenticatedUserUserJunctionDAO to solve 2 and 3 because it's already written and AuthenticatedAgentJunctionDAO to solve 1 and 4. It already does 1.

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,160 @@
...
+
+    // Check agent or user to authorize the request as.
+    User authorizedUser = agent != null ? agent : user;
+
+    // Check junction object relation to authorized user.
+    boolean authorized =
+        ( SafetyUtil.equals(junctionObj.getTargetId(), authorizedUser.getId()) ||
+        SafetyUtil.equals(junctionObj.getSourceId(), authorizedUser.getId()) );
+
+    return ( auth.check(x, permission) && agent == null ) || authorized;
```
kgrgreer: Faster to check agent == null first since then you don't do the expensive check() if it isn't. Also, faster to check authorized first. So the following is equivalent but faster:
  return authorized || ( agent == null && auth.check(x, permission) && agent == null );

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  // Returns predicated delegate based on user and agent in context.
+  public DAO getFilteredDAO(X x, String permission) {
+    User user = (User) x.get("user");
+    User agent = (User) x.get("agent");
+    AuthService auth = (AuthService) x.get("auth");
+
+    if ( auth.check(x, permission) && agent == null )
```
kgrgreer: Reverse sides of && for better performance.

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: Sure, but entity will become the user within the context. I envisioned it as a way to preempt that idea.

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,160 @@
...
+      return getDelegate().put_(x, junctionObj);
+    }
+
+    // Check agent or user to authorize the request as.
+    User authorizedUser = agent != null ? agent : user;
+
+    // Check junction object relation to authorized user.
+    boolean authorized = SafetyUtil.equals(junctionObj.getTargetId(), authorizedUser.getId());
+
+    if ( junctionObj != null && ! authorized ) {
```
TW80000: I think this null check should be before line 61, otherwise it's not doing anything really.

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,160 @@
...
+ * Restrict users from creating a junction object where they are the source user (Agent)
+ * Restrict users from finding and selecting junction objects they are not referenced to. (Agent or Entity)
+ * Restrict users from removing junction objects where they aren't the target user or source user. (Agent or Entity)
+ */
+public class AuthenticatedAgentJunctionDAO
+  extends ProxyDAO
+{
+
+  public final static String GLOBAL_AGENT_JUNCTION_READ   = "agentJunction.read.*";
+  public final static String GLOBAL_AGENT_JUNCTION_UPDATE = "agentJunction.update.*";
```
TW80000, line 33: Shouldn't we have another permission for creating a new junction as well? Eg: `"agentJunction.create"`

zahnno:
### Troy' comments were handled offline and agreed upon. Here's the summary of our conversation:

patrick [10:54 AM]
The authenticatedUserUserJunctionDAO you created doesn't account for the possiblilty of the agent within the context

It currently exists within nanopay aswell and would need to be ported over
and by making the authenticatedAgentJunctionDAO simpler, I'm not sure what you mean by that
> On a different note, I think we need a 4th requirement, which is to prevent the group from being changed by the agent or the entity to one that has permissions the caller doesn't have. It's the same way we prevent users from updating their own group in AuthenticatedUserDAO.

^ I agree
We can address this later on\

Troy [10:59 AM]
> The authenticatedUserUserJunctionDAO you created doesn't account for the possiblilty of the agent within the context

I don't think it should be checking the agent. If we check the agent then how are you supposed to act as a business and add someone to your business?

> It currently exists within nanopay aswell and would need to be ported over

True, but I don't see why it couldn't be ported over. It doesn't use any nanopay-specific models or code.

patrick [11:00 AM]
We'll need to extend this authenticatedJunctionDAO to account for permissions allowing users to add to the junction

Troy [11:01 AM]
> and by making the authenticatedAgentJunctionDAO simpler, I'm not sure what you mean by that

It would be simpler because it wouldn't have to duplicate all of the logic from the generic AuthenticatedUserUserJunctionDAO

patrick [11:01 AM]
This extended authenticatedJunction will be extend within our nanopay platform
AuthenticatedUserUserJunctionDAO was created for something other than the agentJunctionDAO
I believe it satisfies conditions required for the partners functionality

> True, but I don't see why it couldn't be ported over. It doesn't use any nanopay-specific models or code.

Troy [11:02 AM]
> AuthenticatedUserUserJunctionDAO was created for something other than the agentJunctionDAO

It was written so we wouldn't have to duplicate authentication decorator logic every time we have a junction DAO between two users. This is a perfect use case for it.

patrick [11:03 AM]
^ it'll require alternations already provided by the AuthenticatedAgentJunctionDAO

A verbal conversation continued to conclude that the AuthenticatedAgentJunctionDAO will be extended to include business permissions and prevent groups from being added to the userJunction will be addressed later.

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
TW80000:
```suggestion
    return authorized || auth.check(x, permission);
```

Old:
If I'm acting as someone and the agent is target or source: true
If I'm acting as someone and the agent is not target or source: false
If I'm not acting as someone and the user is target or source: true
If I'm not acting as someone and the user is not target or source: Permission check on user's group

New:
If I'm acting as someone and the agent is target or source: true
If I'm acting as someone and the agent is not target or source: Permission check on junction group
If I'm not acting as someone and the user is target or source: true
If I'm not acting as someone and the user is not target or source: Permission check on user's group

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
TW80000: Same reason as above. If we did want a junction group to give an agent such a permission, we should check it.

```suggestion
    if ( auth.check(x, permission) )
```

TW80000 (changes requested):

src/foam/nanos/auth/AgentUserAuthService.java
zahnno: This would be valid but since I'm passing an id into the method it doesn't make much of a difference. However I'll get into the habit of using inX since it avoids potential context related issues.

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
zahnno: I don't see a use case for adding it for the current authentication logic.

src/tests
```diff
@@ -3,5 +3,5 @@ p({"class":"foam.nanos.test.Test","id":"MDAO test","description":"MDAO test","se
...
+p({"class":"foam.dao.SequenceNumberDAOTest","id":"SequenceNumberDAO test","description":"SequenceNumberDAO tests"})r({"class":"foam.nanos.test.Test","id":"AuthenticatedJunctionDAOTest"})
```
TW80000: Did you mean to remove the AuthenticatedJunctionDAOTest?

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,164 @@
...
+      return getDelegate().put_(x, junctionObj);
+    }
+
+    // Check agent or user to authorize the request as.
+    User authorizedUser = agent != null ? agent : user;
+
+    // Check junction object relation to authorized user.
+    boolean authorized = SafetyUtil.equals(junctionObj.getTargetId(), authorizedUser.getId());
+
+    if ( junctionObj != null && ! authorized ) {
```
TW80000: The null check is redundant since there's one on line 52.

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
```diff
@@ -0,0 +1,164 @@
...
+  public AuthenticatedAgentJunctionDAO(X x, DAO delegate) {
+    super(x, delegate);
+  }
+
+  /**
+   * Allows put if user or agent is target of the junction object.
+   * If agent exists within context, condition checks on agent user.
+   */
+  @Override
+  public FObject put_(X x, FObject obj) {
```
TW80000, line 45: This is missing the code restricts what the group can be set to. If this were merged as is, anyone could give themselves super-admin privileges.

src/tests
zahnno: I updated the AuthenticatedJunctDAOTest. Good catch that line was written to the jrl incorrectly.

TW80000 (changes requested):

src/tests
TW80000: `r({"class":"foam.nanos.test.Test","id":"AuthenticatedJunctionDAOTest"})`

That's at the end of the line. That would remove the test wouldn't it?

src/foam/nanos/auth/AuthenticatedAgentJunctionDAO.java
zahnno: As mentioned before I was going to address this in a later PR and account for the changes in the test but I've went ahead and added it in https://github.com/foam-framework/foam2/pull/1637/commits/3ecb2f98c65d21f438b880540237487b5a4dd3b3

TW80000 (approved):

TW80000:
I think this was superseded by https://github.com/nanoPayinc/NANOPAY/pull/5495.

jlhughes:
superceded by https://github.com/nanoPayinc/NANOPAY/pull/5495

PR #1639, opened 2018-10-24
Add 'System' flag to User model

sirenChen:
Please see tickets 4119 in Product Backlog.
Add 'System' flag to User model.


src/foam/nanos/auth/User.js
```diff
@@ -286,6 +286,11 @@ foam.CLASS({
       class: 'DateTime',
       name: 'lastModified',
       documentation: 'Last modified date.'
+    },
+    {
+      class: 'Boolean',
+      name: 'system',
+      documentation: 'Indicate system accounts.'
```
jlhughes, line 294: please explicitly default to false:
value: false

jlhughes:
If possible, please create a TestCase.  The foam side testing is a little different, you can inquire on the foam channel if you can't find examples.

src/foam/nanos/auth/User.js
```diff
@@ -290,6 +290,7 @@ foam.CLASS({
     {
       class: 'Boolean',
       name: 'system',
+      value: false,
```
kgrgreer, line 293: We already have the 'type' property. Couldn't we just add 'system' as a new type?

src/foam/nanos/auth/AuthenticatedUserDaoTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+
+    //test case
+    AuthenticatedUserDAO_AdminUpdateSystemFlag(adminContext);
+    AuthenticatedUserDAO_AdminCreateUserWithSystemFlag(adminContext);
+    AuthenticatedUserDAO_NonAdminUpdateSystemFlag(nonAdminContext);
+    AuthenticatedUserDAO_NonAdminCreateUserWithSystemFlag(nonAdminContext);
+  }
+
+  private void AuthenticatedUserDAO_AdminUpdateSystemFlag(X adminContext) {
```
kgrgreer: Don't make methods private. Just makes harder to test.

src/foam/nanos/auth/AuthenticatedUserDaoTest.java
```diff
@@ -0,0 +1,145 @@
+package foam.nanos.auth;
+
+import foam.core.X;
+import foam.dao.DAO;
+import foam.dao.MDAO;
+import foam.test.TestUtils;
+import foam.util.Auth;
+import foam.util.SafetyUtil;
+
+public class AuthenticatedUserDaoTest extends foam.nanos.test.Test {
```
kgrgreer: Dao -> DAO

src/foam/nanos/auth/User.js
jlhughes: @kgrgreer  We plan to sub-class User into these types such as BusinessUser so the type as it exists now will go away.

jlhughes (commented):
comment deleted

src/foam/nanos/auth/User.js
jlhughes: defaults to false, no need.

PR #1640, opened 2018-10-24
Avoid launching the same side menu item contents

KateMengjunXia:
Avoid loading the same side menu item contents when they are already loaded.

### DIFF
--- a/src/foam/nanos/u2/navigation/SideNavigationItemView.js
+++ b/src/foam/nanos/u2/navigation/SideNavigationItemView.js
@@ -48,7 +48,7 @@ foam.CLASS({
           .style({'padding-left': paddingLeft +  'px', 'font-size': fontSize + 'px', 'opacity': opacity})
           .on('click', function() {
             var menu = view.data;
-            if (menu.handler !== 'foam.nanos.menu.SubMenu' && (window.location.hash.substring(1) !== menu.id)) {
+            if (menu.handler != 'foam.nanos.menu.SubMenu' && (window.location.hash.substring(1) !== menu.id)) {
               menu.launch(view.__context__, view);
             } 
             view.expanded = ! view.expanded;
--- a/src/foam/nanos/u2/navigation/SideNavigationItemView.js
+++ b/src/foam/nanos/u2/navigation/SideNavigationItemView.js
@@ -48,10 +48,8 @@ foam.CLASS({
           .style({'padding-left': paddingLeft +  'px', 'font-size': fontSize + 'px', 'opacity': opacity})
           .on('click', function() {
             var menu = view.data;
-            if ( menu.handler != 'foam.nanos.menu.SubMenu' ) {
-              if ((window.location.hash.substring(1) !== (menu.id))) {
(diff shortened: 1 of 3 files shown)


src/foam/nanos/u2/navigation/SideNavigationItemView.js
```diff
@@ -48,12 +48,10 @@ foam.CLASS({
           .style({'padding-left': paddingLeft +  'px', 'font-size': fontSize + 'px', 'opacity': opacity})
           .on('click', function() {
             var menu = view.data;
-            menu.children.select().then(function(c) {
-              if ( menu.handler != 'foam.nanos.menu.SubMenu' ) {
-                menu.launch(view.__context__, view);
-              }
-              view.expanded = ! view.expanded;
-            });
+            if (menu.handler != 'foam.nanos.menu.SubMenu' && (window.location.hash.substring(1) !== menu.id)) {
```
kgrgreer: Should import currentMenu and use that instead of hash. Will explain tomorrow.

src/foam/nanos/u2/navigation/SideNavigationItemView.js
```diff
@@ -49,7 +49,11 @@ foam.CLASS({
           .on('click', function() {
             var menu = view.data;
             if (menu.handler != 'foam.nanos.menu.SubMenu' && (view.currentMenu.id !== menu.id)) {
-              menu.launch(view.__context__, view);
+              if (menu.handler == 'foam.nanos.menu.LinkMenu') {
```
kgrgreer: Format like:
if ( menu.handler == 'foam.nanos.menu.LinkMenu' ) {

KateMengjunXia:
An attempt to fix the following situation, where menu items 1,2,3 are links to elements of document A, 4,5,6 are links to elements of document B. Previously, if A is loaded, and 4 is clicked, nothing will happen, because document.getElementsByName(4) won't work when B is not loaded. My attempt to fix this bug has a problem, the setTimeout in line 78 isn't a good practice, however, without it, handleClick(4) will be called before execution of handleClick(A) is done (before A is done loaded). Let me know if you have any suggestion on how to solve this.

A
 - 1
 - 2
 - 3

B
 - 4
 - 5
 - 6

KateMengjunXia:
> An attempt to fix the following situation, where menu items 1,2,3 are links to elements of document A, 4,5,6 are links to elements of document B. Previously, if A is loaded, and 4 is clicked, nothing will happen, because document.getElementsByName(4) won't work when B is not loaded. My attempt to fix this bug has a problem, the setTimeout in line 78 isn't a good practice, however, without it, handleClick(4) will be called before execution of handleClick(A) is done (before A is done loaded). Let me know if you have any suggestion on how to solve this.
> 
> A
> 
> * 1
> * 2
> * 3
> 
> B
> 
> * 4
> * 5
> * 6

Added a property "anchor" to DocumentFileMenu to handle linking to places in the a document.
However, the above mentioned problem still exists: document.getElementsByName (called by setAnchor in line 80) won't find the element until launch (line 77) is done execution.

PR #1641, opened 2018-10-24
Cache auth checks on client.

mcarcaso:


src/foam/nanos/auth/CachedAuthServiceProxy.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'CachedAuthServiceProxy',
+  extends: 'foam.nanos.auth.ProxyAuthService',
```
kgrgreer, line 10: I think this should be a decorator rather than extending ProxyAuthService.

src/foam/nanos/auth/CachedAuthServiceProxy.js
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  properties: [
+    {
+      class: 'Map',
+      name: 'cache',
+    },
+  ],
+  methods: [
+    function init() {
+      this.onDetach(this.user$.dot('group').sub(function() {
```
kgrgreer, line 26: User never changes, so you could just do this.user.group$.sub()

mcarcaso: Isn't this technically a little more robust if user ever does change in the future? What's the downside of using .dot? I wouldn't think the overhead would be that much more but I could be wrong.

src/foam/nanos/auth/CachedAuthServiceProxy.js
mcarcaso: Isn't extending a proxy a means to create a decorator? This way, I don't need to worry about any other methods on the interface working because they'll automatically get passed through to the delegate.

mcarcaso:
Ping?

src/foam/nanos/auth/CachedAuthServiceProxy.js
kgrgreer: Sorry, my mistake. I read Stub instead of Proxy somehow.

src/foam/nanos/auth/CachedAuthServiceProxy.js
kgrgreer: Yes, it is more robust, but all of our code is designed assuming user doesn't change, so you could go with the simpler code.

PR #1642, opened 2018-10-25
Start u2 mlangs

mcarcaso:
Creates two toE-able sinks and a demo showing their use.

To run demo, start a local server from foam2 route and go to http://localhost:8000/?model=foam.u2.mlang.Demo

### DIFF
--- a/src/foam/u2/mlang/Demo.js
+++ b/src/foam/u2/mlang/Demo.js
@@ -6,7 +6,11 @@ foam.CLASS({
     'foam.dao.EasyDAO',
     'foam.u2.mlang.Table',
     'foam.u2.mlang.Pie',
+    'foam.u2.mlang.Sequence',
     'foam.mlang.sink.Count',
+  ],
+  implements: [
+    'foam.mlang.Expressions',
   ],
   properties: [
     {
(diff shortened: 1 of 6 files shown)


src/foam/u2/mlang/Demo.js
```diff
@@ -0,0 +1,90 @@
+foam.CLASS({
```
kgrgreer: Can/should we put this somewhere under the demo package with the rest of the demos rather than in the main code base?

src/foam/u2/mlang/Pie.js
```diff
@@ -0,0 +1,38 @@
...
+  name: 'Pie',
+  extends: 'foam.mlang.sink.GroupBy',
+  requires: [
+    'foam.graphics.DataSource',
+    'foam.graphics.PieGraph',
+  ],
+  properties: [
+    {
+      name: 'graphArgs',
+      value: {
```
kgrgreer: This will make it harder to configure. Make individual properties. X and Y can be expressions of radius if not set.

src/foam/u2/mlang/Pie.js
```diff
@@ -0,0 +1,38 @@
...
+        if ( ! seriesValues.length ) seriesValues = [0];
+        var p = this.PieGraph.create(graphArgs);
+        p.seriesValues = seriesValues;
+        return p;
+      },
+    },
+  ],
+  methods: [
+    function toE(_, x) {
+      return x.E().add(this.graph_$);
```
kgrgreer, line 36: Could you instead:
function toE(args, X) {return this.graph_.toE(args, X); }

mcarcaso: I need to use the graph_$ slot because the view can change based on the expression.

src/foam/u2/mlang/Table.js
```diff
@@ -0,0 +1,46 @@
...
+      name: 'columns',
+    },
+  ],
+  methods: [
+    function put(o) {
+      if ( ! this.of ) this.of = o.cls_
+      this.SUPER(o);
+    },
+    function toE(_, x) {
+      var self = this;
```
kgrgreer: Just call this.view.toE();

src/foam/u2/mlang/Pie.js
mcarcaso: Why would this make it harder to configure? If you want to customize these args you just pass in a map to clobber this one and that map will make its way to the view.

src/foam/u2/mlang/Table.js
mcarcaso: need the view$ slot here as well because it can change based on expression.

src/foam/u2/mlang/Demo.js
mcarcaso: I moved it to the foam/demos package.

src/foam/u2/mlang/Pie.js
mcarcaso: The PieGraph could definitely use some work. It doesn't have very reasonable defaults and getting the radius, x/y, margin, and height/width right is very manual right now and I like to think we can make this easier to use. Once that's all fixed up, I should be able to ditch the default value here completely and a developer can trust it'll render fine without any tweak, but have the option to tweak if they desire.

kgrgreer: If I only want to change the radius but leave the colours or margin the same, then how would I do this? Also, this doesn't work as well from a FOAM tag.

mcarcaso: Oh gotcha. Yeah you're right about that but I do plan on removing these defaults from here completely once the PieGraph has reasonable defaults. That should solve the issue that you have with this right?

src/foam/u2/mlang/Pie.js
kgrgreer: ok

src/foam/u2/mlang/Pie.js
adamvy: Even with better defaults it would make sense to split these in to separate properties so its easier to only override the ones you want.  As we just discussed in person :)

mcarcaso: Done. I think longer term we might want to have a separate model that the Pie mlang and PieGraph pull in as traits but for now separate args will do the trick.

PR #1644, opened 2018-10-26
Swift Reference property changes

mcarcaso:
In swift, use `Any?` for the type of the Reference property because it doesn't carry much meaning. It's mostly just be passed into the find function that's generated by a reference property. These changes also make the return type of the find function optional. This change could break any current users of references but the fix is as simple as unwrapping the optional with a `!` or `?`

Other changes:
- Don't set the default value of `url` to `null` in `foam.nanos.auth.Group`. It's a string property so it shouldn't be set to `null`.
- Handle nulls in SwiftLib.js
- Hide a warning in ScrollingViewController


kgrgreer:
Please resolve conflicts.

PR #1650, opened 2018-11-02
Password view entropy add on

deejes:
Adds NewPasswordView that contains a password strength bar and some text reflective of how much complexity a password contains, as a deference against brute force attack.


kirkeaton (approved):

src/foam/u2/view/NewPasswordView.js
```diff
@@ -0,0 +1,149 @@
...
+      .start().
+      start('div').addClass('strenght-indicator').
+        start('div').addClass('outer').
+          start('div').addClass('strength').addClass(this.strength$).end().
+        end().
+        start('div').addClass('message').addClass(this.textStrength$).
+            add(this.textStrength$.map(function(textStrength) {
+            switch ( textStrength ) {
+              case ('text1'):
+                return 'That\'s a really weak password';
```
jlhughes, line 119: can these use 'messages' for locale'?

deejes: Not sure exactly what you mean, could you elaborate or point me to an example?

src/files.js
```diff
@@ -149,6 +149,7 @@ FOAM_FILES([
   { name: "foam/util/EmailTest" },
   { name: "foam/util/PasswordTest" },
   { name: "foam/util/SecurityUtilTest" },
+  { name: "foam/util/PasswordStrengthGauge" },
```
kgrgreer, line 152: This is in package foam.util but you require it below as just foam.PasswordStrengthGauge. In either case, we should find a better package than just foam/util.

kgrgreer:
The library we're bringing in is 803kb. The whole FOAM1 Gmail app (including FOAM) is only 120k, so this is very high overhead for such a small feature. Can we find a smaller library for doing this or do the check on the server where size doesn't matter (using a Java library)?

kgrgreer (changes requested):
Library too large.

kirkeaton:
@kgrgreer My argument for not putting this server side was that it would lead to an increased amount of unnecessary network calls. It would require a network call for every change on the input field.

jlhughes:
@eatonk @kgrgreer A compromise/alternative is to send a Bloom Filter of words client side, and our own logic for entropy.  Believe @deejes is looking into this.  Also, in the past it was that one did not want to use words, but more recently a string of words seems to be the way to go.  What is our policy on this?

deejes:
After speaking with @eatonk, went ahead with a server side implementation. Making a seperate pr for that on, closing this one.

PR #1651, opened 2018-11-05
Agent Junction Status

zahnno:
Provides agent junction' a status capable of dictating if a junction is valid providing the ability for users to act as others. This status can be toggle-able and provides users the opportunity to "turn off" other users from acting as them.


TW80000 (commented):
Wouldn't this make more sense as a simple 'enabled' property?

PR #1653, opened 2018-11-05
Foam demo

hchoura:


src/com/foam/demos/addClass/Controller.js
```diff
@@ -0,0 +1,81 @@
+/**
+ * @license Copyright 2018 The FOAM Authors. All Rights Reserved.
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'com.foam.demos.addClass',
```
kgrgreer, line 7: We don't own the foam.com domain, so we shouldn't use that package. Either use com.foamdev or just foam.

PR #1659, opened 2018-11-07
Modifications to the Side Navigation View

nanoscott:
- Making the color of the side navigation the same as the color of the top navigation
- Fixing positioning of the stack view when using the NavigationView
- Adding custom CSS class to each level of the side navigation


src/foam/nanos/u2/navigation/NavigationView.js
```diff
@@ -7,18 +7,20 @@ foam.CLASS({
...
   css: `
     ^ {
-      position: fixed;
       z-index: 10001;
     }
     ^top-nav {
-      width: calc(100% - 200px);
+      width: 100%;
       display: inline;
+      z-index: 10003;
```
KateMengjunXia: Why does z-index need to be specified here? It is already specified on a class level.

src/foam/nanos/u2/navigation/SideNavigation.js
```diff
@@ -59,7 +59,7 @@ foam.CLASS({
         .start()
           .addClass(this.myClass('side-nav'))
           .select(dao, function(menu) {
-            return foam.nanos.u2.navigation.SideNavigationItemView.create({ data: menu, level: 0 }, this);
+            return foam.nanos.u2.navigation.SideNavigationItemView.create({ data: menu, level: 0, expanded: true }, this);
```
KateMengjunXia, line 62: expanded is already set to true in SideNavigationItem.js.

src/foam/nanos/u2/navigation/SideNavigationItemView.js
```diff
@@ -37,9 +37,11 @@ foam.CLASS({
       var paddingLeft = view.level * 10 + 15;
       var fontSize = view.level * -1 + 18;
       var opacity = view.level * -0.1 + 0.9;
+      var childIsExpanded = (view.level + 1) == 1;
 
       this
         .addClass(this.myClass())
+        .addClass(`side-navigation-level-${this.level}`)
```
KateMengjunXia: Where is this class defined?

PR #1661, opened 2018-11-08
Update countryDAO & regionDAO

Henry0422:
Added `"authenticate": false` for both `countryDAO` & `regionDAO`
And removed `AuthenticatedDAO` in `countryDAO`


adamvy:
You should keep the AuthenticatedDAO, we don't want random people updating our list of countries/regions.

AuthenticatedDAO has an option to allow anonyous reads, which is what we want in this case I believe.  Authenticate writes, but allow anonymous read.

Alternatively decorate with a ReadOnlyDAO so no clients can change the contents, authenticated or not.

jlhughes:
ReadOnlyDAO would be fine on the client as new countries and regions always need some code support they are associated with bank accounts, corridors, ... locale. 
But since this is in foam, we could add the readonly in the nanopay service setup.

kgrgreer:
The AuthenticatedDAO's still need to be there otherwise anyone can just connect and update those DAO's. They were already configured property to authenticate writes but not reads. Just remove authentication of the service level, which you did.

kgrgreer:
@jlhughes They need setup for nanopay, but this is generic FOAM/NANOS code, where they don't generally need any extra processing. Just putting back the AuthenticatedDAO that only authenticates writes works for both NP and non-NP users.

Henry0422:
Put `AuthenticatedDAO` back based on Kgr's review. Thank you for reviewing and helping me.

TW80000 (approved):

PR #1662, opened 2018-11-08
Do_not_merge Client login auth service

deejes:
ClientAuthService uses SessionClientBox, which doesn't return an error message on login, and redirects to login screen. It also re-attempts failed logins after a successful login.

These ClientLoginAuthService avoids this problem by using AuthServiceClientBox instead of SessionClientBox.

This code duplicates code already added by @eatonk in np, but the plan is to remove that and point anything that uses it to the classes in this PR.


kirkeaton (changes requested):
You're missing one thing. You need to modify the "auth" service declaration in `foam2/src/services` so that it uses your new ClientLoginAuthService

kirkeaton (approved):

deejes:
@kgrgreer had some comments. Do not merge till discussed and resolved.

PR #1666, opened 2018-11-09
Side Navigatoin Improvements

nanoscott:
Fix unexpected overlapping
Adding padding so that the side bar doesn't overlap the top menu
Adding custom class to each level of the side menu

### DIFF
--- a/src/foam/nanos/u2/navigation/NavigationView.js
+++ b/src/foam/nanos/u2/navigation/NavigationView.js
@@ -7,7 +7,7 @@ foam.CLASS({
 
   css: `
     ^top-nav {
-      width: calc(100% - 0px);
+      width: 100%;
       display: inline;
       z-index: 10001;
       position: fixed;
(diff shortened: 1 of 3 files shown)


src/foam/nanos/u2/navigation/NavigationView.js
```diff
@@ -6,19 +6,19 @@ foam.CLASS({
   documentation: 'Navigation bars',
 
   css: `
-    ^ {
-      position: fixed;
-      z-index: 10001;
-    }
     ^top-nav {
-      width: calc(100% - 200px);
+      width: calc(100% - 1px);
```
nanoscott: I think the width should just be 100%, but ever time I put it at 100% I lose all of the items on the side menu. Using the calc is the only way that I can keep those showing up.

src/foam/nanos/u2/navigation/NavigationView.js
```diff
@@ -7,7 +7,7 @@ foam.CLASS({
 
   css: `
     ^top-nav {
-      width: calc(100% - 1px);
+      width: calc(100% - 0px);
```
kgrgreer: How is this different than 100%?

PR #1667, opened 2018-11-09
Remove ChartConfig and add color properties to bar/pie charts that can be editted from the Card.

mcarcaso:


src/org/chartjs/Bar.js
```diff
@@ -14,13 +14,15 @@ foam.CLASS({
...
       // TODO: Support multiple datasets via nested groupby
+      var colors = this.colors;
       var groups = data.groups;
       var keys = data.sortedKeys();
       var data = {
         labels: keys,
         datasets: [
           {
             label: data.arg2.label || data.arg2.model_.label,
+            backgroundColor: colors,
```
adamvy: I don't think we want to pass the colours here, we don't want each bar to have a different colour, we want each dataset to have a unique colour for when we plot a GROUP_BY(GROUP_BY()).

mcarcaso: good call. Just using a single color for now and when we support nested group bys I'll add more colors.

PR #1673, opened 2018-11-13
F/np 3149/dig authentification

minsunny:
https://github.com/nanoPayinc/NANOPAY/issues/3149

as-is : if POST, it was always set as put command, so wrong error message was shown
to-be : set as given parameter cmd, and then it handles an appropriate message.

Attached a Curl test data below : login by a wrong user then throw "authentication failure."

curl -X POST \
'localhost:8080/service/dig?dao=countryDAO&cmd=put&user=wronguser@nanopay.net&password=wrongpw1' \
-d '{"id":"DIG", "code":"DIG", "name":"DIG"}' \
-H 'accept: application/json' \
-H 'cache-control: no-cache' \
-H 'content-type: application/json'


jlhughes:
so do we still support HTTP PUT, POST, DELETE? Not just our commands, but HTTP operations.

minsunny:
> so do we still support HTTP PUT, POST, DELETE? Not just our commands, but HTTP operations.

Well.. They (HTTP PUT, POST, DELETE) are still supported, but Even HTTP PUT can access all the commands on DIG not only the command PUT.

jlhughes:
Where is POST converted to 'put' now? does it occur elsewhere? Also DELETE. what does it become?

src/foam/nanos/http/HttpParametersWebAgent.java
```diff
@@ -128,71 +128,72 @@ public void execute(X x) {
...
             logger.debug("id", req.getParameter("id"));
-          }
-          break;
-
-        case "remove":
-          command = Command.remove;
-          parameters.set("id", req.getParameter("id"));
-          logger.debug("id", req.getParameter("id"));
-          break;
+            break;
```
jlhughes: So at this point it looks like it could be reduced further as the flow for cmd is the same for both branches of 
    if ( ! "application/x-www-form-urlencoded".equals(contentType) ) {

jlhughes (approved):

PR #1676, opened 2018-11-13
Refactor Email service interface

sirenChen:
refs:
[4370](https://github.com/nanoPayinc/NANOPAY/issues/4370)
[4474](https://github.com/nanoPayinc/NANOPAY/issues/4474)

- [x] Refactor the EmailService interface. We need pass the user context into the email service so we can get the correct Domain URL from http request.

- [x] Make **emailFrom**, **emailReplyTo**, **emailDisplayName** fully configurable.
    i.  First, look for those value in the EmailMessage.
    ii.  If nothing found in the EmailMessage, look for the value in the group setting.
    iii. If nothing found in the group setting, use the default value. 

- [x] Make Subject of Email Template configurable
    i. If the subject doesn't provided when we creating the EmailMessage, use the subject in the template.
    ii. You can now use template arguments in both template subject and body.

### DIFF
--- a/src/foam/nanos/notification/email/DAOEmailService.js
+++ b/src/foam/nanos/notification/email/DAOEmailService.js
@@ -91,23 +91,11 @@ return config_;`
     {
       name: 'sendEmail',
       javaCode: `
-        getDao().put(emailMessage);
-`
-    },
-    {
-      name: 'sendEmailWithX',
-      javaCode: `
         getDao().put_(x, emailMessage);
 `
     },
     {
       name: 'sendEmailFromTemplate',
-      javaCode: `
-sendEmailFromTemplateWithX(getX(), user,  emailMessage, name, templateArgs);
-`
-    },
-    {
-      name: 'sendEmailFromTemplateWithX',
       javaCode: `
 String group = user != null ? (String) user.getGroup() : null;
 EmailTemplate emailTemplate = DAOResourceLoader.findTemplate(getX(), name, group);
@@ -135,7 +123,7 @@ if (foam.util.SafetyUtil.isEmpty(emailMessage.getSubject())) {
   emailMessage.setSubject(templateSubject.render(model));
 }
 
-sendEmailWithX(x, emailMessage);
+sendEmail(x, emailMessage);
 `
     }
   ]
--- a/src/foam/nanos/notification/email/DAOEmailService.js
+++ b/src/foam/nanos/notification/email/DAOEmailService.js
@@ -95,7 +95,19 @@ return config_;`
 `
     },
     {
+      name: 'sendEmailWithX',
+      javaCode: `
+        getDao().put_(x, emailMessage);
+`
+    },
+    {
       name: 'sendEmailFromTemplate',
+      javaCode: `
+sendEmailFromTemplateWithX(getX(), user,  emailMessage, name, templateArgs);
+`
+    },
+    {
+      name: 'sendEmailFromTemplateWithX',
       javaCode: `
 String group = user != null ? (String) user.getGroup() : null;
 EmailTemplate emailTemplate = DAOResourceLoader.findTemplate(getX(), name, group);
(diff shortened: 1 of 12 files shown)


src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -293,6 +316,57 @@ if ( getAuthenticate() ) {
...
+Group group         = (Group) groupDAO.find(user.getGroup());
+AppConfig appConfig = group.getAppConfig(x);
+
+String domain = appConfig.getUrl().contains("ablii") ? "ablii.com" : "nanopay.net";
+
+if ( SafetyUtil.isEmpty(emailMessage.getFrom()) ) {
+  emailMessage.setFrom(
+    ! SafetyUtil.isEmpty(group.getFrom()) ?
+      group.getFrom() :
+      getFrom().substring(0, getFrom().indexOf("@") + 1) + domain
```
jlhughes: Why would the from not have the correct domain?

src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -293,6 +316,57 @@ if ( getAuthenticate() ) {
...
+foam.nanos.session.Session session = x.get(foam.nanos.session.Session.class);
+
+DAO userDAO         = (DAO) x.get("localUserDAO");
+DAO groupDAO        = (DAO) x.get("groupDAO");
+
+User user           = (User) userDAO.find(session.getUserId());
+Group group         = (Group) groupDAO.find(user.getGroup());
+AppConfig appConfig = group.getAppConfig(x);
+
+String domain = appConfig.getUrl().contains("ablii") ? "ablii.com" : "nanopay.net";
```
jlhughes: Why are explicitly testing for ablii?

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,11 +92,23 @@ return config_;`
       name: 'sendEmail',
       javaCode: `
         getDao().put(emailMessage);
+`
+    },
+    {
+      name: 'sendEmailWithX',
```
jlhughes: Why the double calls? Why not just fix everything? We have to update all calls to pass the context, that’s the entire point, is it not?

src/foam/nanos/notification/email/SMTPEmailService.js
sirenChen: Just an easier way to get the domain from URL. If URL contains "ablii", then the domain is "ablii", otherwise the domain is "nanopay.net".

src/foam/nanos/notification/email/SMTPEmailService.js
sirenChen: If _emailMessage.getFrom()_ and _group.getFrom()_ are both empty, we will have to return _SMTPEmailService.getFrom()_.

The getFrom() in the SMTPEmailService will always return the default value which defined in the services journal file(info@nanopay in this case). For example, if the request is from ablii, we have to replace the domain name with ablii.com

Another option is we can set the domain at SessionServerBox (Just like appConfig), then we will always have correct domain for each request.

src/foam/nanos/notification/email/DAOEmailService.js
sirenChen: Only for the backward compatibility. We have about 10~15 places are calling this method. I think it's better to change them progressively.

jlhughes: Please force the change. Add context to the original method.

src/foam/nanos/notification/email/SMTPEmailService.js
jlhughes: Yes, but now we need a test for every url we may deploy under, this does not scale.
Consider this quote from a recent email:  `...  we use a different domain to send Marketing emails, so we don't affect our domain rating. For example, we should be using hello@e.ablii.com or hello@email.ablii.com, rather than ablii.com. Same thing goes for nanopay.net.`

Since the outbound domain may not match the request url, simply use, in order:

1. email parameters
2. group parameters
3. smtpemail parameters

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -91,7 +91,7 @@ return config_;`
     {
       name: 'sendEmail',
       javaCode: `
-        getDao().put(emailMessage);
+        getDao().put_(x, emailMessage);
```
jlhughes: Change this to getDao().inX(x).put(emailMessage);

jlhughes (approved):

PR #1678, opened 2018-11-14
Add isDescendantOf method to Group model

TW80000:
@KateMengjunXia


jlhughes:
This seems like something to add in Relationships.

TW80000:
@jlhughes I don't see the connection between this and relationships. What do you mean?

jlhughes:
@TW80000 groups have parents setup by a Relationship, so this feature of walking the parent tree seems like a generic feature of Relationships.

PR #1679, opened 2018-11-14
Fix Ablii forgot password screen redirect to Nanopay sign in view

sirenChen:
refs:
[4435 ](https://github.com/nanoPayinc/NANOPAY/issues/4435)

Please merge this and [4533](https://github.com/nanoPayinc/NANOPAY/pull/4533) at the same time

### DIFF
--- a/src/foam/nanos/auth/resetPassword/EmailView.js
+++ b/src/foam/nanos/auth/resetPassword/EmailView.js
@@ -154,7 +154,7 @@ foam.CLASS({
         .start('p').addClass('link')
           .add('Sign in.')
           .on('click', function() {
-            self.stack.push({ class: 'foam.nanos.auth.SignInView' });
+            self.stack.push({ class: 'net.nanopay.sme.ui.SignInView' });
           })
         .end()
       .end();
--- a/src/foam/nanos/auth/resetPassword/EmailView.js
+++ b/src/foam/nanos/auth/resetPassword/EmailView.js
@@ -129,6 +129,11 @@ foam.CLASS({
     {
       class: 'EMail',
       name: 'email'
+    },
+    {
+      class: 'String',
+      name: 'signInView',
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/resetPassword/EmailView.js
```diff
@@ -154,7 +154,7 @@ foam.CLASS({
         .start('p').addClass('link')
           .add('Sign in.')
           .on('click', function() {
-            self.stack.push({ class: 'foam.nanos.auth.SignInView' });
+            self.stack.push({ class: 'net.nanopay.sme.ui.SignInView' });
```
TW80000: This would be putting a reference to a NANOPAY-specific view into the foam2 repo, which is supposed to be a generic framework if I'm not mistaken. I don't think this makes sense to do.

sirenChen: So another option would be create another forgot password view under nanopay.sme.ui

TW80000: Yeah. A third option would be to modify the foam EmailView to take a parameter for the sign in view it pushes onto the stack here that defaults to the foam one but can be set to a custom one if needed. I think that way avoids duplicating code.

src/foam/nanos/auth/resetPassword/EmailView.js
```diff
@@ -129,6 +129,11 @@ foam.CLASS({
     {
       class: 'EMail',
       name: 'email'
+    },
+    {
+      class: 'String',
```
TW80000, line 134: This works, but usually when we want a view as a property I've seen a `foam.u2.ViewSpec` used. I think it's more flexible that way.

[Example](https://github.com/foam-framework/foam2/blob/master/src/foam/comics/DAOControllerView.js#L98)

PR #1680, opened 2018-11-14
Invalid Inputs

kennyqiyenkan:
Purpose:
Adds an error flavour text below inputs that are invalid. The message will be taken from the returned value of `validateObj` in a property.

Changes:
Changed DetailPropertyView to include the above feature.
Changed TextField to be onKey = true by default.


adamvy:
I believe this PR is deprecated now?  Error message support was added elsewhere already I believe.  @mcarcaso correct?

Can we close this PR?

jlhughes:
Superseded

PR #1693, opened 2018-11-16
Make SendEmailNotificationDAO more configurable

TW80000:
I want to be able to change the user DAO used in SendEmailNotificationDAO in nanopay because we want to use `bareUserDAO` instead of `localUserDAO` so you can send emails to Businesses, not just
Users.


TW80000:
Out of date

PR #1697, opened 2018-11-17
Dashboard and Card styling

blakegreen32:
Added spacing between dashboard cards, setup correct header styles and naming, setup consistent height for cards

### DIFF
--- a/src/foam/dashboard/view/Card.js
+++ b/src/foam/dashboard/view/Card.js
@@ -58,14 +58,19 @@ foam.CLASS({
       border-radius: 2px;
       background: white;
       margin: 8px;
+      padding-bottom: 10px;
     }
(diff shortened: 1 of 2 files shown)


src/foam/dashboard/view/Card.js
```diff
@@ -53,20 +53,26 @@ foam.CLASS({
...
-}
+    ^ {
+      border: 2px solid #dae1e9;
+      border-radius: 2px;
+      background: white;
+      margin: 8px;
+      padding-bottom: 10px;
+    }
 
-^header {
```
adamvy, line 62: Update the forumla in the contentHeight/contentWidth expression to match how much width/height is left for the graph after your CSS changes.

PR #1698, opened 2018-11-18
tree view default selection as first element

shreymahendru:


src/foam/u2/view/TreeView.js
```diff
@@ -251,21 +251,27 @@ foam.CLASS({
 
   methods: [
     function initE() {
-      var M   = this.ExpressionsSingleton.create();
-      var of  = this.lookup(this.relationship.sourceModel);
+      var M = this.ExpressionsSingleton.create();
```
adamvy: I prefer the code before where the values were all lined up nicely in columns :)

src/foam/u2/view/TreeView.js
```diff
@@ -251,21 +251,27 @@ foam.CLASS({
...
-        select(dao, function(obj) {
-          return self.TreeViewRow.create({
-            data: obj,
-            relationship: self.relationship,
-            expanded: self.startExpanded,
-            formatter: self.formatter
-          }, this);
-        });
+      var element = this.addClass(this.myClass());
+      dao.select().then((sink) => {
```
adamvy: Doing dao.select().then is dramatically different from doing this.select().

In a U2 Element, this.select(dao, function(){}) hooks up dynamically so that it re-renders when the dao changes.

I think what you could do instead and would be a much smaller change would be

```javascript
var first = true;
this.select(dao, function(obj) {
  if ( first && ! self.selection ) self.selection = obj;
  first = false;
  ...
```

That's a much smaller diff and therefore less risk.

src/foam/u2/view/TreeView.js
```diff
@@ -108,7 +108,7 @@ foam.CLASS({
         }, this.selection$, this.data$.dot('id'))).
         start('span').
           style({
-            visibility: this.hasChildren$.map(function(c) { return c ? 'visible' : 'hidden'; }),
+            'visibility': this.hasChildren$.map(function(c) { return c ? 'visible' : 'hidden'; }),
```
kgrgreer: Why the quotes? This is JS, not JSON.

src/foam/u2/view/TreeView.js
kgrgreer: Why mess up the spacing?

src/foam/u2/view/TreeView.js
kgrgreer: Agreed.

PR #1700, opened 2018-11-21
Update auth service to work with user subclasses

TW80000:
Now that we're starting to support acting as a Business in nanopay, some bugs came up where the DAOs used in FOAM are incorrect because we need it to be able to refer to the `bareUserDAO` in nanopay, which contains all instances of User and its subclasses. However, since FOAM doesn't have any subclasses of User, I thought it made more sense for us to subclass the auth service in nanopay and override the start method where we can set this new class-level DAO variable called "userSubclassDAO" to the nanopay `bareUserDAO` rather than adding `bareUserDAO` to FOAM directly.

Required by https://github.com/nanoPayinc/NANOPAY/pull/4784

### DIFF
--- a/src/foam/nanos/auth/UserAndGroupAuthService.java
+++ b/src/foam/nanos/auth/UserAndGroupAuthService.java
@@ -31,6 +31,7 @@ public class UserAndGroupAuthService
   protected DAO userDAO_;
   protected DAO groupDAO_;
   protected DAO sessionDAO_;
+  protected DAO userSubclassDAO_;
 
   public final static String CHECK_USER_PERMISSION = "service.auth.checkUser";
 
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -141,21 +147,18 @@ public User login(X x, long userId, String password) throws AuthenticationExcept
...
-    User user = (User) userDAO_.find(userId);
+    User user = (User) userSubclassDAO_.find(userId);
     User contextUser = (User) userAndGroupContext(x, user, password);
     return contextUser;
   }
 
   public User loginByEmail(X x, String email, String password) throws AuthenticationException {
-    Sink sink = new ArraySink();
-    sink = userDAO_.where(MLang.EQ(User.EMAIL, email.toLowerCase())).limit(1).select(sink);
+    User user = (User) userDAO_.find(MLang.EQ(User.EMAIL, email.toLowerCase()));
```
daudich, line 165: Why is this also not `userSubclassDAO_`?

daudich (commented):
Should the `userDAO_` in `updatePassword` be also changed to `userSubclassDAO_`?

src/foam/nanos/auth/UserAndGroupAuthService.java
TW80000: Because subclasses of User can have the same email address as a User, but we don't want to find the subclass, we want to find the User.

TW80000:
@daudich Not for us, because (nanopay's) subclasses of User don't have passwords and you can't log in as them directly. That's not necessarily true for other foam2-based applications though I guess.

PR #1702, opened 2018-11-22
duplicate CICO for all accounts view

KateMengjunXia:


KateMengjunXia:
Added to wrong class.

PR #1703, opened 2018-11-22
Dig post request

nanoNeel:
Adds POST requests to DIG
REFS: [here](https://github.com/nanoPayinc/NANOPAY/issues/4041)

### DIFF
--- a/src/foam/nanos/dig/DIG.js
+++ b/src/foam/nanos/dig/DIG.js
@@ -11,7 +11,9 @@ foam.CLASS({
 
   documentation: 'Data Integration Gateway - Perform DAO operations against a web service',
 
-  requires: ['foam.net.web.HTTPRequest'], 
+  requires: ['foam.net.web.HTTPRequest'],
+
+  imports: ['appConfig'],
 
   tableColumns: [
     'id',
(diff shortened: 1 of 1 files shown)


src/foam/nanos/dig/DIG.js
```diff
@@ -122,12 +127,43 @@ foam.CLASS({
...
   methods: [
+  ],
+
+  actions: [
+    {
+      name: 'postButton',
+      label: 'Send Request',
+      code: function() {
+        var req = this.HTTPRequest.create({
+          url: window.location.origin + this.digURL,
```
jlhughes: replace with user.getGroup(x).getAppConfig(x).getUrl()

src/foam/nanos/dig/DIG.js
```diff
@@ -148,7 +150,7 @@ foam.CLASS({
       label: 'Send Request',
       code: function() {
         var req = this.HTTPRequest.create({
-          url: window.location.origin + this.digURL,
+          url: this.appConfig.URL.value + this.digURL.substring(1),
```
jlhughes, line 153: This is incorrect. You need to get the user, get the group and get the appConfig from the group. 
@ZacZhang and @sirenChen can provide guidance.

jlhughes (changes requested):

src/foam/nanos/dig/DIG.js
jlhughes: AppConfig appears to be setup correctly on the client.

jlhughes (approved):

PR #1705, opened 2018-11-23
Update email notification

Henry0422:
Moved `userDAO_ = (DAO) x.get("localUserDAO");` into the constructor.


kgrgreer:
What behaviour does this change or issue does it fix?

zahnno:
It let's us place whatever userDAO we'd like to use in the emailnotification service by extending the sendNotificationDAO.

Henry0422:
It is out of date since there is a code refactoring.

PR #1711, opened 2018-11-27
get the most updated agent

shreymahendru:
this is a fix for when the agent is updated but when calling the getCurrentAgent gives you the stale user object

### DIFF
--- a/src/foam/nanos/auth/AgentUserAuthService.java
+++ b/src/foam/nanos/auth/AgentUserAuthService.java
@@ -36,7 +36,7 @@ public class AgentUserAuthService
     agentJunctionDAO_  = (DAO) getX().get("agentJunctionDAO");
   }
 
-  /* 
+  /*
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -114,7 +113,9 @@ public User getCurrentAgent(X x) throws AuthorizationException {
     X sessionContext = session.getContext();
     // Get agent from session context
     User agent = (User) sessionContext.get("agent");
-
+    if (agent != null) {
```
kgrgreer: spacing:
if ( agent != null ) {

kgrgreer:
So this happens when you edit an agent while they're already logged in?

shreymahendru:
> So this happens when you edit an agent while they're already logged in?

Yes, primarily when you update the agent and do a refresh, You get the stale agent.

PR #1714, opened 2018-11-27
Proactively load permissions

nanoscott:
Lazy loading of default permissions was causing some records to be hidden during initialization, so changing the loading to to proactive instead of waiting for first use.


daudich:
LGTM

kgrgreer:
But we should keep in mind that this is side-stepping but not fixing some
underlying bug.

PR #1718, opened 2018-11-29
Polymorphic authorization

TW80000:
This is a DAO that lets models implement their own methods for custom
authorization logic and runs those methods when appropriate.

It's a polymorphic approach to authorization rather than writing many
custom authorization decorators that have to worry about the class of
the objects they're trying to check.

On top of that, it moves authorization code onto the model and removes
the need for a lot of extra files, namely the custom authorization DAO
decoraters we have.

----

This PR also puts it to use by moving the relevant logic out of AuthenticatedUserDAO into the User model (which now implement Authorizable) and replacing the AuthenticatedUserDAO on `userDAO` with a generic `AuthorizationDAO` to call those methods.


src/foam/java/Authorizable.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.INTERFACE({
```
adamvy, line 7: Why not just put the real interface methods here instead of a "dummy" interface that is only defined in java?

TW80000: I was following the pattern of what we do for the `Validatable` interface. I'll put them in.

TW80000: Hmm, I removed Authorizable.java and updated Authorizable.js to this:

```JavaScript
/**
 * @license
 */

foam.INTERFACE({
  package: 'foam.nanos.auth',
  name: 'Authorizable',

  documentation: `
    A model should implement this interface if it is authorizable, meaning some
    users are allowed to operate on (create, read, update, or delete) that
    object but others are not.
  `,

  methods: [
    {
      name: 'authorizeOnCreate',
      javaReturns: 'void',
      args: [
        { name: 'x', javaType: 'foam.core.X' }
      ],
      javaReturns: 'void',
      javaThrows: ['AuthorizationException'],
    },
    {
      name: 'authorizeOnRead',
      javaReturns: 'void',
      args: [
        { name: 'x', javaType: 'foam.core.X' }
      ],
      javaReturns: 'void',
      javaThrows: ['AuthorizationException'],
    },
    {
      name: 'authorizeOnUpdate',
      javaReturns: 'void',
      args: [
        { name: 'x', javaType: 'foam.core.X' },
        { name: 'oldObj', javaType: 'foam.core.FObject' }
      ],
      javaReturns: 'void',
      javaThrows: ['AuthorizationException'],
    },
    {
      name: 'authorizeOnDelete',
      javaReturns: 'void',
      args: [
        { name: 'x', javaType: 'foam.core.X' }
      ],
      javaReturns: 'void',
      javaThrows: ['AuthorizationException'],
    }
  ]
});
```

but it doesn't compile this way because the Java code for the interface never gets generated.

mcarcaso: Interfaces don't need java code, they're just method definitions. Can you paste what's generated?
Also you have javaReturns twice in each method but that shouldn't be affecting anything.

mcarcaso: Oh you mean the entire file isn't generated? That might be because you added it to the blacklist in genjava2.js :p

TW80000: @mcarcaso Thanks, removing from the blacklists fixed it. I didn't realize that's what they were, my bad.

mcarcaso: All good. You were just following an example that happened to be wrong. We should fix up Validatable to be the same  but we can do that in a separate PR.

PR #1719, opened 2018-11-29
Dig post request

nanoNeel:
- Adds better way to display results for DIG
- Adds POST request to SUGAR
- Adds JSONJ support
- REFS: [POST request](https://github.com/nanoPayinc/NANOPAY/issues/4041), [JSONJ](https://github.com/nanoPayinc/NANOPAY/issues/4861)

### DIFF
--- a/src/foam/nanos/dig/DigWebAgent.java
+++ b/src/foam/nanos/dig/DigWebAgent.java
@@ -510,6 +510,17 @@ public class DigWebAgent
       outputterHtml.outputEndTable();
       outputterHtml.outputEndHtml();
       out.println(outputterHtml.toString());
+    }else if ( format == Format.JSONJ ) {
+      //output error in jsonJ format
+
+      JSONParser jsonParser = new JSONParser();
+      jsonParser.setX(x);
+      foam.lib.json.Outputter outputterJson = new foam.lib.json.Outputter(OutputterMode.STORAGE);
+      outputterJson.setOutputDefaultValues(true);
+      outputterJson.setOutputClassNames(true);
+      outputterJson.outputJSONJFObject(error);
+      out.println(outputterJson.toString());
+
     } else {
       // TODO
     }
--- a/src/foam/nanos/dig/DigWebAgent.java
+++ b/src/foam/nanos/dig/DigWebAgent.java
@@ -213,6 +213,66 @@ public class DigWebAgent
           outputException(x, resp, format, out, error);
 
           return;
+        } else if (Format.JSONJ == format ) {
+          String dataJson = "[";
+          String dataJsonJ[] = data.split("\\s*\\r?\\n");
+          for (String i:dataJsonJ){
+            if (i.startsWith("p(") && i.endsWith(")")) {
+              dataJson += i.substring(2, i.length()-1) + ',';
+            }
+          }
+          dataJson += "]";
+         
+        // JSON part from above
+        JSONParser jsonParser = new JSONParser();
+          jsonParser.setX(x);
+          foam.lib.json.Outputter outputterJson = new foam.lib.json.Outputter(OutputterMode.NETWORK);
+          outputterJson.setOutputDefaultValues(true);
+          outputterJson.setOutputClassNames(true);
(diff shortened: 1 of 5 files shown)


src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -450,6 +510,17 @@ protected void outputException(X x, HttpServletResponse resp, Format format, Pri
       outputterHtml.outputEndTable();
       outputterHtml.outputEndHtml();
       out.println(outputterHtml.toString());
+    }else if ( format == Format.JSONJ ) {
```
jlhughes: space between } and 'else'

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -213,6 +213,66 @@ public void execute(X x) {
...
+          String dataJsonJ[] = data.split("\\s*\\r?\\n");
+          for (String i:dataJsonJ){
+            if (i.startsWith("p(") && i.endsWith(")")) {
+              dataJson += i.substring(2, i.length()-1) + ',';
+            }
+          }
+          dataJson += "]";
+         
+        // JSON part from above
+        JSONParser jsonParser = new JSONParser();
```
jlhughes: indent

src/foam/nanos/dig/DigWebAgent.java
jlhughes: hard to tell from this PR, what is the difference between this JSONJ block and that above?

src/foam/nanos/dig/SUGAR.js
```diff
@@ -231,27 +255,70 @@ foam.CLASS({
...
+      hidden: true,
+      label: 'Send POST Request',
+      code: async function() {
+        if ( ! (this.postURL === '') ) {
+          var req = this.HTTPRequest.create({
+            url: this.appConfig.URL.value + this.postURL.substring(1),
+            method: 'POST',
+            contentType: 'url',
+            payload: this.postData.substring(1),
+          }).send();
```
jlhughes, line 304: Doesn't have to happen in this PR (can be future ticket), but we need to add http authentication info with the sessionid so these posts can proceed with an additional login.

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -213,6 +213,66 @@ public void execute(X x) {
           outputException(x, resp, format, out, error);
 
           return;
+        } else if (Format.JSONJ == format ) {
+          String dataJson = "[";
+          String dataJsonJ[] = data.split("\\s*\\r?\\n");
+          for (String i:dataJsonJ){
+            if (i.startsWith("p(") && i.endsWith(")")) {
```
jlhughes: Also, you shouldn't need to do this. Have a look in the 'replay' of the JDAO as it has logic to apply the json/j operations, unless of course it does the same thing.

src/foam/nanos/dig/DigWebAgent.java
nanoNeel: This JSONJ block is for the outputException method which is called when ever there is an error. Without this block the webpage won't load and you would have no idea if it failed or not. The JSONJ block above is for the put command.

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -213,6 +213,65 @@ public void execute(X x) {
           outputException(x, resp, format, out, error);
 
           return;
+        } else if (Format.JSONJ == format ) {
+          String dataJson = "[";
+          String dataJsonJ[] = data.split("\\r?\\n");
+          for (String i:dataJsonJ){
+            i = i.trim();
+            dataJson += i.substring(2, i.length()-1) + ',';
```
jlhughes: Ok, i preferred the test for 'p'
`         if (i.startsWith("p(") && i.endsWith(")")) {`
as that ensures we only put 'p' operations, the blind substring can let through 'r' operations on command 'put'.  
Is this really what the JDAO does?

jlhughes (approved):

PR #1724, opened 2018-12-01
DAOControllerView EmptyStates

blakegreen32:
Setup empty states for DAOControllerViews


zahnno (approved):
LG

TW80000 (changes requested):
I have one problem with this implementation. `isEmptyStateShowing` defaults to true if the controller provides an empty state view. It doesn't get set to false until the count query on the DAO finishes, which is asynchronous. Because of that along with the fact that we're always showing either the normal controller view or the empty state view, we get a flicker where when you first go to the page you see the empty state for a split second and then it changes to the normal view when the query finishes and it finds out the DAO isn't empty.

I'd rather we have a third state where the query is still running during which nothing is shown to avoid the flicker. Maybe add a property like `loading :: Boolean` that defaults to true and gets set to false after the count query finishes. Then `isEmptyStateShowing` could also depend on `loading` being false and the entire view code would be hidden if loading were true.

TW80000 (approved):

src/foam/comics/DAOControllerView.js
```diff
@@ -220,6 +224,18 @@ foam.CLASS({
       } else {
         this.onEdit(null, null, obj.id);
       }
+    },
+
+    function fetchDaoCount() {
```
kgrgreer, line 229: rename Dao to DAO

src/foam/comics/DAOControllerView.js
```diff
@@ -220,6 +224,18 @@ foam.CLASS({
...
+      this.data.data.select(this.COUNT()).then((count) => {
+        if ( count.value === 0 ) {
```
kgrgreer, line 231: This would be clearer:

if ( count.value === 0 && this.data.emptyStateView ) {
  this.start().tag(this.data.emptyStateView).end();
} else {
  this.start().add(this.data.border).end();
}

PR #1729, opened 2018-12-05
Update child group email settings

sirenChen:
- [x] When you update the email settings, all the child groups settings will update as well.
- [x] When the child group email settings is empty, get the values from the parent group.
- [x] If the child group doesn't want to inherit those values, you can also set the value for child group individually.


zahnno (commented):
It'd be nice to see a select sink for this but otherwise looks good to me

src/foam/nanos/auth/UpdateChildEmailSetting.java
```diff
@@ -0,0 +1,115 @@
...
+  @Override
+  public FObject put_(X x, FObject obj) {
+    Group group    = (Group) obj;
+    Group oldGroup = (Group) this.getDelegate().find_(x, group.getId());
+
+    if ( oldGroup == null ) {
+      return super.put_(x, obj);
+    }
+
+    if ( ! SafetyUtil.equals(group.getDisplayName(), oldGroup.getDisplayName()) ) {
```
jlhughes, line 29: I would rather only update the child if the child value is set. 
This matches the behaviour of looking up the parent value when the child is not set.

src/foam/nanos/auth/UpdateChildEmailSetting.java
```diff
@@ -0,0 +1,115 @@
...
+
+  }
+
+  public void getParentValue(X x, Group childGroup, Group parentGroup) {
+
+    if ( parentGroup == null ) {
+      return;
+    }
+
+    if ( ! SafetyUtil.isEmpty(parentGroup.getFrom()) ||
```
jlhughes, line 97: These should be independent.   Three tests. 
This allows configuration where the parent is/was never set and child is.

src/foam/nanos/auth/UpdateChildEmailSetting.java
kgrgreer, line 29: Does this design result in the same record possibly being updated three times, ones for each field?

src/foam/nanos/auth/UpdateChildEmailSetting.java
```diff
@@ -0,0 +1,115 @@
...
+import foam.dao.ArraySink;
+import foam.dao.DAO;
+import foam.dao.ProxyDAO;
+import foam.mlang.MLang;
+import foam.util.SafetyUtil;
+
+import java.util.ArrayList;
+import java.util.function.BiConsumer;
+
+public class UpdateChildEmailSetting extends ProxyDAO {
```
kgrgreer, line 14: Document what this class is for / does.

src/foam/nanos/auth/UpdateChildEmailSetting.java
```diff
@@ -0,0 +1,115 @@
...
+      return super.find_(x, id);
+    }
+
+    return group;
+  }
+
+  public void updateChildEmailSetting(X x, Group group, BiConsumer<Group, Group> update) {
+
+    ArraySink sink = (ArraySink) this.getDelegate().where(MLang.EQ(
+      Group.PARENT, group.getId()
```
kgrgreer, line 70: Easier to just process directly in select() sink than to copy into an ArrayList then iterate.

jlhughes:
Abandoned.

PR #1730, opened 2018-12-06
Country and Region enhancement

sirenChen:
Lookup country in order:

1. 2 digit ISO code
2. 3 digit ISO code
3. Name
4. Cache
5. Alternative Name


src/countries
```diff
@@ -1,13 +1,13 @@
...
-p({"class":"foam.nanos.auth.Country", "code":"JM", "name":"Jamaica"})
-p({"class":"foam.nanos.auth.Country", "code":"LB", "name":"Lebanon"})
-p({"class":"foam.nanos.auth.Country", "code":"MX", "name":"Mexico"})
-p({"class":"foam.nanos.auth.Country", "code":"MY", "name":"Malaysia"})
-p({"class":"foam.nanos.auth.Country", "code":"RS", "name":"Serbia"})
-p({"class":"foam.nanos.auth.Country", "code":"TT", "name":"Trinidad and Tobago"})
-p({"class":"foam.nanos.auth.Country", "code":"UK", "name":"United Kingdom"})
-p({"class":"foam.nanos.auth.Country", "code":"US", "name":"USA"})
-p({"class":"foam.nanos.auth.Country", "code":"ZA", "name":"South Africa"})
+p({"class":"foam.nanos.auth.Country", "code":"BR", "ISOCode":"BRA","name":"Brazil"})
```
jlhughes: for ISOCode - replace with iso### whatever the number is.

src/foam/nanos/auth/CountryAndRegionCacheService.java
```diff
@@ -0,0 +1,163 @@
...
+import foam.nanos.notification.email.EmailService;
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+/**
+ * Query the country and region by code, name, or alternative names.
+ */
+public class CountryAndRegionCacheService extends ContextAwareSupport implements NanoService {
```
jlhughes: Can you split these into two classes.

src/foam/nanos/auth/Country.js
```diff
@@ -18,10 +18,19 @@ foam.CLASS({
...
+      documentation: '3 digit ISO country code'
+    },
     {
       class: 'String',
       name: 'name',
       documentation: 'Country name.'
+    },
+    {
+      class: 'StringArray',
+      name: 'alternativeName'
```
kgrgreer: Since this is an array, shouldn't it be called 'alternativeNames' (plural)?

src/foam/nanos/auth/CountryAndRegionCacheService.java
```diff
@@ -0,0 +1,163 @@
...
+
+  private DAO countryDAO;
```
kgrgreer: protected instead of private makes easier to test.

src/foam/nanos/auth/CountryAndRegionCacheService.java
kgrgreer: Yes, seems to be two entirely different things: email notifications and cachine.

Also, why is caching needed? Data is already stored in memory, so how does this help?

jlhughes: @kgrgreer I was talking about the split of Country and Region.

jlhughes: re: caching. we can use IN to test the alternates. 
IN(country.getAlternatives())
This can replace all the caching code.

sirenChen: I wasn't aware of the MLang.IN before.

src/foam/nanos/auth/Country.js
sirenChen: Yes, it should be called 'alternativeNames' !

src/countries
sirenChen: Change to ISO_3166_1

src/foam/nanos/auth/Country.js
```diff
@@ -16,12 +16,21 @@ foam.CLASS({
     {
       class: 'String',
       name: 'code',
-      documentation: 'Country code.'
+      documentation: 'ISO 3166-1 alpha-2 Country codes.'
+    },
+    {
+      class: 'String',
+      name: 'ISO_3166_1',
```
kgrgreer: This is going to make for odd method names. Can we rename to iso31661Code or iso3166Code?

src/foam/nanos/auth/RegionService.java
```diff
@@ -0,0 +1,80 @@
...
+import foam.nanos.notification.email.EmailMessage;
+import foam.nanos.notification.email.EmailService;
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+/**
+ * Query the region by code, name, or alternative names.
+ */
+public class RegionService extends ContextAwareSupport implements NanoService {
```
kgrgreer, line 18: Separate email from look. Lookup can just be a static method rather than a service since its stateless.

src/foam/nanos/auth/CountryService.java
```diff
@@ -0,0 +1,86 @@
...
+import foam.nanos.notification.email.EmailMessage;
+import foam.nanos.notification.email.EmailService;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Query the country by code, name, or alternative names.
+ */
+public class CountryService extends ContextAwareSupport implements NanoService {
```
kgrgreer, line 18: Separate email from look. Lookup can just be a static method rather than a service since its stateless.

src/foam/nanos/auth/Country.js
sirenChen: Rename to iso31661Code

PR #1731, opened 2018-12-07
Scroll table and top nav fixes

TW80000:
* Fix bug where you can't see last item in table
* Make ScrollTableView row height configurable
  * We need this in Ablii because table rows are 48px tall instead of 40px
* Fix top nav not showing when already logged in (this was a mistake introduced in my old PR that was merged yesterday) 
* Added option to ScrollTableView to make the height dynamic so that it won't go off the bottom of the page

Should fix https://github.com/nanoPayinc/NANOPAY/issues/5158

### DIFF
--- a/src/foam/u2/view/ScrollTableView.js
+++ b/src/foam/u2/view/ScrollTableView.js
@@ -16,6 +16,14 @@
     'foam.u2.view.TableView'
   ],
 
+  constants: [
+    {
+      type: 'Int',
+      name: 'TABLE_HEAD_HEIGHT',
+      value: 40
+    }
+  ],
+
   properties: [
     {
       class: 'foam.dao.DAOProperty',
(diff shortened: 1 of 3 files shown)


src/foam/u2/view/ScrollTableView.js
```diff
@@ -78,20 +107,24 @@
...
+                height: this.rowHeight * limit + this.TABLE_HEAD_HEIGHT,
+                width: 18,
+                size$: this.daoCount$,
+              });
             })).
           end().
         end().
       end();
+
+      if ( this.fitInScreen ) this.onload.sub(this.updateTableHeight);
```
kgrgreer, line 127: Should you also listen for resize events? If so, do in a future PR.

src/foam/u2/view/ScrollTableView.js
```diff
@@ -16,6 +16,14 @@
     'foam.u2.view.TableView'
   ],
 
+  constants: [
+    {
+      type: 'Int',
+      name: 'TABLE_HEAD_HEIGHT',
+      value: 40
```
kgrgreer, line 23: Could the view figure this out on its own?

src/foam/u2/view/ScrollTableView.js
TW80000: Made a ticket for this here: https://github.com/foam-framework/foam2/issues/1736

src/foam/u2/view/ScrollTableView.js
TW80000: Included in https://github.com/foam-framework/foam2/issues/1736

PR #1734, opened 2018-12-07
Make User.id final as changing it does not change the account, but ra…

jlhughes:
…ther creates another (as expected).

Issue: [5279](https://github.com/nanoPayinc/NANOPAY/issues/5279)

NOTE: this does not affect the UI, it appears that it can be edited, but is ignored on put.  Another ticket will cover adding UI support for 'final'. 

NP-5279 #time 30m

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -53,6 +53,7 @@ foam.CLASS({
     {
       class: 'Long',
       name: 'id',
+      final: true,
       tableWidth: 45
     },
     {


TW80000:
What does marking a property as `final` do? Is `final` already supported in FOAM?

jlhughes:
@TW80000 'final' - once set cannot be updated/changed, and yes had been supported for some time.

kgrgreer:
All primary keys should be final, not just this one. This should be enforced in the DAOController. We should also add a "Copy" action since this limitation was used to make it easier to copy objects by specifying a new id.

jlhughes:
@kgrgreer I'll create a ticket to capture your comments.

PR #1740, opened 2018-12-11
Only update non-frozen objects

jlhughes:
NP-5304 #time 30m

### DIFF
--- a/src/foam/nanos/auth/LastModifiedAwareDAO.js
+++ b/src/foam/nanos/auth/LastModifiedAwareDAO.js
@@ -21,13 +21,15 @@ foam.CLASS({
     {
       name: 'put_',
       code: function(x, obj) {
-        if ( foam.nanos.auth.LastModifiedAware.isInstance(obj) ) {
+        if ( foam.nanos.auth.LastModifiedAware.isInstance(obj) &&
+             ! obj.isFrozen() ) {
           obj.lastModified = new Date();
         }
         return this.SUPER(x, obj);
       },
       javaCode: `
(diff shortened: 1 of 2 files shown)


kgrgreer:
Where does this happen? Shouldn't we just clone() the object and then update it?

jlhughes:
@kgrgreer If the object is frozen, then there are no pending changes, so no journal update, so no reason to update the lastModified/lastModifiedBy.

kgrgreer:
Then couldn't we just return right away and not do anything? Or, throw an
exception. Someone could modify an object and then freeze it.

jlhughes:
@kgrgreer how to differentiate between never cloned and cloned and frozen?  These are the LastModifiedAwareDAO and LastModifiedByAwareDAOs, I don't think they should fail any operation.  This scenario only occurred during testing.  I'll find the test scenario.

kgrgreer:
I could trick the system by making changes and then freezing the object
before put()-ing. This would prevent LastModifiedAware and LastModifiedBy
from working, which could be a security or accountability issue if I could
then blame others for my updates or back-date updates. I think it should
just be an exception to to put() and frozen object.

jlhughes:
The scenario occurs after the change to disable Accounts rather then delete.  On delete/remove we clone the account, set enabled false, then put.  The Account DAO has a cascade of operations on put for finding/setting-up defaults, but the test was in the process of deleting all of these. 
I'll back out the test for frozen, leave it as it is, and address the tests.   I'm thinking all the Account DAOs now need to be aware of disabled accounts.

PR #1745, opened 2018-12-14
Fix ScrollTableView onWheel

lchanmann:
Address: https://github.com/nanoPayinc/NANOPAY/issues/5370

We can use skip change as an indicator for data change in ScrollTableView and if skip change is not detected we don't prevent default mouse wheel event.


TW80000 (approved):
LGTM

PR #1746, opened 2018-12-14
Add Authorization to emailMessageDAO

TW80000:
Only added checks for reading the DAO for now since I don't know if
we're going to allow regular users to create, update or delete email
messages. But at the moment it's totally without authorization checks,
meaning any user can see all emails sent by querying it through the
console or DIG, seeing sensitive information.

Closes https://github.com/nanoPayinc/NANOPAY/issues/5372


adamvy:
"Only added checks for reading the DAO for now since I don't know if
we're going to allow regular users to..."

We should disallow all actions by regular users until someone presents a use case for it.  Otherwise we might forget to close this later.

TW80000:
> We should disallow all actions by regular users until someone presents a use case for it. Otherwise we might forget to close this later.

True. I did think that but I was worried it might break all of the emails we send out right now, because I think it's usually the user context used when sending an email, and they won't have permission to do that by default if I add the create logic in this PR. So we'd have a brief period where all (or most) emails sent out would be broken, at least I think so.

adamvy:
> So we'd have a brief period where all (or most) emails sent out would be broken, at least I think so.

Not unless you fix that behaviour also in the same PR :)

adamvy:
Should probably turn on authenticateRead in the Authenticated DAO

TW80000:
> Not unless you fix that behaviour also in the same PR :)

Truth

> Should probably turn on authenticateRead in the Authenticated DAO

It does by default using the two-argument constructor. [Code](https://github.com/foam-framework/foam2/blob/master/src/foam/dao/AuthenticatedDAO.java#L25)

PR #1748, opened 2018-12-14
Add charset=utf-8 to HTTPBox.

mcarcaso:


src/foam/net/web/HTTPRequest.js
```diff
@@ -162,6 +162,7 @@ foam.CLASS({
           this.headers['Content-Type'] = 'application/x-www-form-urlencoded';
           break;
         }
+        this.headers['Content-Type'] += '; charset=utf-8'
```
mcarcaso: I don't know enough about character encoding to know if I shouldn't do this across all HTTPRequests we send. Do any of you have any thoughts?
I could make this configurable via a property and give it a default value of utf-8, or I can do something different if you have any ideas.

mcarcaso: Spoke to @adamvy offline and he suggested putting this fix on HTTPBox because we might not always want utf-8 on HTTPRequests but we always want it when talking over the HTTPBox.

adamvy:
LGTM, although in the future it might not always be JSON, but that's not a problem for right now.

In the future we will be doing content-type: application/octet-stream and the box on the server will know to decode as utf-8 text and then parse as json or whatever else, all that logic will live in boxes and the http request will just be a means of sending some raw data ot the other end.

PR #1752, opened 2018-12-17
Email services restructure

TW80000:
## Changes

* Added an `AuthenticatedDAO` decorator to `emailTemplateDAO` with `authenticateRead` set to true
* Removed the `smtpEmailService` service because there should only be one way to send an email, which is the service named "email"
* Deleted the `SMTPEmailMessageDAO` decorator because you shouldn't be able to send emails by putting to `emailMessageDAO`
* Created a `ProxyEmailService` that can be extended to create decorators for `EmailService`s
* Created two new decorators for `EmailService`s:
    1. `TemplateEmailService`, which just contains the logic that processes emails sent from a template, and
    2. `UnimplementedEmailService`, which just logs a warning message when you try to call any of its methods. This is the default `EmailService` in FOAM and should be replaced with another `EmailService` by developers using FOAM in their projects. In nanopay we replace it with an instance of `SMTPEmailService`
* Simplified `DAOEmailService` and made it into an `EmailService` decorator. Now it simply puts to `emailMessageDAO` then calls the appropriate method (`sendEmail` or `sendEmailFromTemplate`) on the delegate


kgrgreer:
Why this: "authenticateRead set to false"?

src/foam/nanos/auth/CountryService.java
```diff
@@ -72,7 +71,7 @@ protected void notifySupport(String query) {
     emailMessage.setSubject ("Unknown Country");
     emailMessage.setBody    ("User just added an unknown country " + query);
```
jlhughes: This is legacy code, but this is where I think we should use notification templates so we don’t have hard coded emails in code.   Other parts of the system are using a notification template ‘NOC’, which would be configured per installation and had an appropriate email(s).

TW80000:
> Why this: "authenticateRead set to false"?

I did this because I assumed the code that gets the template from the `emailTemplateDAO` would then fail when the user context was used, but I just tested that assumption and that's not the case. I pushed commit 83c4801 to this branch that sets `authenticateRead` to true.

src/foam/nanos/auth/CountryService.java
TW80000: Sounds good to me. Is there a ticket for that or should I make one? Also, is it alright if that gets done in another PR? This PR is already (indirectly) blocking some stuff I need to do for Ablii.

jlhughes: no ticket at the moment, just noticed it, and yes, ok, if in another PR.

jlhughes (dismissed):

kgrgreer:
Hi Troy,

  After discussing with Joel, we came to the following design:

1. The EmailService should remain since it's a better interface for sending emails since it has methods for sending emails both with and without templates. This helps with discoverability and avoids hidden or hard to understand features.
2. The EMailDAO should not be useable for sending emails and should be renamed to something like EMailLogDAO to make it clear that it only lets you browse sent emails, but not send them.
Sorry for contradicting my earlier self and causing you extra work.
Thanks

TW80000:
> Hi Troy,
> 
> After discussing with Joel, we came to the following design:
> 
> 1. The EmailService should remain since it's a better interface for sending emails since it has methods for sending emails both with and without templates. This helps with discoverability and avoids hidden or hard to understand features.
> 2. The EMailDAO should not be useable for sending emails and should be renamed to something like EMailLogDAO to make it clear that it only lets you browse sent emails, but not send them.
>    Sorry for contradicting my earlier self and causing you extra work.
>    Thanks

No problem. I'll update this PR with the changes.

TW80000:
@kgrgreer @jlhughes  Please see the first post in this thread. I updated it to reflect the changes I made today. This PR is ready for review again. I didn't rename the `emailMessageDAO`, but I did remove the decorator that sends the emails.

jlhughes:
It is not that FOAM doesn't have an email sending implementation, simply that it doesn't have default SMTP configuration.   Nanopay should simply be providing configuration.  The SMTP layer would be used by POP3EmailService as well for example. 
Instead of the UnimplementedEmailService, the SMTP service could complain that it's host and other config is not setup correctly. 
Also, the emailMessage that is saved to the DAO should be the result of the 'send' as additional properties are setup or changed during template send.   So perhaps send/sendFromTemplate could return the email message formatted and sent.

jlhughes:
sorry, didn't realized I accidentally closed this.

TW80000:
Superseded by https://github.com/foam-framework/foam2/pull/1945

PR #1754, opened 2018-12-17
Property permission check

KateMengjunXia:
Added DAO decorator: when doing put_, find_ or select_ on an object, check if any properties require permissions to write / read.
If a user doesn't have the corresponding permissions, then reset value to original one when the user attempts to update the property, or clear the property value before the user attempts to read the property.

Part of https://github.com/nanoPayinc/NANOPAY/pull/5394


src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+            javaType: 'foam.core.FObject'
+          },
+          {
+            name: 'write',
+            javaType: 'Boolean'
+          }
+        ],
+        javaCode: `
+    if ( oldObj != null ) {
+      AuthService auth = (AuthService) x.get("auth");
```
jlhughes: Determining if this DAO should be applied can be done at DAO isntantiation time.  
Also, all the clone above should occur here only if the permissions do indeed need to be checked and the object needs to be modified. 
The default case should be fast.

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+      String of = obj.getClass().getSimpleName().toLowerCase();
+      List list = obj.getClassInfo().getAxiomsByClass(PropertyInfo.class);
+      Iterator e = list.iterator();
+  
+      while( e.hasNext() ) {
+        PropertyInfo axiom = (PropertyInfo) e.next();
+        if ( axiom.getPermissionRequired() ) {
```
kgrgreer: Faster to find all such properties just once and cache.

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+          String axiomName =  axiom.toString();
+          axiomName = axiomName.substring(axiomName.lastIndexOf(".") + 1);
```
kgrgreer: We have different classes with the same name but different packages. This would cause conflicts.

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+
+      return ((ProxySink) super.select_(x, proxySink, skip, limit, order, predicate)).getDelegate();
+    }
+
+    return super.select_(x, sink, skip, limit, order, predicate);
+        `,
+      },
+
+      {
+        name: 'hideProperties',
```
kgrgreer: I don't think you should use the same method for filtering both put()'s and find/select()s as they do different things.

src/foam/core/Property.js
```diff
@@ -163,6 +163,15 @@ foam.CLASS({
      */
     'required',
 
+    /**
+      When set to true, model.propperty.permission is needed for read / write on this property
```
TW80000: Typo: propperty

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+    name: 'PropertyPermissionDAO',
+  
+    javaImports: [
+      'foam.nanos.auth.AuthService',
+      'foam.core.FObject',
+      'foam.core.PropertyInfo',
+      'foam.core.X',
+      'foam.dao.ProxySink',
+      'java.util.Iterator',
+      'java.util.List'
```
TW80000: Please sort

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
+foam.CLASS({
+    package: 'foam.nanos.auth',
+    name: 'PropertyPermissionDAO',
+
```
TW80000: Please add documentation

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+      'foam.core.FObject',
+      'foam.core.PropertyInfo',
+      'foam.core.X',
+      'foam.dao.ProxySink',
+      'java.util.Iterator',
+      'java.util.List'
+    ],
+  
+    extends: 'foam.dao.ProxyDAO',
+    documentation: `A DAO decorator that prevents user from updating / reading
```
TW80000: Typo: "user" -> "users"

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+          }
+        ],
+        javaCode: `
+    if ( oldObj != null ) {
+      AuthService auth = (AuthService) x.get("auth");
+      String of = obj.getClass().getSimpleName().toLowerCase();
+      List list = obj.getClassInfo().getAxiomsByClass(PropertyInfo.class);
+      Iterator e = list.iterator();
+  
+      while( e.hasNext() ) {
```
TW80000:
```suggestion
      while ( e.hasNext() ) {
```

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+        javaCode: `
+    FObject oldObj = getDelegate().find(obj.getProperty("id"));
+    return super.put_(x, hideProperties(x, obj, oldObj, true));
+        `,
+      },
+  
+      {
+        name: 'find_',
+        javaCode: `
+    FObject oldObj = getDelegate().find(id);
```
TW80000: @kgrgreer Should this be `getDelegate().find_(x, id)` since we're the implementer of the method here?

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+    
+    if ( oldObj != null ) {
+      FObject copiedObj = oldObj.fclone();
+      return hideProperties(x, copiedObj, oldObj, false);
+    }
+
+    return super.find_(x, id);
```
TW80000: If we know it's null at this point, can't we just return null instead of doing the find a second time?

TW80000 (changes requested):

src/foam/nanos/auth/PropertyPermissionDAO.js
KateMengjunXia: Isn't most of the code the same for both put and find/select? So would it be better to use one method for put and a different one for find/select?

TW80000 (approved):

src/foam/nanos/auth/PropertyUserPermissionDAO.js
```diff
@@ -0,0 +1,221 @@
+foam.CLASS({
+    package: 'foam.nanos.auth',
```
kgrgreer: indent 2 spaces

src/foam/nanos/auth/PropertyUserPermissionDAO.js
```diff
@@ -0,0 +1,221 @@
...
+    name: 'PropertyUserPermissionDAO',
```
kgrgreer: I think a better name would be PermissionedPropertyDAO.

src/foam/nanos/auth/PropertyUserPermissionDAO.js
```diff
@@ -0,0 +1,221 @@
...
+
+    return null;
+        `,
+      },
+
+      {
+        name: 'select_',
+        javaCode: `
+    if ( sink != null ) {
+      ProxySink proxySink = new ProxySink(x, sink) {
```
kgrgreer: Creating an in-line sub-class is going to prevent this sink from being marshallable over the network, which will prevent this code from working with networked DAO's which will prevent some forms of clustering and scaling. Better to create the Sink as its own model.

src/foam/nanos/auth/PropertyUserPermissionDAO.js
```diff
@@ -0,0 +1,221 @@
...
+    }
+        `,
+      },
+    ],
+  
+    axioms: [
+      {
+        buildJavaClass: function(cls) {
+          cls.extras.push(`
+  protected Map<String, List<PropertyInfo>> propertyMap = new LinkedHashMap<>();
```
kgrgreer: Have protected variable names end with _. ie. propertyMap_.

src/foam/nanos/auth/PropertyUserPermissionDAO.js
```diff
@@ -0,0 +1,221 @@
...
+
+      return ((ProxySink) super.select_(x, proxySink, skip, limit, order, predicate)).getDelegate();
+    }
+
+    return super.select_(x, sink, skip, limit, order, predicate);
+        `,
+      },
+
+      {
+        name: 'clearProperties',
```
kgrgreer: I think their could be a clearer name for this method.

src/foam/nanos/auth/PropertyUserPermissionDAO.js
kgrgreer: What is this map for? Please document.

src/foam/nanos/auth/PropertyUserPermissionDAO.js
```diff
@@ -0,0 +1,221 @@
...
+            javaType: 'foam.core.FObject'
+          },
+          {
+            name: 'oldObj',
+            javaType: 'foam.core.FObject'
+          }
+        ],
+        javaCode: `
+    if ( oldObj != null ) {
+      String of = obj.getClass().getSimpleName().toLowerCase();
```
kgrgreer: Why is the toLowerCase() needed? It just creates an extra object/garbage.

src/foam/nanos/auth/PropertyUserPermissionDAO.js
kgrgreer: Why a LinkedHashMap? Is the order important?

src/foam/nanos/auth/PropertyUserPermissionDAO.js
KateMengjunXia: Because the permission names are in lower case (e.g. user.ro.onboarded)

KateMengjunXia:
Seems like I merged a wrong branch to this one, will fix it soon.

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+        `,
+      },
+
+      {
+        name: 'select_',
+        javaCode: `
+    if ( sink != null ) {
+      ProxySink proxySink = new ProxySink(x, sink) {
```
jlhughes: This sink needs to be its own model/class.

src/foam/nanos/auth/PropertyPermissionDAO.js
```diff
@@ -0,0 +1,133 @@
...
+          if ( ! write ) {
+            hasPermission = hasPermission || auth.check(x, of + ".ro." + axiomName);
+          }
+  
+          if ( ! hasPermission ) {
+            if ( write ) {
+              Object oldValue = oldObj.getProperty(axiomName);
+              axiom.set(obj, oldValue);
+            } else {
+              axiom.clear(obj);
```
jlhughes: This should also occur when oldObj is null, so that you can't create a new object with properties set to particular values that you would normally not be able to update.

jlhughes (changes requested):

src/foam/nanos/auth/PropertyPermissionDAO.js
KateMengjunXia: Updated

jlhughes: @KateMengjunXia did you push or lose the change regarding oldObj?

jlhughes (approved):

PR #1765, opened 2018-12-27
Clear the axiomCache when the parent has axioms installed on it.

### LINKED ISSUE #1764: Stale axiom cache on children when refining parent
Consider the following example:
```
foam.CLASS({
  package: 'refinetest',
  name: 'Test1',
  methods: [
    function method1() {
      console.log('yo');
    }
  ],
});

foam.CLASS({
  package: 'refinetest',
  name: 'Test2',
  extends: 'refinetest.Test1',
});

console.log('Before Refinement');
console.log('Base   class methods:', refinetest.Test1.getAxiomsByClass(foam.core.Method).length);
console.log('Extend class methods:', refinetest.Test2.getAxiomsByClass(foam.core.Method).length);

foam.CLASS({
  package: 'refinetest',
  name: 'Test1Refine',
  refines: 'refinetest.Test1',
  methods: [
    function method2() {
      console.log('yo');
    }
  ],
});

console.log('After Refinement');
console.log('Base   class methods:', refinetest.Test1.getAxiomsByClass(foam.core.Method).length);
console.log('Extend class methods:', refinetest.Test2.getAxiomsByClass(foam.core.Method).length);
```

An extending class should always have at least as many axioms as its parent class but in this case, the extending class has one less method than its parent class in the second set of log statements. This is a result of the axiomCache of a child class not getting cleared when its parent is refined.

mcarcaso:
Fixes https://github.com/foam-framework/foam2/issues/1764 but I'm not sure if this is the best place for a fix.


src/foam/core/Boot.js
```diff
@@ -154,7 +154,10 @@ foam.LIB({
           // Relay 'installAxiom' events from parent class.
           parent.pubsub_ && parent.pubsub_.sub(
             'installAxiom',
-            function(_, a1, a2, a3) { cls.pubsub_.pub(a1, a2, a3); });
+            function(_, a1, a2, a3) {
+              cls.private_.axiomCache = {};
```
adamvy, line 162: Add a comment on why.

mcarcaso: Done

PR #1768, opened 2018-12-28
Authorization bug fixes

### LINKED ISSUE #1767: AuthorizationDAO bugs
* [x] `removeAll_` should only check `authorizeOnDelete`, but it checks `authorizeOnRead` as well right now, which was unintended.
* [x] `remove_` should call `authorizeOnDelete` with the existing object, not the one passed in. This is to avoid people modifying an object in such a way to pass the authorization checks but keep the same id as an existing object, allowing them to delete an object that they shouldn't be allowed to.

TW80000:
Closes #1767 

* When removing, check existing object instead of given one
  * This is to avoid people manipulating an object's properties before
removing it in such a way that they could pass the authorization check
with the modified properties but wouldn't have been able to pass it had
they not modified those properties.
* Minor optimization: Avoid unnecessary second lookup in `find_` method
* Fix bug where read authorization check was being called for deletes
  * It used to call `authorizeOnRead` because calling `this.select_` with
the AuthorizationSink would call `authorizeOnRead` in the `put` method
of the sink. Then the RemoveSink would call `remove_` which would call
`authorizeOnDelete`, which is the call we want.
  * Now it doesn't call `authorizeOnRead`, just `authorizeOnDelete` like it
should.
  * It's important that we call `select_` on the delegate instead of `this`
because otherwise we'd still be hitting the AuthorizationSink which
calls the `authorizeOnRead` method.
  * It's an optimization to pass the delegate to the RemoveSink instead of
`this` because that way we avoid unnecessarily calling
`authorizeOnDelete` a second time when the RemoveSink calls the
`remove_` method on the DAO given to it.

### DIFF
--- a/src/foam/dao/AuthorizationSink.java
+++ b/src/foam/dao/AuthorizationSink.java
@@ -16,29 +16,29 @@ public class AuthorizationSink
   extends ProxySink
 {
   public Authorizer authorizer_;
+  public Boolean checkDelete_;
 
   public AuthorizationSink(X x, Authorizer authorizer, Sink delegate) {
+    this(x, authorizer, delegate, false);
+  }
+
+  public AuthorizationSink(X x, Authorizer authorizer, Sink delegate, Boolean checkDelete) {
     super(x, delegate);
     authorizer_ = authorizer;
+    checkDelete_ = checkDelete;
   }
 
   @Override
   public void put(Object obj, Detachable sub) {
     try {
-      authorizer_.authorizeOnRead(getX(), (FObject) obj);
+      if ( checkDelete_ ) {
+        authorizer_.authorizeOnDelete(getX(), (FObject) obj);
+      } else {
+        authorizer_.authorizeOnRead(getX(), (FObject) obj);
+      }
       super.put(obj, sub);
     } catch ( AuthorizationException e ) {
-      // Do not put to sink if not authorized to read this object.
-    }
-  }
-
-  @Override
-  public void remove(Object obj, Detachable sub) {
-    try {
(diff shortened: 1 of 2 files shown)


src/foam/dao/AuthorizationSink.java
```diff
@@ -16,29 +16,29 @@
...
     authorizer_ = authorizer;
+    checkDelete_ = checkDelete;
   }
 
   @Override
   public void put(Object obj, Detachable sub) {
     try {
-      authorizer_.authorizeOnRead(getX(), (FObject) obj);
+      if ( checkDelete_ ) {
+        authorizer_.authorizeOnDelete(getX(), (FObject) obj);
```
kgrgreer, line 35: Is updating really the same as deleting?

TW80000: I'm not sure what you mean. I don't see how this relates to updating anything.

kgrgreer: put() is called to update an object but you're checking if the user is authorized to delete it. I don't think that updating an object is really the same as deleting it outright, which is what the check is checking.

TW80000: I'm assuming either you thought this was a DAO or I don't understand how sink puts work.

> put() is called to update an object

The `put` method defined for DAO's does, but this is a sink, not a DAO. The `put` method for sinks is called when adding something to a sink, sink puts have nothing to do with updating objects (or at least in this case it doesn't), right? Or am I misunderstanding something?

kgrgreer: I see. My mistake.

PR #1770, opened 2019-01-02
Context menu fix

### LINKED ISSUE #1769: Context menu in tables isn't visible past bottom of table
Example:

screen shot 2018-12-28 at 1 50 03 pm

TW80000:
Closes #1769 

* Fixes bug where the context menu wouldn't be visible past the bottom of the table
* Originally I only intended to remove the artificial restriction we were putting on the height of the dropdown in `getFullHeight` but then I took it further and removed the animation altogether. If we still want the animation I can add it back.

### DIFF
--- a/src/foam/u2/md/OverlayDropdown.js
+++ b/src/foam/u2/md/OverlayDropdown.js
@@ -114,9 +114,9 @@ foam.CLASS({
     function add() {
       // TODO: Replace with content @kgr
       if ( this.addToSelf_ ) {
-        this.SUPER.apply(this, arguments);
+        this.SUPER(...arguments);
       } else {
-        this.dropdownE_.add.apply(this.dropdownE_, arguments);
+        this.dropdownE_.add(...arguments);
       }
 
       return this;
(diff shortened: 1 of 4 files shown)


src/foam/u2/md/OverlayDropdown.js
```diff
@@ -114,78 +83,45 @@ foam.CLASS({
     function add() {
       // TODO: Replace with content @kgr
       if ( this.addToSelf_ ) {
-        this.SUPER.apply(this, arguments);
```
adamvy, line 117: Personally I don't like mixing stylistic changes with functional changes in the same PR.

But you've already written it so I guess no need to revert your changes.  Something to think about going forward.

TW80000: That makes sense. I usually try to do the same thing but instead of a separate PR I try to put all of the style-only changes in one commit. That way it's easy to hide all of the style changes when reviewing because you can just hide that commit when looking at the diff. Seems like different ways of doing the same thing to me, unless there's an advantage you know of to putting them in a separate PR.

PR #1774, opened 2019-01-02
Fix the password word reset not working

sirenChen:
refs [5498](https://github.com/nanoPayinc/NANOPAY/issues/5498)

- [x] fix: email - password reset not working

The real password reset and hashing happens [here](https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/auth/UserPasswordHashingDAO.java#L31)

### DIFF
--- a/src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js
+++ b/src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js
@@ -132,10 +132,8 @@ if ( ! Password.isValid(newPassword) ) {
 
 // update user's password
 userResult = (User) userResult.fclone();
-userResult.setPasswordLastModified(Calendar.getInstance().getTime());
-userResult.setPreviousPassword(userResult.getPassword());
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js
```diff
@@ -132,10 +132,8 @@ if ( ! Password.isValid(newPassword) ) {
 
 // update user's password
 userResult = (User) userResult.fclone();
-userResult.setPasswordLastModified(Calendar.getInstance().getTime());
```
jlhughes, line 135: Reason this and next line are deleted?

sirenChen: please see this [code](https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/auth/UserPasswordHashingDAO.java#L31), as long as the DesiredPassword is not empty, it will update the LastModified and previousPassword there.

PR #1777, opened 2019-01-04
Validate password method on auth service

deejes:
1. Adds a validatePassword password method to AuthService, which in turn calls Password.validatePassword
2. Adds a passwordErrorMessage to the Password model, which should be updated if/when password requirements change.
3. Removes frontEnd password validation, and validatePassword is called by updatePassword when there is an attempt to set a new password.
4. Makes validate password available through the auth service, creating a single source of truth for front end validation to rely and call upon.
5. Fixes styling issues in a number of files.

Next steps would be to go through the system and remove password validation on the front end, and call the validatePassword method on auth service instead.

### DIFF
--- a/src/foam/nanos/auth/AuthService.js
+++ b/src/foam/nanos/auth/AuthService.js
@@ -111,8 +111,7 @@ foam.INTERFACE({
     },
     {
       name: 'validatePassword',
-      javaReturns: 'boolean',
-      swiftReturns: 'Bool',
+      javaThrows: [ 'java.lang.RuntimeException' ],
       swiftThrows: true,
       args: [
         {
--- a/src/foam/nanos/auth/AuthService.js
+++ b/src/foam/nanos/auth/AuthService.js
@@ -110,6 +110,18 @@ foam.INTERFACE({
       ]
     },
     {
+      name: 'validatePassword',
+      javaReturns: 'boolean',
+      swiftReturns: 'Bool',
+      swiftThrows: true,
+      args: [
+        {
+          name: 'potentialPassword',
+          javaType: 'String',
+        }
+      ]
+    },
+    {
       name: 'checkUser',
       javaReturns: 'boolean',
       swiftReturns: 'Bool',
--- a/src/foam/nanos/auth/ChangePasswordView.js
+++ b/src/foam/nanos/auth/ChangePasswordView.js
@@ -139,13 +139,9 @@ foam.CLASS({
   `,
 
   messages: [
-    { name: 'noSpaces', message: 'Password cannot contain spaces' },
-    { name: 'noNumbers', message: 'Password must have one numeric character' },
-    { name: 'noSpecial', message: 'Password must not contain: !@#$%^&*()_+' },
     { name: 'emptyOriginal', message: 'Please enter your original password'},
     { name: 'emptyPassword', message: 'Please enter your new password' },
     { name: 'emptyConfirmation', message: 'Please re-enter your new password' },
-    { name: 'invalidLength', message: 'Password must be 7-32 characters long' },
     { name: 'passwordMismatch', message: 'Passwords do not match' },
     { name: 'passwordSuccess', message: 'Password successfully updated' }
   ],
--- a/src/foam/nanos/auth/ChangePasswordView.js
+++ b/src/foam/nanos/auth/ChangePasswordView.js
@@ -205,33 +205,13 @@ foam.CLASS({
           return;
         }
 
-        // validate new password
+        // check if new password entered
         if ( ! this.newPassword ) {
           this.add(this.NotificationMessage.create({ message: this.emptyPassword, type: 'error' }));
           return;
         }
 
-        if ( this.newPassword.includes(' ') ) {
-          this.add(this.NotificationMessage.create({ message: this.noSpaces, type: 'error' }));
-          return;
-        }
-
-        if ( this.newPassword.length < 7 || this.newPassword.length > 32 ) {
-          this.add(this.NotificationMessage.create({ message: this.invalidLength, type: 'error' }));
-          return;
-        }
-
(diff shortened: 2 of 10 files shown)


src/foam/nanos/auth/AuthService.js
```diff
@@ -109,6 +109,18 @@ foam.INTERFACE({
         }
       ]
     },
+    {
+      name: 'validatePassword',
```
kgrgreer, line 113: How do you know what's wrong with the password if it doesn't like it? An Exception (which is reasonable)? But if that's the case, then we don't really need a return value do we? Validity as implied but just returning without an exception.

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -205,33 +201,13 @@ foam.CLASS({
...
-          return;
-        }
-
-        if ( /[^a-zA-Z0-9]/.test(this.newPassword) ) {
-          this.add(self.NotificationMessage.create({ message: this.noSpecial, type: 'error' }));
-          return;
-        }
-
-        // check if confirmation entered
+        // check if new password confirmation entered
```
kgrgreer, line 210: I think we should get the message from the password validator (via Exception?). Internationalization is going to be a pain, but we can worry about that later.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -277,6 +274,10 @@ public boolean check(foam.core.X x, String permission) {
     return checkPermission(x, new AuthPermission(permission));
   }
 
+  public boolean validatePassword( String newPassword) {
+    return Password.isValid(newPassword);
```
kgrgreer: I think that the validation should be here directly, not in Password.

src/foam/nanos/auth/UserAndGroupAuthService.java
```diff
@@ -316,21 +317,8 @@ public User updatePassword(foam.core.X x, String oldPassword, String newPassword
...
-      throw new RuntimeException("Password must have one capital letter");
-    }
-
-    if ( ! newPassword.matches(".*\\d+.*") ) {
-      throw new RuntimeException("Password must have one numeric character");
-    }
-
-    if ( alphanumeric.matcher(newPassword).matches() ) {
-      throw new RuntimeException("Password must not contain: !@#$%^&*()_+");
+    if (! validatePassword(newPassword)) {
```
kgrgreer: Proper FOAM spacing:
if ( ! validatePassword(newPassword) ) {

src/foam/nanos/auth/resetPassword/ResendView.js
```diff
@@ -92,11 +92,11 @@ foam.CLASS({
   ],
 
   messages: [
-    { name: 'Instructions', message: "We've sent the instructions to your email. Please check your inbox to continue."}
+    { name: 'Instructions', message: 'We\'ve sent the instructions to your email. Please check your inbox to continue.'}
```
kgrgreer: The style guideline isn't to always use single-quotes, it's to use single quotes by default, unless it makes sense to use double quotes for reasons like avoiding escaping strings which include single-quotes.

src/foam/nanos/auth/UserAndGroupAuthService.java
deejes: If we put the validation here, won't it still be the case that there will be two places where we define what appropriate validation is (Password and here)? 
I'm fine with putting the validation here, just trying to understand why its a better choice.
The Password  model seems like a more intuitive place to put the rule, and we can always still overwrite this function to put custom rules in place.

src/foam/nanos/auth/UserAndGroupAuthService.java
deejes: Resolved.

src/foam/nanos/auth/resetPassword/ResendView.js
deejes: Resolved.

src/foam/nanos/auth/UserAndGroupAuthService.java
deejes: For documentation - spoke with kgrgreer offline, and it was decided to put the rule here and make Password call this function instead of hard coding it there. This file will be the source of truth for password validation, and will be the isValid method of Password will be rewritten to call this function instead of the other way around.

jlhughes:
So how does this work with password create/edit where we show strength indicators and where I would expect to apply patterns so I simply cannot enter an invalid password.

src/foam/nanos/auth/ChangePasswordView.js
deejes: Just to confirm that I understand - are you talking about a particular one of these messages or all of them?

Currently, the authService only receives the old password and the new password. 
The check that the new password and the confirmation match is currently only on the front end. 

To move all messages to the service would require that we pass everything to the service, and perform all the checks there. 

I think it's probably cleaner to move all messages and logic to the service.

@kgrgreer

deejes:
Strength indicators are only on the front end, so this shouldn’t change anything on with respect to those. The issue this is trying to resolve is the front end password validation littered across the views. 

We can however now add the strength indicator validation to the service and remove the validation part from the front end.

src/foam/nanos/auth/AuthService.js
deejes: Done .

src/foam/nanos/auth/ChangePasswordView.js
jlhughes: Services - yes all password validation/reset/change and subsequent messaging should reside service side.  Consider and LDAPAuthService - this validation would be occurring in the LDAP server.

deejes: Noted and issue created.

PR #1779, opened 2019-01-07
Fix bug where popup background covers popup in Safari

TW80000:
In Safari, the popup overlay for a certain modal doesn't work in Ablii, see here:

The background is covering the rest of the modal, meaning it can't be interacted with.

I looked up the issue and found a fix that works from [this Stack Overflow thread](https://stackoverflow.com/a/40896165/2512768), although I couldn't actually tell you why it fixes the problem unfortunately.

Necessary for https://github.com/nanoPayinc/NANOPAY/issues/5551

### DIFF
--- a/src/foam/u2/dialog/Popup.js
+++ b/src/foam/u2/dialog/Popup.js
@@ -60,6 +60,12 @@ foam.CLASS({
     }
     ^inner {
       z-index: 3;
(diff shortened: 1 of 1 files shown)


TW80000:
Closing because this isn't happening anymore, even without the changes in this PR. Recently the code that adds the popup to the view changed where it gets added, which must have fixed the issue.

PR #1781, opened 2019-01-08
Fix object cloning

### LINKED ISSUE #1780: Cloning objects in JavaScript doesn't actually clone, just returns reference

TW80000:
Closes #1780 

@Henry0422 and I ran into an issue where cloning a Menu and setting a property on the view on the handler of the menu (a property on a twice-nested object) was also setting the value on the original Menu that we cloned, meaning it wasn't a true clone. 

That lead us to look into it and found that the existing clone method for objects just returns the object to be cloned, which obviously doesn't clone it.

This PR adds an implementation that should work for true deep cloning at arbitrary levels of nesting since it gets called recursively for nested objects, but only with support for the types that stdlib recognizes.

### DIFF
--- a/src/foam/core/stdlib.js
+++ b/src/foam/core/stdlib.js
@@ -777,7 +777,7 @@ foam.LIB({
     function clone(o) {
       const newObj = {};
 
-      for ( let key in o ) {
+      for ( var key in o ) {
         if ( o.hasOwnProperty(key) ) {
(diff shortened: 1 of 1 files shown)


lchanmann:
@TW80000, it seems like this change break some tests https://travis-ci.org/foam-framework/foam2/jobs/482013531.

PR #1782, opened 2019-01-08
Added a snippet view on DIG

minsunny:
https://github.com/nanoPayinc/NANOPAY/issues/5494

Added a snippet view as an enhancement on DIG for nanometer

screen shot 2019-01-08 at 3 35 37 pm

### DIFF
--- a/src/foam/nanos/dig/DIG.js
+++ b/src/foam/nanos/dig/DIG.js
@@ -47,8 +47,7 @@ foam.CLASS({
             return [nspec.id, nspec.id];
           }
         });
-      },
-      value: 'accountDAO'
+      }
     },
     'cmd',
     'format',
--- a/src/foam/nanos/dig/DIG.js
+++ b/src/foam/nanos/dig/DIG.js
@@ -47,7 +47,8 @@ foam.CLASS({
             return [nspec.id, nspec.id];
           }
         });
-      }
+      },
+      value: 'accountDAO'
     },
     'cmd',
     'format',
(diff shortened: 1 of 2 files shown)


jlhughes (approved):

src/foam/nanos/dig/DigSnippetView.js
```diff
@@ -0,0 +1,150 @@
...
+                                    var u = new URL(url);
+                                    var protocol = u.protocol.slice(0, -1)
+                                    return `
+              var password = 'REPLACE_WITH_PASSWORD';
+
+  const ${protocol} = require('${protocol}');
+  var req = ${protocol}.request({
+    hostname: '${u.hostname}',
+    port: ${u.port},
+    path: '${data}',
```
nanoscott: I don't think this is right. The data is the JSON data for the body, is it not?

src/foam/nanos/dig/DigSnippetView.js
```diff
@@ -0,0 +1,150 @@
...
+    method: 'GET',
+    headers: {
+      'accept': 'application/json',
+      'content-type': 'application/json',
+      'cache-control': 'no-cache',
+      'authorization': "Basic " + new Buffer("${user.email}:" + password).toString("base64"),
+    }
+              }, (resp) => {
```
nanoscott: I think you are missing the code which passes in a payload to the API call here. If this is a add/update operation, then you can make the method a PUT and add the data as the message payload instead of passing it in on the URL, which may hit the maximum limits if the data is large.

src/foam/nanos/dig/DIG.js
```diff
@@ -99,9 +99,70 @@ foam.CLASS({
       label: 'URL',
```
nanoscott: Does it make sense to still have this field if we are doing the same thing below? At the least, we shouldn't be duplicating the code.

nanoscott (changes requested):

nanoscott (approved):

PR #1786, opened 2019-01-09
Local and served session Dao

deejes:
Split sessionDAO into localSessionDAO which is not served and sessionDAO which is authenticated and served.


deejes:
Thanks

PR #1787, opened 2019-01-11
Add disabled section support for RichChoiceView

lchanmann:
Needed for: https://github.com/nanoPayinc/NANOPAY/pull/5560/commits/f90c693e28e523ae3c00069f6e3e97389c3b8b47#diff-fd0055649365f16f465f5a61bb8097e8R442

### Screenshot:
screen shot 2019-01-13 at 11 38 27 am

### DIFF
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -29,7 +29,13 @@ foam.CLASS({
               {
                 heading: 'Users',
                 dao: X.userDAO.orderBy(foam.nanos.auth.User.LEGAL_NAME)
-              }
+              },
+              // Set "disabled: true" to render each object as non-selectable row
+              {
+                disabled: true,
+                heading: 'Disabled users',
+                dao: X.userDAO.where(this.EQ(foam.nanos.auth.User.STATUS, this.AccountStatus.DISABLED)),
+              },
             ]
           };
         }
@@ -256,15 +262,22 @@ foam.CLASS({
                   .add(section.heading)
                 .end()
                 .start()
-                  .select(section.filtered || section.dao, function(obj) {
-                    return this.E()
-                      .start(self.rowView, { data: obj })
-                        .on('click', () => {
-                          self.fullObject_ = obj;
-                          self.data = obj;
-                          self.isOpen_ = false;
-                        })
-                      .end();
+                  .callIfElse(! section.disabled, function() {
+                    this.select(section.filtered || section.dao, function(obj) {
+                      return this.E()
+                        .start(self.rowView, { data: obj })
+                          .on('click', () => {
+                            self.fullObject_ = obj;
+                            self.data = obj;
+                            self.isOpen_ = false;
+                          })
+                        .end();
+                    });
+                  }, function() {
+                    this.select(section.filtered || section.dao, function(obj) {
+                      return this.E()
+                        .tag(self.rowView, { data: obj, disabled: true });
+                    });
                   })
                 .end();
             });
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/RichChoiceView.js
```diff
@@ -29,7 +29,13 @@ foam.CLASS({
               {
                 heading: 'Users',
                 dao: X.userDAO.orderBy(foam.nanos.auth.User.LEGAL_NAME)
-              }
+              },
+              // Set "disabled: true" to render each object as non-selectable row
+              {
+                disabled: true,
```
jlhughes, line 35: @TW80000 Why is the User model explicitly referenced in this generic RichChoiceView?
There should be a UserRickChoiceView which extends RichChoiceView and sets the DAO accordingly, or some other reference property.   RichChoiceView should have no explicit knowledge or reference to another model.   This needs to be fixed.

jlhughes (changes requested):

src/foam/u2/view/RichChoiceView.js
lchanmann: It's an example usage in documentation only not the implementation.

jlhughes (approved):
@lchanmann Approved, just make sure the Freeze branch doesn't get included with this one, thanks.

src/foam/u2/view/RichChoiceView.js
```diff
@@ -256,15 +262,22 @@ foam.CLASS({
...
+                            self.fullObject_ = obj;
+                            self.data = obj;
+                            self.isOpen_ = false;
+                          })
+                        .end();
+                    });
+                  }, function() {
+                    this.select(section.filtered || section.dao, function(obj) {
+                      return this.E()
+                        .tag(self.rowView, { data: obj, disabled: true });
```
TW80000: This means that all row views must have a `disabled` property and use it, right? That's one way to do it and it's not necessarily bad, but it would be less code if we put a "disabled" style in RichChoiceView directly that applies to whatever `rowView` was given.

Maybe something like

```CSS
^disabled {
  filter: grayscale(100%) opacity(80%);
}
```

That way is less flexible, but it saves each rowView from having to account for its own disabled style. Maybe that's what we want though, I'm open to either way.

TW80000 (approved):

src/foam/u2/view/RichChoiceView.js
jlhughes: @TW80000 with the style approach, each entry would be listed in natural order but styled differently (default provided) and still not selectable.  
Also the obj could provided a function to control disabled, which we already have with enabledAware.  By default, the function would test enabled, otherwise it is provided.

lchanmann: Here we relay on developer to provide DAOs with the right data to it instead of checking each obj `enabled` property. For example,
```javascript
foam.u2.view.RichChoiceView.create({
  selectionView: { class: 'net.nanopay.auth.ui.UserSelectionView' },
  rowView: { class: 'net.nanopay.auth.ui.UserCitationView' },
  sections: [
    {
        heading: 'Contacts',
        dao: dao1
    },
    {
        disabled: true,
        heading: 'Disabled contacts',
        dao: dao2
    }
  ]
});
```

src/foam/u2/view/RichChoiceView.js
```diff
@@ -259,10 +273,13 @@ foam.CLASS({
                   .select(section.filtered || section.dao, function(obj) {
                     return this.E()
                       .start(self.rowView, { data: obj })
-                        .on('click', () => {
-                          self.fullObject_ = obj;
-                          self.data = obj;
-                          self.isOpen_ = false;
+                        .addClass(section.disabled ? 'disabled' : '')
```
TW80000, line 276: `.enableClass('disabled', section.disabled)` would make more sense here. Not a big deal though.

lchanmann: Fixed: #1791

PR #1790, opened 2019-01-14
Add the global toast notification

Henry0422:
People would utilize `this.add(this.NotificationMessage.create({ message, type }));` lots of times , it would be a good idea to add this in the genearl `ApplicationController.js`.

It will compatible with all application based on foam.


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -339,6 +340,11 @@ foam.CLASS({
     // not navigate to a new screen. Typically for SubMenus
     function lastMenuLaunchedListener(m) {
       this.lastMenuLaunched = m;
+    },
+
+    // This listener is for toast notification message
```
TW80000: Shouldn't this just be a method?

TW80000 (approved):

PR #1792, opened 2019-01-14
Make visibility icon right

Henry0422:
Fix issue #5001 in the app. https://github.com/nanoPayinc/NANOPAY/issues/5001
Eye icon is switched around.


src/foam/u2/view/PasswordView.js
```diff
@@ -36,10 +36,23 @@ foam.CLASS({
...
+      name: 'VISIBILITY_OFF',
+      value: '/foam2/src/foam/u2/images/visibility-off.svg'
+    }
+  ],
+
   properties: [
     {
       name: 'visibilityIcon',
-      value: '/foam2/src/foam/u2/images/visibility.svg'
+      value: '/foam2/src/foam/u2/images/visibility-off.svg'
```
kgrgreer: Isn't this redundant, since you have the same value as a constant?
Would this be better:
  factory: function() { return this.VISIBILITY_OFF; }

Henry0422: OK

PR #1793, opened 2019-01-15
Export the invoice as csv

sirenChen:
refs : 
[5562](https://github.com/nanoPayinc/NANOPAY/issues/5562) - issue 6
[4068](https://github.com/nanoPayinc/NANOPAY/issues/4068)

We are getting an error when we exporting the invoice as the csv.
interface

This is because when we export the data as csv, we are trying to create a new object if the property is undefined. see [if ( o[p.name] == undefined ) o[p.name] = p.of.id;](https://github.com/foam-framework/foam2/blob/master/src/foam/lib/csv/CSV.js#L89).

But the issue is we can not instantiate this Blob datatype because it's an interface.

The easiest way to fix this is provide an adapt function to this property.

### DIFF
--- a/src/foam/nanos/fs/File.js
+++ b/src/foam/nanos/fs/File.js
@@ -34,7 +34,10 @@ foam.CLASS({
     {
       class: 'Blob',
       name: 'data',
-      documentation: 'File data'
+      documentation: 'File data',
+      adapt: function(oldObj, newObj) {
+        return newObj;
+      }
     },
     {
       class: 'String',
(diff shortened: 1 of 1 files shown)


src/foam/nanos/fs/File.js
```diff
@@ -34,7 +34,10 @@ foam.CLASS({
     {
       class: 'Blob',
       name: 'data',
-      documentation: 'File data'
+      documentation: 'File data',
+      adapt: function(oldObj, newObj) {
```
kgrgreer, line 45: This should be documented, otherwise it looks like a NOP and someone is likely to delete it.

jlhughes (approved):

PR #1795, opened 2019-01-15
Remove pop3 from client side

sirenChen:
Remove the pop3 email service from the client side.


sirenChen:
close this because this has already done at [removed pop3 in requires and imports](https://github.com/foam-framework/foam2/pull/1794)

PR #1797, opened 2019-01-15
Add other sudo method signature to scripts and fix sudo by email

TW80000:


jlhughes:
This fails the build. 
I'll fix it.

PR #1799, opened 2019-01-16
Change session timeout duration

sirenChen:
refs
[5653](https://github.com/nanoPayinc/NANOPAY/issues/5653)

1. Change session timeout duration to 5 mins
2. Change the frequency of the session expire cron job to once a mins.

3. Fix the MDAO.removeAll() method. Previously, it will remove all the entity in the MDAO. But it
should only removes all matching entries from the DAO.


jlhughes:
Why are we changing this in foam. This is a nanopay requirement. Just add the appropriate journal entry on nanopay to set this.

kgrgreer:
Why is this a nanopay requirement? This is going to be very annoying for users to have to re-login every five minutes.

tmiller95:
@kgrgreer see https://github.com/nanoPayinc/NANOPAY/issues/5653

After consulting with Nicko (security) and Tony (UX) it was determined that this was acceptable for both security and experience. 

This is 5 minutes of inactivity though, not 5 minutes being logged in.

PR #1800, opened 2019-01-16
Fix MDAO removeAll methtod

sirenChen:
According to the documentation [See here](https://github.com/sirenChen/foam2/blob/master/doc/guides/Dao.md#removeall)

> `removeAll()` is very similar to `select()`, with the obvious exception that it
> removes all matching entries from the DAO instead of returning them.

But currently mdao.removeAll() will ignore the filtering and delete every entry.

### DIFF
--- a/src/foam/dao/MDAO.java
+++ b/src/foam/dao/MDAO.java
@@ -5,11 +5,7 @@
  */
 package foam.dao;
 
-import foam.core.AbstractFObject;
-import foam.core.ClassInfo;
-import foam.core.FObject;
-import foam.core.PropertyInfo;
-import foam.core.X;
+import foam.core.*;
 import foam.dao.index.*;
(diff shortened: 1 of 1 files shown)


src/foam/dao/MDAO.java
```diff
@@ -165,7 +161,14 @@ public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Pre
 
   public void removeAll_(X x, long skip, long limit, Comparator order, Predicate predicate) {
     synchronized ( lock_.writeLock() ) {
-      state_ = null;
+
```
kgrgreer, line 164: This should be:
if ( predicate == null ) { state_ = null } else { super.select_(x, skip, limit, order, predicate); }

sirenChen: @kgrgreer thank you very much for your help.

PR #1801, opened 2019-01-16
Hide Headers Rich Choice View

zahnno:
Hides headers in rich choice view if no data is presents within the dao provided.

### DIFF
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -264,26 +264,29 @@ foam.CLASS({
           }))
           .add(this.slot(function(sections) {
             return this.E().forEach(sections, function(section) {
-              this
-                .start()
-                  .addClass(self.myClass('heading'))
-                  .add(section.heading)
-                .end()
-                .start()
-                  .select(section.filtered || section.dao, function(obj) {
-                    return this.E()
-                      .start(self.rowView, { data: obj })
-                        .enableClass('disabled', section.disabled)
-                        .callIf(! section.disabled, function() {
-                          this.on('click', () => {
-                            self.fullObject_ = obj;
-                            self.data = obj;
-                            self.isOpen_ = false;
+              var positionElement = this;
+              section.dao.select().then((resp) => {
+                this
+                  .start().hide(!! section.hideOnEmpty && ( ! Array.isArray(resp.array) || ! resp.array.length ))
+                    .addClass(self.myClass('heading'))
+                    .add(section.heading)
+                  .end()
+                  .start()
+                    .select(section.filtered || section.dao, function(obj) {
+                      return positionElement.E()
+                        .start(self.rowView, { data: obj })
+                          .enableClass('disabled', section.disabled)
+                          .callIf(! section.disabled, function() {
+                            positionElement.on('click', () => {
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/RichChoiceView.js
```diff
@@ -264,26 +266,29 @@ foam.CLASS({
...
-                    return this.E()
-                      .start(self.rowView, { data: obj })
-                        .enableClass('disabled', section.disabled)
-                        .callIf(! section.disabled, function() {
-                          this.on('click', () => {
-                            self.fullObject_ = obj;
-                            self.data = obj;
-                            self.isOpen_ = false;
+              var positionElement = this;
+              section.dao.select().then((resp) => {
```
TW80000: It would be better to use a COUNT mlang here I think. I think the current method has to send all of the objects in the DAO across the network so we can get the length of the array on the client, whereas using a COUNT mlang would only have to send the count across the network. Please correct me if I'm wrong.

TW80000: I wonder if this changing from synchronous to asynchronous is going to break the ordering for multiple sections. If there are two sections and the first section's promise takes longer to finish than the second section's promise, then I'm guessing the second section would be rendered before the first section, which we wouldn't want.

src/foam/u2/view/RichChoiceView.js
```diff
@@ -31,9 +31,11 @@ foam.CLASS({
                 dao: X.userDAO.orderBy(foam.nanos.auth.User.LEGAL_NAME)
               },
               // Set "disabled: true" to render each object as non-selectable row
+              // Set "hideOnEmpty: true" to hide headers if not objects are present in dao provided.
               {
                 disabled: true,
                 heading: 'Disabled users',
+                hideOnEmpty: true,
```
TW80000: "On" in a name makes me think it's an event listener, like onClick. A section being empty isn't an event (imo) so I think a better name would be `hideIfEmpty`.

TW80000 (changes requested):

src/foam/u2/view/RichChoiceView.js
zahnno: Nope you're right. I'll make the change

zahnno: I'm not 100% sure but I don't think that would be the case considering the select response is contained within a forEach function created by an element. @kgrgreer might be able to confirm this.

src/foam/u2/view/RichChoiceView.js
zahnno: I was juggling between the two but I'll take that as a vote for if instead of on. :+1:

kgrgreer (commented):
LGTM

src/foam/u2/view/RichChoiceView.js
TW80000: I tested and can confirm that with these changes, the order of the sections depends on how long it takes for the count selection to resolve. I don't think it should work that way. I think the order of the sections should depend on the order that they were given in the code.

I think we should be able to use `Promise.all()` to fix this.

lchanmann:
Nice job!

TW80000 (approved):

PR #1802, opened 2019-01-17
Refactor logout disabled user DAO

lchanmann:
Needed for: https://github.com/nanoPayinc/NANOPAY/pull/5354/commits/fcf26afeaa707cf5d30756622f8319622c571081

Refactor: extract isDisabled method

Decoupling the logic for checking if user is disabled allows descendant class to override and define its own definition of a disabled user to be logged out. For example, below DAO will randomly logout user with a probability of 0.1.

```javascript
foam.CLASS({
  name: 'DummyDAO',
  extends: 'foam.nanos.auth.LogoutDisabledUserDAO',

  methods: [
    {
      name: 'isDisabled',
      javaCode: `
        Random rnd = new Random();
        return rnd.nextInt(10) == 0;
      `
    }
  ]
});
```

### DIFF
--- a/src/foam/nanos/auth/LogoutDisabledUserDAO.js
+++ b/src/foam/nanos/auth/LogoutDisabledUserDAO.js
@@ -20,11 +20,7 @@ foam.CLASS({
         User newUser = (User) obj;
         User oldUser = (User) getDelegate().find(newUser.getId());
 
-        if (
-          oldUser != null
-          && oldUser.getEnabled()
-          && !newUser.getEnabled()
-        ) {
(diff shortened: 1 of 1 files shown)


jlhughes:
refactory isDisabled - this would suggest we have this now, is that the case? 
What service/model has this?

jlhughes:
Please do the following (recap)
Remove EnabledAware on User and replace all uses with the isEnabled method. 
I recommend having isEnabled take a context especially if you are expecting other future implementations.

lchanmann:
LogoutDisabledUserDAO class is removed in https://github.com/foam-framework/foam2/pull/1807. This PR is no longer relevant.

PR #1805, opened 2019-01-18
Table Sorting Icon

blakegreen32:
Added correct table sorting icon according to zeplin mocks, icon only displays if column has a label. Also took out close button from modal header as all modals now have an outer close button.


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -185,17 +178,17 @@ foam.CLASS({
           add(this.slot(function(columns_) {
             return this.E('tr').
               forEach(columns_, function(column) {
+                console.log(column.label);
```
kennyqiyenkan: console.log

kennyqiyenkan (dismissed):

src/foam/u2/ModalHeader.js
```diff
@@ -79,8 +79,7 @@ foam.CLASS({
 
     this
       .addClass(this.myClass())
-      .start().addClass('title').add(this.title).end()
-      .start(this.CLOSE_MODAL).addClass('close').end();
```
Henry0422: It is not a good idea remove this `Close` button, it will affect all the app which is using the `ModalHeader.js`

Henry0422 (changes requested):

Henry0422 (changes requested):
Please double check the code in ModalHeader.js before merge..

TW80000 (changes requested):
According to the design, there should be an 8px margin between the column heading and the icon. Please add that in. Something like

```CSS
^ th > img {
  margin-left: 8px;
}
```

should work. I noticed that we're also currently adding a space after the label. That leads to too much space between the label and the icon, so the space should be removed in favour of the 8px left-margin on this icon.

TW80000 (approved):
Tested and it LGTM, thanks!

Henry0422:
I tested this PR as well. Looks good.

Henry0422 (changes requested):
Sorry, one more thing. Can I ask if we can customize the generate the CSV file using `csvDriver`?
Right now, I got some information messed up in the generated CSV file. 
It might only happen to me?

PR #1810, opened 2019-01-20
Cleanup fdescribe

lchanmann:


jlhughes (approved):

lchanmann:
Failing tests because https://github.com/foam-framework/foam2/issues/1812

lchanmann:
Thanks @mcarcaso for helping on troubleshooting issues with `foam.register` and `__context__`.

lchanmann:
@kgrgreer @TW80000 and @mcarcaso please take a look when you a chance.

test/src/box/mixed.js
```diff
@@ -19,7 +19,7 @@ describe('mixed registry types', function() {
   // TODO(markdittmer): Come up with a non-manual test that will confirm all
   // registries clean up properly.
 
-  it('should appropriately manage mixed registry types', function(done) {
+  xit('should appropriately manage mixed registry types', function(done) {
```
TW80000: Could you please put a comment in that explains why this test is disabled? Same for the test in `pipeline.js`.

test/src/core/stdlib.js
```diff
@@ -893,7 +897,7 @@ describe('foam.Object', function() {
 
 });
 
-describe('foam.mmethod', function() {
+describe('foam.method', function() {
```
TW80000: Maybe you thought `mmethod` was a typo? That's the actual name of the method being tested though, it contains two m's. Unless I'm missing something here.

test/src/dao/EnabledAwareDAO.js
```diff
@@ -4,7 +4,7 @@
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
-fdescribe('EnabledAwareDAO', function() {
+describe('EnabledAwareDAO', function() {
```
TW80000, line 7: Just want to make sure I understand what's happening here. Was this `fdescribe` causing other tests not to execute?

test/src/core/stdlib.js
```diff
@@ -841,12 +845,12 @@ describe('foam.Object', function() {
...
   it('clone', function() {
     var a = { d: "hello" };
-    expect(foam.Object.clone(a)).toBe(a);
+    expect(foam.Object.clone(a)).toEqual(a);
   });
   it('equals', function() {
     var a = { d: "hello" };
     var b = { d: "hello" };
-    expect(foam.Object.equals(a, b)).toBe(false);
+    expect(foam.Object.equals(a, b)).toBe(true);
```
TW80000, line 853: We should add another test for 'is' like this:

```JavaScript
  it('is', function() {
    var a = { d: "hello" };
    var b = { d: "hello" };
    expect(foam.Object.is(a, b)).toBe(false);
    expect(foam.Object.is(a, a)).toBe(true);
    expect(foam.Object.is(a, null)).toBe(false);
  });
```

src/foam/core/stdlib.js
```diff
@@ -827,7 +827,23 @@ foam.LIB({
...
     },
-    function equals(a, b) { return a === b; },
+    function keys(o) { return Object.keys(o); },
+    function equals(a, b) {
+      if ( foam.Object.is(a, b) ) return true;
+      if ( ! foam.Object.isInstance(a)
+        || ! foam.Object.isInstance(b) ) return false;
+      if ( this.keys(a).length !== this.keys(b).length ) return false;
+
+      for ( var key in a ) {
```
TW80000, line 837: This won't check enumerable properties. I'm not sure if we want to check those or not anyway though. If we did, we could use [getOwnPropertyNames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames).

Just another note in case it matters, by checking `hasOwnProperty` in the loop, we're only going to be looking at properties owned directly by the objects. Which means we won't be checking any properties that are inherited through the prototype chain. Again, I'm not sure whether we want to or not, but I wanted to point it out in case that matters.

TW80000 (commented):
Mostly LGTM. Just a few comments/questions to make sure I understand the requirements/intentions and a few small requested changes.

test/src/box/mixed.js
lchanmann: ForkBox was removed in the pressure-wash cleanup. Maybe we should also remove the tests that use it.

test/src/core/stdlib.js
lchanmann: Oh yeah, you're right.

test/src/dao/EnabledAwareDAO.js
lchanmann: Yes and the CI only runs this describe for the quite some times. We were trying to fix it last time but the object cloning was in the way #1812.

src/foam/core/stdlib.js
lchanmann: Good point. @kgrgreer, do you think we need to account for non-enumerable properties when comparing objects using `foam.Object.equals`?

kgrgreer: We're really only using this for map-like objects, so I think this is fine. Following prototype inheritance could involve comparing the whole system.

test/src/box/mixed.js
lchanmann: Removed.

lchanmann:
All comments are addressed. Thanks, @TW80000 for reviewing.

mcarcaso (approved):
Sorry for being a little late to the party here but this is looking good to me. Thanks for cleaning this all up!

test/src/core/stdlib.js
```diff
@@ -743,7 +743,11 @@ describe('foam.Date', function() {
     expect(foam.util.equals(new Date(7487474), new Date(23423432))).toBe(false);
 
     expect(foam.util.equals((new Date(7487474), null))).toBe(false);
-    expect(foam.util.equals(new Date(7487474), 7487474)).toBe(true);
+    // NOTE: In pressure-wash mass cleanup, foam.util.equals was modified to
+    //       only compare objects of the same type and returns false otherwise.
+    //       Question: Do we want to preserve the behavior of comparing
+    //                 a date to a number?
+    expect(foam.util.equals(new Date(7487474), 7487474)).toBe(false);
```
kgrgreer, line 750: I think this should be false.

PR #1813, opened 2019-01-21
Refactor agent auth service

deejes:
Expose canActAs in the AgentAuthService interface.

This allows logic in subclasses of AgentUserAuthService to be accessed via the service in the context.


src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -42,6 +42,27 @@ foam.INTERFACE({
         }
       ]
     },
+    {
+      name: 'canActAs',
+      javaReturns: 'boolean',
+      swiftReturns: 'Bool',
+      swiftThrows: true,
```
zahnno, line 54: Add an array of exceptions that might be thrown in javaThrows

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -42,6 +42,27 @@ foam.INTERFACE({
         }
       ]
     },
+    {
+      name: 'canActAs',
+      javaReturns: 'boolean',
```
kgrgreer, line 48: Add documentation.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -54,35 +55,16 @@ public User actAs(X x, User entity) throws AuthenticationException {
...
+      return null;
     }
 
-    /*
-      Finds the UserUserJunction object to see if user can act as the passed in user.
-      Source (agent) users are permitted to act as target (entity) users, not vice versa.
-    */
     UserUserJunction permissionJunction = (UserUserJunction) agentJunctionDAO_.find(AND(
-      EQ(UserUserJunction.SOURCE_ID, agent.getId()),
+    EQ(UserUserJunction.SOURCE_ID, agent.getId()),
```
zahnno: Indentation is off.

src/foam/nanos/auth/AgentAuthService.js
deejes: Done

zahnno:
Looks fine to me other than the minor comments

src/foam/nanos/auth/AgentAuthService.js
deejes: @zahnno will talk to you about this, have a few questions.

deejes: Sorry, super redundant comment, second only to this one.

deejes: Done

deejes:
@kgrgreer any other comments/concerns? 
I know you told me to change the name, I have that as a task for a braindead day.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -42,6 +42,32 @@ foam.INTERFACE({
         }
       ]
     },
+    {
+      name: 'canActAs',
+      documentation: `returns a boolean indicating whether the provided agent has adequete permission to actAs the provided entity`,
```
TW80000: Typos:

```suggestion
      documentation: `Returns a boolean indicating whether the provided agent has adequate permission to act as the provided entity.`,
```

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -100,6 +82,45 @@ public User actAs(X x, User entity) throws AuthenticationException {
...
+      }
+      
+      /*
+      Finds the UserUserJunction object to see if user can act as the passed in user.
+      Source (agent) users are permitted to act as target (entity) users, not vice versa.
+      */
+      UserUserJunction permissionJunction = (UserUserJunction) agentJunctionDAO_.find(AND(
+        EQ(UserUserJunction.SOURCE_ID, agent.getId()),
+        EQ(UserUserJunction.TARGET_ID, entity.getId())
+        ));
```
TW80000: The indentation is one level too deep starting on this line and ending on line 116.

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -84,6 +85,7 @@ public Session authenticate(X x) {
     // query parameters
     String              email        = req.getParameter("user");
     String              password     = req.getParameter("password");
+    String              entityId        = req.getParameter("entityId");
```
TW80000:
```suggestion
    String              entityId     = req.getParameter("entityId");
```

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -121,7 +123,7 @@ public Session authenticate(X x) {
 
     //
     // Support for Basic HTTP Authentication
-    // Redimentary testing: curl --user username:password http://localhost:8080/service/dig
+    // Redimentary testing: curl --user username:password http://localhost:8080/service/dig?entityId=1234
```
TW80000: Typo:

```suggestion
    // Rudimentary testing: curl --user username:password http://localhost:8080/service/dig?entityId=1234
```

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -168,8 +170,19 @@ public Session authenticate(X x) {
           .put(HttpServletResponse.class, resp), email, password);
 
         if ( user != null ) {
+          // if user is attempting to and (Can)actAs another entity, set the entity in session context
```
TW80000:
```suggestion
          // If user is attempting to and can act as another entity, set the entity in session context.
```

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -168,8 +170,19 @@ public Session authenticate(X x) {
...
+          if ( ! SafetyUtil.isEmpty(entityId) ) {
+            AgentAuthService agentService = (AgentAuthService) x.get("agentAuth");
+            DAO localUserDAO = (DAO) x.get("localUserDAO");
+            User entity = (User) localUserDAO.find(Long.parseLong(entityId));
+            if (agentService.canActAs(x, user, entity)) {
```
TW80000: Missing spaces

```suggestion
            if ( agentService.canActAs(x, user, entity) ) {
```

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -168,8 +170,19 @@ public Session authenticate(X x) {
...
+              // set agent in session
+              session.setContext(session.getContext().put("entity",entity));
```
TW80000:
```suggestion
              session.setContext(session.getContext().put("entity", entity));
```

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -168,8 +170,19 @@ public Session authenticate(X x) {
...
+            }
+          }
           return session;
-        } else {
+          }
```
TW80000: This was correct before.

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -201,7 +214,11 @@ public void execute(X x) {
 
     if ( session != null && session.getContext() != null && session.getContext().get("user") != null ) {
       if ( auth.check(session.getContext(), permission_) ) {
-        getDelegate().execute(x.put(Session.class, session).put("user", session.getContext().get("user")));
+        if (session.getContext().get("entity") != null) {
```
TW80000:
```suggestion
        if ( session.getContext().get("entity") != null ) {
```

TW80000 (changes requested):

src/foam/nanos/http/AuthWebAgent.java
deejes: Sorry, not sure what you mean by this @TW80000 . Fixed the rest of the styling issues.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -42,6 +42,32 @@ foam.INTERFACE({
         }
       ]
     },
+    {
+      name: 'canActAs',
+      documentation: `Returns a boolean indicating whether the provided agent has adequete permission to actAs the provided entity.`,
```
TW80000: Typo: Should be `adequate`

TW80000 (changes requested):

TW80000 (approved):

nanoscott (approved):

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -100,6 +82,45 @@ public User actAs(X x, User entity) throws AuthenticationException {
     return user;
   }
 
+  public boolean canActAs(X x, User agent, User entity) {
+    try {
+      Group group = (Group) groupDAO_.find(entity.getGroup());
+      
+      if ( group == null ) {
+        throw new AuthorizationException("Entity must exist within a group.");
```
adamvy, line 90: Why are these exceptional cases rather than just returning false for "no you can't act as this non existant group, or no you can't act as a group that is disabled?"

deejes: I'm honestly not sure  - @zahnno can you take this?

If that would be a better way to do it, happy to. I'm honestly not sure what the trade-offs are.

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -201,7 +213,11 @@ public void execute(X x) {
 
     if ( session != null && session.getContext() != null && session.getContext().get("user") != null ) {
       if ( auth.check(session.getContext(), permission_) ) {
-        getDelegate().execute(x.put(Session.class, session).put("user", session.getContext().get("user")));
+        if ( session.getContext().get("entity") != null ) {
```
adamvy: This is confusing.

I think it would be more consistent if we _always_ had x.user as the actual user who made the request, and x.effectiveUser as the agent/user that they are acting as.

In the case where they did not request to act as somebody else, then effectiveUser == user, and all code which needs to check permissions always checks effectiveUser.

Fewer conditionals is always better, especially when it comes to security.

I'm guessing that most of our code currently checks X.user for authentication?  IF we dont' want to go update all that code, we could add X.realUser as the actual user who made the request, and treat X.user as the effective user they are acting as.  We can do that work in authenticate() and this line can just be

x.put("user" session.getcontext().get("user").put("realUser", session.getContext().get("realUser")), no need for this separate "entity" key in the context, just set user/realUser in the session context appropriately in authenticate()

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -121,7 +123,7 @@ public Session authenticate(X x) {
 
     //
     // Support for Basic HTTP Authentication
-    // Redimentary testing: curl --user username:password http://localhost:8080/service/dig
+    // Rudimentary testing: curl --user username:password http://localhost:8080/service/dig?entityId=1234
```
adamvy, line 143: I don't know if there is a requirement I'm missing here, but entityId feels like a poor name for this parameter, something like "actAs=1234" or (su= or doas= to steal names from unix/openbsd)

deejes: Okay, will make the change, thanks!

src/foam/nanos/http/AuthWebAgent.java
jlhughes: @adamvy This particular point has been debated back and forth, eventually deciding on X.user for authentication and X.agent for the agent the user is acting on behalf of. 
@deejes Please make use of Adam's simplified context setup.

src/foam/nanos/auth/AgentUserAuthService.java
adamvy: If we do remove the exceptions, we can remove them from the Interface declaration as well.

It seems easier to code against if its just a true/false value.  I think the key question is, is the client ever going to need to behave differently for cases where the group is disabled/doesn't exist as opposed to just not being a member of the appropriate group?  Most of the time I expect the client not to care, but I can imagine in a debugging scenario a more detailed error message would be useful.

Perhaps that's handled better by an error on the actual request though, if I try to make a request acting as user XYZ, it would be helpful to know that the request failed because there is no group XYZ, or group XYZ is disabled, meaning i likely have a configuration error or I picked the wrong group, as opposed to not being a member of group XYZ.

Upon further consideration, I'm not even sure that we should return an false if the group is disabled.  It seems reasonable to me to have canActAs return true, if I'm a member of the appropriate group, but to then have my requests fail later because the group is disabled.  If we ever add more states to whether a group is allowed to do perform actions we're not going to want to come back and update this method as well, I think this method should really just be telling me (You should be able to make requests on behalf of user XYZ, whether or not those will succeed due to the group being disabled or over quota is a different question)

src/foam/nanos/http/AuthWebAgent.java
adamvy: Or to put it different.  I can act as an account that is disabled, but in doing so, a disabled account probably won't be able to do anything.  The disabled state is not a factor on whether I can act as them or not.

adamvy: @jlhughes Sounds good, those names are reasonable.  Simplified context setup is the more important goal, rather than complex conditionals.

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: > Why are these exceptional cases rather than just returning false for "no you can't act as this non existant group, or no you can't act as a group that is disabled?"

Just to clarify, you're not acting as a group, you're acting as another User (or in Ablii's case a Business, which extends the User model).

One of the requirements of Ablii was that users (people) should be able to be part of multiple businesses. The way we implemented that was allowing `User`s to "act as" other `User`s. 

Another requirement was that users will need different permissions for different businesses. We implemented that by storing a group id on each junction between a user and a business. That group becomes the "effective" user's group while acting as a business. The junction group is the effective group, not the `User`'s group and not the `Business`'s group.

We considered it exceptional to have an invalid group on a junction when trying to act as another User because that's the group that's the "effective" group, the group that's going to be checked by the auth service. So if the group referenced on the junction doesn't exist, then you're effectively not going to have a group.

I think your last paragraph is reasonable though, I just wanted to clarify the requirements we were working under and how it currently works.

src/foam/nanos/http/AuthWebAgent.java
TW80000: > I think it would be more consistent if we always had x.user as the actual user who made the request, and x.effectiveUser as the agent/user that they are acting as.

Two things:

> and x.effectiveUser as the agent/user that they are acting as

The **entity** is the User being acted as. The **agent** is the User who is acting as the entity.

In the context, if a user is being acted as, then `x.get("user")` is the entity and `x.get("agent")` is the agent. If a user is not being acted as, then `x.get("user")` is just the user and `x.get("agent")` is null.

> x.user as the actual user who made the request

Who are you referring to by "actual user" here? The agent or the entity, using the definitions above?

In Ablii, we want to be doing everything as the Business, which is the entity, which is `x.get("user")`. That's because the id is used in many places such as relationships and authentication code. We want you to see the Business's invoices, contacts, and otherwise, not the user's.

I'm not sure how that translates to using the API, but I wanted to clarify our intentions when we designed it.

> IF we dont' want to go update all that code, we could add X.realUser as the actual user who made the request, and treat X.user as the effective user they are acting as.

We didn't want to update all that code, and that's exactly what we did. Just replace what you call `X.realUser` with `X.agent`. `X.user` is the business, the user you're acting as, and `X.agent` is you, which is rarely accessed.

src/foam/nanos/http/AuthWebAgent.java
TW80000: I agree that `entity` was a poor name for the User being acted as because it's too general. But that should at least clarify why @deejes named it `entityId`.

src/foam/nanos/http/AuthWebAgent.java
adamvy: I appreciate the explanation, I should clarify that when I say that I find the code "confusing" it doesn't necessarily mean that am I actively "confused" but rather that I think that the code will likely lead to confusion of the next developer who comes along in 3 months and tries to understand it :).

> The entity is the User being acted as. The agent is the User who is acting as the entity.

My issue with that naming is that "entity" is a very general term, and we now have 3 names to describe 2 things.  We have "entity" "agent" and "user" but we only really use "agent" and "user"

My main objection/feedback on the PR is not the specific names we use (although I think there are clearer options) but that the code in this PR to construct the context uses a lot of contidionals.  Currently the code sets "entity" in a context in one place, then later pulls it out and uses it as a conditional on whether to put "agent" and "user" or just "user" in the context is unnecessary complication.  It would be safer and simpler to just always put "user" and "agent" in the context, then we never need to refer to "entity" in the context, which is a far too general name for this use case.

src/foam/nanos/http/AuthWebAgent.java
kgrgreer: doas seems too similar to daos for me

src/foam/nanos/http/AuthWebAgent.java
kgrgreer: I agree that the naming is confusing. Agent, entity, user all seem synonymous. Something along the lines of realUser and effectiveUser, and then just always populating both values would be easier to understand.

deejes:
being addressed by #1845

PR #1815, opened 2019-01-22
Fix bug where properties were not being looked up properly

TW80000:
The lowercase property name should be used in the permission string, not in the `getProperty` lookup.

@Henry0422 

Marks https://github.com/nanoPayinc/NANOPAY/issues/5755 as ready for testing


Henry0422:
Cool.

TW80000:
Please note, the actual error was originating from [line 185](https://github.com/foam-framework/foam2/pull/1815/files#diff-0ba4a26a89b694dfe1a848a961c815f6R185), where the `axiomName` variable was being passed to `getProperty`. This was causing `getProperty` to return `null` for properties where the name was moved to lowercase because the name was wrong. For example, `getProperty` expects `"emailVerified"`, not `"emailverified"`.

PR #1816, opened 2019-01-22
Add EnabledAwareDummy for tests

lchanmann:
EnabledAwareDAOTest should not depend on User class to be a EnabledAware
type. Using EnabledAwareDummy in tests make EnabledAwareDAOTest isolated
from changes happening in classes that implement EnabledAware e.g., User.


jlhughes (approved):
@adamvy @mcarcaso let us know if you are ok with the Dummy test model for this case.

kgrgreer:
Move to 'test' sub-package.

PR #1826, opened 2019-01-23
Update RichChoiceView to support being disabled

### LINKED ISSUE #1824: RichChoiceView does not support the "Disabled" visibility mode

TW80000:
Closes #1824

### DIFF
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -232,8 +232,11 @@ foam.CLASS({
         .addClass(this.myClass())
         .start()
           .addClass(this.myClass('selection-view'))
-          .on('click', function() {
-            self.isOpen_ = ! self.isOpen_;
+          .enableClass('disabled', this.mode$.map((mode) => mode === foam.u2.DisplayMode.DISABLED))
+          .callIf(this.mode === foam.u2.DisplayMode.RW, function() {
+            this.on('click', function() {
+              self.isOpen_ = ! self.isOpen_;
+            });
           })
           .start()
             .addClass(this.myClass('custom-selection-view'))
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -233,10 +233,10 @@ foam.CLASS({
         .start()
           .addClass(this.myClass('selection-view'))
           .enableClass('disabled', this.mode$.map((mode) => mode === foam.u2.DisplayMode.DISABLED))
-          .callIf(this.mode === foam.u2.DisplayMode.RW, function() {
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/RichChoiceView.js
```diff
@@ -232,8 +232,11 @@ foam.CLASS({
         .addClass(this.myClass())
         .start()
           .addClass(this.myClass('selection-view'))
-          .on('click', function() {
-            self.isOpen_ = ! self.isOpen_;
+          .enableClass('disabled', this.mode$.map((mode) => mode === foam.u2.DisplayMode.DISABLED))
+          .callIf(this.mode === foam.u2.DisplayMode.RW, function() {
```
adamvy: There's a but here.

Using callIf like this means that if mode was DISABLED when we first rendered, then the click listener won't be hooked up, but if we later change mode to ENABLED, the click listener will still not be attached.  It's not dynamic.

TW80000: Yeah, I wasn't sure how to get around this. I know using an Action would let us use `isEnabled`, but the combination of trying to set the content of the action to the custom `selectionView` as well as dealing with overriding all of the default action styles made me want to avoid using an action. Should I bite the bullet and try to get it to work as an action or is there an easy alternative to using `callIf` the way it already is? I guess I could use a slot and conditionally put the on click listener in the callback, right?

adamvy: You could do that, but it seems simpler to just move your conditional logic into the click listener.  Rather than jump through hoops to prevent attaching the listener when we might not need it, just always attach the listener and add logic to your event handler to take appropriate action based upon the current state.

TW80000: Well that makes a lot more sense, haha. Thanks, updated.

jlhughes (approved):

src/foam/u2/view/RichChoiceView.js
```diff
@@ -232,8 +232,11 @@ foam.CLASS({
         .addClass(this.myClass())
         .start()
           .addClass(this.myClass('selection-view'))
+          .enableClass('disabled', this.mode$.map((mode) => mode === foam.u2.DisplayMode.DISABLED))
           .on('click', function() {
-            self.isOpen_ = ! self.isOpen_;
+            if ( self.mode === foam.u2.DisplayMode.RW ) {
```
adamvy, line 237: It would also be worth at a later date adding support such that if the view transitions to DISABLED state while its currently open, then it automtically gets closed.

TW80000: https://github.com/foam-framework/foam2/pull/1827

PR #1836, opened 2019-01-25
Check property permission

KateMengjunXia:
In DetailView, check if user has read / write permissions to properties that require them, and change property visibilities accordingly.

### DIFF
--- a/src/foam/u2/DetailPropertyView.js
+++ b/src/foam/u2/DetailPropertyView.js
@@ -22,6 +22,10 @@ foam.CLASS({
 
   documentation: 'View for one row/property of a DetailView.',
 
+  imports: [
+    'auth'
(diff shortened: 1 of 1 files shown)


src/foam/u2/DetailPropertyView.js
```diff
@@ -54,9 +58,24 @@ foam.CLASS({
   `,
 
   methods: [
-    function initE() {
+    async function initE() {
       var prop = this.prop;
 
+      if ( prop && prop.permissionRequired )  {
```
kgrgreer: Does this mean that we won't display fields where permission is not required?

jlhughes (approved):

jlhughes (approved):

PR #1837, opened 2019-01-25
changing network retry attempts on timeout

Nauna:
closes: https://github.com/nanoPayinc/NANOPAY/issues/5844

### DIFF
--- a/src/foam/box/RetryBox.js
+++ b/src/foam/box/RetryBox.js
@@ -103,7 +103,7 @@ foam.CLASS({
     {
       name: 'maxAttempts',
       documentation: 'Set to -1 to infinitely retry.',
-      value: 3
+      value: 0
     }
   ],
 


src/foam/box/RetryBox.js
```diff
@@ -103,7 +103,7 @@ foam.CLASS({
     {
       name: 'maxAttempts',
       documentation: 'Set to -1 to infinitely retry.',
-      value: 3
+      value: 0
```
adamvy, line 106: Would be better to remove RetryBox from ClientBuilder.js for the time being, and leave a comment there explaining why, because we want to turn this back on when our Exceptions from the server are better categorized.

adamvy: Looks like the retry is actually done in EasyDAO, we could remove it from there, and/or add retryAttempts as an EasyDAO parameter and set that to 0 in ClientBuilder, either is fine with me.

adamvy: or even retry: true/false to EasyDAO would work.

Nauna:
after talking with @mcarcaso and @adamvy this isn't the best solution.

Another implementation is being developed

PR #1838, opened 2019-01-25
DO_NOT_MERGE Create session with system context

deejes:
1. When creating a new session object, call getX() instead of emptyX()
2. Set user to null to avoid leaking of system user
3. Pass system context when calling new Session()


adamvy:
I'm concerned that this approach will lead to information leaks or security problems later.  As we'll have to remember to update Session.js to plug any information from the system context we don't want going forward, like if we add "agent" globally for instance.

I'll have to do a bit more research on how we use the Session context but I feel like a better approach would be to stop storing Context as a property on the Session and instead introduc t amethod called "augmentSession(x)" which takes an existing context and adds in the session information such as "user" and "agent."

Let me know if that looks reasonable.  I'm have to do a bit more reading to see everywhere we use sessions myself as well to see.

kgrgreer:
Adam, we had to remember to block global values either way. The real solution is to create an explicit SystemContext which is a private sub-Context of the globalContext, then there wouldn't be anyway for values to leak out of it.

adamvy:
If the Session's context now contains all services do we still need to copy the "user" into another context in AuthWebAgent.execute()?  This feels very redundant.

adamvy:
I guess we need some context from the request so that reply boxes know where to send their values, but if we are doing that then why do we need to have the system context in the session's context?

adamvy:
I'm probably missing context, why do we need this change at all?

kgrgreer:
No, we don't need to copy values like user anymore, which is what @deejes was fixing in the first place. He has that in other changes he's made but not submitted yet.

adamvy:
Okay that makes sense, it's probably better if I review the changes together than, or at least start the PR description with an indication of why we're making the change.

If we're centralizing context creations for sessions and requests in the session object then I think those changes should all be committed at once, as it stands this change adds an additional spot where we have to make sure not to leak sensitive context values.

deejes:
What are sensitive context values? If the list is static we can exclude them on session creation. 

Will merge this with my other work, and make a combined pr.

adamvy:
> What are sensitive context values? If the list is static we can exclude them on session creation.

"user" and "agent" are definitely sensitive.  I don't know of others, but those appear to be the only ones that we are sure to mask currenty.

deejes:
being addressed by #1845

PR #1843, opened 2019-01-28
Table Sorting Icon Polishing

blakegreen32:
Added new table sorting icons for resting, ascending and descending. Added functionality to display correct images as user is sorting the table. User experience is much better with a resting sorting icon, it will be obvious to the user that they can use the table sorting functionality.


kgrgreer:
Hi Blake, I'll come to your desk tomorrow to have a look at the running code. The code itself looks good.

kgrgreer:
Looks good, but I think the resting icon should be less visible. This would provide greater contrast.

PR #1844, opened 2019-01-28
Session expiration based on user group

sirenChen:
1. Add a session soft limits in the group, so we can configure it per group.
2. Initial a session timer in ApplicationController.
3. Refresh the session timer in the SessionClientBox.
4. Configure the onSessionTimeout callback and modal in Nanopay, see [here](https://github.com/nanoPayinc/NANOPAY/pull/5939)

refs: https://github.com/nanoPayinc/NANOPAY/issues/5756

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -120,7 +120,16 @@ foam.CLASS({
     },
     {
       class: 'Long',
-      name: 'softSessionLimit'
+      name: 'softSessionLimit',
+      documentation: `
+        Users will be asked for providing a feedback once the soft session limit has been reached.
+        If the user doesn't provide any feedback, system will force the user logout.
+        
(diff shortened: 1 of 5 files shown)


jlhughes (approved):

src/foam/nanos/auth/Group.js
```diff
@@ -118,6 +118,10 @@ foam.CLASS({
       name: 'replyTo',
       value: null
     },
+    {
+      class: 'Long',
+      name: 'softSessionLimit'
```
jlhughes: Please document this. regarding how it's used and what values are appropriate and in what units - milliseconds, seconds.  And what values are appropriate like 1 minute.   And how it interacts with Session limits.

sirenChen: Documentation added

PR #1845, opened 2019-01-28
Refactor auth web agent and session context

deejes:
1. Change session to create a context with all the props of the passed in context, but null user

2. Refactor AuthWebAgent


src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -100,6 +82,45 @@ public User actAs(X x, User entity) throws AuthenticationException {
     return user;
   }
 
+  public boolean canActAs(X x, User agent, User entity) {
```
adamvy, line 85: Is there a better noun than "entity" we can use?  entity is very general, perhaps a name like "requestedUser" or "targetUser"

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -100,6 +82,45 @@ public User actAs(X x, User entity) throws AuthenticationException {
...
+    try {
+      Group group = (Group) groupDAO_.find(entity.getGroup());
+      
+      if ( group == null ) {
+        throw new AuthorizationException("Entity must exist within a group.");
+      }
+      
+      if ( ! group.getEnabled() ) {
```
adamvy, line 93: I'm curious about this error case.  Is there any harm in continuing even if the group is disabled?  Presumably the security should be enforced at another level such that a disabled group cannot perform any actions.  Whether or not I can act as a target group should be true regardless of whether that target group is enabled or not, it just so happens that by acting as a disabled group I can't actually perform any actions.

deejes: Going to ask @zahnno and @TW80000 to take this one

src/foam/nanos/auth/AgentUserAuthService.java
deejes: I agree - does agency make sense?  feel like both "requestedUser" and "targetUser" still aren't specific enough, and if we are calling the user who is acting as someone else the agent, agency might make sense?

@TW80000 @zahnno

src/foam/nanos/session/Session.js
```diff
@@ -55,7 +55,8 @@ foam.CLASS({
       class: 'Object',
       name: 'context',
       javaType: 'foam.core.X',
-      javaFactory: 'return foam.core.EmptyX.instance().put(Session.class, this);',
+      // Put a null user to prevent sytem user from leaking into subcontexts
+      javaFactory: 'return getX().put("user", null).put(Session.class, this);',
```
adamvy, line 59: Let's null out "agent" as well.  I think both "agent" and "user" should always be set.

src/foam/box/SessionServerBox.java
```diff
@@ -42,7 +42,7 @@ public void send(Message msg) {
         Session            session    = (Session)     sessionDAO.find(sessionID);
 
         if ( session == null ) {
-          session = new Session();
+          session = new Session(getX());
           session.setId(sessionID);
           session.setRemoteHost(req.getRemoteHost());
           session.setContext(getX().put(Session.class, session));
```
adamvy, line 48: We can probably remove this line now right? as the context factory in session should build the appropriate session object?

adamvy: I guess not because the session factory will null out the user, hmm.

src/foam/nanos/auth/AgentUserAuthService.java
adamvy: "agency" seems reasonable given the constraints.

src/foam/util/Auth.java
```diff
@@ -30,7 +30,7 @@ public static X sudo(X x, String email) {
   public static X sudo(X x, User user) {
     if ( user == null ) throw new RuntimeException("Unknown user");
 
-    Session session = new Session();
+    Session session = new Session(x);
```
adamvy, line 33: Set "agent" as well.

deejes:
There are a bunch of comments on this code from pr's #1813 and #1838

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: "puppet" & "puppeteer" (joking)
"foregroundUser" & "backgroundUser"
"effectiveUser" & "ineffectiveUser"
"operativeUser" & "inoperativeUser"
"activeUser" & "inactiveUser"
"user" & "coUser"

I agree that agent is a perfect term for the one user, but I can't think of a great word for the other user. I asked on the English language Stack Exchange site [here](https://english.stackexchange.com/questions/483328/what-word-refers-to-a-person-being-acted-on-behalf-of), will let you know if there are any good answers.

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: I agree with you that we should be able to remove this and everything should still work as expected because we should have many checks in many places that will catch it down the line. I think this was added here just because it seemed like a case that would be clearly incorrect for our use case, so we wanted to flag it immediately with a good error message rather than let something else catch it down the line and have to backtrack. I'm not opposed to removing it.

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: According to [this link](https://en.oxforddictionaries.com/definition/principal), **principal** might be the word we're looking for.

PR #1848, opened 2019-01-28
Set default max attempts for RetryBox to 0 for nanos DAOs

mcarcaso:
This should be a temporary band-aid around a bug on nanos misleading the client into thinking it needs to retry transactions that actually completed successfully.

### DIFF
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -321,16 +321,18 @@ return delegate;
       generateJava: false,
       factory: function() {
         // TODO: This should come from the server via a lookup from a NamedBox.
-        return this.SessionClientBox.create({
-          delegate: this.RetryBox.create({
+        var box = this.TimeoutBox.create({
+          delegate: this.remoteListenerSupport ?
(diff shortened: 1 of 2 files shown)


src/foam/dao/EasyDAO.js
```diff
@@ -311,18 +311,26 @@ return delegate;
...
         // TODO: This should come from the server via a lookup from a NamedBox.
-        return this.SessionClientBox.create({ delegate: this.RetryBox.create({ delegate:
-          this.TimeoutBox.create({ delegate:
-          this.remoteListenerSupport ?
-              this.WebSocketBox.create({ uri: this.serviceName }) :
-              this.HTTPBox.create({ url: this.serviceName })
-        })})});
+        return this.SessionClientBox.create({
+          delegate: this.RetryBox.create({
+            maxAttempts: this.retryBoxMaxAttempts,
```
kgrgreer, line 331: If 0, couldn't we just not add the RetryBox?

mcarcaso: Updated

PR #1849, opened 2019-01-29
Added UI support for final

KateMengjunXia:
https://github.com/nanoPayinc/NANOPAY/issues/5290

### DIFF
--- a/src/foam/core/Property.js
+++ b/src/foam/core/Property.js
@@ -529,15 +529,15 @@ foam.CLASS({
               writable: false,
               configurable: true // ???: is this needed?
             });
+
+            if ( ! factory || oldValue !== undefined )  prop.visibility = 'RO';
           }
 
           // If this is the result of a factory setting the initial value,
           // then don't fire a property change event, since it hasn't
           // really changed.
-          if ( ! factory || oldValue !== undefined ) {
-            if ( prop.final ) prop.visibility = 'RO';
+          if ( ! factory || oldValue !== undefined )
             this.pubPropertyChange_(prop, oldValue, newValue);
-          }
 
           // FUTURE: pub to a global topic to support dynamic()
 
--- a/src/foam/core/Property.js
+++ b/src/foam/core/Property.js
@@ -534,8 +534,10 @@ foam.CLASS({
           // If this is the result of a factory setting the initial value,
           // then don't fire a property change event, since it hasn't
           // really changed.
-          if ( ! factory || oldValue !== undefined )
+          if ( ! factory || oldValue !== undefined ) {
+            if ( prop.final ) prop.visibility = 'RO';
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

src/foam/core/Property.js
```diff
@@ -534,8 +534,10 @@ foam.CLASS({
           // If this is the result of a factory setting the initial value,
           // then don't fire a property change event, since it hasn't
           // really changed.
-          if ( ! factory || oldValue !== undefined )
+          if ( ! factory || oldValue !== undefined ) {
+            if ( prop.final ) prop.visibility = 'RO';
```
TW80000: We have an `if ( isFinal ) { ... }` block a few lines above, wouldn't this make more sense there?

KateMengjunXia: I agree. Just updated the code.

kgrgreer (changes requested):
I disagree with this PR.
If a property has final: true that means that you can only set the value once, and then it becomes a constant.
visibility: FINAL means that the value can only be edited in create mode, but becomes RO in edit mode. This should be the defaault for primaryKeys or values like createdBy or createdAt. A final: true property wouldn't be editable in the GUI because the value would get stuck on your first proposed value.

TW80000:
> A final: true property wouldn't be editable in the GUI because the value would get stuck on your first proposed value.

It seems to me like it would make more sense if all `final: true` properties automatically were `visibility: FINAL` as well since they seem to be the same thing in spirit. If our view implementation is setting the property before the form is submitted, then isn't that a problem? When would you want a `final: true` property that wasn't `visibility: FINAL`? @kgrgreer

kgrgreer:
Not the same thing. Say I'm in a GUI and I'm editing a unit test. It has a unique name property which is the primary key and is visibility: final. I might call it test1, then think that isn't a good name and rename it to 'CreatedByAware test', and then rename it finally to 'CreatedByAware creation test'. In this case I update the name property three times, or in on key mode was true then I updated it for each character I typed. If the property were final: true the value would lock up after I set it to test1, or even to just 't' in on key mode. As you're updating the GUI the property value keeps getting updated, but this would be prevented with final: true. final: is final once initially set to any value, visibility: final is final once the object is stored in the DAO. The first one is just used for a few internal high-performance use-cases and isn't at all useful for GUI's.  Java has 'final' variables which is what this feature is emulating.

TW80000:
@kgrgreer Thanks for the explanation. I was arguing that we could change the way our inputs works to not actually set the values for the properties they're there for and leave the actual setting of the property to the form submission. That way the values would only be set once and therefore we could make all `final: true` properties `visibility: FINAL` with no problems, but after thinking about it more I disagree with that argument anyway.

PR #1851, opened 2019-01-29
added double quotes to jsonj keys

nanoNeel:
REFS https://github.com/nanoPayinc/NANOPAY/issues/5334


kgrgreer:
JSON/J already isn't real JSON, so why do we care if it's strict?

jlhughes:
Just trying to be consistent.  All of our hand crafted repo journals use quoted keys.

PR #1852, opened 2019-01-29
updated RulerDAO

smirnovaka:


src/foam/nanos/ruler/RulerDAO.js
```diff
@@ -31,6 +31,36 @@
       class: 'String',
       name: 'daoKey',
       documentation: 'The dao name that rule needs to be applied on.'
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.mlang.predicate.Predicate',
```
kgrgreer: Store the list you get back from doing the select() on the ruleDAO, not the predicate itself. This code still performs the dao query every time you call applyRules. Pass the list to applyRules(), not the predicate that it would use to find the rules.

PR #1853, opened 2019-01-29
Permissions Blacklist

blakegreen32:
Added Blacklist model with properties id, description and entityType. entityType is an Enum of type BlacklistEntityType which only has 2 values USER and BUSINESS for differentiating between different entity types. We can add new Entity types as we wish. Also added blacklistDAO so we can add and configure Blacklists in the GUI.


jlhughes:
Why are going with Enums?  Why not just models: we have User and Business models.  Then we don't have to deploy/update to support other models.  Also Business does not exist on the foam side.

PR #1855, opened 2019-01-31
Add NSpec class to session context

lchanmann:
Needed for: https://github.com/nanoPayinc/NANOPAY/pull/5930/commits/85a7a7e8ebd1a30b105b73d05d9bd3709fb577f2#diff-db975485d409dd9b601ee43def875fffR63.

This allows access to currently running DAO service. E.g., getting dao name when inside a decorator

```
String daoName = x.get(NSpec.class).getName();
```

### DIFF
--- a/src/foam/box/SessionServerBox.java
+++ b/src/foam/box/SessionServerBox.java
@@ -58,7 +58,8 @@ public class SessionServerBox
             new PrefixLogger(
                 new Object[] { user == null ? "" : user.getId() + " - " + user.label(), "[Service]", spec.getName() },
                 (Logger) session.getContext().get("logger")))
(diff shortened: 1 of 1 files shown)


kgrgreer:
I don't think this is always useable for what you're trying to do. This will only set the NSpec if you access it first from a remote client, but not if accessed from the server directly from a test or cron job. Better to have the NSPecFactor populate the NSPec.

kgrgreer (changes requested):
See comment in conversation.

lchanmann:
https://github.com/foam-framework/foam2/commit/e49d605c2e3631de84e7917e81566cca5a2b7bdf is better. Thanks @kgrgreer

PR #1858, opened 2019-01-31
Update service configuration when service record changed

lchanmann:
An NSpec service can have service configuration however updating the service configuration from the UI does not change the service object already in context. For example, to enable the below service

screen shot 2019-01-31 at 1 35 34 pm

It requires server restart for the change to take place which is inconvenient. Instead, we can listen for changes on nSpecDAO and update the service configuration accordingly.

### DIFF
--- a/src/foam/nanos/boot/Boot.java
+++ b/src/foam/nanos/boot/Boot.java
@@ -6,10 +6,7 @@
 
 package foam.nanos.boot;
 
-import foam.core.Detachable;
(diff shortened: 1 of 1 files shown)


jlhughes:
Woo hoo. Thanks. This has been on the TODO for a long time.

kgrgreer:
I think there are some limitations with this that we should improve on in future updates. It doesn't let you change the class of the service or use the serviceScript feature to redo decoration.

PR #1859, opened 2019-01-31
Disable Auth.sudo in PRODUCTION

jlhughes:
CPF-115 #time 1h

Review, but presently I'm unable to get the Test case to run.  Will fix that before we merge.


TW80000:
These are the places we use `Auth.sudo` aside from test code that I could find:

screen shot 2019-01-31 at 2 16 07 pm

Until we fix all of these places, I don't think we can merge this without breaking things.

TW80000 (changes requested):
These are the places we use `Auth.sudo` aside from test code that I could find:

screen shot 2019-01-31 at 2 16 07 pm

Until we fix all of these places, I don't think we can merge this without breaking things.

PR #1861, opened 2019-01-31
Prevent invalid date in DateView

Henry0422:
Fixed the issue that DateView accept `2019/02/90` issue using **Safari**.


src/foam/u2/DateView.js
```diff
@@ -34,19 +38,36 @@ foam.CLASS({
...
+  listeners: [
+    function checkDateFormat(value) {
+      oldDate = value.split(/[^0-9]/).map((str) => {
+        return parseInt(str, 10);
+      }).join('');
+
+      var newDate = new Date(oldDate);
+      newDate = `${newDate.getFullYear()}${newDate.getMonth() + 1}${newDate.getDate()}`;
+      if ( newDate !== oldDate ) {
+        ctrl.notify('Invalid Date.', 'error');
```
zahnno, line 68: Please remove the notify and replace with a console error.

Henry0422: Console.error won't be a good way. We need a way to notify the user their input is not appropriate

src/foam/u2/DateView.js
```diff
@@ -34,19 +38,36 @@ foam.CLASS({
...
           function(value) {
             return new Date(value);
           }
       );
     }
+  ],
+
+  listeners: [
+    function checkDateFormat(value) {
+      oldDate = value.split(/[^0-9]/).map((str) => {
```
TW80000, line 61: Missing `var`

Henry0422:
OK, this might break the other DateView.....

TW80000 (changes requested):

TW80000:
I don't think I like the approach in this PR anymore. The best solution to this problem would be to create a FOAM date-picker that would be used in all browsers and wouldn't let users select invalid dates in the first place.

jlhughes:
I’m in agreement with @TW80000. It does seem absurd that a user can pick an invalid date.  With regard to launch, since it’s beta, we can restrict browser choice, and then deal with this issue properly which is mostly likely a FOAM date picket.

kgrgreer:
FOAM1 has /js/foam/u2/md/DatePicker.js

PR #1863, opened 2019-02-01
Fix routing journals

daudich:
- Fixed `routingJournals`:
  - synchronization issues
  - recursion issues
- Added `FindReplayDAO` decorator to return the underlying `innerDAO` for replaying records directly into it instead of going through all of the decorators.

Blocked because of this: https://github.com/foam-framework/foam2/pull/1878


src/foam/dao/EasyDAO.js
```diff
@@ -64,25 +64,62 @@ foam.CLASS({
...
-      MDAO:   'foam.dao.MDAO'
+      IDB: 'foam.dao.IDBDAO',
+      LOCAL: 'foam.dao.LocalStorageDAO',
+      MDAO: 'foam.dao.MDAO'
     }
   },
 
+  classes: [
+    {
+      name: 'ReplaySharedJournal',
```
mcarcaso: please flag this as 'java'.
i.e. add this:
```
flags: ['java'],
```
This doesn't do much yet, but it will allow this to get stripped when we start stripping the js payload. Keep this in mind for and properties/methods/classes you make that are strictly for java.

src/foam/dao/EasyDAO.js
```diff
@@ -349,6 +392,14 @@ return delegate;
     {
       name: 'testData',
       generateJava: false
+    },
+    {
+      name: 'isSharedJournal',
```
mcarcaso, line 380: Do we need this and `journaled`? Can we collapse it into `journalType` where the options are: none, single, shared? Defaults to none.

src/foam/dao/RoutingJDAO.js
```diff
@@ -19,21 +19,71 @@ foam.CLASS({
       class: 'String',
       name: 'service',
       documentation: 'Name of the service'
+    },
+    {
+      class: 'Boolean',
+      name: 'replayed',
+      documentation: 'Has this journal been replayed yet?',
+      value: false
```
mcarcaso: You can remove this line. `class: 'Boolean'` has a default value of false so this is redundant.

src/foam/dao/RoutingJDAO.js
```diff
@@ -19,21 +19,71 @@ foam.CLASS({
...
     }
   ],
 
+  axioms: [
+      {
+        name: 'javaExtras',
+        buildJavaClass: function (cls) {
+          cls.extras.push(`
+            public final static Object ROUTING_JDAO_REPAYED_CMD = new Object();
```
mcarcaso: Do you mean ROUTING_JDAO_REPLAYED_CMD?

This seems odd to me. Why does the RoutingDAO know/care about replaying? What I think you want is a java equivalent of the PromisedDAO that we have in JS. It's a DAO that blocks all calls on it until its promise resolves.

src/foam/dao/RoutingJournal.js
```diff
@@ -9,9 +9,25 @@ foam.CLASS({
...
+    'foam.dao.java.FindReplayDAO'
+  ],
+
+  axioms: [
+    {
+      name: 'javaExtras',
+      buildJavaClass: function(cls) {
+        cls.extras.push(`
+          // Boolean to check if the journal file is being or has been replayed
+          protected volatile boolean journalReplayed_ = false;
```
mcarcaso, line 26: I think this would also not be needed if you had a PromisedDAO.

src/foam/dao/java/FindReplayDAO.java
```diff
@@ -0,0 +1,48 @@
...
+
+import foam.dao.DAO;
+import foam.dao.ProxyDAO;
+import foam.core.X;
+
+/**
+ * This DAO is a decorator around the innerDAO to return the DAO when replaying
+ * entries directly into the DAO.
+ */
+public class FindReplayDAO
```
mcarcaso: Always model instead of hand written java class...unless you have a reason not to?

src/foam/dao/java/FindReplayDAO.java
```diff
@@ -0,0 +1,48 @@
...
+  extends ProxyDAO {
+
+  public final static Object FIND_REPLAY_DAO_CMD = new Object();
```
mcarcaso: Can you use the context instead to store a 'replayDAO' that RoutingJournal can look up? Then I don't think you'd need this FindReplayDAO at all.

mcarcaso (commented):
Feel free to msg me if you have any questions about my comments.

src/foam/dao/EasyDAO.js
```diff
@@ -388,13 +439,13 @@ return delegate;
 
       if ( ! daoModel ) {
         this.warn(
-          "EasyDAO: Unknown DAO Type.  Add '" + daoType + "' to requires: list."
+          'EasyDAO: Unknown DAO Type.  Add \'' + daoType + '\' to requires: list.'
```
TW80000, line 425:
```suggestion
          `EasyDAO: Unknown DAO Type.  Add ${daoType} to requires: list.`
```

src/foam/dao/EasyDAO.js
```diff
@@ -388,13 +439,13 @@ return delegate;
...
         );
       }
 
       if ( this.name && daoModel.getAxiomByName('name') ) params.name = this.name;
       if ( daoModel.getAxiomByName('autoIndex') ) params.autoIndex = this.autoIndex;
-      //if ( this.seqNo || this.guid ) params.property = this.seqProperty;
+      // if ( this.seqNo || this.guid ) params.property = this.seqProperty;
```
TW80000: Commented out code can be removed.

src/foam/dao/EasyDAO.js
```diff
@@ -413,13 +464,13 @@ return delegate;
...
-          this.mdao = this.MDAO.create({of: params.of});
+          this.mdao = this.MDAO.create({ of: params.of });
           dao = this.CachingDAO.create({
             cache: this.dedup ?
               this.mdao :
-              this.DeDupDAO.create({delegate: this.mdao}),
+              this.DeDupDAO.create({ delegate: this.mdao }),
             src: dao,
-            of: this.model});
+            of: this.model });
```
```suggestion
            of: this.model
          });
```

src/foam/dao/EasyDAO.js
```diff
@@ -430,29 +481,29 @@ return delegate;
         });
       }
 
-      if ( this.seqNo && this.guid ) throw "EasyDAO 'seqNo' and 'guid' features are mutually exclusive.";
+      if ( this.seqNo && this.guid ) throw 'EasyDAO \'seqNo\' and \'guid\' features are mutually exclusive.';
```
```suggestion
      if ( this.seqNo && this.guid ) throw `EasyDAO 'seqNo' and 'guid' features are mutually exclusive.`;
```

src/foam/dao/EasyDAO.js
```diff
@@ -430,29 +481,29 @@ return delegate;
...
 
-      if ( this.syncWithServer && this.isServer ) throw "isServer and syncWithServer are mutually exclusive.";
+      if ( this.syncWithServer && this.isServer ) throw 'isServer and syncWithServer are mutually exclusive.';
 
       if ( this.syncWithServer || this.isServer ) {
         if ( ! this.syncProperty ) {
           this.syncProperty = cls.SYNC_PROPERTY;
           if ( ! this.syncProperty ) {
-            throw "EasyDAO sync with class " + cls.id + " invalid. Sync requires a sync property be set, or be of a class including a property 'sync_property'.";
+            throw 'EasyDAO sync with class ' + cls.id + ' invalid. Sync requires a sync property be set, or be of a class including a property \'sync_property\'.';
```
TW80000, line 482: I suggest using backticks when you find yourself needing to escape quotes in strings in JavaScript. I think it makes it more readable without the slashes.

src/foam/dao/EasyDAO.js
```diff
@@ -523,11 +574,13 @@ return delegate;
                 return;
               }
 
-              self.log("Loading test data");
+              self.log('Loading test data');
               Promise.all(foam.json.parse(self.testData, self.of, self).map(
-                function(o) { return delegate.put(o); }
+                function(o) {
+ return delegate.put(o);
```
TW80000, line 548: Please fix indentation.

src/foam/dao/RoutingJDAO.js
```diff
@@ -19,21 +19,71 @@ foam.CLASS({
...
+    {
+      class: 'Boolean',
+      name: 'replayed',
+      documentation: 'Has this journal been replayed yet?',
+      value: false
     }
   ],
 
+  axioms: [
+      {
```
TW80000: Please fix indentation.

src/foam/dao/java/FindReplayDAO.java
```diff
@@ -0,0 +1,48 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
TW80000: 2019

src/foam/dao/java/FindReplayDAO.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package foam.dao.java;
+
+import foam.dao.DAO;
+import foam.dao.ProxyDAO;
+import foam.core.X;
+
+/**
+ * This DAO is a decorator around the innerDAO to return the DAO when replaying
```
TW80000: This is confusing. What is the `innerDAO`? Which DAO are you referring to in "... return the DAO when ..."? Which DAO are you referring to in "... directly into the DAO"?

src/foam/dao/java/FindReplayDAO.java
```diff
@@ -0,0 +1,48 @@
...
+ * entries directly into the DAO.
+ */
+public class FindReplayDAO
+  extends ProxyDAO {
+
+  public final static Object FIND_REPLAY_DAO_CMD = new Object();
+
+  public FindReplayDAO (DAO d){
```
TW80000: Missing space

src/foam/dao/java/FindReplayDAO.java
```diff
@@ -0,0 +1,48 @@
...
+    setDelegate(d);
+  }
+
+  public Object cmd (Object obj){
```
TW80000: Missing space

TW80000: Isn't the implementation supposed to go in the CO method (`cmd_` in this case)? By default the NCO methods (`cmd` in this case) just call the CO methods, see AbstractDAO. I think you have it the other way around here.

src/foam/dao/EasyDAO.js
daudich: Done, thanks!

src/foam/dao/EasyDAO.js
daudich: That works. I was trying to add a change that would result in the least possible disruption with the rest of settings. Will make this change.

src/foam/dao/EasyDAO.js
daudich: I agree. I think this was old code, I haven't touched this. but ESLint cleaned it for me.

src/foam/dao/java/FindReplayDAO.java
daudich: The `innerDAO` nomenclature has been borrowed from the EasyDAO model. It can be anything MapDAO, MDAO, etc. But point taken, I have re-worded the sentence and added more clarification.

src/foam/dao/java/FindReplayDAO.java
daudich: Changed.

src/foam/dao/java/FindReplayDAO.java
daudich: No reason. Will convert to a modelled class.

src/foam/dao/RoutingJDAO.js
daudich: Using PromisedDAO now.

src/foam/dao/RoutingJournal.js
daudich: No, I think I will need this. Wnen the first service initiates a `replay`, all of the records in the journal will be attempted to be read into the DAOs. Other DAOs would also initiate a replay as the PromisedDAO is a decorator around individual DAOs, and we want to avoid multiple DAOs calling `replay` on the journal.

PR #1864, opened 2019-02-01
Address validation improvements

TW80000:
* Add country and region validation to Address model
* Changed the order of the properties so that the validation messages show up in the order we want
* Made distinction between when a field is empty and when a field is filled out but invalid
* Made distinction between postal code and zip code for Canada and USA
* When the countryId property changes, reset the regionId property

### DIFF
--- a/src/foam/nanos/auth/Address.js
+++ b/src/foam/nanos/auth/Address.js
@@ -72,44 +72,6 @@ foam.CLASS({
       }
     },
     {
-      class: 'String',
-      name: 'suite',
-      documentation: 'Suite pertaining to address.',
-      width: 16,
-      validateObj: function(suite) {
-        var suiteRegex = /^[a-zA-Z0-9 ]{1,70}$/;
-
-        if ( suite.length > 0 && ! suiteRegex.test(suite) ) {
-          return 'Invalid address line.';
-        }
-      }
-    },
-    {
-      class: 'String',
-      name: 'city',
-      documentation: 'City pertaining to address.',
-      required: true,
-      validateObj: function(city) {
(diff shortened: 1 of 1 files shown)


daudich (approved):
LGTM

jlhughes (approved):

mcarcaso (commented):
Does `Address` need `countryId` or can we get away with only having a `regionId`? If you only want the `countryId` for view purposes, you should consider creating a custom view for `regionId` that has a country dropdown in it.

jlhughes:
@mcarcaso region is province/state.  Are you suggesting that we calculate country from region?  With regards to use, we need a model that returns both country and region.   Also with regard to data input, country is selected first, then the region is filtered.

mcarcaso:
Yeah, that's what I'm suggesting. Regions have a countryId so isn't it
redundant to store a country and a region in address? The validation
requires a region and that the region's country matches the address'
country. All that validation can be removed if you only store and require a
region. Then create a custom view for regions that shows a country dropdown
first, and a region dropdown that reacts to the country dropdown.

jlhughes:
Along with a relationship between country and region so we can easily 'find' the other(s).  
And if the region were a reference in Address then: 
Region region = address.findRegion(x);
Country country = region.findCountry(x);
And on the java side the findFoo(x) are synchronous. 

It would require a number of code changes where we use the address.  Perhaps as a bridge we can keep countryId as deprecated.

jlhughes:
In the immediate we have a launch on Monday, so these changes while nice, are just not in the cards at the moment.  The validation is needed so I think we should go with this and open another ticket/PR to make the changes we've discussed along with migration code.

PR #1868, opened 2019-02-04
Fix ChoiceView not recognizing falsey values

### LINKED ISSUE #1867: ChoiceView doesn't recognize falsey data values

TW80000:
Closes #1867 

This commit changes the conditions under which ChoiceView recognizes a 'valid' value for its `data` property.

Only `null` and `undefined` will be considered invalid if this commit is merged. All other values will be considered valid.

This change makes `0` a valid value. Currently `0` is an invalid value.

### DIFF
--- a/src/foam/u2/view/ChoiceView.js
+++ b/src/foam/u2/view/ChoiceView.js
@@ -187,7 +187,7 @@ foam.CLASS({
     function initE() {
       // If no item is selected, and data has not been provided, select the 0th
       // entry.
-      if ( ! this.data && ! this.index ) {
+      if ( this.data == null && ! this.index ) {
         this.index = 0;
       }
 
@@ -249,7 +249,7 @@ foam.CLASS({
       code: function() {
         var d = this.data;
         if ( this.choices.length ) {
-          this.choice = ( d && this.findChoiceByData(d) ) || this.defaultValue;
+          this.choice = ( d != null && this.findChoiceByData(d) ) || this.defaultValue;
         }
       }
     },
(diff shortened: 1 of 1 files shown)


mcarcaso:
Hmm, you're making it better by making 0, false, or an empty string valid but I'm not sure I agree with the premise that null is invalid. Why should it be?

mcarcaso:
I'll approve this because I don't think you're making things any worse but this view could use some simplification I think. Null shouldn't be invalid but it's already considered invalid so shouldn't be any regressions here.

mcarcaso (approved):

TW80000:
> Hmm, you're making it better by making 0, false, or an empty string valid but I'm not sure I agree with the premise that null is invalid. Why should it be?

I'm not sold it is either, but here are my thoughts trying to play devil's advocate:

ChoiceView is a view for a user to make a selection from a list of options. Each option has a corresponding value behind the scenes. The list of options _may_ contain a placeholder element, which represents a lack of selection, but such an element is not required. If such an option exists, its value is the only "invalid" value that can be selected.

We're considering whether the value for an invalid selection should be `undefined` OR `null` or whether it should be only `undefined`, leaving `null` as a valid value.

If `null` is a valid value, than one of the non-placeholder options may have it as its value. If one non-placeholder option has null as its value, then what would the other non-placeholder options have as their values?

I can't think of a situation in which it's sensible to do such a thing, but that doesn't mean there isn't one obviously. However, if there actually is no situation in which that makes sense, then there's no point in counting `null` as a valid value anyway.

mcarcaso:
I think undefined should be the lack of a value and null be treated as any other value. There shouldn't be anything wrong with me setting choices to:
[ [1, 'One!'], [null, 'Null :('], [2, 'Deuce'] ]

If data is set to something that's not in choices, I don't think the view should take it upon itself to change the value because then simply rendering a view will change the data and I don't think that should happen unless the user interacts with the view.

PR #1869, opened 2019-02-04
Refactor auth web agent with boot context

deejes:
Background info in pr #1845


src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -42,6 +42,32 @@ foam.INTERFACE({
...
+      ],
+      swiftThrows: true,
+      args: [
+        {
+          name: 'x',
+          javaType: 'foam.core.X',
+          swiftType: 'Context'
+        },
+        {
+          name: 'agent',
```
kgrgreer, line 58: I would like better names for agent and entity.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -100,6 +87,46 @@ public User actAs(X x, User entity) throws AuthenticationException {
...
+
+      if ( ! group.getEnabled() ) {
+        throw new AuthorizationException("Entity' group must be enabled.");
+      }
+
+      /*
+      Finds the UserUserJunction object to see if user can act as the passed in user.
+      Source (agent) users are permitted to act as target (entity) users, not vice versa.
+      */
+      UserUserJunction permissionJunction = (UserUserJunction) agentJunctionDAO_.find(AND(
```
kgrgreer, line 106: Isn't there a method on the relationship or in the object to access the junction more conveniently?

src/foam/nanos/auth/AgentAuthService.js
deejes: So would I, but I think that's a different issue than what's being addressed here.

src/foam/nanos/auth/AgentUserAuthService.java
deejes: No, I don't think there is.

src/foam/nanos/auth/AgentAuthService.js
```diff
@@ -42,6 +42,32 @@ foam.INTERFACE({
         }
       ]
     },
+    {
+      name: 'canActAs',
+      documentation: `Returns a boolean indicating whether the provided agent has adequate permission to actAs the provided entity.`,
+      javaReturns: 'boolean',
+      swiftReturns: 'Bool',
+      javaThrows: [
```
TW80000: Your implementation doesn't throw either of these. It catches any exception thrown, logs it, and returns false.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -100,6 +87,46 @@ public User actAs(X x, User entity) throws AuthenticationException {
...
+  public boolean canActAs(X x, User agent, User entity) {
+    try {
+      Group group = (Group) groupDAO_.find(entity.getGroup());
+
+      if ( group == null ) {
+        throw new AuthorizationException("Entity must exist within a group.");
+      }
+
+      if ( ! group.getEnabled() ) {
+        throw new AuthorizationException("Entity' group must be enabled.");
```
TW80000: Typo: Entity' -> Entity's

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -169,7 +171,15 @@ public Session authenticate(X x) {
           .put(HttpServletRequest.class,  req)
           .put(HttpServletResponse.class, resp), email, password);
 
-        if ( user != null ) return session;
+        if ( user != null ) {
+          if ( ! SafetyUtil.isEmpty(actAs) ) {
+            AgentAuthService agentService = (AgentAuthService) x.get("agentAuth");
+            DAO localUserDAO = (DAO) x.get("localUserDAO");
+            User entity = (User) localUserDAO.find(Long.parseLong(actAs));
```
TW80000: Could this throw an exception if I put `&actAs=string` in the query instead of a valid long? Should we catch an handle that by throwing a descriptive error message?

deejes: YES!

src/foam/nanos/auth/AgentAuthService.js
kgrgreer: I don't think so. Name discussion is a suitable topic for code reviews.

src/foam/nanos/auth/AgentUserAuthService.java
kgrgreer: I think you could do something like:
 agent.getEntities().getJunctionDAO().find(entity.getId())
or maybe just:
 agent.getEntities().getJunctionDAO().find(entity)

Which is shorter and doesn't require you know the junction key or the method of joining with agent.

src/foam/nanos/http/AuthWebAgent.java
deejes: Have addressed this, but not comprehensively from a DIG point of view. Have created a different task for that.

src/foam/nanos/auth/AgentAuthService.js
deejes: Discussed offline. Name changes will be addressed in a different PR.

src/foam/nanos/auth/AgentUserAuthService.java
deejes: We're trying to the junction object, and not the entity - it has a status and a group associated with it that we need to check and assign respectively.

deejes: Is there a way to do that through relationships?

src/foam/nanos/http/AuthWebAgent.java
deejes: So Joel and I talked about this, I fixed this locally, but it still doesn't write the appropriate message to an api consumer. Going to address that in a different pr.

src/foam/nanos/auth/AgentUserAuthService.java
kgrgreer: agent.getEntities().getJunctionDAO() gets the JunctionDAO, which contains the junctions.

src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -169,7 +171,20 @@ public Session authenticate(X x) {
...
-        if ( user != null ) return session;
+        if ( user != null ) {
+          if ( ! SafetyUtil.isEmpty(actAs) ) {
+            AgentAuthService agentService = (AgentAuthService) x.get("agentAuth");
+            DAO localUserDAO = (DAO) x.get("localUserDAO");
+            try {
+              User entity = (User) localUserDAO.find(Long.parseLong(actAs));
+              agentService.actAs(session.getContext(), entity);
+            } catch (java.lang.NumberFormatException e) {
+              logger.error("actAs must be a number:" + e);
```
kgrgreer, line 182: Can't we support email as well, since that's what user use to login, not their userId.

src/foam/nanos/auth/AgentUserAuthService.java
deejes: sure, but then the rest of the query will be the same right? Happy to change if you'd like, but doesn't seem super consequential to me. If anything, the way it is might be marginally faster? Let me know either way.

src/foam/nanos/http/AuthWebAgent.java
deejes: That should work, but it would have to be the email of the business, and in order to look that up they would (I think) need the id anyway. Also, there may be a business and a user with the same email ID, which might be confusing.

Otherwise the way it works in the gui is they are shown a list of businesses they can actAs when they sign in, and pick one. We should have a call for them to get a similar list of businesses, but after that I think id as the identifier still makes sense to me.

TW80000: > Can't we support email as well, since that's what user use to login, not their userId.

@kgrgreer Correct, but you don't use an email to choose which business to act as.

src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -63,10 +63,9 @@ public User actAs(X x, User entity) throws AuthenticationException {
       Finds the UserUserJunction object to see if user can act as the passed in user.
       Source (agent) users are permitted to act as target (entity) users, not vice versa.
     */
-    UserUserJunction permissionJunction = (UserUserJunction) agentJunctionDAO_.find(AND(
-      EQ(UserUserJunction.SOURCE_ID, agent.getId()),
-      EQ(UserUserJunction.TARGET_ID, user.getId())
-    ));
+    UserUserJunction permissionJunction = (UserUserJunction) agentJunctionDAO_.find(
+        EQ(UserUserJunction.TARGET_ID, entity.getId())
```
TW80000: Was checking `sourceId` actually redundant? I don't think it was. `agentJunctionDAO` is just an EasyDAO, there's nothing special about it afaict that would be inferring the source id based on the user making the call.

Looking into that also made me realize there's no security on `agentJunctionDAO` at the moment. We've refined `UserUserJunction` to implement the `Authorizable` interface, but we don't actually have an `AuthorizationDAO` decorating the `agentJunctionDAO` to call those methods. So we should add an `AuthorizationDAO` to `agentJunctionDAO`.

All that being said though, I could be wrong or missing something. Did you test this? It seems to me that it'll just return the first junction with a matching targetId, which won't necessarily give you the one you want.

kgrgreer: You're right. The other relationship DAO's pre-join the DAO, but I see that the Junction DAO doesn't (but probably should). Please revert. sorry

deejes: Great catch @TW80000, thanks for following up.

Made the change, but can't test on up to date branches till the foam/np conflict is fixed.

PR #1870, opened 2019-02-04
dig web agent

smirnovaka:
output correct obj


jlhughes:
Will be addressed by Minsun.

PR #1871, opened 2019-02-04
deleted aware DAO fix

shreymahendru:


jlhughes:
What exactly is the ‘fix’?

src/foam/nanos/auth/DeletedAwareDAO.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public void removeAll_(X x, long skip, long limit, Comparator order, Predicate predicate) {
+    getDelegate().select_(x, new RemoveSink(x, this), skip, limit, order, predicate);
+  }
+
+  private boolean canReadDeleted(X x, FObject obj) {
+
+    String deletePermission;
+    if(name_ == null) {
```
jlhughes: @shreymahendru this can be done in the constructor (I believe). ProxyDAOs should have an 'of' property with the model.

src/foam/nanos/auth/DeletedAwareDAO.java
```diff
@@ -0,0 +1,110 @@
+package foam.nanos.auth;
+
+import foam.core.FObject;
+import foam.core.X;
+import foam.dao.*;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+
+/*
```
kgrgreer: If you comment this with /** instead of /* then the comments will be included in the javadocs.

src/foam/nanos/auth/DeletedAwareDAO.java
kgrgreer: Spacing should be:
if ( name_ == null ) {

src/foam/nanos/auth/DeletedAwareDAO.java
```diff
@@ -0,0 +1,110 @@
...
+    String deletePermission;
+    if(name_ == null) {
+      String of = obj.getClass().getSimpleName().toLowerCase();
+      deletePermission = of + ".read.deleted";
+    } else {
+      deletePermission =  name_ + ".read.deleted";
+    }
+
+
+    if( obj instanceof DeletedAware
```
kgrgreer: add space after if

src/foam/nanos/auth/DeletedAwareDAO.java
```diff
@@ -0,0 +1,110 @@
...
+    Sink deletedAwareSink = new DeletedAwareSink(x, sink);
+    getDelegate().select_(x, deletedAwareSink, skip, limit, order, predicate);
+
+    return sink;
+  }
+
+  @Override
+  public FObject remove_(X x, FObject obj) {
+
+    if ( obj instanceof DeletedAware ) {
```
kgrgreer, line 78: Won't this always be the case?

src/foam/nanos/auth/DeletedAwareDAO.java
```diff
@@ -0,0 +1,110 @@
...
+
+    return obj;
+  }
+
+  @Override
+  public void removeAll_(X x, long skip, long limit, Comparator order, Predicate predicate) {
+    getDelegate().select_(x, new RemoveSink(x, this), skip, limit, order, predicate);
+  }
+
+  private boolean canReadDeleted(X x, FObject obj) {
```
kgrgreer: Making methods private makes them more difficult to test.

kgrgreer: This can be done once in removeAll() or remove rather than for each object. Then just apply the extra where() or not rather than having to check for each object.

src/foam/nanos/auth/DeletedAwareDAO.java
shreymahendru: 👍

src/foam/nanos/auth/DeletedAwareDAO.java
shreymahendru: Most likely. But this is just a defensive check. Also acts as a null check cause calling `instanceof` on a `null` obj returns false.

src/foam/nanos/auth/DeletedAwareDAO.java
shreymahendru: This method is applied to the find function and select's decorated Sink; where it checks if the user can read a deleted obj. Example: if this obj is deleted, but the user still gets it back when find/select is called, with the `deleted` prop on the obj as `true`.
This is not checking if the user can remove an obj or not. That shouldn't be this decorator's responsibility, in my opinion.

shreymahendru: 👍

src/foam/nanos/auth/DeletedAwareDAO.java
```diff
@@ -0,0 +1,102 @@
...
+  private String name_;
+
+  private class DeletedAwareSink extends ProxySink {
+
+    public DeletedAwareSink(X x, Sink delegate) {
+      super(x, delegate);
+    }
+
+    @Override
+    public void put(Object obj, foam.core.Detachable sub) {
```
jlhughes, line 37: put should return an FObject.

jlhughes (changes requested):

jlhughes (approved):

lchanmann:
@shreymahendru why can't we extend on the JS file instead of manually converting it to Java? Can you also add tests that illustrate the issues leading to this fix?

src/foam/nanos/auth/DeletedAwareDAO.java
```diff
@@ -0,0 +1,102 @@
...
+      return obj;
+    }
+
+    return null;
+  }
+
+  @Override
+  public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
+    Sink deletedAwareSink = new DeletedAwareSink(x, sink);
+    getDelegate().select_(x, deletedAwareSink, skip, limit, order, predicate);
```
jlhughes, line 70: Change this to something like:
```
      AuthService authService = (AuthService) getX().get("auth");
      if ( authService.check(x, deletePermission) ) {
        getDelegate().select_(x, sink, skip, limit, order, predicate);
      } else {
        getDelegate()
          .where(EQ(getOf().getClassInfo().getAxiomByName("deleted"), false)) //<-- I think, could also test for this as a replacement for the obj instanceof DeletedAware. 
          .select_(x, sink, skip, limit, order, predicate);
      } 


```

PR #1872, opened 2019-02-05
Fixing Deleted aware tests

shreymahendru:
fixing deletedAware tests 
and fixing a bug in PermissionedPropertyDAO.js, where `find` instead of `find_` is being called on the delegate. This causes the delegates to get the system context and not the user context.

### DIFF
--- a/src/foam/nanos/auth/DeletedAwareDAOTest.js
+++ b/src/foam/nanos/auth/DeletedAwareDAOTest.js
@@ -44,9 +44,7 @@ foam.CLASS({
       ],
       javaCode: `
         DAO delegate = new MDAO(DeletedAwareDummy.getOwnClassInfo());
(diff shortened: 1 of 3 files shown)


src/foam/nanos/auth/DeletedAwareDAOTest.js
```diff
@@ -44,9 +44,7 @@ foam.CLASS({
       ],
       javaCode: `
         DAO delegate = new MDAO(DeletedAwareDummy.getOwnClassInfo());
-        DAO dao = (DAO) new DeletedAwareDAO.Builder(x)
-          .setDelegate(delegate)
-          .build();
+        DAO dao = (DAO) new DeletedAwareDAO(x, delegate);
```
jlhughes, line 47: Here is an example where the js is preferred, as the builder is so helpful for configuration.

jlhughes (approved):

PR #1876, opened 2019-02-06
deleted aware changes

shreymahendru:


src/foam/nanos/auth/DeletedAwareDAO.java
```diff
@@ -90,15 +83,9 @@ public void removeAll_(X x, long skip, long limit, Comparator order, Predicate p
     getDelegate().select_(x, new RemoveSink(x, this), skip, limit, order, predicate);
   }
 
-  public boolean canReadDeleted(X x, FObject obj) {
+  public boolean canReadDeleted(X x) {
     String deletePermission =  name_ + ".read.deleted";
```
kgrgreer, line 87: In a new PR, make this value be an instance variable. No use creating new garbage on every call.

shreymahendru: Sorry that was ugly. I forgot to change that

PR #1878, opened 2019-02-06
Java Promised axiom generation

mcarcaso:
Also does a bunch of flagging and makes java generation respect flags.


mcarcaso:
Since the test gets mangled by the journal, here's what it looks like:
```
promisedDAO = new foam.dao.PromisedDAO.Builder(x)
  .setOf(foam.nanos.auth.Country.getOwnClassInfo())
  .build();

new Thread() {
  public void run() {
    mdao = new foam.dao.MDAO(foam.nanos.auth.Country.getOwnClassInfo());
    country = new foam.nanos.auth.Country();
    country.setCode("CN");
    country.setName("CHINA");
    mdao.put(country);
    promisedDAO.setPromise(mdao);
  }
}.start();

test(promisedDAO.select().getArray().size() == 1, "DAO has 1 element");
```

PR #1879, opened 2019-02-06
add switch demo

hchoura:


src/com/foamdev/demos/switch/Controller.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
```
kgrgreer, line 9: Document what this is demonstrating.

src/com/foamdev/demos/switch/Controller.js
```diff
@@ -0,0 +1,57 @@
...
+        //other manner to present the data.
+        /*choices: [
+          ['selOpt1','option 1' ] ,
+          ['selOpt2','option 2' ],
+          ['selOpt3','option 3' ]
+        ]*/
+      },
+    },
+    {
+      name: 'setNameHere',
```
kgrgreer: What is this property for? I think it confuses the example.

hchoura: This is a slot.
This property allow to add functionality to the ChoiceView.

src/com/foamdev/demos/switch/Controller.js
```diff
@@ -0,0 +1,57 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+// This example demonstrate the manner to use ChoiceView (combobox) in FOAM
+
+foam.CLASS({
+  package: 'com.foamdev.demos.switch',
+  name: 'Controller',
+  extends: 'foam.u2.Controller',// it will automatically exports: ['this as data']
+  //extends: 'foam.u2.Element',
```
kgrgreer: remove this line

src/com/foamdev/demos/switch/Controller.js
```diff
@@ -0,0 +1,57 @@
...
+  //you can avoid extending any model to get the default model presentation
+
+  //exports: [ 'as data' ],
```
kgrgreer: remove

src/com/foamdev/demos/switch/Controller.js
kgrgreer: I don't understand how this works.

hchoura: Since we can pass an Array of [value, text] choices. This property will show the text. ( In this case, it is just to present the feature).

PR #1882, opened 2019-02-07
add data binding example

hchoura:


src/com/foamdev/demos/dataBinding/Controller.js
```diff
@@ -0,0 +1,94 @@
...
+      this.start('div').add('Invoice:').end().
+      start('div').add('Quantity: ').
+        start(this.QTY).attrs({
+          onKey: true
+        }).end().
+      end(). //or start('div').add('Quantity: ').add(this.QTY).end().
+      start('div').add('Costs: ').add(this.COST).end().     
+      start('div').add('Total: ').add(this.TOT).end().
+      start('div').add('Total :',
+        /*foam.core.ExpressionSlot.create({
```
kgrgreer: delete this code if not used
maybe add some comments to explain this is an alternate approach to using the expression

PR #1884, opened 2019-02-07
add applying a css style to an element

hchoura:
TODO apply the css style (color) just for the clicked element.


src/com/foamdev/demos/addCssToClass/Controller.js
```diff
@@ -0,0 +1,83 @@
...
+      name: 'selectedAnimal',
+      value: 'cat'
+    },
+    'data',
+    {
+      // class: 'String',
+      name: 'colorStyle',
+      value: 'red',
+      expression: function(colorStyle) {
+        return this.myClass(true ? 'color' : 'red' ? 'color' : 'green');
```
kgrgreer: This expression looks strange. Won't it always just return 'color'?

PR #1885, opened 2019-02-07
Bootstrap

hchoura:


src/com/foamdev/demos/bootstrap/ControllerForms.js
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+foam.CLASS({
+  package: 'com.foamdev.demos.bootstrap',
+  name: 'ControllerForms',
+  extends: 'foam.u2.Element',
+
+  exports: [ 'as data' ],
+
+  imports: [ 'window' ], //To add bootstrap JavaScript
```
kgrgreer: Is this import used in this example?

src/com/foamdev/demos/bootstrap/ControllerForms.js
```diff
@@ -0,0 +1,194 @@
...
+      placeholder: 'First name',
+    //value:'Mark'
+    },
+    {
+      class: 'Boolean',
+      name: 'checkBoxTest',
+      value: true,
+      view: {
+        class: 'foam.u2.CheckBox',
+        onKey: true
```
kgrgreer: Is onKey used for CheckBox? If set it updates the model on every keystroke rather just when the user leaves the text field, but checkboxes don't have any incremental changes, the value just toggles between true and false when you click on it.

src/com/foamdev/demos/bootstrap/ControllerForms.js
```diff
@@ -0,0 +1,194 @@
...
+            //TODO to test
+            })).
+          end().
+        end().
+        start('div').addClass('invalid-feedback').add('Example invalid feedback text').end();
+    }
+  ],
+  listeners: [
+    function MyEvent() {
+      var forms = document.getElementsByClassName('needs-validation');
```
kgrgreer, line 179: This is a very SLOPy (SeLector Oriented Programming) style rather than MVC, which would be much cleaner. myEvent should be an action rather than a listener and its isEnabled button should check if the object is valid.

PR #1886, opened 2019-02-07
Fix bug with reference properties in queries

TW80000:
I noticed the query `owner=101` wasn't working in the general search box in the DAO controller, which is supposed to be a valid MQL query. I looked into it and noticed that since the property `owner` was defined via a relationship, it was a `Reference`, not a `Long`. The query parser was only recognizing numeric properties by checking if they were an instance of `foam.core.Int`, so it was treating all `Reference` properties as containing string values. Therefore the query parser was generating an `IN_IC(OWNER, ["101"])` mlang instead of `EQ(OWNER, 101)`.

This commit fixes that issue by checking if the property is a reference to a long property and if so, treating its value as a number, not a string.

I did a full-project search for other places we might be doing the same thing and only found one place, so it's presumably fixed there as well.

### DIFF
--- a/src/foam/parse/QueryParser.js
+++ b/src/foam/parse/QueryParser.js
@@ -340,7 +340,10 @@ foam.CLASS({
             var values = v[2];
             // Int is actually the parent of Float and Long, so this captures all
             // numeric properties.
-            var isNum = foam.core.Int.isInstance(prop);
+            var isNum = foam.core.Int.isInstance(prop) ||
+              foam.core.Reference.isInstance(prop) &&
+              foam.core.Int.isInstance(prop.of.ID);
+
             var isFloat = foam.core.Float.isInstance(prop);
 
             var isDateField = foam.core.Date.isInstance(prop) ||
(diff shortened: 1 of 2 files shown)


src/foam/parse/QueryParser.js
```diff
@@ -72,18 +72,18 @@ foam.CLASS({
       value: function(alt, literal, literalIC, not, notChars, optional, range,
           repeat, seq, seq1, str, sym) {
         return {
-          START: sym('query'),
-          query: sym('or'),
+          'START': sym('query'),
```
kgrgreer: Why add quotes?

TW80000: It was a rule in the linter that didn't like some of the keys in the object being quoted and some not being quoted. I think it's happy when they're consistent, not mixed. I can disable that rule if you want.

TW80000: https://eslint.org/docs/rules/quote-props

TW80000: Reverted

PR #1887, opened 2019-02-07
example how to add HTML Element to the view

hchoura:


src/com/foamdev/demos/addHTMLElement/Controller.js
```diff
@@ -0,0 +1,38 @@
...
+  actions: [
+    {
+      name:  'addElementToDSL',
+      label: 'add a DSL',
+      help:  'Add HTML element to the view',
+
+      code: function() {
+        this.start('div').add('here you are').end();
+      }
+    },
```
kgrgreer: No trailing commas this line or next.

src/com/foamdev/demos/addHTMLElement/Controller.js
```diff
@@ -0,0 +1,38 @@
...
+  methods: [
+    function initE() {
+      this.
+        start(this.ADD_ELEMENT_TO_DSL, {label: 'add element to the HTML using the DSL'}).end();
+    }
+  ],
+  actions: [
+    {
+      name:  'addElementToDSL',
+      label: 'add a DSL',
```
kgrgreer: I think most users will find the term DSL confusing or foreign. We should use the term DOM or virtual-DOM.

PR #1890, opened 2019-02-08
Fixed returning only one obj when put

minsunny:
https://nanopay.atlassian.net/browse/CPF-142

When put array Dig data, only one data returned as a returning data.
Fixed this issue by adding Object[] type on json/Outputter to parse.


minsunny:
it is on wrong branch

PR #1894, opened 2019-02-11
Updated placeholder profile picture

KateMengjunXia:
Previous image path was incorrect
https://nanopay.atlassian.net/projects/CPF/issues/CPF-171?filter=allopenissues&orderby=priority%20DESC&keyword=placeholder

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -173,7 +173,7 @@ foam.CLASS({
       documentation: 'User\' profile picture.',
       view: {
         class: 'foam.nanos.auth.ProfilePictureView',
-        placeholderImage: 'images/ic-placeholder.png'
+        placeholderImage: 'images/merchant/ic-placeholder.png'
       }
     },
     {


src/foam/nanos/auth/User.js
```diff
@@ -173,7 +173,7 @@ foam.CLASS({
       documentation: 'User\' profile picture.',
       view: {
         class: 'foam.nanos.auth.ProfilePictureView',
-        placeholderImage: 'images/ic-placeholder.png'
+        placeholderImage: 'images/merchant/ic-placeholder.png'
```
jlhughes, line 176: can we move the image please to images/ . 
merchant should be deprecated and everything moved.

jlhughes (changes requested):

src/foam/nanos/auth/User.js
TW80000, line 176: The merchant folder is part of our nanopay repo, not part of foam. However, an even bigger issue is that we have a direct reference to nanopay for the images service in foam at the moment: https://github.com/foam-framework/foam2/blob/master/src/services#L16

jlhughes: Good catch, this property/view can be a refinement on the nanopay side.

KateMengjunXia: I'll fix it. Thanks for pointing out.

KateMengjunXia: Moved this change to nanopay refinementm and copied the picture from images/merchant to images
https://github.com/nanoPayinc/NANOPAY/pull/6185

PR #1899, opened 2019-02-12
Fix bug related to toast notifications

TW80000:
We have a controller that extends `ApplicationController`, and in it we register an alternative view in place of `NotificationMessage`. That means that all child views should use the replacement view instead of `NotificationMessage` when trying to create one.

However, we primarily call the `notify` method to notify the user of something rather than add a `NotificationMessage` manually. And since the `notify` method is defined on `ApplicationController`, which `require`s 'foam.u2.dialog.NotificationMessage', the old view is being shown. I'm assuming this is because the require happens before our controller registers the view replacement.

This commit fixes the issue by creating the notification in a different way, specifically one in which the class will be looked up in the context _after_  the replacement has been registered.

### DIFF
--- a/src/foam/nanos/controller/ApplicationController.js
+++ b/src/foam/nanos/controller/ApplicationController.js
@@ -42,7 +42,6 @@ foam.CLASS({
     'foam.nanos.u2.navigation.FooterView',
     'foam.u2.stack.Stack',
     'foam.u2.stack.StackView',
-    'foam.u2.dialog.NotificationMessage',
     'foam.nanos.session.SessionTimer',
     'foam.u2.dialog.Popup'
   ],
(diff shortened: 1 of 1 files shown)


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -42,7 +42,6 @@ foam.CLASS({
     'foam.nanos.u2.navigation.FooterView',
     'foam.u2.stack.Stack',
     'foam.u2.stack.StackView',
-    'foam.u2.dialog.NotificationMessage',
```
mcarcaso, line 45: You might want to keep this. Otherwise we can't easily figure out that this class depends on 'foam.u2.dialog.NotificationMessage'

mcarcaso (approved):

TW80000:
Closing in favour of a different approach.

PR #1901, opened 2019-02-12
Fix RulerDAO

lchanmann:


jlhughes (commented):
Don’t forget to re-ensable the RulerDAOTest in foam/src/tests.js

PR #1906, opened 2019-02-13
Exclude relationship properties from search filter

lchanmann:
The search filter uses GroupBy on properties however it doesn't make sense to GroupBy on a DAO property (aka. relationship).

### DIFF
--- a/src/foam/u2/view/ReciprocalSearch.js
+++ b/src/foam/u2/view/ReciprocalSearch.js
@@ -104,10 +104,7 @@ foam.CLASS({
         }
 
         return of.getAxiomsByClass(foam.core.Property)
-            .filter((prop) => prop.searchView
-              && ! prop.hidden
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/ReciprocalSearch.js
```diff
@@ -104,7 +104,10 @@ foam.CLASS({
         }
 
         return of.getAxiomsByClass(foam.core.Property)
-            .filter((prop) => prop.searchView && ! prop.hidden)
+            .filter((prop) => prop.searchView
+              && ! prop.hidden
+              && ! foam.dao.OneToManyRelationshipProperty.isInstance(prop)
```
adamvy: Rather than hard coding these specific properties, lets update or refine those property types to have "null" as their search view.  This might not forever be the only place we ask a property for its search view, its better that the property be honest about whether it has a suitable search view or not.

adamvy:
Great thanks!

PR #1907, opened 2019-02-13
F/np 5683/support X on sugar

minsunny:
Related to this : https://github.com/foam-framework/foam2/pull/1817/files

I found out it was not merged on master, so I would like to request to be merged.


src/foam/nanos/dig/SUGAR.js
```diff
@@ -69,28 +70,17 @@ foam.CLASS({
...
-               if ( fm.args[j].javaType.toString() == 'foam.core.X' ) {
-                  return false;
-               }
-               return true;
-             }
-             // return true;
-          }).map(function(m) { return m.name; }).sort();
-
-        if ( filteredMethod.length > 0 ) {
+        var methodName = methods.map(function(m) { return m.name; }).sort();
```
kgrgreer: I think a better name would be 'methodNames', since there are more than one name. The name 'methodName' is misleading since it suggests only one method name.

minsunny: done :) Thx

jlhughes: @minsunny did you push the change to the name?

PR #1911, opened 2019-02-16
Report when TargetDAOKey is not found in context

jlhughes:


src/foam/dao/RelationshipDAO.js
```diff
@@ -69,7 +69,15 @@ foam.CLASS({
 
         return delegate;
       },
-      javaFactory: 'return ((foam.dao.DAO) getX().get(getTargetDAOKey())).inX(getX());',
+      javaFactory:`
+      try {
+        return ((foam.dao.DAO) getX().get(getTargetDAOKey())).inX(getX());
```
kgrgreer, line 74: FYI: You could also use java asserts for this kind of thing.

jlhughes: We only noticed this in 'production', an assert would not have reported it - assuming assertions are disabled in production.

PR #1914, opened 2019-02-19
XMl outputter format changes and bug fixes

minsunny:
https://nanopay.atlassian.net/browse/CPF-201

1. XMl outputter format changes on java & js, so now printed out like this format:
 <objcets><object></object><object></object></objects>

screen shot 2019-02-19 at 11 44 14 am

2. failing Export xml fixed

3. when Dig data put, after it's gone successfully with CSV, XML, it returns input data as Json does.

### DIFF
--- a/src/foam/core/stdlib.js
+++ b/src/foam/core/stdlib.js
@@ -940,9 +940,7 @@ foam.LIB({
               'and no deafult method provided');
         }
 
-        if ( type[uid] || opt_defaultMethod ) {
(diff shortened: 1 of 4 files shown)


src/foam/core/stdlib.js
```diff
@@ -940,9 +940,7 @@ foam.LIB({
               'and no deafult method provided');
         }
```
kgrgreer, line 942: Why remove the 'if' statement?

kgrgreer:
Changes directly committed by me.

PR #1915, opened 2019-02-19
Adding JournalType for RoutingJournals

daudich:
- Added enums for distinguishing between journal types.
- Changed the service file entries to use JournalTypes.
- Added some documentation to routingJournals.


src/foam/dao/EasyDAO.js
```diff
@@ -86,7 +92,9 @@ foam.CLASS({
       /** The developer-friendly name for this EasyDAO. */
       class: 'String',
       name: 'name',
-      factory: function() { return this.of.id; }
+      factory: function() {
+ return this.of.id;
```
mcarcaso: What's up with this indentation?

src/foam/dao/EasyDAO.js
```diff
@@ -99,8 +107,10 @@ foam.dao.DAO delegate = getInnerDAO() == null ?
 
 if ( delegate instanceof foam.dao.MDAO ) setMdao((foam.dao.MDAO)delegate);
 
-if ( getJournaled() ) {
+if ( getJournalType().equals(JournalType.SINGLE_JOURNAL) ) {
   delegate = new foam.dao.java.JDAO(getX(), delegate, getJournalName());
+} else if ( getJournalType().equals(JournalType.SHARED_JOURNAL) ) {
+  delegate = new RoutingJDAO.Builder(getX()).setDelegate(delegate).setService(getName()).setJournal((RoutingJournal) getX().get("routingJournal")).build();
```
mcarcaso: Break this into multiple lines to make it easier to read:
```
new RoutingJDAO.Builder(getX())
  .setDelegate(delegate)
  .setService(getName())
  .setJournal((RoutingJournal) getX().get("routingJournal"))
  .build();
```

src/foam/dao/EasyDAO.js
```diff
@@ -197,9 +207,10 @@ return delegate;
     },
     {
       /** Keep a history of all state changes to the DAO. */
-      class: 'Boolean',
-      name: 'journaled',
-      value: false
+      class: 'foam.core.Enum',
+      of: 'foam.dao.JournalType',
+      name: 'journalType',
+      value: 'JournalType.NO_JOURNAL'
```
mcarcaso: I don't think this works but you might be getting NO_JOURNAL by luck because it's the first value for the enum. I think you want.
```
value: 'NO_JOURNAL'
```

mcarcaso: Also flag this for java only
```
flags: ['java']
```

src/foam/dao/JournalType.js
```diff
@@ -0,0 +1,21 @@
+foam.ENUM({
```
mcarcaso, line 18: You'll probably need the license header.

mcarcaso (approved):
LGTM just some small change requests.

src/foam/dao/EasyDAO.js
```diff
@@ -61,22 +61,28 @@ foam.CLASS({
...
+  imports: ['document'],
 
   constants: [
     {
       // Aliases for daoType
       name: 'aliases',
-      flags: [ 'js' ],
+      flags: ['js'],
       value: {
-        ARRAY:  'foam.dao.ArrayDAO',
```
adamvy: Personally I would prefer not to mix formatting cleanup PRs with functional changing PRs.  It makes it harder to review.

src/foam/dao/EasyDAO.js
```diff
@@ -86,7 +92,9 @@ foam.CLASS({
       /** The developer-friendly name for this EasyDAO. */
       class: 'String',
       name: 'name',
-      factory: function() { return this.of.id; }
+      factory: function() {
```
adamvy: I prefer the one liner, its so short.

src/foam/dao/EasyDAO.js
```diff
@@ -527,11 +523,13 @@ return delegate;
                 return;
               }
 
-              self.log("Loading test data");
+              self.log('Loading test data');
               Promise.all(foam.json.parse(self.testData, self.of, self).map(
-                function(o) { return delegate.put(o); }
```
adamvy: Again one liner is easier to read IMO.

src/foam/dao/RoutingJDAO.js
```diff
@@ -11,75 +11,87 @@ foam.CLASS({
...
+        then this is a find called from within a journal replay; bypass all of
+        the decorators. Otherwise, find on the decorator.`,
       flags: null,
       javaCode: `
-if ( x.get("replayingJournal") == getJournal() ) {
-  return getDelegate().find_(x, id);
-} else {
-  return super.find_(x, id);
-}
+        if ( x.get("replayingJournal") == getJournal() ) {
```
adamvy, line 62: This seems like a really odd thing to have to check for, and a really fragile condition.  Why is the default behaviour not appropriate in this case?

mcarcaso: While replaying, we only want DAO operations from the replaying journal to go through and any other operations should be blocked until replaying is complete.

adamvy: I read the documentation comment so I understand now.  But this still is really really fragile.

With regular journals we solve this by passing the inner dao to the journal for replay.  It would be better if we could find a way to still do that.

Perhaps in EasyDAO where we construct the RoutingJDAO, we could also do x.get("routingJournal") and give it the underlying DAO to use for the given name.  Something like x.get("routingJournal").setDAOForName(getName(), deelgate);

src/foam/dao/EasyDAO.js
daudich: It was the JS linter. I will fix it.

src/foam/dao/EasyDAO.js
daudich: It was my JS Linter. I will bear this in mind for the future.

src/foam/dao/EasyDAO.js
daudich: It was the JS linter. Will fix it.

src/foam/dao/EasyDAO.js
daudich: JS linter. Fixed.

src/foam/dao/RoutingJDAO.js
daudich: @adamvy Do you still want me to fix this?

My initial plan was to use a decorator with a command and wrap around the `innerDAO`.

src/foam/dao/EasyDAO.js
```diff
@@ -61,22 +61,28 @@ foam.CLASS({
...
-      flags: [ 'js' ],
+      flags: ['js'],
       value: {
-        ARRAY:  'foam.dao.ArrayDAO',
+        ARRAY: 'foam.dao.ArrayDAO',
         CLIENT: 'foam.dao.RequestResponseClientDAO',
-        IDB:    'foam.dao.IDBDAO',
-        LOCAL:  'foam.dao.LocalStorageDAO',
-        MDAO:   'foam.dao.MDAO'
+        IDB: 'foam.dao.IDBDAO',
```
kgrgreer: Why are you breaking alignment and making the code harder to read?

src/foam/dao/EasyDAO.js
kgrgreer: indentation is broken.

src/foam/dao/RoutingJDAO.js
adamvy: Yes I'd still like to address this, I'd rather not merge the code as is as it makes me nervious :).

For the non-routing journal case, the EasyDAO or the NSpec serviceScript is responsible for making sure the journal has a reference to the underlying dao for replay purposes.  I think we can preserve this in the routing journal case.

Looks like @mcarcaso is working on something along these lines here: https://github.com/foam-framework/foam2/pull/1918/files so let's coordinate with him.

src/foam/dao/EasyDAO.js
daudich: You are looking at an outdated version of the code. I have fixed this now.

src/foam/dao/EasyDAO.js
daudich: This would be the JS linter too. Will fix.

daudich:
Moved: https://github.com/foam-framework/foam2/pull/1936

PR #1916, opened 2019-02-19
Modify the Promised axiom to be cross-platform.

mcarcaso:
The issue with the previous approach is that you can't override methods for java in a class extending a Promised class because the methods get flagged for js/swift. This change makes all platforms have a bit more similar behaviour with this axiom.


adamvy:
What I don't like about this is that even after the waiting is done, you still call maybeWAitFor which is synchronized, this may hurt performance in highly concurrent situations.  Would be better if after this was started we changed state to one without the check, ideally then the JIT would eventually inline this decorator entirely if its in a hot spot.

mcarcaso:
Oh good point. I'll take a stab at that and let you know when it's updated.

mcarcaso:
Ok I updated this to be more like swift's implementation of the blocking axiom which didn't use synchronized and used a semaphore instead. I'm not sure if a semaphore is the right tool for the job here though. I release all waiting threads via:
```
semaphore.release(semaphore.getQueueLength());
```
but my understanding is that getQueueLength is just an estimate because that length could change from another thread so that might not flush all threads if something makes it into the condition to wait, but gets pre-empted before it starts waiting and during that time the property gets set and all waiters get flushed. Is that a possibility? My concurrency game isn't great so any thoughts/suggestions are greatly appreciated. Is there a concurrency tool other than a semaphore I'm missing?

src/foam/dao/BlockingDAO.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'BlockingDAO',
+  extends: 'foam.dao.AbstractDAO',
+  properties: [
+    {
+      class: 'foam.native.Blocking',
+      name: 'blocking',
```
adamvy: Convention is, or is moving towards just naming these all as "delegate"

src/foam/core/Promised.js
```diff
@@ -129,38 +109,35 @@ foam.CLASS({
...
         foam.core.Proxy.create({
           name: stateName,
-          flags: ['js', 'swift'],
           of: this.of,
-          delegates: methodNames,
-          forwards: [],
+          forwards: methodNames,
           factory: function() {
-            return this[pendingState].create();
+            return this[pendingState].create({obj: this});
```
adamvy: spacing should be `.create({ obj: this });`

src/foam/core/Promised.js
```diff
@@ -23,36 +23,17 @@ foam.CLASS({
...
-              });
-            };
-        }
+      expression: function(name, property, type) {
         return type ?
           function() {
-            var self = this;
             var args = arguments;
-            return this[property].then(function(d) {
+            return this.obj[property].then(function(d) {
```
adamvy: can we rename obj to delegate?

src/foam/core/Promised.js
```diff
@@ -129,38 +109,35 @@ foam.CLASS({
...
+      // Use all methods from here on in to satisfy the interface.
+      // Methods that aren't part of this.methods should never get called
+      // but they need to be added to make compilers happy.
+      methods = delegate.getOwnAxiomsByClass(foam.core.Method);
       for ( var i = 0 ; i < methods.length ; i++ ) {
         pendingMethods.push(foam.core.PromisedMethod.create({
           name: methods[i].name,
-          flags: ['js', 'swift'],
           property: myName,
-          type:  methods[i].type,
```
adamvy, line 163: Why drop the type?

mcarcaso: Since the model that we're applying to implements the interface of the `of`, the types and args are pulled in automatically.

src/foam/java/refinements.js
```diff
@@ -1957,57 +1953,54 @@ foam.CLASS({
...
+  flags: ['java'],
+  properties: [
+    {
+      name: 'javaCode',
+      getter: function() {
+        return `
 try {
-  maybeWaitFor${name}();
-  ${m2.javaType != 'void' ? 'return ' : ''}get${foam.String.capitalize(name)}().${m2.name}(${m2.args.map(a => a.name).join(', ')});
+  synchronized ( getObj() ) {
```
adamvy: I'm not sure the value of this synchronized lock?  Shouldn't getObj be synchronized?

mcarcaso: Oh maybe I'm mistaken but I thought I couldn't do wait() or notifyAll() without being in a synchronized block.

src/foam/core/Promised.js
mcarcaso: Done

src/foam/core/Promised.js
mcarcaso: Good call. Done

PR #1918, opened 2019-02-20
Make RoutingJournal talk directly to innerDAO

mcarcaso:
Previously, the RoutingJournal would register itself in the context it gives to the DAOs it's running replay operations on. Now, RoutingJDAOs register their inner daos with the journal so the journal can talk directly to them.


src/foam/dao/RoutingJDAO.js
```diff
@@ -49,37 +50,18 @@ new Thread() {
...
     {
       name: 'put_',
       flags: null,
       javaCode: `
-if ( x.get("replayingJournal") == getJournal() ) {
-  return getDelegate().put_(x, obj);
-} else {
-  return super.put_(x.put("service", getService()), obj);
-}
+return super.put_(x.put("service", getService()), obj);
```
adamvy, line 57: Rather than create a new context on every put/remove in the system (could be slow), let's decorate the journal that each routingJDAO has to know the service name and pass that along to the underlying routing journal as a plain old argument.

src/foam/dao/RoutingJDAO.js
```diff
@@ -32,6 +32,7 @@ foam.CLASS({
       of: 'foam.dao.RoutingJournal',
       name: 'journal',
       javaPostSet: `
+getJournal().setDAOForName(getService(), getDelegate());
```
adamvy, line 35: Is this the right place for this?  I think the EasyDAO and/or NSpec.serviceScript should do it.   The EasyDAO knows exactly which decorators should be before/after the JDAO.

src/foam/dao/RoutingJournal.js
```diff
@@ -112,7 +122,9 @@ try {
               int length = line.trim().length();
               line = line.trim().substring(2, length - 1);
 
-              dao = (foam.dao.DAO) x.get(service);
+              // Touch the service to potentially trigger factory.
+              x.get(service);
```
adamvy, line 126: @daudich use of cmd_ might be appropriate again here.  Rather than have to do the magic of x.get() to trigger a factory that we know will update our map of dao names, we can just do dao = x.get().cmd_(GET_REPLAY_JOURNAL).

If we updated the original JDAO to respect this command we might not even have to write a new RoutingJDAO.

adamvy:
Okay summarizing some thoughts here.

I don't think we need a RoutingJDAO at all.  I think with a few minor we can just use the existing JDAO.

First we make JDAO respect the GET_REPLAY_DAO command as @daudich had in his original PR.

Then we write a RoutingJournal class that _does not_ implement the Journal interface, but instead it implements something like
```
interface RoutingJournal {
  put(foam.core.Context x, String name, FObject obj).
  replay(foam.core.Context x);
}
````

We also implemented a class RoutedJournal, which bridges Journal to RoutingJournal
```
class RoutedJournal implemented Journal {
  String name_;
  RoutingJournal journal_;
  put_(foam.core.Context x, FObject obj) {
    journal_.put(x, name_, obj);
  }
}
```

Then in EasyDAO and NSpec.serviceScript when we construct the JDAO, if we want to use routing what we do is

```
new JDAO.Builder(x).setJournal(new RoutedJournal(name, x.get("routingJournal"))
```

In RoutingJournal we implement replay as
```
synchronized replay(foam.core.Context x) {
if ( replayed_ ) return;
foreach(Record r) {
  x.get(r.daoName()).cmd_(GET_REPLAY_DAO).put_(x, r.object);
}
}
```

adamvy:
The main advantages I see of this approach is that it doesn't really require any additional wiring, and the wiring that it does have seems fairly clear to me.

Some care will have to be done around how JDAO triggers the replay.  Trigger a DAO context factory will likely call back through to RoutingJournal.replay() which is synchronized.  Synchronized methods are re-entrant in Java apparently

https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html

Which means we won't deadlock but we have to be sure we don't replay twice.  Will mull this over over lunch.

adamvy:
Technically we wouldn't even have to add cmd_(GET_REPLAY_DAO), since JDAO would call through to replay() with the dao it wants to replay with, RoutedJournal could pass that along to the RoutingJournal, but the control flow there would probably be unclear so its not necessarily a good idea.

PR #1922, opened 2019-02-21
New RoutingJournal Approach

adamvy:
This approach re-uses the existing FileJournal and JDAO with slight modifications.

Instead of writing a single special Journal class that all writes/reads should go thread, we let each DAO have its own FileJournal but structure them so that they write the same underlying file/writer.

We ensure that each journal event is written in full by synchronizing journal operations on a lock, this lock is shared between all journals that write to the same file.

For replay, each DAO will walk the entire journal on its own, skipping any entires that are not tagged for it.  This approach is slower than if we read the whole file in one pass and dispatched events to each related DAO, but it also makes ordering easier to deal with, doesn't require modifing the journal interface and lets the different DAOs be lazily loaded when they are actually needed.  While not ideal, this performance hit is only a one time cost during startup.

A later revision can look at buffering journal events or replaying all the writes in a single pass.


adamvy:
@daudich I couldn't add you as a reviewer but please review.

I'm out of time to work on this today, so I didn't include the JournalType Enum, but if somebody wants to add that to the branch that would make me happy.

src/foam/dao/FileJournal.js
```diff
@@ -171,14 +152,27 @@ foam.CLASS({
           throw new RuntimeException(t);
         }
       `
+    },
+    {
+      class: 'Object',
```
mcarcaso, line 157: Random thought for another PR but could we model a file as a multiton on the filename and use synchronized methods to ensure they share a lock?

src/foam/dao/java/JDAO.js
```diff
@@ -27,13 +27,12 @@ foam.CLASS({
...
             // create journal
             setJournal(new foam.dao.FileJournal.Builder(x)
-              .setDao(delegate)
               .setFilename(filename)
               .setCreateFile(true)
               .build());
 
             // create a composite journal of repo journal
-            // and runtime journal and load them all
+// and runti me journal and load them all
```
mcarcaso, line 35: lol "runti me"

mcarcaso (approved):
I like this simplification :)

src/foam/dao/java/JDAO.js
daudich: Should be "runti us" there's no "me" in team.

adamvy: lol i'm not even sure how I changed this line.  Will fix.

src/foam/dao/FileJournal.js
adamvy: Not exactly.  The trick here is that we need to use a lock to ensure that complete "records" are written to the file atomically, which can consist of multiple actual "write" calls to the file.  As a journal entry and the comment describing it should be written together.

When/if we model Journal entries then we can remove the synchronization here and only apply at the appropriate level where the entry is actually written to the stream.

jlhughes (changes requested):
A requirement of this system to have journal entries processed 'in order'. 
At time X Account A has bank info Y
Transactions are replayed against this info.
later at time Y Account A has bank info Z
subsequent Transactions are replayed with this new info. 

Allowing each DAO to process the entire journal does  not provide the support we need.

adamvy:
Noted, will change to suit.

src/foam/dao/FileJournal.js
```diff
@@ -171,14 +152,27 @@ foam.CLASS({
...
+      name: 'lock',
+      javaType: 'java.util.concurrent.locks.Lock',
+      documentation: 'The lock to synchronize writing records on.  If multiple file journals write to the same file they should share a lock.',
+      javaFactory: 'return new java.util.concurrent.locks.ReentrantLock();',
```
daudich, line 161: Could you please make this a fair lock to ensure that the ordering is correct when we read back the records? This will also be useful to guarantee that if you stop reading at any point in the journal, you have the correct and complete state of the system up until that point.

src/foam/dao/FileJournal.js
```diff
@@ -237,9 +237,12 @@ foam.CLASS({
         }
       ],
       javaCode: `
-        BufferedWriter writer = getWriter();
+        java.io.Writer writer = getWriter();
         writer.write(data);
-        writer.newLine();
+        writer.write(System.getProperty("line.separator"));
+        // Should we flush() here?  It's probably faster not too as the writer
```
daudich, line 243: I would argue on the "yes" side. If the thread (or program) gets killed before the buffer is written to file we don't end up with a corrupted record.

src/foam/dao/FileJournal.js
```diff
@@ -283,11 +286,18 @@ foam.CLASS({
...
-        try ( BufferedReader reader = getReader() ) {
+        try {
+          BufferedReader reader = new BufferedReader(new FileReader(getFile()));
+
           for ( String line ; ( line = reader.readLine() ) != null ; ) {
             if ( SafetyUtil.isEmpty(line)        ) continue;
             if ( COMMENT.matcher(line).matches() ) continue;
 
+            if ( ! SafetyUtil.isEmpty(getPrefix()) ) {
+              if ( ! line.startsWith(getPrefix()) ) continue;
```
daudich, line 297: So this would mean that we re-reading the entire journal file for every DAO? O(nm) : n is the number for lines in the journal and m is the total number of DAOs?

src/foam/dao/FileJournal.js
daudich: Please disregard my previous comment :)  From our discussion this afternoon: this wouldn't matter because it would never occur (famous last words) that two records being written in the same journal for two different DAOs are dependent on the order of being written and read.

The reason that this wouldn't occur is because the only that would need to happen is if the `put_` is being overridden in a decorator and the logic `put`s on two DAOs. `put` is `synchronized` and there only ever exists one object for the DAO in the context.

This has to be noted somewhere in the code.

Cheers!

jlhughes:
Superceded by RoutingJournal work by @KernelDeimos - after discussions with Adam.

PR #1923, opened 2019-02-22
sendTo - on emailTemplate

Nauna:
adding property and functionality for setting sendTo on email templates

Requirement Came From:

Needing to send an email to a specific address. The work around was to hard code the emailMessage and the address, and use the emailService.sendEmail(x,EmailMessage).

However this is not a dynamic or scalable solution.

So instead by adding the sendTo property in the emailTemplate we can now specify the default address to send an email to, use the existing emailService.sendEmailFromTemplate(...) and have one location where email information exists (templates).

### DIFF
--- a/src/foam/nanos/notification/email/DAOEmailService.js
+++ b/src/foam/nanos/notification/email/DAOEmailService.js
@@ -127,7 +127,7 @@ if (foam.util.SafetyUtil.isEmpty(emailMessage.getSubject())) {
 // and sendTo is provided in the template, use the sendTo from template
 String sendTo = emailTemplate.getSendTo();
 if ( foam.util.SafetyUtil.isEmpty(emailMessage.getTo()) &&
-! foam.util.SafetyUtil.isEmpty(emailTemplate.getsendTo()) ) {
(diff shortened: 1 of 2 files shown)


Nauna:
This and one other PR that is waiting for this merge closes 
https://nanopay.atlassian.net/browse/CPF-240

jlhughes (approved):

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -123,10 +123,16 @@ if (foam.util.SafetyUtil.isEmpty(emailMessage.getSubject())) {
   emailMessage.setSubject(templateSubject.render(model));
 }
 
+// If emailTemplate has a send To address use that.
+String sendTo = emailTemplate.getSendTo();
+if (sendTo != null && !foam.util.SafetyUtil.isEmpty(sendTo)) {
```
jlhughes: Sorry if I wasn't clear on this earlier.  Similar to displayName, if the emailMessage does not have a sendTo, then use the template.

jlhughes (changes requested):

jlhughes (approved):

PR #1925, opened 2019-02-22
Adding the ability to read .0 journals from a different directory than JOURNAL_HOME

daudich:
We need a way to specify reading of initial or initial setup journals from the Jar's resources directory instead of the `JOURNAL_HOME` directory.


jlhughes:
Does this work? Storage needs support loading a file from resource, such as:
```
ClassLoader classLoader = getClass().getClassLoader();
File file = new File(classLoader.getResource("file/test.xml").getFile());

```

src/foam/dao/java/JDAO.js
```diff
@@ -32,11 +37,20 @@ foam.CLASS({
               .setCreateFile(true)
               .build());
 
-            // create a composite journal of repo journal
-            // and runtime journal and load them all
+            /* Create a composite journal of repo journal and runtime journal
+              and load them all.*/
+            X initialStorageX = x;
+            try {
```
adamvy: Do we really need a try/catch for this?  I think a simpler if/else is better than blindly catching all possible exceptions.

src/foam/dao/java/JDAO.js
```diff
@@ -32,11 +37,20 @@ foam.CLASS({
...
+              String initialStorageSource = System.getProperty("INITIAL_JOURNAL_SOURCE");
+              initialStorageX = x.put(foam.nanos.fs.Storage.class,
+                  new Storage(initialStorageSource));
```
adamvy: What is the format we need for INITIAL_JOURNAL_SOURCE to point to a resource in a JAR?

jlhughes: "resource"?

it results in a classLoader call so it's not a complete path. 
path part would be "journals/name.0"

It's rather more just a flag to load from resource.

src/foam/dao/java/JDAO.js
daudich: But it does throw exceptions. Re-throwing doesn't make sense. If the string isn't set or is empty, then it should default to regular rather than fail JDAO completely.

jlhughes: The dot.zero file does not need to exist.

src/foam/dao/java/JDAO.js
daudich: Using a classLoader in Storage now.

jlhughes (approved):

jlhughes (approved):

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -62,7 +62,8 @@ bodyAsByteArrayIsSet_ = false;`
     {
       class: 'String',
       name: 'sendTo',
-      documentation: `This property will set who the email is being sent too.`
+      documentation: `This property will set to whomever the email is being sent
+        to.`
```
ronin00, line 66: Isn't this injecting linefeed into the string and splitting it over 2 lines?

src/foam/dao/java/JDAO.js
```diff
@@ -32,11 +37,17 @@ foam.CLASS({
...
-            // create a composite journal of repo journal
-            // and runtime journal and load them all
+            /* Create a composite journal of repo journal and runtime journal
+              and load them all.*/
+            X resourceStorageX = x;
+            if ( System.getProperty("RESOURCES_JOURNAL_SOURCE") != null ) {
+              resourceStorageX = x.put(foam.nanos.fs.Storage.class,
+                  new Storage(System.getProperty("RESOURCES_JOURNAL_SOURCE"), true));
+            }
+
```
ronin00: Can RESOURCE_JOURNAL_SOURCE property be not set?  What happens then? Or does it matter -- just assume its set?

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -123,11 +123,11 @@ if (foam.util.SafetyUtil.isEmpty(emailMessage.getSubject())) {
...
 
-// If sendTo isn't set in the message
-// and sendTo is provided in the template, use the sendTo from template
+/* If To isn't set in the message and sendTo is provided in the template, use
+  the sendTo from template. */
 String sendTo = emailTemplate.getSendTo();
-if ( foam.util.SafetyUtil.isEmpty(emailMessage.getTo()) &&
-  ! foam.util.SafetyUtil.isEmpty(emailTemplate.getsendTo()) ) {
+if ( emailMessage.getTo().length == 0  &&
+  ! foam.util.SafetyUtil.isEmpty(emailTemplate.getSendTo()) ) {
```
ronin00, line 130: Why this change?  Just puzzled.

src/foam/nanos/notification/email/EmailTemplate.js
jlhughes: yes.  There are foam tools to build documentation from the documentation properties, not sure how things are rendered.

src/foam/nanos/notification/email/DAOEmailService.js
jlhughes: The PR for the Email was merge earlier today so guessing Dhiren's was having git/merge issues.

src/foam/nanos/notification/email/EmailTemplate.js
ronin00: I guess if its rendered as HTML the LF will be ignored.

src/foam/nanos/notification/email/DAOEmailService.js
daudich: The property `to` in EmailMessage.js is an array, whereas `isEmpty` assumes it is a string. This was throwing compilation errors. Hence, I had to change the logic to fix things.

src/foam/nanos/notification/email/EmailTemplate.js
daudich: I split it over two lines to maintain the code style. The line feed is ignored during rendering.

PR #1929, opened 2019-02-24
When Storage request is jar resource read as InputStream

jlhughes:


adamvy:
This works for now.

In the future I'd like to remove this conditional from Storage and FileJournal.  I think we need a separate journal implementation for reading from the JAR resources, as it stands FileJournal won't work as a real journal if you set it to a resource, since it can't write to the file, only replay it.  I will look at cleaning this up as part of my Routing Journal work.

Issue #1897, opened 2019-02-11
DIG doesn't encode query correctly.


kgrgreer:
DIG isn't encoding the query string properly. So if my query is countryId=CA, it writes the URL as q=countryId=CA, which doesn't work. It should be encoded and be q=countryId%3DCA to work properly,

FYI: @nanoscott

minsunny:
@jlhughes @kgrgreer This can be close? Seems working well and already encoded inside

kgrgreer:
I think it's better if we do encode it. The = character isn't a problem but the '&' symbol would be.

kgrgreer:
Example: if my query was:
businessName:Simon & Simon

Then the '&' would interfere with the URL and it would think everything after the & was the next argument.

kgrgreer:
Ex.: this would be an invalid URL ?daoKey=businessDAO&q:Simon&Simon

nanoscott:
This should be closed and we should not create issues or track them within GitHub issues. I have created a corresponding issue JIRA so that we can track it in our current sprint.

https://nanopay.atlassian.net/browse/CPF-336

@kgrgreer .. please close this bug within GitHub issues.

minsunny:
It has been merged.
https://github.com/foam-framework/foam2/pull/1934/files

PR #1934, opened 2019-02-25
Fixed getting the query part encoded on DIG

minsunny:
https://github.com/foam-framework/foam2/issues/1897

https://nanopay.atlassian.net/browse/CPF-336

if, its query part is encoded not to interfere with URL

### DIFF
--- a/src/foam/nanos/dig/DIG.js
+++ b/src/foam/nanos/dig/DIG.js
@@ -131,33 +131,30 @@ foam.CLASS({
         if ( subject ) {
           url += query ? "&" : "?";
           query = true;
-          url += "subject=" + subject;
+          url += "subject=" + encodeURIComponent(subject);
         }
         if ( q ) {
-          //q_ is encoded query part
-          q_ = query ? "&" : "?";
+          url += query ? "&" : "?";
           query = true;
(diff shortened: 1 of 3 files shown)


src/foam/nanos/dig/DIG.js
```diff
@@ -133,12 +134,15 @@ foam.CLASS({
           url += "subject=" + subject;
         }
         if ( q ) {
-          url += query ? "&" : "?";
+          //q_ is encoded query part
+          q_ = query ? "&" : "?";
           query = true;
-          url += "q=" + q;
+          q_ += "q=" + encodeURIComponent(q);
```
kgrgreer: Couldn't we just change this line to:
url += "q=" encodeURIComponent(q);

and keep the rest of the code the same?

minsunny: At the end of the code we are supposed to encode the whole URI. After encoding query part's "&" to %26 then it is encoded again. So I separate and encoded query part from the URl.

kgrgreer: Have you tested? Create an object with an '&' some field (maybe a script or test) and then see if this works.

PR #1935, opened 2019-02-25
Json property auth

KateMengjunXia:
User PermissionedOutputter in DigWebAgent.java to filter out properties based on permission settings.
Related PR: https://github.com/nanoPayinc/NANOPAY/pull/6324


src/foam/lib/json/Outputter.java
```diff
@@ -251,9 +251,13 @@ protected void outputDate(java.util.Date date) {
   }
 
   protected Boolean maybeOutputProperty(FObject fo, PropertyInfo prop, boolean includeComma) {
+
+    if ( ! this.propertyPredicate(fo, prop) ) {
```
kgrgreer: No need for 'this.' in Java.

src/foam/lib/json/Outputter.java
```diff
@@ -265,6 +269,10 @@ protected Boolean maybeOutputProperty(FObject fo, PropertyInfo prop, boolean inc
     return true;
   }
 
+  protected  boolean propertyPredicate( FObject fo, PropertyInfo prop) {
+    return ( outputDefaultValues_ || prop.isSet(fo) );
```
kgrgreer: No need for parenthesis since there's no large expression with a higher priority.

src/foam/lib/json/PermissionedOutputter.java
```diff
@@ -0,0 +1,26 @@
...
+
+  public PermissionedOutputter(foam.core.X x, OutputterMode mode) {
+    super(mode);
+    this.x_ = x;
+  }
+
+  @Override
+  protected boolean propertyPredicate(FObject fo, PropertyInfo prop) {
+    if ( prop.getPermissionRequired() ) {
+      AuthService auth = (AuthService) this.x_.get("auth");
```
kgrgreer: Could lookup 'auth' in constructor.

src/foam/lib/json/PermissionedOutputter.java
```diff
@@ -0,0 +1,26 @@
...
+      String propName = prop.toString();
```
kgrgreer: These next two lines generate a lot of garbage which will take more CPU to allocate and cleanup than the work you're doing. Couldn't you just do prop.getName()?

src/foam/lib/json/Outputter.java
```diff
@@ -265,6 +269,10 @@ protected Boolean maybeOutputProperty(FObject fo, PropertyInfo prop, boolean inc
     return true;
   }
 
+  protected  boolean propertyPredicate( FObject fo, PropertyInfo prop) {
```
kgrgreer: This method should be in an interface: PropertyPredicate. Then this class can use the Strategy pattern, not the Template pattern, as you've done. This would mean that you could reuse the PermissionedPropertyPredicate for the XMLOutputter and other Outputters as well. Also, we could then get rid of the mode_ variable and make Network and Storage be constant predicates.

KateMengjunXia: > This method should be in an interface: PropertyPredicate. Then this class can use the Strategy pattern, not the Template pattern, as you've done. This would mean that you could reuse the PermissionedPropertyPredicate for the XMLOutputter and other Outputters as well. Also, we could then get rid of the mode_ variable and make Network and Storage be constant predicates.

Why Strategy pattern? Isn't PermissionedPropertyPredicate identical for JSON outputter, XML outputter, etc?

kgrgreer: Yes, PermissionedPropertyPredicate should be identical for all outputters. But you don't have that interface here, it's just a template method that you override by subclassing. You don't have it as a standalone reusable object.

src/foam/lib/NetworkPropertyPredicate.js
```diff
@@ -0,0 +1,35 @@
+foam.CLASS({
+    package: 'foam.lib',
+    name: 'NetworkPropertyPredicate',
+    extends: 'foam.lib.PermissionedPropertyPredicate',
```
mcarcaso: Why is this extending a `PermissionedPropertyPredicate`? I'm not sure I see why they should be related.

src/foam/lib/NetworkPropertyPredicate.js
```diff
@@ -0,0 +1,35 @@
...
+    javaImports: [
+        'foam.nanos.auth.AuthService',
+        'foam.lib.PermissionedPropertyPredicate'
+    ],
+    
+    methods: [
+      {
+        name: 'propertyPredicateCheck',
```
mcarcaso, line 11: Since this is part of the interface, you don't need the type, or args. They're pulled in automatically.

src/foam/lib/PermissionedPropertyPredicate.js
```diff
@@ -0,0 +1,40 @@
...
+    package: 'foam.lib',
+    name: 'PermissionedPropertyPredicate',
+    implements: [ 'foam.lib.PropertyPredicate'],
+    javaImports: [
+      'foam.nanos.auth.AuthService'
+    ],
+    
+    methods: [
+      {
+        name: 'propertyPredicateCheck',
```
mcarcaso, line 11: Same here. Delete the type and args and let FOAM fill them in by you implementing the interface.

src/foam/lib/PropertyPredicate.js
```diff
@@ -0,0 +1,26 @@
+foam.INTERFACE({
+  package: 'foam.lib',
+  name: 'PropertyPredicate',
+
+  methods: [
+    {
+      name: 'propertyPredicateCheck',
+      type: 'boolean',
```
mcarcaso: type: 'Boolean'

The bulk of the type resolution happens in here: https://github.com/foam-framework/foam2/blob/master/src/foam/core/Type.js

Primitives like Boolean, Integer, Float, etc. are special and know how to output in both swift and java. Use those names when referring to primitives. Use either those special types or model IDs only. If it's not a modelled object, use javaType (assuming you're doing java dev).

src/foam/lib/PropertyPredicate.js
```diff
@@ -0,0 +1,26 @@
...
+      args: [
+        {
+          name: 'x',
+          type: 'foam.core.X'
```
mcarcaso: type: 'Context'

src/foam/lib/PropertyPredicate.js
```diff
@@ -0,0 +1,26 @@
...
+        },
+        {
+          name: 'fo',
+          type: 'FObject'
+        },
+        {
+          name: 'prop',
+          type: 'foam.core.PropertyInfo'
```
mcarcaso: `foam.core.PropertyInfo` is not a modelled object. It only exists for java so use `javaType` instead of `type`.

src/foam/lib/StoragePropertyPredicate.js
```diff
@@ -0,0 +1,35 @@
...
+    extends: 'foam.lib.PermissionedPropertyPredicate',
+    javaImports: [
+        'foam.nanos.auth.AuthService',
+        'foam.lib.PermissionedPropertyPredicate'
+    ],
+    
+    methods: [
+      {
+        name: 'propertyPredicateCheck',
+        type: 'boolean',
```
mcarcaso: Get rid of type and args here too

src/foam/lib/StoragePropertyPredicate.js
```diff
@@ -0,0 +1,35 @@
+foam.CLASS({
+    package: 'foam.lib',
+    name: 'StoragePropertyPredicate',
+    extends: 'foam.lib.PermissionedPropertyPredicate',
```
mcarcaso: Why extend PermissionedPropertyPredicate here too?

src/foam/box/HTTPBox.js
```diff
@@ -103,8 +103,8 @@ foam.CLASS({
         cls.extras.push(foam.java.Code.create({
           data: `
 protected class Outputter extends foam.lib.json.Outputter {
```
mcarcaso: Is there a reason foam.lib.json.Outputter isn't modelled? If it was, we could use the class property here instead of javaExtras.

src/foam/lib/json/Outputter.java
```diff
@@ -32,33 +34,43 @@ protected SimpleDateFormat initialValue() {
     }
   };
 
-  protected PrintWriter   writer_;
-  protected OutputterMode mode_;
```
mcarcaso, line 36: Seems like we're ditching the idea of having predefined outputter configurations. Why?

src/foam/lib/NetworkPropertyPredicate.js
KateMengjunXia: Will make PermissionedPropertyPredicate, NetworkPropertyPredicate and StoragePropertyPredicate independent implementations of the interface, and add an implementation that allows a combination of predicates.

adamvy:
Overally I like this PR, my main concern is the many different outputter classes for each use case.  It might be easier and more extensible to do an EasyOutputter facade similar to EasyDAO with setCheckPermissions(true).setStorageTransient(true).setNetworkTransient(true) which then builds the appropriate predicate as needed.

PR #1936, opened 2019-02-25
Adding journaltypes for easydao

daudich:
This closes: https://github.com/foam-framework/foam2/pull/1915

This is needed for the RoutingJournals redesigning.


src/files.js
```diff
@@ -24,551 +24,553 @@
 if ( typeof window !== 'undefined' ) global = window;
 
 FOAM_FILES([
```
adamvy: This is a huge diff.  Can you look at disabling your JS formatter?  Until FOAM decides upon a standard code formatter to use better to just let code live as is.

In fact the style guide the FOAM uses recommends leaving style changing changes to their own PRs.

https://google.github.io/styleguide/jsguide.html#policies-reformatting-existing-code

adamvy:
The actual functional changes look fine but I would really prefer if you didn't run all the code through a formatter.  What is actually a ~200line diff looks like a 600 line diff.

src/files.js
daudich: Fixed

src/foam/dao/EasyDAO.js
```diff
@@ -61,7 +61,11 @@ foam.CLASS({
     'foam.dao.RequestResponseClientDAO',
     'foam.dao.SequenceNumberDAO',
     'foam.dao.SyncDAO',
-    'foam.dao.TimingDAO'
+    'foam.dao.TimingDAO',
+    {
+      path: 'foam.dao.JournalType',
+      flags: ['java']
```
adamvy: Any reason why we have to flag this for only java?  Would be better to just switch to JournalType for both JS and Java.

daudich: Since journals are only being used on the server side, and we are trying to limit the amount of JS being served to the client. I thought this is what we wanted.

adamvy: You're not going to save a lot of code by stripping this requirement.  Basically just the size of the JournalType .js file will be saved.  But at the cost of decreased understandability of the codebase.  Consistency between platforms aids in understandability, so better to keep them consistent when we can.

daudich: Removed

kgrgreer (approved):

PR #1937, opened 2019-02-26
MultiView

Nauna:
Requirement have a view that handles multiple views for a single property.

Difference to DualView by fact that MultiView can handle any number of views per property, verses DualView which can handle a max of 2 views.


kgrgreer:
Nice!

Issue #1910, opened 2019-02-16
Report when Relationship TargetDAOKey not found 


jlhughes:
The Relationship TargeDAOKey, when not explicitly specified, is calculated based on the target property.  If this is incorrect, then a NullPointerException is generated when the missing TargetDAO is referenced.  We know when this occurs so add reporting/logging.

kgrgreer:
The following commit adds one check, but there are several other places that need similar support:

commit 4124b06f774ced6384a7f95119e801e08ab07e0a (HEAD -> master, origin/master, origin/HEAD)
Author: Kevin Greer <kgrgreer@gmail.com>
Date:   Fri Feb 22 10:58:00 2019 -0500

    Add assert that targetDAOKey is present.

PR #1939, opened 2019-02-26
Added alert when targetDAOKey not found

minsunny:
https://github.com/foam-framework/foam2/issues/1910

But swiftFactory parts are not handled yet.

### DIFF
--- a/src/foam/dao/Relationship.js
+++ b/src/foam/dao/Relationship.js
@@ -403,7 +403,8 @@ foam.CLASS({
       label: '',
       factory: function() {
         var targetDAO = this.__context__[this.targetDAOKey];
-        foam.assert(targetDAO, 'Missing DAO for targetDAAOKey', this.targetDAOKey);
+        foam.assert(targetDAO, 'Missing DAO for targetDAOKey', this.targetDAOKey);
+
         return foam.dao.ReadOnlyDAO.create({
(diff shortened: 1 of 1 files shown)


src/foam/dao/Relationship.js
```diff
@@ -403,21 +403,30 @@ foam.CLASS({
...
+        try {
+          return new foam.dao.ReadOnlyDAO.Builder(getX()).
+            setDelegate(new foam.dao.ManyToManyRelationshipDAO.Builder(getX()).
+              setRelationship(this).
+              setDelegate((foam.dao.DAO)getX().get(getTargetDAOKey())).
+              build()).
+            build();
+        } catch ( NullPointerException e ) {
+          foam.nanos.logger.Logger logger = (foam.nanos.logger.Logger) getX().get("logger");
+          logger.error("TargetDAOKey", getTargetDAOKey(), "not found.", e);
```
jlhughes, line 425: Any way to output what the relationship is? Else the error is hard to track down.

minsunny: screen shot 2019-02-27 at 3 20 35 pm

When I set wrong targetKey and test many to many relationship on User, I got above assert.

jlhughes: That's enough, thanks.

jlhughes (approved):

PR #1941, opened 2019-02-27
Ruler enhancement POC - Part3

lchanmann:
Storing rule execution history: https://docs.google.com/document/d/1SyenzyRPf-iaBEIHO3b0K1IZ1gkaaWba7SrJjplEW_4/edit#heading=h.h2q17ycyjj7h


src/foam/nanos/ruler/Rule.js
```diff
@@ -180,7 +192,9 @@
       javaCode: `
         getAsyncAction().applyAction(x, obj, oldObj, ruler);
         if ( ! getAfter()
-          && Operations.REMOVE != getOperation()
+          && Operations.CREATE == getOperation()
+          || Operations.UPDATE == getOperation()
+          || Operations.CREATE_OR_UPDATE == getOperation()
         ) {
           ruler.getDelegate().put_(x, obj);
```
jlhughes, line 199: this is an asynchronous, applyAction should return a promise so you can perform the getDelegate.put on it's completion.  Or does applyAction already do this when it completes?

src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -61,18 +62,66 @@ public void stop() {
...
         for (Rule rule : rules) {
           if ( stops_.get() ) return;
 
-          if ( rule.f(getX(), obj, oldObj) ) {
+          currentRule_ = rule;
+          if ( rule.f(getX(), obj, oldObj)
+            && rule.getAsyncAction() != null
+          ) {
             rule.asyncApply(getX(), obj, oldObj, RuleEngine.this);
+            saveHistory(rule, obj);
```
jlhughes, line 103: saveHistory saves the result of the rule, but with async, it's not complete yet.  Believe apply should saveHistory on completion.

src/foam/nanos/ruler/Rule.js
lchanmann: This block will be run inside a thread pool on the server side. https://github.com/lchanmann/foam2/blob/RulerEnhancementPOC_part3/src/foam/nanos/ruler/RuleEngine.java#L102.

src/foam/nanos/ruler/RuleEngine.java
lchanmann: This is inside the thread pool and it would complete as is.

jlhughes (approved):

PR #1945, opened 2019-03-04
Email refactor

Nauna:
Idea is to clean up where the email properties are being set. 

This PR conglomerates the property setting into the 'email' service and takes it out of the property settings from SMTPEmailService.

With this PR email implementation:
email(1) ->emailMessageDAO(2) -> smtpEmailService(3)

Function of:
(1) email - set the properties and template for the emailMessage. Using a priorities hierarchy.
(2) emailMessageDAO -  storing the emails that are being sent and calling smtpEmailService
(3) smtpEmailService - creating a MimeMessage that is ready for transport

Issue: setting email default values. 
This used to be set on the SMTPEmailService, however because of migrating the properties for emailMessages to one location, we are now populating the email, in the email(Service). 

Since this is on foam, and we don't want business defaults in foam, email(Service) is overridden in this PR: https://github.com/nanoPayinc/NANOPAY/pull/6330

to set the default nanopay email properties for {from, displayName, replyTo}. 

Added Feature:
-adding replyTo on EmailTemplates for easier customization of emails being sent.


src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +111,173 @@ return config_;`
...
 
-// If the displayName doesn't set in the message
-// and the displayName provided in the template, use the displayName from template
-if ( foam.util.SafetyUtil.isEmpty(emailMessage.getDisplayName()) &&
-  ! foam.util.SafetyUtil.isEmpty(emailTemplate.getDisplayName()) ) {
-  JtwigTemplate templateDisplayName = JtwigTemplate.inlineTemplate(emailTemplate.getDisplayName(), config);
-  emailMessage.setDisplayName(templateDisplayName.render(model));
-}
+        if ( user == null || user.getId() == 0 ) {
+          throw new RuntimeException("User Error @ DAOEmailService.fillInEmailProperties()");
```
jlhughes: please generate a log message with the real error - that user is null at this point - with a stack trace. 
Same with those below, please.

jlhughes (changes requested):
Couple comments, looks good so far. I'd like to @TW80000  to review as well.

src/foam/nanos/notification/email/DAOEmailService.js
Nauna: will figure out tomorrow, but if someone reviews this and knows how to print a stack trace outside of a catch block ... well that would be helpful.

jlhughes: re: stack trace - log.warning(message, new Exception())

src/foam/nanos/notification/email/SMTPEmailMessageDAO.js
```diff
@@ -9,11 +9,15 @@ foam.CLASS({
...
+  javaImports: [
+    'foam.nanos.notification.email.SMTPEmailService'
+  ],
+
   methods: [
     {
       name: 'put_',
       javaCode: `
-        EmailService service = (EmailService) x.get("smtpEmailService");
+      SMTPEmailService service = (SMTPEmailService) x.get("smtpEmailService");
```
TW80000: Indentation is wrong. Missing two spaces at the beginning of the line.

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -57,13 +58,20 @@ bodyAsByteArrayIsSet_ = false;`
     },
     {
       class: 'String',
-      name: 'displayName'
+      name: 'displayName',
+      documentation: `Displayed as the name in the email from field.`
```
TW80000, line 62: Is the image below right?

screen shot 2019-03-05 at 8 59 12 am

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
-  ! foam.util.SafetyUtil.isEmpty(emailTemplate.getSendTo()) ) {
-  emailMessage.setTo(new String[] {sendTo});
-}
+        sendEmail(x, emailMessage);
+      `
+    },
+    {
+      name: 'fillInEmailProperties',
+      documentation: `
+      Prioritisations:
```
TW80000: Should be "Prioritization:". Or we could say "Precedence" or "Order of precedence" here instead. Personally I think precedence is closer to what we're defining here.

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
+          logger.warning("User null or unverified through DAO: failing at DAOEmailService.fillInEmailProperties()", new Exception());
+        }
 
-sendEmail(x, emailMessage);
-`
+        EnvironmentConfiguration config = getConfig(user.getGroup());
+        DAO groupDAO   = (DAO) x.get("groupDAO");
+        Group group    = (Group) groupDAO.find(user.getGroup());
+
+        if ( group == null || group.getId() == "0" ) {
```
TW80000: We don't need to check the group's id here. The null check if sufficient.

(Also, "0" is a valid group name. "" isn't.)

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
-}
+        if ( user == null || user.getId() == 0 ) {
+          logger.warning("User null or unverified through DAO: failing at DAOEmailService.fillInEmailProperties()", new Exception());
+        }
 
-sendEmail(x, emailMessage);
-`
+        EnvironmentConfiguration config = getConfig(user.getGroup());
+        DAO groupDAO   = (DAO) x.get("groupDAO");
+        Group group    = (Group) groupDAO.find(user.getGroup());
```
TW80000, line 174: Missing `.inX(x)`

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
+
+        if ( group == null || group.getId() == "0" ) {
+          logger.warning("group null or unverified through DAO: failing at DAOEmailService.fillInEmailProperties()", new Exception());
```
TW80000, line 177: What do you mean by "unverified"? Why don't we just say "Group not found."?

Also, why is a `new Exception()` being included in the warning with no message?

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
+          } else {
+            emailMessage.setDisplayName(
+              ! SafetyUtil.isEmpty(group.getDisplayName()) ? group.getDisplayName() : getDisplayName()
+            );
+          }
+        }
+
+        // SUBJECT:
+        // Since subject is very specific to each email there is no group field or default value for this property.
+        if ( foam.util.SafetyUtil.isEmpty(emailMessage.getSubject()) &&
```
TW80000, line 220: When conditions span multiple lines, I find the following format easier to read:

```Java
if (
  foam.util.SafetyUtil.isEmpty(emailMessage.getSubject()) &&
  ! foam.util.SafetyUtil.isEmpty(emailTemplate.getSubject())
) {
  // Body...
}
```

It makes it visually clear what is part of the condition and what is part of the body so it's easier to scan.

Just a suggestion though.

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
+          ! foam.util.SafetyUtil.isEmpty(emailTemplate.getSubject()) ) {
+            JtwigTemplate templateSubject = JtwigTemplate.inlineTemplate(emailTemplate.getSubject(), config);
+            emailMessage.setSubject(templateSubject.render(model));
+        }
+
+        // SEND TO:
+        // Since sendTo is very specific to each email there is no group field or default value for this property.
+        if ( emailMessage.getTo().length == 0 &&
+          ! foam.util.SafetyUtil.isEmpty(emailTemplate.getSendTo()) ) {
+            JtwigTemplate templateSendTo= JtwigTemplate.inlineTemplate(emailTemplate.getSendTo(), config);
```
TW80000: Missing space before `=`

mcarcaso:
I think we can simplify things a bit if we structure some of this differently.

The EmailService interface should lose the `sendEmailFromTemplate` method. An email service's concern is only to send an email. You can construct an email using a template before handing it to the EmailService.

The DAOEmailService should essentially just be `dao.put(email)`. It shouldn't care about filling an email or anything else.

For putting together an email from a template, we'll probably need a new interface with a method similar to your `fillInEmailProperties` method but I'm not 100% sure that's the right signature exactly. I'll have to put a little more thought into that and we can discuss ideas for that offline.

I'm not sure why we need the SMTPEmailMessageDAO if all email sending goes through the emailService. When/why do we need that?

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
+          name: 'emailMessage',
+          javaType: 'final foam.nanos.notification.email.EmailMessage'
+        }
+      ],
+      javaCode:
+        `
+        foam.nanos.session.Session session = x.get(foam.nanos.session.Session.class);
+
+        DAO userDAO         = (DAO) x.get("localUserDAO");
+        User user           = (User) userDAO.find(session.getUserId());
```
TW80000, line 256: Missing `.inX(x)`

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
+
+        // 1. If the user doesn't login at this time, get the user from localUserDao
+        // 2. If the user is the system user, get the real user from localUserDao
+        if ( user == null || user.getId() == 1 ) {
+
+          Sink sink = new ArraySink();
+          sink = userDAO.where(MLang.EQ(User.EMAIL, emailMessage.getTo()[0]))
```
TW80000: Missing `.inX(x)`

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
+            .limit(1).select(sink);
+
+          List list = ((ArraySink) sink).getArray();
+          if ( list == null || list.size() == 0 ) {
```
TW80000: Instead of selecting with limit 1 and checking if the array had a result or not, you can just pass an mlang to `find` and it will do the same thing for you behind the scenes.

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +112,173 @@ return config_;`
...
+        Logger logger = (Logger) x.get("logger");
 
-// If the displayName doesn't set in the message
-// and the displayName provided in the template, use the displayName from template
-if ( foam.util.SafetyUtil.isEmpty(emailMessage.getDisplayName()) &&
-  ! foam.util.SafetyUtil.isEmpty(emailTemplate.getDisplayName()) ) {
-  JtwigTemplate templateDisplayName = JtwigTemplate.inlineTemplate(emailTemplate.getDisplayName(), config);
-  emailMessage.setDisplayName(templateDisplayName.render(model));
-}
+        if ( user == null || user.getId() == 0 ) {
```
TW80000: This is already caught inside of `findUser` so `user` will never be null here.

TW80000 (changes requested):

src/foam/nanos/notification/email/EmailTemplate.js
Nauna: yes

jlhughes:
@mcarcaso regarding redesign, yes, we've had a couple discussions/tickets on this.  The DAO was grafted in and these few changes clean things up and get us closer to where we need to be without restricting future changes.  We can table additional refactoring based on your and Troy's design ideas.

src/foam/nanos/notification/email/DAOEmailService.js
Nauna: 👍

src/foam/nanos/notification/email/DAOEmailService.js
Nauna: inX() makes sense when creating/calling the dao, however not after, from my understanding.

from: foam2/doc/guides/StyleGuide.md
##### When using a DAO as a regular client from outside of a decorator for that DAO, use the non-CO methods and use inX() to specify the context. ie.
```javascript
   DAO userDAO = ((DAO) x.get("userDAO")).inX(x);
   userDAO.find(id);
```

src/foam/nanos/notification/email/DAOEmailService.js
Nauna: yup good call

Nauna:
@mcarcaso 
> I think we can simplify things a bit if we structure some of this differently.
> 
> The EmailService interface should lose the `sendEmailFromTemplate` method. An email service's concern is only to send an email. You can construct an email using a template before handing it to the EmailService.

So correct me if I'm wrong, but I think naming is the only issue here.
When we call email.service what is happening is creating the emailMessage from the template and grabbing necessary data from the order of precedences of properties.

email(service) === populateEmailMessageWithPrecedenceFields(service)

> The DAOEmailService should essentially just be `dao.put(email)`. It shouldn't care about filling an email or anything else.

We could separate out email(service).sendEmailFromTemplate() and then just have email(service).sendEmail() which does *getDao().inX(x).put(emailMessage);*

But I personally already don't like that we have a chain of services that are called for emails. 
### Email Services:
email --> emailMessageDAO --> smtpEmailService

if I'm not mistaken, the suggestion would then be for templated Emails:
templatePopulation --> email(dao.put(emsg);) --> emailMessageDAO(storing emails) --> smtpEmailService(transporting emails)

and then we'd also have the issue of calling a different service if we just wanted to send an EmailMessage directly - which we can do right now with email(service).sendEmail();

> ...
> I'm not sure why we need the SMTPEmailMessageDAO if all email sending goes through the emailService. When/why do we need that?

SMTPEmailMessageDAO really just passes the EmailMessage through to the next service on the chain(smtpEmailService).
This file is a decorator on the EmailMessageDAO(service).

src/foam/nanos/notification/email/DAOEmailService.js
TW80000: It makes no difference where you put it. If you want to put it where you define the variable then you don't need to put it at call sites and vice versa.

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -92,55 +113,159 @@ return config_;`
...
+        Logger logger = (Logger) x.get("logger");
+        foam.nanos.session.Session session = x.get(foam.nanos.session.Session.class);
+
+        DAO userDAO         = ((DAO) x.get("localUserDAO")).inX(x);
+        User user           = (User) userDAO.find(session.getUserId());
+
+        // 1. If the user doesn't login at this time, get the user from localUserDao
+        // 2. If the user is the system user, get the real user from localUserDao
+        if ( user == null || user.getId() == 1 ) {
+          user = (User) userDAO.find(MLang.EQ(User.EMAIL, emailMessage.getTo()[0]));
```
TW80000, line 261: I know it was this way when you started making changes, but there could be multiple Users/Businesses/Contacts in `localUserDAO` with the same email address. Shouldn't we update this condition to also check if `loginEnabled = true`? That way it will only give us the User since all contacts and businesses have `loginEnabled = false`. Alternatively, we could find in `localUserUserDAO` instead of `locaUserDAO` to accomplish the same thing.

TW80000 (changes requested):

mcarcaso:
Nah this is a bit more than just naming. It's separating the responsibilities in a more clear way. Currently, models that should be really simple have more responsibilities than they should and are too complicated as a result.

So after thinking about it, here's how I think this should be designed:


EmailService: Interface with only a `send(x, email)` method.

SMTPEmailService: Implementation of EmailService that attempts to send the email you give it via SMTP. It should not modify the email in any way during this process. If somebody tries to send an email from an email address that the SMTP credentials don't have permissions to send from then it errors out and that's that.

EmailServiceDAO: DAO decorator that, on put, sends the email to the emailService. It doesn't care what kind of email service it is. It just blindly takes the email and calls emailService.send(x,email).

EmailTemplate: Should have an apply(x, email) method. Calling this method copies any fields from the template into the given email.


With these components, the prod setup could be something like:
emailService = SMTPEmailService
emailMessageDAO = EmailServiceDAO(delegate: JDAO)

And the dev setup could be the same except with a NullEmailService, or with an emailMessageDAO that's simply a JDAO.

So how do you actually send emails from a template?
```
Email e = new Email.builder(x).setTo(someEmailAddress);
EmailTemplate et = emailTemplateDAO.find("some-template");
et.apply(x, e);
emailMessageDAO.put(e);
```

Maybe those components alone aren't enough to do what you want to do but I think we can get where you need to go with decorating the various components I listed above.

So the flow of data is something like:
email -> email dao -> email service

However, an alternative would be:
email -> email service -> email dao
To take this approach, you could get rid of the EmailServiceDAO and replace it with a DAOEmailService. An email service that simply puts the email into a dao. I'm kinda torn between the two approaches but I think the concerns of each model are much more clear under either of these designs.

src/foam/nanos/notification/email/DAOEmailService.js
jlhughes: The email system should only be aware of localUserDAO, so testing for loginEnabled seems a reasonable compromise.  All the other user daos are nanopay specific.  Or the DAOEmailService can take the userDAO as an argument, defaulting to localUserDAO, then nanopay can override it in service setup.

TW80000: Right, I forgot this was in FOAM. I think `loginEnabled` makes the most sense then.

TW80000:
> Nah this is a bit more than just naming. It's separating the responsibilities in a more clear way. Currently, models that should be really simple have more responsibilities than they should and are too complicated as a result.
> 
> So after thinking about it, here's how I think this should be designed:
> 
> EmailService: Interface with only a `send(x, email)` method.
> 
> SMTPEmailService: Implementation of EmailService that attempts to send the email you give it via SMTP. It should not modify the email in any way during this process. If somebody tries to send an email from an email address that the SMTP credentials don't have permissions to send from then it errors out and that's that.
> 
> EmailServiceDAO: DAO decorator that, on put, sends the email to the emailService. It doesn't care what kind of email service it is. It just blindly takes the email and calls emailService.send(x,email).
> 
> EmailTemplate: Should have an apply(x, email) method. Calling this method copies any fields from the template into the given email.
> 
> With these components, the prod setup could be something like:
> emailService = SMTPEmailService
> emailMessageDAO = EmailServiceDAO(delegate: JDAO)
> 
> And the dev setup could be the same except with a NullEmailService, or with an emailMessageDAO that's simply a JDAO.
> 
> So how do you actually send emails from a template?
> 
> ```
> Email e = new Email.builder(x).setTo(someEmailAddress);
> EmailTemplate et = emailTemplateDAO.find("some-template");
> et.apply(x, e);
> emailMessageDAO.put(e);
> ```
> Maybe those components alone aren't enough to do what you want to do but I think we can get where you need to go with decorating the various components I listed above.
> 
> So the flow of data is something like:
> email -> email dao -> email service
> 
> However, an alternative would be:
> email -> email service -> email dao
> To take this approach, you could get rid of the EmailServiceDAO and replace it with a DAOEmailService. An email service that simply puts the email into a dao. I'm kinda torn between the two approaches but I think the concerns of each model are much more clear under either of these designs.

I opened a PR a while back doing some of this: https://github.com/foam-framework/foam2/pull/1752

Particularly doing the `email -> email dao -> email service` approach you mentioned.

I'm not sure why it was dropped, I can't remember anymore.

Nauna:
Ok I will be going over everyone's points today and I believe we can all get what we want.

@TW80000 will be merging your PR into my branch. Combining our refactors.

Will reopen branch soon

Nauna:
going point by point:

> EmailService: Interface with only a `send(x, email)` method.
> 
> SMTPEmailService: Implementation of EmailService that attempts to send the email you give it via SMTP. It should not modify the email in any way during this process.

Ok does this with this PR.

>  If somebody tries to send an email from an email address that the SMTP credentials don't have permissions to send from then it errors out and that's that.

Doesn't currently error out, it just returns empty handed. We can add an error.

> EmailServiceDAO: DAO decorator that, on put, sends the email to the emailService. It doesn't care what kind of email service it is. It just blindly takes the email and calls emailService.send(x,email).

So right now let's consider EmailServiceDAO == emailMessageDAO. 
and emailMessageDAO.put() does the following because of a decorator SMTPEmailMessageDAO
```
SMTPEmailService service = (SMTPEmailService) x.get("smtpEmailService");
service.sendEmail(x, (EmailMessage) obj);
```
So the SMTPEmailService/emailMessageDAO do blindly take the email and call emailService.send

> EmailTemplate: Should have an apply(x, email) method. Calling this method copies any fields from the template into the given email.

This is an interesting concept but let's consider that there are four places that email properties can be set: emailMessage, emailTemplate, group properties, and then we have default values. 

In this PR all the setting of these properties in one location. Which allows us to priorities the setting of these properties. For example if we have an email template but want to reuse the template for an email but want to simply change the subject, we can do that by explicitly setting the subject on emailMessage (in the decorator/service)  before going through the email service pipeline.

> With these components, the prod setup could be something like:
> emailService = SMTPEmailService
> emailMessageDAO = EmailServiceDAO(delegate: JDAO)
> 
> And the dev setup could be the same except with a NullEmailService, or with an emailMessageDAO that's simply a JDAO.
> 
> So how do you actually send emails from a template?
> 
> ```
> Email e = new Email.builder(x).setTo(someEmailAddress);
> EmailTemplate et = emailTemplateDAO.find("some-template");
> et.apply(x, e);
> emailMessageDAO.put(e);
> ```
> Maybe those components alone aren't enough to do what you want to do but I think we can get where you need to go with decorating the various components I listed above.

This structure although interesting and modular is overtly complicated. Right now to send and email you have two choices:
```
EmailService email = (EmailService) x.get('email');
1) email.sendEmail(x, emsg)
2) email.sendEmailFromTemplate(x, user, emsg, "some-template", args);
```

Troy's PR #1752 does abstract out the template setting to what I understand to be diagrammed below:
image

But I am yet to be convinced that this is good/better. It does seem like what Troy had written in his PR is what Mike is talking about. Guys(@TW80000 and @mcarcaso ) can you please explain what makes this structure better than what we currently have.

> So the flow of data is something like:
> email -> email dao -> email service

This is the current flow (which has not been adjusted in my PR- properties and some stuff have changed but the flow was not adjusted), thus feel your suggestions are satisfied.

image

> However, an alternative would be:
> email -> email service -> email dao

I do think this last suggestion (email -> email service -> email dao) -is something that Kevin and Joel would agree to - adjusting the order in which these services take place, based on https://github.com/foam-framework/foam2/pull/1752#issuecomment-449472156 and https://github.com/foam-framework/foam2/pull/1752#issuecomment-450356705

Adjustments as I see:
So if everyone agrees, 
1) I could move around the services, so SMTPEmailService is called prior to a put into the EmailMessageDAO.

and

2) Set error handling in SMTPEmailService in case properties are not set correctly instead of just returning.

src/foam/nanos/notification/email/DAOEmailService.js
Nauna: loginEnabled is a simple check, so I've done it this way.

jlhughes:
re: email -> email service -> email dao
I would prefer email -> email dao -> email service to facilitate testing.  In non-PROD the email dao can be decorator to test for successful generation of email.  With this QA would not have to configure an smtp server.  Also this allows sending to queue if the smtp service goes offline; email messages could be flagged as 'sent', then on restart or service online queued emails can be processed.  Presently we don't handle smtp send rejections, but if we did, then the email message could be updated with the send failure.

mcarcaso:
Handling and resending SMTP rejections is something I haven't thought much about but if we want to handle that we should make sure the design allows for this in a good way. I always thought the emailMessageDAO was just to be used for logging purposes, not to actually have some meaningful behaviour like queuing and sending messages. If you want to queue, batch, and/or retry sending emails, those might be best as different EmailService decorators.

I think we should all sit down and discuss all of the end goals here and how to best design all of this. Let's take this offline for now.

PR #1948, opened 2019-03-05
Change the timeout box to 1 min.

sirenChen:
refs: 
https://nanopay.atlassian.net/browse/CPF-370

The temp fix is change the timeout to 1 min. Once @minsunny finish the cacheDAO, we can change this back to 5s.


kgrgreer:
Wouldn't it be safer to just remote the TimeoutBox rather than having retries a minute later?

PR #1949, opened 2019-03-05
Make scrolling speed proportional to input

TW80000:
Before this change, every single wheel event would scroll by at least 1 row. That made it difficult to scroll small amounts since even tiny gestures on a trackpad would cause several rows to be skipped.

This PR changes that behaviour so that it keeps track of the deltas from the events and only changes the rows when the delta is greater than or equal to the height of 1 row. This makes it much easier to control how fast you want to scroll.

### DIFF
--- a/src/foam/u2/view/ScrollTableView.js
+++ b/src/foam/u2/view/ScrollTableView.js
@@ -87,10 +87,9 @@
       `
     },
     {
-      type: 'Int',
+      class: 'Int',
       name: 'accumulator',
-      value: 0,
-      documentation: `Used internally to mimic native scrolling speed.`,
+      documentation: 'Used internally to mimic native scrolling speed.',
       adapt: function(_, v) {
         return v % this.rowHeight;
       }
--- a/src/foam/u2/view/ScrollTableView.js
+++ b/src/foam/u2/view/ScrollTableView.js
@@ -85,6 +85,15 @@
       documentation: `
         A reference to the table element we use in the fitInScreen calculations.
       `
+    },
+    {
+      type: 'Int',
+      name: 'accumulator',
+      value: 0,
+      documentation: `Used internally to mimic native scrolling speed.`,
+      adapt: function(_, v) {
+        return v % this.rowHeight;
+      }
     }
   ],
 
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/ScrollTableView.js
```diff
@@ -85,6 +85,15 @@
       documentation: `
         A reference to the table element we use in the fitInScreen calculations.
       `
+    },
+    {
+      type: 'Int',
+      name: 'accumulator',
+      value: 0,
+      documentation: `Used internally to mimic native scrolling speed.`,
```
kgrgreer: Faster to use ' instead of `, plus no need since it is just a single line string.

src/foam/u2/view/ScrollTableView.js
```diff
@@ -85,6 +85,15 @@
       documentation: `
         A reference to the table element we use in the fitInScreen calculations.
       `
+    },
+    {
+      type: 'Int',
+      name: 'accumulator',
+      value: 0,
```
kgrgreer: 0 is already the default.

src/foam/u2/view/ScrollTableView.js
```diff
@@ -141,11 +150,10 @@
...
       code: function(e) {
         var negative = e.deltaY < 0;
-        // Convert to rows, rounding up. (Therefore minumum 1.)
-        var rows = Math.ceil(Math.abs(e.deltaY) / 40);
-        var oldSkip = this.skip;
+        var rows = Math.floor(Math.abs(this.accumulator + e.deltaY) / this.rowHeight);
+        this.accumulator += e.deltaY;
         this.skip = Math.max(0, this.skip + (negative ? -rows : rows));
-        if ( e.deltaY !== 0 && oldSkip != this.skip ) e.preventDefault();
+        if ( e.deltaY !== 0 ) e.preventDefault();
```
kgrgreer, line 155: Would it be better if we just always prevented default?

TW80000: That would prevent us from scrolling horizontally.

src/foam/u2/view/ScrollTableView.js
TW80000: I thought so, but if I remove this line it breaks scrolling.

However, I just tried changing line 90 from `type: 'Int'` to `class: 'Int'` and that worked without the value being set explicitly. I'm assuming that's a bug.

src/foam/u2/view/ScrollTableView.js
TW80000: What makes it faster? I wouldn't expect there to be a performance difference.

PR #1950, opened 2019-03-05
Fix journaltype for rulerdao in services

daudich:


jlhughes:
Merging as the corresponding merge on EasyDAO has already occurred.

PR #1952, opened 2019-03-05
Modelling Group and User Auth Service

daudich:


jlhughes (approved):

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -0,0 +1,630 @@
...
+        context.`,
+      type: 'foam.nanos.auth.User',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'user',
+          of: 'foam.nanos.auth.User'
```
adamvy, line 210: Should be type:

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -0,0 +1,630 @@
...
+        and group enabled flags.`,
+      type: 'boolean',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'user',
+          of: 'foam.nanos.auth.User'
```
adamvy, line 327: Should be type:

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -0,0 +1,630 @@
...
+        },
+        {
+          name: 'permission',
+          of: 'java.security.Permission'
```
adamvy, line 331: should be javaType:

adamvy:
This doesn't compile.

Needs to be added to classes.js and there's a test in NANOPAY that needs to be updated.

I imagine you're not doing a clean build so your code is still picking up compiled class from the hand written code.

adamvy:
Once you do add it to classes.js you'll notice a bunch of new warnings when generating java code from the usage of "of:" instead of type/javaType.  Be sure to fix all those.

PR #1954, opened 2019-03-06
Limit input length when precision is specified

KateMengjunXia:


src/foam/u2/FloatView.js
```diff
@@ -54,10 +54,29 @@ foam.CLASS({
...
+      var data = this.data$;
+      var view = this.attrSlot(null, this.onKey ? 'input' : null);
+
+      if ( ! foam.Undefined.isInstance(this.data) ) view.set(this.formatNumber(this.data));
+
+      view.sub(function() {
+        var text = view.value;
+        data.set(this.textToData(text));
+
+        if ( text.indexOf('.') > 0 && text.length - text.indexOf('.') - 1 > this.precision ) {
```
kgrgreer: Would it be simpler and still work to do something like:
parseFloat(parseFloat(value).toFixed(this.precision));

src/foam/u2/FloatView.js
```diff
@@ -54,10 +54,29 @@ foam.CLASS({
...
+          view.set(text.substring(0, text.indexOf('.') + this.precision + 1));
+        }
+
+      }.bind(this));
```
kgrgreer: You are binding from the view to the data, but for the view to work properly you also need to bind from the data to the view.

src/foam/u2/FloatView.js
KateMengjunXia: If 0.0 is entered, then your code would force the view to show 0, but I think we should leave it like 0.0

src/foam/u2/FloatView.js
KateMengjunXia: Since data will change only in the case when view is changed, wouldn't it be more simple to just put the logic inside data.sub(...) here?

PR #1955, opened 2019-03-06
Add agent id to session

TW80000:
Without the agent id, there's no way to tell the difference between two
sessions where user A and user B are both acting as user C because only
the id of user C is saved on the session. This makes it impossible to
query the sessions where a particular user is an agent.

Originally I thought we could just set the userId to the agent's id, but
that would break a large number of places that expect the userId on the
session to be set to what it's currently set to.

Required for https://nanopay.atlassian.net/browse/CPF-188

### DIFF
--- a/src/foam/nanos/auth/AgentUserAuthService.java
+++ b/src/foam/nanos/auth/AgentUserAuthService.java
@@ -82,7 +82,7 @@ public class AgentUserAuthService
     // Set user and agent objects into the session context and place into sessionDAO.
     Session session = x.get(Session.class);
     session.setUserId(user.getId());
-    session.setUserId(agent.getId());
+    session.setAgentId(agent.getId());
     session.setContext(session.getContext().put("user", user));
     session.setContext(session.getContext().put("agent", agent));
     return user;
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -82,6 +82,7 @@ public User actAs(X x, User entity) throws AuthenticationException {
     // Set user and agent objects into the session context and place into sessionDAO.
     Session session = x.get(Session.class);
     session.setUserId(user.getId());
+    session.setUserId(agent.getId());
```
adamvy: Was this supposed to be session.setAgentId() ?

adamvy:
I was under the impression we were renaming userId/agentId to realUserId/effectiveUserId.  Or did we land somewhere else on that?

src/foam/nanos/auth/AgentUserAuthService.java
TW80000: Yes, thanks. Fixed by 7be3e47.

TW80000:
> I was under the impression we were renaming userId/agentId to realUserId/effectiveUserId. Or did we land somewhere else on that?

I think that's still the plan. I don't think anybody's been assigned to do it yet though.

kgrgreer:
More generally, we want to just store your "credentials" in the context and then have your credentials have methods for accessing your display vs. your auth id. Then we could decorate credentials to do things like actingAsAgent(realUser(kgr), realUser(troy)), or add new credential types for OAuth like tokens or machines or whatever.

PR #1958, opened 2019-03-07
Allow the parameter X to be an empty value

minsunny:
https://nanopay.atlassian.net/browse/CPF-393

- Under foam

Allow the parameter X to be an empty value or whatever.
Since the X is replaced with a real context value later.


deejes:
Looks good, I will test once and merge. Thanks!

deejes:
Tested, LGTM

jlhughes (approved):

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -101,7 +101,7 @@ public void execute(X x) {
               Class paramObj_ = null;
```
mcarcaso, line 101: This and the line below are unused. Can we get rid of them?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -101,7 +101,7 @@ public void execute(X x) {
...
               Object obj_ = null;
 
-              if ( ! SafetyUtil.isEmpty(p.getParameter(pArray[j].getName())) ) {
+              if ( ! SafetyUtil.isEmpty(p.getParameter(pArray[j].getName())) || typeName.equals("foam.core.X") ) {
```
mcarcaso, line 104: At first, it wasn't clear to me why this condition was being added here and I think it could warrant its own conditional block along with a comment explaining why. How does something like this look?
```
              if ( typeName.equals("foam.core.X") ) {
                // Contexts should not be passed in as arguments so force the use of
                // the current context when a foam.core.X argument is encountered.
                arglist[j] = x;
              } else if ( ! SafetyUtil.isEmpty(p.getParameter(pArray[j].getName())) ) {
                arglist[j] = setParameterType(x, resp, out, p, typeName, pArray[j]);
              } else {
```
If we do this, we can remove the `foam.core.X` case in the `setParameterType` method.

Somewhat unrelated: I was also being a bit thrown off by the name `setParameterType` because it's not really setting a type, it's setting a value based on the type. We might want to consider renaming that.

minsunny: Thank you, let me update when I have a chance asap.

deejes:
@minsunny going to leave it to you resolve mike's issues in another PR.

Thanks.

PR #1960, opened 2019-03-07
Fixed data view binding

KateMengjunXia:


adamvy:
Is there a more complete description or bug report on what exactly was broken with the old data binding?

KateMengjunXia:
It causes a stack overflow when FloatView's subviews are used (e.g Currency view is used when entering a transfer amount).
I don't think there's a ticket for this, since the issue was spotted right after the code had been merged

adamvy:
Cool, Joel linked me to https://github.com/foam-framework/foam2/pull/1954 as context, as I didn't see that go by earlier.

adamvy:
Kate and I were just discussing this.  There appears to be two separate tasks that are sort of intertwined here.

One is the interaction between the user and the actual input element.  Where we want to enforce that they can't type beyond a certain precision, and that they can't type letter and what not.  The other is between the InputElement and the "data" slot, where we need to convert the text to a float and vice versa.  It would be nice if there were a way to separate these two tasks a little more cleanly.

We could almost accomplish what we wanted via just .relateFrom() if FOAM didn't suppress the feedback automatically, and textToData/dataToText are perfect inverses of eachother.

Not worth doing for now as we have code that works and I'm not proposing a clear plan, but something to keep in mind for the next time anyone has to work on a low level view like this.

PR #1962, opened 2019-03-08
Group in context

TW80000:
## Problem definition

Currently when one user acts as another user, we take the group on the junction between them in and set the target user's group to it in the session.

The problem is that when we go to put the user in the context to `userDAO`, it looks exactly like an update where we're trying to set the group. Currently it gets caught in the authorization logic and throws an error saying you can't update someone else's group, which is good, but we don't want to be "updating" the group at all.

## Solution

Instead of setting User B's group to something else in the session, we just put a group directly in the context instead. This becomes the effective group that should be referenced almost everywhere instead of `user.group`.

Users still have groups, but now when we create a session we put a copy of the user's group as the group in the context, and if the user acts as someone else, we replace the group in the context with the group from the junction instead. The auth service and anything else that cares about reading the effective group should just reference the group in the context instead of the user's group.

## Notes
* This PR obviously only includes the changes on the FOAM side. I have some of the changes for the NANOPAY side done but not all of them yet. I'm not sure if that makes this DO_NOT_MERGE or not. I've tested and logging in to both nanopay's portal and Ablii work fine with the changes I have made, as does switching businesses in Ablii.
* Look at commit 8868279 if you want just the important changes. The other commit (f6fbaf1) is just linting and clean up.


src/foam/box/SessionClientBox.js
```diff
@@ -48,7 +48,7 @@ foam.CLASS({
         } else {
 
           // fetch the soft session limit from group, and then start the timer
-          if ( this.group.id !== "" && this.group.softSessionLimit !== 0 ) {
+          if ( this.group && this.group.id !== '' && this.group.softSessionLimit !== 0 ) {
             this.sessionTimer.startTimer(this.group.softSessionLimit);
```
kgrgreer, line 52: Unrelated to this PR, but what is the softSessionLimit and why do we need it?

src/foam/box/SessionServerBox.java
```diff
@@ -67,29 +70,27 @@ public void send(Message msg) {
...
+        // If this service has been configured to require authentication, then
+        // throw an error if there's no user in the context.
         if ( authenticate_ && session.getUserId() == 0 ) {
           msg.replyWithException(new AuthenticationException());
           return;
         }
 
         if ( user != null ) {
-          Group group = (Group) groupDAO.find(user.getGroup());
+          Group group = (Group) x.get("group");
```
kgrgreer, line 85: A side-effect of this is that if we change permissions in our system they will no longer be reflected in logged in users. Users who are already logged in will continue to have those permissions until they logout.

src/foam/box/SessionServerBox.java
```diff
@@ -67,29 +70,27 @@ public void send(Message msg) {
...
             AppConfig appConfig = group.getAppConfig(x);
             x = x.put("appConfig", appConfig);
             session.getContext().put("appConfig", appConfig);
-          } else {
-            logger.error("missing group: ", String.format("User: [%d]", user.getId()));
-            throw new RuntimeException("User without a group.");
+          }
+
+          if ( authenticate_ && ! auth.check(session.getContext(), "service." + spec.getName()) ) {
+            msg.replyWithException(new AuthorizationException("You do not have permission to access to that service."));
```
kgrgreer: Grammar seems a little odd.

src/foam/box/SessionClientBox.js
jlhughes: softSessionLimit present a dialog x time before real session limit prompting the user to click to stay signed in. 
So in Ablii, the idle timeout is set to 10 minutes and the prompt appears 1 minute before.

src/foam/box/SessionServerBox.java
jlhughes: I think this issue is significant.

jlhughes: I suppose if we really needed to revoke privileges we have the ability to manually terminate sessions for users in the affected group.  And thus far not something that's come up.

TW80000: I actually came across this issue while working on the signing officer stuff. @lchanmann pointed it out in a review he did. My solution was to loop through the sessions where the user id (and optionally agent id) matched the user whose permissions we were trying to revoke/change and update the user's group in the context in the session and put it back to the sessionDAO.

See commit https://github.com/nanoPayinc/NANOPAY/pull/6320/commits/373ffea4f8887e65974ecf6187157a154f816060

That only works in a specific situation though, it obviously doesn't cover all potential cases where this could happen.

src/foam/box/SessionServerBox.java
TW80000: Too many to's. Fixed in dffbdd321.

TW80000:
Alright, I have all of foam and nanopay refactored now to support this. I will open the nanopay PR once this PR is merged. Please review @kgrgreer @jlhughes @mcarcaso @adamvy @daudich.

PR #1963, opened 2019-03-09
Ruler enhancement - Part5

lchanmann:


src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -99,14 +102,28 @@ public void execute(X x) {
...
         }
       }
     });
   }
 
+  private void retryAsyncApply(X x, Rule rule, FObject obj, FObject oldObj) {
+    new RetryManager().submit(x, new ContextAgent() {
+      @Override
+      public void execute(X x) {
+        rule.asyncApply(getX(), obj, oldObj, RuleEngine.this);
```
jlhughes, line 121: @lchanmann can you explain asyncApply again for me.  I understand that this execute block is passed to a threadpool.  Does the asyncApply call itself execute asynchronously or is this just a naming convention to distinguish which apply calls are passed to the threadpool?  Thanks.

lchanmann: The execute block above is called by another execute method https://github.com/lchanmann/foam2/blob/RulerEnhancementPOC_part5/src/foam/nanos/ruler/RuleEngine.java#L97 that was passed to a threadpool.

>  Does the asyncApply call itself execute asynchronously or is this just a naming convention to distinguish which apply calls are passed to the threadpool?

asyncApply invokes `asyncAction.applyAction()` on Rule model but does not execute asynchronously by itself. It's just a naming convention so that RuleEngine can pass it to the thread pool.

Rule has both action and asyncAction. asyncActions of all rules in the same group will be grouped together and sent as a single task to the thread pool.

IMG_0843

jlhughes (approved):

PR #1966, opened 2019-03-12
Filter out permissioned property as arguments

KateMengjunXia:
https://nanopay.atlassian.net/browse/CPF-12
When permissioned properties as passed in as predicate arguments, and the user doesn't have the permission, then replace the predicate with TRUE.

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -329,6 +329,32 @@ foam.INTERFACE({
       flags: ['js', 'java'],
       javaSupport: false,
       type: 'foam.mlang.predicate.Predicate',
+    },
+    {
+      name:'authorize',
+      flags: [ 'java' ],
+      type: 'Boolean',
+      args:[
+        {
+          name: 'x',
+          type: 'Context'
+        }
+      ]
+    },
+    {
+      name:'authorizeArg',
+      flags: [ 'java'],
+      type: 'Boolean',
+      args:[
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'arg',
+          type: 'foam.mlang.Expr'
+        }
+      ]
     }
   ]
 });
@@ -390,6 +416,9 @@ foam.CLASS({
   name: 'AbstractPredicate',
   abstract: true,
   implements: [ 'foam.mlang.predicate.Predicate' ],
+  javaImports: [
+    'foam.nanos.auth.AuthService',
+  ],
 
   documentation: 'Abstract Predicate base-class.',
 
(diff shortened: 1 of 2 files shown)


src/foam/mlang/mlang.js
```diff
@@ -457,6 +486,27 @@ foam.CLASS({
...
       swiftCode: '//noop'
+    },
+    {
+      name: 'authorize',
+      javaCode:`
+  return true;
+  `
+    },
+    {
+      name:'authorizeArg',
```
kgrgreer, line 497: The PropertyInfo's implement the Predicate interface, which means that they also have the authorize() method. Move this logic to the PropertyInfo itself and the design will be simpler.

src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -48,6 +48,9 @@ foam.CLASS({
       name: 'select_',
       javaCode: `
       if ( x.get("auth") != null ) {
+        if ( predicate != null && ! predicate.authorize(x) ) {
+          return super.select_(x, sink, skip, limit, order,  new foam.mlang.predicate.True());
```
jlhughes: Two options:
1. Replace just the offending predicate with TRUE. so AND(EQ, EQ....) -> AND(TRUE, EQ, ... ) 
2. Throw an authentication/authorization exception. 
I don’t think collapsing the entire argument predicate to TRUE is appropriate.

src/foam/mlang/mlang.js
KateMengjunXia: It seems like PropertyInfo extends SQLStatement interface, Predicate also extends SQLStatement interface. I don't see any inheritance between PropertyInfo and Predicate. Correct me if I am misunderstanding something.

src/foam/nanos/auth/PermissionedPropertyDAO.js
KateMengjunXia: I think it makes sense to go for your option 2, because the current idea would have the following issue: OR(offending predicate, something evaluates false) => OR(true,  something evaluates false) => true

src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -48,6 +48,9 @@ foam.CLASS({
       name: 'select_',
       javaCode: `
       if ( x.get("auth") != null ) {
+        if ( predicate != null && ! predicate.authorize(x) ) {
+          throw new AuthenticationException("You do not have access to the querying property / properties");
```
jlhughes: need to adjust the wording. 
"Property access restricted."  or "Property access denied", or something along those lines. 
@nanoscott thoughts on this? This is when a query is made against properties that the user does not have 'read' access to.

nanoscott: "Access Denied: Protected property" should work. You want to be clear on the cause.

nanoscott: "Access Denied: Property protected" Or "Access Denied: Query on property not allowed"

KateMengjunXia: Updated exception message.

jlhughes (approved):

jlhughes:
@KateMengjunXia when Kevin returns we can address his ideas.

PR #1967, opened 2019-03-13
Ruler: add canExecute method

smirnovaka:


src/foam/nanos/ruler/Rule.js
```diff
@@ -230,6 +231,11 @@
           ruler.getDelegate().cmd_(x.put("OBJ", obj), getCmd());
         }
       `
+    },
+    {
+      name: 'canExecute',
+      type: 'Boolean',
+      javaCode: `return true;`
```
lchanmann, line 238: value: true

src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -78,17 +75,25 @@ public Object getResult(long ruleId) {
...
       currentRule_ = rule;
       if ( rule.f(getX(), obj, oldObj)
         && rule.getAction() != null
       ) {
-        rule.apply(getX(), obj, oldObj, this);
-        saveHistory(rule, obj);
+        if ( rule.canExecute() ) {
+          updateList.add(rule);
+        } else {
+          return;
```
lchanmann, line 88: break?

src/foam/nanos/ruler/RulerDAO.js
```diff
@@ -66,42 +74,30 @@
...
-      applyRules(x, obj, oldObj,
-        getBeforeGroup( oldObj == null ? Operations.CREATE : Operations.UPDATE
-          , Operations.CREATE_OR_UPDATE
-      ));
-
+      if ( oldObj == null ) {
+        applyRules(x, obj, oldObj, getCreateBefore());
+      } else {
+        applyRules(x, obj, oldObj, getUpdateBefore());
+      }
```
lchanmann, line 81: you can also use conditional operator.

src/foam/nanos/ruler/RulerDAO.js
```diff
@@ -66,42 +74,30 @@
...
-
-      applyRules(x, obj, oldObj,
-        getAfterGroup( oldObj == null ? Operations.CREATE : Operations.UPDATE
-          , Operations.CREATE_OR_UPDATE
-      ));
+      if ( oldObj == null ) {
+        applyRules(x, obj, oldObj, getCreateAfter());
+      } else {
+        applyRules(x, obj, oldObj, getUpdateAfter());
+      }
```
lchanmann, line 87: same as above.

src/foam/nanos/ruler/RulerDAO.js
```diff
@@ -123,14 +119,77 @@
         }
       ],
       javaCode: `
-
       for ( Object key : sink.getGroupKeys() ) {
         List<Rule> group = ((ArraySink) sink.getGroups().get(key)).getArray();
         if ( ! group.isEmpty() ) {
-          new RuleEngine(x, this).execute(group, obj, oldObj);
+          new RuleEngine(x, getDelegate()).execute(group, obj, oldObj);
```
lchanmann, line 125: Use RulerDAO as the delegate of rule engine so that RulerDAO.cmd_ would be called https://github.com/foam-framework/foam2/pull/1947/commits/7edd451b5f8c20e4ad485d5648023aad9f4098a2.

src/foam/nanos/ruler/RulerDAO.js
```diff
@@ -123,14 +119,77 @@
...
+        EQ(Rule.AFTER, false)
+      )).orderBy(new Desc(Rule.PRIORITY)).select(GROUP_BY(Rule.RULE_GROUP, new ArraySink()));
+      setRemoveBefore(removedBefore);
+      GroupBy removedAfter = (GroupBy) ruleDAO.where(AND(
+        EQ(Rule.OPERATION, Operations.REMOVE),
+        EQ(Rule.AFTER, true)
+      )).orderBy(new Desc(Rule.PRIORITY)).select(GROUP_BY(Rule.RULE_GROUP, new ArraySink()));
+      setRemoveAfter(removedAfter);
+        `
+    },
```
lchanmann, line 183: It makes the rule groups static as updateRules will only be called once in the constructor.

src/foam/nanos/ruler/RulerDAOTest.java
```diff
@@ -24,9 +24,9 @@ public void runTest(X x) {
     x = TestUtils.mockDAO(x, "ruleHistoryDAO");
 
     ruleDAO = (DAO) x.get("ruleDAO");
+    createRule(x);
     userDAO = new RulerDAO(x, (DAO) x.get("localUserDAO"), "localUserDAO");
     ruleHistoryDAO = (DAO) x.get("ruleHistoryDAO");
-    createRule(x);
```
lchanmann, line 29: There will also be rules created after RulerDAO is initialized. Shouldn't we handle that?

lchanmann (changes requested):
According to our F2F conversation, canExecute will not address the root problem you are trying to solve i.e., transactional capability support.

src/foam/nanos/ruler/RuleEngine.java
smirnovaka: no, return

src/foam/nanos/ruler/RulerDAO.js
smirnovaka: yes, that is intentional

src/foam/nanos/ruler/RulerDAOTest.java
smirnovaka: yes

src/foam/nanos/ruler/Rule.js
penzital: > According to our F2F conversation, canExecute will not address the root problem you are trying to solve i.e., transactional capability support.

yes, it does solve it. The problem was that the design didn't look so good but there wasn't a better one

smirnovaka:
> According to our F2F conversation, canExecute will not address the root problem you are trying to solve i.e., transactional capability support.

yes, it does solve it. The problem was that the design didn't look so good but there wasn't a better one

PR #1971, opened 2019-03-16
Create RichTextArea for editing html text

molurin1:
**Notes**
RichTextArea currently allows for basic formattings such as Bold, Italic, Underline, Font Size, Font Face, Justification, Indenting etc. It does not currently support Links and Images but these can be added.


mcarcaso (commented):
Looking good but I think we can model this a little more and cut down on a bunch of code while providing more functionality. In rich text editors, the buttons have some state to them and we'll want that behaviour too. For example, if I'm typing in bold, the bold button should be highlighted in some way to tell me that the next character I type will be bolded. Here's a fiddle I came across that shows what I'm talking about: http://jsfiddle.net/vvCzD/
We'll chat about this more offline.

src/foam/u2/view/FontFace.js
```diff
@@ -0,0 +1,147 @@
...
+    },
+    {
+      name: 'window',
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.SUPER();
+      this
```
mcarcaso: You can shorten this if you want by doing:
```
      this
        .startContext({ data: this })
          .start('div', null, this.menu$).addClass('dropdown-content').hide()
            .forEach(this.cls_.getOwnAxiomsByClass(foam.core.Action), function(a) {
              .start('div').start(a).addClass('richtext-button').end().end()
            })
          .end()
        .end();
```

src/foam/u2/view/RichTextArea.js
```diff
@@ -0,0 +1,329 @@
...
+      }
+    },
+  ],
+
+  methods: [
+    function initE() {
+      this.SUPER();
+      this
+        .start('div').addClass('foam-flow-Document')
+          .start('iframe', null, this.frame$).addClass('iframeContainer')
```
mcarcaso: I'm not sure we need to do this in an iframe anymore. Foam1 did it this way because content editable support in a div was scarce but now it's pretty widespread so I think we can simplify our lives a little by making this just a content editable div.

src/foam/u2/view/RichTextArea.js
```diff
@@ -0,0 +1,329 @@
...
+          } else {
+            actions[i].children[1].menu.hide();
+          }
+        }
+      }
+    }
+  ],
+
+  actions: [
+    {
```
mcarcaso: Do you need these actions still? I think you pulled all of these out into their own components.

mcarcaso (commented):
I really like how the actions are grouped into their own little components!

src/foam/u2/view/RichTextArea.js
molurin1: No, would remove it.

src/foam/u2/view/RichTextArea.js
molurin1: Would remove this class, it is not used again. Moved it to RichTextEditor

molurin1: Using div in RichTextEditor. Just ignore RichTextArea.js. Would remove it.

logs/logs
```diff
@@ -2002,3 +2002,122 @@ p({"created":"2019-02-09T16:59:12.976Z","severity":1,"createdBy":1,"message":"[J
 p({"created":"2019-02-09T16:59:13.014Z","severity":1,"createdBy":1,"message":"[JDAO],rules,Successfully read 1 entries from file: rules"})
 p({"created":"2019-02-09T16:59:13.015Z","severity":2,"createdBy":1,"message":"Direct constructor use is deprecated. Use Builder instead."})
 p({"created":"2019-02-09T16:59:13.016Z","severity":1,"createdBy":1,"message":"1 - admin admin,[Service],testDAO,Created Service,ruleDAO"})
+
```
mcarcaso, line 2005: Undo these changes.

src/cronjobs
```diff
@@ -1,3 +1,5 @@
...
 p({"class":"foam.nanos.cron.Cron","minute":0,"hour":1,"dayOfMonth":-1,"month":-1,"dayOfWeek":-1,"scheduledTime":"2018-06-01T05:00:00.0Z","enabled":false,"id":"expired notification","description":"remove expired notifications","lastRun":null,"lastDuration":0,"server":true,"status":1,"code":"import foam.nanos.cron.RemoveExpiredNotificationCronjob;\ncron = new RemoveExpiredNotificationCronjob();\ncron.execute(x);\nprint(\"s\");","output":"","notes":""})
 p({"class":"foam.nanos.cron.Cron","minute":0,"enabled":true,"id":"Renew rule history cron","description":"Cron for renewing rule history - run every hour.","scheduled":false,"code":"import foam.nanos.ruler.cron.RenewRuleHistoryCron; \ncron = new RenewRuleHistoryCron();\ncron.execute(x);\nprint(\"done\");"})
+
```
mcarcaso, line 4: Undo these too

src/foam/u2/texteditor/ToolBar.js
```diff
@@ -0,0 +1,169 @@
...
+      this.refresh();
+
+      this.SUPER();
+      this.addClass('richtext-actions');
+      this
+        .start(this.DetailView, { data: this.DefaultTextFormats.create(), showActions: true, title: ''})
+          .addClass('toolbar-item')
+        .end()
+        .start(this.Popup, {
+          button: this.Image.create({ data: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABUAAAAVCAYAAACpF6WWAAAABGdBTUEAALGPC/xhBQAAAAFzUkdCAK7OHOkAAAAgY0hSTQAAeiYAAICEAAD6AAAAgOgAAHUwAADqYAAAOpgAABdwnLpRPAAAAAZiS0dEAAAAAAAA+UO7fwAAAAlwSFlzAAAASAAAAEgARslrPgAAANpJREFUOMvtkyFuw0AQRd+sCiwrICC8zilKAnqIHKAjXyA3aHKDEtN1YY5Q6CP4CAtSKSDAYQXRTohBW7WODYycz3b/7pv/V1qYtKTLLMtyHmPc9IU5595VNTzcODcXkdcBISsguDHqdyYNIRyyLFt+3zOzPfAEfIrI6teV4w9oURSzNE0XHTMaVW2891/t+qKqoTNpkiRrMyv/I5rZDtj2qT/umzrn6hjjDkBEXoBH4Gxmb61fDYaqag3UAN775xba5Hneq/Lo9ScO/fNHmdmHiATgNMS7azxdAX9WQ9Rz12fXAAAAJXRFWHRkYXRlOmNyZWF0ZQAyMDE5LTAzLTEzVDAzOjE0OjAzKzAwOjAwnW3t/gAAACV0RVh0ZGF0ZTptb2RpZnkAMjAxOS0wMy0xM1QwMzoxNDowMyswMDowMOwwVUIAAAAodEVYdHN2ZzpiYXNlLXVyaQBmaWxlOi8vL3RtcC9tYWdpY2stN2xlZW1xSTPhtKVNAAAAAElFTkSuQmCC'}),
```
mcarcaso, line 141: Can we move these long data urls out of here into constants so the u2 is a little easier to read?

src/foam/u2/texteditor/Popup.js
```diff
@@ -0,0 +1,46 @@
...
+  package: 'foam.u2.texteditor',
+  name: 'Popup',
+  extends: 'foam.u2.View',
+
+  properties: [
+    {
+      name: 'button'
+    },
+    {
+      name: 'view'
```
mcarcaso, line 11: Let's make this a ViewSpec

src/foam/u2/texteditor/Popup.js
```diff
@@ -0,0 +1,46 @@
...
+                self.document.addEventListener('click', function(e2) {
+                  if ( e1.target === e2.target ) return;
+                  if ( ! self.hidden ) self.hidden = true;
+                  self.document.removeEventListener('click', arguments.callee);
+                });
+              }
+            })
+          .end()
+          .start('div').addClass('toolbar-menu')
+            .style({
```
mcarcaso, line 39: I recently learned of the "hide" method in u2 that you can use here instead.
You should be able to replace this call to `style` with a call to `hide(self.hidden$)`

src/foam/u2/texteditor/ToolBar.js
```diff
@@ -0,0 +1,169 @@
...
+          view: this.DetailView.create({ data: this.FontSize.create(), showActions: true, title: ''})
```
mcarcaso, line 142: Since we'll be using ViewSpecs for the `view` property, you'll need to change this to something like:
```
view: { class: 'foam.u2.DetailView', data: this.FontSize.create() .....
```

src/foam/u2/view/RichTextEditor.js
```diff
@@ -0,0 +1,59 @@
...
+    package: 'foam.u2.view',
+    name: 'RichTextEditor',
+    extends: 'foam.u2.View',
+  
+    requires: [
+      'foam.u2.texteditor.ToolBar',
+    ],
+  
+    exports: [
+     'document'
```
mcarcaso, line 17: Probably don't need these exports.

src/foam/u2/view/RichTextEditor.js
```diff
@@ -0,0 +1,59 @@
...
+    `,
+  
+    properties: [
+      {
+        class: 'Boolean',
+        name: 'onKey',
+        attribute: true
+      },
+      {
+        name: 'textarea',
```
mcarcaso, line 38: Why do you need to store this?

src/notifications
```diff
@@ -0,0 +1,3 @@
+
+p({"class":"foam.nanos.notification.Notification","id":1,"body":"Cron STARTED - sessionExpiry Remove expired sessions.","emailArgs":{}})
```
mcarcaso, line 2: Revert this file.

src/foam/u2/texteditor/ToolBar.js
```diff
@@ -0,0 +1,169 @@
...
+    'foam.u2.texteditor.ToolBarItemView',
+    'foam.u2.tag.Image'
+  ],
+
+  exports: [
+    'doc_ as document'
+  ],
+
+  css: `
+    .richtext-actions {
```
mcarcaso, line 30: When outputting css it's typically best to use `^` to ensure we don't conflict with css classes potentially used elsewhere. `^` gets replaced with the class name to ensure the css is unique to this class. So this would get changed to something like
`^actions {`
and then when you want to use it down below, you'd replace 
`this.addClass('richtext-actions');`
with
`this.addClass(this.myClass('actions'));`

mcarcaso (changes requested):

src/foam/u2/view/RichTextEditor.js
molurin1: You are right, not using it again

molurin1:
moved here https://github.com/foam-framework/foam2/pull/2078

PR #1972, opened 2019-03-18
Use native scrollbar for ScrollTableView

TW80000:
Uses the native scrollbar in `ScrollTableView` instead of the canvas-based one it's currently using.

### DIFF
--- a/src/foam/comics/DAOControllerView.js
+++ b/src/foam/comics/DAOControllerView.js
@@ -62,7 +62,8 @@ foam.CLASS({
     }
 
     ^container {
-      display: flex;
+      display: grid;
+      grid-template-columns: fit-content(100%) auto;
     }
 
     ^container > * + * {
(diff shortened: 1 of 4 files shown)


src/foam/comics/DAOControllerView.js
```diff
@@ -62,7 +62,8 @@ foam.CLASS({
     }
 
     ^container {
-      display: flex;
+      display: grid;
```
TW80000, line 65: The changes in this file are required for this to work in Safari. There's a bug in Safari where `position: sticky` elements won't display correctly if they have a `display: flex` ancestor.

src/foam/u2/Element.js
```diff
@@ -523,7 +523,7 @@ foam.CLASS({
   listeners: [
     {
       name: 'paint',
-      isMerged: 100,
+      isFramed: true,
```
TW80000, line 526: This isn't strictly necessary, but it fixes an issue where the table wasn't re-rendering until the end of the scroll.

PR #1976, opened 2019-03-21
CSV Outputter

Nauna:
Main addition was a fix to the CSV Outputter.

But CSVSink.js (CSV Outputter) was redesigned, to extend AbstractSink and to define new property axioms toCSV() and toCSVLabel()

CSVSinkDemo.js is a demo for testing CSVSink with various property types. Will be needing this in the future for rebuilding CSV Parser.

Terminal cmd to run demo: ```python -m SimpleHTTPServer```
URL for testing: ```http://localhost:8000/foam2/?model=foam.test.CSVSinkDemo```

closing: 
https://github.com/nanoPayinc/NANOPAY/issues/5850
and 
CPF-196 #time 3d
CPF-196 #In Review done with @mcarcaso

### DIFF
--- a/src/foam/comics/DAOController.js
+++ b/src/foam/comics/DAOController.js
@@ -92,7 +92,7 @@ foam.CLASS({
       class: 'Boolean',
       name: 'exportCSVEnabled',
       documentation: 'True to enable export csv button',
-      value: false
+      value: true
     },
     {
       class: 'Boolean',
--- a/src/foam/comics/DAOController.js
+++ b/src/foam/comics/DAOController.js
@@ -91,8 +91,7 @@ foam.CLASS({
     {
       class: 'Boolean',
       name: 'exportCSVEnabled',
-      documentation: 'True to enable export csv button',
-      value: true
+      documentation: 'True to enable export csv button'
     },
     {
       class: 'Boolean',
--- a/src/foam/dao/CSVSink.js
+++ b/src/foam/dao/CSVSink.js
@@ -1,0 +1,13 @@
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'CSVSink',
+  extends: 'foam.dao.AbstractSink',
+  properties: [
+    {
+      class: 'String',
+      name: 'csv',
+      view: 'foam.u2.tag.TextArea',
+      value: 'TODO'
+    }
+  ]
+});
--- a/src/foam/dao/CSVSink.js
+++ b/src/foam/dao/CSVSink.js
@@ -2,12 +2,171 @@ foam.CLASS({
   package: 'foam.dao',
   name: 'CSVSink',
   extends: 'foam.dao.AbstractSink',
+
+  documentation: ``,
+
+  requires: [
+    'foam.core.FObject',
+    'foam.dao.AbstractSink',
+    'foam.lib.json.OutputterMode',
+    'foam.util.SafetyUtil'
+  ],
+
   properties: [
     {
       class: 'String',
       name: 'csv',
-      view: 'foam.u2.tag.TextArea',
-      value: 'TODO'
+      view: 'foam.u2.tag.TextArea'
+    },
+    {
+      class: 'Class',
+      name: 'of',
+      visibility: 'HIDDEN'
+    },
+    {
+      name: 'props',
+      expression: function(of) {
+        return of.getAxiomsByClass(foam.core.Property)
+          .filter( (p) => ! p.transient || ! p.storageTransient );
+      },
+      visibility: 'HIDDEN'
+    },
+    {
+      class: 'Boolean',
+      name: 'isHeadersOutput',
+      visibility: 'HIDDEN'
+    },
+    {
+      class: 'Boolean',
+      name: 'isNewLine',
+      value: true,
+      visibility: 'HIDDEN'
+    }
+  ],
+
+  methods: [
+    function output(value) {
+      if ( ! this.isNewLine ) this.csv += ',';
+      this.isNewLine = false;
+      this.output_(value);
+    },
+    {
+      name: 'output_',
+      code:
+        foam.mmethod(
+          {
+            String: function(value) {
+              this.csv += `"${value.replace(/\"/g, '""')}"`;
+            },
+            Number: function(value) {
+              this.csv += value.toString();
+            },
+            Boolean: function(value) {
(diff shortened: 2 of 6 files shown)


src/foam/comics/DAOController.js
```diff
@@ -92,7 +92,7 @@ foam.CLASS({
       class: 'Boolean',
       name: 'exportCSVEnabled',
       documentation: 'True to enable export csv button',
-      value: false
+      value: true
```
mcarcaso: Undo this change or remove the `exportCSVEnabled` property completely.

src/foam/dao/CSVSink.js
```diff
@@ -0,0 +1,172 @@
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'CSVSink',
+  extends: 'foam.dao.AbstractSink',
+
+  documentation: ``,
+
+  requires: [
+    'foam.core.FObject',
```
mcarcaso: I don't think you need any of these requires.

src/foam/dao/CSVSink.js
```diff
@@ -0,0 +1,172 @@
...
+      this.props.forEach((element) => {
+        element.toCSV(obj, this, element);
+      });
+      this.newLine_();
+    },
+
+    function reset() {
+      ['csv', 'isNewLine', 'isHeadersOutput']
+        .forEach( (s) => this.clearProperty(s) );
+      return this;
```
mcarcaso: I don't think you need the return statement here.

src/foam/dao/CSVSink.js
```diff
@@ -0,0 +1,172 @@
...
+  name: 'FObjectPropertyCSVRefinement',
+
+  refines: 'foam.core.FObjectProperty',
+
+  properties: [
+    {
+      name: 'toCSV',
+      class: 'Function',
+      value: function(obj, outputter, prop) {
+        if ( ! prop.of ) return; // TODO JSON
```
mcarcaso: Filling this in should be easy enough so let's do it:
```
if ( ! prop.of ) {
  outputter.output(obj ? obj[prop.name] : null);
  return;
}
```

src/foam/dao/CSVSink.js
```diff
@@ -0,0 +1,172 @@
...
+          .forEach((axiom) => {
+            axiom.toCSV(obj ? obj[prop.name] : null, outputter, axiom);
+          });
+      }
+    },
+    {
+      name: 'toCSVLabel',
+      class: 'Function',
+      value: function(outputter, prop) {
+        if ( ! prop.of ) return; // TODO JSON
```
mcarcaso: Same here:
```
if ( ! prop.of ) {
  outputter.output(prop.name);
  return;
}
```

src/foam/u2/ExportModal.js
```diff
@@ -103,12 +103,12 @@ foam.CLASS({
...
       exportDriver = foam.lookup(exportDriver.driverName).create();
 
-      this.downloadCSV(this.exportData, exportDriver);
+      this.downloadCSVv(this.exportData, exportDriver);
     }
   ],
 
   listeners: [
-    function downloadCSV(data, exportDriver) {
+    function downloadCSVv(data, exportDriver) {
```
mcarcaso: Give this a different name. downloadCSV_ perhaps. Or move the body of this method into the downloadCSV function since it's the only caller of it.

mcarcaso (changes requested):

src/foam/comics/DAOController.js
Nauna: I'm just setting this property back to being false. 

But ... Did we want to get rid of it? Don't we want default true??

I'm going to make a ticket for the parser portion of this, and put this in as a note on TODO

mcarcaso: Yeah we want to get rid of it but doing so might break some things right now so it's best to leave this untoutched for now and not change anybody's behaviour just yet.

Having this `exportCSVEnabled` is more code for less functionality than simply having the `exportEnabled` flag. We should only need the `exportEnabled` and then let the user decide how they'd like to export. Also, having the `exportCSVEnabled` creates a bad precedent for having a separate flag for all of the output types. Before you know it, we'd have an `exportJSONEnabled`, `exportXMLEnabled`, etc. and there's not much value in that when we could have a single `exportEnabled` to rule them all.

Nauna: LOL OK got it. Makes sense. Thanks as always for explaining 👍 🥇

Nauna: https://nanopay.atlassian.net/browse/CPF-695

src/foam/dao/CSVSink.js
```diff
@@ -0,0 +1,176 @@
...
+    {
+      class: 'Class',
+      name: 'of',
+      visibility: 'HIDDEN'
+    },
+    {
+      name: 'props',
+      expression: function(of) {
+        return of.getAxiomsByClass(foam.core.Property)
+          .filter( (p) => ! p.transient || ! p.storageTransient );
```
kgrgreer: I think this should be ! p.networkTransient.

src/foam/dao/CSVSink.js
```diff
@@ -0,0 +1,176 @@
...
+      class: 'Function',
+      value: function(outputter, prop) {
+        if ( ! prop.of ) {
+          outputter.output(prop.name);
+          return;
+        }
+        // mini decorator
+        var prefixedOutputter = {
+          output: function(value) {
+            outputter.output(`${prop.name}.${value}`);
```
kgrgreer: Faster to do just: prop.name + '.' + value

mcarcaso (approved):

src/foam/dao/CSVSink.js
```diff
@@ -0,0 +1,176 @@
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'CSVSink',
+  extends: 'foam.dao.AbstractSink',
+
+  documentation: `Sink runs the csv outputter, and contains the resulting string in this.csv`,
```
kgrgreer: Only one line, just just use regular quotes, which are faster: '

Nauna: I'll be more careful with this in the future. Sorry about that. Good catch.

PR #1977, opened 2019-03-22
Added hello service tests

KateMengjunXia:


src/foam/nanos/hello/ClientHelloService.js
```diff
@@ -0,0 +1,29 @@
...
+  properties: [
+    {
+      class: 'String',
+      name: 'serviceName'
+    },
+    {
+      class: 'Stub',
+      name: 'delegate',
+      of: 'foam.nanos.hello.HelloService'
+    //   factory: function() {
```
adamvy: Just delete this code rather than commenting it out.

src/foam/nanos/hello/HelloTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+import foam.box.HTTPBox;
+import foam.core.X;
+
+public class HelloTest extends foam.nanos.test.Test {
+
+  @Override
+  public void runTest(X x) {
+    ClientHelloService helloService = new ClientHelloService.Builder(x).setDelegate(new HTTPBox.Builder(x).setUrl("http://localhost:8080/service/hello").build()).build();
+    String result = helloService.hello("");
```
adamvy: Give something more than the empty string as a parameter.

src/foam/nanos/hello/tests.jrl
```diff
@@ -0,0 +1,2 @@
+p({"class":"foam.nanos.hello.HelloTest","id":"HelloTest"})
+p({"class":"foam.nanos.test.Test","id":"HelloTestJS","server":false,"status":2,"code":"var hello = foam.nanos.hello.ClientHelloService.create({\n  delegate: foam.box.HTTPBox.create({\n    url: 'service/hello'\n  }, x)\n}, x);\nhello.hello().then(function(msg) { log(msg); });"})
```
adamvy: Instead of log(msg) the test should probably assert that msg equals the expected value.

adamvy:
FYI @jlhughes 

Kate has a minimal test case here that successfully uses HTTPBox to communicate Java Server to Java Server.  Technically this is from a single Java Server to itself, but its still happening via an HTTP Request.

src/foam/nanos/hello/tests.jrl
```diff
@@ -0,0 +1,2 @@
+p({"class":"foam.nanos.hello.HelloTest","id":"HelloTest"})
+p({"class":"foam.nanos.test.Test","id":"HelloTestJS","server":false,"status":2,"code":"var hello = foam.nanos.hello.ClientHelloService.create({\ndelegate: foam.box.HTTPBox.create({\nurl: 'service/hello'\n}, x)\n}, x);\nhello.hello(\"JS\").then(function(msg) { foam.assert(msg === (\"helloJS\"), \"ClientHelloService succeed\") });"})
```
adamvy: Sorry I didn't provide enough detail.  In nanos tests we want to use test(condition, message) rather than foam.assert.

So test(msg == 'whatever', "Didn't received expected greeting.") for example.

src/foam/nanos/hello/ClientHelloService.js
```diff
@@ -0,0 +1,23 @@
...
+      name: 'serviceName'
+    },
+    {
+      class: 'Stub',
+      name: 'delegate',
+      of: 'foam.nanos.hello.HelloService'
+    }
+  ]
+});
+
```
adamvy: Stray whitespace here.

src/foam/nanos/hello/tests.jrl
adamvy: Minor detail but it's also good to end files with a newline, especially line delimited files like journals.

 You can see that github is warning about that with the little symbols at the end of the diff here and in HelloService.js

Ideally you should configure your editor to do this automatically for you.

PR #1978, opened 2019-03-22
DO NOT MERGE

adamvy:
@KateMengjunXia  FYI


src/foam/nanos/mrac/ClusterDAO.js
```diff
@@ -0,0 +1,182 @@
...
+       )
+       .select(new ArraySink())).getArray();
+      DAO[] newClients = new DAO[arr.size()];
+      for ( int i = 0; i < arr.size(); i++ ) {
+        ClusterConfig config = (ClusterConfig) arr.get(i);
+        if ( "localhost".equals(config.getId()) ) {
+          setConfig(config);
+          continue;
+        }
+        DAO client = new ClusterClientDAO.Builder(x).setServiceName(getNSpec().getName()).setConfig(config).build();
```
adamvy: Instead of creating a ClusterClientDAO here we can create a ClientDAO and call setDelegate(new HTTPBox()) like we did in our HelloService test case.

src/foam/nanos/mrac/ClusterDAO.js
```diff
@@ -0,0 +1,182 @@
...
+        },
+        {
+          name: 'obj',
+          type: 'foam.core.FObject'
+        }
+      ],
+      javaCode: `
+      foam.core.FObject o = getDelegate().put_(x, obj);
+      for ( DAO client : getClients() ) {
+        client.put(o);
```
adamvy: And here, instead of calling client.put(), we can call client.cmd_(x, new RPCMessage(...)) to get the cmd_ to our client.

src/foam/nanos/mrac/ClusterDAO.js
```diff
@@ -0,0 +1,182 @@
...
+        },
+        {
+          name: 'obj',
+          type: 'foam.core.FObject'
+        }
+      ],
+      javaCode: `
+      foam.core.FObject o = getDelegate().remove_(x, obj);
+      for ( DAO client : getClients() ) {
+        client.put(o);
```
adamvy: Same here, we can call client.cmd_(new RPCMessage(...)), but make sure we make a 'remove' RPC message.

src/foam/nanos/mrac/ClusterDAO.js
```diff
@@ -0,0 +1,182 @@
...
+      }
+      return o;
+     `
+    },
+    {
+      name: 'cmd_',
+      javaCode: `
+      if ( obj instanceof ClusterRequest ) {
```
adamvy: Then finally we can re-write this method to unpack RPCMessages and apply them to the delegate.  You can look at foam/java/Skeleton.js for how we do this normally.

adamvy: A future improvement would be to add a method to RPCMessage called ".apply" which takes an object and does the method dispatch for us.

src/foam/nanos/mrac/ClusterResponse.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
```
adamvy: No need for this object.

adamvy:
Summary of steps to complete client/server functionality.

1. Remove ClusterClientDAO
2. In ClusterDAO create a series of ClientDAO's instead with the appropriate HTTPBox as a delegate.
3. In ClusterDAO conver all foreach(DAO dao : clients) {}, to call cmd_() on the client, pass an RPCMessage
4. Remove ClusterRequest/Response and just use RPCMessage
5. Add a method to RPCMessage named ".apply()" which takes an object and does the method dispatch work that is currently done in Skeleton.js

adamvy:
On second though, don't use RPCMessage.  Rename ClusterRequest to ClusterCommand and use that instead.  That's the proper way to follow the "Chain of Command" design pattern.

jlhughes:
PR existed only for review/comment purposes.

PR #1979, opened 2019-03-22
ScrollTableView native scrolling support

TW80000:
Refactors `ScrollTableView` to support real native vertical scrolling.

## Changes

* Made the ScrollTableView use real scrolling instead of just replacing the data in the table.
  * This was the main point of the PR and all of the other changes are caused by this.
  * See the image below for the structure of the DOM elements to accomplish this.
  * There is always one `<table>` with three `<tbody>`'s in it. As the user scrolls we add and remove `<tbody>` elements. Each of them has their own DAO with a different skip value. The bulk of the code in this PR is just tracking these `<tbody>`'s and what data to feel them.
* Added a background colour CSS macro so the sticky table header could set it's background colour.
  * It used to be transparent which was fine when the rows didn't move, but now that the rows move, they would be visible behind the table header.
* Removed `fitInScreen :: Boolean` as an optional flag on `ScrollTableView`
  * This flag used to make sure the table height didn't extend off of the bottom of the screen. Now it's always on. Not sure if this is quite what we want. We probably want it as an option still that can be overwritten.
* Changed the way `OverlayDropdown` is positioned. Now it's position is based on the coordinates of the mouse click event instead of the coordinates of it's DOM element.
  * The issue was that since the DOM element was inside the `<td>` element in the table and one of the containers containing the table needs to have `overflow: hidden`, the overlay would be cut off if it were near the bottom of the table. So since we can't put it directly in the table anymore, I chose to add it to the controller and then use the mouse coordinates to position it since we can't use the DOM anymore.

IMG_20190322_154914600

### DIFF
--- a/src/foam/comics/DAOControllerView.js
+++ b/src/foam/comics/DAOControllerView.js
@@ -102,8 +102,7 @@ foam.CLASS({
       name: 'summaryView',
       factory: function() {
         return this.data.summaryView || this.importedSummaryView || {
-          class: 'foam.u2.view.ScrollTableView',
-          fitInScreen: true
+          class: 'foam.u2.view.ScrollTableView'
         };
       }
     },
--- a/src/foam/nanos/controller/AppStyles.js
+++ b/src/foam/nanos/controller/AppStyles.js
@@ -37,7 +37,6 @@ foam.CLASS({
       font-weight: 900;
       line-height: 1.5;
       letter-spacing: 0.4px;
-      background: transparent;
     }
     .foam-u2-view-TableView-row > th > td {
       font-size: 12px;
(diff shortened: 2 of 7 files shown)


TW80000:
There's one optimization I know of that we could still apply to this but I haven't made yet. When a user scrolls more than three pages in a single event (most often this happens when quickly dragging the scrollbar with the mouse on a large table) it currently loops over the `scroll{Up,Down}OnePage` method, which removes one tbody and adds another. Therefore it would still be drawing and removing every tbody between the starting and ending positions of the scroll rather than just removing all three tbodies and redrawing three new ones at the ending position.

kgrgreer:
Has conflicts.

TW80000:
> Has conflicts.

Resolved.

PR #1980, opened 2019-03-22
LRUCachingDAO 

minsunny:
on foam
https://nanopay.atlassian.net/browse/CPF-311

https://github.com/nanoPayinc/NANOPAY/pull/6451

- initially for javascript client.
- only handle find, put, remove
- maintain an inner MapDAO as Cache
- on cache miss, call delegate (ClientDAO)
- on put, remove cache entry
- on delegate find
- found: add to cache
- not found: add key:NULL to cache to indicate that value does not exist in delegate, return null.

LRUCahingDAO applied on balanceDAO, userDAO,  invoiceDAO first,
Changed on Services file.


src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+    function remove(x, o) {
+      var self = this;
+      return self.trackingDAO.remove(o);
+    }
+  ],
+
+  listeners: [
+    /** Adds the put() item to the tracking dao, runs cleanup() to check the dao size. */
+    function onPut(obj) {
+      var self = this;
```
jlhughes, line 153: Is it possible for onPut and onRemove to call the class methods put and remove above?

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+      return this.trackingDAO.find_(x, id).then(function(o) {
+        if ( o ) {
+          return o;
+        } else {
+          return self.delegate.find_(x, id).then(function(o) {
+            if ( o ) {
+              self.put(x, o)
+              return o;
+            } else {
+              return null;
```
jlhughes, line 123: It may be future work, but are we planning on storing a NULL flag for misses?

src/foam/dao/CacheType.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2019

src/foam/dao/CacheType.js
```diff
@@ -0,0 +1,31 @@
...
+
+  values: [
+    {
+      name: 'NONE',
+      documentation: 'NO CacheDAO used',
+      label: 'None'
+    },
+    {
+      name: 'FULL',
+      documentation: 'Using Regular CacheDAO',
```
kgrgreer: 'Use full caching.'

src/foam/dao/CacheType.js
```diff
@@ -0,0 +1,31 @@
...
+foam.ENUM({
+  package: 'foam.dao',
+  name: 'CacheType',
+
+  documentation: 'Set CacheType Value on EasyDao',
+
+  values: [
+    {
+      name: 'NONE',
+      documentation: 'NO CacheDAO used',
```
kgrgreer: 'No caching'

src/foam/dao/CacheType.js
```diff
@@ -0,0 +1,31 @@
...
+      label: 'None'
+    },
+    {
+      name: 'FULL',
+      documentation: 'Using Regular CacheDAO',
+      label: 'Full'
+    },
+    {
+      name: 'LRU',
+      documentation: 'Using LRU CacheDAO',
```
kgrgreer: Use a LRU (Least Recently Used) partial cache

src/foam/dao/EasyDAO.js
```diff
@@ -406,7 +414,7 @@ return delegate;
 //             name: this.model.id + "_" + daoModel.id + "_" + this.name
 //           });
 //         }
-        if ( this.cache ) {
+        if ( this.cacheType == 'FULL') {
```
kgrgreer: Insert space before closing ).

kgrgreer: Should be == foam.dao.CacheType.FULL

src/foam/dao/EasyDAO.js
```diff
@@ -415,6 +423,14 @@ return delegate;
             src: dao,
             of: this.model});
         }
+
+        if ( this.cacheType == 'LRU' ) {
```
kgrgreer: Should be == foam.dao.CacheType.LRU

src/foam/dao/EasyDAO.js
```diff
@@ -465,7 +481,7 @@ return delegate;
         });
         dao.syncRecordDAO = foam.dao.EasyDAO.create({
           of: dao.SyncRecord,
-          cache: true,
+          cacheType: 'FULL',
```
kgrgreer: Should be foam.dao.CacheType.FULL

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2019

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+      ]
+    }
+  ],
+
+  properties: [
+    {
+      /** The maximum size to allow the target dao to be. */
+      class: 'Int',
+      name: 'maxSize',
+      value: 100
```
kgrgreer: 10000 might be a better default

src/foam/dao/EasyDAO.js
```diff
@@ -415,6 +423,14 @@ return delegate;
             src: dao,
             of: this.model});
         }
+
+        if ( this.cacheType == 'LRU' ) {
+          this.mdao = this.MDAO.create({of: params.of});
+          dao = this.LRUCachingDAO.create({
+            dao : this.mdao,
```
kgrgreer: I think we could specify a factory on the 'dao' property of LRUCachingDAO so it would be easier to use. Setting it to an MDAO with 'of' being the same as the 'of' of the delegate is what users will do 99% of the time, so we should make this default.

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+    },
+    {
+      /** By starting at the current time */
+      class: 'Int',
+      name: 'lastTimeUsed_',
+      factory: function() { return Date.now(); }
+    },
+    {
+      class: 'foam.dao.DAOProperty',
+      name: 'dao'
```
kgrgreer: A more informative name like 'cacheDAO' would make the code easier to understand.

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+  ],
+
+  methods: [
+    function init() {
+      this.SUPER();
+      var self = this;
+
+      self.delegate.listen(self.QuickSink.create({
```
kgrgreer: This works, but is an unusual way to implement a decorator. It would be more efficient to just decorate the put(), remove() and removeAll() methods.

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+        putFn: this.onPut,
+        removeFn: this.onRemove
+      }));
+    },
+
+    /** Calculates a timestamp to use in the tracking dao. */
+    function getTimestamp() {
+      // Just increment on each request.
+      return this.lastTimeUsed_++;
```
kgrgreer: If this is just a seqNo (which is good) then why make it be a timestamp at all, rather than just a seqNo starting at 0?

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+
+    /** to keep the dao size. */
+    function cleanup() {
+      var self = this;
+      self.trackingDAO
+        .orderBy(self.DESC(self.LRUCacheItem.TIMESTAMP))
+        .skip(self.maxSize)
+        .select({
+          put: function(obj) {
+            self.dao.remove(obj);
```
kgrgreer: You also need to remove LRUCacheItem from the trackingDAO.

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+      class: 'Int',
+      name: 'maxSize',
+      value: 100
+    },
+    {
+      /** Tracks the age of items in the target dao. */
+      name: 'trackingDAO',
+      of: 'foam.dao.DAO',
+      factory: function() {
+        return this.MDAO.create({ of: this.LRUCacheItem });
```
kgrgreer: Add an index on the timestamp/seqNo field so that its efficient when you do the ORDER_BY, otherwise this will be very slow.

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+        });
+    },
+
+    function find_(x, id) {
+      var self = this;
+
+      return this.trackingDAO.find_(x, id).then(function(o) {
+        if ( o ) {
+          return o;
+        } else {
```
kgrgreer, line 116: No need for 'else' if you've already returned on the line above.

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,154 @@
...
+          }
+        });
+    },
+
+    function find_(x, id) {
+      var self = this;
+
+      return this.trackingDAO.find_(x, id).then(function(o) {
+        if ( o ) {
+          return o;
```
kgrgreer, line 115: At this point, you should also update the timestamp of the object in the itemCache.

src/foam/dao/LRUCachingDAO.js
kgrgreer: The ItemCache should contain a boolean to say if the object exists or not. It will happen a lot that the DAO will be used to try to find a value that doesn't exist, and we need to cache that it doesn't exist otherwise we will still be going to the network every time.

src/foam/dao/EasyDAO.js
```diff
@@ -170,12 +172,18 @@ return delegate;
...
     {
-      /** Enable local in-memory caching of the DAO. */
-      class: 'Boolean',
-      name: 'cache',
-      generateJava: false,
-      value: false
+      class: 'foam.core.Enum',
+      of: 'foam.dao.CacheType',
+      name: 'cacheType',
+      value: 'NONE' /* 'None'*/
```
lchanmann, line 186: Should it be
```suggestion
      value: foam.dao.CacheType.NONE /* 'None'*/
```

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,174 @@
...
+      of: 'foam.dao.DAO',
+      factory: function() {
+        return this.MDAO.create({ of: this.LRUCacheItem });
+      }
+    },
+    {
+      /** By starting at 0 */
+      class: 'Int',
+      name: 'lastTimeUsed_',
+      factory: function() { return 0; }
```
lchanmann: can we use value instead of factory?

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,174 @@
...
+  ],
+
+  documentation: `
+     Least Recently Used CachingDAO :
+     Manages a DAO\'s size by removing old items. Commonly applied inside a cache to limit the cache\'s size.
+  `,
+
+  classes: [
+    {
+      /** Links an object id to a last-accessed seqNo */
```
lchanmann: can we use documentation rather than code comment? There are also a few similar places below.

lchanmann (changes requested):

src/foam/dao/EasyDAO.js
kgrgreer: Both work because the adapt method for enums will make the conversion.
@mcarcaso Is it better to just do 'NONE' for class loading purposes?

minsunny: Then will leave it as "NONE"

src/foam/dao/LRUCachingDAO.js
kgrgreer: Yes, documentation: is better.

src/foam/dao/EasyDAO.js
```diff
@@ -415,6 +424,13 @@ return delegate;
             src: dao,
             of: this.model});
         }
+
+        if ( this.cacheType == foam.dao.CacheType.LRU ) {
+          this.mdao = this.MDAO.create({of: params.of});
```
kgrgreer: This line isn't needed, is it? I don't think this.mdao is used by this cache.

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,174 @@
...
+      name: 'LRUCacheItem',
+      properties: [
+        {
+          name: 'id',
```
kgrgreer: No trailing comma.

src/foam/dao/LRUCachingDAO.js
kgrgreer: ping

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,174 @@
...
+      name: 'trackingDAO',
+      of: 'foam.dao.DAO',
+      factory: function() {
+        return this.MDAO.create({ of: this.LRUCacheItem });
+      },
+      documentation: 'Tracks the age of items in the target dao'
+    },
+    {
+      class: 'Int',
+      name: 'lastTimeUsed_',
```
kgrgreer: The name 'lastTimeUsed_' is misleading since it isn't a time. Maybe something like 'nextSeqNo' would be better.

src/foam/dao/CacheType.js
```diff
@@ -0,0 +1,31 @@
...
+    },
+    {
+      name: 'FULL',
+      documentation: 'Using full caching',
+      label: 'Full'
+    },
+    {
+      name: 'LRU',
+      documentation: 'Using a LRU (Least Recently Used) partial cache',
+      label: 'LRU'
```
kgrgreer: Label defaults to the same value as 'name' if not specified, so this line isn't needed.

src/foam/dao/EasyDAO.js
kgrgreer: ping

src/foam/dao/LRUCachingDAO.js
```diff
@@ -0,0 +1,174 @@
...
+  implements: [ 'foam.mlang.Expressions' ],
+
+  requires: [
+    'foam.dao.MDAO',
+    'foam.dao.QuickSink'
+  ],
+
+  documentation: `
+     Least Recently Used CachingDAO :
+     Manages a DAO\'s size by removing old items. Commonly applied inside a cache to limit the cache\'s size.
```
kgrgreer: I don't think you need \' since this string isn't quoted with 's but is quoted with `.

jlhughes:
Closed - replaced by javascript side caching.

PR #1981, opened 2019-03-22
Fix predicate authorize permission names

lchanmann:
- permission names are expected to be lowercase

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -453,7 +453,13 @@ foam.CLASS({
 
     if ( prop.getPermissionRequired() ) {
       AuthService auth = (AuthService) x.get("auth");
-      return auth.check(x, prop.getClassInfo().getObjClass().getSimpleName() + ".rw." + prop.getName()) || auth.check(x, prop.getClassInfo().getObjClass().getSimpleName() + ".ro." + prop.getName());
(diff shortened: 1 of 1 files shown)


lchanmann:
PermissionedPropertyDAO.select_ calls Predicate.authorize and breaks a bunch of places on NANOPAY where there is a predicate on a permissioned field. Eg.

https://github.com/nanoPayinc/NANOPAY/blob/development/nanopay/src/net/nanopay/onboarding/email/NewUserOnboardedEmailDAO.js#L55 and https://github.com/nanoPayinc/NANOPAY/blob/development/nanopay/src/net/nanopay/bank/CanReceiveCurrencyDAO.java#L61, etc.,

> #1966 got merged since March 13, why hasn't it broken NANOPAY then but now?

We haven't pegged FOAM for a while until this morning. The previous one was https://github.com/nanoPayinc/NANOPAY/pull/6380 which doesn't include #1966.

@KateMengjunXia @jlhughes @kgrgreer

KateMengjunXia (commented):
Why %s instead of rw and ro?

lchanmann:
> Why %s instead of rw and ro?

So that we can reuse permission string and provide the variation (e.g., rw and ro) when needed.

jlhughes (approved):

PR #1984, opened 2019-03-25
Improve error message for unauthorized property access via mlangs

TW80000:
## Changes

* Changed the interface from returning a boolean to return type void and throwing an exception if need be. This way, deeply nested mlangs can throw specific and useful errors and the call site doesn't need to try and figure out what caused the call to `authorize` to return false to create an error message.
* Changed the exception thrown from an `AuthenticationException` to an `AuthorizationException`. I did this because `AuthenticationException`s will cause the web client to show the sign in screen.
* Changed the error message to include the specific model and property that caused the exception to be thrown, making it easier to debug.
* Remove `authorizeArg` method from interface
* Add an authorize method to Expr interface, which PropertyInfo implements
* Move logic previously in `authorizeArg` to `authorize` on PropertyInfo

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -149,6 +149,17 @@ foam.INTERFACE({
     {
       name: 'partialEval',
       type: 'foam.mlang.Expr'
+    },
+    {
+      name: 'authorize',
+      flags: [ 'java' ],
+      type: 'Void',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        }
+      ]
     }
   ]
 });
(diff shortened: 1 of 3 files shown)


TW80000:
@KateMengjunXia Please review as well.

KateMengjunXia (approved):

lchanmann:
Nice job. Thanks @TW80000

src/foam/mlang/mlang.js
```diff
@@ -441,29 +441,30 @@ foam.CLASS({
...
-  return true;
-  `
+      javaCode: `//noop`
     },
     {
-      name:'authorizeArg',
-      javaCode:`
+      name: 'authorizeArg',
+      javaCode: `
   if ( arg instanceof foam.core.PropertyInfo ) {
```
kgrgreer: I don't think this check is needed. Or this method. Just move this code to AbstractPropertyInfo and then rely on polymorphism to handle this for you. Then instead of calling authorizeArg just call arg.authorize().

TW80000: Resolved by 0ec5a4592.

jlhughes (approved):
Looks good.

kgrgreer:
Much nicer.

PR #1988, opened 2019-03-26
Add explicit service permissions check to SessionServerBox 

tharmaman:
As outlined in ticket [CPF-694](https://nanopay.atlassian.net/browse/CPF-694), after having removed the service.* permissions wildcard from both basicUser and SME, we need to enforce strict service permissions for security purposes.

Previously in SessionServerBox.java, on lines 89 and 90, we were ignoring the specific service checks.
sessionServerBoxCommentedOut

I have uncommented the above two lines and left a comment block describing why we needed to do so for any future reader.

### DIFF
--- a/src/foam/box/SessionServerBox.java
+++ b/src/foam/box/SessionServerBox.java
@@ -86,8 +86,13 @@ public class SessionServerBox
 
           if ( authenticate_ && ! auth.check(session.getContext(), "service." + spec.getName()) ) {
             logger.debug("missing permission", group != null ? group.getId() : "NO GROUP" , "service." + spec.getName());
-            // msg.replyWithException(new NoPermissionException("No permission"));
-            // return;
+            /**
+             * for security purposes we uncommented the following two lines
+             * we needed to remove basicUser and SME wildcard permissions to service.*
+             * and enforce explicit service permissions
+             */
+            msg.replyWithException(new NoPermissionException("No permission"));
+            return;
           }
 
           // padding this cause if group is null this can cause an NPE


src/foam/box/SessionServerBox.java
```diff
@@ -86,8 +86,13 @@ public void send(Message msg) {
 
           if ( authenticate_ && ! auth.check(session.getContext(), "service." + spec.getName()) ) {
             logger.debug("missing permission", group != null ? group.getId() : "NO GROUP" , "service." + spec.getName());
-            // msg.replyWithException(new NoPermissionException("No permission"));
-            // return;
+            /**
+             * for security purposes we uncommented the following two lines
```
jlhughes: Remove the comment - as it's nanopay specific.  Thanks.

jlhughes (approved):

src/foam/box/SessionServerBox.java
```diff
@@ -88,6 +88,14 @@ public void send(Message msg) {
...
           Group group = (Group) x.get("group");
 
+          if ( authenticate_ && ! auth.check(session.getContext(), "service." + spec.getName()) ) {
+            logger.debug("missing permission", group != null ? group.getId() : "NO GROUP" , "service." + spec.getName());
+            msg.replyWithException(new NoPermissionException("No permission"));
+            return;
+          }
+
+          // padding this cause if group is null this can cause an NPE
+          // technically the user shouldn't be created without a group
```
kgrgreer, line 98: What does "padding this cause" mean?

I think that the group may not be set if the user is acting as an agent.

@TW80000, why is that?

TW80000: @shreymahendru wrote that comment originally in https://github.com/foam-framework/foam2/commit/62795f6ae94f8ad6e4b94d61229089a87a7e8e53. I think "padding this cause" means "wrapping this in a null check because".

> I think that the group may not be set if the user is acting as an agent.

That shouldn't be true. When you act as a user we set the group in the context to the junction group. See https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/auth/AgentUserAuthService.java#L89

PR #1992, opened 2019-03-27
Fix issues with sign in and sign up

TW80000:
Part of an incoming nanopay PR to fix an issue regarding signing up for Ablii. The issue is that when you sign up for Ablii, it should show you the "validate email" screen, but it was showing it briefly and then immediately switching to the dashboard (which is the default view for that group). This was happening because `onGroupUpdate` was getting called after `onUserLoad`. `onUserLoad` would load the validate email screen, but then shortly after when the group was fetched, the `onGroupUpdate` method would cause the group's default menu to load.

To fix this I combined the two listeners into one that gets called after the user, agent, and group are all loaded. This way we can return early from the function if needed, like we want to when we show the verify email screen.

### DIFF
--- a/src/foam/nanos/controller/ApplicationController.js
+++ b/src/foam/nanos/controller/ApplicationController.js
@@ -141,10 +141,7 @@ foam.CLASS({
     {
       class: 'foam.core.FObjectProperty',
       of: 'foam.nanos.auth.Group',
-      name: 'group',
-      postSet: function(oldValue, newValue) {
-        if ( newValue ) this.onGroupUpdate(newValue);
-      }
+      name: 'group'
     },
     {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -363,32 +360,27 @@ foam.CLASS({
      *   - Update the macros list based on the group
      *   - Go to a menu based on either the hash or the group
      */
-    async function onGroupUpdate(group) {
-      this.setPortalView(group);
+    function onUserAgentAndGroupLoaded() {
```
TW80000, line 363: Open to suggestions for a better name than this.

PR #1994, opened 2019-03-28
Fix issues with email templates

TW80000:
In commit 8868279cc998c73e27b745263783c02c57c26191 I made changes to
access the group in the context rather than the user's group. I didn't
realize that these places in particular still should reference the group
on the user being passed in rather than the user in the context. The
change I made in that commit caused the wrong email template to be used
because it was now based on the group of the user doing the put, not the
user who's getting an email generated for them.

509e2e1 reverts those changes to exactly how they were before.

The changes to `SessionServerBox` make sure that the URL is set in the appConfig even when a user is signing up.

### DIFF
--- a/src/foam/box/SessionServerBox.java
+++ b/src/foam/box/SessionServerBox.java
@@ -18,9 +18,9 @@ import foam.nanos.logger.*;
 import foam.nanos.logger.PrefixLogger;
 import foam.nanos.session.Session;
 import foam.util.SafetyUtil;
-import java.util.Date;
 import javax.naming.NoPermissionException;
 import javax.servlet.http.HttpServletRequest;
+import org.eclipse.jetty.server.Request;
 
 public class SessionServerBox
   extends ProxyBox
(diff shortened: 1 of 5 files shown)


src/foam/box/SessionServerBox.java
```diff
@@ -85,12 +85,23 @@ public void send(Message msg) {
           return;
         }
 
+        // If there is no user in the session (which happens when a user is
+        // signing up, for example) then set the URL in the app configuration to
+        // the URL that the request is coming from.
+        if ( req != null && ! SafetyUtil.isEmpty(req.getRequestURI()) ) {
+          AppConfig appConfig = (AppConfig) x.get("appConfig");
+          appConfig = (AppConfig) appConfig.fclone();
+          String configUrl = ((Request) req).getRootURL().toString();
```
jlhughes, line 94: So the url at this point is 'http', while in front of the load balancers is 'https', for example.  There is a flag in appConfig forceHttps which you need to use to craft the url correctly.

jlhughes (changes requested):

jlhughes (approved):

PR #1995, opened 2019-03-28
Add a calendar date picker to the u2 view library.

mcarcaso:
It's not super pretty but I prefer this over most date pickers I've used before:

image

Demo link if you run `python -m SimpleHTTPServer` from cmd line:
`http://localhost:8000/index.html?model=foam.u2.view.date.DateTimePicker`

### DIFF
--- a/src/foam/u2/view/date/AbstractDateView.js
+++ b/src/foam/u2/view/date/AbstractDateView.js
@@ -1,0 +1,139 @@
+foam.CLASS({
+  package: 'foam.u2.view.date',
+  name: 'AbstractDateView',
+  extends: 'foam.u2.Element',
+  properties: [
+    {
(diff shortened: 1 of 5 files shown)


jlhughes:
@mcarcaso with @kgrgreer port of the foam1 calendar to foam2 can we close this PR?

kgrgreer:
I would prefer Mike's if we can update it to Logan's UX design. The foam1
one brings in a lot of dependencies.

jlhughes:
@mcarcaso UX styling is here: https://zpl.io/2p4BK8E
Suspect you'll need to ask for access.

mcarcaso:
We can certainly update the UX to match Logan's designs later. In the meantime at least we have a functional but ugly date picker.

PR #1996, opened 2019-03-29
Fix password reset not having user in context

TW80000:
Fixes https://nanopay.atlassian.net/browse/CPF-710 and https://nanopay.atlassian.net/browse/CPF-772.


kgrgreer (changes requested):
Let's meet to discuss when you're available.

jlhughes (approved):

jlhughes (approved):

PR #1997, opened 2019-03-29
V2.1 email refactor

Nauna:
Related to : https://github.com/foam-framework/foam2/pull/1945

Description:
EmailsUtility.java is the default class that is the developers class to call. Examples of this usage are in the files listed under ```Code Change```.
The EmailsUtility then after processing all the properties - passes the emailMessage through to the email pipeline. 
Email Pipeline begins with the emailMessageDAO(service) which has two responsibilities(EmailServiceDAO.js). One it stores the emailMessage in a journal and two it passes the emailMessage to the next service in the pipeline.
The second service after emailMessageDAO is the email(service). This default service in FOAM is the NullEmailService, and in nanopay is the SMTPEmailService.

#### Refactored Architect changes:
-```src/foam/nanos/notification/email/EmailServiceDAO.js : ``` Decorator on emailMessage(service) that passes the emailMessage(eM) to the emailService and stores eM into journal.
-```src/foam/nanos/notification/email/DAOEmailService.js :``` removed
-```src/foam/nanos/notification/email/EmailService.js :``` simplified to only have one method
-```src/foam/nanos/notification/email/EmailTemplate.js :``` added functions to handle setting default properties to an emailMessage
-```src/foam/util/email/EmailsUtility.java :``` the static utility functions that are needed for the population of properties and error checking prior a put into the email service.
-```src/foam/nanos/app/EmailConfig :``` adding default properties for emails.
-```src/foam/nanos/notification/email/SMTPEmailService.js :``` clean up to handle only the mimeMessage and transport of emailMessage
-```src/foam/nanos/notification/email/NullEmailService.js :``` since extends EmailService, needed to be simplified to one method override.
-```src/foam/nanos/notification/email/services.jrl```
-```tools/classes.js```

#### Code Change: Changes to email service calls based on above file refactor
-```src/foam/nanos/auth/email/EmailDocService.js```
-```src/foam/nanos/auth/email/EmailTokenService.js```
-```src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js```
-```src/foam/nanos/notification/SendEmailNotificationDAO.java```

#### Clean UP: Changes to email service calls based on above file refactor
-```src/foam/nanos/auth/email/EmailVerificationWebAgent.java```
-```src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js :``` changes to account for the SMTPEmailService not existing on default.
-```src/foam/nanos/notification/email/scripts.jrl```


src/foam/nanos/app/AppConfig.js
```diff
@@ -72,6 +72,21 @@ foam.CLASS({
     {
       class: 'String',
       name: 'supportEmail'
+    },
+    {
+      class: 'String',
+      name: 'emailsFrom',
+      value: 'info@nanopay.net'
```
mcarcaso: Don't include anything nanopay specific in the open source repo.

src/foam/nanos/app/AppConfig.js
```diff
@@ -72,6 +72,21 @@ foam.CLASS({
     {
       class: 'String',
       name: 'supportEmail'
+    },
+    {
```
mcarcaso: I'm not sure about including these in the AppConfig since this also gets sent to the client and we might not want to ship this to the client. I wonder if we can create either an EmailService or DAO decorator to fill in this stuff on an email.

src/foam/nanos/notification/email/DAOEmailService.js
```diff
@@ -57,90 +52,18 @@ foam.CLASS({
...
-      .and()
-    .and()
-  .build();
-}
-return config_;`
-    },
     {
       name: 'sendEmail',
       javaCode: `
         getDao().inX(x).put(emailMessage);
```
mcarcaso: Do we need this implementation at all anymore or can we delete this entire file?

src/foam/nanos/notification/email/EmailServiceDAO.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
mcarcaso: 2019 :p , and there's a smaller header you can use
```
/**
 * @license
 */
```

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -54,16 +82,160 @@ bodyAsByteArrayIsSet_ = false;`
...
+          }
+        }
+
+        // build model, and confirm it was built
+        JtwigModel model = JtwigModel.newModel(templateArgs);
+        if ( model == null ) throw new NoSuchFieldException("JtwigModel is Null");
+
+        // Try to locate user group
+        DAO groupDAO = ((DAO) x.get("groupDAO")).inX(x);
+        Group group = null;
```
mcarcaso: You can do:
```
Group group = user != null ? (Group) groupDAO.find(user.getGroup()) : null
```
And then remove the next 3 lines.

Though there's quite a bit of logic going on here for simply finding the group and I wonder if it's all necessary. Don't we usually put a group in the context and could we just be using that? If it's not in the context, is that a sign that something's wrong or is falling back to the user's group reasonable? Also, when would `groupDAO.find(user.getGroup())` ever return null?

src/foam/nanos/notification/email/EmailsUtility.java
```diff
@@ -0,0 +1,118 @@
+package foam.nanos.notification.email;
```
mcarcaso: Why not use `foam.util.Email`?

src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -243,160 +205,49 @@ return config_;`
...
-emailMessage.setBody(templateBody.render(model));
-
-// If subject has already provided, then we don't want to use template subject.
-if (SafetyUtil.isEmpty(emailMessage.getSubject())) {
-  JtwigTemplate templateSubject = JtwigTemplate.inlineTemplate(emailTemplate.getSubject(), config);
-  emailMessage.setSubject(templateSubject.render(model));
-}
+      javaCode:
+        `
+        if ( ! this.getEnabled() ) return;
```
mcarcaso: Why even have this flag?

src/foam/nanos/notification/email/DAOEmailService.js
Nauna: Deleting the file. Thx for the catch

src/foam/nanos/notification/email/EmailServiceDAO.js
Nauna: :|

src/foam/nanos/notification/email/SMTPEmailService.js
Nauna: On the nanopay side we are overriding the email(service) with the smtp implementation. If this is to remain in the build, we would want this off by default.

Will talk to you in person

src/foam/nanos/app/AppConfig.js
Nauna: New ticket to encompass a future fix: https://nanopay.atlassian.net/browse/CPF-870

src/foam/nanos/app/AppConfig.js
Nauna: New ticket to encompass a future fix: https://nanopay.atlassian.net/browse/CPF-870

src/foam/nanos/notification/email/SMTPEmailService.js
Nauna: New ticket to encompass a future fix: https://nanopay.atlassian.net/browse/CPF-870

src/foam/nanos/app/AppConfig.js
Nauna: So defaults were removed from here and set to be populated in service.jrl. Added a line to nanopay/src/services.jrl to overridden in the nanopay repo.

src/foam/core/Action.js
```diff
@@ -94,14 +94,14 @@ foam.CLASS({
...
     {
       class: 'Function',
       name: 'isEnabled',
       label: 'Enabled',
       help: 'Function to determine if action is enabled.',
-      value: null
+      value: function() { return true; }
     },
     {
       class: 'Function',
```
Nauna: Changes here are not mine. Shows Troy made these changes a day ago. Odd that this is here?

Nauna:
Just finished adding emailPropertyService(service) which is called from intended convenience function in EmailsUtility.java

emailPropertyService has the purpose of abstracting out the setting of email properties based on order of precedence.

This works by storing an EmailPropertyService[] on ChainedPropertyService.

EmailPropertyService[] = 
{
new ApplyEmailTemplateService(), 
new GroupEmailTemplateService(), 
new DefaultPropertyEmailTemplateService()
}

src/foam/nanos/notification/email/EmailPropertyService.js
```diff
@@ -0,0 +1,39 @@
...
+
+foam.INTERFACE({
+  package: 'foam.nanos.notification.email',
+  name: 'EmailPropertyService',
+
+  methods: [
+    {
+      name: 'apply',
+      type: 'foam.nanos.notification.email.EmailMessage',
+      // javaThrows: ['java.lang.NoSuchFieldException'],
```
mcarcaso: Remove?

src/foam/nanos/notification/email/ApplyEmailTemplateService.js
```diff
@@ -0,0 +1,72 @@
...
+
+  javaImports: [
+    'foam.nanos.logger.Logger',
+    'foam.util.SafetyUtil'
+  ],
+
+  methods: [
+    {
+      name: 'apply',
+      type: 'foam.nanos.notification.email.EmailMessage',
```
mcarcaso: When implementing an interface you can omit the args and type of a method as that will be pulled in for you.

src/foam/nanos/notification/email/ApplyEmailTemplateService.js
```diff
@@ -0,0 +1,72 @@
...
+      documentation: 'application of template args to emailTemplate and then onto the emailMessage',
+      javaCode: `
+      Logger logger = (Logger) x.get("logger");
+      String templateName = (String)templateArgs.get("template");
+      if ( SafetyUtil.isEmpty(templateName) ) return emailMessage;
+
+      // STEP 1) Find EmailTemplate
+      EmailTemplate emailTemplateObj = DAOResourceLoader.findTemplate(x, templateName, group);
+      if ( emailTemplateObj == null ) {
+        logger.warning("@ApplyEmailTemplateService: emailTemplate not found.");
```
mcarcaso: Maybe add the template name in the log so we know what to look for if we ever fall down this path.

src/foam/nanos/notification/email/ApplyEmailTemplateService.js
```diff
@@ -0,0 +1,72 @@
...
+        return emailMessage;
+      } 
+
+      // STEP 2) Apply Template to emailMessage
+      try {
+        emailMessage = emailTemplateObj.apply(x, group, emailMessage, templateArgs);
+      } catch (Exception e) {
+        logger.warning("@ApplyEmailTemplateService: emailTemplate.apply has failed, with a thrown exception. ", e);
```
mcarcaso: Same here. Might want to add template name to log.

src/foam/nanos/notification/email/ApplyEmailTemplateService.js
```diff
@@ -0,0 +1,72 @@
...
+        return null;
```
mcarcaso: Why return null rather than throwing an exception?

src/foam/nanos/notification/email/DefaultPropertyEmailTemplateService.js
```diff
@@ -0,0 +1,81 @@
...
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.notification.email',
+  name: 'DefaultPropertyEmailTemplateService',
+
+  documentation: 'Used in conjuction with ChainedPropertyService',
```
mcarcaso: This has no relation to ChainedPropertyService. We just happen to use it in conjunction with it but I wouldn't make mention of it here since they're completely independent.

src/foam/nanos/notification/email/DefaultPropertyEmailTemplateService.js
```diff
@@ -0,0 +1,81 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.notification.email',
+  name: 'DefaultPropertyEmailTemplateService',
```
mcarcaso: I'd maybe call this something like EmailConfigEmailPropertyService since it's an EmailPropertyService that uses the EmailConfig.

src/foam/nanos/notification/email/ApplyEmailTemplateService.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.notification.email',
+  name: 'ApplyEmailTemplateService',
```
mcarcaso: I'd call this EmailTemplateApplyEmailPropertyService since it's an EmailPropertyService that does an EmailTemplate.apply

src/foam/nanos/notification/email/DefaultPropertyEmailTemplateService.js
```diff
@@ -0,0 +1,81 @@
...
+          documentation: 'Template arguments'
+        }
+      ],
+      javaCode: `
+        Logger logger = (Logger) x.get("logger");
+        EmailConfig emailConfig = (EmailConfig) x.get("emailConfig");
+
+        // Service property check
+        if ( emailConfig == null
+          || SafetyUtil.isEmpty(emailConfig.getReplyTo())
```
mcarcaso: I don't think making all these checks is necessary and it also doesn't scale well if/when more properties are added to the email config. Also, an empty string for display name is technically valid. I'm not sure replyTo is even a required field for emails either.

src/foam/nanos/notification/email/DefaultPropertyEmailTemplateService.js
```diff
@@ -0,0 +1,81 @@
...
+          || SafetyUtil.isEmpty(emailConfig.getDisplayName())
+          || SafetyUtil.isEmpty(emailConfig.getFrom()) 
+        ) {
+          logger.error( "EmailConfig service has invalid property settings.");
+          return emailMessage;
+        }
+
+        // REPLY TO:
+        if ( SafetyUtil.isEmpty(emailMessage.getReplyTo()) ) {
```
mcarcaso: I think you want to make `isPropertySet` checks rather than isEmpty because an empty string is valid in some cases.

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -42,10 +52,23 @@ foam.CLASS({
       documentation: 'Template body',
       view: { class: 'foam.u2.tag.TextArea', rows: 40, cols: 150 },
       javaSetter:
-`body_ = val;
-bodyIsSet_ = true;
-bodyAsByteArray_ = null;
-bodyAsByteArrayIsSet_ = false;`
+        `body_ = val;
```
mcarcaso: This javaSetter provides no value from what I can tell. Ditch it if that's correct.

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -54,16 +77,133 @@ bodyAsByteArrayIsSet_ = false;`
...
+
+        return fillInEmailProperties_(x, emailMessage, model, config);
+      `
     },
     {
-      class: 'String',
-      name: 'sendTo',
-      documentation: `This property will set to whomever the email is being sent
-        to.`
+      name: 'fillInEmailProperties_',
```
mcarcaso: Why have this separate method rather than just doing it all inline in the method above? If there's no reason for the separation, I'd just inline it.

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -54,16 +77,133 @@ bodyAsByteArrayIsSet_ = false;`
...
+        },
+        {
+          name: 'config',
+          javaType: 'EnvironmentConfiguration'
+        }
+      ],
+      javaCode: `
+        // BODY:
+        if ( SafetyUtil.isEmpty(emailMessage.getBody()) ) 
+        {
```
mcarcaso: This isn't really our style for java. If you want a formatter to come into play, open the generated java file, run a formatter on it, find the method you implemented, and copy/paste it back here.

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -54,16 +77,133 @@ bodyAsByteArrayIsSet_ = false;`
...
+          javaType: 'org.jtwig.JtwigModel'
+        },
+        {
+          name: 'config',
+          javaType: 'EnvironmentConfiguration'
+        }
+      ],
+      javaCode: `
+        // BODY:
+        if ( SafetyUtil.isEmpty(emailMessage.getBody()) )
```
mcarcaso: All these isEmpty checks should probably be isPropertySet instead. This way, if I want to override the subject with an empty string before sending the email along this pipeline, the value won't get clobbered.

src/foam/nanos/notification/email/GroupEmailTemplateService.js
```diff
@@ -0,0 +1,79 @@
...
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.notification.email',
+  name: 'GroupEmailTemplateService',
+
+  documentation: 'Used in conjuction with ChainedPropertyService',
```
mcarcaso: Remove mention of this and describe what this class does instead:
Fills unset properties on an email with values from the group.

src/foam/nanos/notification/email/GroupEmailTemplateService.js
```diff
@@ -0,0 +1,79 @@
...
+      documentation:
+      `Looks for group email properties on passed in group and then up the chain to root group.
+       Note: It is possible that only one of the properties is set on the group.`,
+      javaCode: `
+        Group grp = null;
+        do {
+          grp = (Group)((DAO) x.get("groupDAO")).find(group);
+          if ( grp == null ) return emailMessage;
+          group = grp.getParent();
+        } while ( SafetyUtil.isEmpty(grp.getReplyTo()) && SafetyUtil.isEmpty(grp.getDisplayName()) );
```
mcarcaso: Do we ONLY want to use the replyTo of a group if the displayName is set?
For example if I have:
Group A - replyTo: "group@mail.com" and displayName: "Group A"
Group B - child of group A, displayName: "Group B"

What should the behaviour be? I think I'd expect it to be an email from "Group B <group@mail.com>".

src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -243,160 +200,49 @@ return config_;`
...
+        if ( emailMessage.getTo().length < 1 || SafetyUtil.isEmpty(emailMessage.getFrom())) {
```
mcarcaso: Technically I can send an email to only people in the CC or BCC field. I'm not sure this check adds much value.

src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -243,160 +200,49 @@ return config_;`
...
+          throw new IllegalStateException("Email that is being transported out does not have basic fields set.");
+        }
 
-sendEmail(x, emailMessage);`
+        ((FixedThreadPool) getThreadPool()).submit(x, new ContextAgent() {
```
mcarcaso: Oh this kicks it out into its own thread huh? I'm not sure I like that happening in here. When I call `sendEmail` on the SMTPEmailService, I'd expect the email to be sent when the method returns. I think you'll want to make a decorator that calls an EmailService in a separate thread so we can move this feature around in the chain of EmailServices we're creating.

src/foam/nanos/notification/email/ChainedPropertyService.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.notification.email',
+  name: 'ChainedPropertyService',
```
mcarcaso, line 9: This model seems incomplete...

mcarcaso (commented):
Sorry for the delay here I completely lost track of this.

src/foam/nanos/notification/email/ApplyEmailTemplateService.js
Nauna: It was just a structure decision. If we throw an exception, then we'd have to define javaThrows in EmailPropertyService, with this we could throw exceptions throughout all implementing classes but then we'd have to add try-catch to the for loop in EmailsUtility. 

I made the decision to minimize the code, by just returning null.

src/foam/nanos/notification/email/GroupEmailTemplateService.js
Nauna: 👍 Thx for the suggestion 😉

src/foam/nanos/notification/email/GroupEmailTemplateService.js
Nauna: Ya to be honest I pondered this too. Glad you caught it. Ok so my thoughts are that if we are setting the values on a group then we are customizing this group and thus would be setting the values we want on the group.

But now that I'm thinking about it again, if I change one property to be customized, I'd expect the other properties to remain the same.

I'll make these changes. With next push

src/foam/nanos/notification/email/SMTPEmailService.js
Nauna: REally. HMMMMMM so should we check if they are all empty?

src/foam/nanos/notification/email/SMTPEmailService.js
Nauna: image

So don't really get the point of a decorator or where this decorator would be set. Why don't we just remove the thread?

src/foam/nanos/notification/email/ChainedPropertyService.js
Nauna: lol ... ok let me think about it.

src/foam/nanos/notification/email/ApplyEmailTemplateService.js
mcarcaso: I think this is worth the extra code. You'll only need to add the javaThrows on the interface method and foam will fill it in on all your implementations so it shouldn't be THAT much extra code and I think it's more "correct" to throw an exception when something that should be exceptional happens.

src/foam/nanos/notification/email/SMTPEmailService.js
mcarcaso: Nah SMTPEmailService shouldn't care about this sort of validation. If we want to enforce that sort of thing we should implement a decorator for validating an email. SMTPEmailService should take an email and try to blindly send it over SMTP.

src/foam/nanos/notification/email/SMTPEmailService.js
mcarcaso: I'm guessing it was put in because SMTP needs to hit the network to actually send the email and we don't want it blocking. SMTPEmailService shouldn't care about this.

src/foam/nanos/notification/email/ChainedPropertyService.js
mcarcaso: Should be pretty easy. The sendEmail method just iterates over the FObjectArray and call sendEmail on each of those.

Nauna:
Things that are remaining:
1) need to adjust the SMTP enabled feature. Current idea is to adjust this in build scripts. Currently this is set to true in the accompanying nanopay PR, that overrides the emailService in (service)email from NullEmailService to SMTPEmailService.

2) the from property that is being passed from EmailMessage to MimeMessage in the SMTPEmailService is not getting set. What is getting displayed is the username that we are using to authenticate our email. Only thing I found, after much testing and reading was this : https://coderanch.com/t/433494/java/Sending-mails-address

which basically says we cannot change it. Although previously we were setting it. No idea when this happened or why.

Nauna:
Things that remain: (cont)
3. Need to store which emails were actually sent and which have failed. Currently we store the emails before the email gets sent. Perhaps we should just reverse the code call on this?

4. Removed the try catch, because a valid error is thrown that stops the email sending, but nothing else. However reviewer please confirm this:

Test:
image


result:
image

src/foam/nanos/notification/email/EmailServiceDAO.js
```diff
@@ -0,0 +1,34 @@
...
+  name: 'EmailServiceDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  properties: [
+    {
+      name: 'emailService',
+      documentation: `This property determines how to process the email.`,
+      of: 'foam.nanos.notification.email.EmailService',
+      class: 'FObjectProperty',
+      javaGetter:
```
mcarcaso: Make a factory instead. That way it's just the default and can be overwritten.

src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -32,45 +30,37 @@ foam.CLASS({
...
+                this.password_ = password;
+              }
+
+              @Override
+              protected PasswordAuthentication getPasswordAuthentication() {
+                return new PasswordAuthentication(this.username_, this.password_);
+              }
+            }
+
+            protected Session session_ = null;
```
mcarcaso: Make this a property and give it a javaFactory that does everything in the `start` method and then remove the `start` method.

src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -111,291 +101,106 @@ protected EnvironmentConfiguration config_ = null;`
...
-        {
-          name: 'x',
-          type: 'Context'
-        },
-        {
-          name: 'emailMessage',
-          type: 'foam.nanos.notification.email.EmailMessage'
+      javaCode:
+      `
+      if ( ! this.getEnabled() ) return;
```
mcarcaso: This boolean can go now.

mcarcaso (changes requested):
Looks good. Just a couple smaller things.

src/foam/nanos/notification/email/SMTPEmailService.js
```diff
@@ -67,10 +65,20 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'Boolean',
-      name: 'enabled',
-      value: true,
-      javaFactory: 'return true;'
+      name: 'session_',
+      class: 'Object',
```
mcarcaso, line 71: if you give this a javaType you won't need to cast it below.

src/foam/nanos/notification/email/EmailServiceDAO.js
```diff
@@ -9,14 +9,27 @@ foam.CLASS({
...
   properties: [
     {
       name: 'emailService',
       documentation: `This property determines how to process the email.`,
       of: 'foam.nanos.notification.email.EmailService',
       class: 'FObjectProperty',
-      javaGetter:
-      `return (foam.nanos.notification.email.EmailService)getX().get("email");`
+      factory: function() {
+        return this.email;
```
mcarcaso: Unless you plan on implementing the JS put_ method, I wouldn't include this factory.

mcarcaso (approved):

src/foam/nanos/auth/CountryService.java
```diff
@@ -72,7 +70,7 @@ protected void notifySupport(String query) {
     emailMessage.setSubject ("Unknown Country");
     emailMessage.setBody    ("User just added an unknown country " + query);
     emailMessage.setTo(new String[]{"ops@nanopay.net"});
-    emailService.sendEmail(getX(), emailMessage);
+    EmailsUtility.sendEmailFromTemplate(getX(), null, emailMessage, null, null);
```
mcarcaso, line 73: Sorry I didn't notice that we were using this to send non-templated emails. I think this design still needs a little work but at least we're in a better state than before so let's roll with this given the current time crunch. Please file a task to revisit this and remove the use of this utility class in favour of a service that knows how to send emails given a template. We'll have to think about that interface a little but we'll tackle that later.

Nauna: https://nanopay.atlassian.net/browse/CPF-1290

Nauna:
Still remaining:

https://nanopay.atlassian.net/browse/CPF-1248  : "Images in Emails not working"
https://nanopay.atlassian.net/browse/CPF-1290 : "Email services adjustment"
https://nanopay.atlassian.net/browse/CPF-1237 : "rework services{emailMessageDAO, email}"
https://nanopay.atlassian.net/browse/CPF-1224 : "Add permission to Operations group to access emails."
https://nanopay.atlassian.net/browse/CPF-1208 : "Add the ability to resend an email"

PR #2004, opened 2019-04-01
Implement comparable property info

smirnovaka:


src/foam/core/AbstractPropertyInfo.java
```diff
@@ -73,6 +73,9 @@ public void diff(FObject o1, FObject o2, Map diff, PropertyInfo prop) {
     }
   }
 
+  public int compareTo(Object obj) {
+    return getName().compareTo(((PropertyInfo)obj).getName());
```
kgrgreer: Should compare class getClassInfo() also, not just name, so that User.ID != Account.ID.

jlhughes (approved):

PR #2007, opened 2019-04-02
 Make documentation changes for User and Address

LillianWong:


src/foam/nanos/auth/Address.js
```diff
@@ -27,31 +27,34 @@ foam.CLASS({
...
       class: 'Boolean',
       name: 'verified',
-      documentation: 'Identifies if address has been verified.'
+      documentation: 'Verifies that the address exists.'
     },
     {
       class: 'Boolean',
       name: 'deleted',
-      documentation: 'Marks address as deleted.'
+      documentation: 'Indicates the address as deleted.'
```
jlhughes: Perhaps: When 'true', Indicates that this address record has been marked as deleted.  To normal privileged users this address record does not exist, but it still exists in the system as historical data may refer to it.  Only an administrator can see deleted records.

src/foam/nanos/auth/Address.js
```diff
@@ -79,7 +82,8 @@ foam.CLASS({
       targetDAOKey: 'countryDAO',
       name: 'countryId',
       of: 'foam.nanos.auth.Country',
-      documentation: 'Country address.',
+      documentation: `A generated name that doesn't contain any special characters which represents
```
jlhughes: countryId is a foreign key into the CountryDAO.  By chance the ID matches the human readable ISO country code.

src/foam/nanos/auth/Address.js
```diff
@@ -97,7 +101,8 @@ foam.CLASS({
       targetDAOKey: 'regionDAO',
       name: 'regionId',
       of: 'foam.nanos.auth.Region',
-      documentation: 'Region address.',
+      documentation: `A generated name that doesn't contain any special characters which represents
```
jlhughes: Similar to countryId.

src/foam/nanos/auth/Address.js
```diff
@@ -221,23 +226,23 @@ foam.CLASS({
...
       name: 'longitude',
-      documentation: 'Longitude of address location.'
+      documentation: 'The longitude of the address location.'
     },
     {
       class: 'FObjectArray',
       of: 'foam.nanos.auth.Hours',
       name: 'hours',
-      documentation: 'Opening and closing hours for this address',
+      documentation: 'The opening and closing hours for this address',
```
jlhughes: ... if the address represents a business.

src/foam/nanos/auth/User.js
```diff
@@ -33,8 +33,9 @@ foam.CLASS({
     'foam.util.SafetyUtil',
     'static foam.mlang.MLang.EQ'
   ],
-
-  documentation: '',
+  
+  documentation: `Authenticates a user in the system.
```
jlhughes: The User model/record represents a user of the system.  Has nothing to do with authentication itself.

PR #2010, opened 2019-04-02
Create a code editor and use it on the nanos Script's code property.

mcarcaso:
The editor was just copied from codemirror.net. Here's what it looks like in nanos:

image

Kinda sucks that it's pulling in so much code so I won't be heart broken if we don't wanna go forward with this but I'm finding it useful for writing scripts.


kgrgreer:
How big is it?

mcarcaso:
11k lines

kgrgreer:
<500k if we only load JS and Java support.

PR #2012, opened 2019-04-03
Code Editor

mcarcaso:
Create a view that wraps a third party editor from https://ace.c9.io


jlhughes (commented):
Always amazed at what we can do in foam with such little code.

src/io/c9/ace/Lib.js
```diff
@@ -0,0 +1,20 @@
+foam.CLASS({
```
kgrgreer: Why not use foam.LIB() instead? Because this is more lazy?

mcarcaso: I did it this way because foam.LIBs don't really store any state and I want to store this in a variable and a constant provides this functionality. But now that I think about it, I guess I could use foam.LIB and pass this function to memoize0? Maybe that's the better way to go since creating an instance of this "Lib" makes no sense.

mcarcaso: Ok now it's using foam.LIB and I put a fallback to a TextArea in case it fails to load. Let me know what you think.

PR #2016, opened 2019-04-03
Give DetailView's data factory result a context

mcarcaso:


jlhughes (commented):
Curious - what does this fix/correct?  Where would I see the difference with respect to testing or directing someone to this change when I see some behaviour.  Thanks.

kgrgreer:
This would affect the use of relationships from the object. They wouldn't be able to find the dao's from the context if they don't have a context.

mcarcaso:
I ran into this when trying to make a ViewMenu in nanos where the json was something like:
`{"class": "foam.u2.DetailView", "of": "foo.Bar"}`
and the instance of `foo.Bar` had the global context which was missing all of the services.

PR #2019, opened 2019-04-04
Add IsClassOf predicate

lchanmann:
To check if the class of object if a specified class.

### DIFF
--- a/src/foam/mlang/MLang.java
+++ b/src/foam/mlang/MLang.java
@@ -167,4 +167,16 @@ public class MLang
       .setArg2(new PredicatedExpr(o2))
       .build();
   }
+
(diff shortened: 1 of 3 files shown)


mcarcaso:
Is the InstanceOf predicate what you're actually looking for?
https://github.com/foam-framework/foam2/blob/32bc8b00a2cfb7eec192eddcdde6dc43c80b47c5/src/foam/mlang/mlang.js#L2156

There is a difference between your predicate and that one but I'm not sure if that's intentional. Yours wouldn't match a subclassed object with a predicate for its parent class. For example:
If I have:
`Vehicle`
`Car extends Vehicle`
`CLASS_OF(Vehicle)` being applied to a `Car` would be `false`. Is that the behaviour you want?

PR #2020, opened 2019-04-04
Configured Jetty Server To Interpret the "X-Forwarded-For" Header

tharmaman:
This PR is in regards to ticket [CPF-364](https://nanopay.atlassian.net/browse/CPF-364).

When using a Load Balancer, the Jetty webservers needed to be configured correctly otherwise users keep getting logged out.

This is because since we are using a **Load Balancer** in production, when a user's request gets to our Jetty server after having passed through the Load Balancer, **our Jetty server sends the response to the IP of the Load Balancer instead of the originating connection**. This is where the **X-Forwarded-For** header comes into play.

As a cause, what I did to allow our Jetty server's to interpret the X-Forwarded-For header, is that I configured the connectors to be able to interpret the "X-Forwarded-For" header by passing in an instance of the **ForwardRequestCustomizer** to the **HTTPConfiguration** of each connector as detailed [here](https://www.eclipse.org/jetty/documentation/9.4.x/configuring-connectors.html#_x_forward_for_configuration).


kgrgreer:
Don't we want to only accept X-Forwarded-For from specific hosts, ie. only from the load-balancers? Do we need to configure this from code, or can we do it from the standard Jetty config files?

jlhughes (dismissed):

tharmaman:
> Don't we want to only accept X-Forwarded-For from specific hosts, ie. only from the load-balancers? Do we need to configure this from code, or can we do it from the standard Jetty config files?

I think this configuring can be done with the code...

Looking at the [ForwardedRequestCustomizer API](https://www.eclipse.org/jetty/javadoc/current/org/eclipse/jetty/server/ForwardedRequestCustomizer.html#getRemoteAddr(java.lang.String)), they have getRemoteAddr method, which according to the normal [javadocs](https://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getRemoteAddr%28%29) (I'm assuming they will follow the same definitions), **returns the Internet Protocol (IP) address of the client or last proxy that sent the request**. 

What we could do is have a whitelist of IPs and use the abovementioned getter to check if the last address is on the whitelist (i.e. the last proxy that sent the request would be the load balancer's IP).

I'm just a little but unsure how to configure this in the connectors as I've done with the code currently to just enable the ability to interpret X-Forwarded-For.

Thoughts @daudich?

PR #2022, opened 2019-04-05
Fix ClassReferenceParser

lchanmann:
Returns ClassInfo of the model when forClass_ is a modelled class, otherwise returns the Class itself.

ClassInfo is used as the underlying java type for the property of type "Class".
Thus when unmarshalling from json string it should be converted back into a ClassInfo.

For example, for the model below
```js
foam.CLASS({
  name: 'Thing',
  properties: [
    { name: 'targetClass', class: 'Class' }
  ]
});
```
When unmarshalling
```json
{
  "class": "Thing",
  "targetClass": {
    "class": "__Class__",
    "forClass_": "OtherThing"
  }
}
```
The targetClass should be the ClassInfo of "OtherThing" and not OtherThing.class.


kgrgreer:
Are you sure this is the correct behaviour? It seems odd that we would return a Class or ClassInfo. If someone needs a ClassInfo instead of a Class then shouldn't they just resolve it themselves?

lchanmann:
There is an issue when unmarshalling the above example because the PropertyParser tries to set the "targetClass" property and it expects a ClassInfo but ClassReferenceParser return a Class so the parser throws ClassCastException. 

Script to re-produce:
```
import foam.lib.json.*;
import foam.nanos.auth.User;

import static foam.mlang.MLang.*;

pred = INSTANCE_OF(User.class);
out = new Outputter();
jsonString = out.stringify(pred);
print(jsonString);  // json string

parser = new JSONParser();
parser.setX(x);
obj = parser.parseString(jsonString, Predicate.class);
print(obj);  // null
```

Since "Class" type on the Java side is simply a ClassInfo, we could probably make it always returns ClassInfo. Also, currently on the outputter side we don't support generating value of pure Java class type. https://github.com/foam-framework/foam2/blob/master/src/foam/lib/json/Outputter.java#L206:L241.

lchanmann:
@kgrgreer do you think we make ClassReferenceParser to only return ClassInfo?

adamvy:
This problem is alluded to my TODO left right by the code you're updating.

```
// TODO(adamvy): Should Class vs ClassInfo be separate types?
// What should we store in Class properties.
```

The question being, should properties of { class: 'Class' }, store the ClassInfo or the Java Class object?  In JS these are the same thing, but in Java and Swift and any language that has its own class system they are separate.

AbstractDAO.OF for example is a "Class" property but stores a ClassInfo in java. 

https://github.com/foam-framework/foam2/blob/master/src/foam/dao/AbstractDAO.js#L63

We should answer this question, remove the TODO and make sure the parsers/outputters on all platforms do the same thing.

lchanmann:
> We should answer this question .. should properties of { class: 'Class' }, store the ClassInfo or the Java Class object?

Yes, we should decide how to handle this properly.

Our new type system for "Class" type also defaults to ClassInfo on the Java side.

https://github.com/foam-framework/foam2/blob/master/src/foam/core/Type.js#L329

It seems like we don't have custom javaType: 'java.lang.Class' on properties of { class: 'Class' } in our codebase. Would it be sensible to only store ClassInfo? 

And if we would like/need to support storing both, it might be also better to separate them as @adamvy has mentioned. @kgrgreer, any thoughts?

kgrgreer:
We really want to store the Model, but decided to call models classes in JS to make it more familiar to people. What we really want is the Model/JS-Class/ClassInfo. We can get the actual Java class from the ClassInfo. I would just keep this named Class and clearly document that it is storing the classInfo/JS-Model. If we need an actual Class property, we probably just want a "class name" like we use in NSpec, in which case we can just use String since classes aren't actually marshallable or editable.

lchanmann:
Updated. @kgrgreer @adamvy, please take a look when you have a chance.

lchanmann:
@kgrgreer, is there something else we need to address here?

adamvy:
This looks reasonable, but I think we can optimize a bit.

First off, in https://github.com/foam-framework/foam2/pull/2051 I added an end to end serialization test that serializes a complex object to the server and back.  I had to disable the part of the test which deals with Class/ClassInfo references.  Can you merge your PR with master so you have those changes, and uncomment the lines in my test to verify that these changes pass the end to end test?


Secondly, I think we can take a shortcut for Class properties like we do for Date properties and just serialized the name of the class as a string, instead of the more complex `{ class: "__Class__" ...` construct.  We should still use `__Class__` by default, but when dealing with a Class property it should have custom serialization/deserialization to just serialize the class name.

Take a look at #2051 for to see the places where we put in special optimizatino for DAte properties.  In particular the Date property in JS the AbstractDatePropertyInfo.java object.

lchanmann:
Updated, please take another look. Thanks @adamvy.

src/tests
```diff
@@ -40,3 +40,6 @@ p({"class":"foam.nanos.test.Test","id":"e2e-json-serialization-test","descriptio
...
 p({"class":"foam.nanos.test.Test","id":"e2e-json-serialization-test","status":1})
+p({"class":"foam.nanos.test.Test","id":"e2e-json-serialization-test","code":"var jsonParser = foam.json.Parser.create({ strict: true });\n\nvar initialObject = foam.test.AllSerializableProperties.createPopulated();\n\nvar jsSerialized = foam.json.Network.stringify(initialObject);\n\nlog(\"JS Serialization: \", jsSerialized);\n\nvar afterJSSerialization = jsonParser.parseString(jsSerialized);\n\ntest(foam.util.equals(initialObject, afterJSSerialization), \"After JS Serialization object is still equivalent.\");\n\nx[\"serializationTestEchoService\"].echo(afterJSSerialization).then(function(afterJavaSerialization) {\n  var diff = initialObject.diff(afterJavaSerialization);\n  for ( var keys = Object.keys(diff), i = 0, key ; key = keys[i++] ; ) {\n    log(\"Differs on property \", key, \": \", diff[key]);\n  }\n  test(foam.util.equals(initialObject, afterJavaSerialization), \"After Java Serialization object is still equivalent.\");\n});\n"})
```
adamvy: I think we can cleanup this journal and just replace the line 39 with your updated line 43.

tools/classes.js
```diff
@@ -346,6 +346,7 @@ var classes = [
 
 
   // Testing support
+  'foam.test.AllProperties',
```
adamvy: I don't think the test requires the AllProperties class, as it serializes AllSerializableProperties.

adamvy:
This is looking good, what I'd love to do at this point is add custom serialization/deserialization for the Class property so that it can just send a the class id as a string.  Similar to how Date property just sends a number as opposed to the whole `{ class: '__Timestamp__' ...}` construct.

lchanmann:
Added passClassesByReference boolean to JSON Outputter and override toJSON in Class property so that it outputs string or `{ class: '__Class__', ... }` with respect to the value of passClassesByReference.

If passClassesByReference is true, Class objects are passed as `{ class: '__Class__', ... }`. Otherwise, passes the class.id as string.

The server-side JSON Outputter is unchanged meaning we still store { "class": "\_\_Class\_\_", ... } in the journal files. @adamvy @kgrgreer, do you think we are okay with this?

src/foam/core/JSON.js
```diff
@@ -186,6 +186,12 @@ foam.CLASS({
       help: 'If true, Property objects are passed as __Property__ references rather than by value.',
       value: true
     },
+    {
+      class: 'Boolean',
+      name: 'passClassesByReference',
+      help: 'If true, Class objects are passed as __Class__ references rather than by value.',
```
adamvy: This isn't really an accurate description and I don't think we ever will need to change this flag.  There's no way to pass a class object by value, as classes are not serializable.  They are always passed by name/id, but its just a question of whether we use the long form `{ "class": "__Class__",  "name": "Foo" }` or the short form `"Foo"`

This isn't a decision that the JSONOutputter needs tomake or even be aware of.  In the Class property toJSON, we should always just output the id, same as we do for Dates, we always just output the raw date value directly rather than the long form `{ "class":"__Timestamp__"...` version.

See here: https://github.com/foam-framework/foam2/blob/master/src/foam/core/types.js#L87 in either case we are returning a simple value.  For dates we have extra config on whether we want to format as a number or a string, but for Class references we don't need to make this distinction.

src/foam/core/types.js
```diff
@@ -342,6 +342,13 @@ foam.CLASS({
   extends: 'Property',
 
   properties: [
+    {
+      name: 'toJSON',
+      value: function toJSON(value, outputter) {
+        return value == null ? null :
```
adamvy: This should just be `return value.id`

src/foam/test/AllProperties.js
```diff
@@ -104,7 +104,8 @@ foam.CLASS({
     },
     {
       class: 'foam.u2.view.TableCellFormatter',
-      name: 'tableCellFormatterProp'
+      name: 'tableCellFormatterProp',
+      generateJava: false
```
adamvy: Should be able to revert this as we're not generating this class at all.

adamvy:
I still think we want to special case Class property serialization so that it only outputs the id.  For dates we do this here https://github.com/foam-framework/foam2/blob/master/src/foam/core/types.js#L87 for JS and here https://github.com/foam-framework/foam2/blob/master/src/foam/core/AbstractDatePropertyInfo.java#L72 for Java.

For Class properties we should add an AbstractClassPropertyInfo with a custom toJSON() that is just outputString(((foam.core.ClassInfo)value).getId()); for java.

src/foam/core/AbstractClassPropertyInfo.java
```diff
@@ -0,0 +1,10 @@
+package foam.core;
+
+import foam.lib.json.Outputter;
+
+public abstract class AbstractClassPropertyInfo extends AbstractObjectPropertyInfo {
+  @Override
+  public void toJSON(Outputter outputter, Object value) {
+    outputter.outputClassInfoValue((ClassInfo) value);
```
adamvy: This should just be outputString(((ClassInfo)value).getId()) to be consistent with the JS should it not?

adamvy: Oh nvm, I see what you did.  I don't think we really need to split outputClassInfo and outputClassInfoValue.  The only reason this was done for dates is that how we actually output a date value differs depending on the outputter outputReadableDates flag, but we have no such flag for ClassInfos, the ClassInfo value is always output the same way.

lchanmann: The outputString and outputClassInfo methods of Outputter class have "protected" access modifiers thus cannot be called from AbstractClassPropertyInfo. 

The Outputter class also has a public output method and AbstractPropertyInfo, the ancestor of this class (AbstractClassPropertyInfo), already have toJSON method that calls outputter.output(value) so having AbstractClassPropertyInfo seems to not have any benefits.

Also on server-side, I wonder whether we should store the long form { "class": "\_\_Class\_\_", ... } or short form (class name as string) of Class objects in the journal files.

adamvy: "The outputString and outputClassInfo methods of Outputter" making those public rather than adding a new outputClassInfoValue method is simpler.

"Also on server-side, I wonder whether we should store the long form { "class": "__Class__", ... } or short form (class name as string) of Class objects in the journal files."

"In the journal files" is a special case we shouldn't consider directly.  When outputting a Class property it should output as a plain string, when outputting  a class in some other context like a generic Object property it will be `{ "class": "__Class__", ...`

So in the journals it will typically just be the Class name, but it could be the long form depending on context.

lchanmann: Removed outputClassInfoValue in https://github.com/foam-framework/foam2/pull/2022/commits/28bc409f12ec9ca3497c2d7dd4ac022be2684d6a.

lchanmann:
Addressed your comments, @adamvy. Thanks.

PR #2030, opened 2019-04-08
Remember table column preferences

TW80000:
https://nanopay.atlassian.net/browse/CPF-855

## Changes

* Tables will now "remember" the columns you choose to show/hide and use those preferences any time a table with that model is being displayed.
* Simplified the code in EditColumnsView.

### DIFF
--- a/src/foam/u2/view/EditColumnsView.js
+++ b/src/foam/u2/view/EditColumnsView.js
@@ -1,4 +1,4 @@
-/** 
+/**
  * @license
  * Copyright 2017 The FOAM Authors. All Rights Reserved.
  * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/EditColumnsView.js
```diff
@@ -43,74 +35,58 @@ foam.CLASS({
...
+      var index = 0;
+      this
+        .addClass(this.myClass())
+        .forEach(this.columns, function(columnName) {
+          var axiom = this.table.getAxiomByName(columnName);
+          var localStorageKey = this.table.id + '.' + columnName;
+          var valueFromLocalStorage = localStorage.getItem(localStorageKey);
+
+          // This will be true only when seeing a new model for the first time.
+          if ( valueFromLocalStorage === null ) {
```
TW80000: Currently I have it set up so that there's always a value in localStorage. Another way we could do it is to only set a value in localStorage if the checkbox has been unchecked, and remove the entry in localStorage when it's checked again. So we'd only be storing the properties that we don't want to show.

I wasn't sure which way was better, but I'm happy to switch it if anyone thinks the other way is better.

jlhughes: So when a new property is added to the model tableColumns[], they will not see it.  And this is regardless if they have explicitly made a selection or not.  
Given this, I would prefer the negative case, where we record the 'not-selected'. Then they are aware of new properties we add and can consciously disable them. 
On the flip side, not-showing what's new might help testing - if they were validating what is shown,  but this would be a quick update - to deselect and run the test again.

TW80000: > So when a new property is added to the model tableColumns[], they will not see it.

That wasn't the intention, and I just tested and that's not happening for me.

jlhughes: Ok, if that's not the case, then I'm fine with the currently coded approach.

TW80000: Updated to the other approach anyway since it requires less code and less space in localStorage.

jlhughes (approved):

PR #2032, opened 2019-04-08
ReferenceView fixes

mcarcaso:
- Adds toSummary to FObjects and uses that to fuel the ReferenceView options.
- Adds placeholder to ReferenceView.

### DIFF
--- a/src/foam/u2/view/ReferenceView.js
+++ b/src/foam/u2/view/ReferenceView.js
@@ -38,7 +38,7 @@ foam.CLASS({
     {
       name: 'objToChoice',
       value: function(obj) {
-        return obj.toSummary();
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/ReferenceView.js
```diff
@@ -31,31 +31,14 @@ foam.CLASS({
...
-          }
-
-          f = function(obj) {
-            return [obj.id, obj.id];
-          };
-
-          return f(obj);
-        };
+      value: function(obj) {
+        return obj.toSummary();
```
kgrgreer: return array including id

mcarcaso: woops. fixed.

PR #2043, opened 2019-04-12
Add option to ignore actions defined on model in context menu

TW80000:
Necessary for https://nanopay.atlassian.net/browse/CPF-886


jlhughes (commented):
I would prefer we do this via permissions.

PR #2045, opened 2019-04-15
Make maybeCall and isAvailableFor check for permission

TW80000:
This makes `isAvailable` return a Promise, so `isEnabled` should probably return a Promise as well. Just wanted to see what you thought of this before doing that.

### DIFF
--- a/src/foam/core/Action.js
+++ b/src/foam/core/Action.js
@@ -153,10 +153,11 @@ foam.CLASS({
         slot ;
     },
 
-    function isAvailableFor(data) {
-      return this.isAvailable ?
+    async function isAvailableFor(ctx, data) {
+      var available = this.isAvailable ?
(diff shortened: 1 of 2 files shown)


kgrgreer (changes requested):
I disapprove this PR. I don't think we should just change the sync/async nature of methods as required. If we do, I would say then it needs to be a new method. With this change, isAvailableFor becomes async but isEnabledFor stays sync. The correct thing to do is to use createIsAvailable$(), which is already async by virtue of using a Slot. Like Promises, Slots are async, but better than Promises they can also change their values in the future. We have the convention of having methods that return slots have '$' as a suffix to their name, but we should also have some convention for methods that return promises. Maybe 'Async', so isAvailableForAsync(data) (although, we don't need that in this case).

PR #2047, opened 2019-04-15
Style data picker

nanoNeel:
refs: https://nanopay.atlassian.net/browse/CPF-939


jlhughes (approved):

jlhughes:
DO_NOT_MERGE waiting on an nanopay PR with required layout updates, as this new default will break a number of nanopay UIs.

TW80000 (changes requested):
Please fix the issues we talked about in person. Namely:

1. Position of the calendar not correct when opening after scrolling down a bit.
2. Clicking on a date should close the calendar (optional, I just expected it to behave that way)
3. Shouldn't default to today if you clear it. This breaks things like validation where we want to know if the user picked a date or not.

Henry0422:
The position of the date pickers is slightly different using Firefox, I am able to scroll the new invoice form of the send request payment, but the date picker is fixed at the bottom.

It needs to go with the date text field and it might be tricky to adjust the CSS position/layout of the date picker.

PR #2048, opened 2019-04-15
Modified Session Model To Account For Proxied Requests

tharmaman:
This PR is in regards to ticket [CPF-867](https://nanopay.atlassian.net/browse/CPF-867).

Rather than making a new whitelistedXForwardedForDAO, Joel and I decided that it would be unnecessary as we could just modify the Session model instead.

Here are the following changes that were made:
1. Rather than the Session model only including one remoteHost property which stores the IP returned from req.getRemoteHost(), I replaced that property with two new properties: **sourceHost** and **proxyHost**.

**sourceHost** and **proxyHost** were put in place to account for requests who have the **X-Forwarded-For** header:
     * sourceHost would be the IP detailed in the X-Forwarded-For header
     * proxyHost would be what ever req.getRemoteHost returns (which is always the last node that sent the request which in this case is the proxy)
     * in cases without proxies, sourceHost would be req.getRemoteHost and proxyHost would be null

2. Modified the logic in **AuthWebAgent.java** and **SessionServerBox.java** which previously contained session.getRemoteHost() to account for sourceHost and proxyHost by factoring in cases for proxies and non-proxies

**BONUS**: Added "_android/nanos_example_client/.settings/_" to the foam2 .gitignore after talking to numerous people facing the same problem of always having to remove this file in particular from git in order to make a proper commit to foam2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ build/*
 nano.log*
 /cronjobs
 .vscode/
+android/nanos_example_client/.settings/
--- a/src/foam/box/SessionServerBox.java
+++ b/src/foam/box/SessionServerBox.java
@@ -51,17 +51,30 @@ public class SessionServerBox
         session = new Session();
         session.setId(sessionID);
 
-        // check if 'x-forwarded-for' exists
+        /**
+         * Since we are now accounting for requests which are sent through proxies which have the X-Forwarded-For header
+         * We need to take care of 2 cases:
+         * 1. if the request was sent through a proxy (X-Forwarded-For != null)
+         * 2. if the request was not sent through a proxy (X-Forwarded-For == null)
+         */
         if ( req.getHeader("X-Forwarded-For") != null ) {
-          // set x forwarded for as sourceHost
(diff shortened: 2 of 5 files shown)


jlhughes:
Regarding whitelist, since the Session is updated and saved to the SessionDAO we can use the RuleEngine to enforce the whitelist.

kgrgreer:
This approach basically implements the same logic twice, once for WebAgents and once for Boxes, but if we instead just handled this in foam.nanos.jetty.HttpServer then we would only have to do this once. Also, that object is where we already configure other HTTP server options, so would be a reasonable place to configure the whitelist. Also, doing it there would ensure that unauthenticated services would also get the correct IP-address. Jetty already provides code to do this: https://www.eclipse.org/jetty/javadoc/9.4.8.v20171121/org/eclipse/jetty/server/ForwardedRequestCustomizer.html which we could install in HttpServer if the whitelist is set.

kgrgreer (changes requested):
See comments in discussion.

jlhughes:
@tharmaman we can close this PR yes?

PR #2049, opened 2019-04-15
Creating an AppConfigService

mcarcaso:
Create an AppConfigService and use that directly in the ClientBuilder instead of querying nSpecDAO for appConfig.

The motivation here is to remove the reliance on asking the nspec for the service because that's a permissioned property and the appConfig should be fetchable while unauthenticated.

This also fixes a potential race condition so the ClientBuilder waits for the appConfig before resolving the promise.


src/foam/nanos/app/ContextLookupAppConfigService.js
```diff
@@ -0,0 +1,22 @@
...
+  package: 'foam.nanos.app',
+  name: 'ContextLookupAppConfigService',
+  implements: [
+    'foam.nanos.app.AppConfigService'
+  ],
+  methods: [
+    {
+      name: 'getAppConfig',
+      type: 'foam.nanos.app.AppConfig',
+      javaCode: `
```
kgrgreer, line 17: Have this method take a context as the first argument and then look the AppConfig up from there. This is better because we customize the appConfig in the user's session with their root URL.

src/foam/nanos/client/ClientBuilder.js
```diff
@@ -56,26 +57,33 @@ foam.CLASS({
...
+          var appConfigService = self.ClientAppConfigService.create({
+            delegate: self.RetryBox.create({
+              maxAttempts: 2,
+              delegate: self.HTTPBox.create({
+                method: 'POST',
+                url: 'service/appConfigService'
+              })
+            })
+          });
+          var appConfigPromise = appConfigService.getAppConfig().then(function(appConfig) {
```
kgrgreer: I don't think we need to build this service. Won't it get created along with the others from the nSpecDAO?

src/foam/nanos/client/ClientBuilder.js
```diff
@@ -56,26 +57,33 @@ foam.CLASS({
...
+              name: 'appConfig',
               factory: function() {
-                return foam.json.parse(appConfig, null, this.__subContext__);
+                return appConfig.clone(this.__subContext__);
               }
             });
 
             var version   = appConfig.version;
-            if ( "CLIENT_VERSION" in localStorage ) {
+            if ( 'CLIENT_VERSION' in localStorage ) {
```
jlhughes, line 80: NOTE: AppConfig version is now set during the running of Script 'main'.  The version is extracted from the jar manifest and set on AppConfig.

src/foam/nanos/app/ContextLookupAppConfigService.js
mcarcaso: Done

src/foam/nanos/client/ClientBuilder.js
mcarcaso: Reworked it a little to use nSpecDAO

src/foam/nanos/client/ClientBuilder.js
mcarcaso: Oh good to know. That shouldn't affect this code much though as that's the server setting up the app config that eventually makes its way to the client and will ultimately get checked here. Or is there something I'm missing I need to account for?

mcarcaso: Spoke with Joel offline. This is just good information to know but he also pointed out that groups have a getAppConfig that I'll need to be mindful of. Shouldn't affect this PR so I'm going to merge it now but I think we can use this infrastructure to clean some other things up.

PR #2051, opened 2019-04-16
Improve serialization: Particularly Dates

adamvy:
Before the default Date serialization was to as a number or a formatted string, with special support for Object property to serialize as { class: __Timestamp__, value: #### }.

Now we always serialize as { class: __Timestamp__ }... however Date properties contains a custom toJSON to only serialize as the number or formatted string.

This means that serialization should work out of box no matter where we do it, and Date properties should continue to be efficient.

We also fixed a bug in the __Class__ support for serializing class references.

This change includes an end to end JSON Serialization test. The test makes an object with most of the known serializable properties, goes to/from JSON once on the client, send the object to the server, goes to/from JSON on the server, send the object back to the client, and then compares that result with the initial object.

There is an outstanding bug due to disagreement on Java on whether we should deserialize class references to ClassInfo or the real Class object.  As such this portion of the test is disabled.


src/foam/lib/json/Outputter.java
```diff
@@ -38,6 +38,7 @@ protected SimpleDateFormat initialValue() {
   protected boolean       outputShortNames_    = false;
   protected boolean       outputDefaultValues_ = false;
   protected boolean       outputClassNames_    = true;
+  protected boolean       outputReadableDates_ = true;
```
kgrgreer, line 41: Does everything still work if this is false?

src/foam/core/JSON.js
```diff
@@ -382,11 +381,12 @@ foam.CLASS({
...
     function outputDate(o) {
-      if ( this.formatDatesAsNumbers ) {
-        this.out(o.valueOf());
-      } else {
-        this.out(JSON.stringify(o));
-      }
+      // Serialize an unambiguous timestamp.  Date and DateTime
+      // property can provide an alternative toJSON() and
+      // adapt/javaJSONParser mechanism to save space.
+      this.out('{"class":"__Timestamp__","value":');
```
kgrgreer, line 387: I disagree with this approach for two reasons:
1. It is non standard and will seem odd and unusable by non-FOAM users.
2. It will be slow because it will use extra bandwidth and CPU when communicating between FOAM instances, but between FOAM instances we should be able to encode it however we like to maximize performance.

src/foam/nanos/test/SerializationEchoService.js
```diff
@@ -0,0 +1,16 @@
+foam.CLASS({
+  package: 'foam.nanos.test',
+  name: 'SerializationTestEchoService',
```
mcarcaso: Remove this file? It's a duplicate of SerializationTestEchoService.js

src/foam/core/JSON.js
mcarcaso: What do you suggest we do instead? There's no real standard for dates in json so we're trying to invent our own here.
For models with date properties, we can provide alternate toJSON and adapt/javaJSONParser to maximize performance so I don't think we lose much here with FOAM to FOAM communication.

adamvy: > It is non standard and will seem odd and unusable by non-FOAM users.

So is `__Class__` and `__Property__`.  There is no standard way of serializing Dates to JSON.

> It will be slow because it will use extra bandwidth and CPU when communicating between FOAM instances, but between FOAM instances we should be able to encode it however we like to maximize performance.

Only when sending a Date not as part of a Date property.  What alternative do we have?  The previous strategy of just serializing a Date as a number means that we lose the type information and we end up having difficult to debug errors where method arguments come in as Longs or Ints instead of Dates.

In practice we won't use much more bandwidth because any time we're using a Date property, it has custom serialization to send just the date as a number.  But in cases where we're not a Date property, like an Array, method arguments, some value of a Map, then serialization will still work.

src/foam/lib/json/Outputter.java
adamvy: Yes.

This flag is so that cronjobs appear in the journals as readable strings in case people are editing by hand or reading the journal by hand.  I'd be happy to remove it if we don't care about that use case and always serialize as numbers.

src/foam/nanos/test/SerializationEchoService.js
adamvy: Yes added by mistake, removed.

src/foam/core/JSON.js
adamvy: Another thing we could do is entirely remove support from foam.core.JSON for outputting things that don't exist in JSON (Dates, Classes, Properties. Enums).  And to serialize those types require support to be added to the appropriate Properties and Arguments.

This may provide a cleaner separation of concerns and easier understanding.  JSON Outputter doesn't know about (and can error on) if you try to output any value that doesn't exist in JSON. And any "place" that can store these other non-JSON types can translate them to whatever JSON primitive they deem appropriate.  A plain number or string for dates, an ID string for classes, a plain ordinal for enums (optionally with a type if necessary).

adamvy:
ping

adamvy:
As disucssed offline with Kevin, this does not affect performance of DAO operations, and will be going ahead.

PR #2052, opened 2019-04-17
uncaught exception - notification in ablii fail because of this

Nauna:
https://nanopay.atlassian.net/browse/CPF-350

#CPF-350 was caused here because Long.parseLong throws a NumberFormatException that was not being handled. 

Hence my solution was to wrap in a try catch, if we have a failed attempt to convert token to a long then simply token is not in the right format and verification should simply return false.

### DIFF
--- a/src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
+++ b/src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
@@ -86,7 +86,13 @@ foam.CLASS({
     {
       name: 'verifyToken',
       javaCode: `
-        long code      = Long.parseLong(token, 10);
+        long code;
+        try {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
```diff
@@ -86,7 +86,13 @@ foam.CLASS({
     {
       name: 'verifyToken',
       javaCode: `
-        long code      = Long.parseLong(token, 10);
+        long code;
+        try {
+          code = Long.parseLong(token, 10);
```
kgrgreer, line 91: 10 is the default radix, so you can just do:
code = Long.parseLong(token);

PR #2054, opened 2019-04-17
Create reusable component for list of actions

TW80000:
Required for https://nanopay.atlassian.net/browse/CPF-966.

## Changes

* Moved the logic that generates the context menu in the table view into it's own view so that it can be reused.
  * In particular, I want to use it in the `tableCellFormatter` for properties on models being displayed in the table. For example, on an Invoice model that has a list of attachments, we can generate a dropdown action list view where each action opens the attachment in a new tab.
* Updated the "three vertical dots" icon to match the designs.

### Resting and disabled are the same (as specified by design)
Screen Shot 2019-04-17 at 11 44 35 AM

### Hover (disabled entries won't transition to this state)
Screen Shot 2019-04-17 at 11 44 48 AM

### Active
Screen Shot 2019-04-17 at 11 44 55 AM

### The new view being reused in a different tableCellFormatter
Screen Shot 2019-04-17 at 11 46 23 AM

### DIFF
--- a/src/foam/u2/view/OverlayActionListView.js
+++ b/src/foam/u2/view/OverlayActionListView.js
@@ -155,9 +155,7 @@ foam.CLASS({
       this.overlay_.forEach(this.data, function(action) {
         this.
           start().
-            show(action.createIsAvailable$(self.ConstantSlot.create({
-              value: self.obj
-            }))).
+            show(action.createIsAvailable$(self.obj$)).
             addClass(self.myClass('action')).
             add(action.label).
             on('click', function(evt) {
--- a/src/foam/u2/view/OverlayActionListView.js
+++ b/src/foam/u2/view/OverlayActionListView.js
@@ -13,6 +13,7 @@ foam.CLASS({
 
   requires: [
     'foam.core.ConstantSlot',
+    'foam.core.ExpressionSlot',
     'foam.u2.md.OverlayDropdown'
   ],
 
@@ -22,35 +23,35 @@ foam.CLASS({
 
   properties: [
     {
-      type: 'Action[]',
+      class: 'FObjectArray',
+      of: 'foam.core.Action',
       name: 'data'
     },
     {
-      type: 'FObject',
       name: 'obj'
     },
     {
-      type: 'URL',
+      class: 'URL',
       name: 'activeImageURL',
       value: 'images/Icon_More_Active.svg'
     },
     {
-      type: 'URL',
+      class: 'URL',
       name: 'restingImageURL',
       value: 'images/Icon_More_Resting.svg'
     },
     {
-      type: 'URL',
+      class: 'URL',
       name: 'hoverImageURL',
       value: 'images/Icon_More_Hover.svg'
     },
     {
-      type: 'URL',
+      class: 'URL',
       name: 'disabledImageURL',
       value: 'images/Icon_More_Disabled.svg'
     },
     {
-      type: 'URL',
+      class: 'URL',
       name: 'imageURL_',
       expression: function(restingImageURL, hoverImageURL, disabledImageURL, activeImageURL, hovered_, disabled_, active_) {
         if ( disabled_ ) {
@@ -64,16 +65,28 @@ foam.CLASS({
       }
     },
     {
-      type: 'Boolean',
+      class: 'Boolean',
       name: 'hovered_'
     },
     {
-      type: 'Boolean',
+      class: 'Boolean',
       name: 'disabled_'
     },
     {
-      type: 'Boolean',
+      class: 'Boolean',
       name: 'active_'
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.u2.Element',
+      name: 'overlay_',
+      factory: function() {
+        return this.OverlayDropdown.create();
+      }
+    },
+    {
+      class: 'Boolean',
+      name: 'overlayInitialized_'
     }
   ],
 
@@ -82,11 +95,11 @@ foam.CLASS({
       padding: 10px;
     }
 
-    ^action.disabled {
+    ^action[disabled] {
       color: #aaa;
     }
 
-    ^action:hover:not(.disabled) {
+    ^action:hover:not([disabled]) {
       cursor: pointer;
       background-color: %ACCENTCOLOR%;
     }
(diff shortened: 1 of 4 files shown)


src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,172 @@
...
+    'foam.u2.md.OverlayDropdown'
+  ],
+
+  imports: [
+    'ctrl'
+  ],
+
+  properties: [
+    {
+      type: 'Action[]',
```
adamvy: We still have to specify class: 'FObjectArray', class: 'FObjectProperty', class: 'Boolean', etc for properties at the moment.

In the future all you will have to specify is type: '', but that's not fully hooked up yet.

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,172 @@
...
+
+      var overlay = this.OverlayDropdown.create();
+      this.active_$.follow(overlay.opened$);
+      this.disabled_$.follow(foam.core.ExpressionSlot.create({
+        args: this.data.map((a) => a.createIsAvailable$(this.obj$)),
+        code: (...rest) => ! rest.reduce((l, r) => l || r, false)
+      }));
+
+      this.
+        callIf(this.data.length > 0, function() {
```
adamvy: No need for `callIf(length > 0` just to protect a `forEach(this.data`  If there's no items in data, forEach won't iterate anything.

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,172 @@
...
+          addClass('noselect').
+          start().
+            addClass(this.myClass('icon')).
+            style({ 'background-image': this.imageURL_$ }).
+            on('mouseover', this.onMouseOver).
+            on('mouseout', this.onMouseOut).
+            callIf(this.data.length > 0, function() {
+              this.on('click', function(evt) {
+                if ( self.disabled_ ) return;
+                overlay.open(evt.clientX, evt.clientY);
```
adamvy: I would move building the overlay contents down here.  No sense building all the action views for the overlay until we're about to render it.  Otherwise we're spending all the work every row in the table, but the user is unlikely to click on any particular row.

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,172 @@
...
+          // container with `overflow: hidden` then this overlay won't be cut
+          // off.
+          this.ctrl.add(overlay);
+        }).
+
+        start('span').
+          addClass('noselect').
+          start().
+            addClass(this.myClass('icon')).
+            style({ 'background-image': this.imageURL_$ }).
```
adamvy, line 138: If we're not going to be dynamic on data, we probably don't need to be dynamic on imageURL_ either.

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,172 @@
...
+      cursor: pointer;
+    }
+  `,
+
+  methods: [
+    function initE() {
+      var self = this;
+
+      var overlay = this.OverlayDropdown.create();
+      this.active_$.follow(overlay.opened$);
```
adamvy: Need to make sure we .onDetach() these bindings, otherwise we'll be leaking memory.

src/foam/u2/view/OverlayActionListView.js
adamvy: Oh NVM, I see that imageURL_ is in fact dynamic.

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,172 @@
...
+        }).
+
+        start('span').
+          addClass('noselect').
+          start().
+            addClass(this.myClass('icon')).
+            style({ 'background-image': this.imageURL_$ }).
+            on('mouseover', this.onMouseOver).
+            on('mouseout', this.onMouseOut).
+            callIf(this.data.length > 0, function() {
```
adamvy, line 139: If we have no actions to display, we should consider not hooking up any listeners including mouseOut/mouseOver and make the button look disabled.

Then we barely have to do any work for rows with no actions.

src/foam/u2/view/TableView.js
```diff
@@ -70,13 +70,13 @@ foam.CLASS({
...
 
-    ^vertDots {
+    ^ .vertDots {
       font-size: 20px;
       font-weight: bold;
       padding-right: 12px;
     }
 
-    ^noselect {
+    ^ .noselect {
```
adamvy: Why the change?

CSS engines are typically faster at matching explicit class names like `^noselect`, rather than having to match on parent/child class name relationships like `^ .noselect`

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -206,11 +199,11 @@ foam.CLASS({
...
                     }).
-                    add(' ', view.vertMenuIcon).
-                    addClass(view.myClass('vertDots')).
-                    addClass(view.myClass('noselect'));
+                    tag(view.Image, { data: '/images/Icon_More_Resting.svg' }).
+                    addClass('vertDots').
+                    addClass('noselect');
                   }).
-                  style({ width: 40 }).
+                  style({ width: 60 }).
```
adamvy: I think this should actually be width: '60px'

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -276,47 +274,14 @@ foam.CLASS({
...
-                          overlay.open(evt.clientX, evt.clientY);
-                        });
-                      }).
-                      add(view.vertMenuIcon).
-                    end().
-                  end();
-                });
+                start('td').
+                  addClass(view.myClass('context-menu-cell')).
+                  tag({
```
adamvy: The style I would like to move to for `tag(` and `start(` calls like this is to add
```
requires: [
  'foam.u2.view.OverlayActionListView'
]
```

and do

```
.tag(this.OverlayActionListView, { data: actions, obj: obj })
```

This will make the build tool's job easier.

src/foam/u2/view/OverlayActionListView.js
TW80000: I'll change it, but it seems to work anyway the way it is. Any idea why that would be?

adamvy: Because you essentially just had a regular Property axiom installed, instead of a more type specific Array property.  The base property still works as a property, it just doesn't do any of the extra functionality like type checking and adapting that the more specific property types do.  And if you're not relying on that actual functionality then it will all still work.

You can even specify properties as just a name:

```
properties: [
  'data'
]
```

This creates a plain property named data, which still works like a property.  We use things like
```
properties: [
  {
    class: 'Array'
    name: 'prop'
  }
]
```
Because Array property adds things like a useful default factory and type checking if you try to set it to a non-property.

TW80000: Got it, thank you!

src/foam/u2/view/TableView.js
TW80000: I made that change because I want to use that CSS class in both `UnstyledTableView` and the `OverlayActionListView` but `^noselect` will tie it to the table view.

Maybe this should be moved to our global styles.

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,172 @@
...
+      name: 'disabledImageURL',
+      value: 'images/Icon_More_Disabled.svg'
+    },
+    {
+      type: 'URL',
+      name: 'imageURL_',
+      expression: function(restingImageURL, hoverImageURL, disabledImageURL, activeImageURL, hovered_, disabled_, active_) {
+        if ( disabled_ ) {
+          return `url(${disabledImageURL})`;
+        } else if ( active_ ) {
```
kgrgreer: No need for the else's if you've already returned in the if.

src/foam/u2/view/TableView.js
adamvy: I see.

The downside of sharing CSS like this is that OverlayActionListView is now no longer fully componentized, it _has_ to be rendered inside a TableView in order for it to behave properly, and this dependency is not obvious in any way to a reader of either model.

Personally I would rather just repeat the CSS in OverlayActionListView, _or_ add some top level FOAM stylesheet that is _always_ included and contains some utility classes like .noselect

I don't think we have such a global stylesheet yet, but it might be worth investigating if we have some CSS classes that we want to reuse like this.

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,182 @@
...
+          end().
+        end();
+    },
+
+    function initializeOverlay() {
+      var self = this;
+      this.overlay_.forEach(this.data, function(action) {
+        this.
+          start().
+            show(action.createIsAvailable$(self.ConstantSlot.create({
```
adamvy: `self.obj$` instead of `self.ConstantSlot.create...` would be shorter to type, but it would also create a dynamic binding on self.  If self.obj isn't ever going to change then I guess this works :).

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -0,0 +1,182 @@
...
+              value: self.obj
+            }))).
+            addClass(self.myClass('action')).
+            add(action.label).
+            on('click', function(evt) {
+              self.overlay_.close();
+              action.maybeCall(self.obj.__subContext__, self.obj);
+            }).
+            attrs({
+              disabled: action.createIsEnabled$(self.obj$).map(function(e) {
```
adamvy, line 166: Wait why are we using self.obj$ here but a ConstantSlot above?  We should be consistent.  `self.obj$` is probably the most idiomatic and straightforward to read.

TW80000: Updated to make them consistent.

src/foam/u2/view/TableView.js
TW80000: Good point about dependencies between components. I just copied the CSS to the new view and reverted the relevant changes to CSS selectors for the TableView.

adamvy:
Looks great!

PR #2055, opened 2019-04-18
Add DAO decorator to cause one property to reflect the value of another

TW80000:
https://nanopay.atlassian.net/browse/CPF-854

## Changes

* Added `LinkPropertiesDAO`, which lets you set up a dependency between properties across models.
  * You supply:
    * a `targetProperty` whose value you want to "follow" or "reflect" the value of a different property (possibly on another model), which is called the
    * `sourceProperty`.
    * A `relationshipProperty` is also supplied, which is what's used to look up the target objects in the target DAO, which is specified by
    * the `targetDAOKey`

We can add properties like `payeeName` and `payerName` to the `Invoice` model and then use this decorator to keep `payeeName` and `payerName` updated whenever `Business.businessName` or `User.legalName` changes. In this example:
*  `Invoice.payeeName` is the target property,
* `Business.businessName` is the source property,
* `Invoice.payeeId` is the relationship property, and
* `"invoiceDAO"` is the target DAO key.

### DIFF
--- a/src/foam/dao/LinkPropertiesDAO.js
+++ b/src/foam/dao/LinkPropertiesDAO.js
@@ -1,0 +1,72 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'LinkPropertiesDAO',
+  extends: 'foam.dao.ProxyDAO',
+  flags: ['java'],
+
+  documentation: `
+    When you have two models with a relationship between them and you want a
+    property on one of the models to reflect the value of a property on the
+    other model, you can use this decorator to make that happen.
+  `,
+
+  javaImports: [
+    'foam.core.FObject',
+    'foam.core.PropertyInfo',
+    'java.util.List',
+    'static foam.mlang.MLang.EQ',
+  ],
+
+  properties: [
+    {
+      javaType: 'foam.core.PropertyInfo',
+      javaInfoType: 'foam.core.AbstractObjectPropertyInfo',
+      name: 'sourceProperty',
+      documentation: `This property will be watched for changes.`
+    },
+    {
+      javaType: 'foam.core.PropertyInfo',
+      javaInfoType: 'foam.core.AbstractObjectPropertyInfo',
+      name: 'targetProperty',
+      documentation: `This property will follow the value of 'sourceProperty'.`
+    },
+    {
+      javaType: 'foam.core.PropertyInfo',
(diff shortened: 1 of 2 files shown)


src/foam/dao/LinkPropertiesDAO.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'LinkPropertiesDAO',
+  extends: 'foam.dao.ProxyDAO',
+  flags: ['java'],
+
+  documentation: `
+    When you have two models with a relationship between them and you want a
```
jlhughes, line 14: "with a relationship between them" - models are not using a foam.RELATIONSHIP correct? 
Can we use RELATIONSHIP as it specifies the relationship property and also dao keys, or is this overkill.

TW80000: We could probably do that. In the specific case of invoices though we'd need to update it to have a real relationship between `Invoice` and `User`, as right now it's just a `Reference` property on `Invoice`.

adamvy: Can you give a more concrete example of where/why we want this?

Can we achieve this by adding broader support to reference properties or create a new type of property?

TW80000: > Can you give a more concrete example of where/why we want this?

We have `Invoice` and `User`. `Invoice` has two Reference properties to `User`s: `payeeId` and `payerId`. In the TableView we display those properties as columns and we can use `tableCellFormatter` to do a lookup on the referenced `User` and display it's name, because the default `tableCellFormatter` is just the referenced `User`'s id, which is the value that the Reference property holds. This is all good until the user tries to search or sort by `payeeId` or `payerId`. Then it searches and sorts by the ids rather than the names that we're displaying.

One of the many possible solutions to this problem that we're considering is to add two dummy properties to `Invoice`: `payerName` and `payeeName` that will be of type String and hold the appropriate names. This way displaying is faster because we don't need to do the lookup every time, and sorting and searching works without needing extra UI support that requires more work on the user's part.

In order for those dummy properties to stay up to date, we'd need to decorate `localUserDAO` with a decorator that would update the appropriate `Invoice`s in `invoiceDAO` whenever a `User`'s name changes.

This PR implements that decorator, but generalized to any two models.

> Can we achieve this by adding broader support to reference properties or create a new type of property?

Maybe, that's one of the other solution areas we've thought about. If there were a way for MLangs on the server to do a lookup and check some property on another model then we could use that for the search and sort values rather than the id.

TW80000:
After talking with @kgrgreer in person, I don't think we want to take this approach.

PR #2056, opened 2019-04-19
Whitelisting X-Forwarded-For Requests In foam.nanos.jetty.HttpServer

tharmaman:
Will be posting this PR in Slack on Monday so I don't disturb anyone's Easter weekend!
===========================================

This PR is in regards to ticket [CPF-867](https://nanopay.atlassian.net/browse/CPF-867).

Managed to find a way to implement the whitelisting in foam.nanos.jetty.HttpServer as per @kgrgreer's request. What I ended up having to do was make a subclass **foam.nanos.jetty.WhitelistedForwardedRequestCustomizer** with Jetty's ForwardedRequestCustomizer as the super.

Instead of adding ForwardedRequestCustomizer as a customizer when setting up our Http Connection Factories, I instead added foam.nanos.jetty.WhitelistedForwardedRequestCustomizer. By doing this, we have more freedom in customizing jetty to fit our needs. We are able to both pass a WhiteList and the context into the customizer now. 

I added the **forwardedForProxyWhitelist** as an additional property to **foam.nanos.jetty.HttpServer**.  This allows us to pass in the whitelisted IP addresses in the journal when the "http" service is called as the service calls **foam.nanos.jetty.HttpServer**. I have placed an empty array in the journal as this turns the whitelist off. But as soon as you add values to the array then the whitelist turns on. I included an empty array for staging, nanopay/src and production services journals where the http service is called which I plan to push after this PR is approved.

Within **foam.nanos.jetty.WhitelistedForwardedRequestCustomizer,** the constructor takes two arguments as mentioned above: the whitelist and the context. I had to override the **customize** method as that is where I was also able to make it solely focused on X-Forwarded-For. As if we were to solely use the ForwardedRequestCustomizer, they would also include handling for the rest of the X-Forwarded headers and could potentially cause unexpected behaviours in the future. We can add the new header cases as required.

In total there **4 cases** in how the **main control flow** is handled with this customizer:

1. If there is **no X-Forwarded-For header,** let the request go through as normal
2. If there is an **X-Forwarded-For header, but the whitelist is not set up** (empty array in the journals), then the request will pass and have the X-Forwarded-For header set as it's new remote address
3. If there is an **X-Forwarded-For header, but the whitelist is set up and the request is authorized**, then the request will also pass and have the X-Forwarded-For header set as it's new remote address
4. If there is an **X-Forwarded-For header, but the whitelist is set up and the request is unauthorized**, then the request will be denied. The Proxy IP address will be logged in addition to **throwing a NotAuthorizedException** to stop the program. In the nanopay journals I linked **/nanopay/src/net/nanopay/ui/unauthorizedAccess/unauthorizedAccess.html**  to render when the NotAuthorizedException is thrown. I will make the PR for the nanopay stuff once this PR is good.

There was a bug with how Jetty's HostPort class parses IPv6 addresses without a port number and square brackets. At least with how it's written in the ForwardedRequestCustomizer, they do not automatically normalize IPv6 addresses. **This is a problem because AWS injects the X-Forwarded-For headers for IPv6 addresses without the above two features like so:  2001:DB8::21f:5bff:febf:ce22**

What Jetty's HostPort class would do when an naked IPv6 address such as above was passed to it,  is that **it would truncate the last octet thinking that it is the port number returning an address like: 2001:DB8::21f:5bff:febf** 

This is because after inspecting the control flow, it would not consider the case of naked IPv6 addresses and instead think that it's an IPv4 address since there are no square brackets around it. It would as a cause, think the value after the last semicolon is the port number because that is how an IPv4 address with a port number would be formatted: 127.0.0.1:3000

So what I had to do handle the naked IPv6 case was:
1. X-Forwarded-For address through **two RegEx expressions** to check if it was a **valid yet naked IPv6 address**
2. then we would **enclose square brackets** around the address to normalize the address. 
3. After the IPv6 address was normalized, I would then pass it into host port.

**P.S.** In the case of a chain of X-Forwarded-For's such as: 
X-Forwarded-For: client, proxy1, proxy2

The code will always parse the **LEFTMOST** address when wiring X-Forwarded-For as the new remote host. As per [RFC](https://tools.ietf.org/html/rfc2616#section-4.2), the left-most value holds the oldest header value. Therefore if passed through multiple proxies it should be the client's. 

**P.P.S.** Hope I got the licensing right as my rite of passage 😉


src/foam/nanos/jetty/WhitelistedForwardedRequestCustomizer.java
```diff
@@ -0,0 +1,210 @@
...
+
+  // these are patterns being declared here to handle proper normalization of IPv6 addresses
+  private static final Pattern IPV6_STD_PATTERN = Pattern.compile("^(?:[0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$");
+  private static final Pattern IPV6_HEX_COMPRESSED_PATTERN = Pattern.compile("^((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)::((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)$");
+
+  /**
+   * A constructor which can take in an Set of whitelisted Proxy IP Addresses
+   * @param forwardedForProxyWhitelist a set of whitelisted IP addresses
+   */
+  public WhitelistedForwardedRequestCustomizer( Set<String> forwardedForProxyWhitelist, X x ) {
```
jlhughes: Move X to be the first parameter.

src/foam/nanos/jetty/WhitelistedForwardedRequestCustomizer.java
```diff
@@ -0,0 +1,210 @@
...
+ * For more info on X-Forwarded-For:
+ * @see <a href="http://en.wikipedia.org/wiki/X-Forwarded-For">Wikipedia: X-Forwarded-For</a>
+ *
+ * For implementation of ForwardedRequestCustomizer as per Jetty version 9.4.x:
+ * @see <a href="https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-server/src/main/java/org/eclipse/jetty/server/ForwardedRequestCustomizer.java">Jetty: ForwardedRequestCustomizer</a>
+ */
+public class WhitelistedForwardedRequestCustomizer extends ForwardedRequestCustomizer
+{
+  private Logger logger;
+  private String _forwardedForHeader = HttpHeader.X_FORWARDED_FOR.toString();
```
jlhughes: Is this not a static value?

src/foam/nanos/jetty/WhitelistedForwardedRequestCustomizer.java
```diff
@@ -0,0 +1,210 @@
...
+  private Set<String> forwardedForProxyWhitelist;
+  private X x;
+
+  // these are patterns being declared here to handle proper normalization of IPv6 addresses
+  private static final Pattern IPV6_STD_PATTERN = Pattern.compile("^(?:[0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$");
```
jlhughes: @kgrgreer @adamvy do we have a Thread Local Pattern?

src/foam/nanos/jetty/WhitelistedForwardedRequestCustomizer.java
tharmaman: Done.

src/foam/nanos/jetty/WhitelistedForwardedRequestCustomizer.java
tharmaman: Done. Converted it to a constant as well.

src/foam/nanos/jetty/WhitelistedForwardedRequestCustomizer.java
jlhughes: This is something I can do.  We'll merge now and add a Thread safe pattern in another PR.

PR #2057, opened 2019-04-22
Add remove operation as valid cmd

nanoMichal:
Rule Engine: Rules that were to be applied on REMOVE operation were causing infinite loops. 
Add REMOVE as a valid command.

### DIFF
--- a/src/foam/nanos/ruler/Rule.js
+++ b/src/foam/nanos/ruler/Rule.js
@@ -199,7 +199,7 @@
       ],
       javaCode: `
         getAction().applyAction(x, obj, oldObj, ruler);
-        if ( ! getAfter() && getCmd() != null ) {
+        if ( ! getAfter() ) {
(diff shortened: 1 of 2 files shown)


lchanmann:
@nanoMichal, REMOVE operation caused infinite loop because the `obj` passed to RulerDAO.cmd_ is null and `super.cmd(obj)` call simply delegates back to `cmd_` of the RulerDAO.

It might be better if we change https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/ruler/RulerDAO.js#L102 to `getDelegate().cmd(obj);` as it would prevent self-recurring from the first place.

nanoMichal:
@lchanmann made the change

lchanmann:
LGTM, thanks @nanoMichal.

lchanmann:
@kgrgreer @jlhughes, please review.

PR #2058, opened 2019-04-22
Trim leading and trailing spaces in name

KateMengjunXia:
https://nanopay.atlassian.net/browse/CPF-725

### DIFF
--- a/src/foam/core/Property.js
+++ b/src/foam/core/Property.js
@@ -387,6 +387,7 @@ foam.CLASS({
       var eFactory    = this.exprFactory(prop.expression);
       var FIP         = factory && ( prop.name + '_fip' ); // Factory In Progress
       var fip         = 0;
+      var trim        = prop.trim;
 
       // Factory In Progress (FIP) Support
       // When a factory method is in progress, the object sets a private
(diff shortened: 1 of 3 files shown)


src/foam/core/Property.js
```diff
@@ -522,6 +523,7 @@ foam.CLASS({
           if ( assertValue ) assertValue.call(this, newValue, prop);
 
           if ( preSet ) newValue = preSet.call(this, oldValue, newValue, prop);
+          if ( trim ) newValue = newValue.trim();
```
kgrgreer: I think this should go before adapt and assertValue. Do you agree?

KateMengjunXia: Yes

jlhughes:
@kgrgreer Sounds like the original intent was to default to true.  If we break something with default: true, then it was already broken.

KateMengjunXia:
> @kgrgreer Sounds like the original intent was to default to true. If we break something with default: true, then it was already broken.

Great catch. It seems like when default to true, there are cases where oldValue is a string and newvalue is of some other type, and trim doesn't work.

jlhughes:
OK, so with the additional tests for type 'string', can we switch this to default: true?

src/foam/core/Property.js
mcarcaso: Sorry for the retro review here, but why have trim baked into foam.core.Property? Isn't this just a String concept? We generate an adapt on String that should do this work.

mcarcaso: See: https://github.com/foam-framework/foam2/pull/2106/files

PR #2060, opened 2019-04-23
Fix bug in Action

TW80000:
Fixes https://nanopay.atlassian.net/browse/CPF-1004

---

When `maybeCall` is given no data argument, then `isEnabledFor` was
throwing an error because it can't read the 'slot' method of undefined.
This commit fixes that by using `foam.Function.withArgs` instead, which
won't access anything on `data` unless the `isEnabled` method has
parameters, which it doesn't by default.

Note that this is what `isAvailableFor` has been doing this whole time. I'm not sure why they were different.

This came up when adding an Action to a RichChoiceView that doesn't have
an `isEnabled` method and doesn't take any data either. It would fail
when clicked because the call it makes to `isEnabledFor` was throwing an
error.

I also removed the places where we check if `isAvailable` or `isEnabled` exist since they have default values and will therefore always exist.


kgrgreer:
Let's discuss in person.

src/foam/core/Action.js
```diff
@@ -153,9 +153,7 @@ foam.CLASS({
     },
 
     function isEnabledFor(data) {
-      return this.isEnabled ?
-        data.slot(this.isEnabled).get() :
-        true ;
+      return foam.Function.withArgs(this.isEnabled, data);
```
kgrgreer: I prefer to have isEnabled be null when not used rather than an empty function because it avoids the not negligible amount of work to hookup and then remove the listeners when not needed.

TW80000: Ok, I'll revert the change I made a week or two ago that sets the default functions and update everywhere they're referenced to do null checks.

TW80000: Resolved by 96c3728

PR #2061, opened 2019-04-24
Enhance Ace Editor

lchanmann:
This is to allow configuring Ace editor with `config` as an object literal. For example,
```
foam.CLASS({
  name: 'Thing',
  properties: [
    {
        name: 'jsonString',
        view: {
          class: 'io.c9.ace.Editor',
          config: {
            width: 600, height: 200,
            mode: 'JSON'
          }
        }
    }
  ]
});
```

### DIFF
--- a/src/io/c9/ace/Editor.js
+++ b/src/io/c9/ace/Editor.js
@@ -33,15 +33,11 @@ foam.CLASS({
       name: 'editor'
     },
     {
+      class: 'FObjectProperty',
+      of: 'io.c9.ace.Config',
       name: 'config',
       factory: function() {
         return this.Config.create();
-      },
-      adapt: function(_, nu) {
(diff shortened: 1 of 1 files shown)


src/io/c9/ace/Editor.js
```diff
@@ -36,6 +36,12 @@ foam.CLASS({
       name: 'config',
```
mcarcaso, line 38: I think you'll get the same effect if you make the property an FObjectProperty. i.e. try adding:
```
  class: 'FObjectProperty',
  of: 'io.c9.ace.Config'
```
to the property and remove the `adapt`

lchanmann: Works like a charm, thanks @mcarcaso.

mcarcaso (approved):

PR #2065, opened 2019-04-24
Make VIEW the default controller mode

TW80000:
https://nanopay.atlassian.net/browse/CPF-988

## Notes

* Made the default `controllerMode` of `DAOUpdateControllerView` `VIEW`.
* Moved the cancel action off of `DAOUpdateController` and on to `DAOUpdateControllerView`
  * This was so we can always show the cancel button on the left and not include it on the right. I can filter the list of actions we show on the right instead if people think that's a better solution.
* Updated the layout of the actions so cancel is always on the left and the other actions are on the right
* Updated the width of the view and made it horizontally scrollable if it's too wide
* Added and "Edit" button that lets the user switch to edit mode

### Read-only mode

Screen Shot 2019-04-24 at 10 39 50 AM

### After clicking "Edit"

Screen Shot 2019-04-24 at 10 51 50 AM

### Showing horizontal overflow (it's scrollable)

Screen Shot 2019-04-24 at 10 41 01 AM

### DIFF
--- a/src/foam/comics/DAOUpdateControllerView.js
+++ b/src/foam/comics/DAOUpdateControllerView.js
@@ -37,10 +37,25 @@ foam.CLASS({
   ],
 
   css: `
+    ^ {
+      width: 1024px;
+      margin: auto;
+    }
+    ^action-container {
+      display: flex;
+      justify-content: space-between;
+      margin: 8px 0;
+    }
+    ^detail-container {
+      overflow-x: scroll;
+    }
+    /* TODO: Remove references to nanopay */
     ^ .net-nanopay-ui-ActionView {
       background: #59aadd;
       color: white;
-      margin-right: 4px;
+    }
+    ^ .net-nanopay-ui-ActionView + .net-nanopay-ui-ActionView {
+      margin-left: 4px;
     }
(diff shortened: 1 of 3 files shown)


jlhughes:
I like this.

src/foam/comics/DAOUpdateControllerView.js
```diff
@@ -31,14 +33,30 @@ foam.CLASS({
...
+    ^detail-container {
+      overflow-x: scroll;
+    }
+    /* TODO: Remove references to nanopay */
     ^ .net-nanopay-ui-ActionView {
       background: #59aadd;
       color: white;
-      margin-right: 4px;
+    }
+    ^ .net-nanopay-ui-ActionView + .net-nanopay-ui-ActionView {
```
kgrgreer: We shouldn't have nanopay CSS in foam code.

TW80000: Fixed in 69c1837.

TW80000:
Note that this removes the red colour from the delete button. I plan on doing another PR that gives Actions a way of specifying that they are destructive and should be rendered with special styling.

PR #2068, opened 2019-04-25
Fix the redirection enabled issue when login with 2FA not success

Henry0422:
Fix issue: CPF-841


kgrgreer:
The bigger issue is that you are actually logged in and have authenticated access to all services before the 2FA is finished. If that were working properly then this PR wouldn't be required.

Henry0422:
This PR is out-of-date; please check the new PR: https://github.com/foam-framework/foam2/pull/2100

PR #2070, opened 2019-04-26
Pass the correct context to `maybeCall`

TW80000:
Fixes https://nanopay.atlassian.net/browse/CPF-1032.

### DIFF
--- a/src/foam/u2/view/OverlayActionListView.js
+++ b/src/foam/u2/view/OverlayActionListView.js
@@ -160,7 +160,7 @@ foam.CLASS({
             add(action.label).
             on('click', function(evt) {
               self.overlay_.close();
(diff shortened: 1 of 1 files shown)


kgrgreer:
Why is this correct?

TW80000:
> Why is this correct?

We use the view's context everywhere else where we call `maybeCall`, and that's how it worked before the changes I made to the context menu last week. See https://github.com/foam-framework/foam2/commit/09a6e59e362297946c3702c0bf0f0bf8ca1c552d#diff-ab9fced374e44e291872aa99f9fd2da2L296. The change to using the object's context broke many of the context menu actions in Ablii.

kgrgreer:
But if the view exports anything, then it won't be visible in __context__ as it is only in __subContext__.

PR #2073, opened 2019-04-26
Added searchPlaceholder and search img icon to the RCV view

zahnno:
Updated UI within the rich choice view. 

- Included a scope search image.
- Implemented a custom search placeholder.
- Hide header if none is present.
- Accounted for absence of header in styles.

### DIFF
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -225,7 +225,8 @@ foam.CLASS({
     {
       class: 'String',
       name: 'searchPlaceholder',
-      documentation: 'Replaces search box placeholder with passed in string.'
+      documentation: 'Replaces search box placeholder with passed in string.',
+      value: 'Search...'
     },
     {
       type: 'Action',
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/RichChoiceView.js
```diff
@@ -270,7 +291,7 @@ foam.CLASS({
                   .addClass('search')
                   .add(self.FILTER_.clone().copyFrom({ view: {
                     class: 'foam.u2.view.TextField',
-                    placeholder: 'Search...',
+                    placeholder: this.searchPlaceholder ? this.searchPlaceholder : 'Search...',
```
TW80000: Can just set the `value` of the `searchPlaceholder` property instead.

src/foam/u2/view/RichChoiceView.js
```diff
@@ -286,7 +307,7 @@ foam.CLASS({
               var index = 0;
               return this.E().forEach(sections, function(section) {
                 this
-                  .start().hide(!! section.hideIfEmpty && resp[index].value <= 0)
+                  .start().hide(!! section.hideIfEmpty && resp[index].value <= 0 || ! section.heading)
```
TW80000, line 314: Why do you want to hide sections that don't have headings?

TW80000 (changes requested):
You added an image and added styles for the image but the image is never actually inserted into the view.

src/foam/u2/view/RichChoiceView.js
TW80000: Oh, I see, you're just hiding the heading itself, not the whole section.

zahnno: It's not hiding the section, it's hiding the header is none is passed in. Also did this so the styles applied to self.myClass('heading') aren't applied into the view.

PR #2078, opened 2019-04-27
RichtextEditor

molurin1:
**Notes**
RichTextArea currently allows for basic formattings such as Bold, Italic, Underline, Font Size, Font Face, Justification, Indenting etc. It does not currently support Links and Images but these can be added.


src/foam/u2/texteditor/Popup.js
```diff
@@ -0,0 +1,69 @@
...
+              if ( ! self.hidden ) {
+                self.document.addEventListener('click', function(e2) {
+                  if ( e1.target === e2.target ) return;
+                  if ( ! self.hidden ) self.hidden = true;
+                  self.document.removeEventListener('click', arguments.callee);
+                });
+              }
+            })
+          .end()
+          .start('div').addClass(this.myClass('menu')).hide(self.hidden$)
```
kgrgreer, line 64: .end() doesn't line up with .start()

src/foam/u2/texteditor/Popup.js
```diff
@@ -0,0 +1,69 @@
...
+  methods: [
+    function initE() {
+      var self = this;
+      self.SUPER();
+      self
+          .start()
+            .add(self.button$)
+            .on('click', function(e1) {
+              self.hidden = ! self.hidden;
+              if ( ! self.hidden ) {
```
kgrgreer, line 55: Are you ever going to unhide the popup or is it now hidden forever? If forever, just call self.remove() to remove it from the DOM. Then there's no need to remove the event listener.

molurin1: @kgrgreer Yes plan is to be able to hide and unhide the popup.

PR #2079, opened 2019-04-27
Add additional MDAOs for intervals and cron script to roll up PMs.

jlhughes:
Additional pmInfoDAOs (MDAO) added for 5 minute, 1 hour, 1 day.
Cron script running every minute replaces the pmInfoDAO, and copies
the contents into the other interval DAOs.
The 5 minute, 1 hour, 1 day DAOs are cleared every 5 minutes, 1 hour, 1 day respectively.

Added menu entry for Performance - 1 Day.

The interval MDAOs are intended to be used with the future TimeSeriesDAO.

PR #2080, opened 2019-04-27
Add OMs (Operation Measures) similar to PMs to be used by TM (Thresho…

jlhughes:
…ld Monitor) for service monitoring.

Operation Measures peg or count a single event (compared to PMs which also count but record duration).

Operation Measures are needed to distinguish between request and response for example, to track when a service is stalled or degraded. Track SLA compliance.

Add omInfoDAOs (MDAO) added for 5 minute, 1 hour, 1 day.
Cron script running every minute replaces the omInfoDAO, and copies
the contents into the other interval DAOs.
The 5 minute, 1 hour, 1 day DAOs are cleared every 5 minutes, 1 hour, 1 day respectively.

Added menu entry for Operation Measures - 1 Day.

The interval MDAOs are intended to be used with the future TimeSeriesDAO.


src/foam/nanos/om/DAOOMLoggerCopySink.java
```diff
@@ -0,0 +1,47 @@
+/**
```
TW80000, line 1: This file is added in https://github.com/foam-framework/foam2/pull/2079 as well, just a heads up. We don't want to add it twice.

src/foam/nanos/om/OM.js
```diff
@@ -0,0 +1,53 @@
...
+    {
+      class: 'Class',
+      name: 'classType'
+    },
+    {
+      class: 'String',
+      name: 'name'
+    },
+    {
+      name: 'created',
```
TW80000, line 27: Could implement `CreatedAware` and use `CreatedAwareDAO` instead.

src/foam/nanos/om/services.jrl
```diff
@@ -0,0 +1,7 @@
+p({"class":"foam.nanos.boot.NSpec", "name":"omLogger",                         "lazy":true,  "serve":false, "serviceClass":"foam.nanos.om.DAOOMLogger"})
+p({"class":"foam.nanos.boot.NSpec", "name":"omInfoDAO",                        "lazy":true,  "serve":true,  "serviceScript":"return new foam.dao.ProxyDAO.Builder(x).setDelegate(new foam.dao.MDAO(foam.nanos.om.OMInfo.getOwnClassInfo())).build();","client":"{\"of\":\"foam.nanos.om.OMInfo\"}" })
```
TW80000, line 2: Same comments as the other PR. Why are there plain ProxyDAO decorators on all of these? We should be wrapping these with `AuthenticationDAO`s.

src/foam/nanos/om/OMInfo.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
TW80000: 2019

src/foam/nanos/om/OMDAO.java
```diff
@@ -0,0 +1,80 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
TW80000: Shouldn't this be 2019?

src/foam/nanos/om/OMDAO.java
```diff
@@ -0,0 +1,80 @@
...
+package foam.nanos.om;
+
+import foam.core.FObject;
+import foam.core.X;
+import foam.dao.DAO;
+import foam.dao.ProxyDAO;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+
+public class OMDAO
```
TW80000, line 16: I notice that we don't do anything with selects here. Was that intentional?

src/foam/nanos/om/DAOOMLogger.java
```diff
@@ -0,0 +1,60 @@
...
+    // TODO: could reuse the OMInfo by also using it as the lock object
+    OMInfo info = new OMInfo.Builder(getX())
+      .setClsName(om.getClassType().getId())
+      .setName(om.getName())
+      .build();
+
+    DAO dao = (DAO) getX().get(DAO_NAME);
+
+    synchronized ( getLock(info) ) {
+      OMInfo i = (OMInfo) dao.find(info);
```
TW80000: Missing `inX(x)`.

src/foam/nanos/om/OM.js
jlhughes: Yes, since we put OMs immediately, CreatedAwareDAO will work,  was in the PM mindset where we create but not put until later.

src/foam/nanos/om/services.jrl
jlhughes: These were just copied over from the main ‘services’.  Want to switch to EasyDAO, but it has updates in another pending PR.

src/foam/nanos/om/OMDAO.java
jlhughes: Yes, noticed the same.  This is copy of PMDAO - which I know explicitly did not include select.

src/foam/nanos/om/OM.js
jlhughes: Going to leave this out for now as it adds complexity on the proxy delegate replacement on rollup.  When we decide which way we are going - this or Candlestick then we can make use of CreatedAware.

src/foam/nanos/om/services.jrl
jlhughes: removed.

jlhughes:
replaced by CPF-129

PR #2083, opened 2019-04-29
adding a property check on User.label()

Nauna:
Adding an extra property check to User.label().

I thought this was a safer strategy to have if we seem to be prioritising business name of a user.

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -336,6 +336,7 @@ foam.CLASS({
       },
       javaCode: `
         if ( ! SafetyUtil.isEmpty(getOrganization()) ) return getOrganization();
+        if ( ! SafetyUtil.isEmpty(getBusinessName()) ) return getBusinessName();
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/User.js
```diff
@@ -336,6 +336,7 @@ foam.CLASS({
       },
       javaCode: `
         if ( ! SafetyUtil.isEmpty(getOrganization()) ) return getOrganization();
+        if ( ! SafetyUtil.isEmpty(getBusinessName()) ) return getBusinessName();
```
jlhughes, line 339: please update the javascript to match. 
**suggestion for line 335**
`       return this.organization || this.businessName || ( this.lastName ? this.firstName + ' ' + this.lastName : this.firstName );
`

jlhughes (changes requested):

src/foam/nanos/auth/User.js
Nauna: 😳 😰 😣 😢sorry i missed it

jlhughes (approved):

PR #2084, opened 2019-04-29
Rule engine small fixes

lchanmann:
- Handle exception in rule predicate
- Add availability check


src/foam/nanos/ruler/Rule.js
```diff
@@ -168,13 +169,24 @@
...
+          name: 'isAsync',
+          type: 'Boolean'
         }
       ],
       javaCode: `
-        return getEnabled()
-          && getPredicate().f(
-            x.put("NEW", obj).put("OLD", oldObj)
-          );
+        Boolean isAvailable = null != (isAsync ? getAsyncAction() : getAction());
```
smirnovaka: Why do we combine these two different checks in one? Shouldn't we separate logic for rule.predicate and action == null? If we were to check if a rule is applicable it will be impossible to tell whether predicate is false or action is null

lchanmann: Yeah, that makes sense to not combine them. Also, we can check if action != null before checking if rule is applicable so we don't evaluate the predicate at all if it is the case.

src/foam/nanos/ruler/test/DummyErroneousPredicate.java
```diff
@@ -0,0 +1,14 @@
+package foam.nanos.ruler.test;
+
+import foam.mlang.predicate.AbstractPredicate;
+
+/**
+ * DummyErroneousPredicate - always throws exception when evaluated.
```
kgrgreer, line 6: What is this for?

lchanmann: It's for simulating a predicate that throws exception and being used for testing RulerDAO - https://github.com/lchanmann/foam2/blob/SmallFixRuleEngine/src/foam/nanos/ruler/test/RulerDAOTest.java#L211.

jlhughes (approved):

PR #2087, opened 2019-04-29
Fix validation logic

TW80000:
There were errors being thrown when because validateObj was sometimes a
function and sometimes an array. Errors were also being thrown because
sometimes `this.__context__.data$` was undefined. This commit fixes both
of those issues.

It also adds a call to `onDetach` to make sure the listener is destroyed
when appropriate.


kgrgreer (changes requested):
Don't merge. Just submit new PR to comment out my PR if it is breaking things. Half way between two designs right now but should standardize and not do both and then have to do isInstance() checks.

PR #2089, opened 2019-04-29
Allow Session IP Binding

tharmaman:
This PR is in regards to [CPF-241](https://nanopay.atlassian.net/browse/CPF-241)

After enabling the load balancer whitelisting, we can now remove the temporary fix we had in **SessionServerBox.java** in regards to allowing indexing by using **both session id and remote IP**. Now it we will **only be indexing according to session id** and not allow multiple IPs for a single session.

Removed the commented code and passed the AuthenticationException to force log out the user and prompt them to re-login like so.

logOutNormal

There is still one bug which exists upon hard-refreshing for both when asked to re-login or upon hard-coding a new session token into localStorage as the user will get a white page of death. 

whitePageOfDeath

Can create another ticket for this as it deals with something more deep complex. Could thus far trace this all the way to the **promise** method of **foam.nanos.client.ClientBuilder**


src/foam/box/SessionServerBox.java
```diff
@@ -64,10 +64,9 @@ public void send(Message msg) {
         if ( ! SafetyUtil.equals(session.getRemoteHost(), req.getRemoteHost()) ) {
           // If an existing session is reused with a different remote host then
           // logout the session and force a re-login.
-//          logger.warning("Attempt to use session create for ", session.getRemoteHost(), " from ", req.getRemoteHost());
-//          session.setContext(getX().put(Session.class, session));
-//          session.setRemoteHost(req.getRemoteHost());
-//          sessionDAO.put(session);
+          // logger.warning("Attempt to use session create for ", session.getRemoteHost(), " from ", req.getRemoteHost());
+          msg.replyWithException(new AuthenticationException("Attempt to use session from invalid host."));
```
kgrgreer: I think this is giving the client(/potential hacker) too much information. We should log this attempt, but just return the same exception as we would for an unknown session (which doesn't happen). So we should actually just key the session off of the ip-address and session-id rather than just session-id. Then you could use the same session-id from different hosts without conflict.

jlhughes: I agree that message gives away too much, just return the empty AuthenticationException. 

With x-forward-for your ip won't change.  
Regarding session/ip pairing, if we ignore ip, then one could more 'easily' brute force an attach by just spoofing a sessionid.  At least with the ip/session pair they have to spook the ip as well. 
With x-forward-for the original issue cause by the load-balancers is resolved.

jlhughes:
```suggestion
          msg.replyWithException(new AuthenticationException());
```

jlhughes (approved):

PR #2092, opened 2019-04-30
Let ActionView support sizes

TW80000:
So we can support things like

Screen Shot 2019-04-30 at 9 52 08 AM

for Liquid and

Screen Shot 2019-04-30 at 9 52 56 AM

for Ablii, as well as different button sizes specified by the model rather than custom CSS everywhere.

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -376,6 +376,7 @@ FOAM_FILES([
   { name: "foam/u2/TableView", flags: ['web'] },
   { name: "foam/u2/TableSelection", flags: ['web'] },
   { name: "foam/u2/Scroller", flags: ['web'] },
+  { name: "foam/u2/ButtonSize", flags: ['web'] },
   { name: "foam/u2/UnstyledActionView", flags: ['web'] },
   { name: "foam/u2/ActionView", flags: ['web'] },
   { name: "foam/u2/DetailPropertyView", flags: ['web'] },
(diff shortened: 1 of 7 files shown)


kgrgreer:
I don't agree with this approach. An action might appear in a desktop toolbar or context menu or in a cellphone or tablet sized/styled application. Or just in dramatically differently styled applications: like Ablii vs Liquid. Overspecifying appearance in Model isn't very flexible. I would rather see just a 'styles' map where you could add hints like 'important', 'destructive', etc. that could be interpreted by the UI(s) however they like.

TW80000:
> I don't agree with this approach. An action might appear in a desktop toolbar or context menu or in a cellphone or tablet sized/styled application. Or just in dramatically differently styled applications: like Ablii vs Liquid. Overspecifying appearance in Model isn't very flexible. I would rather see just a 'styles' map where you could add hints like 'important', 'destructive', etc. that could be interpreted by the UI(s) however they like.

What exactly do you disagree with? I'm confused because what you describe is exactly what I'm trying to do with `isSecondary`, `isDestructive`, `isLink`, and `buttonSize` properties. Those are the hints provided by the action, and the views decide how to interpret those hints. For example, in NANOPAY I have a branch with an ActionView for Ablii and an ActionView for Connect, each of which interprets those hints based on how buttons in those applications have been designed. Other views that display actions can choose to ignore or interpret the hints differently, but the vital part is that the hints are there in the first place so that they can be interpreted.

I can move those hints to a map property if you want, though.

TW80000:
Closing in favour of https://github.com/foam-framework/foam2/pull/2095.

PR #2095, opened 2019-05-01
DO_NOT_DELETE: Update Action View

TW80000:
# WARNING
This visually breaks a number of buttons in Ablii and Connect. I've fixed many of them in this branch and in a local branch, but I'm sure I didn't get all of them. We should be careful about when we merge this because we don't want to be in a situation where we merge this and peg nanopay to a commit past this one before the fixes on the nanopay side are ready.

## Changes

* Updates `foam.u2.ActionView`'s CSS to match this design system:

Screen Shot 2019-05-01 at 4 50 39 PM

* Added `ButtonSize` enum and `size` property on `foam.u2.ActionView`
* Added `ButtonStyle` enum and `buttonStyle` property on `foam.u2.ActionView`
  * The styles are `PRIMARY` (the default), `SECONDARY`, `TERTIARY`, and `UNSTYLED`, which is useful when you don't want the button styles but you still want the features of actions
* Merged `UnstyledActionView` into `foam.u2.ActionView` now that you can get an unstyled action view through the `buttonStyle` property instead
* Added a CSS colour macro for header colour
* Replaced a number of hard-coded colours with references to named CSS colour macros
* Updated the submenu text so that it's no longer white, which is illegible with the new accent colour, which is a light blue

Changing the CSS for the base ActionView causes a lot of places that were overriding the CSS of actions to break. I fixed a bunch of them here but I'm sure I didn't get them all. Here is a list of the places I've fixed already:

* Sign in page
* Clear button in ReciprocalSearch pane
* Footer buttons/links

I also made use of the new button size support by setting the primary action in the DAO controller to large and actions rendered in table cells to small.

I also replaced all references to nanopay's ActionView with references to foam's newly updated ActionView.

## Screenshots

Screen Shot 2019-05-01 at 5 04 35 PM
Screen Shot 2019-05-02 at 3 58 21 PM
Screen Shot 2019-05-01 at 5 05 05 PM


src/foam/u2/view/PasswordView.js
```diff
@@ -79,7 +79,7 @@ foam.CLASS({
 
       this.addClass(this.myClass()).start().
         addClass('input-field-container').
-        start(foam.u2.tag.Input, {
+        start(foam.u2.TextField, {
```
adamvy: Ideally we should be doing `requires: [ 'foam.u2.TextField' ]` and then `start(this.TextField` but that's a minor issue, something to keep in mind.

src/foam/nanos/u2/navigation/FooterView.js
```diff
@@ -82,6 +84,11 @@ foam.CLASS({
     ^ .mini-links {
       float: left;
     }
+    ^ button {
+      border: none !important;
```
adamvy: Is there a way to do this without `!important` ?  Using !important makes it harder to override later, as there's not `!evenMoreImportant` in CSS.

adamvy:
The approach looks good, added a few thoughts that occured to me while reading.

src/foam/u2/view/PasswordView.js
TW80000: Resolved by ae634b2.

src/foam/nanos/u2/navigation/FooterView.js
TW80000: I added a fourth `ButtonStyle` value in the enum: `UNSTYLED`, which can be used in cases like this where we want to use an action for all the features that come with them, but don't want the default button styles. This will let us remove a lot of CSS overrides like this file was doing.

src/foam/comics/DAOControllerView.js
```diff
@@ -149,14 +149,17 @@ foam.CLASS({
             .callIfElse(this.data.primaryAction, function() {
               this.startContext({ data: self })
                 .start()
-                  .add(self.data.primaryAction)
+                  .tag(self.data.primaryAction, { size: 'LARGE' })
```
kgrgreer, line 152: Are we getting rid of 'size'? Is this PR still valid?

TW80000: No, size is just moving from the Action itself to ActionView.

kgrgreer:
Has a merge conflict.

TW80000:
> Has a merge conflict.

@kgrgreer Resolved

PR #2098, opened 2019-05-02
Add javaValidateObj to properties and make all FObject in java validatable

mcarcaso:
Included a simple test in the journal that is:
```
valid = new foam.nanos.demo.DemoObject.Builder(x).setIsValid(true).build();
valid.validate(x);

invalid = new foam.nanos.demo.DemoObject.Builder(x).setIsValid(false).build();
thrown = false;
try {
    invalid.validate(x);
} catch (Exception e) {
    thrown = true;
}
test(thrown, "Thrown!");
```


src/foam/nanos/demo/DemoObject.js
```diff
@@ -30,6 +30,16 @@ foam.CLASS({
     {
       class: 'Int',
       name: 'value'
+    },
+    {
+      class: 'Boolean',
+      name: 'isValid',
+      value: true,
+      javaValidateObj: `
+        boolean isValid = (boolean) obj.getProperty("isValid");
```
kgrgreer, line 39: This line of code is odd. Better to just do: ((DemoObject) obj).getIsValid();

kgrgreer:
has merge conflicts

src/foam/nanos/demo/DemoObject.js
mcarcaso: Removed this all together and did an example on foam.nanos.auth.Phone. Noticed we don't generate anything for messages in java so I added that while I was at it. We'll probably want to change the way we generate them when we get more i18n stuff together but for now I figure this is fine. Let me know if you disagree and I can change it up.

PR #2100, opened 2019-05-02
Add TwoFactorAuthService

Henry0422:
Fix the 2FA bypassing issue https://nanopay.atlassian.net/browse/CPF-841
Complete the main task of https://nanopay.atlassian.net/browse/CPF-1036


src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
```diff
@@ -86,7 +86,7 @@ foam.CLASS({
     {
       name: 'verifyToken',
       javaCode: `
-        long code      = Long.parseLong(token);
+        long code;
```
Henry0422, line 89: Fix the issue in https://nanopay.atlassian.net/browse/CPF-350

mcarcaso (approved):

src/services
```diff
@@ -69,7 +69,7 @@ p({"class":"foam.nanos.boot.NSpec", "name":"serviceProviderDAO",               "
 
 p({"class":"foam.nanos.boot.NSpec", "name":"emailToken",                       "lazy":true,  "serve":true,  "serviceClass":"foam.nanos.auth.email.EmailTokenService","boxClass":"foam.nanos.auth.token.TokenServiceSkeleton","client":"{\"class\":\"foam.nanos.auth.token.ClientTokenService\"}"})
 p({"class":"foam.nanos.boot.NSpec", "name":"verifyEmail",                      "lazy":true,  "serve":false, "authenticate": false, "serviceClass":"foam.nanos.auth.email.EmailVerificationWebAgent"})
-p({"class":"foam.nanos.boot.NSpec", "name":"twofactor",                        "lazy":true,  "serve":true,  "boxClass":"foam.nanos.auth.twofactor.OTPAuthServiceSkeleton", "serviceClass":"foam.nanos.auth.twofactor.GoogleTOTPAuthService", "client":"{\"class\":\"foam.nanos.auth.twofactor.ClientOTPAuthService\"}"})
+p({"class":"foam.nanos.boot.NSpec", "name":"twofactor",                        "lazy":true,  "serve":true,  "authenticate": false, "boxClass":"foam.nanos.auth.twofactor.OTPAuthServiceSkeleton", "serviceClass":"foam.nanos.auth.twofactor.GoogleTOTPAuthService", "client":"{\"class\":\"foam.nanos.auth.twofactor.ClientOTPAuthService\"}"})
```
mcarcaso, line 72: Let's talk about this.

mcarcaso (changes requested):

adamvy:
We're doing this work out of order.  We have to redesign work to do on AuthService in general first.

First we need to model a number of Credential objects, such as UserAndPasswordCredentials, and SecondFactorCredentials.

Then we need to update the auth service in two ways.  We need an `auth(Credentials)` method which takes any credentials object.  And `check()` needs to be modified to throw an exception rather than return a boolean.

The idea is that, when the user makes a request to the server, the server will call `auth()` to see if the session has the necessary credentials.  If they don't then it will throw an exception such as `MissingCredentialsException(UserAndPasswordCredentials.class)`

If the client receives a MissingCredentialsException, it can then render a DetailView for the type of Credentials that were indicated as missing in the exception.  After which it can retry the request if it wants.

The current mechanism of just returning a boolean means we don't have enough information in the client to know what credentials we're missing.

As an example, UserPasswordAuthService should look something like
`
For instace UserAndPasswordAuthService
`
class UserPasswordAuthService { 
  auth(X, Credentials c) {
    if ( ! ( c instanceof UserPasswordCredentails ) {
      return getDelegate().auth(c);
    }
    if ( c is valid user and password ) {
      update X.session
    }
    throw InvalidCredentialsException
  }

  check(x, permission) {
    if ( X.session does not have valid UserPasswordCredentials ) {
      throw MissingCredentialsExceptoin(UserPasswordCredentials.class)
    }

     // do any other checks we want relating to user/password
    getDelegate().check(x, permission)
  }
}
`

We need to fix the design of AuthService first before we can go implement the TwoFactorAuthService.

adamvy:
Ugh the formatting of my pseudocode is broken, pasting again here

`
class UserPasswordAuthService {
  auth(X, Credentials c) {
    if ( ! ( c instanceof UserPasswordCredentails ) {
      return getDelegate().auth(c);
    }
    if ( c is not a valid user and password ) {
      throw InvalidCredentialsException(c);
    }

    update X.session;
  }
  

  check(x, permission) {
    if ( X.session does not have valid UserPasswordCredentials ) {
      throw MissingCredentialsExceptoin(UserPasswordCredentials.class)
    }
    
    // do any other checks we want relating to user/password
     
    getDelegate().check(x, permission)
  }
}
`

adamvy:
Sorry I'm bad at github apparnetly, here's the pseudocode for real.

```
class UserPasswordAuthService {
  auth(X, Credentials c) {
    if ( ! ( c instanceof UserPasswordCredentails ) {
      return getDelegate().auth(c);
    }
    if ( c is not a valid user and password ) {
      throw InvalidCredentialsException(c);
    }

    update X.session;
  }
  

  check(x, permission) {
    if ( X.session does not have valid UserPasswordCredentials ) {
      throw MissingCredentialsExceptoin(UserPasswordCredentials.class)
    }
    
    // do any other checks we want relating to user/password
     
    getDelegate().check(x, permission)
  }
}
```

adamvy:
Okay as discussed, this is a good fix for the issue for now, and we're going to work on the auth system redesign afterwards.

LGTM

PR #2102, opened 2019-05-03
CandlestickDAOChartCView

kennyqiyenkan:
A chart that is created from chartjs using the data inside a CandlestickDAO.


src/org/chartjs/CandlestickDAOChartCView.js
```diff
@@ -0,0 +1,204 @@
...
+    {
+      class: 'Map',
+      name: 'config',
+      documentation: `
+        The config map that is expected by chartjs. Structure and information can be found in chartjs.org's documentation.
+      `
+    },
+    {
+      class: 'Map',
+      name: 'candlestickMap',
```
mcarcaso: Do you need this as a property?

src/org/chartjs/CandlestickDAOChartCView.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.onDetach(this.data$proxy.listen(this.FnSink.create({ fn: this.dataUpdate })));
+      var self = this;
+      this.add(this.slot(function( config ) {
+        if ( config ) {
+          return this.E().tag(self.ChartCView, { config: self.config });
```
mcarcaso: self.config can be just config

src/org/chartjs/CandlestickDAOChartCView.js
```diff
@@ -0,0 +1,204 @@
...
+      code: function() {
+        // Data (which should be a CandlestickDAO) has been updated
+        if ( ! this.data ) {
+          // data was set to null
+          return;
+        }
+
+        var self = this;
+        // Get all candlesticks categorized by their key, and ordered in ascending order by their closeTime.
+        this.data.orderBy(this.Candlestick.CLOSE_TIME).select(this.GROUP_BY(this.Candlestick.KEY, this.ArraySink.create())).then( function(a) {
```
mcarcaso: There's a bunch of jumping around here that makes this kinda tough to read.
generateConfig goes to generateData which goes to generateDataSets, etc.

Check out https://github.com/foam-framework/foam2/pull/2103 and the test that's in the comment. I think using that approach to get the dataset together can help clean this up a bunch and I think the entire config can be done in one method since I don't see us needing to reuse much of these methods.

kennyqiyenkan: I personally find that to be pretty hard to read. I'll see what I can do simplify the functions

PR #2105, opened 2019-05-03
Fix bug where editing fails when you go through a relationship table row

TW80000:
https://nanopay.atlassian.net/browse/CPF-1060

## Problem

When you edit an FObject in the DAO controller view, you get brought to a DetailView for that FObject. If that FObject has relationship properties, they will get rendered as DetailViews in tabs. In each relationship property's DetailView, there is only one property displayed (for ManyToManyRelationshipProperties): `dao :: DAOProperty`. The default view for DAOProperties is an InlineBrowserView, which contains a table view. When you double click a row in that table view, you get brought to a new DetailView for that object. However, now the underlying DAO is the relationship property, which is currently decorated with a `ReadOnlyDAO` decorator. This means that when you try to edit the FObject you double-clicked on in the relationship property table and click "Save", it throws an error because you can't put to a `ReadOnlyDAO`.

## Changes

* Removed the `ReadOnlyDAO` decorator on the `dao` property of the `ManyToManyRelationshipImpl` model
* Fixed linter errors in `DetailView`

## Discussion

I'm not aware of why it was added in the first place. I'm not saying it wasn't necessary, I'm just saying that I'm not aware of the reason. I'm expecting there was a good reason and so this won't be merged. In that case, does anyone have any ideas on how I can do this? I tried making a custom DetailView for `ManyToManyRelationshipImpl` but the data of that view is always set to the `dao` property's value anyway, so that doesn't fix anything. I was hoping that if the detail view specified the data then the property's value wouldn't be used, but that's not the case because of this bit of code that overwrites it: https://github.com/foam-framework/foam2/blob/master/src/foam/u2/Element.js#L2145-L2147. If I could override `toE` on just that `ManyToManyRelationshipImpl.dao` property then it would work, but I couldn't find a way to do that.

### DIFF
--- a/src/foam/u2/DetailView.js
+++ b/src/foam/u2/DetailView.js
@@ -70,10 +70,10 @@ foam.CLASS({
       // TODO: Make an FObjectArray when it validates properly
       preSet: function(_, ps) {
         foam.assert(ps, 'Properties required.');
-        for ( var i = 0 ; i < ps.length ; i++ ) {
+        for ( var i = 0; i < ps.length; i++ ) {
           foam.assert(
               foam.core.Property.isInstance(ps[i]),
-              "Non-Property in 'properties' list:",
+              `Non-Property in 'properties' list:`,
               ps);
         }
         return ps;
@@ -82,7 +82,9 @@ foam.CLASS({
         if ( ! of ) return [];
         return this.of.getAxiomsByClass(foam.core.Property).
           // TODO: this is a temporary fix, but Visibility.HIDDEN should be included and could be switched
-          filter(function(p) { return ! ( p.hidden || p.visibility === foam.u2.Visibility.HIDDEN ); });
+          filter(function(p) {
+            return ! ( p.hidden || p.visibility === foam.u2.Visibility.HIDDEN );
+          });
       }
     },
     {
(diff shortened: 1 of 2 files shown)


jlhughes (approved):

jlhughes (approved):

src/foam/u2/DetailView.js
```diff
@@ -70,10 +70,10 @@ foam.CLASS({
       // TODO: Make an FObjectArray when it validates properly
       preSet: function(_, ps) {
         foam.assert(ps, 'Properties required.');
-        for ( var i = 0 ; i < ps.length ; i++ ) {
+        for ( var i = 0; i < ps.length; i++ ) {
```
kgrgreer, line 73: The original version is the properly formatted one.

src/foam/u2/DetailView.js
```diff
@@ -70,10 +70,10 @@ foam.CLASS({
...
           foam.assert(
               foam.core.Property.isInstance(ps[i]),
-              "Non-Property in 'properties' list:",
+              `Non-Property in 'properties' list:`,
```
kgrgreer, line 76: The left line follows our/Google's convention, which is to use single quotes unless your string contains single quotes, in which case you should use double quotes.

TW80000:
@kgrgreer I'll try to update our linter settings to reflect that.

PR #2108, opened 2019-05-03
fixing an incorrect url for USBankAccounts

Nauna:
So this view is only used as the default for FileProperty. 

However the issue with blobService, was the url that was returned contained the wrong Id. 

Switching this to data.address lets us use the preset url on files and avoid the incorrect result returned from blobService.

### DIFF
--- a/src/foam/u2/view/FileView.js
+++ b/src/foam/u2/view/FileView.js
@@ -14,23 +14,16 @@ foam.CLASS({
     'foam.nanos.fs.File'
   ],
 
-  imports: [
-    'blobService'
-  ],
-
   properties: [
     'data'
   ],
 
   methods: [
     function initE() {
-      var view = this;
       this.setNodeName('span')
         .start('input').attrs({ type: 'file' }).on('change', this.onChange).end()
         .add(this.slot(function(data) {
-          var file = data && data.data;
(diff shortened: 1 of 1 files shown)


TW80000 (approved):

jlhughes (approved):

src/foam/u2/view/FileView.js
```diff
@@ -14,23 +14,16 @@ foam.CLASS({
...
   methods: [
     function initE() {
-      var view = this;
       this.setNodeName('span')
         .start('input').attrs({ type: 'file' }).on('change', this.onChange).end()
         .add(this.slot(function(data) {
-          var file = data && data.data;
-          var url = file && view.blobService.urlFor(file);
-          return ! url ? this.E('span') : this.E('a').attrs({ href: url }).add('Download')
+          return ! data ? this.E('span') : this.E('a').attrs({ href: data.address }).add('Download');
```
kgrgreer, line 26: Very minor coding style comment: better to remove the ! and then just reverse the order of the arguments. Makes easier to read.

Also, if you don't need the 'span' when the data is null, you could just return 'null', and I think it would still work. So then you could just do:
return data && this.E('a').attrs({ href: data.address }).add('Download');

PR #2110, opened 2019-05-05
When pulling in constants from a library, Check if objects have a `cl…

mcarcaso:
…ass` property and create an instance of them if so.

When the builder runs, it serializes everything and can result in constants in a foam.LIB to go from something like:
```
    Compact: foam.json.Outputter.create({
      pretty: false,
      formatDatesAsNumbers: true,
      formatFunctionsAsStrings: false,
      outputDefaultValues: false,
      pretty: false,
      strict: false
    },
```
to this:
```
Compact: {
                       class: "foam.json.Outputter",
                       indentStr: "",
                       nlStr: "",
                       postColonStr: "",
                       alwaysQuoteKeys: false,
                       formatDatesAsNumbers: true,
                       formatFunctionsAsStrings: false,
                       outputDefaultValues: false,
                       pretty: false,
                       strict: false
},
```

This PR makes constants in foam.LIB convert objects with a `class` property to an instance of that class.

### DIFF
--- a/src/foam/core/lib.js
+++ b/src/foam/core/lib.js
@@ -124,7 +124,13 @@ foam.LIB = function LIB(model) {
       typeof model.constants === 'object',
       'Constants must be a map.');
 
-    for ( var key in model.constants ) root[key] = model.constants[key];
+    for ( var key in model.constants ) {
+      var v = root[key] = model.constants[key];
+      if ( foam.Object.isInstance(v) && v.class ) {
+        v = foam.lookup(v.class).create(v);
+      }
+      root[key] = v;
+    }
   }
 
   if ( model.methods ) {


jlhughes (approved):

src/foam/core/lib.js
```diff
@@ -124,7 +124,13 @@ foam.LIB = function LIB(model) {
       typeof model.constants === 'object',
       'Constants must be a map.');
 
-    for ( var key in model.constants ) root[key] = model.constants[key];
+    for ( var key in model.constants ) {
```
adamvy, line 127: I know I fixed this before somewhere but I can't find the code, no idea what happened to it.  The main limitation here is that we're one-offing a solution for constants and not methods or any other axioms that LIBs may eventually have.

A more general solution would be to call `foam.json.parse` here, or maybe to improve the non-strict mode of JSON Outputter to output literal calls to `foam.json.Outputter.create()` as opposed to `{ class: 'foam.json.Outputter' }` but there may be other issues with that.

This is okay for now but we should keep this in mind.

PR #2112, opened 2019-05-05
Move the code to adapt a date to foam.Date.toDate and make it handle …

mcarcaso:
…objects with class: '__Timestamp__'. Use this method to adapt values.


jlhughes:
This and 2111 address issues with CalendarDatePicker when running from foam-bin.js.

adamvy:
As discussed with Kevin I believe we're going to drop this.  `__Timestamp__` shouldn't leak anywhere beyond the JSON outputter, and Kevin is going to add general support for adapt running on default values.

PR #2114, opened 2019-05-06
Fix CalendarDatePicker defaulting to today's date

TW80000:
This is undesirable because we want frontend validation to be able to catch when a User hasn't picked a date yet.

### DIFF
--- a/src/foam/u2/view/date/CalendarDatePicker.js
+++ b/src/foam/u2/view/date/CalendarDatePicker.js
@@ -14,27 +14,7 @@ foam.CLASS({
   properties: [
     {
       class: 'Date',
-      name: 'data_',
-      factory: function() { return new Date(); },
-      postSet: function(_, n) {
-        if ( this.skipDateUpdate ) {
-          this.skipDateUpdate = false;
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/date/CalendarDatePicker.js
```diff
@@ -125,22 +105,24 @@ foam.CLASS({
...
     function initE() {
       var self = this;
-      this.startContext({ data_: this.data_ }).
+      this.startContext({ data: this.data }).
         addClass(this.myClass('calendar_table')).
-        add(this.slot(function(data_) {
-          var month = data_.getMonth();
-          var year = data_.getYear() + 1900;
+        add(this.slot(function(data) {
+          data = data || new Date();
```
TW80000, line 111: This is the important change that @mcarcaso suggested in person. If the data is null or undefined, we'll just use today's date internally in this function to calculate how the calendar should be rendered. That is what the `data_` property was originally created to address, but doing assignment here is simpler and avoids the issue where `data` was being set when `data_` changed.

PR #2116, opened 2019-05-06
 Fix the redirect issue if user not acting as business

Henry0422:
Fix issue: https://nanopay.atlassian.net/browse/CPF-1116

It might fix: https://nanopay.atlassian.net/browse/CPF-883


Henry0422:
@mcarcaso @TW80000 Can you review this PR?

TW80000 (approved):

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -183,18 +183,19 @@ foam.CLASS({
     function init() {
       this.SUPER();
       var self = this;
+
+      window.onpopstate = async function(event) {
+        var hid = location.hash.substr(1);
+        if ( hid ) {
+          var menu = await this.client.menuDAO.find(hid);
```
TW80000: We might want to add a null check to make sure `this.client` is set here.

TW80000: We might want to add a null check to make sure `this.client` is set here.

TW80000:
Whoops, didn't realize this was FOAM. Let me know if you want me to open a PR reverting this.

PR #2119, opened 2019-05-08
CPF-1233/addNameAttributes #comment added name attributes to foam ele…

noodlemoodle:
…ments for testing #time 3hr

### DIFF
--- a/src/foam/u2/UnstyledActionView.js
+++ b/src/foam/u2/UnstyledActionView.js
@@ -93,6 +93,7 @@ foam.CLASS({
 
   methods: [
     function initE() {
+      this.attrs({name: this.action.name});
       this.initCls();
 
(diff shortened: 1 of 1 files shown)


src/foam/u2/UnstyledActionView.js
```diff
@@ -93,6 +93,7 @@ foam.CLASS({
 
   methods: [
     function initE() {
+      this.attrs({name: this.action.name});
```
TW80000: This should be moved inside the `if ( this.action ) { ... }` block below so we know the action is set.

Also, our style convention is to use spaces inside of objects written on a single line, like this: `{ name: this.action.name }`.

TW80000 (changes requested):

TW80000 (approved):

PR #2127, opened 2019-05-09
Added Helper Tooltips to SectionedDetailPropertyView

tharmaman:


tharmaman:
Going to move the icon to foam/u2/images

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -37,11 +37,11 @@ foam.CLASS({
           return self.E()
             .start(self.Rows)
               .add(prop.label$)
-              .start(self.Cols)
-                .start(self.Col, { flex: 1 })
-                  .add(prop)
+              .start(self.Cols, { alignmentType: foam.u2.layout.AlignmentType.START })
+                .add(prop)
+                .start({class: 'foam.u2.tag.Image', data: 'images/question-icon.svg'})
+                  .show(prop.help$.map((help) => typeof help === 'string' && help.length > 0)).attrs({ title: prop.help$ })
```
kgrgreer: does just this work:
  .show(prop.help$) 

Since both an empty string or a '' will both be converted to false.

kgrgreer: Easier to read if you put the .attrs() on the next line.

kgrgreer: Actually, the prop.help isn't going to change its value, so you don't need to use the slot and can just do .show(prop.help), or even better, don't even add it if the help isn't set:
.callIf(prop.help, function() { this.start().attrs().end(); }).

tharmaman: Done

PR #2130, opened 2019-05-09
StringArrayRowView replacement

mcarcaso:


mcarcaso:
Merging because the view that I'm replacing doesn't work anyway. Can fix any comments up after the fact.

PR #2131, opened 2019-05-10
Let the permission required to see an action be customizable

TW80000:
Sometimes we might want to re-use existing permissions. For example, I might want to use a permission related to the data being worked on in the view rather than a permission related to the action itself and view it's defined on.


jlhughes (approved):
We are blocked on this. Approving, we'll post review.

adamvy:
lgtm

PR #2132, opened 2019-05-10
Improved action permission support

TW80000:
* maybeCall respects permissions
* permissions can apply to enabled now, not just available
* you can specify multiple permissions for each

Lets you do things like this:

### Multiple permissions
```JavaScript
actions: [
  {
    name: 'foo',
    permissionConfig: { enabled: ['currency.read.CAD', 'invoice.pay'] }
  }
]
```

### Different permissions for available vs enabled
```JavaScript
actions: [
  {
    name: 'foo',
    permissionConfig: {
      enabled: ['currency.read.CAD']
      available: ['invoice.pay']
    }
  }
]
```

### Same as before (still works)
```JavaScript
actions: [
  {
    name: 'foo',
    permissionRequired: true
  }
]
```

### DIFF
--- a/src/foam/core/Action.js
+++ b/src/foam/core/Action.js
@@ -220,7 +220,7 @@ foam.CLASS({
 
     function hasPermissions_(configKey, data$) {
       if ( this.permissionConfig == null ) return Promise.resolve(true);
-      return Promise.all(this.permissionConfig[configKey].map((p) => this.checkPermission(data$.get().__subContext__, p))).then((values) => values.reduce((l, r) => l && r), true);
+      return Promise.all(this.permissionConfig[configKey].map((p) => this.checkPermission(data$.get().__subContext__, p))).then((values) => values.reduce((l, r) => l && r, true));
     }
   ]
 });
--- a/src/foam/core/Action.js
+++ b/src/foam/core/Action.js
@@ -24,7 +24,8 @@ foam.CLASS({
   requires: [
     'foam.core.ConstantSlot',
     'foam.core.ExpressionSlot',
-    'foam.core.PromiseSlot'
+    'foam.core.PromiseSlot',
+    'foam.nanos.auth.ActionPermissionConfiguration'
   ],
 
   documentation: 'An Action is a method with extra GUI support.',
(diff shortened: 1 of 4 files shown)


jlhughes (approved):
Blocking. Will post review.

src/foam/core/Action.js
```diff
@@ -157,20 +171,22 @@ foam.CLASS({
...
 
     function createIsAvailable$(data$) {
-      return this.createSlot_(this.isAvailable, data$);
+      return this.createSlot_(this.isAvailable, 'available', data$);
     },
 
-    function maybeCall(ctx, data) {
-      // TODO: permission check
-      if ( this.isEnabledFor(data) && this.isAvailableFor(data) ) {
+    async function maybeCall(ctx, data) {
```
adamvy, line 177: This is going to cause problems if an Action ever does window.open() to say open a link in another window it will run up against popup blockers which only let you open windows during synchronous handling of a 'click' event handler.

Not a game breaking restriction but something to keep in mind.

src/foam/u2/ActionView.js
```diff
@@ -344,11 +344,11 @@ foam.CLASS({
       if ( this.action ) {
         this.attrs({ name: this.action.name });
 
-        if ( this.action.isAvailable ) {
+        if ( this.action.isAvailable || this.action.permissionConfig != null ) {
```
adamvy, line 347: This seems overly complex.  We don't need to update all the views to understand permissions, all we need to do us ensure that the isAvailable and isEnabled slot's created by the action _also_ take permissions into account if necessary.

I think we should just put enabledPermissions and availablePermissions as properties on Action, and then expand the createIsEnabled and createIsAvailable code to create slots which also check the permissions.

As a result the interface to an Action never changes and _all_ the views which currently check isEnabled/isAvailable will pick up permission support automatically.

TW80000: > This seems overly complex. We don't need to update all the views to understand permissions, all we need to do us ensure that the isAvailable and isEnabled slot's created by the action _also_ take permissions into account if necessary.

They already do. `createIsEnabled$` and `createIsAvailable$` already take permissions into account. The only reason we have these `if`s in ActionView is as a tiny optimization to avoid setting up the slots if they won't be doing anything anyway. ActionView would work just as well without these checks and other ActionViews will as well.

> I think we should just put enabledPermissions and availablePermissions as properties on Action...

I can change the interface to let you set those instead of providing an `ActionPermissionConfiguration` object if you think that's a better interface.

PR #2135, opened 2019-05-10
Check permissions in createVisibilityFor

mcarcaso:


src/foam/core/Slot.js
```diff
@@ -435,6 +434,13 @@ foam.CLASS({
     },
     {
       name: 'value',
+      preSet: function(_, n) {
+        if ( n && n.then ) {
+          this.promise = n;
```
adamvy, line 439: There's a race condition.

Assume I have an expression which returns a promise, and then triggers again and returns a value.

```
ExpressionSlot.create({
  code: function(a, b) {
    return a ?
      new Promise(function(r) { setTimeout(function() { r(100); }, 1000); }) :
      b;
  }
});
```

If a is first true, and then false, we will set the value to b, and then after 1 second the value will be set to `100`, but this is incorrect as it should be `b` since the current input to the expression is `a = false`.

adamvy: As discussed we should be able to fix this by clearing the promise property if `n` is not a promise.

PR #2138, opened 2019-05-10
Fix bug in Rule Engine

TW80000:
There's a bug where after each sync rule executes the object is put to
the underlying DAO. This was only intended to happen for async actions
according to Chanmann.

Say you have a rule that performs some validation and throws an error if some condition is met to prevent the put to the underlying DAO from happening. If this rule is executed after any other rule, the code that I'm removing in this PR will cause the put to happen after the first rule runs and before the validation rule can run.


lchanmann:
LGTM, thanks @TW80000.

PR #2141, opened 2019-05-11
Update tab style

TW80000:
Updates the tab style to match the Liquid designs.

I can leave `foam.u2.Tabs` as is and create a new component in the nanopay repo if we'd rather not lose the existing style of tabs.

Screen Shot 2019-05-11 at 3 13 11 PM

### DIFF
--- a/src/foam/u2/DetailView.js
+++ b/src/foam/u2/DetailView.js
@@ -188,7 +188,7 @@ foam.CLASS({
             add(self.title$).
           end();
 
-        var tabs = foam.u2.Tabs.create().style({ width: '1200px' });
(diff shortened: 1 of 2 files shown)


src/foam/u2/DetailView.js
```diff
@@ -188,7 +188,7 @@ foam.CLASS({
             add(self.title$).
           end();
 
-        var tabs = foam.u2.Tabs.create().style({ width: '1200px' });
+        var tabs = foam.u2.Tabs.create({}, self);
```
TW80000, line 191: Setting the parent is required for CSS macro expansion to work.

jlhughes (approved):

PR #2143, opened 2019-05-12
add two way data binding examples

hchoura:


src/com/foamdev/demos/TwoWayDataBinding/Controller.js
```diff
@@ -0,0 +1,84 @@
+/**
```
kgrgreer, line 1: We already have com.google.foam.demos.sevenguis.TempConv. What is this version demonstrating differently? Just how do do it as a Controller or with a more customized view?

PR #2146, opened 2019-05-14
Add ace editor classes

jlhughes:
Add ace editor js models to classes to enable 'scripts' code syntax highlighting/formatting. 

Also remove spaces and special characters from Enum value names.


jlhughes:
Follow up: The classes entries should not be required.  Adding all to files.js does not help.

mcarcaso:
Pretty sure we don't want those added to classes.js. Shouldn't break anything but will generate some large enums on java.

PR #2147, opened 2019-05-14
#comment added more name attributes

noodlemoodle:


src/foam/u2/view/RichChoiceView.js
```diff
@@ -325,6 +325,7 @@ foam.CLASS({
                     .select(section.filtered || section.dao, (obj) => {
                       return this.E()
                         .start(self.rowView, { data: obj })
+                          .attrs({ name: (obj.name).replace(/[^\w\d]/g,'') })
```
kgrgreer, line 328: Is the the proper use of the 'name' attribute? Why are we doing this?

jlhughes: believe this should be 
`obj.cls_.id`

kgrgreer: Not needed. Troy will set the name and the value on the top-level <div>.
FYI: @TW80000

PR #2149, opened 2019-05-14
Notification updates

TW80000:
## Changes

* Update the way emails are sent based on notifications
  * The 'broadcasted' and 'groupId' properties were not being respected before. Now emails will be sent to everyone if 'broadcasted' is true and to everyone in a particular group if 'groupId' is set.
* Fix the NotificationRowView a bit

### DIFF
--- a/src/foam/nanos/notification/SendEmailNotificationDAO.java
+++ b/src/foam/nanos/notification/SendEmailNotificationDAO.java
@@ -8,25 +8,16 @@ package foam.nanos.notification;
 
 import foam.core.FObject;
 import foam.core.X;
-import foam.dao.ArraySink;
 import foam.dao.DAO;
 import foam.dao.ProxyDAO;
-import foam.mlang.sink.Map;
 import foam.nanos.app.AppConfig;
 import foam.nanos.auth.User;
-import foam.nanos.logger.Logger;
 import foam.nanos.notification.email.EmailMessage;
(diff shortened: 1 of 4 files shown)


jlhughes (approved):

src/foam/nanos/notification/SendEmailNotificationDAO.java
```diff
@@ -28,48 +37,84 @@ public SendEmailNotificationDAO(X x, DAO delegate) {
...
 
-    if ( oldNotif != null )
-      return super.put_(x, obj);
+    // Check if the notification should send emails.
+    if ( oldNotif != null || ! notif.getEmailIsEnabled() ) return super.put_(x, obj);
+
+    // A list of user ids we're going to send emails to.
+    Set ids = new HashSet<Long>();
+
+    // If `broadcasted` is set, add all users in the system to the list.
```
smirnovaka: this logic is already implemented. The bug can be fixed by putting 
`notification.setId(0L);` on SendNotificationDAO.java:51

smirnovaka (changes requested):

src/foam/nanos/notification/SendEmailNotificationDAO.java
TW80000: I see, thanks. I didn't know about `SendNotificationDAO`.

smirnovaka (approved):

PR #2155, opened 2019-05-14
Create Storage interface, and various implementations

nick-prat:
Standardizes the storage interface, and offers File System, Resource, and Fallback implementations


src/foam/flow/DocumentationFolderDAO.js
```diff
@@ -35,28 +35,26 @@ sink = prepareSink(sink);
 foam.dao.Sink         decorated = decorateSink_(sink, skip, limit, order, predicate);
 foam.dao.Subscription sub       = new foam.dao.Subscription();
 
-java.nio.file.FileSystem fs = java.nio.file.FileSystems.getDefault();
-java.nio.file.DirectoryStream<java.nio.file.Path> contents;
+// Load files from both the Jar and regular FileSystem
+java.util.Map<String, java.io.InputStream> iRStreamMap = new foam.nanos.fs.Storage(getDir(), true).getDirectoryAsStream("");
```
adamvy: Do we really need fallback support?  I don't think we'll ever actually be serving from both places at the same time, we either serve from the filesystem or serve from the JAR.  Rather than redo most of the code of this class we only need to be smarter about which FileSystem we start from.

We can crib from how HttpServer does it here: https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/jetty/HttpServer.js#L114

Basically if nanos.webroot is set, then we use that as the resource base.  We can do something like

```
String documentFs = System.getProperty("nanos.document.fs");
if ( documentFs == null ) {
  documentFs = System.getProperty("user.dir");
}
java.nio.file.FileSystem fs = java.nio.file.FileSystem.getFileSystem(new URI(documentFs));
```

and then set nanos.document.fs appropriately when running from a JAR or not.

and then leave the rest as is.  The above code might need some tweaking.

jlhughes: @adamvy But what about edited flow documents - flow documents shipped in the jar will then have a later version outside the jar.

jlhughes (approved):

src/foam/flow/DocumentationFolderDAO.js
adamvy: @jlhughes that's what we should have, we shouldn't be updating the JAR IMO, the jar we ship should be immutable.

src/foam/nanos/fs/Storage.java
```diff
@@ -32,6 +31,59 @@ public Storage (String root, boolean isResource) {
...
+
+    return fs;
+  }
+
+  public Path getPath(String name) {
+    FileSystem fs = getFS();
+    if ( fs == null ) return null;
+
+    Path path;
+    if ( isResource() ) {
```
adamvy: This class is getting really complicated, I think we should split it up into multiple implementations of the same interface.  One implementation for a regular directory and another for reading from resources.

I also don't think we should be returning Path objects, but instead we should abstract the caller away from paths and instead provide methods to get an iStream or to read all bytes of a file.

src/foam/flow/DocumentationFolderDAO.js
```diff
@@ -35,31 +35,18 @@ sink = prepareSink(sink);
...
-try {
-  contents = java.nio.file.Files.newDirectoryStream(fs.getPath(getDir()), "*.flow");
-} catch ( java.io.IOException e ) {
-  throw new RuntimeException(e);
-}
-
-for ( java.nio.file.Path path : contents ) {
-  if ( sub.getDetached() ) break;
-
+java.nio.file.DirectoryStream<java.nio.file.Path> paths = new foam.nanos.fs.ResourceStorage(getDir()).getDirectoryStream("", "*.flow");
```
adamvy: We could make a composite UnionStorage class which composes two Storage classes together.  Then we can just use an instance of UnionStorage for our Storage class in the DocumentationFolderDAO.

UnionStorage would read from both child storages when enumerating the directory and unify them, and it would only write to one of them.  Letting us read fro mthe JAR and the documentation folder, bu tonly write to the documentation folder.

As it stands this won't work when we don't run from a JAR i believe.

src/foam/flow/DocumentationFolderDAO.js
```diff
@@ -82,13 +69,29 @@ if ( ! id.matches("^[a-zA-Z0-9_-]+$") ) {
     {
       name: 'put_',
       javaCode: `
-verifyId(((String)getPK(obj)));
+String id = (String)getPK(obj);
+verifyId(id);
 
-java.nio.file.FileSystem fs = java.nio.file.FileSystems.getDefault();
+java.nio.file.Path path = new foam.nanos.fs.FileSystemStorage(getDir()).getPath(id + ".flow");
+if ( ! java.nio.file.Files.exists(path) ) {
```
adamvy: This could probably live inside the FileSystemStorage class, it could create the parent directory automatically if it doesn't exist, then client code doesn't need to know or care, they just ask for an outputStream as usual.

src/foam/nanos/fs/Storage.java
```diff
@@ -1,50 +1,22 @@
...
+
+  java.io.File get(String name);
+
+  byte[] getBytes(String name);
+
+  OutputStream getOutputStream(String name);
+
+  InputStream getInputStream(String name);
+
+  DirectoryStream<Path> getDirectoryStream(String name);
```
adamvy: I'm wondering if we should expose Paths at all.  Maybe getDirectoryStream should just return a List<String> which is a list of files in the directory.  In fact if you make it Set<String> then it would more or less automatically deduplicate names in the case of a UnionStorage implementation.

src/foam/u2/view/ReciprocalSearch.js
```diff
@@ -113,6 +113,30 @@ foam.CLASS({
           predicate$: this.data$
         });
       }
+    },
```
adamvy: Are these related to this PR?

nick-prat: No, they weren't.

PR #2156, opened 2019-05-14
Action permission fixes

mcarcaso:
- Make the auth checks for enabled/available happen in parallel
- Fix a couple of spots where we were checking if the action has isEnabled/isAvailable to not make that  check because it doesn't account for permissions. Best is to just trust the createIsEnabled$ and createIsAvailable$ methods.
- Make createIsAvailable$ and createIsEnabled$ methods context oriented.
- Restructure the Action code a little so there's less of it.


src/foam/core/Action.js
```diff
@@ -134,59 +134,64 @@ foam.CLASS({
...
-    function andSlotAndPromise(slot, promise) {
-      return this.andSlots(slot, this.PromiseSlot.create({
-        promise: promise
-      }));
-    },
 
-    function checkPermission(x, permission) {
-      if ( ! this.permissionConfig || ! x.auth ) return Promise.resolve(true);
-      return x.auth.check(null, permission);
+    function hasPermissions_(x, data, property) {
```
mcarcaso: I don't need data here. Fixing...

src/foam/core/Action.js
```diff
@@ -134,59 +134,58 @@ foam.CLASS({
...
-        ? foam.Function.withArgs(this.isEnabled, data)
-        : true;
-    },
-
-    function createIsEnabled$(data$) {
-      return this.createSlot_(this.isEnabled, 'enabled', data$);
+    function createPermissionlessCheckFor_(data$, property) {
+      if ( ! this[property] ) return foam.core.ConstantSlot.create({value: true});
+      return foam.core.ExpressionSlot.create({
+        obj$: data$,
```
kgrgreer: I don't think ExpressionSlot supports 'obj' changing, so you could just do obj: data$.get(), which would be more efficient and generate less garbage/listeners.

src/foam/core/Action.js
```diff
@@ -134,59 +134,58 @@ foam.CLASS({
...
-          return a && b;
-        }
-      });
-    },
 
-    function andSlotAndPromise(slot, promise) {
-      return this.andSlots(slot, this.PromiseSlot.create({
-        promise: promise
-      }));
+    function hasPermissions_(x, property) {
```
kgrgreer: Naming the second argument 'property' is a bit confusing because I would then expect it to be a Property. Maybe 'fn' would be a better name.

src/foam/core/Action.js
```diff
@@ -134,59 +134,58 @@ foam.CLASS({
...
-
-    function isEnabledFor(data) {
-      return this.isEnabled
-        ? foam.Function.withArgs(this.isEnabled, data)
-        : true;
-    },
-
-    function createIsEnabled$(data$) {
-      return this.createSlot_(this.isEnabled, 'enabled', data$);
+    function createPermissionlessCheckFor_(data$, property) {
```
kgrgreer: Naming the second argument 'property' is confusing because I expected it would be a Property. Maybe naming something like 'f' or 'fn' would be clearer. And then calling like:
this.createPermissionlessCheckFor_(data$, this.isEnabled)
would be simpler and clearer.

src/foam/core/Action.js
```diff
@@ -134,59 +134,58 @@ foam.CLASS({
...
+    function maybeCall(ctx, data) {
       var data$ = this.ConstantSlot.create({ value: data });
-      if (
-        this.isEnabledFor(data) &&
-        this.isAvailableFor(data) &&
-        await this.hasPermissions_('enabled', data$) &&
-        await this.hasPermissions_('available', data$)
-      ) {
+      Promise.all([
+        Promise.resolve(this.createPermissionlessCheckFor_(data$, 'isEnabled').get()),
```
kgrgreer: Do you need to have Promise.resolve()? Doesn't Promise.all() also accept non-promises?

src/foam/core/Action.js
```diff
@@ -134,59 +134,58 @@ foam.CLASS({
...
     },
 
-    function isAvailableFor(data) {
-      return this.isAvailable
-        ? foam.Function.withArgs(this.isAvailable, data)
-        : true;
+    function createIsEnabled$(x, data$) {
+      return foam.core.ArraySlot.create({
+        slots: [
+          foam.core.PromiseSlot.create({
```
adamvy: Can we avoid creating the PromiseSlot in the case where we _don't_ have permissions or an auth service?  If we leave it as a ConstantSlot then we can save even more listeners from being hooked up in what is currently a common case of no permissions on an action.

src/foam/core/Action.js
```diff
@@ -134,59 +134,58 @@ foam.CLASS({
...
-    async function maybeCall(ctx, data) {
+    function maybeCall(ctx, data) {
       var data$ = this.ConstantSlot.create({ value: data });
-      if (
-        this.isEnabledFor(data) &&
-        this.isAvailableFor(data) &&
-        await this.hasPermissions_('enabled', data$) &&
-        await this.hasPermissions_('available', data$)
-      ) {
+      Promise.all([
```
adamvy: Let's add a comment here noting that because the action is called asynchronously, if it attempts to open any windows with window.open then it will get rejected by the browser's popup blocker.

As you can only do window.open synchronously from an onclick event handler.

It would be extra nice but not as elegant if we avoided the async path in the case where we don't have permissions or we don't have an auth service to check again.

kgrgreer: I agree.

src/foam/core/Action.js
```diff
@@ -113,80 +112,99 @@ foam.CLASS({
...
-    },
+      // Helper method for creating isEnabled/isAvailable slots.
 
-    function isEnabledFor(data) {
-      return this.isEnabled
-        ? foam.Function.withArgs(this.isEnabled, data)
-        : true;
-    },
+      var slot = expression ?
+          foam.core.ExpressionSlot.create({
```
kgrgreer: I think you can just do data.slot(expression).

src/foam/core/stdlib.js
```diff
@@ -704,6 +704,17 @@ foam.LIB({
           a.splice(i, 1);
         }
       }
+    },
+    function unique(a, comparator) {
```
kgrgreer, line 708: Add docs.

src/foam/core/Action.js
adamvy: Done

src/foam/core/stdlib.js
adamvy: done

src/foam/core/Action.js
kgrgreer: It still shows the old code.

PR #2166, opened 2019-05-16
Fix ruleDAO.listen

lchanmann:
Remove enabled=true predicate from ruleDAO because it prevents ruleDAO.listen from updating rules list when a rule is set to enabled=false.


kgrgreer:
Put back. Listen first then add the where().

lchanmann:
We still need where `EQ(Rule.DAO_KEY, getDaoKey())` before listen because it is going to be used by many DAOs.

PR #2169, opened 2019-05-17
ReciprocalSearch Performance Improvements

TW80000:
Should slightly improve the performance of ReciprocalSearch since we don't set up any of the search views until the user checks the box to show it.

Also improved GroupAutocompleteSearchView performance by changing the way the datalist is updated when the DAO changes.

### DIFF
--- a/src/foam/u2/view/TextField.js
+++ b/src/foam/u2/view/TextField.js
@@ -42,56 +42,62 @@ foam.CLASS({
     'type',
     'placeholder',
     'autocompleter',
-    'autocompleteList_'
+    'autocompleteList_',
+    'elm_'
   ],
 
   methods: [
     function initE() {
-      var e = this.start(this.view, {
+      this.elm_ = this.start(this.view, {
         data$:            this.data$,
         label$:           this.label$,
         alwaysFloatLabel: this.alwaysFloatLabel,
         type:             this.type,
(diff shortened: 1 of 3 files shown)


src/foam/u2/view/TextField.js
```diff
@@ -42,56 +42,62 @@ foam.CLASS({
...
         alwaysFloatLabel: this.alwaysFloatLabel,
         type:             this.type,
         onKey:            this.onKey
       });
-      e.attrs({
+      this.elm_.attrs({
         placeholder: this.placeholder$,
       });
-      e.end();
+      this.elm_.end();
```
kgrgreer: This line doesn't do anything. It just returns the parent, which you don't use.

src/foam/u2/view/TextField.js
```diff
@@ -42,56 +42,62 @@ foam.CLASS({
...
-      var e = this.start(this.view, {
+      this.elm_ = this.start(this.view, {
         data$:            this.data$,
         label$:           this.label$,
         alwaysFloatLabel: this.alwaysFloatLabel,
         type:             this.type,
         onKey:            this.onKey
       });
-      e.attrs({
+      this.elm_.attrs({
```
kgrgreer: Why not just chain this onto the previous line like normal?

PR #2170, opened 2019-05-17
update grp.getAppConfig()

Nauna:
grp.getAppConfig() previously picked up the group from the context and set the properties of the context group onto a copy of appConfig and returned.

This was a problem for when the logged in user was someone from paymentOps and was triggering an email to an ablii user.

The solution/changes include:
1) using the ```Group group = this;``` instead of ```Group group = x.get("group");```
2) If properties are not set on the current group, iterate through to root(getParent()) and set properties. Taking president with the first property settings found. With ablii users, the properties are set on sme.

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -211,39 +211,39 @@ foam.CLASS({
         }
       ],
       javaCode: `
-AppConfig config = (AppConfig) ((AppConfig) x.get("appConfig")).fclone();
+        AppConfig config = (AppConfig) ((AppConfig) x.get("appConfig")).fclone();
+        String configUrl = "";
 
-String configUrl = config.getUrl();
+        // FIND URL
+        Group group = (Group) x.get("group");
+        if ( ! SafetyUtil.isEmpty(group.getUrl()) ) {
+          configUrl = group.getUrl();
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/Group.js
```diff
@@ -211,39 +211,57 @@ foam.CLASS({
...
+        DAO groupDAO              = (DAO) x.get("groupDAO");
 
-String configUrl = config.getUrl();
+        while ( group != null && ! (urlFound && supportEmailFound)) {
+          configUrl          = urlFound ? configUrl : group.getUrl();
+          configSupportEmail = supportEmailFound ? configSupportEmail : group.getSupportEmail();
+      
+          // Once true, stay true
+          urlFound          = urlFound   ? urlFound   : ! SafetyUtil.isEmpty(configUrl);
+          supportEmailFound = supportEmailFound ? supportEmailFound : ! SafetyUtil.isEmpty(configSupportEmail);
```
jlhughes, line 230: Little optimization - to break when both have been found.  Otherwise looks good.

Nauna: good catch thx 👍

jlhughes (approved):

PR #2183, opened 2019-05-18
Add validationPredicates to Property

mcarcaso:
This allows us to create cross-platform validation. See Phone.js and Address.js for examples.

### DIFF
--- a/src/foam/core/Validation.js
+++ b/src/foam/core/Validation.js
@@ -118,17 +118,13 @@ foam.CLASS({
       of: 'foam.core.ValidationPredicate',
       name: 'validationPredicates',
       factory: function() {
+        var self = this;
         var a = [];
         if ( foam.Number.isInstance(this.min) ) {
           a.push({
             args: [this.name],
-            predicate: {
-              class: 'foam.mlang.predicate.RegExp',
-              arg1: {
-                class: 'foam.mlang.FObjectPropertyExpr',
-                property: this.name
-              },
-              regExp: new RegExp('^.{'+this.min+',}$')
+            predicateFactory: function(e) {
+              return e.REG_EXP(self, new RegExp('^.{'+this.min+',}$'));
             },
             errorString: `${this.label} must be at least ${this.min} character${this.min>1?'s':''}`
           });
@@ -136,13 +132,8 @@ foam.CLASS({
         if ( foam.Number.isInstance(this.max) ) {
           a.push({
             args: [this.name],
-            predicate: {
-              class: 'foam.mlang.predicate.RegExp',
-              arg1: {
-                class: 'foam.mlang.FObjectPropertyExpr',
-                property: this.name
-              },
-              regExp: new RegExp('^.{0,'+this.max+'}$')
+            predicateFactory: function(e) {
+              return e.REG_EXP(self, new RegExp('^.{0,'+this.max+'}$'));
             },
             errorString: `${this.label} must be at most ${this.max} character${this.max>1?'s':''}`
           });
--- a/src/foam/core/Validation.js
+++ b/src/foam/core/Validation.js
@@ -17,13 +17,154 @@
 
 foam.CLASS({
   package: 'foam.core',
+  name: 'ValidationPredicate',
+  properties: [
+    {
+      class: 'FObjectProperty',
+      of: 'foam.mlang.predicate.Predicate',
+      name: 'predicate'
+    },
+    {
+      class: 'StringArray',
+      name: 'args'
+    },
+    {
+      class: 'Function',
+      name: 'jsFunc',
+      expression: function(predicate, jsErr) {
+        return function() {
+          if ( ! predicate.f(this) ) return jsErr(this);
+        };
+      }
+    },
+    {
+      class: 'String',
+      name: 'errorString'
+    },
+    {
+      class: 'Function',
+      name: 'jsErr',
+      expression: function(errorString) {
+        return function() { return errorString; };
+      }
+    }
+  ],
+  methods: [
+    function createErrorSlotFor(data) {
+      return this.ExpressionSlot.create({
+        args: this.args.map(a => data[a+'$']),
+        code: this.jsFunc.bind(data)
+      });
+    }
+  ]
+});
+
+foam.CLASS({
+  package: 'foam.core',
   name: 'PropertyValidationRefinement',
   refines: 'foam.core.Property',
 
   properties: [
     {
+      class: 'FObjectArray',
+      of: 'foam.core.ValidationPredicate',
+      name: 'validationPredicates'
+    },
+    {
       name: 'validateObj',
-      expression: function(name, label, required) {
+      expression: function(name, label, required, validationPredicates) {
+        if ( validationPredicates.length ) {
+          var args = Object.keys(validationPredicates
+            .map(vp => vp.args)
+            .flat()
+            .reduce((map, a) => {
+              map[a] = true;
+              return map;
+            }, {}));
+          return [args, function() {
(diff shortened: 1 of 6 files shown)


src/foam/core/Validation.js
```diff
@@ -15,15 +15,156 @@
...
+    },
     {
       name: 'validateObj',
-      expression: function(name, label, required) {
+      expression: function(name, label, required, validationPredicates) {
+        if ( validationPredicates.length ) {
+          var args = Object.keys(validationPredicates
+            .map(vp => vp.args)
+            .flat()
+            .reduce((map, a) => {
```
adamvy: You can use foam.Array.unique() insteat of the trick with Object.keys() and .reduce().  Is a bit more readable.

src/foam/core/Validation.js
```diff
@@ -15,15 +15,156 @@
...
+    },
+  ]
+});
+
+foam.CLASS({
+  package: 'foam.core',
+  name: 'StringPropertyValidationRefinement',
+  refines: 'foam.core.String',
+  properties: [
+    'min',
```
adamvy: Would minLength/maxLength be more readable?

mcarcaso: Think so :)

src/foam/core/Validation.js
mcarcaso: Sweet! Done.

src/foam/nanos/auth/Address.js
```diff
@@ -149,42 +139,31 @@ foam.CLASS({
...
-        var streetNameRegex = /^[a-zA-Z0-9 ]{1,70}$/;
-        if ( ! streetNameRegex.test(streetName) ) {
-          return 'Invalid street name.';
+      validationPredicates: [
+        {
+          args: ['structured', 'streetName'],
+          predicateFactory: function(e) {
+            return e.OR(
+              e.EQ(foam.nanos.auth.Address.STRUCTURED, false),
+              e.REG_EXP(foam.nanos.auth.Address.STREET_NAME, /^\s*.+\s*$/)
```
kgrgreer, line 148: I would rather have the preSet() do a trim(), or add a trim: true property to String properties and then just use EQ here. REG_EXP is big and slow for this and doesn't clean up the string.

kgrgreer: Actually, we already have a trim: property on Strings.

mcarcaso: We don't want active trimming on the value because then you can't add spaces while filling it out in an input.

src/foam/core/Validation.js
```diff
@@ -15,15 +15,151 @@
...
+      of: 'foam.core.ValidationPredicate',
+      name: 'validationPredicates',
+      factory: function() {
+        var self = this;
+        var a = [];
+        if ( foam.Number.isInstance(this.minLength) ) {
+          a.push({
+            args: [this.name],
+            predicateFactory: function(e) {
+              return e.REG_EXP(self, new RegExp('^.{'+this.minLength+',}$'));
```
adamvy: A RegExp for a minimum length feels like overkill.

src/foam/java/refinements.js
```diff
@@ -212,11 +214,28 @@ foam.CLASS({
...
     {
       class: 'String',
-      name: 'javaValidateObj'
+      name: 'javaValidateObj',
+      expression: function(validationPredicates) {
+        return validationPredicates
+          .map(vp => {
+            return `
+if ( ! ${foam.java.asJavaValue(vp.predicate)}.f(obj) ) {
+  throw new IllegalStateException(${foam.java.asJavaValue(vp.errorString)});
```
adamvy, line 223: I think we need to model a specific ValidationException that the JS side can catch and handle specially.

src/foam/nanos/auth/Address.js
```diff
@@ -51,31 +51,20 @@ foam.CLASS({
...
       displayWidth: 50,
       documentation: 'An unstructured field for the main postal address.',
-      validateObj: function(address1) {
-        var address1Regex = /^[a-zA-Z0-9 ]{1,70}$/;
-
-        if ( address1.length > 0 && ! address1Regex.test(address1) ) {
-          return 'Invalid address line.';
-        }
-      }
+      expression: function(structured, streetNumber, streetName) {
```
adamvy, line 57: Are we truly using both structured and unstructured addresses in the system?

src/foam/nanos/auth/Address.js
```diff
@@ -51,31 +51,20 @@ foam.CLASS({
...
       displayWidth: 50,
-      documentation: 'An unstructured field for the sub postal address.',
-      validateObj: function(address2) {
-        var address2Regex = /^[a-zA-Z0-9 ]{1,70}$/;
-
-        if ( address2.length > 0 && ! address2Regex.test(address2) ) {
-          return 'Invalid address line.';
-        }
-      }
+      documentation: 'An unstructured field for the sub postal address.'
```
adamvy, line 67: "sub postal address" is meaningless to me.

src/foam/core/Validation.js
mcarcaso: Can't disagree with that :)
What do you suggest instead? Do we have an mlang expression for length of a string or something? If not, should I create one or did you have something else in mind?

src/foam/java/refinements.js
mcarcaso: Agreed. We must do something already because I've seen some server side errors bubble up to the client but I'm not entirely sure how that's happening.

src/foam/core/Validation.js
adamvy: mlang expression for string length would be useful I think

src/foam/java/refinements.js
adamvy: WizardView in particular may want to handle ValidationExpression well, as it could jump to the appropriate section if it wanted.

src/foam/nanos/auth/Address.js
mcarcaso: Don't know off the top of my head and I'd love to remove this stuff but I'm afraid of non-backwards compatible changes right now.

src/foam/nanos/auth/Address.js
mcarcaso: Me too. Once this crunch is over I want to remove a ton of stuff from this model.

src/foam/core/Validation.js
mcarcaso: On it

adamvy:
We still need object level, cross property validation so you can do things like: I need properties A B C to be valid, or properties D and E.

mcarcaso:
> We still need object level, cross property validation so you can do things like: I need properties A B C to be valid, or properties D and E.

Good idea. Any idea how you'd expect to declare that on a property?

adamvy:
I don't think you can.  You have to declare it at a Model level.  The Model has its own requirements for validation that extends beyond the Properties.

This would typically be done on during the validate() method on the object.

adamvy:
Maybe we can get by without it for now.

src/foam/core/Validation.js
mcarcaso: Done

src/foam/nanos/auth/Address.js
adamvy: Yeah me too, I hear you.  Backwards compatability is what has me nervous about adding this expression, but I think this is probably the best idea at the moment.

src/foam/nanos/auth/Address.js
adamvy: I meant that the documentation string could maybe be updated, but that's minor lol.

PR #2187, opened 2019-05-20
Remove the validation on HumanNameTrait because it's not great valida…

mcarcaso:
…tion and it gets in the way of any validation that a implementor of this interface might want.


jlhughes (approved):
The length limit came from iso 20022, but should be implemented/restricted elsewhere.
The restriction of numbers in names is just plain odd. 
And in anycase, this should not be on the foam side.

PR #2193, opened 2019-05-21
Create a ChoiceWithOtherView which takes a choice view and an otherKe…

mcarcaso:
…y and shows another input if otherKey is selected.

image

image

### DIFF
--- a/src/foam/u2/view/ChoiceWithOtherView.js
+++ b/src/foam/u2/view/ChoiceWithOtherView.js
@@ -1,0 +1,98 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'ChoiceWithOtherView',
+  extends: 'foam.u2.View',
+  requires: [
+    'foam.u2.layout.Rows'
+  ],
+  properties: [
+    {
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/ChoiceWithOtherView.js
```diff
@@ -0,0 +1,98 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
```
adamvy, line 7: In the future I think we can just inline this into the ChoiceView as an optional feature.

TW80000:
This only really works for string type properties, right? I don't think that's an issue, just wanted to point it out since the regular choice view can be used for more than just strings.

mcarcaso:
> This only really works for string type properties, right? I don't think that's an issue, just wanted to point it out since the regular choice view can be used for more than just strings.

I don't think so actually. You should be able to set `otherKey` to any kind of object and same goes for the `otherDefault` but if I'm being completely honest I only really tested for Strings :p

TW80000:
> > This only really works for string type properties, right? I don't think that's an issue, just wanted to point it out since the regular choice view can be used for more than just strings.
> 
> I don't think so actually. You should be able to set `otherKey` to any kind of object and same goes for the `otherDefault` but if I'm being completely honest I only really tested for Strings :p

Yeah, never mind, I should have given it more thought. I see now that you can set the `otherView` as well.

PR #2201, opened 2019-05-22
Cpf 1421/tooltip view

noodlemoodle:


src/foam/u2/view/Tooltip.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  ], 
+
+  listeners: [
+    {
+      name: 'loadTooltip',
+      isFramed: true,
+      code: function() {
+        var self = this;
+        setTimeout(function(){
```
kgrgreer, line 83: Timeout is risky. Better to use an on load.

src/foam/u2/view/Tooltip.js
```diff
@@ -0,0 +1,111 @@
...
+          if ( ! self.el() ) return;
+          self.domRect = self.el().getBoundingClientRect(); 
+          self.screenWidth = self.window.innerWidth;
+          self.screenHeight = self.window.innerHeight;
+          self.scrollY = self.window.scrollY;
+          var above = self.domRect.top - scrollY > self.screenHeight / 2;
+          var left = self.domRect.left > self.screenWidth / 2;
+          self.start()
```
kgrgreer, line 91: Better to add tooltip to document than to this element itself since that can mess up sizing. Then best to remove when done so if hover again after scrolling proper position can be computed again.

PR #2202, opened 2019-05-22
CPF-1595 #comment fixed to show correct placeholder

noodlemoodle:


mcarcaso (commented):
I don't think this is where we want the fix to happen. Let's take this offline.

PR #2204, opened 2019-05-22
Differentiate an unselected value vs other value.

mcarcaso:


src/foam/u2/view/ChoiceWithOtherView.js
```diff
@@ -22,6 +22,15 @@ foam.CLASS({
       name: 'otherKey',
       required: true
     },
+    {
+      documentation: `
```
adamvy, line 28: Fix after launch.

This is a weird implementation detail to expose to the user.  We should be maintaining the choices of the choice view to contain an "other" choice regardless of whether the choices come from a DAO or a choices property.

PR #2211, opened 2019-05-24
A few mlang Java implementations

TW80000:
- Log the partially evaluated predicate since that's what gets used
- Add Java implementation for `partialEval` for `In` mlang
  - Fixes unoptimized queries on MDAO when we're selecting from a ManyToManyRelationshipDAO since it uses `In` on the server side. Without the Java implementation of `partialEval`, if no junctions were found, then it was selecting with a query like `id in []`, which was causing the "unoptimized query" warning.
  - I think the query parser on the client is doing this optimization for us, but since I was using a ManyToManyRelationshipDAO on the backend for the feature I'm working on, it was never going through the query parser.
- Add Java implementation for `toString` for `ArrayConstant`
  - This was just to make debugging easier since " ? " isn't very helpful, which is the current value shown for `Constant` and `ArrayConstant`.

### DIFF
--- a/src/foam/dao/MDAO.java
+++ b/src/foam/dao/MDAO.java
@@ -159,7 +159,7 @@ public class MDAO
     // TODO: if plan cost is >= size, log a warning
     if ( state != null && predicate != null && plan.cost() > 1000 && plan.cost() >= index_.size(state) ) {
       Logger logger = (Logger) x.get("logger");
-      logger.error(predicate.createStatement(), " Unindexed search on MDAO");
+      logger.error(simplePredicate.createStatement(), " Unindexed search on MDAO");
     }
 
     plan.select(state, sink, skip, limit, order, simplePredicate);
--- a/src/foam/dao/MDAO.java
+++ b/src/foam/dao/MDAO.java
@@ -156,10 +156,10 @@ public class MDAO
       plan = index_.planSelect(state, sink, skip, limit, order, simplePredicate);
     }
 
-    // TODO: if plan cost is >= size, log a warning
     if ( state != null && predicate != null && plan.cost() > 1000 && plan.cost() >= index_.size(state) ) {
       Logger logger = (Logger) x.get("logger");
-      logger.error(simplePredicate.createStatement(), " Unindexed search on MDAO");
+      if ( ! predicate.equals(simplePredicate) ) logger.debug(String.format("The original predicate was %s but it was simplified to %s.", predicate.toString(), simplePredicate.toString()));
+      logger.warning("Unindexed search on MDAO", simplePredicate.toString());
     }
 
     plan.select(state, sink, skip, limit, order, simplePredicate);
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -703,11 +703,16 @@ foam.CLASS({
     function toIndex(tail) {
       return this.arg1 && this.arg1.toIndex(tail);
     },
-
-    function toString() {
-      return foam.String.constantize(this.cls_.name) + '(' +
-          this.arg1.toString() + ', ' +
-          this.arg2.toString() + ')';
+    {
+      name: 'toString',
+      code: function() {
+        return foam.String.constantize(this.cls_.name) + '(' +
+            this.arg1.toString() + ', ' +
+            this.arg2.toString() + ')';
+      },
+      javaCode: `
+        return String.format("%s(%s, %s)", getClass().getSimpleName(), getArg1().toString(), getArg2().toString());
+      `
     },
     {
       name: 'prepareStatement',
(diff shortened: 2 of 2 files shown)


src/foam/dao/MDAO.java
```diff
@@ -159,7 +159,7 @@ public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Pre
     // TODO: if plan cost is >= size, log a warning
     if ( state != null && predicate != null && plan.cost() > 1000 && plan.cost() >= index_.size(state) ) {
       Logger logger = (Logger) x.get("logger");
-      logger.error(predicate.createStatement(), " Unindexed search on MDAO");
```
adamvy, line 162: I'm trying to think if there might be value to log the predicate that the user actually requested along with the one that we partialEval'd to.  Might be useful to log both.

src/foam/mlang/mlang.js
```diff
@@ -1449,6 +1449,12 @@ foam.CLASS({
 
   requires: [ 'foam.mlang.Constant' ],
 
+  javaImports: [
```
adamvy, line 1457: Personally I much prefer writing fully qualified names in the java code for three reasons.

1) If we're writing so much java code that using javaimports saves a lot of time, then we might be writing too much code.

2) If we use fully qualified names in the javaCode itself, then its easy to cut and paste that code if we need to move it around, as the code is more self contained.

3) javaImports quickly gets out of date, if we remove some code its extra mental overhead to read the rest of the class and see if we are still using all the imports.  As a result the list tends to grow and never shrink.

But like I said, this is just a preference not (yet) a rule.  So fix it if you agree, or leave it if you don't agree.  At a later date we could debate and decide which one is truly better :)

src/foam/mlang/mlang.js
```diff
@@ -1621,7 +1645,7 @@ foam.CLASS({
     },
     {
       name: 'createStatement',
-      javaCode: 'return " ? "; '
+      javaCode: 'return getValue().toString(); '
```
adamvy: I think this needs to be getValue().toSQL or something.  toString will not escape it properly for SQL.

adamvy: Do we need to fix createStatement for this PR?  I'd rather leave that for a separate job, we need to be very careful that we don't introduce SQL injection vulnerabilities.  And I think that's best avoided by not repurposing toString() when we mean toSQL().

TW80000: Oh, is `createStatement` something to do with SQL? I didn't realize that. It's what's being used by MDAO when logging the unoptimized query warning. Should it just be using `toString` there instead of `createStatement`?

src/foam/mlang/mlang.js
TW80000: 1. I don't understand what you mean by this
2. Fair enough
3. Fair enough

The reason I prefer to not use the full name is because I find it makes the code harder to read. Personally, to me it's visual noise, especially in an already verbose language like Java. The package name just adds superfluous text that the reader has to scan (and almost always ignore). The noise is especially annoying in nested code like mlangs where everything is prefixed by `foam.mlang.MLang.`. The only time I personally opt to add the full path is when there's a naming conflict.

I'd argue that code is read far more often than it's modified, and that's why I'd argue that it's more valuable to optimize for readability than optimizing for copy-and-pasting and modification.

src/foam/mlang/mlang.js
TW80000: Reverted changes to `createStatement` and used `toString` instead.

src/foam/dao/MDAO.java
TW80000: Added one in 3b1b1cf

src/foam/mlang/mlang.js
adamvy: 1 is kind of a pre-answer to a common argument.  Adding javaImports saves us typing because we don't have to type the fully qualified name every time we refer to something in the method body.  So many people would argue that its worth having becuase it saves typing.  My counter-argument to this point, is that if we're writing so much javaCode by hand that having javaImports is saving us significant amounts of effort, then that is actually a symptom of a deeper problem.  Namely that we're probably missing the chance to model something at a higher level and we should address _THAT_ issue rather than trying to optimize the rate at which we write java code.

But that's not the argument you're making so its less relevant.

Readability is an interesting point.  I agree that code is read more often than it is written.  I think I feel that the code is just as readable if not more readable with fully qualified names.  Often I care if its the Jetty HttpServer or the nanos one.  Or its the foam.nanos.auth.User vs the net.nanopay.auth.User, having the fully qualified name there means that I don't have to refer to a different part of the file to determine which one being referenced.

But I agree that in some cases its just noise that you have to skip over.  I'll have to mull on this further, since I'm now not sure which I prefer.  Interesting discussion, let's move ahead with the code as is.

src/foam/mlang/mlang.js
adamvy: Yeah createStatement comes from the foam.dao.SQLStatement interface.  I believe someone was working on PostgreSQL integration, not sure if its finished or not but that's what it's intended for.

adamvy (approved):

PR #2212, opened 2019-05-26
Set tableWidths, and disable edit and create

jlhughes:
CPF-882 #time 30m


TW80000 (approved):
Didn't test but code LGTM.

PR #2214, opened 2019-05-27
Cpf 129/operation measures

nanoNeel:


jlhughes (approved):

src/foam/nanos/om/DAOOMLogger.java
```diff
@@ -0,0 +1,74 @@
...
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.nanos.om;
+
+import foam.core.ContextAwareSupport;
+import foam.dao.DAO;
+
+public class DAOOMLogger
```
mcarcaso: Why not modelled?

src/foam/nanos/om/DAOOMLogger.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  protected Object getLock(OMInfo info) {
+    int hash = info.getClsName().hashCode() * 31 + info.getName().hashCode();
+    return locks_[(int) (Math.abs(hash) % locks_.length)];
+  }
+
+  @Override
+  public void log(OM om) {
+    if ( ! om.getClassType().getId().equals("foam.dao.OMDAO") ) {
```
mcarcaso: What's going on here? Why do you care if OM is in the class ID?

src/foam/nanos/om/OM.js
```diff
@@ -0,0 +1,65 @@
...
+      javaCode: `
+    if ( x == null ) return;
+    OMLogger logger = (OMLogger) x.get(DAOOMLogger.SERVICE_NAME);
+    if ( logger != null ) logger.log(this);
+`
+    },
+    {
+      name: 'doFolds',
+      javaCode: `
+    fm.foldForState(getClassType().getId()+":"+getName(), getCreated(), 0);
```
mcarcaso: Why 0 and not 1? I guess it doesn't make much difference either way because we're simply counting here but I feel like 1 makes a tad more sense. Either way, you might want to add a comment explaining why you're logging a constant value.

src/foam/nanos/om/OMDAO.java
```diff
@@ -0,0 +1,80 @@
...
+package foam.nanos.om;
+
+import foam.core.FObject;
+import foam.core.X;
+import foam.dao.DAO;
+import foam.dao.ProxyDAO;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+
+public class OMDAO
```
mcarcaso: Model this.

src/foam/nanos/om/OMDAO.java
```diff
@@ -0,0 +1,80 @@
...
+  void init() {
+    putName_       = getOf().getId() + ":put";
+    findName_      = getOf().getId() + ":find";
+    removeName_    = getOf().getId() + ":remove";
+    removeAllName_ = getOf().getId() + ":removeAll";
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    OM om = new OM.Builder(x).setClassType(OMDAO.getOwnClassInfo()).setName(putName_).build();
```
mcarcaso: How many other services are we going to want to add OMs to? If we plan on doing OMs for more than just DAOs, we should generate these classes similar to how we generate proxies.

src/foam/nanos/om/OMInfo.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.om',
+  name: 'OMInfo',
```
mcarcaso: What's the point of this class? Isn't this about the same as the Candlestick?

src/foam/nanos/om/services.jrl
```diff
@@ -0,0 +1,12 @@
+p({"class":"foam.nanos.boot.NSpec", "name":"omLogger",                          "lazy":true,  "serve":false, "serviceClass":"foam.nanos.om.DAOOMLogger"})
+p({"class":"foam.nanos.boot.NSpec", "name":"omInfoDAO",                         "lazy":true,  "serve":true,  "serviceScript":"return new foam.dao.ProxyDAO.Builder(x).setDelegate(new foam.dao.MDAO(foam.nanos.om.OMInfo.getOwnClassInfo())).build();","client":"{\"of\":\"foam.nanos.om.OMInfo\"}" })
+p({"class":"foam.nanos.boot.NSpec", "name":"omDAO",                             "lazy":true,  "serve":true,  "serviceScript":"return new foam.nanos.analytics.FoldManagerDAODecorator.Builder(x).setDelegate(new foam.dao.MDAO(foam.nanos.om.OM.getOwnClassInfo())).setFm((foam.nanos.analytics.DAOFoldManager) x.get(\"omFoldManager\")).build();","client":"{\"of\":\"foam.nanos.om.OM\", \"remoteListenerSupport\": true}" })
```
mcarcaso, line 1: Do we actually care to have a dao of OMs or do we only care about the running totals? If we don't plan on actually storing these, we can probably ditch the OM model completely and make a decorator that goes straight into the FoldManager to reduce a little bit of garbage we'd be creating.

mcarcaso (commented):
I think we can reduce a bunch of the code here by creating:

An 'OMLoggerProperty'. Creating it will look something like this:
```
{
  class: 'OMLoggerProperty',
  name: 'omLogger',
  methods: ['put_', 'select_', 'find_' ] // Not providing this defaults to all methods
}
```
Using this, we can implement an OMLoggerDAO like this:
```
foam.CLASS({
  name: 'OMLoggerDAO',
  extends: 'foam.dao.ProxyDAO',
  properties: [
    {
      class: 'OMLoggerProperty',
      name: 'omLogger'
    }
  ]
});
```
(note you probably don't need that part immediately but could reduce the amount of code you actually end up writing)


I'd change the OMLogger interface to:
```
foam.INTERFACE({
  package: 'foam.nanos.om',
  name: 'OMLogger',
   methods: [
    {
      name: 'log',
      args: [
        {
          name: 'om',
          type: 'String'
        }
      ]
    }
  ]
});
```
A method for OMDAO would look something like this:
```
    try {
      return super.find_(x, id);
    } finally {
      getOmLogger().log("find_");
    }
```
Then created a FoldManagerOMLogger that looks something like:
```
foam.CLASS({
  name: 'FoldManagerOMLogger',
  implements: [
    'foam.nanos.om.OMLogger'
  ],
  properties: [
    {
      class: 'FObjectProperty',
      of: 'foam.nanos.analytics.FoldManager',
      name: 'fm'
    }
  ],
  methods: [
    {
      name: 'log',
      javaCode: `
getFm().foldForState(om, new java.util.Date(), 1);
      `
    }
  ]
});
```
You'll probably also want a PrefixOMLogger that's something like:
```
foam.CLASS({
  name: 'PrefixOMLogger',
  extends: 'foam.nanos.om.ProxyOMLogger',
  properties: [
    {
      class: 'String',
      name: 'prefix'
    }
  ],
  methods: [
    {
      name: 'log',
      javaCode: `
getDelegate().log(prefix + om);
      `
    }
  ]
});
```

Finally, putting it all together on a transactionDAO would look something like this:
```
OMLogger oml = new PrefixOMLogger.Builder(x)
  .setPrefix("TransactionDAO")
  .setDelegate(new FoldManagerOMLogger.Builder(x)
    .setFm(x.get("foldManager"))
    .build());

return new OMLoggerDAO.Builder(x)
  .setOmLogger(oml)
  .setDelegate(x.get("transactionDAO"));
```

With this approach, there's no need for OM, OMInfo, DAOOMLoggerCopySink, or the DAOOMLogger and I don't think it's any less flexible than the current solution but there's a chance I'm missing something that your solution provides that mine doesn't so let me know if I'm mistaken in my thinking.

jlhughes:
Completely refactor to just the one OM.js model, OMDAO and the three 1 minute, 1 hour, 1 day, ... candlestick configuration

src/foam/nanos/om/OMDAO.java
jlhughes: removed

src/foam/nanos/om/OMDAO.java
jlhughes: class removed.

src/foam/nanos/om/DAOOMLogger.java
jlhughes: removed.

src/foam/nanos/om/DAOOMLogger.java
jlhughes: removed

src/foam/nanos/om/OMInfo.js
jlhughes: removed.

src/foam/nanos/om/services.jrl
jlhughes: OM model is now just convenience for putting to candlestick.

mcarcaso:
I'm a little confused here and feel like maybe something is missing. What's actually creating the OMs and calling `log` on them?

jlhughes:
@mcarcaso 
// wherever you want an OM
new OM(name).log(x); 
// Similar to:
pm = new PM(..); pm.log(x);

PR #2216, opened 2019-05-27
Update the notify method to pass in the DOM element

Henry0422:
Allow passing in DOM element into the toast notification.


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -357,8 +357,15 @@ foam.CLASS({
     },
 
     // This method is for toast notification message
-    function notify(message, type) {
-      this.add(this.NotificationMessage.create({ message, type }));
+    function notify(contentElement, type) {
```
mcarcaso: Rename this to `content` because it's not always an actual element and that name could be a tad misleading.

mcarcaso: You could even restructure/shorten this method if you do this:
```
var args = { type };
args[foam.String.isInstance(content) ? 'message' : 'contentElement'] = content;
this.add(this.NotificationMessage.create(args));
```
But I'm thinking the NotificationMessage should just have a single arg and figure out how to display what you give it.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -357,8 +357,15 @@ foam.CLASS({
...
+      if ( foam.String.isInstance(contentElement) ) {
+        this.add(this.NotificationMessage.create({
+          message: contentElement,
+          type: type
+        }));
+      } else {
+        this.add(this.NotificationMessage.create({ contentElement, type }));
```
mcarcaso: Doesn't NotificationMessage need to be updated too?

kgrgreer (approved):

PR #2221, opened 2019-05-28
Small fix SectionedDetailView

lchanmann:
Clear SectionedDetailView "of" property so that it respects data.cls_ after "data" property is changed.


lchanmann:
Thanks @mcarcaso and @tharmaman for helping.

tharmaman (approved):

PR #2223, opened 2019-05-29
add HTMLView element to AllViews

hchoura:


src/com/google/foam/demos/u2/AllViews.html
```diff
@@ -4,6 +4,8 @@
     <script language="javascript" src="../../../../../foam/nanos/nanos.js"></script>
     <script language="javascript" src="AllViews.js"></script>
     <script language="javascript" src="../../../../../foam/u2/md/CheckBox.js"></script>
+    <script language="javascript" src="../../../../../foam/u2/HTMLView.js"></script>
```
kgrgreer, line 7: Couldn't we just add this to files.js instead?

PR #2225, opened 2019-05-30
Move most of the look and feel properties off of Group and onto a new LookAndFeel model

TW80000:
https://nanopay.atlassian.net/browse/CPF-934

## Changes
* Moved most of the look and feel properties off of Group and onto a new LookAndFeel model
* Renamed `groupCSS` to `customCSS`
* Added a default FOAM LookAndFeel

## Description
* Added a `lookAndFeelDAO` and `lookAndFeelService`, which has one method that returns the most appropriate `LookAndFeel` for you
* When you hit the `ApplicationController` it pulls a `LookAndFeel` based on the app name before you log in, then it pulls again in case there's a more specific one you can be using after you sign in


src/foam/nanos/auth/LookAndFeelService.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.INTERFACE({
+  package: 'foam.nanos.auth',
+  name: 'LookAndFeelService',
```
adamvy: Does this need to be its own service or could we just talk to a LookAndFeelDAO?

TW80000: I couldn't think of a nice way for it would work if you were accessing the DAO directly. Returning it from `find` seems weird since you're not passing in an id. Returning from `select` could work I guess if you left it up to the caller to only take the first object.

src/foam/nanos/auth/LookAndFeel.js
```diff
@@ -0,0 +1,114 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'LookAndFeel',
```
adamvy: Personally this feels more like a Theme than LookAndFeel.  I also don't think it should live the auth package, its only loosely auth related.

src/foam/nanos/auth/LookAndFeelService.js
adamvy: My gut instinct is telling me that the service adds overhead to try to defer the logic on which look and feel object to the server.

If the client could just query the DAO and select the theme they want, that would also open the door to features like allowing users to select and switch between available themes if we ever wanted that support without having to add any additional logic to the service.

Basically it seems like less work to just access the DAO from the client, which also leaves options open for more flexibility later at not added cost.

TW80000: I agree. I'll remove the service and update ApplicationController to hit the DAO directly.

src/foam/nanos/auth/LookAndFeel.js
TW80000: I agree with both points.

TW80000: Resolved by 4fdb80c.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -406,6 +388,36 @@ foam.CLASS({
...
+      } catch (err) {
+        this.notify(this.LOOK_AND_FEEL_NOT_FOUND, 'error');
+        console.error(err);
+        return;
+      }
+
+      this.useCustomElements();
+    },
+
+    // Use custom elements if supplied by the Theme.
```
kgrgreer: If you put comments inside the function then the FOAM doc tool can pick them up. Ex.:function useCustomElements() {
  /* Use custom elements if supplied by the Theme. */

src/foam/nanos/theme/ClientThemeService.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.theme',
+  name: 'ClientThemeService',
```
kgrgreer: Why is this needed rather than just letting FOAM generate the client along with all of the other services?

TW80000: I'm removing the service anyway, but I'll keep that in mind for the future.

src/foam/nanos/auth/LookAndFeelService.js
TW80000: Resolved by 25a838ff26f04893188bebf09972c7703bf5177e.

src/foam/nanos/controller/ApplicationController.js
TW80000: Does it pick up `// ...` style comments as well?

TW80000: Nvm, looks like it does.

TW80000: Resolved by 439b82623.

adamvy (approved):

PR #2231, opened 2019-05-31
CPF-1697 Add foam.nanos.app.EmailConfig, causing nspec table view to fail.

jlhughes:


src/io/c9/ace/Lib.js
```diff
@@ -11,7 +11,7 @@ foam.LIB({
       name: 'ACE',
       code: foam.Function.memoize0(function() {
         return new Promise(function(resolve, reject) {
-          var url = global.FOAM_ROOT + '../node_modules/ace-builds/src-min-noconflict/ace.js';
+          var url = (global.FOAM_ROOT && global.FOAM_ROOT + '../' || '')  + 'node_modules/ace-builds/src-min-noconflict/ace.js';
```
adamvy, line 14: Let's leave a TODO to fix this up :)  We shouldn't have to do fancy conditionals like this, I think we need a cleaner story for how FOAM loads external modules, but we can do that later.

adamvy: And a ticket as you just said in slack

PR #2232, opened 2019-06-03
Permission refactor before migration

TW80000:
## Changes

* Refactored permissions and groups so that there's a many-to-many relationship between permissions and groups

## Notes

This PR represents the first half of the changes. In this PR we define the relationship and create the junction DAO (named `groupPermissionJunctionDAO`). We also create a migration script that loops over every permission in every group and adds junctions for each combination to the junctionDAO.

The other half of the changes will be coming in another PR that should only be merged after the migration script has run. In that PR, we remove the `permissions` property on `Group` (which is an `FObjectArray` of `Permission`s) and rename the property defined on `Group` by the relationship from `permissions2` to `permissions`. Then we update all places in the codebase where we were referencing the old property and update them to work with the new property.

Related to: https://nanopay.atlassian.net/browse/CPF-1627


src/foam/nanos/auth/Relationships.js
```diff
@@ -4,22 +4,14 @@
...
-  forwardName: 'permissions',
+  forwardName: 'permissions2',
   inverseName: 'groups',
-  sourceProperty: {
-    hidden: true
-  },
-  targetProperty: {
-    hidden: true
-  }
+  junctionDAOKey: 'groupPermissionJunctionDAO'
```
adamvy, line 13: I think this is the default?

adamvy (approved):

adamvy:
I didn't merge, leaving it up to you if you to decide and coordinate if we want to migrate for this launch.

TW80000:
> I didn't merge, leaving it up to you if you to decide and coordinate if we want to migrate for this launch.

Thanks. @jlhughes and I will coordinate on that.

PR #2233, opened 2019-06-03
Create an AxiomCloner and fix up a couple bugs

mcarcaso:
The two bugs that were found remove the need for the AxiomCloner but I still think it's the better way to go rather than cloning axioms directly like we do because it can allow us to be lazier about instantiating classes.

### DIFF
--- a/src/foam/core/AxiomCloner.js
+++ b/src/foam/core/AxiomCloner.js
@@ -1,0 +1,57 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.core',
+  name: 'AxiomCloner',
+  documentation: 'An axiom that clones an axiom from another model.',
+  properties: [
+    {
+      class: 'Class',
+      name: 'from'
+    },
+    {
+      class: 'String',
+      name: 'axiom'
+    },
+    {
+      class: 'String',
+      name: 'name',
+      transient: true,
+      expression: function(axiom) {
+        return axiom + '_cloner';
+      }
+    },
+    {
+      class: 'Map',
+      name: 'config'
+    }
+  ],
+  methods: [
+    function installInClass(cls) {
+      var axiom = this.from.getAxiomByName(this.axiom);
+      if ( ! axiom ) {
+        throw `Cannot find ${this.axiom} on ${this.from.id}`;
+      }
+      axiom = axiom.clone().copyFrom(this.config);
+      cls.installAxiom(axiom);
+    }
+  ]
+});
+
+foam.SCRIPT({
+  package: 'foam.core',
+  name: 'AxiomClonerConvenienceMethod',
+  code: function() {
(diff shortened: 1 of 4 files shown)


src/foam/java/refinements.js
```diff
@@ -233,7 +233,7 @@ if ( ! ${foam.java.asJavaValue(vp.predicate)}.f(obj) ) {
     {
       name: 'asJavaValue',
       code: function() {
-        return `${this.sourceCls_.id}.${foam.String.constantize(this.name)}`;
+        return `${this.forClass_}.${foam.String.constantize(this.name)}`;
```
mcarcaso, line 236: This is the reason why I started down this path. sourceCls_ wasn't coming through as I was expecting it to but I shouldn't have been using that property in the first place since it's technically not a property and shouldn't be relied upon.

src/foam/core/Property.js
```diff
@@ -308,6 +308,10 @@ foam.CLASS({
       if ( this.forClass_ && this.forClass_ !== c.id && prop === this ) {
         // Clone this property if it's been installed before.
         prop = this.clone();
+
+        // sourceCls_ isn't a real property so it gets lost during the clone.
+        prop.sourceCls_ = c;
```
mcarcaso, line 313: This seems right but nothing should be broken in its absence because we shouldn't really be relying on this property. The comment around it here: https://github.com/foam-framework/foam2/blob/11463bc73914489079a9781ca148bf82b8531b75/src/foam/core/FObject.js#L149 says it's for the describe method.

src/foam/core/AxiomCloner.js
```diff
@@ -0,0 +1,57 @@
...
+      if ( ! axiom ) {
+        throw `Cannot find ${this.axiom} on ${this.from.id}`;
+      }
+      axiom = axiom.clone().copyFrom(this.config);
+      cls.installAxiom(axiom);
+    }
+  ]
+});
+
+foam.SCRIPT({
```
adamvy: Just foam.LIB({ name: 'foam', methods: [ function axiomCloner() {} ] }) would be shorter.

src/foam/core/Property.js
adamvy: I don't think this is necessary.  installAxiom() sets it, and you're calling installAxiom in your AxiomClone.installInClass().

This code path shouldn't even be hit.  forClass_ shouldn't be set on the cloned axiom that AxiomCloner creates.  We maybe need to clear it, or update the logic on clone() or have a specific transient type for properties such that they don't get cloned.  Maybe override cloneProperty for forClass_ would be easiest.

mcarcaso: forClass_ does get copied over since it's a real property on foam.core.Property. I don't think cloning looks at transient properties. And sourceCls_ gets set during installAxiom but gets cleared when installAxiom calls installInClass.

adamvy: I'm saying we _shouldn't_ copy over forClass_, not that we don't copy over forClass_.

mcarcaso: Chatted offline about this to clarify. This is just a fix for an unrelated bug.

PR #2234, opened 2019-06-03
disable button when asynchronous action

molurin1:


src/foam/u2/ActionView.js
```diff
@@ -346,7 +350,7 @@ foam.CLASS({
         this.attrs({ name: this.action.name });
 
         this.enableClass(this.myClass('unavailable'), this.action.createIsAvailable$(this.__context__, this.data), true);
-        this.attrs({ disabled: this.action.createIsEnabled$(this.__context__, this.data).map((e) => e ? false : 'disabled') });
+        this.attrs({ disabled: this.disabled ? 'disabled' : this.action.createIsEnabled$(this.__context__, this.data).map((e) => e ? false : 'disabled') });
```
adamvy: This means that we only hook up the slot if disabled was not set on first render.  Instead you should create a new slot which computes the disabled attribute based upon action.createIsEnable$ and this.disabled$

adamvy (changes requested):
Would it be possible to take the "running" state in to account in the enabled slot that we get from createIsEnabled$()?  Then none of our action views would have to change or even be aware of async vs sync actions, as they could still just hook up createIsEnabled$() to both.

Additionally we could check if the action is running during Action.maybeCall() and prevent it from being run multiple times from code as well as from the GUI.

There would be a bit of complication here in that we need to track the running state on a per-object basis, but you might be able to use a WeakMap for that.

```
properties: [
  { name: 'runningMap', factory: function() { return new WeakMap(); }, transient: true }
],
...
function createIsEnabled$(x, data) {
  var running = foam.core.SimpleSlot.create({ value: false });
  this.runningMap.set(data, running);
...
  return foam.core.ExpressionSlot({
    args: [ running, ... ],
...
});

function maybeCall(x, data) {
  var running = this.runningMap.get(data);
  if ( running && running.get() ) { console.warn("Attempted to call action that is in progress."); }
  if ( running ) running.set(true);
  var ret = this.code.call(data...);
  if ( ret && ret.then ) {
    ret.then(function() { running && running.set(false); }, function() { running.set(false; });
  }
  return ret;
}
```

Something like that.  Could be cleaned up by creating the running slot during maybecall and during createIfEnable$ if it exists or not.  Could also check if the return value of ret matches Action.isAsync, but we might not even need isAsync, as we can just decide its async depending on whether it returns a promise or not.

What do you think?

molurin1:
Way better approach!

src/foam/core/Action.js
```diff
@@ -122,7 +122,8 @@ If empty than no permissions are required.`
       name: 'enabledPermissions',
       documentation: `Permissions required for the action to be enabled.
 If empty than no permissions are required.`,
-    }
+    },
+    { name: 'runningMap', factory: function() { return new WeakMap(); }, transient: true }
```
adamvy: Add hidden: true and documentation:

src/foam/core/Action.js
```diff
@@ -167,7 +168,19 @@ If empty than no permissions are required.`,
     },
 
     function createIsEnabled$(x, data) {
-      return this.createSlotFor_(x, data, this.isEnabled, this.enabledPermissions);
+      var running = foam.core.SimpleSlot.create({ value: false });
```
adamvy: I think we should check if the runningMap already has a value for data.

If we create two isEnabled$ slots for the same combination of data and Action, then I think we can share running state.  If an action is rendered twice on a single view for instance, then clicking on the action in one button will successfully disable both buttons while running.

src/foam/core/Action.js
```diff
@@ -177,10 +190,21 @@ If empty than no permissions are required.`,
     function maybeCall(x, data) {
       var self = this;
       function call() {
-        self.code.call(data, x, self);
+        var running = self.runningMap.get(data);
```
adamvy: Yeah let's put this in a utility function so we can just call getRunning$(data) both here and createIsEnabled$().  Then its readable and we're guaranteed to only have one running slot per action+data combination.

src/foam/core/Action.js
```diff
@@ -177,10 +190,21 @@ If empty than no permissions are required.`,
...
+        if ( ! running ) {
+          running = foam.core.SimpleSlot.create({ value: false });
+          if ( data ) self.runningMap.set(data, running);
+        }
+        if ( running.get() ) { console.warn("Attempted to call action that is in progress."); }
```
adamvy: This should probably be a call to x.warn().

Should we return early here and prevent calling the action?  Bit of a design decision to make here about the failure case.

On the one hand if we don't call the action that's in progress we protect against data binding and other code mistakes that might trigger the action twice in quick succession.  On the other hand, if the action returns a promise that never resolves, then the user will be unable to try the action again until, possibly ever.

Think about which failure case we care more about and which provides the better user experience, then add a comment here describing why we still call the action or why not.

molurin1: I think the problem of quick successive action trigger can easily be handled by a debounce which I think ActionView.js is already doing. Having User stuck forever on the other hand is less desirable.

src/foam/core/Action.js
```diff
@@ -167,20 +176,52 @@ If empty than no permissions are required.`,
...
 
     function createIsAvailable$(x, data) {
       return this.createSlotFor_(x, data, this.isAvailable, this.availablePermissions);
     },
 
+    function getRunning$(data) {
+      var running = this.runningMap.get(data);
+      if ( ! running ) {
+        running = foam.core.SimpleSlot.create({ value: false });
+        if ( data ) this.runningMap.set(data, running);
```
adamvy, line 200: When would data not be true?

PR #2235, opened 2019-06-04
Cache auth-related DAOs

TW80000:
* Cache groupDAO, permissionDAO, and groupPermissionJunctionDAO
* Move service to main services file so it gets picked up by nanos


adamvy:
So we improved the rendering time already by reducing the number of re-renders.  Can we drop this PR for now and revive it if we need it?

kgrgreer:
I think permissionDAO should definitely be cached. The Junction DAO should probably be cached at least for the PermissionTableView.

TW80000:
Closing since this isn't especially important. Feel free to re-open.

PR #2241, opened 2019-06-04
Clean up a bunch of warnings.

mcarcaso:


src/foam/u2/Element.js
```diff
@@ -2128,10 +2128,19 @@ foam.CLASS({
       name: 'attribute',
       value: false
     },
+    {
+      class: 'String',
+      name: 'placeholder'
```
mcarcaso, line 2133: I've seen us try to set `placeholder` on a property in a few spots or override the view to include a placeholder so this seems like it would be pretty convenient to have.

src/foam/core/debug.js
```diff
@@ -451,7 +451,7 @@ foam.CLASS({
   methods: [
     function installInClass(c, superImport) {
       // Produce warning for duplicate imports
-      if ( superImport ) {
+      if ( superImport && superImport.equals(this) ) {
```
mcarcaso, line 454: This is to suppress a warning that can happen when you do something like:
```
foam.CLASS({
  name: 'ParentClass',
  imports: [ 'foo' ],
})
foam.CLASS({
  name: 'SubClass',
  extends: 'ParentClass',
  imports: [ 'bar as foo' ]
})
```

src/foam/u2/Element.js
```diff
@@ -2128,10 +2128,19 @@ foam.CLASS({
...
+    },
     {
       class: 'foam.u2.ViewSpec',
       name: 'view',
-      value: { class: 'foam.u2.TextField' }
+      expression: function (placeholder) {
```
kgrgreer, line 2138: TextField has a fromProperty() method which is where it can copy settings out of the property. This should be done there.

kgrgreer: commit 99ffd1d132b91d24cf78746d7f4657f419fd1cd0 (HEAD -> master, origin/master, origin/HEAD)
Date:   Tue Jun 4 22:05:38 2019 -0400

    Fix TextField style of setting placeholder.

mcarcaso:
Oh good call. Thanks for cleaning that up :)

PR #2243, opened 2019-06-05
test/trap null email service

jlhughes:


src/foam/nanos/notification/email/EmailServiceDAO.js
```diff
@@ -35,7 +35,11 @@ foam.CLASS({
       name: 'put_',
       javaCode:
       `
-        getEmailService().sendEmail(x, (EmailMessage)obj);
+        EmailService service = getEmailService();
+        if ( service != null ) {
+          ((foam.nanos.logger.Logger) x.get("logger")).debug("EmailServiceDAO emailService null");
```
mcarcaso, line 40: I think you want this debug statement in an else block?

PR #2246, opened 2019-06-05
Convert view values to viewspec.

mcarcaso:


kgrgreer:
I don't think we should have to do this. The ViewSpec is supposed to take view class names as Strings. I think we should fix it there.

PR #2247, opened 2019-06-05
PipelinePMDAO

michaelconnor468:
changed to deal with Reentrancy and made code cleaner


src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,114 @@
...
+
+public class PipelinePMDAO
+  extends ProxyDAO
+{
+  protected String putName_;
+  protected String findName_;
+  protected String removeName_;
+  protected String removeAllName_;
+
+  public PipelinePMDAO(X x, DAO delegate) {
```
kgrgreer, line 24: Also needs the daoKey or name of the DAO to report. Otherwise something like AuthenticateDAO will have the same PM key for all DAO decorator chains and you won't know which one you're recording, or rather, you'll actually be combining the results from different DAO's together.

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,114 @@
...
+      if( secondaryDelegate instanceof ProxyDAO ) {
+        ((ProxyDAO) delegate).setDelegate(new PipelinePMDAO(getX(), secondaryDelegate));
+      }
+    }
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    PM pm = new PM();
+    pm.setClassType(PMDAO.getOwnClassInfo());
```
kgrgreer: Use PipelinePMDAO, not PMDAO.

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  private void createPipeline() {
+    DAO delegate = getDelegate();
+    DAO secondaryDelegate;
+    if( delegate instanceof ProxyDAO ) {
+      secondaryDelegate = ((ProxyDAO) delegate).getDelegate();
+      ((ProxyDAO) delegate).setDelegate(new EndPipelinePMDAO(getX(), secondaryDelegate));
+      delegate = ((ProxyDAO) delegate).getDelegate();
+      if( secondaryDelegate instanceof ProxyDAO ) {
```
kgrgreer: space between 'if' and '('

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,114 @@
...
+    putName_       = getDelegate().getClass().getName() + ":pipePut";
+    findName_      = getDelegate().getClass().getName() + ":pipeFind";
+    removeName_    = getDelegate().getClass().getName() + ":pipeRemove";
+    removeAllName_ = getDelegate().getClass().getName() + ":pipeRemoveAll";
+  }
+
+  private void createPipeline() {
+    DAO delegate = getDelegate();
+    DAO secondaryDelegate;
+    if( delegate instanceof ProxyDAO ) {
```
kgrgreer: spacing

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,114 @@
...
+    PM pm = new PM();
+    pm.setClassType(PMDAO.getOwnClassInfo());
+    pm.setName(findName_);
+    X pipeX = getX().put("pipePmStart", pm);
+    return super.find_(pipeX, id);
+  }
+
+  @Override
+  public FObject remove_(X x, FObject obj) {
+    PM pm = new PM();
```
kgrgreer: The next four lines could be made into a reusable method.

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,114 @@
...
+  @Override
+  public void removeAll_(X x, long skip, long limit, Comparator order, Predicate predicate) {
+    PM pm = new PM();
+    pm.setClassType(PMDAO.getOwnClassInfo());
+    pm.setName(removeAllName_);
+    X pipeX = getX().put("pipePmStart", pm);
+    super.removeAll_(pipeX, skip, limit, order, predicate);
+  }
+
+  public class EndPipelinePMDAO extends ProxyDAO {
```
kgrgreer: You can't have two public classes in the same file, can you?

michaelconnor468: This is an inner class; it compiles ok

src/foam/dao/PipelinePMDAO.java
michaelconnor468: I think the init method makes this distinction when creating the names by appending the name of the delegate before the type of operation. I will however refactor this into its own variable to make things cleaner and more explicit.

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,116 @@
...
+    removeName_    = delegateName_ + ":pipeRemove";
+    removeAllName_ = delegateName_ + ":pipeRemoveAll";
+  }
+
+  /** Creates the PM pipeline by adding an EndPipelinePMDAO after of this class only if it is a ProxyDAO. 
+  *   If the delegate of that is also a ProxyDAO, creates a new PipelinedPMDAO in the chain beofre it which repeats this procedure recursively. */
+  private void createPipeline() {
+    DAO delegate = getDelegate();
+    DAO secondaryDelegate;
+    if ( delegate instanceof ProxyDAO ) {
```
kgrgreer: Put this check in EasyDAO instead. PMPipelineDAO shouldn't be called on non-decorators.

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,116 @@
...
+      delegate = ((ProxyDAO) delegate).getDelegate();
+      if ( secondaryDelegate instanceof ProxyDAO ) {
+        ((ProxyDAO) delegate).setDelegate(new PipelinePMDAO(getX(), secondaryDelegate));
+      }
+    }
+  }
+
+  /** Creates the PM that will measure the performance of each operation and creates a new context with it as a variable which the EndPipelinePMDAO
+   *  will use to access the pm after it is passed onto it through the arguments of the DAO operations */
+  private X createPM(String name) {
```
kgrgreer: Take x as a parameter and use instead.

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,116 @@
...
+    PM pm = new PM();
+    pm.setClassType(PipelinePMDAO.getOwnClassInfo());
+    pm.setName(name);
+    return getX().put("pipePmStart", pm);
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    X pipeX = createPM(putName_);
```
kgrgreer: return super.put_(createPMX(x, putName_), obj);

src/foam/dao/PipelinePMDAO.java
kgrgreer: remove 'public'

src/foam/dao/EasyDAO.js
```diff
@@ -149,9 +149,9 @@ return delegate;
       name: 'innerDAO'
     },
     {
+      /**  Creates pipelinePMDAO's around each decorator to measure their performance */
```
kgrgreer: documentation: 'Creates pipelinePMDAO's around each decorator to measure their performance'

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,110 @@
...
+public class PipelinePMDAO
+  extends ProxyDAO
+{
+  protected String putName_;
+  protected String findName_;
+  protected String removeName_;
+  protected String removeAllName_;
+  protected String delegateName_;
+
+  public PipelinePMDAO(X x, DAO delegate) {
```
kgrgreer, line 24: If you do this instead: public PipelinePMDAO(X x, ProxyDAO delegate) {
then can't even be called with a non ProxyDAO.

PR #2250, opened 2019-06-06
Csv outputter prop list

Nauna:
for CPF-1656


src/foam/comics/DAOController.js
```diff
@@ -172,6 +172,10 @@ foam.CLASS({
       class: 'String',
       name: 'detailView',
       value: 'foam.u2.DetailView'
+    },
+    {
+      name: 'propList',
```
adamvy, line 177: Please add documentation to this property so we know what its for.

How does this differ from tableColumns on the model we're exporting?

PR #2260, opened 2019-06-07
Permission refactor after migration

TW80000:
The second half of https://github.com/foam-framework/foam2/pull/2232. See #2232 for an explanation.


jlhughes (approved):
This has been tested.

PR #2262, opened 2019-06-07
Better RichChoiceView keyboard support

TW80000:
This PR adds support for tabbing to `RichChoiceView`s just like any other native input. The border colour changes as well to visually distinguish that it has been focused. You can press enter when focused to open the dropdown and tab between the options in the list, then press enter to make your selection.


src/foam/u2/view/RichChoiceView.js
```diff
@@ -267,16 +279,18 @@ foam.CLASS({
...
 
       this
-        .attrs({ name: this.name })
+        .attrs({
+          name: this.name,
+          tabIndex: 0
+        })
         .addClass(this.myClass())
+        .on('keydown', function(evt) {
+          if ( evt.which === 13 || evt.keyCode === 13 ) self.toggle();
```
adamvy: Mozilla Developer Network now recommends using evt.key to check the key.

This would be a good time to write a utility or polyfill so we only need to call a single function to check key sizes.

Maybe foam.web.key(evt) which can check evt.key evt.keyCode evt.charCode evt.which, etc. Then our keyboard event handling will be shorter and more cross platform everywhere we do things like this :)

TW80000: I'm with you on the spirit of the idea, but after looking into it I don't think it's going to work.

`key` is preferred and `which`, `keyCode`, and `keyIdentifier` are all deprecated. The problem is that the types of those properties vary, so I'm not sure what the method signature would look like. `key` is a string like "Enter" and `which` and `keyCode` are integers like 13. So would our utility method take one of the two (ie: either the string or the int) and then have to use a huge table to convert between them? I don't particularly like the idea of creating that table.

An alternative is to allow them to supply either the string or the int, but that's obviously not a great plan since they have to be aware of the implementation at that point practically anyway.

And finally if we stick to just one (string, let's say) and don't make the lookup table, then it's no different from the caller just checking `key` directly.

On top of all of that, the developer has to know to use the utility in the first place.

So I'm struggling to see how we can make a useful utility for this case.

TW80000:
@adamvy I updated the code to check `key` instead of the other attributes. [MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Browser_compatibility) says `key` is supported by all current browsers.

Also please see my reply to your comment and review again when you can. Thanks.

src/foam/u2/view/RichChoiceView.js
adamvy: The utility method would take the keyboard event and return the key identifier for that event.  It would just use .key if it exists, or it would interpret keyCode/charCode/which as needed.

A table to convert from deprecated keyCode/charCode/which to the non-deprecated "key" identifier would be an entirely reasonable thing to have.  I agree it wouldn't be fun to create, but solving it now prevents us having to play whack a mole to find these bugs later or crossing our fingers hoping that we catch them all in review.  For somebody less familiar with the situation, reading code that checks "keyCode" "which" or any of the existing properties on the event seems entirely reasonably, its only when you have a reviewer with really specific domain knowledge that you will catch it.

It might not even require a very big table.  Most characters return their printed representation, it's only control and unprintable characters that require mapping to some known name.  So through some clever combination of String.fromCodePoint and checking the range that the key code lies in might not be that much work.

The developer has to know to use the utility but that can be said of literally any feature we provide.  I would rather solve this problem than train every developer to recognize this edge case of web programming.  Even to many experienced web programmers, code that checks "keyCode" looks entirely reasonable.  Additionally we can take steps to force or encourage the developer to use our interface.  If we decorate the listener during .on() in foam.u2.Element, we don't have to give them the raw keyboard event.  We could instead return them our own custom modeled keyboard event.  Or we could do other fun things like modify the KeyboardEvent.prototype to remove keyCode/charCode when running in developer mode, preventing any FOAM dev from getting useful results from those properties.

A lot of the FOAM framework (and any web framework) is providing better and more consistent interfaces to the browser than are provided out of box.  It's part of why we have the u2 package in the first place.  The web is a strange and messy platform. The less domain specific knowledge people need to build apps in FOAM, the fewer bugs they will create and the more productive they will be.

TW80000: @adamvy Updated to include the utility. Please review again.

src/foam/u2/view/RichChoiceView.js
```diff
@@ -59,6 +63,14 @@ foam.CLASS({
       z-index: 1;
     }
 
+    ^:focus {
+      outline: none;
+    }
+
+    ^:focus ^selection-view {
+      border-color: %SECONDARYCOLOR%;
```
kgrgreer, line 71: consider:
border-color: /*%SECONDARYCOLOR%*/ somereasonablecolor;

So that it has a default value when not used with Nanos.

src/foam/web/KeyboardUtil.js
```diff
@@ -0,0 +1,199 @@
...
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.LIB({
+  name: 'foam.web.KeyboardUtil',
+
+  documentation: 'A utility library to make it easier to work with keyboard events.',
+
+  constants: {
```
kgrgreer, line 12: How does this compare to Element.NAMED_CODES and keyboardShortcuts support?

TW80000: I hadn't seen that stuff before, but I just had a look at it and this utility method seems to be most similar to `Element.evtToCharCode`. The point of it is to provide a reliable way to get a string representing the key a user pressed from a KeyboardEvent. Looks like there are some differences between `Element.evtToCharCode` and this though. `Element.evtToCharCode` uses deprecated properties on KeyboardEvent and uses non-standard strings from the codes, but handles modifier keys like ctrl, alt, and shift and combinations thereof. `KeyboardUtil.getKey` is more standards-compliant and supports more browsers and keys, but doesn't include the modifier keys in the string it returns.

src/foam/u2/view/RichChoiceView.js
TW80000: I can do another PR after the colour palette PR merges to do this across the board or just modify that one.

src/foam/web/KeyboardUtil.js
```diff
@@ -0,0 +1,199 @@
...
+      'Esc': 'Escape',
+    }
+  },
+
+  methods: [
+    function getKey(keyboardEvent) {
+      /**
+       * Returns a string that represents the key pressed.
+       *
+       * WARNING: Only works for `keydown` or `keyup` events if the browser does
```
adamvy: First sentence is confusing.  "Only works for ... if the browser does not support 'KeyboardEvent.key'"  Surely this function works when the browser _does_ support KeyboardEvent.key as well?

src/foam/u2/view/RichChoiceView.js
kgrgreer: Yes, better to do across the board in another PR.

src/foam/web/KeyboardUtil.js
TW80000: Thanks for catching that, fixed. I also added a TODO comment to note that we can indeed support keypress events on old browsers if we add another lookup table and pass the event name as a parameter.

src/foam/u2/view/RichChoiceView.js
TW80000: Done in https://github.com/foam-framework/foam2/pull/2298.

src/foam/web/KeyboardUtil.js
adamvy: Can we unify the two?

adamvy: modifier keys seems like it would be useful to include

PR #2263, opened 2019-06-07
allow DAO to be passed through stack to DAOUpdateControllerView

RichardZhu123:


mcarcaso:
FYI I think this is backwards. `dao? as importedDAO` is what we want.

mcarcaso:
https://github.com/foam-framework/foam2/commit/4cb94a021e36b63e233a9637d9e49c116ff15b16

PR #2264, opened 2019-06-10
[do not merge] adjusting csvSink

Nauna:
working progress. Not ready for real review.


src/foam/dao/CSVSink.js
```diff
@@ -11,6 +11,14 @@ foam.CLASS({
 
   documentation: 'Sink runs the csv outputter, and contains the resulting string in this.csv',
 
+  implements: [
+    'foam.core.Serializable'
+  ],
+
+  javaImports: [
```
Nauna: probably unnecessary, but build doesn't like foam.java

Nauna:
currently getting errors:

Unknown property foam.core.FObject.name: obj
ERROR: Unhandled promise rejection  Required property foam.core.FObjectArray.of not defined.

so I'm confused where to use foam.java.PropertyInfo or foam.core.Property.

Which also plays into how to properly do the javaToCSV() and javaToCSVLabel().

src/foam/dao/CSVSink.js
```diff
@@ -40,22 +70,43 @@ foam.CLASS({
       name: 'isNewLine',
       value: true,
       visibility: 'HIDDEN'
+    },
+    {
+      name: 'axioms',
+      factory: function() {
+        return this.of.getAxiomByName('tableColumns').columns;
```
adamvy: I think we should name these columns and columns_ instead of axioms and props, to be consistent with TableView.

src/foam/dao/CSVSink.js
```diff
@@ -40,22 +70,43 @@ foam.CLASS({
...
+      args: [
+        { name: 'value' }
+      ],
+      code: function(value) {
+        if ( ! this.isNewLine ) this.csv += ',';
+        this.isNewLine = false;
+        this.output_(value);
+      },
+      javaCode: `
+        if ( ! getIsNewLine() ) setCsv(getCsv() + ",");
```
adamvy: We should use a StringBuilder in java rather than repeated string concatenation, it will be much faster.

src/foam/dao/CSVSink.js
```diff
@@ -40,22 +70,43 @@ foam.CLASS({
...
       name: 'output_',
+      args: [
+        { name: 'value' }
+      ],
       code:
         foam.mmethod(
           {
             String: function(value) {
-              this.csv += `"${value.replace(/\"/g, '""')}"`;
+              this.csv += `\"${value.replace(/\"/g, '\"\"')}\"`;
```
adamvy: This change should be unnecessary, I don't think you need to backslash esacpe double quotes inside a backquote in JS.

src/foam/dao/CSVSink.js
```diff
@@ -40,22 +70,43 @@ foam.CLASS({
...
 
   methods: [
-    function output(value) {
-      if ( ! this.isNewLine ) this.csv += ',';
-      this.isNewLine = false;
-      this.output_(value);
+    {
+      name: 'output',
+      args: [
+        { name: 'value' }
```
adamvy, line 80: Should give this a type, type: 'Any' may be appropriate.

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +127,95 @@ foam.CLASS({
             Null: function(value) {}
           }, function(value) {
             this.output_(value.toString());
-          })
+        }),
+      javaCode: `
+        switch (value) {
+          case value == null:
```
adamvy: I don't think this is valid java code?

You'll need to do a series of if () else if () else if (), similar to how JSON Outputter operates.

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +127,95 @@ foam.CLASS({
...
           }, function(value) {
             this.output_(value.toString());
-          })
+        }),
+      javaCode: `
+        switch (value) {
+          case value == null: 
+            break;
+          case value instanceof String:
+            setCsv(getCsv() +  value.replace("\\"", "\\"\\""));
```
adamvy: We should also wrap the string in quotes

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +127,95 @@ foam.CLASS({
...
+            setCsv(getCsv() + value.toString());
+            break;
+          case value instanceof Boolean: 
+            setCsv(getCsv() + value.toString());
+            break;
+          case value instanceof Date:
+            output_(value.toDateString());
+            break;
+          case value instanceof Array:
+            output_(foam.json.Pretty.stringify(value));
```
adamvy: The Pretty outputter can add newlines, which would break the CSV format, most CSV handlers don't allow newlines.

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +127,95 @@ foam.CLASS({
...
+      },
+      javaCode: `
+        setCsv(getCsv() + "\\n");
+        setIsNewLine(true);
+      `
     },
+    {
+      name: 'put',
+      args: [
+        { class: 'FObject', name: 'obj' }
```
adamvy: type: 'FObject' not class

src/foam/dao/CSVSink.js
adamvy: Actually upon further thought we might not need these conditionals at all.  If you add outputCSV() to PropertyInfo, you can just call that for each axiom we want to output rather than doing the formatting here.

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +127,95 @@ foam.CLASS({
...
+          }
+          newLine_();
+          setIsHeadersOutput(true);
+        }
 
-      this.props.forEach((element) => {
-        element.toCSV(obj, this, element);
-      });
-      this.newLine_();
+        for (FObject element : props) {
```
adamvy: PropertyInfo's are not FObjects.  Should be for (PropertyInfo : axoims)

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +127,95 @@ foam.CLASS({
...
+          element.javaToCSV(obj, this, element);
```
adamvy: Shouldn't need to pass in element, as the "this" of the javaToCSV method will be the elemen.

src/foam/dao/CSVSink.js
```diff
@@ -145,7 +257,7 @@ foam.CLASS({
 
   properties: [
     {
-      name: 'toCSV',
+      name: 'javaToCSV',
```
adamvy: Presumably we still want the JS implementations as well, so the client, so  Iwouldn't delete toCSV.

Take a look at src/foam/java/refinements.js for javaCsvParser and/or javaCloneProperty.

You need to add a javaOutputCSVLabel and javaOutputCSVValue or similar properties to Property as class: 'String', update the createJavaPropertyInfo_ method, and then update the methods factory of src/foam/java/PropertyInfo.js

adamvy:
In answer to your questions about Property vs PropertyInfo.

There are two entities named PropertyInfo.  foam.java.PropertyInfo is a modeled/javascript class, which generates code for a Java PropertyInfo instance.  It lives in foam/java/PropertyInfo.js.  foam.core.PropertyInfo is a plain java interface that lives in foam/core/PropertyInfo.java.

We do not actually generate foam.core.Property in to Java at the moment.  Instead we take a Property and convert it to a foam.java.PropertyInfo, which then generates the code we need for the property constant and class info on a Java class.  This transformation from foam.core.Property to foam.java.PropertyInfo happens in foam/java/refinements.js createJavaPropertyInfo_.

src/foam/dao/CSVSink.js
Nauna: https://www.w3schools.com/java/java_switch.asp

and adding the epiphany idea - second - comment here.

Nauna: Currently:
So we are creating a default toCSV on properties(js/java) which can be overridden for any type we want unique behavior.
The output( and output_ ) functions, basically convert csv output of each property to string.

So epiphany idea as I understand it: make toCSV/toCSVLabel(for both js and java) a function that returns a string, and then we can rid ourselves of output functions in CSVSink

correct??

src/foam/dao/CSVSink.js
Nauna: should "this" not be the outputter(csvSink)?

Nauna:
PERSONAL notes on progress (: - S ) will need some clarification on something so stashing this for now.

FILES (changes in last work commit)

FOAM

AbstractEnumPropertyInfo: line(40) - comment out toCSV override
AbstractPropertyInfo: line(46-49) - comment out toCSV override
PropertyInfo.java: line(46-47) - add abstract methods to base
CSVSink: line(19(added), 23-26(added), console.logs, and System.out's)
PropertyInfo.js: line(112-113(removed properties toCSV/toCSVLabel), 295-298&304-307(changed args and body of methods))
Refinements.js: line(229-243(removed prp def), commented out override* 1120-1139)
Outputted.java: line(179) comment out invalid format toCSV call
CSVDriver.js: line(31-58) rewrite function to java

NANOPAY - NOT COMMITED - redo when re-run

iso20022/refinements.js lines(59-68) comment out override

ISSUES:
1) CSVSink.java never called. When exportDAO is in js we call sink.reset from js, and then never go back into the sink. Also if in java the call to this function fails.
2) need to make toCSV/label prop on propertyInfo.
3) re-add the commented out overrides mentioned above in FILES

PR #2267, opened 2019-06-10
Moved groupPermissionJunction files to src directory and deleted the …

tharmaman:
…old files, as well as added wildcard permissions to the system user


TW80000 (approved):

src/groupPermissionJunctions
```diff
@@ -0,0 +1,2 @@
+p({"class":"foam.nanos.auth.GroupPermissionJunction","sourceId":"admin","targetId":"*"})
```
jlhughes, line 1: please add the jrl ending, trying to standardize on that. .

jlhughes: foam doesn't support .jrl.

PR #2269, opened 2019-06-11
Extend pipe pmdao

michaelconnor468:
fixed issue of pipe fragmentation and added logic to ensure proper restructuring of a fragmented pipe


src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  /** Creates the PM that will measure the performance of each operation and creates a new context with it as a variable which the EndPipelinePMDAO
+   *  will use to access the pm after it is passed onto it through the arguments of the DAO operations */
+  X createPMX(String name, X x) {
+    PM pm = new PM();
+    pm.setClassType(PipelinePMDAO.getOwnClassInfo());
+    pm.setName(name);
+    return x.put("pipePmStart", pm);
```
kgrgreer, line 59: insert this line:
pm.init_();

It will cause startTime to be set, otherwise PM's will record incorrect time.

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,135 @@
...
+    ((ProxyDAO) delegate).setDelegate(new EndPipelinePMDAO(getX(), secondaryDelegate));
+    delegate = ((ProxyDAO) delegate).getDelegate();
+    if ( secondaryDelegate instanceof ProxyDAO ) {
+      ((ProxyDAO) delegate).setDelegate(new PipelinePMDAO(getX(), secondaryDelegate));
+    }
+  }
+
+  /** Creates the PM that will measure the performance of each operation and creates a new context with it as a variable which the EndPipelinePMDAO
+   *  will use to access the pm after it is passed onto it through the arguments of the DAO operations */
+  X createPMX(String name, X x) {
```
kgrgreer: Convention is to put X as first parameter.

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public void removeAll_(X x, long skip, long limit, Comparator order, Predicate predicate) {
+      PM pm = (PM) x.get("pipePmStart");
+      if ( pm != null )
+        pm.log(x);
+      super.removeAll_(x, skip, limit, order, predicate);
+    }
+  }
+
+  void removePipelineFragmentation() {
```
kgrgreer: What does this do?

src/foam/dao/PipelinePMDAO.java
```diff
@@ -0,0 +1,135 @@
...
+  /** Creates the PM pipeline by adding an EndPipelinePMDAO after of this class only if it is a ProxyDAO. 
+  *   If the delegate of that is also a ProxyDAO, creates a new PipelinedPMDAO in the chain beofre it which repeats this procedure recursively. */
+  void createPipeline() {
+    removePipelineFragmentation();
+    DAO delegate = getDelegate();
+    DAO secondaryDelegate;
+    secondaryDelegate = ((ProxyDAO) delegate).getDelegate();
+    ((ProxyDAO) delegate).setDelegate(new EndPipelinePMDAO(getX(), secondaryDelegate));
+    delegate = ((ProxyDAO) delegate).getDelegate();
+    if ( secondaryDelegate instanceof ProxyDAO ) {
```
kgrgreer: And check and that it isn't a PMPipelineDAO.

src/foam/dao/PipelinePMDAO.java
michaelconnor468: It essentially just cleans up any "junk" pipePMDAO decorators to prevent double layering of pipePMDAO's. Say you decorate account, then localAccount is also decorated as a result and then if you decorate localAccount you are essentially doubling down except on the initial decorator.

kgrgreer: I don't think this is needed now that you added the check to see if the delegate is a PMPipelineDAO.

PR #2270, opened 2019-06-11
#comment removed check for if array is empty

noodlemoodle:


TW80000 (approved):

src/foam/core/JSON.js
```diff
@@ -360,9 +360,6 @@ foam.CLASS({
       if ( this.outputOwnPropertiesOnly && ! o.hasOwnProperty(p.name) )
         return false;
 
-      if ( foam.Array.isInstance(v) && v.length == 0 )
```
kgrgreer, line 363: Instead of just deleting can you comment it out with an explanation?

PR #2271, opened 2019-06-11
Fix CSS generation

TW80000:
The problem this fixes:

We have 2 views: `IntView` and `TextField`. `IntView` extends `TextField`. `TextField` has a bunch of CSS.

Currently, if an `IntView` gets created before a `TextField` does, then when a `TextField` gets created later, its CSS is not generated. The CSS is generated for the `IntView` though.

These changes avoid that problem though and the CSS gets generated for both.

I'm not exactly sure how or why this fixes the issue though.


kgrgreer:
I've merged, but need to do some tests to make sure this isn't causing some other issues.

PR #2272, opened 2019-06-11
Simplify FileArrayDAODecorator

TW80000:
I did this while trying to fix a bug but this file didn't end being the cause of the bug.

Just thought I'd open a PR for the refactor in case anyone thought this version was more readable. If not then I'll just close this PR.


adamvy:
I like it, thanks for the cleanup!

PR #2274, opened 2019-06-11
File upload view

mcarcaso:
Create a view that can take a file and stores it as a string.

FileUpload

The video just shows this view in action. It's just a button with a checkbox beside it. The checkbox is checked when the value of the property isn't empty. The video just shows two views bound to eachother. The uploaded document set the `uploaded` field and then when I clear the text of that field, the checkbox on the upload field goes away.

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -423,6 +423,7 @@ FOAM_FILES([
   { name: "foam/u2/view/RadioView", flags: ['web'] },
   { name: "foam/u2/view/TextField", flags: ['web'] },
   { name: "foam/u2/view/TreeView", flags: ['web'] },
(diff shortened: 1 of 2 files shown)


adamvy:
What for?  We have Blob and File properties, as well as a file upload service so you can actually upload files as is rather than store as a string.

mcarcaso:
> What for? We have Blob and File properties, as well as a file upload service so you can actually upload files as is rather than store as a string.

This is so we can just turn a file directly into a string and not deal with it as a blob. I had this same view in foam1 and used it a bunch when I was doing things with CSVs on the client.

PR #2279, opened 2019-06-12
table cell formatter displays id for FObjectProperties

RichardZhu123:
Previously, TableViews displaying FObjectProperties in a column would not display anything, throwing a null value error.

### DIFF
--- a/src/foam/u2/view/TableCellFormatter.js
+++ b/src/foam/u2/view/TableCellFormatter.js
@@ -127,7 +127,7 @@ foam.CLASS({
       name: 'tableCellFormatter',
       value: function(obj) {
         this.start()
-          .add(obj.id)
+          .add(obj.toSummary())
         .end();
       }
     }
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/TableCellFormatter.js
```diff
@@ -121,7 +121,17 @@ foam.CLASS({
...
   refines: 'foam.core.FObjectProperty',
 
-  properties: [ [ 'tableCellFormatter', null ] ]
+  properties: [
+    {
+      class: 'foam.u2.view.TableCellFormatter',
+      name: 'tableCellFormatter',
+      value: function(obj) {
+        this.start()
+          .add(obj.id)
```
jlhughes: FObjectProperty should have a toSummary() method which should used in placed of this,. 
And this summary can overridden in the model, making this default tablecellformatter very useful.

jlhughes (changes requested):

src/foam/u2/view/TableCellFormatter.js
kgrgreer: FObject's have toSummary() methods. So you could call obj.toSummary() instead of obj.id, and then make sure interested classes have meaningful implementations of toSummary().

RichardZhu123: done!

jlhughes (approved):

PR #2281, opened 2019-06-12
Fixes the canvas sizing issue for CandlestickDAOChartView

kennyqiyenkan:
As title says.

### DIFF
--- a/src/org/chartjs/CandlestickDAOChartView.js
+++ b/src/org/chartjs/CandlestickDAOChartView.js
@@ -81,16 +81,12 @@ foam.CLASS({
     {
       class: 'Int',
       name: 'width',
-      factory: function() {
-        return 750;
-      }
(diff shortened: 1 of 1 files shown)


src/org/chartjs/CandlestickDAOChartView.js
```diff
@@ -75,14 +77,28 @@ foam.CLASS({
       class: 'foam.mlang.ExprProperty',
       name: 'yExpr',
       factory: function() { return this.Candlestick.AVERAGE; }
+    },
+    {
+      class: 'Int',
+      name: 'width',
+      factory: function() {
```
kgrgreer: Why factory instead of just doing value: 750?

kennyqiyenkan: true. will change.

PR #2282, opened 2019-06-12
Cpf 946/dao cont improvements

tharmaman:


src/foam/comics/v2/DAOUpdateView.js
```diff
@@ -71,34 +76,30 @@ foam.CLASS({
...
-        alert('TODO');
-      }
-    }
   ],
   methods: [
     function initE() {
       var self = this;
       this.SUPER();
+      
+      const originalName = this.data.name
```
mcarcaso, line 101: Why declare this up here like this instead of just doing data.name below?

mcarcaso (approved):

PR #2286, opened 2019-06-14
Created userFileUploadWebAgent

minsunny:
This must be merged with https://github.com/nanoPayinc/NANOPAY/pull/7176

https://nanopay.atlassian.net/browse/CPF-1139

FileUploadWebAgent to upload files(AdditionalDocuments) on a given user via API


src/foam/box/RequestWrapper.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.servlet.ServletInputStream;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import org.apache.commons.io.IOUtils;
+
+/**
+ * HTTP request body can be read/parsed only once
+ * as the client has sent it only once and won't send it multiple times.
```
adamvy, line 19: Can you describe in what situation we need this wrapper?  Currently this will create a complete copy of every request, before we've done any sort of authentication of size checks, this could lead to denial of service attacks if someone repeatedly sends us gigantic requests.

I don't think we should be reading the body as a byte array, and also trying to parse it with getParameter(), we should instead parse the query string only if we want query string parameters as well to do our own parsing of the body.

adamvy: Additionally copying every request just in case we want to parse it both ways will hurt performance, even if they are not malicious requests.

minsunny: @adamvy  We are supposed to parse 'multipart/form-data' and it has to be parsed by parseRequest() which is supported by common file upload lib. By the way I noticed it returns as empty and found out this.

minsunny: https://liferay.dev/forums/-/message_boards/message/4550900

Why is parseRequest() returning no items?
This most commonly happens when the request has already been parsed, or processed in some other way. Since the input stream has aleady been consumed by that earlier process, it is no longer available for parsing by Commons FileUpload.

minsunny: After a few investigations, I found out wrapper can solve this problem.

jlhughes:
Replaced by Refactor with DocumentDAO.

PR #2287, opened 2019-06-14
Model PMDAO, NSpecAware and update LoggingDAO

jlhughes:
- PMDAO now modelled
- LoggingDAO refactored in nanos.logger.LoggingDAO and combines js logging in dao.LoggingDAO with java DAO Logging.
- EasyDAO uses new logger.LoggingDAO
- EasyDAO supports DeletedAware, CreatedAware, LastModifiedAware
- EasyDAO.init_ will fail if 'of' is not set.
- EasyDAO now uses MDAO rather than MapDAO. 
- All services EasyDAO configurations add setOf where missing.
- NSpecAware modelled and NSpecFactory will now walk created NSpec chain and test/set each for ContextAware, NSpecAware.


src/foam/dao/PMDAO.js
```diff
@@ -0,0 +1,201 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2019

src/foam/nanos/boot/NSpecAware.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2019

src/foam/nanos/boot/NSpecFactory.java
```diff
@@ -46,10 +47,18 @@ public Object create(X x) {
...
-        if (logger != null) logger.info("Created Service", spec_.getName());
-
-        if (ns_ instanceof ContextAware) ((ContextAware) ns_).setX(x_.getX());
-        if (ns_ instanceof NanoService) ((NanoService) ns_).start();
+        Object ns = ns_;
+        while ( ns != null ) {
+          if (ns instanceof ContextAware) ((ContextAware) ns).setX(x_.getX());
+          if (ns instanceof NSpecAware) ((NSpecAware) ns).setNSpec(spec_);
+          if (ns instanceof NanoService) ((NanoService) ns).start();
+          if (ns instanceof ProxyDAO) {
```
kgrgreer, line 55: What is this for?

jlhughes: Previously, this test and set only occured on the top spec.  But in serviceScript, if you wrapped your dao in proxyDAO or a DAO which was not NSpecAware, then the 'set' never occurred. EasyDAO is NSpecAware and this makes sure it gets set.  This also came up during simple clustering setup where the NSpec was needed but not set because only the top dao was inspected.

src/foam/dao/PMDAO.js
jlhughes:
```suggestion
 * Copyright 2019 The FOAM Authors. All Rights Reserved.
```

src/foam/nanos/boot/NSpecAware.js
jlhughes:
```suggestion
 * Copyright 2019 The FOAM Authors. All Rights Reserved.
```

src/foam/nanos/boot/NSpecFactory.java
kgrgreer: Should we then call dao.cmd(spec_), then any DAO's in the chain that are interested can get the NSpec?

PR #2290, opened 2019-06-15
If relationship called by system user then attempt to use unauthentic…

jlhughes:
…ated local dao to avoid the unnessary overhead of authentication checks.


jlhughes:
Replaced by a Refinement on the nanopay side, as the 'local' naming convention is nanopay.

PR #2292, opened 2019-06-15
New CSV parser

mcarcaso:
This approach allows properties to customize how they get parsed and potentially have multiple columns.

This includes a demo and a menu item for viewing the demo. The demo allows you to paste a CSV in a text box and parse it into the specified class.

To see the demo, add this to your menu journal:
```
p({"class":"foam.nanos.menu.Menu","id":"csv","label":"CSV","handler":{"class":"foam.nanos.menu.ViewMenu"}})
p({"class":"foam.nanos.menu.Menu","id":"csv","handler":{"class":"foam.nanos.menu.ViewMenu","view":{"class":"foam.u2.DetailView","of":"foam.demos.csv.CsvUploadDemo"}}})
```

And click on the CSV menu.


mcarcaso:
Sounds like there's a growing need for this and it's working so I'll remove the "DO NOT MERGE" on it. I'm not sure what to actually name this stuff as `CSVParser2` is probably not the greatest name. Any suggestions on where this should go and what to call the `fromCSVLabelMapping` property would also be appreciated.

src/foam/lib/csv/CSVParser2.js
```diff
@@ -0,0 +1,165 @@
...
+        };
+      },
+      actions: {
+        unquotedText: function (a) {
+          return a.join('');
+        },
+        quotedText: function (a) {
+          return a.join('');
+        },
+        escapedQuote: function () { return '"'; }
```
kgrgreer: Why is this an action instead of a symbol:?

src/foam/lib/csv/CSVParser2.js
```diff
@@ -0,0 +1,165 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
```
kgrgreer: Document how this is different than the regular one and why I would want to use it.

src/foam/lib/csv/CSVParser2.js
mcarcaso: It's kind of subtle to see but this replaces `""` (i.e. two double quotes) with just one `"`.

src/foam/lib/csv/CSVParser2.js
mcarcaso: Added documentation and renamed file. Name sound ok?

PR #2293, opened 2019-06-16
DAOFoldManager and DAOReduceManager open/close time expressions

mcarcaso:
Allow the DAOFoldManager and DAOReduceManager to take expressions for openTime and closeTime. This will allow us to properly to monthly or quarterly reducing or folding.


src/foam/nanos/analytics/DAOReduceManager.js
```diff
@@ -42,19 +61,18 @@ foam.mlang.sink.GroupBy dataToReduce = (foam.mlang.sink.GroupBy) getSourceDAO()
...
   for ( Object o : ((foam.dao.ArraySink) dataToReduce.getGroups().get(key)).getArray() ) {
     Candlestick c = (Candlestick) o;
-    long cCloseTimeMs = c.getCloseTime().getTime();
-    long rCloseTimeMs = cCloseTimeMs - ( cCloseTimeMs % periodMs ) + periodMs;
-    if ( ! reducedData.containsKey(rCloseTimeMs) ) {
-      reducedData.put(rCloseTimeMs, new Candlestick.Builder(x)
-        .setOpenTime(new Date(rCloseTimeMs - periodMs))
-        .setCloseTime(new Date(rCloseTimeMs))
+    Date rCloseTime = (Date) getCloseTimeExpr().f(c.getCloseTime());
+    if ( ! reducedData.containsKey(rCloseTime) ) {
```
adamvy, line 60: Can we make Candlestick a sink? Or a CandlestickSink so that we don't have to do this looping ourselves?  Would be much better if we just wrote.

```
getSourceDAO().where().select(GROUP_BY(Candlestick.KEY, CANDLESTICK()))
```

Where CANDLESTICK is a sink which creates a new candelstick on first put, and reduces into the existing one on every subsequent put.

src/foam/nanos/analytics/DAOFoldManager.js
```diff
@@ -20,34 +20,45 @@ foam.CLASS({
...
+    },
+    {
+      class: 'foam.mlang.ExprProperty',
+      name: 'closeTimeExpr'
     }
   ],
   methods: [
     {
       name: 'foldForState',
-      // TODO: use a more efficient key based lock
```
adamvy: Is this TODO no longer relevant?  The method is still synchronized.

src/foam/nanos/analytics/DAOReduceManager.js
adamvy: Or maybe there's room for a "REDUCE" sink which takes a "reducable" and calls .reduce on its delegate for every put.

mcarcaso: Yeah I like how you're thinking about this but I'm thinking we do that in a follow-up PR. This PR opens the door to being able to do monthly/quarterly candlesticks and we'll follow this up with better naming of our interfaces/methods for this stuff and a sink or two for folding and reducing.

src/foam/nanos/analytics/DAOFoldManager.js
mcarcaso: My bad, didn't mean to remove it. Started looking at actually addressing the TODO but stepped back because I don't have time to really get into it at the moment.

kgrgreer: I removed/improved the synchronization.

mcarcaso: Yup so this comment can finally go :)
Thanks Kevin!

PR #2294, opened 2019-06-16
NullDAO constructor taking 'of'. 

jlhughes:


src/foam/dao/NullDAO.js
```diff
@@ -22,6 +22,18 @@ foam.CLASS({
 
   documentation: 'A Null pattern (do-nothing) DAO implementation.',
 
+  properties: [
+    {
+      name: 'of',
+      factory: function() {
+        return this.cls_;
```
kgrgreer: Isn't this going to return NullDAO? Is that useful? Better to set value: foam.core.FObject?

jlhughes: Had to pick something.  Need to satisfy the getOf().getId() in PMInfoDAO. 
I'll update. 
What should the js be?  just   return foam.core.FObject;   ?
In java FObject is an interface.  Not sure what to use or call?

kgrgreer: Does AbstractFObject work?

jlhughes:
Use Builder instead.

PR #2298, opened 2019-06-17
Colour scheme support on theme

TW80000:
* Adds properties to Theme that represent a colour palette
* Removes the old colour properties and updates all references to reference the new colours

The default FOAM colour scheme is pictured below.

Screen Shot 2019-06-17 at 3 43 27 PM


adamvy:
My main concern with this PR is that if breaks all views when not running under nanos.  We should ensure that those macros _always_ expand to reasonable defaults even if you're not running under nanos.

TW80000:
> My main concern with this PR is that if breaks all views when not running under nanos. We should ensure that those macros _always_ expand to reasonable defaults even if you're not running under nanos.

Yeah, I'll change everything to the form `/*%WHATEVER%*/ #abc123`.

adamvy:
Sounds good

TW80000:
> Sounds good

@adamvy Updated.

PR #2303, opened 2019-06-18
DAOContCRUDWiringAndRefactors

tharmaman:


src/foam/comics/v2/DataSummaryCard.js
```diff
@@ -0,0 +1,69 @@
...
+      name: 'controllerMode',
+      factory: function() {
+        return this.ControllerMode.VIEW;
+      }
+    },
+    {
+      class: 'FObjectArray',
+      of: 'Reference',
+      name: 'choices',
+      factory: function() {
```
kgrgreer: Isn't this already the default factory for FObjectArray?

src/foam/comics/v2/DataSummaryCard.js
```diff
@@ -0,0 +1,69 @@
...
+      this
+        .addClass(this.myClass())
+        .start(this.CardBorder)
+          .start(this.Rows)
+            .start(this.Cols).addClass(this.myClass('card-header'))
+              .start().add().addClass().end()
+              .forEach()
+            .end()
+          .end()
+        .end()
```
kgrgreer: ;

tharmaman: Done.

src/foam/comics/v2/DataSummaryCard.js
tharmaman: Removed.

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -94,7 +99,14 @@ foam.CLASS({
...
       name: 'delete',
       code: function() {
-        alert('TODO');
+        var self = this;
+        this.config.dao.remove(this.data).then(function() {
+          self.finished.pub();
+          self.stack.back();
+        }, function(e) {
+          self.exception = e;
+          self.throwError.pub();
```
adamvy: Who is listening for throwError?  It might be useful to add the exception as an argument to pub() so that receivers don't have to go looking for it.

adamvy (approved):

src/foam/comics/v2/DAOSummaryView.js
tharmaman: Done. Added the exception as an argument to pub() to DAOSummaryView, DAOCreateView and DAOUpdateView.

src/files.js
```diff
@@ -632,6 +632,7 @@ FOAM_FILES([
   { name: "foam/comics/v2/DAOUpdateView" },
   { name: "foam/comics/v2/NamedViews" },
   { name: "foam/comics/v2/ViewSpecWithJava" },
+  { name: "foam/comics/v2/DataSummaryCard" },
```
mcarcaso: Let's leave this out for now since it's not fully baked and not really being used yet (unless I'm missing something?)

src/foam/comics/v2/DAOBrowseControllerView.js
```diff
@@ -80,7 +80,8 @@ foam.CLASS({
         this.stack.push({
           class: 'foam.comics.v2.DAOCreateView',
           data: this.data.of.create({}, this),
-          config$: this.config$
+          config$: this.config$,
+          of: this.data.of
```
mcarcaso, line 84: Do you need both `of` and `data`? It feels redundant to have both. If the DAOCreateView knows the `of`, it could just create the `data` for you. And the `of` can be found from the `config` so you may only need the config and nothing more. What do you think?

src/foam/comics/v2/DAOCreateView.js
```diff
@@ -49,6 +58,13 @@ foam.CLASS({
       class: 'FObjectProperty',
       name: 'data'
     },
+    {
+      class: 'FObjectProperty',
+      name: 'workingData',
+      expression: function(data) {
+        return data.clone(this.__subContext__)
```
mcarcaso: For create, I don't think you need to have a workingData. Just have `data` and give it an expression based on the config$of.

src/foam/comics/v2/DAOCreateView.js
```diff
@@ -68,13 +84,30 @@ foam.CLASS({
       }
     }
   ],
+  actions: [
+    {
+      name: 'save',
+      code: function() {
+        var self = this;
+        this.data = this.workingData;
+        this.data.createdBy = this.__subContext__.user.id;
```
mcarcaso: `createdBy` and `owner` shouldn't be filled in here.

src/foam/comics/v2/DAOCreateView.js
```diff
@@ -68,13 +84,30 @@ foam.CLASS({
...
+        this.data.owner = this.__subContext__.user.id;
+        this.config.dao.put(this.data).then(function() {
```
mcarcaso: I think you can simplify this method to be something like
```
this.config.dao.put(this.data).then(o => {
  this.data = o;
  this.finished.pub();
  this.stack.back();
}, e => {
  this.throwError.pub(e);
})
```

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -139,7 +150,7 @@ foam.CLASS({
               .end()
 
               .start(config$viewBorder)
-                .start(foam.u2.detail.SectionedDetailView, { data: data })
+                .start(foam.u2.detail.SectionedDetailView, { data: data.clone(self.__subContext__) })
```
mcarcaso: Ditch this cloning.

src/foam/comics/v2/DAOUpdateView.js
```diff
@@ -59,6 +59,13 @@ foam.CLASS({
       class: 'FObjectProperty',
       name: 'data'
     },
+    {
+      class: 'FObjectProperty',
+      name: 'workingData',
+      expression: function(data) {
+        return data.clone(this.__subContext__)
```
mcarcaso: Ditch the __subContext__

src/foam/comics/v2/DAOUpdateView.js
```diff
@@ -83,12 +90,12 @@ foam.CLASS({
       name: 'save',
       code: function() {
         var self = this;
-        this.config.dao.put(this.data.clone()).then(function() {
+        this.data.copyFrom(this.workingData);
+        this.config.dao.put(this.data).then(function() {
```
mcarcaso: I'd do similar here as mentioned above:
```
this.config.dao.put(this.workingData).then(o => {
  this.data = o;
  this.finished.pub();
  this.stack.back();
}, e => {
  this.throwError.pub(e);
})
```

src/foam/comics/v2/DAOBrowseControllerView.js
tharmaman: Remembered I was facing a weird bug before where the config of wasn't working, gunna retest to see if it works now

src/files.js
tharmaman: Removing

src/foam/comics/v2/DAOBrowseControllerView.js
tharmaman: It's chill we don't even use of in DAOCreateView because we crated it in data my bad

tharmaman: Ah nvm, we needed to pass in an **of** property otherwise the facet won't trigger and render the AccountDAOCreateView

src/foam/comics/v2/DAOCreateView.js
tharmaman: Removed createdBy,

But need to set owner otherwise I get the following error thrown upon creating an account:
"You do not have permission to create an account for another user."

I'm assuming the default/garbage value is 0 and so it doesn't get automatically set.

mcarcaso: You don't want that in this generic DAOCreateView though since it applies to any kind of object. We'll chat offline about this.

src/foam/comics/v2/DAOBrowseControllerView.js
mcarcaso: Oh good call I overlooked that. Ok let's roll with this.

src/foam/comics/v2/DAOCreateView.js
tharmaman: Done.

src/foam/comics/v2/DAOSummaryView.js
tharmaman: Done.

src/foam/comics/v2/DAOUpdateView.js
tharmaman: Done just left it as
`data.clone(this)`

src/foam/comics/v2/DAOCreateView.js
tharmaman: Done. Removed it from the generic DAOCreateView and overrided the create function in the AccountDAOCreateView to add it in.

src/foam/comics/v2/DAOUpdateView.js
tharmaman: Done. Did the same for DAOCreateView.

mcarcaso (approved):

PR #2305, opened 2019-06-18
Pm stack trace

michaelconnor468:


src/foam/nanos/pm/PM.js
```diff
@@ -60,12 +63,28 @@ foam.CLASS({
...
-    PMLogger pmLogger = (PMLogger) x.get(DAOPMLogger.SERVICE_NAME);
-    if ( pmLogger != null ) {
-      pmLogger.log(this);
-    }
+      setEndTime(new java.util.Date());
+      if ( x == null ) return;
+      PMLogger pmLogger = (PMLogger) x.get(DAOPMLogger.SERVICE_NAME);
+      DAO pmInfoDAO = (DAO) x.get("pmInfoDAO");
+      if ( pmInfoDAO != null ) {
+        java.util.List pmInfoList = ( (ArraySink) (pmInfoDAO.where(foam.mlang.MLang.EQ(PMInfo.NAME, this.getName())).select(new ArraySink())) ).getArray();
```
jlhughes: @kgrgreer This seems expensive on each PM.

kgrgreer: This shouldn't be done in the PM, it should be done in the PMLogger. It already has access to PMInfo so can avoid duplicating all of this work.

kgrgreer: In DAOPMLogger in around line 64.

PR #2306, opened 2019-06-18
Table multi select support

TW80000:
## Notes
* Adds support to `UnstyledTableView` to let the user select multiple rows. It can be turned on via a flag property named `multiSelectEnabled` and when it's on, the `selectedObjects` property is updated to reflect the selections the user has made.
* There's a checkbox in the header that selects all rows when checked and deselects all rows when unchecked.
* Supports the ScrollTableView by making sure checkboxes know their state even when they're loaded in and out of the DOM as the user scrolls.
* It isn't integrated with the DAO controller yet (ie: you can't do anything once you've made selections), but I wanted to put this up for review before I start integrating this with other things.

## Screenshot with some rows selected
Screen Shot 2019-06-18 at 3 26 12 PM

### DIFF
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -154,12 +154,16 @@ foam.CLASS({
     },
     {
       name: 'idsOfObjectsTheUserHasInteractedWith_',
-      value: new Set()
+      factory: function() {
+        return {};
+      }
     },
     {
       name: 'checkboxes_',
       documentation: 'The checkbox elements when multi-select support is enabled. Used internally to implement the select all feature.',
-      value: new Map()
+      factory: function() {
+        return {};
+      }
     },
     {
       class: 'Boolean',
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -136,6 +138,38 @@ foam.CLASS({
...
+      class: 'Map',
+      name: 'selectedObjects',
+      documentation: `
+        The objects selected by the user when multi-select support is enabled.
+        It's a map where the key is the object id and the value is the object.
+      `
+    },
+    {
+      name: 'idsOfObjectsTheUserHasInteractedWith_',
+      value: new Set()
```
TW80000: make factory

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -136,6 +138,38 @@ foam.CLASS({
...
+    },
+    {
+      name: 'checkboxes_',
+      documentation: 'The checkbox elements when multi-select support is enabled. Used internally to implement the select all feature.',
+      value: new Map()
```
TW80000: make factory

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -173,6 +207,42 @@ foam.CLASS({
...
+                end();
+
+                // Set up a listener so we can update the existing CheckBox
+                // views when a user wants to select all or select none.
+                view.onDetach(slot.value.dot('data').sub(function(_, __, ___, newValueSlot) {
+                  var checked = newValueSlot.get();
+                  view.allCheckBoxesEnabled_ = checked;
+
+                  if ( checked ) {
+                    view.data.select().then(function(arraySink) {
```
TW80000: Just pass function to select() directly.

PR #2307, opened 2019-06-18
Update json outputter

KateMengjunXia:
nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/7209


src/foam/lib/PropertyPredicate.js
```diff
@@ -0,0 +1,26 @@
...
+    {
+      name: 'propertyPredicateCheck',
+      type: 'Boolean',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'fo',
```
kgrgreer: This field should be documented. Do we need to old and new objects? When outputting deltas, we would need to know what value was previously stored in order to know if we need to output this value or not. For example, for arrays we need to know the previous value to know if we can not output default values or not. Or, should this be handled in some other way?

KateMengjunXia: I think comparing old and new is handled by whoever calls propertyPredicateCheck. For example, in the json's Outputter.java:

protected boolean maybeOutputPropertyDelta(FObject oldFObject, FObject newFObject, PropertyInfo prop) {
    if ( propertyPredicate_ != null && ! propertyPredicate_.propertyPredicateCheck(this.x_, newFObject, prop) ) {
      return false;
    }

    return prop.compare(oldFObject, newFObject) != 0;
 }

src/foam/lib/StoragePropertyPredicate.js
```diff
@@ -0,0 +1,18 @@
...
+  methods: [
+    {
+      name: 'propertyPredicateCheck',
+      javaCode: `
+return ! prop.getStorageTransient();
+`
+    }
+  ]
+});
+
```
adamvy: Clean up the whitespace here.

src/foam/lib/PropertyPredicate.js
adamvy: But we could roll the old vs new check into the predicate right?  Adding a predicate for ValueUpdated.  Then a delta JSON outputter could just be implemented as `AND(ValueChanged, Permissioned, StorageTransient)` rather than having a separate stringifyDelta method.

Would be a cleaner interface.  We should also put a better name for this like "obj" instead of "fo"

adamvy (approved):
LGTM

PR #2310, opened 2019-06-19
checking currency class and outputting format

Nauna:


Nauna:
This is a quick fix. Working on better solution. But clients want this asap. Plus have Kevin's ok for this temp fix.

PR #2313, opened 2019-06-20
DO NOT MERGE: ExpressionSlot fallback to context

mcarcaso:
Allow ExpressionSlot to fallback onto the context it was created in and make the Section's createIsAvailableFor's ExpressionSlot use the context that's passed in. With this, the isAvailable for SectionAxioms can use things like controllerMode in their expression.

### DIFF
--- a/src/foam/core/Slot.js
+++ b/src/foam/core/Slot.js
@@ -420,7 +420,12 @@ foam.CLASS({
 
         var args = foam.Function.argNames(this.code);
         for ( var i = 0 ; i < args.length ; i++ ) {
-          args[i] = obj.slot(args[i]);
+          try {
+            args[i] = obj.slot(args[i]);
+          } catch {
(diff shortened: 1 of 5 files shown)


src/foam/core/Slot.js
```diff
@@ -420,7 +420,12 @@ foam.CLASS({
 
         var args = foam.Function.argNames(this.code);
         for ( var i = 0 ; i < args.length ; i++ ) {
-          args[i] = obj.slot(args[i]);
+          try {
+            args[i] = obj.slot(args[i]);
+          } catch {
+            args[i] = this.__context__[args[i]+'$'] ||
+              foam.core.ConstantSlot.create();
```
mcarcaso, line 427: Normally, if a slot isn't found on the object there's an error. This will now swallow that error allowing for typos easier.

mcarcaso:
Chatted about this and we both agree it's not a good avenue to go down.

PR #2316, opened 2019-06-20
Cpf 2066/performance troubshooting

smirnovaka:


src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -127,26 +127,28 @@ foam.CLASS({
...
       }
     }
     propertyMap_.put(of, properties);
   }
+    
+  List properties = propertyMap_.get(of);
+  Iterator e = properties.iterator();
+  while ( e.hasNext() ) {
+    PropertyInfo axiom = (PropertyInfo) e.next();
+    axiom.clear(obj);
```
TW80000, line 150: I don't think this makes sense. We don't just want to clear every permissioned property every time, we only want to clear it if they don't have the appropriate permission.

src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -127,26 +127,28 @@ foam.CLASS({
...
     List<PropertyInfo> properties = new ArrayList<>();
     List list = oldObj.getClassInfo().getAxiomsByClass(PropertyInfo.class);
     Iterator e = list.iterator();
     while ( e.hasNext() ) {
       PropertyInfo axiom = (PropertyInfo) e.next();
       if ( axiom.getPermissionRequired() ) {
-        properties.add(axiom);
-        maybeRemove(axiom, of,auth, x, obj, oldObj);
+        boolean authCheck = auth.check(x, of + ".rw." + axiom.getName().toLowerCase()) || auth.check(x, of + ".ro." + axiom.getName().toLowerCase());
+        if ( ! authCheck ) {
```
TW80000, line 138: This doesn't make sense. You're only doing an auth check on the first user and putting the axioms in the map based on whether that particular user has permission or not. It needs to do the auth check every time because different users will have different permissions.

src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -260,8 +227,12 @@ foam.CLASS({
     {
       name: 'put',
       javaCode: `
-        FObject fo = ((FObject) obj).fclone();
-        getDelegate().put(dao.maybeRemoveProperties(getX(), fo), sub);
+  FObject oldObj = dao.getDelegate().find(((FObject) obj).getProperty("id"));
```
TW80000, line 230: This regresses back to a find on every put to the sink. @kgrgreer fixed this yesterday.

TW80000 (changes requested):

src/foam/nanos/auth/PermissionedPropertyDAO.js
kgrgreer: Yes, keep the current code. for this method.

smirnovaka: sorry my bad, copy paste error.

src/foam/nanos/auth/PermissionedPropertyDAO.js
kgrgreer: Right idea, but in wrong place. Move to HidePropertiesSink so that it is only cached/reused for a single select(), and doesn't leak between users.

PR #2320, opened 2019-06-20
Add async action agency

smirnovaka:


src/foam/core/CompoundContextAgency.js
```diff
@@ -32,10 +33,12 @@ foam.CLASS({
         { name: 'x', type: 'X' }
       ],
       javaCode: `CompoundException e = new CompoundException();
+Logger logger = (Logger) x.get("logger");
 for ( Runnable agent : getAgents() ) {
   try {
     agent.run();
   } catch (Throwable t) {
+    logger.info(t.getMessage());
```
jlhughes: This should at least be a warning.

src/foam/nanos/ruler/AsyncAgency.java
```diff
@@ -0,0 +1,19 @@
+package foam.nanos.ruler;
```
kgrgreer: Add copyright.

src/foam/nanos/ruler/AsyncAgency.java
```diff
@@ -0,0 +1,19 @@
...
+
+import foam.core.Agency;
+import foam.core.ContextAgent;
+import foam.core.X;
+
+/**
+ * AsyncAgency is passed to async rule.action
+ * for immediate execution of agent.
+ */
+public class AsyncAgency implements Agency {
```
kgrgreer: I would name this class after what it does, rather than where you're currently using it. So, something like: DirectAgency, because it runs it directly.

See: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html

PR #2321, opened 2019-06-21
change PermissionedPropertyDAO.maybeRemove

smirnovaka:


src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -271,6 +244,9 @@ foam.CLASS({
       buildJavaClass: function(cls) {
         cls.extras.push(`
   private PermissionedPropertyDAO dao;
+  
+  /** map of properties of a model that require model.permission.property for read / write operations **/
+  protected Map<String, List<PropertyInfo>> propertyMap_ = new HashMap<>();
```
kgrgreer, line 249: Remove this.

src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -43,7 +43,7 @@ foam.CLASS({
   FObject oldObj = getDelegate().find_(x, id);
 
   if ( oldObj != null ) {
-    return maybeRemoveProperties(x, oldObj);
+    return maybeRemoveProperties(x, oldObj, propertyMap_);
```
kgrgreer: return maybeRemoveProperties(x, oldObj, new HashMap());

src/foam/nanos/auth/PermissionedPropertyDAO.js
kgrgreer: Keep this one, remove the other one.

src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -120,35 +120,39 @@ foam.CLASS({
...
+      FObject obj = oldObj.fclone();
+      AuthService auth = (AuthService) x.get("auth");
+      if ( ! propMap.containsKey(of) ) {
+        List<PropertyInfo> properties = new ArrayList<>();
+        List list = oldObj.getClassInfo().getAxiomsByClass(PropertyInfo.class);
+        Iterator e = list.iterator();
+        while ( e.hasNext() ) {
+          PropertyInfo axiom = (PropertyInfo) e.next();
+          if ( axiom.getPermissionRequired() ) {
+            boolean authCheck = auth.check(x, of + ".rw." + axiom.getName().toLowerCase()) || auth.check(x, of + ".ro." + axiom.getName().toLowerCase());
```
TW80000, line 140: This doesn't make sense. The way it was before was correct. We want to store the list of properties that require permission checks in the map. All we need to do is check if the property requires a permission check. Here you're actually checking if the user (who could be anyone) in the context has the permission and only adding it if they don't have either, then caching that result for everyone else, even though other users will have different permissions.

TW80000: Never mind, I see how it works now. Since we're providing a fresh HashMap every time `find` is called, we're not actually caching anything, so it works. And when we call it from HidePropertiesSink, we're using a Map that's local to that instance of the sink. And since we create a new HidePropertiesSink on every select, it works as expected and saves us from doing auth checks for every object that gets put to the sink. Instead, we only have to do the auth checks once and they'll be cached for the other calls to HidePropertiesSink.put during the select. Sorry for the confusion.

smirnovaka: The way it was before was so that on select for a single user we would be checking the same permissions and same properties for every single object even though we already just did the check.
I am passing a new map for each select/find so there is no global caching for all users as you say. Each individual map contains `of` of an object and properties that current user doesn't have permissions to see so that we don't have to do any duplicate auth checks and simply clear axioms that are not permitted. Does it not make sense?

PR #2323, opened 2019-06-21
Add text key for 2FA

Henry0422:
I would prefer to let `generateKey` method return an array which can contain both QR and text-key.

GoogleTOTPAuthService extends AbstractTOTPAuthService
AbstractTOTPAuthService extends AbstractOTPAuthService
AbstractOTPAuthService implements OTPAuthService
OTPAuthService is an interface.

### DIFF
--- a/src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
+++ b/src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
@@ -56,9 +56,6 @@ foam.CLASS({
           x.get("user")) ;
         DAO userDAO = (DAO) x.get("localUserDAO");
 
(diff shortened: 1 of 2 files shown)


adamvy:
This is an incomplete change.  You have left behind the "generateQrCode" argument in the interface, but it is no longer being respected.

I don't think like this idea in general.  I think generateKey should always just return the key, and a separate utility can be called to convert that to a QR code if required.  Returning a pair of strings is unintuitive.  Perhaps you could model an OTPKey as an object which has "key" and "qr-code" as properties.

Henry0422:
New PR: https://github.com/foam-framework/foam2/pull/2361

PR #2325, opened 2019-06-22
DO NOT MERGE: Virtual properties experiment

mcarcaso:
Create a Person which contains a property on it that has the behaviour of what, I think, we want a virtual property to have.

Note that is just to get the conversation around virtual properties started.


src/foam/dao/MDAO.java
```diff
@@ -80,11 +80,15 @@ synchronized void setState(Object state) {
   }
 
   public FObject objIn(FObject obj) {
-    return obj.fclone().freeze();
+    FObject o = obj.fclone();
```
mcarcaso, line 83: Because the javaGetter wants to use getX(), we can't lose the context on object in and out of the MDAO. This feels bad but I'm not sure what other path to take.

src/foam/demos/virtualprop/Person.js
```diff
@@ -0,0 +1,55 @@
...
+        this.country_Name = undefined;
+      }
+    }
+  ],
+  axioms: [
+    {
+      name: 'javaExtras',
+      buildJavaClass: function(cls) {
+        country_NameIsSet_ = true;
+        cls.getField('country_NameIsSet_').initializer = 'true';
```
mcarcaso, line 51: The JSON outputter won't send properties that aren't set and I'm not sure any other ways of making it think something is set other than removing the `storageTransient` on the property but then we're storing a value that's never actually used.

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -57,7 +57,7 @@ foam.CLASS({
         return foam.u2.view.FnFormatter.create({
           class: 'foam.u2.view.FnFormatter',
           f: function(value, obj, axiom) {
-            this.add(value);
+            this.add(obj[axiom.name + '$']);
```
mcarcaso, line 60: I noticed that properties don't change in the tableview when they change and this fixes it. Not sure if we'd want to do this. It'll cause lots of bindings.

TW80000:
I've been thinking a lot about the problem I think you're trying to solve here lately. Often we have models that reference a user and when we display a table of that model we want to see the user's name, not the id. Or other similar situations. At least I think that's the problem you're trying to solve here.

We want to be able to sort by the virtual property and apply mlangs to them as well. I don't know if having an expression that returns a promise will work on client-side MDAOs since mlangs aren't async-aware afaik. Just a part of the problem to consider.

Another part of the problem I've been thinking about is whether we "pull" from the source value via a `find` every time or whether we "push" to the virtual property with a `put` every time the source value changes. I think the first approach is more performant when we write to the model with the source value more than we read the model with the virtual property, but the second approach is faster when we read the model with the virtual property more than we write model with the source value. Just another thing to consider.

The idea seems similar to "views" in databases, where there's a "virtual table" that doesn't actually store the data but you can query it like real tables (although modifying data is usually only possible under certain restrictions). I was actually thinking about this last night and wanted to know how other databases handle views. I found [documentation](https://www.postgresql.org/docs/current/rules.html) on how PostgreSQL handles it. In summary, they have what they call a "rule system" that takes a query tree from the SQL parser and returns a new, possibly-modified version of the query tree where references to views in the query tree are replaced with queries on the "real" source tables that the view is a view of. Then it hands that off to the query optimizer, planner, and executer to run the query. I'm not sure if we could do something similar since we don't have virtual models/tables but virtual properties on real models. Also, I don't think we have the same fine-grained control as SQL for selecting only a subset of the columns (properties) of a model. We always just return the whole model via a find or select.

Anyway, just noticed this in the list of PRs and thought it sounded similar to what I've been thinking about so I wanted to write down my thoughts.

jlhughes:
Closed - abandoned for now.

PR #2326, opened 2019-06-24
add StringBufferLogger

hchoura:
Logger logger = new ProxyLogger(new StringBufferLogger());
root_.put("logger", logger);
logger.info("Registering:", "Test");

result
2019.06.24.19.45.41.112,INFO,Registering:,Test

### DIFF
--- a/src/foam/nanos/logger/StringBufferLogger.java
+++ b/src/foam/nanos/logger/StringBufferLogger.java
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
 package foam.nanos.logger;
 
 import java.util.logging.ConsoleHandler;
--- a/src/foam/nanos/logger/StringBufferLogger.java
+++ b/src/foam/nanos/logger/StringBufferLogger.java
@@ -1,0 +1,80 @@
+package foam.nanos.logger;
+
+import java.util.logging.ConsoleHandler;
+import java.util.logging.Formatter;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+public class StringBufferLogger extends AbstractLogger {
+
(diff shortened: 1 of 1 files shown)


jlhughes:
Is this used to register our nanos logging as a handler for java logging?

src/foam/nanos/logger/StringBufferLogger.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class StringBufferLogger extends AbstractLogger {
+
+  protected java.util.logging.Logger logger_; 
+  
+  public StringBufferLogger() {
+    logger_ = java.util.logging.Logger.getAnonymousLogger();
+    logger_.setUseParentHandlers(false);
+    logger_.setLevel(Level.ALL);
+    Handler handler = new ConsoleHandler();
```
kgrgreer, line 23: Does this mean that the output is also going to the console? Do we need to use java.util.Logger? Does it provide any value in this case?

kgrgreer:
The intent of this class is to capture log output for tests/scripts/crons/etc. so that we can include the output in their output sections.

jlhughes:
This can also be used to capture logs before the file logger is in place, then we can dump this to the file logger.

src/foam/nanos/logger/StringBufferLogger.java
```diff
@@ -0,0 +1,80 @@
+package foam.nanos.logger;
```
kgrgreer, line 7: Add copyright header.

src/foam/nanos/logger/StringBufferLogger.java
hchoura: Does this mean that the output is also going to the console? yes
Do we need to use java.util.Logger? yes, I follow the same principle as StdoutLogger
Does it provide any value in this case?

kgrgreer: I think the use case for this is capturing logging from Scripts. Is that correct? If so, I don't think we also want the output to go to the stdout.

PR #2327, opened 2019-06-24
Set session created time properly

Henry0422:
https://nanopay.atlassian.net/browse/CPF-1678

### DIFF
--- a/src/foam/box/SessionServerBox.java
+++ b/src/foam/box/SessionServerBox.java
@@ -18,6 +18,7 @@ import foam.nanos.logger.*;
 import foam.nanos.logger.PrefixLogger;
 import foam.nanos.session.Session;
 import foam.util.SafetyUtil;
+import java.util.Date;
 import javax.naming.NoPermissionException;
 import javax.servlet.http.HttpServletRequest;
 import org.eclipse.jetty.server.Request;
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

adamvy (approved):

src/foam/box/SessionServerBox.java
```diff
@@ -53,6 +54,7 @@ public void send(Message msg) {
         session = new Session();
         session.setId(sessionID == null ? "anonymous" : sessionID);
         session.setRemoteHost(req.getRemoteHost());
+        session.setCreated(new Date());
```
kgrgreer, line 57: Why is this necessary? Here is the definition of the created property:
    {
      class: 'DateTime',
      name: 'created',
      factory: function() { return new Date(); },
      javaFactory: 'return new Date();'
    },

which should do this already. Do we have a bug in FOAM?

adamvy: Factories are not triggered at object creation time, but at first access, so this will return the time that the created property is first accessed.

kgrgreer:
Oh, yah. So we could also just do:
session.getCreated();

adamvy:
Indeed.  I'm good with either personally.  .getCreated() is shorter, .setCreated(new Date()) is clearer.

PR #2328, opened 2019-06-24
Cards and Card Layout

tharmaman:


src/foam/u2/layout/Card.js
```diff
@@ -0,0 +1,52 @@
...
+
+foam.CLASS({
+  package: 'foam.u2.layout',
+  name: 'Card',
+  extends: 'foam.u2.borders.CardBorder',
+
+  properties: [
+    {
+      class: 'Int',
+      name: 'def',
```
mcarcaso: Can you make these a little more descriptive and/or add documentation?
cols - this is the default value for all sizes
mdCols
lgCols
smCols
etc...

src/foam/u2/layout/Card.js
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.SUPER();
+      this.addClass(this.myClass());
+
+      // TODO: find the display mode and adjust accordingly (e.g. for sm divide by 8), for now just use md
```
mcarcaso: Let's put some of this stuff in place now. Import 'displayWidth?' and make decisions based on that. Create an enum that contains the widths we'll be caring about and assume displayWidth is one of those, or give it a default if it's undefined. Make sure this initE is bound to the displayWidth because it can/will change.

src/foam/u2/layout/Cards.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.SUPER();
+      this.addClass(this.myClass());
+    },
+
+    function start(spec, args, slot) {
```
mcarcaso: Let's avoid doing this and just give Card a default margin and anybody rendering them that wants to override the margin can do it via u2. If we find ourselves doing this frequently we'll find a way to conveniently do that but until then I'd rather keep the configuration we expose here to a minimum

src/foam/u2/layout/Card.js
adamvy: Yeah I think we shouldn't use such short names, just use full words, it would be more consistent with the rest of foam.

tharmaman: Done.

src/foam/u2/layout/Card.js
tharmaman: Done.

src/foam/u2/layout/Cards.js
tharmaman: Done.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -137,6 +138,13 @@ foam.CLASS({
   `,
 
   properties: [
+    {
+      class: 'Int',
```
mcarcaso: I think this should be an Enum of the DisplayWidth that you just created. Your factory can use the window.innerWidth to compute the DisplayWidth that should be used.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -240,6 +248,8 @@ foam.CLASS({
     },
 
     function initE() {
+      window.addEventListener('resize', () => this.displayWidth = window.innerWidth);
```
mcarcaso: Do this instead:
```
window.addEventListener('resize', () => this.displayWidth = undefined);
```
This will clear the property which will cause your factory to re-run so you don't need to duplicate the logic of that factory here.

src/foam/u2/layout/Card.js
```diff
@@ -0,0 +1,140 @@
...
+      documentation: `
+        Sets up a standard default column width across all display types
+      `,
+      value: 12
+    },
+    {
+      class: 'Int',
+      name: 'xxsCols',
+      documentation: `
+        The column width for the smaller end of smartphone devices and smartphone portrait screens:
```
mcarcaso: I think the majority of the documentation on these properties are better off on the enum values. I wouldn't call out pixel values in this model. It only cares about DisplayWidths.

src/foam/u2/layout/Card.js
```diff
@@ -0,0 +1,140 @@
...
+        Max-width @ 320px and an 8 column grid
+      `,
+      expression: function(cols) {
+        return cols > 8 ? 8 : cols;
```
mcarcaso: Could do `return Math.min(8, cols);` instead here and below.

src/foam/u2/layout/Card.js
```diff
@@ -0,0 +1,140 @@
...
+
+  methods: [
+    function initE() {
+      this.SUPER();
+      this.addClass(this.myClass());
+
+      // set the default value the xl if display width doesn't exist
+      let flex;
+
+      if ( this.displayWidth ) {
```
mcarcaso: displayWidth can change so you want to be bound to it.
Also, you can enumerate enums so this kinda thing could be reduced a little by using array functions on the enum values but I think there's a better way to reduce it:
```
this.style({
  'flex-grow': this.displayWidth$.map(dw => {
    dw = dw || foam.u2.layout.DisplayWidth.XL;
    flex = this[dw.name.toLowerCase() + 'Cols'] / dw.cols;
  })
})
```
That said, there's something that's not quite sitting well with me here because I'm worried we're going to have to add a bunch of properties to every view that cares about the display width and that can be kind of annoying but maybe we won't actually be doing that as frequently as I think. Let's keep an eye out for us needing to do this frequently and if it becomes a problem we'll maybe turn this sort of thing into a map or maybe even a more formal modelled object.

src/foam/u2/layout/DisplayWidth.js
```diff
@@ -0,0 +1,23 @@
+foam.ENUM({
+  package: 'foam.u2.layout',
+  name: 'DisplayWidth',
+  documentation: `
+      An enum of the responsive display width types
+  `,
+
+  values: [
+      { name: 'XXS', label: 1440, cols: 8 },
```
mcarcaso: You shouldn't use `label` for these values. `label` is typically used when we're rendering the enum in a dropdown and I think the default (which is just name) would be good enough for that value. Change the name of these properties to something like `minWidth` or `maxWidth`. I think I like `minWidth` better because then `XS` would just have a value of `0`. If you do `maxWidth`, `XL` would need a value of `Number.MAX_SAFE_INTEGER` or something which looks a little weird IMO.

src/foam/nanos/controller/ApplicationController.js
tharmaman: Done.

src/foam/u2/layout/Card.js
tharmaman: Done.

src/foam/u2/layout/DisplayWidth.js
tharmaman: Done.

src/foam/nanos/controller/ApplicationController.js
tharmaman: Done.

src/foam/u2/layout/Card.js
tharmaman: Done.

src/foam/u2/layout/Card.js
tharmaman: Done but factory doesn't fire-off upon resizing

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -137,6 +138,33 @@ foam.CLASS({
   `,
 
   properties: [
+    {
+      class: 'Enum',
+      of: 'foam.u2.layout.DisplayWidth',
+      name: 'displayWidth',
+      factory: function() {
+        debugger;
```
adamvy: Clean up debugger statement.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -137,6 +138,33 @@ foam.CLASS({
   `,
 
   properties: [
+    {
+      class: 'Enum',
+      of: 'foam.u2.layout.DisplayWidth',
+      name: 'displayWidth',
```
adamvy, line 144: Let's make this a property and/or function on the Window object itself.  foam.core.Window, or a refinement added in foam.u2.Element.

src/foam/u2/layout/Card.js
```diff
@@ -0,0 +1,110 @@
...
+  css: `
+    ^ {
+      margin: 8px;
+    }
+  `,
+
+  properties: [
+    {
+      class: 'Int',
+      name: 'cols',
```
adamvy: I personally would say "columns" instead of "cols"  I lean towards longer names :)

Possibly even "smallColumns" "xSmallColumns" "xxSmallColumns"

src/foam/u2/layout/Card.js
```diff
@@ -0,0 +1,110 @@
...
+      // need to bind the flex value to displayWidth because it can change
+      this.style({
+        'flex-grow': this.displayWidth$.map(dw => {
+          dw = dw || foam.u2.layout.DisplayWidth.XL;
+          return this[`${dw.name.toLowerCase()}Cols`] / dw.cols;
+        })
+      })
+    }
+  ]
+});
```
adamvy: Newline at end of file.

adamvy (approved):

src/foam/nanos/controller/ApplicationController.js
tharmaman: Done.

src/foam/u2/layout/Card.js
tharmaman: Done.

src/foam/nanos/controller/ApplicationController.js
tharmaman: Fixed all the other stuff, just needs the window refactor like you said.

src/foam/u2/layout/Card.js
tharmaman: Done just appended in column because the enum uses xs, sm, etc.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -137,6 +138,12 @@ foam.CLASS({
...
+      value: foam.u2.layout.DisplayWidth.XL
```
mcarcaso, line 145: `value: 'XL'`

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -473,6 +483,17 @@ foam.CLASS({
         this.footerView_.removeAllChildren();
         this.footerView_.tag({ class: this.theme.footerView });
       }
-    }
+    },
+    {
+      name: 'updateDisplayWidth',
+      isMerged: true,
+      mergeDelay: 5000,
```
mcarcaso, line 490: let's reduce this delay a little. we made it big for testing to ensure it wasn't the cause of the slow resizing and it isn't. maybe reduce to 1000? or put it back to isFramed

src/foam/u2/layout/Card.js
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.SUPER();
+      this.addClass(this.myClass());
+
+      // need to bind the flex value to displayWidth because it can change
```
mcarcaso, line 101: This comment can be removed now right?

mcarcaso (approved):

PR #2329, opened 2019-06-24
Multi line journals

michaelconnor468:
added multi line journal capabilities to filejournal reading logic

### DIFF
--- a/src/foam/dao/FileJournal.js
+++ b/src/foam/dao/FileJournal.js
@@ -290,32 +290,26 @@ foam.CLASS({
       name: 'getCommand',
       documentation: 'reads a syntatically meaningful bluck from the service file',
       type: 'String',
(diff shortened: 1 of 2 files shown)


src/foam/dao/FileJournal.js
```diff
@@ -295,9 +300,21 @@ foam.CLASS({
             if ( COMMENT.matcher(line).matches() ) continue;
 
             try {
-              char operation = line.charAt(0);
-              int length = line.trim().length();
-              line = line.trim().substring(2, length - 1);
+              operation = line.charAt(0);
```
kgrgreer: A cleaner design would be to create a readCommand() method, which just calls reader.readLine() if multiline is disabled but reads multiple lines if it isn't.

PR #2331, opened 2019-06-25
Multi line j

michaelconnor468:


src/foam/dao/FileJournal.js
```diff
@@ -281,6 +285,39 @@ foam.CLASS({
...
+          name: 'reader',
+          type: 'BufferedReader'
+        }
+      ],
+      javaCode: `
+        try {
+          String line = reader.readLine();
+          if ( getMultiLine() ) {
+            StringBuilder sb = new StringBuilder();
+            line = line.trim();
```
adamvy: Do we want to trim every line?  That will remove indentation from scripts.

src/foam/dao/FileJournal.js
```diff
@@ -281,6 +285,39 @@ foam.CLASS({
...
+            while ( ! ( line.charAt( line.length() ) == ')' && line.charAt( line.length() - 1 ) == '}' ) ) {
```
adamvy: line.endsWith("});") is a lot more readable.

src/foam/dao/FileJournal.js
```diff
@@ -290,14 +327,14 @@ foam.CLASS({
         JSONParser parser = getParser();
 
         try ( BufferedReader reader = getReader() ) {
-          for ( String line ; ( line = reader.readLine() ) != null ; ) {
+          for ( String line ; ( line = getUnit(reader) ) != null ; ) {
```
adamvy: Maybe re-name `line` to `entry`

src/foam/dao/FileJournal.js
michaelconnor468: Sure, this helps us check what the lines end with. Also we are not using any indentation sensitive languages within our service scripts so it shouldn't cause issues with parsing.

adamvy: Its not syntactically sensitive, to the machine, but it is relevant to the developers reading and editing the scripts.  I don't want all scripts in the GUI to lose their indentation, that would be really hurt the usability.

src/foam/dao/FileJournal.js
```diff
@@ -281,27 +285,58 @@ foam.CLASS({
...
+          name: 'reader',
+          type: 'BufferedReader'
+        }
+      ],
+      javaCode: `
+        try {
+          String line = reader.readLine();
+          if ( getMultiLine() ) {
+            StringBuilder sb = new StringBuilder();
+            while ( ! line.endsWith("})") ) {
```
adamvy: I think we actually want to check line.equals("})"), this will help protect against lines in js scripts like "var person = Person.create({ name: "adam" })"

adamvy: A proper multiline journal entry shoudl read

```
p({"class":"foam.nanos.script.Script","code":"""doo
some

stuff""", "otherProperty":123
})
```

src/foam/dao/FileJournal.js
```diff
@@ -281,27 +285,58 @@ foam.CLASS({
...
+          name: 'reader',
+          type: 'BufferedReader'
+        }
+      ],
+      javaCode: `
+        try {
+          String line = reader.readLine();
+          if ( getMultiLine() ) {
+            StringBuilder sb = new StringBuilder();
+            while ( ! line.trim().equals("})") ) {
```
adamvy: I still wouldn't trim it, this will create a new String for every line we read

src/foam/dao/FileJournal.js
```diff
@@ -281,27 +285,58 @@ foam.CLASS({
...
+      ],
+      javaCode: `
+        try {
+          String line = reader.readLine();
+          if ( getMultiLine() ) {
+            StringBuilder sb = new StringBuilder();
+            while ( ! line.equals("})") ) {
+              sb.append(line);
+              line = reader.readLine();
+              if ( line == null )
```
adamvy, line 308: Surely this check needs to be earlier, we just called line.equals() but if line was null that would throw a NPE.

PR #2333, opened 2019-06-25
Add a rootMenu to Group and respect it in the TopNavigation.

mcarcaso:
Using this, you can set the rootMenu of a group and the children of that menu will take over the top nav.

### DIFF
--- a/src/foam/nanos/u2/navigation/TopNavigation.js
+++ b/src/foam/nanos/u2/navigation/TopNavigation.js
@@ -137,7 +137,7 @@ foam.CLASS({
         .start()
           .addClass('logged-in-container')
           .tag({ class: 'foam.nanos.u2.navigation.BusinessLogoView' })
-          .start(this.MenuBar, { menuName$: this.group$.dot('rootMenu') })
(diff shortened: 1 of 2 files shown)


src/foam/nanos/u2/navigation/TopNavigation.js
```diff
@@ -132,7 +137,7 @@ foam.CLASS({
         .start()
           .addClass('logged-in-container')
           .tag({ class: 'foam.nanos.u2.navigation.BusinessLogoView' })
-          .start({ class: 'foam.nanos.menu.MenuBar' })
+          .start(this.MenuBar, { menuName$: this.group$.dot('rootMenu') })
```
kgrgreer: Do we need to bind slots? Is the menu going to redraw if the group or group.rootMenu changes? And are they ever going to change?

mcarcaso: Just felt more robust this way in case group ever does change but the reality is that the MenuBar doesn't bind to the menuName so even if it does change, this won't update.

PR #2334, opened 2019-06-25
building Java csvSink

Nauna:
- This work allows the csvSink to be built and run on the server.
- Allows each property to have a javaToCSV/javaToCSVLabel  which is a string that will output specified formatting for the csv outputter(csvSink).
- clean up a few toCSV legacy formatting and overrides

goes hand in hand with this pr: https://github.com/nanoPayinc/NANOPAY/pull/7306


Nauna:
Current State of the world

image

src/foam/dao/CSVSink.js
```diff
@@ -44,13 +49,33 @@ foam.CLASS({
...
+      args: [
+        { name: 'value' }
+      ],
+      code: function(value) {
+        if ( ! this.isNewLine ) this.csv += ',';
+        this.isNewLine = false;
+        this.output_(value);
+      },
+      javaCode: `
+        StringBuilder sb = new StringBuilder();
```
mcarcaso: This isn't the right use of a StringBuilder. You'll want to create a property that you store the StringBuilder in and this method will be something like:
Under properties:
```
{
  name: 'sb',
  flags: ['java'],
  javaType: 'StringBuilder',
  javaFactory: 'return new StringBuilder();'
}
```
and this method turns into:
```
        if ( ! getIsNewLine() ) getSb().append(",");
        setIsNewLine(false);
        output_(value);
```
and then you can implement the sink's `eof` method in java:
```
{
  name: 'eof'
  javaCode: 'setCsv(getSb().toString());'
}
```
And in the other methods, don't create new string builders and just re-use getSb() to append.

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +101,97 @@ foam.CLASS({
...
           }, function(value) {
             this.output_(value.toString());
-          })
+        }),
+      javaCode: `
+        StringBuilder sb = new StringBuilder();
+        sb.append(getCsv());
+        String s = value.toString();
+
+        if (s.indexOf("\\"") != -1 || s.indexOf(",") != -1) {
```
mcarcaso: You might want to use logic similar to the code block above to keep js and java behavior more consistent with each other. You can't use an mmethod but you can use if/else similar to the java json outputter: https://github.com/foam-framework/foam2/blob/master/src/foam/lib/json/Outputter.java#L213

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +101,97 @@ foam.CLASS({
...
     },
+    {
+      name: 'put',
+      code: function(obj) {
+        if ( ! this.of ) this.of = obj.cls_;
 
-    function put(obj) {
-      if ( ! this.of ) this.of = obj.cls_;
+        if ( ! this.isHeadersOutput ) {
+          this.props.forEach((element) => {
```
mcarcaso: Did you break this? Aren't props a StringArray now?

src/foam/java/PropertyInfo.js
```diff
@@ -285,6 +287,20 @@ foam.CLASS({
               { name: 'obj', type: 'foam.core.FObject' }
             ],
             body: this.validateObj
+          },
+          {
+            name: 'toCSV',
+            visibility: 'public',
+            type: 'void',
+            args: [{ name: 'x', type: 'foam.core.X' }, { name: 'obj', type: 'Object' }, { name: 'outputter', type: 'foam.dao.CSVSink' }, { name: 'prop', type: 'Object' }],
```
mcarcaso: Use multiple lines for this so it's easier to read.

src/foam/java/PropertyInfo.js
```diff
@@ -285,6 +287,20 @@ foam.CLASS({
...
+            body: this.toCSV
+          },
+          {
+            name: 'toCSVLabel',
+            visibility: 'public',
+            type: 'void',
+            args: [{ name: 'outputter', type: 'foam.dao.CSVSink' }, { name: 'prop', type: 'Object' }],
```
mcarcaso: Here too

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -60,6 +60,7 @@ foam.CLASS({
     'appConfig',
     'as ctrl',
     'currentMenu',
+    'filteredTableColumns',
```
mcarcaso: You don't want this on the ApplicationController. This ties the ApplicationController to one set of filteredTableColumns but there's no reason we can't render multiple table views at once. I think troy's suggestion was to put this on `foam.comics.DAOController`

mcarcaso (changes requested):

src/foam/dao/CSVSink.js
Nauna: nice!

src/foam/dao/CSVSink.js
Nauna: :( ya i did. Fixed with next push

src/foam/comics/DAOController.js
```diff
@@ -253,16 +261,16 @@ foam.CLASS({
...
-        link.click();
-      });
+      this.csvDriver.exportDAO(this.__context__, data, this.filteredTableColumns)
+        .then(function(result) {
+          result = 'data:text/csv;charset=utf-8,' + result;
+          var encodedUri = encodeURI(result);
+          var link = document.createElement('a');
+          link.setAttribute('href', encodedUri);
+          link.setAttribute('download', 'data.csv');
+          document.body.appendChild(link);
```
adamvy, line 245: We should be cleaning up this link after the fact as well

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +106,99 @@ foam.CLASS({
...
     },
-
-    function newLine_() {
-      this.csv += '\n';
-      this.isNewLine = true;
+    {
+      name: 'newLine_',
+      code: function() {
+        this.csv += '\n';
+        this.isNewLine = true;
```
adamvy: I personally would have called this isFirst, but that's not major

adamvy: Or maybe isFirstColumn

src/foam/dao/CSVSink.js
```diff
@@ -76,34 +106,99 @@ foam.CLASS({
...
+      name: 'eof',
+      javaCode: 'setCsv(getSb().toString());'
+    },
+    {
+      name: 'put',
+      code: function(obj) {
+        if ( ! this.of ) this.of = obj.cls_;
+        var element = undefined;
+        if ( ! this.isHeadersOutput ) {
+          this.props.forEach((name) => {
```
adamvy: I would have done isFirstRow and put the header outputting into its own function.

`if ( this.isFirstRow ) this.outputHeaders`

Not a big deal

src/foam/lib/csv/Outputter.java
```diff
@@ -176,7 +176,7 @@ protected void outputFObject(FObject obj) {
 
     while ( i.hasNext() ) {
       PropertyInfo prop = (PropertyInfo) i.next();
-      prop.toCSV(this, prop.f(obj));
+      output(prop.f(obj));
```
adamvy: Why this change?  shouldn't the property's toCSV still be respected if it wants to control how it outputs itself to CSV?

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -33,7 +33,8 @@ foam.CLASS({
     'ctrl',
     'dblclick?',
     'editRecord?',
-    'selection? as importSelection'
+    'selection? as importSelection',
+    'filteredTableColumns'
```
adamvy: Make this an optional import

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -53,21 +54,28 @@ foam.CLASS({
...
-            of.getAxiomByName(p) :
-            p;
-
-           if ( ! c ) {
-             console.error('Unknown table column: ', p);
-           }
-
-          return c;
-        }).filter(function(c) { return c; });
+        this.filteredTableColumns = [];
```
adamvy: Build the array first as a local variable, and then set the imported value afterwards.  This is for two reasons.  1) Anybody listening on updates to filteredTableColumns would first get a property change even as its set to an empty array. and then 2) property change events don't fire when an array's contents change.

adamvy: Actually you'd be better off implementing this as a reaction.

```
foam.CLASS({
  name: 'UnstyledTableView',
...
  imports: [ 'filteredTableColumns?' ],
  reactions: [ '', 'propertyChange.columns_', 'onColumnsUpdate' ],
  listeners: [
    function onColumnsUpdate() {
      // First check if the optional import was provided.
      if ( this.filteredTableColumns$ ) {
        this.filteredTableColumns$ = this.columns_.map(p => p.name);
      }
    }
  ]
});
```

Then you don't need to modify the expression at all, and if somebody explicitly sets columns_ then filteredTableColumns will still be updated appropriately.

src/foam/lib/csv/Outputter.java
mcarcaso: The behaviour of toCSV has changed and isn't really compatible with this Outputter anymore so I'm guessing the safest thing to do was this. If I'm reading this correctly, this roughly preserves the previous behaviour with the exception of Enums. IMO, we should be aiming to nuke this outputter completely as it's replaced by the CSVSink.

adamvy: I see that toCSV now takes a CSVSink?  That doesn't sound right to me.  CSVSink should just internally use a CSVOutputter.  As it stands we're setting outselves up for confusion and possible bugs as people write javaToCSV for properties but then that doesn't actually get called when the object is output the CSV.

Nauna: So short of re-writing this outputter to function like the new outputter, this was the quick fix.

This I feel should be removed. But need to confirm that no one is using it.

The toCSV that this is expecting is not the same as what we have developed here.

And ultimately all this toCSV that was doing was 'output(prop.f(obj))'

Do you think it will suffice to just create a ticket to investigate its use and replace with the new outputter????

src/foam/u2/view/UnstyledTableView.js
Nauna: interesting

src/foam/lib/csv/Outputter.java
mcarcaso: Yeah the CSVSink is basically an Outputter. Let's do this:
- Rename CSVSink to CSVOutputter
- Rename put to outputFObject (so it's a similar interface to json Outputter)
- Rename reset to flush
- Implement toString which returns csv in js and and sb.toString in java

Then implement a CSVSink which creates an instance of an outputter and on put, calls outputFObject and on eof, does a csv = outputter.toString()

That sound ok?

mcarcaso: So CSVSink turns to something like this (note that there's no javaCode in here which you'll need to fill in):
```
foam.CLASS({
  package: 'foam.dao',
  name: 'CSVSink',
  extends: 'foam.dao.AbstractSink',

  documentation: 'Sink runs the csv outputter, and contains the resulting string in this.csv',

  properties: [
    {
      class: 'String',
      name: 'csv',
      view: 'foam.u2.tag.TextArea'
    },
    {
      class: 'Class',
      name: 'of',
      visibility: 'HIDDEN'
    },
    {
      name: 'props',
      visibility: 'HIDDEN'
    },
    {
      name: 'outputter',
      transient: true,
      factory: function(props) {
        return this.CSVOutputter.create({
          of: this.of,
          props: this.props
        })
      },
    }
  ],

  methods: [
    function put(obj) {
      this.outputter.outputFObject(obj);
    },

    function eof() {
      this.csv = this.outputter.toString();
    },

    function reset() {
      this.outputter.flush();
      this.csv = '';
    }
  ]
});
```

src/foam/dao/CSVSink.js
```diff
@@ -23,87 +24,58 @@ foam.CLASS({
...
+      of: 'foam.lib.csv.CSVOutputter',
+      transient: true,
+      factory: function() {
+        return foam.lib.csv.CSVOutputter.create({
+          of: this.of,
+          props: this.props
+        });
+      },
+      javaFactory: `
+        return new foam.lib.csv.CSVOutputter.Builder(getX()).setOf(getOf()).setProps(getProps()).build();
```
mcarcaso: Style suggestion: If you're setting more than one field, break into multiple lines:
```
return new foam.lib.csv.CSVOutputter.Builder(getX())
  .setOf(getOf())
  .setProps(getProps())
  .build();
```
Is much easier to read and maintain.

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+    'java.lang.String',
+    'java.util.Date'
+  ],
+
+  properties: [
+    {
+      class: 'String',
+      name: 'csv',
+      view: 'foam.u2.tag.TextArea',
+      documentation: 'Used for js version of this outputter.'
```
mcarcaso: If this is for js only then give it `flags: ['js']` and it won't be generated for java.

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+      name: 'of',
+      visibility: 'HIDDEN'
+    },
+    {
+      class: 'StringArray',
+      name: 'props',
+      factory: function() {
+        return this.of.getAxiomByName('tableColumns').columns;
+      },
+      visibility: 'HIDDEN'
```
mcarcaso: I'm not sure you should hide any of these properties. If I'm viewing the CSVOutputter from a DetailView then it's probably because I'm debugging and I'll want to see a lot of these fields.

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+        });
+        this.newLine_();
+        this.isFirstRow = false;
+      },
+      javaCode: `
+        Object propObj;
+        PropertyInfo columnProp;
+        String[] tableColumnNames = getProps();
+
+        for (String propName: tableColumnNames) {
```
mcarcaso: you can just use `getProps()` here and not declare a tableColumnNames above.

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+      },
+      javaCode: 'return getSb().toString();'
+    },
+    {
+      name: 'headerOutput',
+      args: [
+        { type: 'FObject', name: 'obj' }
+      ],
+      code: function() {
+        var element;
```
mcarcaso: Why is this declared up here? It's only used in the loop below.

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+        this.props.forEach((name) => {
+          element = this.of.getAxiomByName(name);
```
mcarcaso: Rename element to axiom or prop.

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+    },
+    {
+      name: 'toString',
+      code: function() {
+        return this.csv;
+      },
+      javaCode: 'return getSb().toString();'
+    },
+    {
+      name: 'headerOutput',
```
mcarcaso: rename this to outputHeader

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+        setIsFirstRow(false);
+      `
+    },
+    {
+      name: 'outputFObject',
+      args: [
+        { type: 'FObject', name: 'obj' }
+      ],
+      code: function(obj) {
+        var element;
```
mcarcaso: Don't declare up here

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+
+        if ( ! this.of ) this.of = obj.cls_;
+
+        if ( this.isFirstRow ) this.headerOutput();
+
+        this.props.forEach((name) => {
+          element = this.of.getAxiomByName(name);
```
mcarcaso: rename to axiom/prop

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+          element.toCSV(x, obj, this, element);
+        });
+
+        this.newLine_();
+      },
+      javaCode: `
+        Object propObj;
```
mcarcaso: Don't declare any of these variables up here

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+        PropertyInfo columnProp;
+        String[] tableColumnNames = getProps();
+
+        if ( ! isPropertySet("of") || getOf() == null ) setOf(obj.getClassInfo());
+
+        if ( getIsFirstRow() ) headerOutput(obj);
+
+        for (String propName : tableColumnNames) {
```
mcarcaso: just use getProps()

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+        });
+
+        this.newLine_();
+      },
+      javaCode: `
+        Object propObj;
+        PropertyInfo columnProp;
+        String[] tableColumnNames = getProps();
+
+        if ( ! isPropertySet("of") || getOf() == null ) setOf(obj.getClassInfo());
```
mcarcaso: Why do you need to `isPropertySet` check?

src/foam/core/PropertyInfo.java
```diff
@@ -43,7 +44,8 @@
   public Parser queryParser();
   public Parser csvParser();
   public void toJSON(foam.lib.json.Outputter outputter, Object value);
-  public void toCSV(foam.lib.csv.Outputter outputter, Object value);
+  public void toCSV(X x, Object obj, foam.lib.csv.CSVOutputter outputter, Object value);
```
mcarcaso: I think we can do this:
`public void toCSV(X x, FObject obj, foam.lib.csv.CSVOutputter outputter);`
And the toCSV method can figure out the value by doing an `Object value = f(obj);`
`PropertyInfos` implement `Expr` which has an `f` method on it that takes an object and gives you an object. In the case of a PropertyInfo, it takes an FObject and gives you the value of that property on the object.

src/foam/core/PropertyInfo.java
```diff
@@ -43,7 +44,8 @@
...
+  public void toCSVLabel(foam.lib.csv.CSVOutputter outputter, Object value);
```
mcarcaso: I don't think you need the `value` in this method at all do you?

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+        this.newLine_();
+        this.isFirstRow = false;
+      },
+      javaCode: `
+        Object propObj;
+        PropertyInfo columnProp;
+        String[] tableColumnNames = getProps();
+
+        for (String propName: tableColumnNames) {
+          propObj = obj.getProperty(propName);
```
mcarcaso: Since this code block is basically identical to the JS version, try to use the same variable names for the same object if you can. columnProp would be renamed to axiom or prop depending on what you renamed it to above.
propObj shouldn't be necessary.

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,227 @@
...
+        Object propObj;
+        PropertyInfo columnProp;
+        String[] tableColumnNames = getProps();
+
+        if ( ! isPropertySet("of") || getOf() == null ) setOf(obj.getClassInfo());
+
+        if ( getIsFirstRow() ) headerOutput(obj);
+
+        for (String propName : tableColumnNames) {
+          propObj = obj.getProperty(propName);
```
mcarcaso: Similar comment here as above. Ditch propObj and rename columnProp to axiom/prop.

src/foam/lib/csv/Outputter.java
mcarcaso: This change can change the behaviour of anything using this Outputter. Please make it a priority to remove this Outputter ASAP so nobody tries to fix it if/when something breaks. Anything using this outputter needs to be updated to use the new outputter.

src/foam/nanos/export/CSVDriver.js
```diff
@@ -10,30 +10,37 @@ foam.CLASS({
...
+      factory: function() {
+        return this.CSVOutputter.create();
+      }
     }
   ],
 
   methods: [
     function exportFObject(X, obj) {
-      return this.outputter.toCSV(obj);
+      // if obj coming in is not from a model with tableColumns,
```
mcarcaso: I don't think this comment adds much value.

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -53,21 +54,28 @@ foam.CLASS({
     },
     {
       name: 'columns_',
+      documentation: `Note: filteredTableColumns was a property set in DAOController
```
mcarcaso, line 57: This shouldn't be intertwined within this expression like this. Let's use a reaction instead. Add this to the model:
```
reactions: [
  ['', 'propertyChange.columns_', 'updateFilteredTableColumns']
],
listeners: [
  {
    name: 'updateFilteredTableColumns',
    code: function() {
      this.filteredTableColumns = this.columns_.map(a => a.name);
    }
  }
]
```

src/foam/core/PropertyInfo.java
Nauna: No we don't. Thx for the reminder

src/foam/u2/view/UnstyledTableView.js
Nauna: considering this should have been done weeks ago. Are you ok with me making a ticket and adjusting this after.?

mcarcaso: Why not just do it now? I pretty much gave you the code you need to use above. It might even be less work to just use that than to create a ticket and describe what needs to be done.

Nauna:
Test Cases:
- Dig: calls CSVSink and CSVOutputter
(To Test) ```1)  http://localhost:8080/service/dig?dao=accountDAO&cmd=select&format=csv```
```2)  http://localhost:8080/service/dig?dao=invoiceDAO&cmd=select&format=csv```
(Note on Test 2) accountDAO call will check defaults but testing invoiceDAO tests nested FObject properties.
- EFTTest : uses CSVUtil
- Export from localhost -> admin -> services -> transactionDAO : uses ExportDriver/ExportModal
- Ablii -> payable -> export as CSV: uses CSVDriver

File: src/foam/dao/CSVSink.js [hunk 31]
```diff
@@ -23,87 +32,88 @@ foam.CLASS({
       visibility: 'HIDDEN'
     },
     {
+      class: 'StringArray',
       name: 'props',
-      expression: function(of) {
-        return of.getAxiomsByClass(foam.core.Property)
-          .filter( (p) => ! p.networkTransient );
+      factory: function(of) {
+        if ( ! of ) return;
```
mcarcaso: Maybe return `[]`?

src/foam/dao/CSVSink.js
```diff
@@ -23,87 +32,88 @@ foam.CLASS({
...
-      visibility: 'HIDDEN'
-    },
-    {
-      class: 'Boolean',
-      name: 'isHeadersOutput',
-      visibility: 'HIDDEN'
+      javaFactory: `
+        if ( ! isPropertySet("of") ) {
+          propsIsSet_ = false;
+          return null;
```
mcarcaso: Same here. Rather than marking it as unset just return an empty array? I'm hesitant to mess with the propIsSet_ cuz that's managed by the model. Also, I could be wrong but I think this will get clobbered right after the factory runs and set to true.

src/foam/dao/CSVSink.js
```diff
@@ -23,87 +32,88 @@ foam.CLASS({
...
     {
-      class: 'Boolean',
-      name: 'isNewLine',
-      value: true,
-      visibility: 'HIDDEN'
+      name: 'outputter',
+      class: 'FObjectProperty',
+      of: 'foam.lib.csv.CSVOutputter',
+      transient: true,
+      factory: function(of, props) {
```
mcarcaso: These arguments won't actually be getting passed in since this is a factory. Ditch these from here and use this.of and this.props below

src/foam/dao/CSVSink.js
```diff
@@ -23,87 +32,88 @@ foam.CLASS({
...
+        var csvOutputter = foam.lib.csv.CSVOutputter.create();
+        if ( of ) {
```
mcarcaso: Is there any harm in removing these checks and just creating the CSVOutputter like:
```
this.CSVOutputter.create({
  of: this.of,
  props: this.props
})
```
?

src/foam/dao/CSVSink.js
```diff
@@ -23,87 +32,88 @@ foam.CLASS({
...
+          csvOutputter.of = of;
+        }
+        if ( props ) {
+          csvOutputter.props = props;
+        }
+        return csvOutputter;
+      },
+      javaFactory: `
+      foam.lib.csv.CSVOutputter csvOutputter = new foam.lib.csv.CSVOutputter.Builder(getX()).build();
+      if ( isPropertySet("of") ) {
```
mcarcaso: Do this without the conditional if you can too.
```
foam.lib.csv.CSVOutputter csvOutputter = new foam.lib.csv.CSVOutputter.Builder(getX())
  .setOf(getOf())
  .setProps(getProps())
  .build();
```

src/foam/dao/CSVSink.js
```diff
@@ -120,15 +130,15 @@ foam.CLASS({
     {
       name: 'toCSV',
       class: 'Function',
-      value: function(obj, outputter, prop) {
-        outputter.output(obj ? obj[prop.name] : null);
+      value: function(obj, outputter) {
```
mcarcaso: x as first arg?

src/foam/dao/CSVSink.js
```diff
@@ -120,15 +130,15 @@ foam.CLASS({
...
+        outputter.outputValue(obj ? obj[this.name] : null);
       }
     },
     {
       name: 'toCSVLabel',
       class: 'Function',
-      value: function(outputter, prop) {
-        outputter.output(prop.name);
+      value: function(x, outputter, obj) {
```
mcarcaso: I don't think you need/want obj here. The code would be:
`outputter.outputValue(this.name);`
(where `this` is the foam.core.Property)

src/foam/dao/CSVSink.js
```diff
@@ -147,36 +157,70 @@ foam.CLASS({
...
+      value: function(x, obj, outputter) {
+        if ( ! this.of ) {
+          outputter.outputValue(obj ? obj[this.name] : null);
           return;
         }
-        prop.of.getAxiomsByClass(foam.core.Property)
+        this.of.getAxiomsByClass(foam.core.Property)
           .forEach((axiom) => {
-            axiom.toCSV(obj ? obj[prop.name] : null, outputter, axiom);
+            axiom.toCSV(x, obj ? obj[this.name] : null, outputter, axiom);
```
mcarcaso: I think you want:
`axiom.toCSV.call(axiom, x, obj && obj[this.name], outputter)`

src/foam/dao/CSVSink.js
```diff
@@ -147,36 +157,70 @@ foam.CLASS({
...
           });
       }
     },
+    {
+      name: 'javaToCSV',
```
mcarcaso, line 157: Add these refinements in java/refinements.js

src/foam/dao/CSVSink.js
```diff
@@ -147,36 +157,70 @@ foam.CLASS({
...
+      class: 'String',
+      value: `
+        if ( get(obj) == null && ! (get(obj) instanceof foam.core.FObject) ) {
```
mcarcaso: This check isn't right. You should be doing the same check as js. If there's no `of` then you just output the object as is.

src/foam/dao/CSVSink.js
```diff
@@ -147,36 +157,70 @@ foam.CLASS({
...
+          outputter.outputValue(getName());
+          return;
+        }
+
+        java.util.List<foam.core.PropertyInfo> nestPropList = ((foam.core.FObject)get(obj)).getClassInfo().getAxiomsByClass(foam.core.PropertyInfo.class);
```
mcarcaso: Here too, you want to base it off the `of` of the FObjectProperty. Using the ClassInfo of the property being stored can lead to bugs if it's a subclass of whatever the `of` is.

src/foam/dao/CSVSink.js
```diff
@@ -147,36 +157,70 @@ foam.CLASS({
...
+        for ( foam.core.PropertyInfo axiom : nestPropList ) {
+          axiom.toCSV(x, get(obj), outputter);
+        }
+      `
+    },
     {
       name: 'toCSVLabel',
       class: 'Function',
-      value: function(outputter, prop) {
+      value: function(x, outputter, obj) {
```
mcarcaso: Shouldn't need `obj` here. Base decisions on this.of

src/foam/java/PropertyInfo.js
```diff
@@ -285,6 +287,28 @@ foam.CLASS({
...
+            body: this.toCSV
+          },
+          {
+            name: 'toCSVLabel',
+            visibility: 'public',
+            type: 'void',
+            args: [
+              { name: 'x',          type: 'foam.core.X' },
+              { name: 'outputter',  type: 'foam.lib.csv.CSVOutputterInterface' },
+              { name: 'obj',        type: 'Object' }
```
mcarcaso: You shouldn't need an obj here.

src/foam/lib/csv/CSVOutputter.js
```diff
@@ -0,0 +1,235 @@
...
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.lib.csv',
+  name: 'CSVOutputter',
+
+  implements: [
+    'foam.lib.csv.CSVOutputterInterface'
```
mcarcaso: I think the more common/consistent way to do this would be to have the interface called CSVOutputter and this model called CSVOutputterImpl

src/foam/lib/csv/FObjectCSVOutputterDecorator.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.lib.csv',
+  name: 'FObjectCSVOutputterDecorator',
```
mcarcaso: Create a ProxyCSVOutputter and make this extend that. This way, we can add methods to the interface and nothing else should need to get outputted.

mcarcaso: Also rename this to PrefixedValueCSVOutputter ?

src/foam/lib/csv/FObjectCSVOutputterDecorator.js
```diff
@@ -0,0 +1,41 @@
...
+  ],
+
+  properties: [
+    {
+      name: 'outputter',
+      class: 'FObjectProperty',
+      of: 'foam.lib.csv.CSVOutputterInterface'
+    },
+    {
+      name: 'preLabelString',
```
mcarcaso: Just call this `prefix`

src/foam/lib/csv/Outputter.java
```diff
@@ -1,235 +0,0 @@
-/**
- * @license
- * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
mcarcaso, line 3: Woohoo!

mcarcaso (approved):
More featured and less code. Beautiful :)

PR #2335, opened 2019-06-26
AuthService Refactoring

TW80000:
## Changes

* Removed the redundant `checkPermission` method on the `AuthService` interface.
* Moved several checks out of `UserAndGroupAuthService` and into a new decorator named `EnabledCheckAuthService` and put that decorator before `CachingAuthService` in the decorator chain. This was to prevent the cache from ignoring changes to a user's or group's `enabled` property.
* Removed the old AuthService caching file and associated test


src/foam/nanos/auth/EnabledCheckAuthService.js
```diff
@@ -0,0 +1,44 @@
+foam.CLASS({
```
kgrgreer, line 1: Add copyright header.

PR #2339, opened 2019-06-26
Multi select for many to many relationship editing

TW80000:
## Changes

* Adds multi-select support to tables when editing many-to-many relationship properties in the detail view
* Fixed a bug where attributes wouldn't be removed from DOM elements under certain conditions
* Removed `addSelection` and `addEnabled`. Just use `select` for both adding and removing instead.
  * Also removed `findRelatedObject` since it was referencing `addEnabled` and I couldn't tell what it was for. @adamvy is it fine to remove `findRelatedObject` or should I add it back? What does it do?

####  Looking at a random group
Screen Shot 2019-06-26 at 2 30 53 PM

#### After clicking "Add"
Screen Shot 2019-06-26 at 2 30 59 PM

#### After making some selections
Screen Shot 2019-06-26 at 2 31 10 PM

#### After clicking "Add"
Screen Shot 2019-06-26 at 2 31 17 PM

#### After clicking "Remove" in the detail view
Screen Shot 2019-06-26 at 2 31 27 PM

#### After clicking the checkbox in the table header to select all
Screen Shot 2019-06-26 at 2 31 36 PM

#### After clicking "Remove"
Screen Shot 2019-06-26 at 2 31 42 PM


jlhughes (commented):
@TW80000 How does this scale? If the relationship had thousands or hundreds of thousand entries?
I like this feature, we'll just have to be smart about which relationships we add it to.

TW80000:
> @TW80000 How does this scale? If the relationship had thousands or hundreds of thousand entries?

It's just a feature of the TableView and DAOController, so its performance is essentially just the performance of our TableView. All it's doing is an unfiltered select on the targetDAO specified in the relationship when adding and a filtered select on the targetDAO when removing. So it's just as performant as checking out a TableView of the targetDAO directly. The performance impact of the multi-select feature itself should be negligible.

> I like this feature, we'll just have to be smart about which relationships we add it to.

This feature would be enabled for all relationships, and I think that's what we want. Just like we can render a TableView of any DAO no matter how many entries there are in it, we can use this feature for editing any many-to-many relationship through the GUI.

jlhughes (approved):

src/foam/comics/BrowserView.js
```diff
@@ -78,13 +78,6 @@ foam.CLASS({
       name: 'selectEnabled',
       documentation: 'True to enable the select button.'
     },
-    {
-      class: 'Boolean',
-      name: 'addEnabled',
```
kgrgreer, line 83: Why is this removed?

src/foam/u2/Element.js
```diff
@@ -1248,14 +1245,17 @@ foam.CLASS({
...
-        }
-      }
+
+      // NOTE: This call might throw an exception if it happens to be called on
+      // an element that's in the OUTPUT state. If that's the case, we don't
+      // want to remove the entry in `this.attributes` or `this.attributeMap`,
+      // otherwise when the element is in the LOADED state later and we try to
+      // remove the attribute, the attribute will have been removed from the
+      // array and map, but the attribute will still be set on the DOM element.
+      this.onRemoveAttr(name);
```
adamvy: We have this same sort of ordering issue for all methods like "removeChild" which calls onRemoveChild.  I'm not sure we should bother to re-order all of them.  Calling anything but load() in the OUTPUT state is an error.

Additionally I'm not wild about the idea of calling onRemoveAttr() before we've actually updated our state, it feels wrong.  I see why you did it, so that the exception triggers before we've updated state and we can therefore successfully remove the attribute on a later attempt, but I don't think it's worth it.  This feels like a bandaid over a deeper bug of whatever code is outputting elements but not calling .load() immediately afterwards, such code shouldn't exist, and if it does we should figure out why.

src/foam/comics/BrowserView.js
TW80000: The 'add' action we have on the DAO controller is just a specific case of the more general 'select' action. Since we can use 'select' to select a bunch of objects to add or to remove, I think there's no need for a special action just for adding. That was my reason for removing it. If I'm missing some reason we need it though, please let me know.

src/foam/u2/Element.js
TW80000: Ok, I'll try to find out what's calling `.load` at the wrong time.

TW80000: @adamvy Fixed. I reverted the commit that made the change you commented on and found a different way to avoid the situation happening in the first place. See 5b1c62d for the fix.

PR #2343, opened 2019-06-28
Make foam.core.Reference properties override-able.

mcarcaso:
In java, you're currently unable to override properties on a foam.core.Reference and this PR unblocks that. Note that it needs to remove a method we're not using on StringPropertyInfo and that was preventing us from being able to do this because the `getWidth` method only gets filled in by foam.core.String and that won't run when installing a foam.core.Reference.


Nauna:
tested and works with the csvSink that Im working on.

jlhughes (approved):

PR #2345, opened 2019-07-01
Validate beanshell script testing for known security vulnerabilities.

jlhughes:


src/foam/nanos/script/Script.js
```diff
@@ -176,7 +181,46 @@ foam.CLASS({
...
+        considered invalid and if so throw an IllegalStateException.
+      `,
+      args: [
+        {
+          name: 'x', type: 'Context'
+        }
+      ],
+      javaThrows: ['IllegalStateException'],
+      javaCode: `
+      String script = getCode();
```
mcarcaso, line 214: I think you can remove this code, and the javaExtras above if you use a validationPredicate on the `code` property. I can show you what that looks like if you want.

kgrgreer:
We use getRuntime().getAvailableProcessors(), so if we disable that, we should make this available in the context. Better to set a custom SecurityManager to do these restrictions. Would me more secure/less error-prone. See:
https://docs.oracle.com/javase/8/docs/api/java/lang/SecurityManager.html

jlhughes:
@kgrgreer Everything I've read indicates the security manager does not help with beanshell. 
Going to pursue OS level restrictions and then revisit this if escaping is still an issue.

PR #2346, opened 2019-07-01
Add defaults to theme and give them the logic to install the CSS so t…

mcarcaso:
…hey can run independently of nanos.

Themes broke a separate project of mine so I needed these changes. If we're ok with this, we can remove the copied logic from ApplicationController but we'll need to take care with the context because the wrapCSS would come from here.

### DIFF
--- a/src/foam/core/Window.js
+++ b/src/foam/core/Window.js
@@ -18,7 +18,9 @@
 foam.CLASS({
   package: 'foam.core',
   name: 'Window',
-
+  requires: [
+    'foam.nanos.theme.Theme'
+  ],
   documentation: function(){/*
     Encapsulates top-level window/document features.
 
@@ -64,7 +66,8 @@ foam.CLASS({
     'setInterval',
     'setTimeout',
     'warn',
-    'window'
+    'window',
+    'theme'
   ],
 
   properties: [
@@ -77,6 +80,14 @@ foam.CLASS({
     {
       name: 'console',
       factory: function() { return this.window.console; }
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.nanos.theme.Theme',
+      name: 'theme',
+      factory: function() {
+        return this.Theme.create();
+      }
     }
   ],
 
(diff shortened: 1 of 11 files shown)


src/foam/nanos/theme/Theme.js
```diff
@@ -102,157 +112,244 @@ foam.CLASS({
     {
       class: 'Color',
       name: 'primary1',
-      section: 'colours'
+      section: 'colours',
+      value: 'unset'
```
TW80000, line 148: Wouldn't it be better to set these to the values in the ["FOAM default" theme](https://github.com/foam-framework/foam2/blob/master/src/themes)?

TW80000 (commented):
Another way of "fixing" this is to simply write the long form the CSS macro directly in the CSS code rather than the short form. See [ActionView](https://github.com/foam-framework/foam2/blob/master/src/foam/u2/ActionView.js#L85-L86) for an example. This was suggested by @kgrgreer and @adamvy, so I did a project-wide find-and-replace to do that across the repo as part of https://github.com/foam-framework/foam2/pull/2298. So that alone should have fixed it for projects that aren't using `ApplicationController`.

> [...] we can remove the copied logic from ApplicationController but we'll need to take care with the context because the wrapCSS would come from here.

Have you tested to see if removing that code from `ApplicationController` breaks our current app?

src/foam/nanos/theme/Theme.js
mcarcaso: Nope. Remember that foam exists without nanos so we shouldn't be doing things that require the use of nanos.

mcarcaso:
> Another way of "fixing" this is to simply write the long form the CSS macro directly in the CSS code rather than the short form. See [ActionView](https://github.com/foam-framework/foam2/blob/master/src/foam/u2/ActionView.js#L85-L86) for an example. This was suggested by @kgrgreer and @adamvy, so I did a project-wide find-and-replace to do that across the repo as part of #2298. So that alone should have fixed it for projects that aren't using `ApplicationController`.

I need to be able to modify themes without running nanos and the current setup makes foam dependent on nanos which shouldn't be the case.

> 
> > [...] we can remove the copied logic from ApplicationController but we'll need to take care with the context because the wrapCSS would come from here.
> 
> Have you tested to see if removing that code from `ApplicationController` breaks our current app?

I haven't. Simply removing that code certainly will break the app and I don't want to break things in this PR, I just want to make things work in a non-nanos environment.

adamvy:
I don't think we should move wrapCSS into Theme.  If we're putting this syntax into core U2 widgets, then we should make the top level installCSS function understand them.

Let's just inline upgrade the window level installCSS to understand the macro syntax, even if all it does is expand every instance of `/*%ANY_MACRO%/ default` to `default`

src/foam/nanos/theme/Theme.js
TW80000: Right, but why does that make `"unset"` a more reasonable default here than `"#202341"`?

mcarcaso: Oh sorry misunderstood your previous comment. I'd rather have the defaults than some arbitrary blue colour. My personal use case is that I am embedding foam widgets in a webpage and I want to just inherit the various styles from the parent node by default.

mcarcaso:
Inlining the css parsing/binding makes sense to me. I'll do that when I get a chance.

src/foam/nanos/theme/Theme.js
TW80000: Ah, I didn't know `unset` was a CSS thing. I see now. Thanks.

src/foam/nanos/theme/Theme.js
```diff
@@ -102,157 +112,244 @@ foam.CLASS({
...
     },
     {
       class: 'Color',
       name: 'black',
-      section: 'colours'
+      section: 'colours',
+      value: 'black'
```
TW80000: Wouldn't this make sense as `unset` as well?

mcarcaso: Maybe? This is mostly a selfish default but I feel like it isn't a totally unreasonable default. black being black doesn't feel particularly unexpected lol. I can switch this to unset for consistency if you think that'd be better.

TW80000: `unset` is more consistent with the other properties, and it should default to `black` in most cases anyway, shouldn't it? Isn't the default `color` on most browsers black? If whoever's using it has a different colour in a parent element we'd want this property's behaviour to be consistent with the other properties, right?

mcarcaso: Yup that make sense. I'll make this consistent with the rest :)

mcarcaso:
Ok I gave Window a theme and create/export a theme from that model.

src/foam/core/Window.js
```diff
@@ -196,12 +207,59 @@ foam.CLASS({
...
+      if ( text ) {
+        var eid = foam.u2.Element.NEXT_ID();
+
+        for ( var i = 0 ; i < this.Theme.MACROS.length ; i++ ) {
+          let m     = this.Theme.MACROS[i];
+          var text2 = this.expandShortFormMacro(this.expandLongFormMacro(text, m), m);
+
+            // If the macro was found, then listen for changes to the property
+            // and update the CSS if it changes.
+            if ( text != text2 ) {
```
mcarcaso, line 248: Note we have a bug here that breaks theme previewing when we use the long form of the CSS. Is this a known bug?

TW80000: I wasn't aware.

mcarcaso: Yeah when the long form matches the value set by the theme, text1 == tex2 so we don't bind things up.

TW80000: Ahh, that makes sense, thanks. Maybe we need to have a smarter way of binding things. Maybe we can do something like

```JavaScript
if ( /%\w+%/g.test(text) ) { ... }
```

to check if we need to set up the binding.

PR #2350, opened 2019-07-02
DO_NOT_MERGE Keep Email Transport connection alive between send calls.

jlhughes:
This is incomplete. created PR to share/discuss. 
outstanding
- using transport listener to report failed email sends
- provide retry at AsyncEmailService
- flag and retry failed emailMessages


jlhughes:
Addressed in another PR.

PR #2356, opened 2019-07-03
Cpf 1816/cico horizontal bar foam

tharmaman:


src/org/chartjs/HorizontalBarDAOChartView.js
```diff
@@ -0,0 +1,138 @@
...
+                  callback: function (value, index, values) {
+                    const dateArray = value.split('/');
+                    const monthNames = [
+                      "January", "February", "March", "April", "May", "June",
+                      "July", "August", "September", "October", "November", "December"
+                    ];
+                    const quarterNames = ['Q1', 'Q2', 'Q3', 'Q4'];
+
+                    switch (self.dateFrequency) {
+                      case net.nanopay.liquidity.ui.dashboard.DateFrequency.MONTHLY:
```
mcarcaso: This isn't part of the foam repo so it shouldn't be here. You can override the config wherever you're rendering this view. I'd keep the config of this model super simple:
```
{ type: 'horizontalBar' }
```

src/org/chartjs/HorizontalBarDAOChartView.js
```diff
@@ -0,0 +1,138 @@
...
+            });
+
+            config.data = {
+              labels: sink.groupKeys,
+              datasets: Object.keys(dataMap).map(k => {
+                var data = {
+                  label: k,
+                  data: dataMap[k]
+                };
+                // var style = self.customDatasetStyling[k] || {};
```
mcarcaso: You might want to preserve this to get different styled bars. Was it causing problems?

src/org/chartjs/PieDAOChartView.js
```diff
@@ -52,6 +50,9 @@ foam.CLASS({
             config.data = {
               datasets: [{
                 data: Object.keys(sink.groups).map(key => sink.groups[key].value),
+                backgroundColor: [
```
mcarcaso: This will make all the content the same colour and I'm not sure we want that. We should make this configurable in some way.

tharmaman: yeah I'll remove this here I handle it in my multiple currencies PR

src/org/chartjs/HorizontalBarDAOChartView.js
tharmaman: Done.

src/org/chartjs/HorizontalBarDAOChartView.js
tharmaman: Done.

src/org/chartjs/HorizontalBarDAOChartView.js
```diff
@@ -0,0 +1,100 @@
...
+
+  listeners: [
+    {
+      name: 'dataUpdate',
+      isFramed: true,
+      code: function () {
+        var self = this;
+        self.data
+          .orderBy(this.yExpr)
+          .select(this.GROUP_BY(this.yExpr, this.GROUP_BY(this.keyExpr, this.SUM(this.xExpr))))
```
adamvy, line 59: Consider using PLOT() if the keys for group_by are giving you issues.

src/org/chartjs/HorizontalBarDAOChartView.js
```diff
@@ -0,0 +1,100 @@
...
+          .then(function (sink) {
+            // Clear data before cloning because it gets clobbered anyway.
+            self.config.data = { datasets: [] };
```
adamvy, line 64: Do we really want to clear the datasets and then set them again?  Will that trigger multiple plains from the Charts?  We could also just build the new datasets and just do self.config.data = ..., followed by i believe an .update() call on the chartjs object.

I guess this works it just seems like unnecessary cloning.

test/src/lib/mlang.js
```diff
@@ -486,7 +486,8 @@ describe('MLang', function() {
     });
   });
 
-  describe('CONTAINS()', function() {
+  describe('CONTA
```
adamvy: ?

src/org/chartjs/HorizontalBarDAOChartView.js
mcarcaso: Maybe? I remember having issues with changing just the datasets on chartjs so we clone the config and re-render the entire chart.

Leave this for now if it works and I'll think about doing it better.

test/src/lib/mlang.js
tharmaman: Whooops fixed it

PR #2357, opened 2019-07-03
Call sink.eof before returning the sink in MDAO's select and remove a…

mcarcaso:
…n eof call from TreeNode that should not longer be needed.


mcarcaso:
Note travis breakage is unrelated to this. Potential fix is here https://github.com/foam-framework/foam2/pull/2358/files but I want Adam to help.

PR #2358, opened 2019-07-03
Ensure there's a reply box in LookupBox.

mcarcaso:
This fixes travis tests.


mcarcaso:
Chatted with Adam offline, he says to fix the test rather than putting this "fix" in. Do not merge this for now.

adamvy:
I will fix tonight

PR #2359, opened 2019-07-03
add listeners to CachingAuthService

smirnovaka:


src/foam/nanos/auth/CachingAuthService.java
```diff
@@ -51,10 +53,10 @@ public Object create(X x) {
   protected static Map<String,Boolean> getPermissionMap(final X x) {
     Session             session = (Session) x.get(Session.class);
     Map<String,Boolean> map     = (Map) session.getContext().get(CACHE_KEY);
+    User                user    = (User) x.get("user");
```
TW80000: This can be moved inside the `if` since we don't use it unless `map` is null. Saves a context lookup for every call to `check` when the cache is already there.

src/foam/nanos/auth/CachingAuthService.java
```diff
@@ -66,7 +68,16 @@ public void eof() {
...
           purgeCache(x);
         }
-      }, TRUE);
+      };
+
+      DAO userDAO       = (DAO) x.get("userDAO");
+      DAO groupDAO      = (DAO) x.get("groupDAO");
+      DAO groupPermissionJunctionDAO = (DAO) x.get("groupPermissionJunctionDAO");
+
+      groupDAO.listen(purgeSink, TRUE);
```
TW80000, line 80: We're setting up these listeners inside `getPermissionMap`, which means we're going to be setting up three listeners for every user in the system. Would it be more efficient to set up 3 listeners when this class is initialized and purge all caches in all sessions when an event happens? Genuine question. I'm not sure if the overhead of all of the listeners would be greater than or less than the overhead of doing the lookup of all sessions.

src/foam/nanos/auth/CachingAuthService.java
```diff
@@ -66,7 +68,16 @@ public void eof() {
...
+      userDAO.listen(purgeSink, EQ(User.ID, user.getId()));
```
TW80000, line 81: Another thought about setting up these listeners. When a session is deleted/finished, are we removing/detaching the listeners? Wouldn't it be a memory leak if we aren't?

TW80000, line 81: We only need to be purging a user's cache if their `enabled` property changes, not on every put. Same applies to when `Group`s change in `groupDAO`. Clearing on every put to `userDAO` and `groupDAO` could result in far more cache misses than is necessary.

src/foam/nanos/auth/CachingAuthService.java
```diff
@@ -87,7 +98,7 @@ public CachingAuthService(AuthService delegate) {
   @Override
   public boolean check(foam.core.X x, String permission) {
     if ( x == null || permission == null ) return false;
-
+    if ( ((User)x.get("user")).getId() == 1 ) return true;
```
TW80000, line 105: To save a context lookup (not sure if worthwhile or not), we could do the lookup here and pass the result as an argument to `getPermissionMap` so it doesn't need to do the same lookup.

src/foam/nanos/auth/CachingAuthService.java
```diff
@@ -66,7 +68,16 @@ public void eof() {
         public void reset(Detachable sub) {
           purgeCache(x);
         }
-      }, TRUE);
+      };
+
+      DAO userDAO       = (DAO) x.get("userDAO");
```
TW80000: We should be listening on `localGroupDAO` and `localGroupDAO` so anything that accesses those directly will still cause the listeners to trigger.

src/foam/nanos/auth/CachingAuthService.java
kgrgreer: agreed

src/foam/nanos/auth/CachingAuthService.java
kgrgreer: Switch to local versions. Is that what you're saying?

src/foam/nanos/auth/CachingAuthService.java
kgrgreer: If we just had three listeners then each session would have to listen for whatever those listeners trigger, which would result in the same amount of work.

src/foam/nanos/auth/CachingAuthService.java
kgrgreer: Yes, we need to unsubscribe once the session is removed. How do we know when a session is stale?

kgrgreer: We would need lots of logic to determine when we actually need to purge the cache, but that is error prone and since users and groups shouldn't change that often, it isn't much overhead to purge caches and rebuild caches.

src/foam/nanos/auth/CachingAuthService.java
kgrgreer: getPermissionMap is also called by a static method, so I don't think it is worth it. You can do millions of context lookups per second anyway.

src/foam/nanos/auth/CachingAuthService.java
TW80000: Yeah

PR #2360, opened 2019-07-04
Attribute map name added to radio input name

zahnno:
I encountered an issue when having two properties each with a radio view with the same choice values 
ie. ( choices: [ [ true, 'yes' ], [ false, 'no' ] ] )

I did some investigation and saw that we were assigning the name of our radio inputs as the first index in our choices. This presents issues when several properties share the same values since HTML registers them as a group - making only one capable of being selected.

The following changes specifies the name on the radio input if an attributeMap is present.

### DIFF
--- a/src/foam/u2/view/RadioView.js
+++ b/src/foam/u2/view/RadioView.js
@@ -70,7 +70,7 @@ foam.CLASS({
           start('input').
             attrs({
               type: 'radio',
-              name: c[0] + '',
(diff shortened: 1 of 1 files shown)


kgrgreer:
I fixed the bug with this:

commit 7c033d9ba7a7694a3099c17def6da3fcda8739b8
Date:   Thu Jul 4 10:01:56 2019 -0400

    Fix RadioView to set name to be name of Property.

PR #2361, opened 2019-07-04
Add text key for 2FA

Henry0422:
Model an OTPKey as an object which has "key" and "qr-code" as properties.
The `generateKeyAndQR` method returns the OTPKey object.

GoogleTOTPAuthService extends AbstractTOTPAuthService
AbstractTOTPAuthService extends AbstractOTPAuthService
AbstractOTPAuthService implements OTPAuthService
OTPAuthService is an interface.


src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
```diff
@@ -65,20 +69,23 @@ foam.CLASS({
...
-
         try {
           EmailConfig service = (EmailConfig) x.get("emailConfig");
           String name = service == null ? "FOAM" : service.getDisplayName();
           String path = String.format("/%s:%s", name, user.getEmail());
           String query = String.format("secret=%s&issuer=%s&algorithm=%s", key, name, getAlgorithm());
           URI uri = new URI("otpauth", "totp", path, query, null);
-          return "data:image/svg+xml;charset=UTF-8," + QrCode.encodeText(uri.toASCIIString(), QrCode.Ecc.MEDIUM).toSvgString(0);
+
+          otpKey.setKey(key);
```
TW80000: This line and line 82 can be removed.

src/foam/nanos/auth/twofactor/OTPKey.js
```diff
@@ -0,0 +1,16 @@
+foam.CLASS({
```
TW80000, line 7: Missing license header.

src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
```diff
@@ -65,20 +69,23 @@ foam.CLASS({
...
+          otpKey.setQrCode("data:image/svg+xml;charset=UTF-8,"
+            + QrCode.encodeText(uri.toASCIIString(), QrCode.Ecc.MEDIUM).toSvgString(0));
+          return otpKey;
         } catch ( Throwable t ) {
-          throw new RuntimeException(t);
+          logger.error("Error when generating QR code: " + t);
```
mcarcaso: Should we be swallowing this error like this or still throwing an exception? To me, something exceptional has happened if we've failed to retrieve a QR code and if I'm requesting an OTPKey, I'll never know that something went wrong.

src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
mcarcaso: Use builder when you can:
```
return new OTPKey().Builder(x)
  .setKey(key)
  .setQrCode("data:image/svg+xml;charset=UTF-8,"+
    QrCode.encodeText(uri.toASCIIString(), QrCode.Ecc.MEDIUM).toSvgString(0))
  .build();
```

src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
```diff
@@ -46,14 +47,16 @@ foam.CLASS({
...
       javaCode: `
+        Logger logger = (Logger) x.get("logger");
+
         User user = (User) (x.get("agent") != null ?
           x.get("agent") :
           x.get("user")) ;
         DAO userDAO = (DAO) x.get("localUserDAO");
 
-        // fetch from user dao to get secret key
+        // fetch from localUserDAO to get updated user before setting TwoFactorSecret
```
mcarcaso, line 56: Personally, I don't find these kinds of comments to be super helpful and would recommend removing comments that simply describe what the code is doing. Comments are important for explaining "why" code is doing what it's doing as oppose to what the code is doing. Describing what the code is doing means a developer has to remember to update these comments any time the code changes and that's often forgotten.

Just my two cents :)

src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
```diff
@@ -65,19 +68,20 @@ foam.CLASS({
...
-          return "data:image/svg+xml;charset=UTF-8," + QrCode.encodeText(uri.toASCIIString(), QrCode.Ecc.MEDIUM).toSvgString(0);
+
+          return new OTPKey.Builder(x)
+            .setKey(key)
+            .setQrCode("data:image/svg+xml;charset=UTF-8,"+
+              QrCode.encodeText(uri.toASCIIString(), QrCode.Ecc.MEDIUM).toSvgString(0))
+            .build();
         } catch ( Throwable t ) {
-          throw new RuntimeException(t);
+          throw new RuntimeException(t);
```
mcarcaso: Remove trailing whitespace. Also it might be helpful to add some messaging to this exception so we can give a useful error message rather than just having the client receive "RuntimeException". Do we have an error type that the client knows how to interpret and display?

src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js
```diff
@@ -46,14 +47,16 @@ foam.CLASS({
 
   methods: [
     {
-      name: 'generateKey',
+      name: 'generateKeyAndQR',
       javaCode: `
+        Logger logger = (Logger) x.get("logger");
```
mcarcaso: I think this is no longer needed. Unless you also want to log errors so we can monitor how frequently we fail to fetch QR codes.

mcarcaso (approved):
Looks good. Just a few more comments but nothing blocking.

TW80000 (approved):

PR #2365, opened 2019-07-04
Table style update

TW80000:
## Changes

* The `TableView` is implemented using `div` elements now instead of the traditional `table`, `tbody`, `thead`, `tr`, `th`, and `td` elements. This frees us from a number of CSS limitations and bugs. The reason I did this was because I came across a bug in the CSS specification/browser implementation of it when setting table elements to `position: sticky`, where the border of the header wouldn't be sticky. There are other annoying quirks around the table elements in HTML that I've come across before, so it's simplest just to use `div`s instead and avoid the issues altogether.f
  * The DOM is a bit uglier now and the HTML isn't as semantic, but it's a tradeoff to get the application behaviour we want, which I would consider worthwhile
* Updated tables to match the designs
* Changed a handful of hard-coded colours to reference the Theme
* Fixed a tiny visual bug where the table and heading wouldn't be aligned when you toggle the filters in the DAO controller

### Showing the table
Screen Shot 2019-07-04 at 4 03 43 PM

### Showing hover and selected colour
Screen Shot 2019-07-04 at 4 03 55 PM

### DIFF
--- a/src/foam/comics/DAOControllerView.js
+++ b/src/foam/comics/DAOControllerView.js
@@ -60,11 +60,8 @@ foam.CLASS({
     ^container {
       display: grid;
       grid-template-columns: fit-content(100%) auto;
+      grid-gap: 10px;
       overflow-x: scroll;
-    }
-
-    ^container > * + * {
-      margin-left: 10px;
     }
 
     ^ .actions {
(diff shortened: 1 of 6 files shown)


mcarcaso (approved):
I'm good with this if it allows us to get the behaviour we want.

PR #2366, opened 2019-07-07
CPF-2218/SecureDAOs

jlhughes:
Addressing Penetration Testing results which revealed a number of improperly authenticated DAOs. 

All DAOs need to be migrated to EasyDAO, but this first set are significant.


src/foam/dao/EasyDAO.js
```diff
@@ -102,20 +107,41 @@ foam.CLASS({
...
 //logger.info(this.getClass().getSimpleName(), "delegate", "NSpec.name", getNSpec().getName(), "of_", of_); //Thread.currentThread().getName());
 foam.dao.DAO delegate = getInnerDAO() == null ?
   new foam.dao.MDAO(getOf()) :
   getInnerDAO();
 
 if ( delegate instanceof foam.dao.MDAO ) {
   setMdao((foam.dao.MDAO)delegate);
+  if ( getIndex() != null &&
+       getIndex().length > 0 ) {
+    // IMPORTANT: Indexes must be added before JDAO replay occurs.
```
kgrgreer: I don't think this is true, but it doesn't hurt.

jlhughes: At the moment this is the case, DAOs are simply broken otherwise.

src/foam/dao/EasyDAO.js
```diff
@@ -388,6 +426,16 @@ return delegate;
...
     },
+    {
+      class: 'FObjectArray',
+      of: 'foam.core.PropertyInfo',
+      name: 'order'
+    },
+    {
+      class: 'FObjectArray',
+      of: 'foam.core.PropertyInfo',
+      name: 'index'
```
kgrgreer, line 487: Just adding on index isn't very useful. Better if this took an array of indices (each of which is an array of PropertyInfos.

src/foam/dao/EasyDAO.js
```diff
@@ -189,6 +222,11 @@ return delegate;
       type: 'foam.dao.DAO',
       name: 'innerDAO'
     },
+    {
+      class: 'Object',
+      type: 'foam.dao.DAO',
+      name: 'decorator'
```
kgrgreer, line 266: This would be more useful if there were more than one.

src/foam/dao/EasyDAO.js
jlhughes: @kgrgreer

src/foam/dao/EasyDAO.js
```diff
@@ -162,6 +188,13 @@ if ( getContextualize() ) {
     build();
 }
 
+if ( getOrder() != null &&
+     getOrder().length > 0 ) {
+  for ( foam.core.PropertyInfo prop : getOrder() ) {
+    delegate = delegate.orderBy(prop);
```
kgrgreer: This will just order by the last property. You want a CompoundComparator, which we currently only support in JS, but we should fix/implement that. Will also require work on MDAO to support.

src/foam/dao/EasyDAO.js
```diff
@@ -162,6 +188,13 @@ if ( getContextualize() ) {
     build();
 }
 
+if ( getOrder() != null &&
+     getOrder().length > 0 ) {
```
kgrgreer, line 207: The 'order' property isn't an array, so .length doesn't make sense.

src/foam/dao/EasyDAO.js
```diff
@@ -576,6 +624,12 @@ return delegate;
         dao = decorated;
       }
 
+      if ( this.order ) {
```
kgrgreer, line 681: Didn't you already do this above? Also, order isn't an array.

src/foam/dao/EasyDAO.js
jlhughes: @kgrgreer It handles a ProxyDAO - walks the chain and sets the delegate on the last.

src/foam/dao/index/AltIndex.java
```diff
@@ -37,13 +37,13 @@ public Object addIndex(Object state, Index i) {
...
     final Object[] sa   = toObjectArray(state);
-          Sink     sink = new AbstractSink() {
+    Sink     sink = new AbstractSink() {
       public void put(Object obj, foam.core.Detachable sub) {
         sa[sa.length-1] = i.put(sa[sa.length-1], (FObject) obj);
       }
     };
 
-    planSelect(sa[0], sink, 0, Long.MAX_VALUE, null, null).select(sa[0], sink, 0, Long.MAX_VALUE, null, null);
+    planSelect(sa, sink, 0, Long.MAX_VALUE, null, null).select(sa[0], sink, 0, Long.MAX_VALUE, null, null);
```
kgrgreer: Should the second reference to sa[0] also be converted to 'sa'?

src/foam/dao/EasyDAO.js
jlhughes: Presently serviceScripts only have one orderBy so this will work for now and we can add support in another PR.

src/foam/dao/EasyDAO.js
jlhughes: @kgrgreer 'order' is an FObjectArray of foam.mlang.order.Comparator

src/foam/dao/index/AltIndex.java
jlhughes: @kgrgreer it's actually ignored. So more correctly would be to pass null.

src/foam/dao/EasyDAO.js
jlhughes: @kgrgreer one is java, this is javascript.

src/foam/dao/index/AltIndex.java
jlhughes: @kgrgreer planSelect returns an AltSelectPlan which uses it's internal state_ rather than the argument state passed on the select call.

src/foam/dao/AuthenticatedDAO.java
```diff
@@ -61,7 +61,7 @@ public FObject put_(X x, FObject obj) {
     } else {
       permission = createPermission("update", id);
     }
-
+    ((foam.nanos.logger.Logger)x.get("logger")).debug("check put permission", permission);
```
TW80000: Should probably be removed before merging.

src/foam/dao/EasyDAO.js
TW80000: @jlhughes Wouldn't we just end up ordering by whatever the last Comparator in the array was? If so, it doesn't make sense for 'order' to be an Array.

src/foam/nanos/auth/Relationships.js
```diff
@@ -10,7 +10,7 @@ foam.RELATIONSHIP({
   targetModel: 'foam.nanos.auth.Permission',
   forwardName: 'permissions',
   inverseName: 'groups',
-  junctionDAOKey: 'groupPermissionJunctionDAO'
+  junctionDAOKey: 'localGroupPermissionJunctionDAO'
```
TW80000: I don't think we want to do this. If we do, I think client-side code related to relationships will try to use `localGroupPermissionJunctionDAO` from the client and fail since it isn't served.

src/services
```diff
@@ -82,14 +83,17 @@ p({"class":"foam.nanos.boot.NSpec", "name":"supportEmailDAO",                  "
...
+//p({"class":"foam.nanos.boot.NSpec","name":"groupPermissionJunctionDAO","description":"The junction DAO for the many-to-many relationship between groups and permissions","serve":true,"serviceScript":"mdao = new foam.dao.MDAO(foam.nanos.auth.GroupPermissionJunction.getOwnClassInfo()); ed = new foam.dao.EasyDAO.Builder(x)\u000a  .setOf(foam.nanos.auth.GroupPermissionJunction.getOwnClassInfo())\u000a  .setAuthenticate(false)\u000a  .setAuthenticateRead(false)\u000a  .setJournalType(foam.dao.JournalType.SINGLE_JOURNAL)\u000a  .setJournalName(\"groupPermissionJunctions\")\u000a  .setPm(true)\u000a  .setMdao(mdao).build(); \u000a\u000areturn new foam.nanos.auth.AuthorizationDAO(x, ed);","client":"{\u000a  \"of\": \"foam.nanos.auth.GroupPermissionJunction\"\u000a}","authNotes":"Decorated with an AuthorizationDAO. Reads are not checked."})
```
TW80000: We shouldn't commit commented out code if we don't need it.

TW80000 (commented):
I'm getting this error when I check out this branch and try to use the client:

```
2019.07.09.13.52.24.566,DEBUG,1348 - admin,[Service],nSpecDAO,returning skeleton exception,java.lang.NullPointerException
	at foam.nanos.auth.CachingAuthService.getPermissionMap(CachingAuthService.java:82)
	at foam.nanos.auth.CachingAuthService.check(CachingAuthService.java:108)
	at foam.nanos.auth.ProxyAuthService.check(ProxyAuthService.java:241)
	at foam.nanos.auth.EnabledCheckAuthService.check(EnabledCheckAuthService.java:42)
	at net.nanopay.meter.compliance.ComplianceAuthService.check(ComplianceAuthService.java:71)
	at foam.nanos.auth.ProxyAuthService.check(ProxyAuthService.java:241)
	at foam.nanos.auth.ProxyAuthService.check(ProxyAuthService.java:241)
	at foam.nanos.auth.ProxyAuthService.check(ProxyAuthService.java:241)
	at foam.nanos.auth.ProxyAuthService.check(ProxyAuthService.java:241)
	at foam.nanos.auth.twofactor.TwoFactorAuthService.check(TwoFactorAuthService.java:31)
	at foam.nanos.auth.PermissionedPropertyDAO.maybeRemoveProperties(PermissionedPropertyDAO.java:90)
	at foam.nanos.auth.PermissionedPropertyDAO.find_(PermissionedPropertyDAO.java:32)
	at foam.dao.ProxyDAO.find_(ProxyDAO.java:235)
	at foam.dao.AuthenticatedDAO.find_(AuthenticatedDAO.java:95)
	at foam.dao.ProxyDAO.find_(ProxyDAO.java:235)
	at foam.dao.DAOSkeleton.send(DAOSkeleton.java:147)
	at foam.dao.SessionDAOSkeleton.send(SessionDAOSkeleton.java:51)
	at foam.box.SessionServerBox.send(SessionServerBox.java:153)
	at foam.nanos.http.ServiceWebAgent.execute(ServiceWebAgent.java:118)
	at foam.nanos.http.NanoRouter.service(NanoRouter.java:85)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:865)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1655)
	at org.eclipse.jetty.websocket.server.WebSocketUpgradeFilter.doFilter(WebSocketUpgradeFilter.java:215)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1642)
	at net.nanopay.security.csp.CSPFilter.doFilter(CSPFilter.java:24)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1642)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:533)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:255)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1253)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:203)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:473)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:201)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1155)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:144)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:132)
	at org.eclipse.jetty.server.Server.handle(Server.java:531)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:352)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:260)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:281)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:102)
	at org.eclipse.jetty.io.ChannelEndPoint$2.run(ChannelEndPoint.java:118)
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:333)
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:310)
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.tryProduce(EatWhatYouKill.java:168)
	at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.run(EatWhatYouKill.java:126)
	at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:366)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:760)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$2.run(QueuedThreadPool.java:678)
	at java.lang.Thread.run(Thread.java:748)
```

src/foam/dao/EasyDAO.js
jlhughes: 'order' is an array, the issue is the dao.orderBy call which needs to compose them in some way.

src/foam/nanos/auth/Relationships.js
jlhughes: @TW80000 why would client side code be accessing this dao?

TW80000: If you go to the DetailView for either of those models (Group and Permission) and try to edit the relationship using the GUI, it's going to try to read from and write to the junctionDAO, which won't be served if you change this to the local version. It would also break the permission matrix, which needs to read from groupPermissionJunctionDAO on the client to know which permissions the different groups have.

src/services
```diff
@@ -90,8 +90,9 @@ p({"class":"foam.nanos.boot.NSpec", "name":"ruleHistoryDAO",                   "
...
-p({"class":"foam.nanos.boot.NSpec","name":"groupPermissionJunctionDAO","description":"The junction DAO for the many-to-many relationship between groups and permissions","serve":true,"serviceScript":"return new foam.dao.EasyDAO.Builder(x).setOf(foam.nanos.auth.GroupPermissionJunction.getOwnClassInfo()).setInnerDAO((foam.dao.DAO) x.get(\"\")).build();","client":"{\"of\": \"foam.nanos.auth.GroupPermissionJunction\"}"})
+p({"class":"foam.nanos.boot.NSpec","name":"groupPermissionJunctionDAO","description":"The junction DAO for the many-to-many relationship between groups and permissions","serve":true,"serviceScript":"return new foam.dao.EasyDAO.Builder(x).setOf(foam.nanos.auth.GroupPermissionJunction.getOwnClassInfo()).setAuthenticate(true).setInnerDAO((foam.dao.DAO) x.get(\"localGroupPermissionJunctionDAO\")).build();","client":"{\"of\": \"foam.nanos.auth.GroupPermissionJunction\"}"})
```
TW80000: We don't need `setAuthenticate(true)`. The `GroupPermissionJunction` model implements the `Authorizable` interface. `groupPermissionJunctionDAO` should be wrapped in an `AuthorizationDAO`.

src/foam/dao/EasyDAO.js
```diff
@@ -162,13 +203,43 @@ if ( getContextualize() ) {
     build();
 }
 
+if ( getOrder() != null &&
+     getOrder().length > 0 ) {
+  // TODO: CompositeDAO or thenBy
```
TW80000, line 208: TODO

Still doesn't make much sense how it's currently written without `ThenBy` since it'll just end up using whatever comparator was last in the list.

src/foam/dao/EasyDAO.js
```diff
@@ -162,13 +203,43 @@ if ( getContextualize() ) {
...
+    delegate = delegate.orderBy(comp);
+  }
+}
+
+if ( getAuthorize() ) {
+  setAuthenticate(false);
+  if ( foam.nanos.auth.Authorizable.class.isAssignableFrom(getOf().getObjClass()) ) {
+    delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate);
+  } else {
+    logger.warning("EasyDAO", "authorize=true but 'of' ",getOf().getId(), "not Authorizable");
```
TW80000, line 219: In this case we're setting `authenticate` to false above but not actually adding an authorization decorator. Isn't that a security hole?

src/foam/dao/EasyDAO.js
```diff
@@ -227,6 +303,12 @@ return delegate;
       generateJava: false,
       value: false
     },
+    {
+      /** Enable standard authorization. */
```
TW80000, line 307: This isn't technically true with the way it's currently set up in this PR. The default `Authorizer` is `AuthorizableAuthorizer`. I don't think we have `StandardAuthorizer` yet, but @penzital is working on it if I'm not mistaken.

src/tests
```diff
@@ -43,7 +43,7 @@ p({"class":"foam.nanos.test.Test","passed":10,"failed":0,"id":"FoldAndReduceTest
...
...(userContext).remove(j);\n\n    // If the put didn't throw, then this test failed.\n    test(false, TEST_MESSAGE);\n  } catch (AuthorizationException e) {\n    passed = e.getMessage().equals(\"Permission denied. You cannot add or remove a permission that you do not have.\");\n    test(passed, TEST_MESSAGE);\n    if ( ! passed ) {\n      print(\"Error message mismatch. Actual was: \" + e.getMessage());\n    }\n  } finally {\n    cleanUp(x);\n  }\n}\n\n\n// Run the tests.\ntry {\n  createGroupWithoutPermissionThrows();\n  addNewPermissionThrows();\n  addGroupWithAdminParentThrows();\n  updateGroupWithoutPermissionThrows();\n  updateGroupParentToAdminThrows();\n  createUserInAdminGroupThrows();\n  updateUserGroupToAdminThrows();\n  updateOwnGroupThrows();\n  createPermissionWithoutPermissionThrows();\n  createPermissionWithPermission();\n  updatePermissionWithoutPermissionThrows();\n  deletePermissionWithoutPermissionThrows();\n  deleteGroupWithoutPermissionThrows();\n  removePermissionFromGroupWithoutPermissionThrows();\n} catch (Throwable e) {\n  e.printStackTrace();\n  test(false, \"An unexpected exception was thrown. Some tests might not have been executed.\");\n}\n","output":""})
```
TW80000, line 46: @jlhughes Can you please paste the updated script as a reply to this comment so I can diff it with the current one?

src/foam/dao/EasyDAO.js
smirnovaka: @jlhughes  can you please clarify. If `authorize` is set to `true` then `authenticate` goes to false automatically?

src/foam/dao/EasyDAO.js
```diff
@@ -162,13 +203,43 @@ if ( getContextualize() ) {
...
+    getPermissionPrefix(),
     getAuthenticateRead(),
     delegate);
 }
 
+if ( getNSpec() != null &&
+     getNSpec().getServe() &&
+     ! getAuthorize() &&
+     ! getAuthenticate() &&
+     ! getReadOnly() ) {
```
smirnovaka, line 234: why is ReadOnly important? Isn't it a common case when served service is not readOnly? Like all the daos the client is writing to?

src/foam/dao/EasyDAO.js
smirnovaka: Yes, currently it will be only authorizing objects that implement authorizable but the StandardAuthorizer is in progress so I think we can keep it this way. Troy is correct(almost) @noodlemoodle is working on it.

src/foam/dao/EasyDAO.js
```diff
@@ -227,6 +303,12 @@ return delegate;
       generateJava: false,
       value: false
     },
+    {
+      /** Enable standard authorization. */
+      class: 'Boolean',
+      name: 'authorize',
+      value: false
```
smirnovaka, line 310: a suggestion, maybe should be a part of second secure refactoring ... Should the default be true?

src/foam/nanos/notification/email/services.jrl
```diff
@@ -1,5 +1,12 @@
-p({"class":"foam.nanos.boot.NSpec", "name":"emailTemplateDAO",                 "lazy":false,  "serve":true,  "serviceScript":"return new foam.dao.PMDAO(x, new foam.dao.PipelinePMDAO(x, new foam.dao.SequenceNumberDAO(new foam.dao.java.JDAO(x, foam.nanos.notification.email.EmailTemplate.getOwnClassInfo(), \"emailTemplates\"))));","client":"{\"of\":\"foam.nanos.notification.email.EmailTemplate\"}" })
+p({"class":"foam.nanos.boot.NSpec", "name":"localEmailTemplateDAO",                 "lazy":false,  "serve":false,  "serviceScript":"return new foam.dao.EasyDAO.Builder(x).setOf(foam.nanos.notification.email.EmailTemplate.getOwnClassInfo()).setAuthenticate(false).setPm(true).setPipelinePm(true).setSeqNo(true).setJournalType(foam.dao.JournalType.SINGLE_JOURNAL).setJournalName(\"emailTemplates\").build();"})
```
smirnovaka, line 1: should it be `authenticate:false`? It's not even served

src/foam/nanos/notification/email/services.jrl
```diff
@@ -1,5 +1,12 @@
...
+
+p({"class":"foam.nanos.boot.NSpec", "name":"emailTemplateDAO",                 "lazy":false,  "serve":true,  "serviceScript":"return new foam.dao.EasyDAO.Builder(x).setOf(foam.nanos.notification.email.EmailTemplate.getOwnClassInfo()).setInnerDAO((foam.dao.DAO) x.get(\"localEmailTemplateDAO\")).build();","client":"{\"of\":\"foam.nanos.notification.email.EmailTemplate\"}" })
```
smirnovaka, line 3: this and and pretty much all other services are set `authorize:false` (not set at all but it is the default). As I mentioned above I think the default should be true or explicitly set to true. I'd imagine only local services would not require authorization

src/foam/dao/EasyDAO.js
jlhughes: If authorize is true and the test for instanceof passes, then we set authenticate to false.  This is temporary as we slowly transition from authenticate to authorize. Similar with read-only, we can't just set it until we go through every EasyDao setup.

src/foam/dao/EasyDAO.js
jlhughes: Yes, this definitely a TODO. It was never my intent to implement all this now.  We need to move as much of the current service setups to EasyDAO.   We also need EasyDAO to turn clustering on.

src/foam/dao/EasyDAO.js
jlhughes: For 'serve'd daos, they should be authenticate, authorized, or read-only.   Nothing should be open read-write.

src/foam/dao/EasyDAO.js
jlhughes: Yes, this is coming later.  This is phase I, we can't convert everything at once.

src/foam/nanos/notification/email/services.jrl
jlhughes: It's EasyDAO and presently authenticate defaults to true.

src/foam/dao/EasyDAO.js
jlhughes: Yes, but no worse than what we currently have.  We have to keep the system running while we make all these changes.

PR #2373, opened 2019-07-08
Predicate DAOMenu

zahnno:
I figured the DAOMenu would do well with the ability to pass in a predicate and an "of" class considering we might want to display records and create objects etc...  that aren't associated to the DAOKey passed but still would like the objects to be saved to that DAO.

Examples of this being helpful is with subclasses

### DIFF
--- a/src/foam/nanos/menu/DAOMenu.js
+++ b/src/foam/nanos/menu/DAOMenu.js
@@ -18,6 +18,10 @@ foam.CLASS({
     },
     {
       class: 'String',
+      name: 'of'
+    },
+    {
+      class: 'String',
       name: 'title'
     },
     {
@@ -28,6 +32,11 @@ foam.CLASS({
       class: 'String',
       name: 'customDAOController',
       displayWidth: 80
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.mlang.predicate.Predicate',
+      name: 'predicate'
     },
     {
       class: 'foam.u2.ViewSpec',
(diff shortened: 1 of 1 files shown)


src/foam/nanos/menu/DAOMenu.js
```diff
@@ -133,9 +142,14 @@ foam.CLASS({
         throw new Error('No DAO found for ' + this.daoKey);
       }
 
+      var dao = X[this.daoKey];
+      dao.of = this.of ? this.of : dao.of;
+
+      var predicatedDAO = dao.where(this.predicate);
```
TW80000, line 148: What if `predicate` isn't set? Shouldn't this be something like
```suggestion
      var predicatedDAO = this.predicate ? dao.where(this.predicate) : dao;
```

src/foam/nanos/menu/DAOMenu.js
```diff
@@ -133,9 +142,14 @@ foam.CLASS({
         throw new Error('No DAO found for ' + this.daoKey);
       }
 
+      var dao = X[this.daoKey];
+      dao.of = this.of ? this.of : dao.of;
```
TW80000, line 146: Wouldn't this be changing the `of` for that DAO everywhere? For example, if you set `of` to Business and `daoKey = "userDAO"`, then you'd be setting the `of` of `userDAO` to `Business` on your client which might affect other parts of the client code that expect it to be `User`. I'm not sure if this assignment will affect that wide of a scope or not though. I'd have to test to see.

adamvy: Yes we shouldn't mutate the `of` property of the dao.

If you want to make sure the DAOController only creates business objects, what you could do is create a new ProxyDAO which has it's own `of` set to `Business` and the delegate of which is a predicated dao for `INSTANCE_OF(Business)`

zahnno: Couldn't we clone the DAO and set the of in the DAOMenu? That would prevent the DAO within the context to be mutated.

nvm looks like that won't work.

jlhughes:
DAOMenu is being deprecated.

PR #2374, opened 2019-07-08
Generic string parser

michaelconnor468:


src/foam/lib/json/MultiLineStringParser.java
```diff
@@ -0,0 +1,80 @@
...
+          unicodeParser.get() : asciiEscapeParser.get();
+        PStream escapePS = ps.apply(escapeSeqParser, x);
+        if ( escapePS != null ) {
+          builder.append(escapePS.value());
+          tail = escapePS;
+
+          c = (Character) escapePS.value();
+        }
+      } else {
+          builder.append(c);
```
kgrgreer: indent two spaces

src/foam/lib/json/MultiLineStringParser.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public MultiLineStringParser() {
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
+    if ( ! ps.valid() ) return null;
+    char delim = ps.head();
+    char lastc = delim;
+
+    for(int i = 0; i < 3; i++) {
```
kgrgreer: Fix spacing.

src/foam/lib/json/StringParser.java
```diff
@@ -9,68 +9,19 @@
 import foam.lib.parse.PStream;
 import foam.lib.parse.Parser;
 import foam.lib.parse.ParserContext;
+import foam.lib.parse.Alt;
 
 public class StringParser
   implements Parser
```
kgrgreer: Couldn't this extend AltParser and configure in consturctor rather than every time in parse() method.

michaelconnor468: It could! I prefer this method since delegation is a tad more readable and natural than extension here. I will move the initialization of the alt object to the body to address the performance issue.

src/foam/lib/json/StringParser.java
```diff
@@ -9,68 +9,19 @@
...
-    @Override
-    public StringBuilder get() {
-      StringBuilder b = super.get();
-      b.setLength(0);
-      return b;
-    }
-  };
+  MultiLineStringParser mp = new MultiLineStringParser();
+  SingleLineStringParser sp = new SingleLineStringParser();
+  Alt altp = new Alt(new MultiLineStringParser(), new SingleLineStringParser());
```
adamvy: Typically what we do is extend ProxyParser and in the constructor do `super(new Alt(...))`

Let's change to that so its consistent.

src/foam/lib/json/MultiLineStringParser.java
```diff
@@ -0,0 +1,80 @@
...
+
+    while ( ps.valid() ) {
+      char c = ps.head();
+
+      if ( c == delim && lastc != ESCAPE && ps.tail().head() == delim && ps.tail().tail().head() == delim ) break;
+
+      PStream tail = ps.tail();
+
+      if ( c == ESCAPE ) {
+        char   nextChar        = ps.tail().head();
```
adamvy: Seems like a lot of duplicated logic between the Multiline and the Singeline parser.

This seems a lot like just:

`Sequence(Literal("\"\"\""), Repeat(Not(Literal("\"\"\""), Alt(AsciiEscape, UnicodeEscape, AnyChar()))), Literal("\"\"\""))`

Rather than implementing the entire loop itself.

Upon further consideration though, I don't think our current StringParser even rejects newlines in strings.  I don't think there's any need for Multiline vs Sinlgeline parsers, the only real different is whether the string is delimited with a single quotation mark or a triple quotation mark so that people don't have to escape every quotation mark in the string.

If we tweak the existing StringParser to check if the string starts with a triple " in addition to the current checks for a single " or ', then we I think we can just continue to use that rather than adding new parser.

adamvy: In StringParser.java, instead of using a single char as a delim we can use a parser similar to how Repeat is implemented.  Something like

```
public class StringParser {
  private Parser delimiterParser = new Alt(new Literal("\""),
                                           new Literal("'"),
                                           new Literal("\"\"\""));
                                     
    ps = ps.apply(delimiterParser, x);
    if ( ! ps ) return null;

    Parser delimiter = new Literal((String)ps.getValue());

    PStream result;
    while ( ps.valid() ) {
      result = ps.apply(delimiter, x);
      if ( result != null ) break;
      ...
    }
  }
}
```

src/foam/dao/FileJournal.js
```diff
@@ -298,21 +298,17 @@ foam.CLASS({
...
-              if ( line == null )
-                return null;
-            }
+          if ( ! line.equals("p({") && ! line.equals("r({") ) return line;
+          if ( line == null ) return null;
+          StringBuilder sb = new StringBuilder();
+          sb.append(line);
+          while( ! line.equals("})") ) {
+            if ( (line = reader.readLine()) == null ) break;
+            sb.append("\\n");
```
adamvy, line 307: What happens if we append a regular newline rather than an escaped one?

src/foam/dao/FileJournal.js
```diff
@@ -298,21 +298,17 @@ foam.CLASS({
...
-            return null;
-          if ( getMultiLine() ) {
-            StringBuilder sb = new StringBuilder();
-            while ( ! line.equals("})") ) {
-              sb.append(line);
-              line = reader.readLine();
-              if ( line == null )
-                return null;
-            }
+          if ( ! line.equals("p({") && ! line.equals("r({") ) return line;
```
adamvy, line 301: If we were really clever we could remove the other condition and just collect lines until we find the next one starting with "p({" or "r({" but that would be a bit tricky to implement as we'd have to buffer lines locally or make a PushbackBufferedReader or something.

Not something to do now, but just a thought I had while reading.

PR #2376, opened 2019-07-09
Simplify validations

Nauna:
This PR goes with:
https://github.com/nanoPayinc/NANOPAY/pull/7067

and is to replace:
https://github.com/foam-framework/foam2/pull/2352

The plan was to do more of a cleanup then a quick fix.


src/foam/core/Validation.js
```diff
@@ -327,10 +327,36 @@ foam.CLASS({
...
+      class: 'FObjectArray',
+      of: 'foam.core.ValidationPredicate',
+      name: 'validationPredicates',
+      factory: function() {
+        if ( this.required ) {
+          var superFactory = foam.core.String.VALIDATION_PREDICATES.FACTORY;
+          return superFactory();
+        }
+        return [];
+      }
```
Nauna, line 375: Above compiles, but I am yet to test if return superFactory() does what I think.

adamvy: "compiles" is not much of a guarantee in JS.  If you haven't tested it yet then it's not ready to be submitted.  It's fine to start PRs before testing if you're looking for feedback on your implementation strategy, but be sure to mark such PRs as "DO NOT MERGE."

What are you trying to achieve?  It looks like you want the same validation predicates as foam.core.String but only if ths field is marked as required?  Is there any reason not to respect "required" in the foam.core.String validation predicates?

At the very least you probably have to do `foam.core.String.VALIDATION_PREDICATES.factory` and `superFactory.call(this);`  but I think we should re-think why we want to do this in the first place.

src/foam/nanos/auth/User.js
```diff
@@ -303,26 +260,18 @@ foam.CLASS({
       documentation: `The date and time that the current password of the User
         will expire.`,
     },
-    // TODO: startDate, endDate,
```
adamvy: Are both these TODOs actually addressed?

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -257,8 +247,8 @@ foam.CLASS({
     {
       name: 'validatePassword',
       javaCode: `
-        if ( SafetyUtil.isEmpty(potentialPassword) || ! (Pattern.compile(PASSWORD_VALIDATE_REGEX)).matcher(potentialPassword).matches() ) {
-          throw new RuntimeException(PASSWORD_VALIDATION_ERROR_MESSAGE);
+        if ( potentialPassword.length() < 6 ) {
+          throw new RuntimeException("PASSWORD_VALIDATION_ERROR");
```
adamvy: Why remove the isEmpty() check?  That protects against password being null.  Additionally why switch to a less informative error message?

src/foam/nanos/auth/User.js
Nauna: No. But there is nothing in the works for them. We have many ideas but not enough time to do them all.

If this becomes a requirement then we can do it.

This just seems like a 1 yr old sticky note.

adamvy: 1 year old sticky notes can still be important.  It's not a question of hard requirements but more just documented alternative design ideas that haven't been resolved.  I've found 3-4 year old TODOs that predicted the very bug that I was investigating.  I see no reason to clean them up

src/foam/nanos/auth/UserAndGroupAuthService.js
Nauna: Point taken.

PASSWORD_VALIDATION_ERROR_MESSAGE put back and isEmpty check added to if statement

The reason I changed was on the thought of simplifying. There is work that needs to be done with passwords outlined in CPF-2259

adamvy:
I'm all for simplification, in particular removing the regex is good, but the safety util check was important to prevent crashes.

src/foam/core/Validation.js
Nauna: First Objective: Simplify password validation to 6 char.
https://nanopay.atlassian.net/browse/CPF-2259 will correct the requirements. Which will be 7 char. But don't want to break anyone in production with a 6 char password right now.

So I added this here to confirm field validation on Passwords.

IF we do nothing but set the default minLength to 6, this breaks anywhere we are incorrectly using a User obj. Ex) CABankPADForm.js
validateInputs()  checks the errors on the user(this.viewData.user.errors_) despite only setting a first and last name.

Thus I tried to make my changes as unobtrusive as possible. Despite User having a password field and password is getting default setting is to have a minLength, I added that we only need this if field is required.

String (which Password extends) does not need to have this. If a minLength is set then there should be validation.

adamvy:
It makes sense to me for String validation the respect the "required" property.  A string field being option with a min and max length makes sense to me.  You don't have to provide it, but if you do then it must be within the specified min and max length.

src/foam/core/Validation.js
Nauna: AND

Sorry about not adding DNM to pr name.

Tested and changed. YOu were right:
foam.core.String.VALIDATION_PREDICATES.factory and superFactory.call(this);

src/foam/nanos/auth/User.js
Nauna: interesting. ok. reversed delete

src/foam/nanos/auth/User.js
```diff
@@ -99,16 +99,7 @@ foam.CLASS({
...
-      validateObj: function(firstName) {
-        if ( ! firstName.trim() ){
-          return 'First Name Required.'
-        } if ( firstName.length > 70 ) {
-          return 'First name cannot exceed 70 characters.';
-        } if( /\d/.test(this.firstName) ) {
-          return 'First name cannot contain numbers';
-        }
-      }
+      documentation: 'The first name of the User.'
```
mcarcaso, line 103: Don't you want a minLength here and below or are you ok with these fields being empty strings?

src/foam/core/Validation.js
mcarcaso: I'm not sure I follow why this is needed either. You should be able to simply set the minLength to 6 here and shouldn't need this additional code.

Also, I'm not sure we've really defined what the `required` flag should do wrt validation. Does it mean the property is set (i.e. enforces that hasOwnProperty returns true)? A property can be set to an empty string so what happens in that case? We should answer those before making use of it here.

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -257,7 +252,7 @@ foam.CLASS({
     {
       name: 'validatePassword',
       javaCode: `
-        if ( SafetyUtil.isEmpty(potentialPassword) || ! (Pattern.compile(PASSWORD_VALIDATE_REGEX)).matcher(potentialPassword).matches() ) {
+        if ( SafetyUtil.isEmpty(potentialPassword) || potentialPassword.length() < 6 ) {
```
mcarcaso, line 255: This method shouldn't be needed at all. If the userDAO has a ValidatingDAO around it, you shouldn't need to do any of this because the user with the updated password will be put into the dao and rejected.


...after looking closer at this service: It does password hashing before putting into the DAO so the DAO will allow it because it will likely hash to something that is valid. I don't think this service should be responsible for that. We already have a UserPasswordHashingDAO which should be doing this. If that decorator comes after the ValidatingDAO then I think you're good to go.

src/foam/nanos/auth/User.js
```diff
@@ -269,14 +234,7 @@ foam.CLASS({
...
       width: 100,
-      storageTransient: true,
-      validateObj: function (password) {
-        var re = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9]).{7,32}$/;
-
-        if ( password.length > 0 && ! re.test(password) ) {
-          return 'Password must contain one lowercase letter, one uppercase letter, one digit, and be between 7 and 32 characters in length.';
-        }
-      }
+      storageTransient: true
```
mcarcaso, line 238: What's the point of desiredPassword vs password?

src/foam/nanos/auth/User.js
Nauna: as default on the model, we are ok at least right now with empty string. Setting defaults here breaks things across the board. 

For example a contact is a user that does not need firstName

src/foam/nanos/auth/UserAndGroupAuthService.js
Nauna: I have no complaints , I opened a ticket to clean passwords. CPF-2259

src/foam/nanos/auth/User.js
Nauna: no idea. I'll add this comment to cpf-2259

Nauna: Just as a point of reference, despite having discusses this in person. 

desiredPassword is a string with the desired password, and password stores the current password as a hash of desiredPassword. (desiredPassword is set to null is set on userDAO.put)

src/foam/nanos/auth/User.js
Nauna: We did talk about this in person, but considering this involves making changes to **Contact** to ensure no requirements, and the fact that some places are using the **User** incorrectly. I'm going to open a ticket for this. And do this separately.

Nauna: ticket: https://nanopay.atlassian.net/browse/CPF-2461

Nauna:
-- this was done with work by Scott in PasswordPolicy

PR #2377, opened 2019-07-09
Give the FObjectView an "of".

mcarcaso:
And use it to provide a default for choices.

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -7,30 +7,38 @@
 foam.CLASS({
   package: 'foam.u2.view',
   name: 'FObjectView',
-  extends: 'foam.u2.Controller',
-
+  extends: 'foam.u2.View',
   documentation: 'View for editing FObjects.',
-
-  // css: '^:read-only { border: none; background: rgba(0,0,0,0); }'
-
   properties: [
     {
       class: 'String',
       name: 'objectClass',
-      displayWidth: 70,
-      view: { class: 'foam.u2.TextField', size: 80 },
+      label: '',
+      visibilityExpression: function(choices) {
+        return choices.length > 1 ?
+          foam.u2.Visibility.RW :
+          foam.u2.Visibility.HIDDEN;
+      },
+      view: function(_, x) {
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/FObjectView.js
```diff
@@ -39,34 +47,44 @@ foam.CLASS({
         }
       }
     },
-    'choices'
-  ],
+    'of',
+    {
+      name: 'choices',
+      expression: function(of) {
+        var models = Object.keys(foam.USED)
```
mcarcaso, line 54: This is pretty inefficient and I'm not sure if there's a better place to get all models that have been created. Is there a better way to walk all of our models?

src/foam/u2/view/FObjectView.js
```diff
@@ -7,30 +7,38 @@
...
+        return foam.u2.view.ChoiceView.create({
+          choices: x.data.choices
+        }, x);
+      },
       postSet: function(oldValue, newValue) {
         if ( newValue !== oldValue ) {
           var m = this.__context__.lookup(newValue, true);
           if ( m ) {
-            this.data = m.create(this.data, this);
+            var n = m.create(null, this);
```
adamvy, line 31: Why was this unrolled into multiple steps?  Not saying its wrong but curious what the difference was functionally.

src/foam/u2/view/FObjectView.js
adamvy: Would be more _efficient_ to cache this list and invalidate the cache every time a new model is loaded, but i'm not sure that's worth the effort.

mcarcaso: Yeah I was thinking the same thing. I was chatting with Kevin earlier and he suggested using foam.UNUSED too which I was afraid to do originally but it would be easy to add. Does anybody object to me going through all the UNUSED models too? It'll effectively flip them all to the USED pile though.

src/foam/u2/view/FObjectView.js
mcarcaso: Doing a copyFrom is quite a bit different from doing it directly in create (which invokes initArgs)

PR #2379, opened 2019-07-09
ChartJS Update [2.7.3 -> 2.8.0]

kennyqiyenkan:
Also added support for backgroundColor for PieDAOChartView


src/org/chartjs/PieDAOChartView.js
```diff
@@ -25,12 +25,15 @@ foam.CLASS({
...
-      name: 'keyExpr',
+      name: 'keyExpr'
     },
     {
       class: 'foam.mlang.ExprProperty',
-      name: 'valueExpr',
+      name: 'valueExpr'
     },
+    {
+      name: 'backgroundColor'
```
mcarcaso: Give this a default value so you don't need to check it before using it below and so you show an example of what it can take. Can this only be a function or can it also just be an array of colors or something?

kennyqiyenkan: One or the other. function or array of colours.

mcarcaso (approved):

mcarcaso (approved):

PR #2382, opened 2019-07-09
Fix bug where prop=value MQL queries wouldn't work

TW80000:
## Refs

https://nanopay.atlassian.net/browse/CPF-2099

## Changes

* Fixes a bug where "prop=value" MQL queries wouldn't work from the web client


jlhughes (approved):
Glad you found this, I thought I was nuts - I kept thinking my attempts at k=v were operator error.

PR #2385, opened 2019-07-10
Make RadioView respect view mode

TW80000:


src/foam/u2/view/RadioView.js
```diff
@@ -68,12 +81,13 @@ foam.CLASS({
         return this.E('div').
           // TODO: why is the radio item getting assigned the same class as the radio whole
           addClass(this.myClass()).
-          start('input').
+          start('input', null, ).
```
mcarcaso: What's with this change?

mcarcaso (approved):

src/foam/u2/view/RadioView.js
TW80000: Oh, thanks for catching that, my bad. I was going to do something like `.tag('input', null, this.someSlot$)` but ended up doing something else and must have forgotten to delete that. I'm surprised I didn't get any errors though, I tested it and it was working.

mcarcaso: hehe JS can be pretty forgiving at times....and super punishing other times :p

PR #2386, opened 2019-07-10
String multi line

michaelconnor468:


src/foam/dao/FileJournal.js
```diff
@@ -298,21 +303,18 @@ foam.CLASS({
...
           String line = reader.readLine();
-          if ( line == null )
-            return null;
+          if ( line == null ) return null;
+          StringBuilder sb = new StringBuilder();
+          sb.append(line);
           if ( getMultiLine() ) {
-            StringBuilder sb = new StringBuilder();
-            while ( ! line.equals("})") ) {
+            while( ! line.trim().endsWith("})") && ! line.trim().startsWith("//") ) {
```
adamvy, line 310: I don't agree with doing .trim().

This means that any comments inside the script will be lost.  Also any instance of }) will end the script early.

We also shouldn't build the StringBuilder before checking getMultiline() otherwise we'll be building a new Stringbuilder for every line in single-lined journals, that's wasteful.

I think what you want for this method is:

```
String line = reader.readLine();
if ( ! getMultiline() ) return line;
if ( line == null ) return null;
if ( line.endsWith("})") return line;

sb.append(line);

do {
  line = reader.readLine();
  if ( line == null ) return null;
  sb.append(line);
  sb.append("\n");
} while ( ! line.equals("})") );

return sb.toString();
```

In the non-multiline case this returns quickly with no extra work than before.

If we're in a multiline journal but the entry appears to be a single line entry `endsWith("})")` then we just return the line as is.

If we're in a multiline journal and the entry appears to be a multi-line entry (does not end with "})") then we buffer lines until we see a line that only consists of "})".

I don't think you need to escape newlines in the buffer, just insert real new-lines, the StringParser should handle them just fine.

src/foam/dao/FileJournal.js
```diff
@@ -327,14 +329,14 @@ foam.CLASS({
         JSONParser parser = getParser();
 
         try ( BufferedReader reader = getReader() ) {
-          for ( String entry ; ( entry = getEntry(reader) ) != null ; ) {
-            if ( SafetyUtil.isEmpty(entry)        ) continue;
-            if ( COMMENT.matcher(entry).matches() ) continue;
+          for ( String line ; ( line = getEntry(reader) ) != null ; ) {
```
adamvy: Why the revert back to line instead of entry?

src/foam/lib/json/StringParser.java
```diff
@@ -36,19 +42,21 @@ public StringParser() {
...
     StringBuilder builder = sb.get();
+    PStream result;
+    char lastc = ps.head();
 
     while ( ps.valid() ) {
       char c = ps.head();
 
-      if ( c == delim && lastc != ESCAPE ) break;
+      result = ps.apply(delimiter, x);
+      if ( result != null && lastc != ESCAPE )  break;
```
adamvy, line 59: Instead of isMultiline, you can just do `ps = result` here, and then `break`  Then outside of the loop we can just do `return ps.setValue(builder.toString());` because the delimiter parser will have consumed PStream till after the delimiter regardless of what it was.

adamvy: I think you can also clean up this logic a bit.  Only attempt to parse the delimiter if `lastc != ESCAPE` no use checking if the next characters are the string delimiter if we know we're going to ignore the result.

src/foam/dao/FileJournal.js
michaelconnor468: took it from a dated branch by accident; will fix

src/foam/lib/json/StringParser.java
michaelconnor468: im not sure what you mean by this

adamvy:
This PR was replaced by #2374 closing.

PR #2388, opened 2019-07-10
 #comment add x-content-type-options cache-control and pragma headers

noodlemoodle:


jlhughes (approved):

src/foam/nanos/http/NanoRouter.java
```diff
@@ -63,7 +63,12 @@ protected void service(final HttpServletRequest req, final HttpServletResponse r
     PM       pm         = new PM(this.getClass(), serviceKey);
 
     resp.setContentType("text/html");
-
+    resp.setHeader("X-Content-Type-Options", "nosniff");
```
kgrgreer, line 66: Please add one line comment for each of these to explain its purpose.

src/foam/nanos/http/NanoRouter.java
```diff
@@ -63,7 +63,12 @@ protected void service(final HttpServletRequest req, final HttpServletResponse r
...
+    resp.setHeader("Cache-Control", "no-cache, no-store, must-revalidate");
```
kgrgreer, line 67: Is this line correct? Don't we want some level of caching to improve performance? Currently we're downloading some small images hundreds of times for the same client session.

noodlemoodle: it looks like the headers are only being applied to responses to post requests

src/foam/nanos/http/NanoRouter.java
noodlemoodle: please review comments: https://github.com/foam-framework/foam2/pull/2390

PR #2391, opened 2019-07-10
Fix extra padding caused by nested SectionedDetailPropertyViews

TW80000:
## Changes

* ~Don't add padding to a `SectionedDetailPropertyView` if it has `SectionedDetailPropertyView`s as descendant elements.~
* Don't add top padding to the first visible `SectionedDetailPropertyView` in a `SectionView` and don't add bottom padding to the last visible one
  * This fixes a bug where `SectionedDetailPropertyView`s that were descendants of other `SectionedDetailPropertyView`s would cause unwanted extra whitespace, since both would be adding padding, but one contains the other.
* Added a period to an error message to be consistent with the other messages
* Add the phone number label on the property instead of labeling the FObjectProperty itself as "Phone #"


src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -355,5 +362,13 @@ foam.CLASS({
             .end();
         }));
     }
+  ],
+
+  listeners: [
+    function loaded() {
+      // Mark that we have descendant elements of the same class. If this is
+      // true, we don't want to add any padding to this element.
+      this.hasDescendants = this.el().querySelector('.foam-u2-detail-SectionedDetailPropertyView') !== null;
```
mcarcaso: What if we remove the padding around this Model completely and make it the SectionedDetailView's responsibility to put padding between the children. Don't put top padding on the first child or bottom padding on the last. This may be annoying to fix up everywhere but it feels more correct to me. What do you think?

TW80000: I can look into it. One thing to note though is that we have to deal with `VerticalDetailView` being nested as well, but we can just do the same thing.

mcarcaso: Ah yeah good catch. It wouldn't surprise me for others to be affected too and even if you don't catch them all here we can fix them as we see them.

TW80000: I've been trying this out and there are two tricky parts so far:

1. Selecting the last `.foam-u2-detail-SectionedDetailPropertyView` in a `.foam-u2-detail-SectionView` isn't possible with CSS since `:last-of-type` doesn't apply to CSS classes, just simple element selectors like `div` or `p`. So since we have other `div`s after the `.foam-u2-detail-SectionedDetailPropertyView`, that won't work. The recommended solution I saw on StackOverflow was to use JavaScript to mark the last one with an extra CSS class. Another idea is to switch from `div`s to `section`s for `.foam-u2-detail-SectionedDetailPropertyView`s. Even in that case, point 2 below is still an issue.
2. Knowing which one is the last one isn't as simple as I thought since each property is shown/hidden based on a visibility slot. So the last property in the array isn't necessarily the last _visible_ property in the array. We can figure this out with some work, but at that point I don't know if it's a better or worse solution than the one I have open in this PR currently.

TW80000: After talking with @mcarcaso in person, we want to try this:

* `SectionView` will create the visibility slots for each of its child `SectionedDetailPropertyView`s and pass them via a new `visibilitySlot` property on `SectionDetailedPropertyView`
  * We don't want to break anything else that depends on `SectionDetailedPropertyView` taking care of its own visibility, so if a `SectionDetailedPropertyView`'s `visibilitySlot` property is set then it'll use that, otherwise it'll make its own (like it's already doing). We could also just duplicate it and it would technically be fine, but it's more efficient just to create one slot instead of two duplicate ones.
  * Since `SectionView` has access to all of its children's visibility slots, it can figure out which of its children are the first and last visible ones and style them accordingly (no top-padding on the first visible child and no bottom-border on the last visible chid)

TW80000: Implemented in 7dab75a.

src/foam/u2/detail/SectionView.js
```diff
@@ -38,37 +55,74 @@ foam.CLASS({
...
+      expression: function(visibilitySlots_) {
+        return visibilitySlots_.get().lastIndexOf(true);
+      }
     }
   ],
   methods: [
     function initE() {
       var self = this;
-      this.addClass(this.myClass());
+      var slots = [];
```
mcarcaso: Hmm. This is going to grow larger every time section, showTitle, or section$title changes. You probably just want to declare this on line 98 and then set the ArraySlot just before returning.

src/foam/u2/detail/SectionView.js
```diff
@@ -38,37 +55,74 @@ foam.CLASS({
...
+        SectionedDetailPropertyViews don't cause unnecessary padding.
+      `,
+      factory: function() {
+        return this.ArraySlot.create();
+      }
+    },
+    {
+      class: 'Int',
+      name: 'firstVisibleIndex_',
+      expression: function(visibilitySlots_) {
```
mcarcaso: Instead of these expressions, what if you do this in initE()
```
var vSlot = foam.core.ProxySlot({delegate$: this.visibilitySlots_$});
this.onDetach(this.firstVisibleIndex_$.follow(vSlot.map(ar => ar.indexOf(true)))
this.onDetach(this. lastVisibleIndex_ $.follow(vSlot.map(ar => ar.lastIndexOf(true)))
```

src/foam/u2/detail/SectionView.js
```diff
@@ -38,37 +55,74 @@ foam.CLASS({
...
             })
             .start(self.Grid)
-              .forEach(section.properties, function (p) {
-                this.start(self.GUnit, { 
-                  columns: p.gridColumns 
-                })
-                  .tag(self.SectionedDetailPropertyView, { prop: p, data$: self.data$ })
-                .end()
+              .forEach(section.properties, function(p, index) {
+                var slot = self.SimpleSlot.create({}, self);
```
mcarcaso: Maybe rather than doing this, can we put this inside the `start` (before your enableClass calls)
```
.call(function() { slots.push(this.visibilitySlot) })
```

TW80000: Resolved by 19661fbb0796bd0e1cc90ec1b74d1e9690800f03.

src/foam/u2/detail/SectionView.js
TW80000: Resolved by 19661fbb0796bd0e1cc90ec1b74d1e9690800f03.

src/foam/u2/detail/SectionView.js
TW80000: What's the advantage of doing it that way over the current way? Currently it's clear from the property's definition that its value is an expression of `visibilitySlots_` and it also avoids the overhead of the ProxySlot.

mcarcaso: It's currently a one-time get. You're not actually subscribing to the visibilitySlots_

TW80000: Resolved by a40bdbd.

src/foam/u2/detail/SectionView.js
```diff
@@ -38,37 +55,71 @@ foam.CLASS({
...
+      name: 'lastVisibleIndex_'
     }
   ],
   methods: [
     function initE() {
       var self = this;
-      this.addClass(this.myClass());
       self.SUPER();
-      
+      var proxySlot = self.ProxySlot.create({ delegate$: self.visibilitySlots_$ });
```
mcarcaso: Sweet! Ok I think we can do a tad more cleanup here. First, remove all of these new properties.
Now this part becomes:
```
var proxySlot = self.ProxySlot.create({ delegate$: self.ConstantSlot.create({value: []}) });
var firstVisibleIndex = proxySlot.map(arr => arr.indexOf(true));
var lastVisibleIndex = proxySlot.map(arr => arr.lastIndexOf(true));
```
See below for comments on other lines as a result...

src/foam/u2/detail/SectionView.js
```diff
@@ -38,37 +55,71 @@ foam.CLASS({
...
-                  .tag(self.SectionedDetailPropertyView, { prop: p, data$: self.data$ })
-                .end()
+              .forEach(section.properties, function(p, index) {
+                this.start(self.GUnit, { columns: p.gridColumns })
+                  .start(self.SectionedDetailPropertyView, {
+                    prop: p,
+                    data$: self.data$
+                  })
+                    .call(function() { slots.push(this.visibilitySlot); })
+                    .enableClass('first', self.firstVisibleIndex_$.map((value) => value === index))
```
mcarcaso: use firstVisibleIndex variable here and lastVisibleIndex on next line.

src/foam/u2/detail/SectionView.js
```diff
@@ -38,37 +55,71 @@ foam.CLASS({
...
+                .end();
               })
             .end()
             .start(self.Cols)
-              .forEach(section.actions, function (a) {
+              .forEach(section.actions, function(a) {
                 this.add(a);
               })
             .end();
+          this.visibilitySlots_ = self.ArraySlot.create({ slots: slots });
```
mcarcaso: this becomes `proxySlot.delegate = self.ArraySlot.create({ slots: slots });`

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -262,28 +262,36 @@ foam.CLASS({
...
 
   methods: [
     function initE() {
       var self = this;
       this.SUPER();
+
       this
-        .show(this.prop.createVisibilityFor(this.data$)
-          .map(m => m != foam.u2.Visibility.HIDDEN))
+        .show(this.visibilitySlot)
```
mcarcaso: Probably not a big deal right now but if `prop` changes, this won't react to it. You'll need a proxy slot here if you want that reaction to work. Something like:
`.show(self.ProxySlot.create({delegate$: self.visibilitySlot$})`

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -262,28 +262,36 @@ foam.CLASS({
...
+    'foam.u2.layout.Rows',
+    'foam.u2.Visibility'
   ],
 
   properties: [
     'prop',
+    {
+      class: 'FObjectProperty',
+      of: 'foam.core.Slot',
+      name: 'visibilitySlot',
```
mcarcaso, line 275: Hmm thinking about this a little more right now. Would it make life simpler if we changed this to:
```
{
  class: 'Enum',
  of: 'foam.u2.Visibility',
  name: 'visibility'
}
```
and in `initE` we do:
```
this.onDetach(this.visibility$.follow(this.ProxySlot.create({
  delegate$: this.prop$.map(p => {
    p.createVisibilityFor(this.data$).map((m) => m !== this.Visibility.HIDDEN)
  }
// brackets...
```

Maybe I'm overthinking this lol

src/foam/u2/detail/SectionView.js
TW80000: Had to do

```
var proxySlot = self.ProxySlot.create({ delegate: self.ConstantSlot.create({value: []}) });
```

instead, setting `delegate` instead of `delegate$`. Otherwise we get the following error: 

Screen Shot 2019-07-22 at 9 10 27 AM


### Explanation (mostly for myself to make sure I really get it):

When you call `copyFrom` with a property slot as a key like `{ delegate$: someSlot }`, it calls https://github.com/foam-framework/foam2/blob/master/src/foam/core/Property.js#L424, and the `toSlot` part of that line creates a `PropertySlot` on whatever property you chose, in this case `delegate`. The type of the value in the `delegate` property is expected to be a `Slot`. When we `linkFrom` two slots, we say "These two slots should always have the same value. When one of them changes, update the other to match." Therefore when we link the `PropertySlot` for the `delegate` property to the `ConstantSlot` we create, we're trying to say: "The value of the `ConstantSlot` should always be the same as the value of the `PropertySlot` for the `delegate` property`". But that's not what we want because the value of the `ConstantSlot` is an array and the value of the `delegate` property should be a slot.

That also explains why it works with the other properties where the value of the property _is_ indeed a slot, such as `{ delegate$: self.visibilitySlot$ }`. In that case we can read it as "Let the `PropertySlot` of the `delegate` property to link to the `PropertySlot` of the `visibilitySlot` value, such that they always have the same value." Since the type of the value of both the `delegate` property and the `visibilitySlot` property is `Slot`, then when one changes we update the other. That's one level, the second level is that setting a property with a Slot as its value actually just `linkFrom`s the two  And the `postSet` on the `delegate` property on `ProxySlot` makes sure we're removing old listeners when the `delegate` changes.

Slot Diagram

TW80000: Resolved by fd57cd2.

src/foam/u2/detail/SectionView.js
TW80000: Resolved by fd57cd2.

src/foam/u2/detail/SectionView.js
TW80000: Resolved by fd57cd2.

src/foam/u2/detail/SectionedDetailPropertyView.js
TW80000: Resolved by b94d49d.

src/foam/u2/detail/SectionedDetailPropertyView.js
TW80000: ~The reason we needed to move the visibility slot into a property is so we could reference it from `SectionView`. I think if we made the change you're suggesting here we'd only be able to access the visibility, but not the visibility slot. Of course we could duplicate the logic but that's what I did initially and we changed it so that wasn't necessary.~

Never mind, I just realized we would still be able to reference `visibility$` from `SectionView`.

mcarcaso: Yeah but I think the way you have it is fine. I went too far lol

src/foam/u2/detail/SectionView.js
mcarcaso: Yes my bad that delegate$ shouldn't have been there in my code to you. Sorry about that. Nice diagram. I wanted to draw something up to help visualize it but couldn't think of a good way of doing it. Your diagram gets the point across. Nice :)

mcarcaso (approved):

PR #2392, opened 2019-07-11
#comment create write only dao

noodlemoodle:


src/foam/dao/PutOnlyDAO.js
```diff
@@ -0,0 +1,55 @@
...
+      {
+        name: 'remove_',
+        javaCode: `throw new UnsupportedOperationException("Cannot remove from PutOnlyDAO");`,
+        swiftCode: `throw FoamError("Cannot remove from PutOnlyDAO")`,
+        code: function remove_() {
+          return Promise.reject('Cannot remove from PutOnlyDAO');
+        }
+      },
+      {
+        name: 'cmd_',
```
jlhughes, line 38: cmd_ should be passed through to delegate.

jlhughes (changes requested):

src/foam/dao/PutOnlyDAO.js
smirnovaka: why? what if cmd reads and returns data?

PR #2395, opened 2019-07-12
Fix bug where limit was happening before authorization check

TW80000:
On calls to select_, AuthenticatedDAO decorates the sink with an AuthenticatedSink. The problem is that when the code eventually hits MDAO, it decorates the sink with a LimitedSink, which gets applied before the AuthenticatedSink instead of after.

For example, this means that if a user might have permission to see 12 objects in some DAO and they limit by 12, they might get less than 12 items back from the select, depending on the arbitrary order MDAO chooses to select items. It will pick the first 12 items and then do the auth checks, of which some might fail. The correct order is to first do the auth checks and then limit.


src/foam/dao/AuthenticatedDAO.java
```diff
@@ -99,27 +102,33 @@ public FObject find_(X x, Object id) {
   public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
     // x.get("auth") != null added so that system user can act during booting when the auth service is not loaded yet.
     if ( authenticateRead_ && x.get("auth") != null ) {
```
kgrgreer: If you move the authenticateRead_ && x.get()... check into augmentPredicate, it will make the code simpler and fix listen_.

TW80000: We don't want the `authenticateRead_` check to apply to `removeAll_`. I can move just the `x.get("auth") != null` check though.

src/foam/mlang/mlang.js
```diff
@@ -3480,6 +3480,43 @@ return true;
   ]
 });
 
+foam.CLASS({
+  package: 'foam.mlang.predicate',
+  name: 'HasPermission',
+  extends: 'foam.mlang.predicate.Binary',
```
kgrgreer: Should this extend Binary? It isn't a normal Binary mlang, in that each argument isn't a Function, it just has two arguments. Then the args could be properly typed.

TW80000: That was something I wasn't sure about. Should it just extend `AbstractPredicate` then?

kgrgreer:
I see.

kgrgreer:
yes

src/foam/mlang/mlang.js
TW80000: Resolved by 65ef3a0.

src/foam/dao/AuthenticatedDAO.java
TW80000: If I just remove the `x.get("auth") != null` check altogether, the server still runs and the tests still pass. Maybe whatever failure this was put in to fix has been fixed elsewhere since this was put in. Will just remove it instead.

TW80000: Resolved by 02d6fa3.

src/foam/dao/AuthenticatedDAO.java
```diff
@@ -97,29 +100,34 @@ public FObject find_(X x, Object id) {
 
   @Override
   public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
-    // x.get("auth") != null added so that system user can act during booting when the auth service is not loaded yet.
-    if ( authenticateRead_ && x.get("auth") != null ) {
-      super.select_(x, new AuthenticatedSink(x, createPermission("read"), sink), skip, limit, order, predicate);
+    if ( authenticateRead_ ) {
```
adamvy, line 103: So you no longer have the `x.get("auth") != null` check anywhere right? I 100% support that because I think it could lead to security problems, however the comment implies that we needed it during boot?  Is that no longer the case?

TW80000: I think so. I removed it and started the server and I could sign in to the portal. Then I tried running the test suite and it passed. So based on that I'm assuming we don't need it anymore.

PR #2396, opened 2019-07-12
Rule engine obj reloading

lchanmann:


src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -161,11 +163,17 @@ private void asyncApplyRules(List<Rule> rules, FObject obj, FObject oldObj) {
         if ( rule.getAsyncAction() != null
           && rule.f(x, obj, oldObj)
         ) {
+          // We assume the original object `obj` is stale when running after rules.
+          // For that, greedy mode is used for object reload. For before rules,
+          // object reload uses non-greedy mode so that changes on the original
+          // object will be copied over to the reloaded object.
+          FObject nu = getDelegate().find_(x, obj);
+          reloadObject(obj, oldObj, nu, rule.getAfter());
```
jlhughes: argument 'nu' is potentially reassigned in reloadObject, why does reloadObject not return the appropriate object?

jlhughes (changes requested):

src/foam/nanos/ruler/RuleEngine.java
lchanmann: Fixed.

jlhughes (approved):
Approved. Still request @kgrgreer reviews and approves.

src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -203,4 +211,61 @@ private void saveHistory(Rule rule, FObject obj) {
...
+      try {
+        old = obj.getClass().newInstance();
+      } catch (Exception e) {
+        // Object instantiation should not fail but if it does fail then return
+        // the original object as the reloaded object.
+        return obj;
+      }
+    }
+
+    old = old.fclone();
```
kgrgreer: If you put this line in an else clause for the above 'if' you'll save unnecessary cloneing when you create an object on line 238.

lchanmann: Oh yeah, you're right.

nanoscott (approved):

src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -161,11 +163,17 @@ private void asyncApplyRules(List<Rule> rules, FObject obj, FObject oldObj) {
         if ( rule.getAsyncAction() != null
           && rule.f(x, obj, oldObj)
         ) {
+          // We assume the original object `obj` is stale when running after rules.
```
adamvy, line 166: Can you describe a but that this fixes?  If your action is dependent upon whether the object has been written in to the DAO yet or not, then I think we're setting ourselves up for race conditions and your rule would be better of running in an agency at a specific time.  Or we should update the async logic to always fetch the latest object from the DAO, what state from the `obj` do you want that won't make it to `nu` ?

src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -203,4 +211,61 @@ private void saveHistory(Rule rule, FObject obj) {
...
+    if ( old == null ) {
+      try {
+        old = obj.getClass().newInstance();
+      } catch (Exception e) {
+        // Object instantiation should not fail but if it does fail then return
+        // the original object as the reloaded object.
+        return obj;
+      }
+    } else {
+      old = old.fclone();
```
jlhughes: @lchanmann I think Kevin mentioned this, it's not needed.

src/foam/nanos/ruler/RuleEngine.java
lchanmann: We have before rules (sync and async) that can update the object by setting its properties. When the next async rule is run the object might not have been saved since saving happen on a separate CMD operation https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/ruler/RulerDAO.js#L208.

As discussed with @kgrgreer, @jlhughes, and @nanoscott, we would have this kind of race condition because there is no synchronization inside the rule engine itself which is something needs to be addressed properly. On the other hand, we would need this temporary fix for our current release to not block the operation team from using IdentityMind integration.

I will be working on creating integration tests for IdentityMind then refactoring the rule engine to support object locking so that reloading would just be a simple `find()`.

lchanmann:
Ticket for adding synchronization on rule engine - https://github.com/foam-framework/foam2/issues/2401.

@kgrgreer can we have this PR in?

PR #2397, opened 2019-07-13
FObjectArrayView and StringArrayView enhancements

mcarcaso:
Create a CollapseableDetailView and a DraftDetailView and use them in the FObjectArrayView.

The resulting behaviour is that an FObjectArrayView just renders a row for each element in it that has a button that can be clicked to show that value. Once edited, the user must click the save button to persist the changes.

Also update StringArrayView to use ArrayView

Demo

### DIFF
--- a/src/foam/u2/view/CollapseableDetailView.js
+++ b/src/foam/u2/view/CollapseableDetailView.js
@@ -1,0 +1,84 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'CollapseableDetailView',
+  extends: 'foam.u2.View',
+  documentation: `
+    A detail view that initially just renders the summary of the object
+    with a "Show" button that, when clicked, shows a detail view of that
+    object.
+  `,
+  requires: [
+    'foam.u2.layout.Cols',
+  ],
+  properties: [
+    {
+      class: 'Boolean',
+      name: 'isCollapsed',
+      value: true
+    },
+    {
+      class: 'foam.u2.ViewSpec',
+      name: 'view',
+      value: { class: 'foam.u2.DetailView' }
+    },
+    {
+      class: 'String',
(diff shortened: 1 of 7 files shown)


mcarcaso:
I originally wanted to use a TableView for the FObjectArrayView but that got complicated since not all objects have an ID and the TableView would require me to put objects into a DAO. Maybe I'll revisit that some day but I think these are all useful views to have kicking around but let me know if you disagree.

src/foam/u2/view/CollapseableDetailView.js
```diff
@@ -0,0 +1,79 @@
...
+      name: 'view',
+      value: { class: 'foam.u2.DetailView' }
+    },
+    {
+      class: 'foam.u2.ViewSpec',
+      name: 'citationView',
+      expression: function(data) {
+        return {
+          class: 'foam.u2.CitationView',
+          of: data && data.cls_
```
adamvy, line 36: Does citation view need an of? It doesn't appear to use it.

src/foam/u2/view/DraftDetailView.js
```diff
@@ -0,0 +1,69 @@
...
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'DraftDetailView',
+  extends: 'foam.u2.View',
+  documentation: `
+    A detail view that holds off on updating the given object until the user clicks save.
```
adamvy, line 12: I would drop a TODO somewhere in here that we need nested property change event support for this to be truly robust.  Otherwise if you change a property of property of your data, you won't update hasDiff.

adamvy (approved):

src/foam/u2/view/CollapseableDetailView.js
mcarcaso: My understanding is that it's needed for the facet to work. Am I wrong on that?

src/foam/u2/view/DraftDetailView.js
mcarcaso: Good call. Done.

PR #2404, opened 2019-07-15
PieDAOChartView Colouring

kennyqiyenkan:
Refactors `backgroundColor` property to `palette` which takes an array/map/function.

If array:
- Cycles through the array and loops depending on number of data

If map:
- Colours section of the pie according to the provided key colour in the map
- If `default` is given in map, keys which have not been accounted for will use that colour
- If no `default` is given, 'lightgray' will be used

If function: 
- must be a function that takes two parameters (keys, data).
- must be a function that returns an array of colours.

This refactor should add flexibility in how we would represent data in the pie chart. Some possible representations include:
- Cycle through a colour palette (so not everything will be represented by the same colour)
- Each key would be represented by a specific colour (USD -> blue, CAD -> red)
- Colour of section would also reflect the range in which the data sits in. (red for data <= 33%, yellow for data between 34% - 66%, green for data > 66%)

### DIFF
--- a/src/org/chartjs/PieDAOChartView.js
+++ b/src/org/chartjs/PieDAOChartView.js
@@ -78,7 +78,7 @@ foam.CLASS({
               // Pass keys and data in their correct order to method passed to be
               // processed
               backgroundColors = self.palette(keys, data);
-            } else {
+            } else if ( backgroundColors.length === 0 ) {
               // Fallback if all else fails. Should never be the case.
               data.forEach(function(_, index) {
                 backgroundColors.push('lightgray');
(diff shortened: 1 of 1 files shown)


src/org/chartjs/PieDAOChartView.js
```diff
@@ -58,12 +55,42 @@ foam.CLASS({
             // a stack overflow.
             self.config.data = { datasets: [] };
             var config = foam.Object.clone(self.config);
+
+            var keys = Object.keys(sink.groups);
+            var data = [];
+            var backgroundColors = [];
+            keys.forEach(key => {
+              data.push(sink.groups[key].value);
+              if ( ! Array.isArray(self.palette) && typeof self.palette === 'object' ) {
```
mcarcaso: foam.Object.isInstance

mcarcaso (approved):

mcarcaso (approved):

PR #2408, opened 2019-07-16
Added a read-only feature to the Rich Choice View

tharmaman:


src/foam/u2/view/RichChoiceView.js
```diff
@@ -285,8 +285,8 @@ foam.CLASS({
         })
         .addClass(this.myClass())
         .start()
-          .addClass(this.myClass('selection-view'))
-          .enableClass('disabled', this.mode$.map((mode) => mode === foam.u2.DisplayMode.DISABLED))
+          .enableClass(this.myClass('selection-view'), this.mode$.map(mode => mode !== foam.u2.DisplayMode.RO))
```
adamvy, line 288: You can also do
```
addClass(this.mode$.map(mode => mode === foam.u2.DisplayMode.RO ? 'selection-view' : null))
```

not sure if that's any more readable, but its an option :)

adamvy (approved):

src/foam/u2/view/RichChoiceView.js
TW80000, line 288: This is the only thing I'm not sure about. We want to always mark this with the `selection-view` CSS class so it can be styled by ancestor elements.

I'm assuming you made this change because some of the styles applying to it are overriding the `.disabled` styles when we don't want them to. In that case I think we should adjust the offending selector(s) to include `:not(.disabled)`, or override the default styles in the `.disabled` CSS by making the selector more specific.

PR #2409, opened 2019-07-16
Added a read-only option to the select tag

tharmaman:


src/foam/u2/tag/Select.js
```diff
@@ -84,9 +84,24 @@ foam.CLASS({
...
-          mode === foam.u2.DisplayMode.RO;
-      this.setAttribute('disabled', disabled);
+      var isReadOnly = mode === foam.u2.DisplayMode.RO;
+
+      /**
+       * For read-only, we need to make it disabled since it is a SELECT element,
+       * otherwise it will still act like a dropdown
+       */
+      var isDisabled = mode === foam.u2.DisplayMode.DISABLED || isReadOnly;
+      this.setAttribute('disabled', isDisabled);
```
adamvy: Where do you remove the stylings if the view transitions from RO to RW ?  Also you can't set the disabled attribute to false, it's truthiness is determined by it being set at all.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#disabled  "A Boolean attribute which, *if present*, ..."

It would be more readable to just do

```
this.enableClass(this.myClass('read-only'), mode === foam.u2.DisplayMode.RO);
this.setAttribute('disabled', mode == foam.u2.DisplayMode.RO || mode == foam.u2.DisplayMode.DISABLED)
```
 and put your styles into a css class.

adamvy: Ignore the comment on the "disabled" truthiness, FOAM2 fixes it up for you automatically, I didn't realize that.  But hte other comment still stands.

tharmaman: @adamvy so I tried it using enabled class before but the class gets clobbered out, so far works when I directly style it:
inspector
code

tharmaman: Or it also works if I give !important to each of the CSS stylings under .read-only but I don't think this is what we want to do:
importantWOrks

PR #2410, opened 2019-07-16
CPF-1978 Temp Fix for Scroll table view

carl-zzz:
https://nanopay.atlassian.net/browse/CPF-1978


TW80000 (changes requested):
I would prefer that we move the dynamic height calculation code to the DAO controller or somewhere else so we keep that behaviour where we want it. Just setting a max-height on the table only solves half of the problem.

PR #2411, opened 2019-07-16
Fixed both rich choice view and choice view to support read-only

tharmaman:


TW80000 (approved):
LGTM

src/foam/u2/view/ChoiceView.js
```diff
@@ -201,18 +203,29 @@ foam.CLASS({
...
+              mode$: self.mode$,
+              size$: self.size$
+            })
+              .attrs({ name: self.name })
+              .enableClass('selection-made', self.index$.map((index) => index !== -1))
+            .end();
+      
+        } else {
+          return self.E()
+            .start()
```
mcarcaso: Shouldn't need this extra start/end. `self.E().add(self.text$)` should do the trick.

mcarcaso (approved):

src/foam/u2/view/ChoiceView.js
tharmaman: Done.

PR #2413, opened 2019-07-17
Add support for PermissionedPropertyDAO

jlhughes:
Add PermissionPropertyDAO support 
Also it wraps AuthenticatedDAO so that AuthenticatedDAO acts on the unmodified object.

### DIFF
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -201,11 +201,6 @@ if ( getOrder() != null &&
   }
 }
 
-if ( getPermissioned() &&
-     ( getNSpec() != null && getNSpec().getServe() ) ) {
-  delegate = new foam.nanos.auth.PermissionedPropertyDAO.Builder(getX()).setDelegate(delegate).build();
-}
-
 if ( getAuthorize() ) {
   if ( foam.nanos.auth.Authorizable.class.isAssignableFrom(getOf().getObjClass()) ) {
     delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate);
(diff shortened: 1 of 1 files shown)


src/foam/dao/EasyDAO.js
```diff
@@ -211,6 +201,11 @@ if ( getOrder() != null &&
   }
 }
 
+if ( getPermissioned() &&
```
smirnovaka: do you think we need `permissioned` property? Wouldn't it be better to add it always without asking for explicit setup?

jlhughes: It defaults to true.  If we don't provide for turning it off then when one DAO setup that needs it can't use EasyDAO and will cause Cluster setup issues.

src/foam/dao/EasyDAO.js
```diff
@@ -236,6 +226,11 @@ if ( getNSpec() != null &&
   logger.warning("EasyDAO", getNSpec().getName(), "Served DAO should be Authenticated, Authorized, or ReadOnly");
 }
 
+if ( getPermissioned() &&
```
smirnovaka, line 229: should authorization go before permissioned property to avoid redundant checks in case authorizationDAO doesn't allow to read at ll?

TW80000: That can cause problems where `PermissionedPropertyDAO` might reset a property to its default value before the authorization logic called by `AuthorizationDAO` can access it.

Say a model `Foo` has a permissioned property `bar` of type `Boolean`, where the default value is false. Lets consider a `find`, for example. We have some instance of `Foo` we're trying to find and we know the real value of `bar` is `true`. `AuthorizationDAO` happens before `PermissionedPropertyDAO` and calls `super.find_` to get the object before running the authorization checks on it. That call to `super.find_` will trigger `PermissionedPropertyDAO.find_`, which clears `bar` back to its default value of false. The `authorizeOnRead` code then gets executed, but it contains `if ( getBar() ) { ... }`. The problem is that `getBar()` will always be false since it has already been reset.

---

There are other ways to solve this problem, but moving the `PermissionedPropertyDAO` outside of `AuthorizationDAO` is one of them.

PR #2415, opened 2019-07-17
Split service notificationDAO into local and non-local.

jlhughes:
Change SendSlackNotificationDAO constructor to only context so delegate can be set later (EasyDAO).


TW80000 (approved):
LGTM, although I didn't build and test.

I'd prefer if we start using the multi-line journal syntax since it's so much easier to read and review, but it's up to you if you want to make those changes in this PR or not.

src/services
```diff
@@ -1,11 +1,13 @@
...
+p({"class":"foam.nanos.boot.NSpec", "name":"localNotificationDAO", "lazy":false, "serve":false, "serviceScript":"return  new foam.dao.EasyDAO.Builder(x).setOf(foam.nanos.notification.Notification.getOwnClassInfo()).setIndex(new foam.core.PropertyInfo[] {foam.nanos.notification.Notification.GROUP_ID}).setAuthenticate(false).setPm(true).setSeqNo(true).setDecorator(new foam.nanos.notification.NotificationTemplateDAO.Builder(x).setDelegate(new foam.nanos.notification.SendNotificationDAO(x, new foam.nanos.notification.SendEmailNotificationDAO(new foam.nanos.notification.SendSlackNotificationDAO(x)))).build()).setJournalType(foam.dao.JournalType.SINGLE_JOURNAL).setJournalName(\"notifications\").build();"})
```
smirnovaka, line 8: `authenticate` here still would be true because that is the default for NSpec. I don't think we should set `authenticate` on EasyDAO as it is going to be removed. Authentication is handled on service level and should be set on NSpec

src/services
```diff
@@ -1,11 +1,13 @@
...
 
+p({"class":"foam.nanos.boot.NSpec", "name":"localLogMessageDAO", "lazy":false, "serve":false, "serviceScript":"new foam.dao.EasyDAO.Builder(x).setOf(foam.nanos.logger.LogMessage.getOwnClassInfo()).setAuthenticate(false).setPm(true).setSeqNo(true).setCreatedAware(true).setCreatedByAware(true).setOrder(new foam.mlang.order.Comparator[] {foam.mlang.MLang.DESC(foam.nanos.logger.LogMessage.CREATED)}).setDecorator(new foam.nanos.logger.RepeatLogMessageDAO.Builder(x).setDelegate(new foam.dao.FixedSizeDAO.Builder(x).setDelegate(new foam.nanos.logger.NotificationLogMessageDAO.Builder(x).build()).build()).build()).setInnerDAO(new foam.dao.WriteOnlyJDAO(x, foam.nanos.logger.LogMessage.getOwnClassInfo(), \"../logs/logs\" /*TODO System.getProperty(\"LOG_HOME\")+\"/logs\"*/)).build();"})
```
smirnovaka, line 3: Same `authenticate` comment here

PR #2417, opened 2019-07-17
Multi line write

michaelconnor468:


src/foam/dao/FileJournal.js
```diff
@@ -111,6 +112,11 @@ foam.CLASS({
       class: 'Boolean',
       name: 'multiLine'
     },
+    {
+      class: 'Boolean',
+      name: 'multiLinePut',
```
adamvy: I would rather name this "outputMultiline" we don't need to say its specific to put or not, that's an implementation detail of FileJournal.

adamvy: I also wouldn't default it to true, only a select few journals should output this way (scripts, tests, cronjobs) the rest should stick to single-line.

src/foam/lib/json/Outputter.java
```diff
@@ -96,17 +97,26 @@ protected void outputNull() {
   }
 
   public void outputString(String s) {
+    boolean hasNewLine = s.indexOf('\n') >= 0;
+    if ( multiLineOutput_ && hasNewLine ) writer_.append("\r\n  ");
```
adamvy: I think this would be much more readable as two completely separate code paths, rather than the same conditional every other line.

```
if ( multiLineOutput_ && s.indexOf('\n') >= 0 ) {
  <do multiline output>
  return
}
<existing single-line output code>
```

It's not worth trying to re-use every single shared line such as "writer_.append(escape(s))" at the cost of making the flow difficult to read.  Sometimes a small amount of duplicated code is better.

It's also better to only check indexOf('\n') when multiLineOutput_ is true, otherwise we pay the cost of searching the string every time even when we don't care if its true or not.

src/foam/lib/json/Outputter.java
```diff
@@ -96,17 +97,26 @@ protected void outputNull() {
...
+    if ( multiLineOutput_ && hasNewLine ) writer_.append("\"\"\r\n    ");
     writer_.append(escape(s));
+    if ( multiLineOutput_ && hasNewLine ) writer_.append("\r\n  \"\"");
     writer_.append("\"");
   }
 
   public String escape(String s) {
-    return s
-      .replace("\\", "\\\\")
+    boolean hasNewLine = s.indexOf('\n') >= 0;
```
adamvy: I would rather have two escape methods for single vs multiline strings, otherwise we're searching the whole string for newlines _again_ every time.

src/foam/lib/json/Outputter.java
```diff
@@ -96,17 +97,26 @@ protected void outputNull() {
...
+    s = s.replace("\\", "\\\\")
       .replace("\"", "\\\"")
       .replace("\t", "\\t")
       .replace("\n","\\n");
+    if ( multiLineOutput_ ) {
+      s = s.replace("\\n", "\r\n    ");
```
adamvy: We shouldn't turn line-feed in to carriage-return line feed.  Additionally if we're going to insert indentation in the outputter, we should ensure we also strip it in the parser, otherwise every time we update an object it will pick up additionall indentation.  Alternatively we should consider not adding extra indetation.

src/foam/lib/json/Outputter.java
```diff
@@ -96,19 +97,38 @@ protected void outputNull() {
   }
 
   public void outputString(String s) {
-    writer_.append("\"");
-    writer_.append(escape(s));
-    writer_.append("\"");
+    if ( multiLineOutput_ && s.indexOf('\n') >= 0 ) {
+      writer_.append("\n  ");
+      writer_.append("\"");
```
adamvy: This is really confusing to read, you can combine these append statements

instead of:
```
append("\"");
append("\"\"\n   ");
```
why not just
```
append("\"\"\"\n   ");
```

src/foam/lib/json/Outputter.java
```diff
@@ -96,19 +97,38 @@ protected void outputNull() {
...
+      writer_.append(escape(s));
+      writer_.append("\"");
+    }
   }
 
   public String escape(String s) {
-    return s
-      .replace("\\", "\\\\")
+    return s.replace("\\", "\\\\")
```
adamvy, line 114: The previous line break of
```
return s
    .replace()
    .replace()
...
```

is easier to read IMO, as all the replace() calls are lined up.

src/foam/dao/FileJournal.js
```diff
@@ -109,7 +110,23 @@ foam.CLASS({
...
-      name: 'multiLine'
+      name: 'multiLineOutput',
+      value: true
+    },
+    {
+      class: 'Object',
+      name:  'multiLineOutputFiles',
+      javaType: 'java.util.HashSet',
+      javaFactory: `
+        HashSet<String> outputFiles = new HashSet<String>();
```
adamvy: I don't think we should do this.  First of all we're hard coding naonpay specific file locations in to the generic FOAM framework.  Secondly we already have the multilineOutput boolean property, just let client code set that property if they want to enable multiline output.

src/foam/lib/json/Outputter.java
```diff
@@ -455,4 +493,12 @@ public void flush() throws IOException {
     if ( stringWriter_ != null ) stringWriter_.flush();
     if ( writer_ != null ) writer_.flush();
   }
+
+  public void makeMultiLine() {
```
adamvy: Could make this follow the standard get/set pattern that most properties follow.

src/foam/lib/json/Outputter.java
```diff
@@ -96,17 +97,33 @@ protected void outputNull() {
...
-      .replace("\n","\\n");
+    return s.replace("\\", "\\\\")
+            .replace("\"", "\\\"")
+            .replace("\t", "\\t")
+            .replace("\n","\\n");
+  }
+
+  public String escapeMultiline(String s) {
+    s = s.replace("\\", "\\\\")
+         .replace("\\\"", "\"")
```
adamvy: I don't think this one line is necessary, we don't need to unescape escaped quotation marks in the string.

src/foam/lib/json/Outputter.java
```diff
@@ -96,17 +97,33 @@ protected void outputNull() {
...
+         .replace("\n","\\n")
```
adamvy: Don't unescape newlines, we're in a multiline output so newlines are okay

src/foam/lib/json/Outputter.java
```diff
@@ -96,17 +97,33 @@ protected void outputNull() {
...
+         .replace("\t", "\\t");
```
adamvy: don't need to escape tabs either I don't think.

src/foam/lib/json/Outputter.java
```diff
@@ -96,17 +97,33 @@ protected void outputNull() {
   }
 
   public void outputString(String s) {
-    writer_.append("\"");
-    writer_.append(escape(s));
-    writer_.append("\"");
+    if ( multiLineOutput_ && s.indexOf('\n') >= 0 ) {
+      writer_.append("\n  \"\"\"\n    ");
+      s = ltrim(s, 4);
```
adamvy: So we're going to trim the 4 space indentation off the string before indentation?  I don't understand why.

adamvy:
Can you make sure to test this by running the app, modifying a script on the scriptDAO and saving it and ensuring that it comes out reasonable in the journal?

Better yet, add a test to the testDAO that parses then outputs a multiline serialization of a Script object and ensure that the output matches the input.

src/foam/lib/json/Outputter.java
michaelconnor468: this is to prevent the indentation from piling up as it's read and outputted; you said this was a problem with adding indentation so it checks to see if it already indented and removes it if needed

src/foam/lib/json/Outputter.java
michaelconnor468: this was there before; I did not add this in so someone probably put it there for a reason

src/foam/lib/json/Outputter.java
michaelconnor468: this is just to add indentation formatting when we run s.replace("\\n", "\n    "); that way all new lines are indented

src/foam/lib/json/Outputter.java
michaelconnor468: we don't, but since the string is multiline the quatation marks don't need to be escaped and it makes things harder to read

michaelconnor468: this simply cleans up the formatting to make it more readable

src/foam/dao/FileJournal.js
```diff
@@ -199,18 +200,32 @@ foam.CLASS({
       synchronized: true,
       javaCode: `
         try {
+          String c = "";
+          if ( multiLineOutput ) {
```
adamvy: I don't think we should toggle multLine on the Outputter every put operation.  That's unnecessary, I would just update the javaFactory for Outputter to read the local multilineOutput property and set it up appropriately once at creation time.

src/foam/lib/json/Outputter.java
adamvy: This is the multiline escape.  What I'm saying is we don't need to escape tabs in multiline mode, we can just leave them as literal tabs.

src/foam/lib/json/Outputter.java
adamvy: I don't think this will work the way you are expecting it to work.  You're trimming the String that is being output at output time but instead we should be trimming away the indentation at read time.

If I have an object with a string proprety, and I start that string with four spaces I expect the JSON outputter to preserve those four spaces in its output so that when I read it back those four spaces are still there.  Similarly if I don't have 4 spaces, if I have 0 spaces or 30 spaces at the start of my line, then I expect those to still be there.  It's my data, the outputter shouldn't be modifying my data except in ways that the parser will undo at read time.  I don't think this trim fits that criteria since it always strips up to 4 spaces from my data, but the parser would then have no idea how many spaces to insert at parse time.

Are you sure this setup preserves all indentation in Scripts after being saved and re-read?

src/foam/lib/json/Outputter.java
adamvy: I still don't think we should be doing this, they may be pre-escaped for a reason.  If my script reads

```
print("\\" hello\\"")
```
this escaping will turn it in to
```
print("" hello"");
```
which is no longer valid code.

Additionally we shouldn't do unescaping in a method named "escape" ;)

src/foam/lib/json/Outputter.java
adamvy: Are you sure?  If you escape newlines to be \\n then later if you do .replace("\n", ...) it won't match anything since the newline is now escaped.

adamvy:
Let's add a nano test to this PR that makes an object with a bunch of different string properties.  In the properties put bunch of random code like content with various newlines, tabs, quotation marks, lines with indentation, lines without indentation, blank lines etc.  And then run it through Outputter in multiline mode and then parse it again and ensure that the String properties are identical.

src/tests
```diff
@@ -52,3 +52,5 @@ p({"class":"foam.nanos.test.Test","id":"FoldAndReduceTest","status":2,"code":"im
...
+p({"class":"foam.nanos.test.Test","id":"Multi Line Output","status":2,"code":"file = java.io.File.createTempFile(\"temp\", null);\nfile.deleteOnExit();\noutputter = new foam.lib.json.Outputter(x, file);\nstring = \"\\\\\\t line1 \\n   \\\\n  \\n line2 \\n line3 \\\" \\\\ \";\nuser = new foam.nanos.auth.User(x);\nuser.setId(24);\nuser.setFirstName(string);\nuser.setLastName(\"Fjorenborenborkenberger\");\n\noutputter.setMultiLine(true);\noutputter.outputJSONJFObject((foam.core.FObject) user);\noutputter.flush();\n\nfilejournal = new foam.dao.FileJournal(x);\nfilejournal.setMultiLineOutput(true);\nfilejournal.setFile(file);\n\nfirstread = filejournal.getEntry(filejournal.getReader());\nfirstread = firstread.substring(2, firstread.length() - 1);\n\nparser = new foam.lib.json.JSONParser();\nfirstparse = parser.parseString(firstread);\n\noutputter.outputJSONJFObject((foam.core.FObject) firstparse);\nsecondread = filejournal.getEntry(filejournal.getReader());\nsecondread = secondread.substring(2, secondread.length() - 1);\n\ntest(firstread.equals(secondread), \"Reparsing preserves JSON\");\n","output":""})
```
adamvy: Let's format this test as a multiline entry, since we now support that :).  Then it will be easier to review

PR #2418, opened 2019-07-18
Cpf 2293/authorization

noodlemoodle:
AuthorizationDAO and standardauthorizer 
- only daos already decorated with authorizationdao are updated
- disabled prevent privilege escalation tests to pass 
- nanopay pr https://github.com/nanoPayinc/NANOPAY/pull/7474


src/foam/dao/EasyDAO.js
```diff
@@ -213,10 +213,9 @@ if ( getOrder() != null &&
 
 if ( getAuthorize() ) {
   if ( foam.nanos.auth.Authorizable.class.isAssignableFrom(getOf().getObjClass()) ) {
-    delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate);
-    setAuthenticate(false);
+    delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate, new foam.nanos.auth.AuthorizableAuthorizer(getPermissionPrefix()));
   } else {
-    logger.warning("EasyDAO", "authorize=true but 'of' ",getOf().getId(), "not Authorizable");
+    delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate, new foam.nanos.auth.StandardAuthorizer(getPermissionPrefix()));
   }
```
jlhughes, line 211: It would be nice to still log this so we know we've made a configuration error.

jlhughes (commented):
This looks good.

src/foam/dao/EasyDAO.js
noodlemoodle: added warning back

jlhughes (approved):

jlhughes (approved):

src/foam/mlang/mlang.js
```diff
@@ -3502,8 +3501,14 @@ foam.CLASS({
       name: 'userContext'
     },
     {
-      class: 'String',
-      name: 'permissionPrefix'
+      class: 'Boolean',
+      name: 'remove'
```
kgrgreer: Document.

src/foam/mlang/MLang.java
```diff
@@ -194,10 +195,17 @@ public static Predicate HAS(Object o) {
       .build();
   }
 
-  public static Predicate HAS_PERMISSION(X userContext, String permissionPrefix) {
+  // used by AuthenticatedDAO
+  public static Predicate HAS_PERMISSION(X userContext, Boolean remove, String permissionPrefix) {
```
kgrgreer, line 199: Split into two. HAS_READ_PERMISSION and HAS_DELETE_PERMISSION.

src/foam/mlang/mlang.js
```diff
@@ -3502,8 +3501,14 @@ foam.CLASS({
       name: 'userContext'
```
kgrgreer, line 3502: Why do we need userContext? Why not just use object's get/setX()? Ask Troy.

src/foam/nanos/auth/AuthorizableAuthorizer.java
```diff
@@ -10,16 +10,16 @@
...
 
 public class AuthorizableAuthorizer implements Authorizer {
-  private static AuthorizableAuthorizer instance_ = null;
 
-  public static AuthorizableAuthorizer instance() {
-    if ( instance_ == null ) {
-      instance_ = new AuthorizableAuthorizer();
-    }
-    return instance_;
+  private String name;
```
kgrgreer: Rename to something more meaningful, lik permissionPrefix_. Make protected rather than private since private just makes testing and debugging more difficult.

src/foam/nanos/auth/AuthorizableAuthorizer.java
```diff
@@ -10,16 +10,16 @@
...
+
+  public AuthorizableAuthorizer(String name) {  
+    this.name = name;
```
kgrgreer: permissionPrefix_ = permissionPrefix;

src/foam/nanos/auth/AuthorizationDAO.java
```diff
@@ -53,29 +62,41 @@ public FObject put_(X x, FObject obj) throws AuthorizationException {
   @Override
   public FObject remove_(X x, FObject obj) {
     Object id = obj.getProperty("id");
-    FObject oldObj = getDelegate().inX(x).find(id);
-    if ( id == null || oldObj == null ) return null;
+    FObject oldObj = getDelegate().inX(x).find(id);
```
kgrgreer: Remove trailing spaces this line and next.

src/foam/nanos/auth/AuthorizationDAO.java
```diff
@@ -53,29 +62,41 @@ public FObject put_(X x, FObject obj) throws AuthorizationException {
...
     authorizer_.authorizeOnDelete(x, oldObj);
     return super.remove_(x, obj);
   }
 
   @Override
   public FObject find_(X x, Object id) {
+    if ( ! authorizeRead_ ) return super.find_(x, id);
+
     FObject obj = super.find_(x, id);
     if ( id == null || obj == null ) return null;
```
kgrgreer: Move id == null check above the find_(), or even first thing in the method.

src/foam/nanos/auth/Authorizer.java
```diff
@@ -18,4 +18,7 @@
   void authorizeOnRead(X x, FObject obj) throws AuthorizationException;
   void authorizeOnUpdate(X x, FObject oldObj, FObject newObj) throws AuthorizationException;
   void authorizeOnDelete(X x, FObject obj) throws AuthorizationException;
+  boolean checkGlobalRead(X x);
+  boolean checkGlobalRemove(X x);
+  String getPermissionPrefix();
```
kgrgreer: I don't think this belongs her.

src/foam/nanos/auth/AuthorizationDAO.java
```diff
@@ -53,29 +62,41 @@ public FObject put_(X x, FObject obj) throws AuthorizationException {
...
     return obj;
   }
 
   @Override
-  public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) {
-    super.select_(x, new AuthorizationSink(x, authorizer_, sink), skip, limit, order, predicate);
-    return sink;
+  public Sink select_(X x, Sink sink, long skip, long limit, Comparator order, Predicate predicate) { 
+
+    if ( authorizeRead_ && ! authorizer_.checkGlobalRead(x) ) predicate = augmentPredicate(x, false, predicate);
```
TW80000, line 85: I don't think the extra `checkGlobal_____` checks are buying us anything here or below in `removeAll_`. This way we _always_ need to check global read/delete every time before doing the other checks, which seems like unnecessary work. If I'm only doing custom authorization checks and not checking permissions, then this will always be unnecessary since the `foo.read.*` permission isn't granted to anyone (except for admin).

src/foam/nanos/auth/StandardAuthorizer.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package foam.nanos.auth;
+
+import foam.core.FObject;
+import foam.core.X;
+import foam.nanos.auth.AuthService;
+import foam.nanos.auth.AuthorizationException;
+
+public class StandardAuthorizer implements Authorizer {
```
kgrgreer, line 14: Add comment.

src/foam/nanos/auth/StandardAuthorizer.java
```diff
@@ -0,0 +1,94 @@
...
+
+  private String name;
```
kgrgreer: Rename to something more meaningful.

src/foam/mlang/MLang.java
TW80000, line 199: I think at this point it makes sense to rename this something like `IS_AUTHORIZED` since that's what it's really checking. Or maybe `IS_AUTHORIZED_TO_READ`.

src/foam/nanos/auth/SystemAuthService.java
```diff
@@ -0,0 +1,23 @@
...
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+package foam.nanos.auth;
+
+import foam.core.X;
+import foam.nanos.auth.User;
+
+public class SystemAuthService
```
kgrgreer, line 11: Document.

src/foam/nanos/auth/SystemAuthService.java
```diff
@@ -0,0 +1,23 @@
...
+  extends ProxyAuthService
+{
+  public SystemAuthService(AuthService delegate) {
+    setDelegate(delegate);
+  }
+
+  @Override
+  public boolean check(foam.core.X x, String permission) {
+    User user = (User) x.get("user");
+    return ( user != null && user.getId() == 1 ) ? true :  getDelegate().check(x, permission);
```
kgrgreer: Same as:
( user != null && user.getId() == 1 ) || getDelegate().check(x, permission);

src/foam/mlang/mlang.js
TW80000: The `remove` flag strikes me as a code smell. It might be better to split this up into `IS_AUTHORIZED_TO_READ` and `IS_AUTHORIZED_TO_DELETE`.

Then each of them does one thing and one thing only, and it's very clear what it's doing. Having totally different functionality hidden behind a flag variable/property can make things less clear.

src/foam/mlang/mlang.js
```diff
@@ -3521,9 +3526,14 @@ foam.CLASS({
...
       javaCode: `
         X x = (X) getUserContext();
-        String permission = getPermissionPrefix() + "." + ((FObject) obj).getProperty("id");
-        AuthService auth = (AuthService) x.get("auth");
-        return auth.check(x, permission);
+        foam.nanos.auth.Authorizer authorizer = getAuthorizer();
+        try {
+          if ( getRemove() ) authorizer.authorizeOnDelete(x, (FObject) obj);
+          else authorizer.authorizeOnRead(x, (FObject) obj);
+        } catch ( Exception e ) {
```
TW80000: Shouldn't this be

```
        } catch ( Throwable t ) {
```
? Same applies to elsewhere in this PR. I don't actually know what the argument for using `Throwable` over `Exception` is, I just know that's what I've seen elsewhere in our codebase and have been doing as well.

src/foam/nanos/auth/AuthorizableAuthorizer.java
TW80000: It would be more accurate to name this `permissionPrefix` based on its usage.

src/foam/mlang/MLang.java
```diff
@@ -194,10 +195,17 @@ public static Predicate HAS(Object o) {
       .build();
   }
 
-  public static Predicate HAS_PERMISSION(X userContext, String permissionPrefix) {
+  // used by AuthenticatedDAO
```
TW80000, line 198: Aren't we getting rid of `AuthenticatedDAO`?

src/foam/nanos/auth/AuthorizableAuthorizer.java
```diff
@@ -44,4 +44,28 @@ public void authorizeOnDelete(X x, FObject obj) throws AuthorizationException {
       ((Authorizable) obj).authorizeOnDelete(x);
     }
   }
+
+  public String createPermission(String op) {
+    return name + "." + op;
+  }
+
+  public boolean checkGlobalRead(X x) {
```
TW80000, line 53: See my comment below. I don't think `checkGlobalRead` and `checkGlobalRemove` are a good idea.

src/foam/nanos/auth/Authorizer.java
TW80000, line 23: This isn't applicable to all `Authorizer` implementations so I don't think it should be part of the interface.

src/foam/nanos/auth/SystemAuthService.java
TW80000: Slightly simpler:

```
    return ( user != null && user.getId() == 1 ) || getDelegate().check(x, permission);
```

src/foam/mlang/MLang.java
noodlemoodle: Changes made to HAS_PERMISSION used for authorizationdao
the one used by authenticatedDAO will no longer be used after its migrated. will remove this method after.

src/services
```diff
@@ -14,7 +14,7 @@ p({"class":"foam.nanos.boot.NSpec", "name":"http",                             "
...
+p({"class":"foam.nanos.boot.NSpec", "name":"auth",                             "lazy":true,  "serve":true,  "authenticate": false, "boxClass":"foam.nanos.auth.AuthServiceSkeleton", "serviceScript":"return new foam.nanos.auth.SystemAuthService.Builder(x).setDelegate(new foam.nanos.auth.twofactor.TwoFactorAuthService.Builder(x).setDelegate(new foam.nanos.auth.PasswordExpiryAuthService.Builder(x).setDelegate(new foam.nanos.auth.EnabledCheckAuthService.Builder(x).setDelegate(new foam.nanos.auth.CachingAuthService(new foam.nanos.auth.UserAndGroupAuthService(x))).build()).build()).build()).build();","client":"{\"class\":\"foam.nanos.auth.CachedAuthServiceProxy\",\"delegate\":{\"class\":\"foam.nanos.auth.ClientLoginAuthService\"}}"})
```
TW80000: Please consider converting to the new multi-line format for journal files with scripts in them. It makes it much easier to read and review. See an example here: https://github.com/foam-framework/foam2/blob/master/src/services#L42-L60.

src/foam/dao/EasyDAO.js
```diff
@@ -304,6 +305,16 @@ return delegate;
       name: 'authorize',
       value: false
     },
+    {
+      class: 'Boolean',
+      name: 'authorizeRead',
```
TW80000: Maybe `authorizeReads`?

src/foam/mlang/mlang.js
TW80000: I tried just using the `getX()` originally but it wouldn't work for some reason. I never ended up tracking down exactly why.

src/foam/nanos/auth/AuthorizableAuthorizer.java
noodlemoodle: renamed

src/foam/mlang/MLang.java
noodlemoodle: yes, this is being used for now since the services using authenticateddao are not migrated

src/services
noodlemoodle: converted

src/foam/mlang/mlang.js
noodlemoodle: removed flag and renamed method

src/foam/nanos/auth/AuthorizationDAO.java
smirnovaka: We should consider cases when system is forced to go through authorized DAO services(e.g. relationships), which would decrease performance if we were to apply auth.check on each record in select/removeAll.
> foo.read.* permission isn't granted to anyone

This confused me. I thought there were many read.* permissions granted to multiple groups

src/foam/mlang/mlang.js
smirnovaka: I don't think it's good. `Throwable` is a superclass for `Exception` as well as for `Error` which is subclass to errors like `StackOverflow`, `OutOfMemory` and everything else. Why would we ever want to catch those? I'd think that even catching `Exception` should be a rare occasion. In many cases when we set try/catch blocks, we know what to expect and that is the type that should be caught. In this specific case should be `AuthorizationException`. If something else goes wrong in the system we need to know

src/foam/nanos/auth/AuthorizationDAO.java
```diff
@@ -61,21 +71,35 @@ public FObject remove_(X x, FObject obj) {
 
   @Override
   public FObject find_(X x, Object id) {
+    if ( id == null ) return null;
+    if ( ! authorizeRead_ ) return super.find_(x, id);
+
     FObject obj = super.find_(x, id);
-    if ( id == null || obj == null ) return null;
+    if ( obj == null ) return null;
```
smirnovaka: `if ( obj != null ) authorizer_.authorizeOnRead(x, obj);`

src/foam/nanos/auth/AuthorizationDAO.java
```diff
@@ -61,21 +71,35 @@ public FObject remove_(X x, FObject obj) {
 
   @Override
   public FObject find_(X x, Object id) {
+    if ( id == null ) return null;
+    if ( ! authorizeRead_ ) return super.find_(x, id);
```
smirnovaka: can you please also check globalRead here

smirnovaka (changes requested):

src/foam/nanos/auth/AuthorizationDAO.java
TW80000: `! authorizeRead_` means we're not doing authorization checks for reads. And if we're not authorizing reads then there's no reason to check the global read permission here.

smirnovaka: the plan is to remove authorizeRead_

noodlemoodle: added

src/foam/dao/AuthenticatedDAO.java
```diff
@@ -123,11 +122,12 @@ public void pipe_(foam.core.X x, foam.dao.Sink sink, foam.mlang.predicate.Predic
...
 
   public Predicate augmentPredicate(X x, Predicate existingPredicate, String operation) {
-      return existingPredicate != null ?
-        AND(
-          HAS_PERMISSION(x, createPermission(operation)),
-          existingPredicate
-        ) :
-        HAS_PERMISSION(x, createPermission(operation));
+    boolean remove = operation.equals("delete");
+    return existingPredicate != null ?
```
kgrgreer, line 126: Doesn't this need to change to use the new MLang's?

src/foam/mlang/MLang.java
```diff
@@ -194,10 +195,23 @@ public static Predicate HAS(Object o) {
       .build();
   }
 
-  public static Predicate HAS_PERMISSION(X userContext, String permissionPrefix) {
-    return new HasPermission.Builder(null)
+  // used by AuthenticatedDAO
+  public static Predicate HAS_PERMISSION(X userContext, Boolean remove, String permissionPrefix) {
```
kgrgreer, line 199: Do we still need this one?

src/foam/dao/AuthenticatedDAO.java
noodlemoodle: this will get removed with the authenticatedDAO once everything is migratde right?

src/foam/mlang/MLang.java
noodlemoodle: this will also be removed once migrated to authorizationdao

src/foam/dao/AuthenticatedDAO.java
kgrgreer: ok

src/foam/mlang/MLang.java
kgrgreer: ok

src/foam/mlang/MLang.java
kgrgreer: ok

PR #2421, opened 2019-07-18
Fixed up the back to account for plural DAOs

tharmaman:


src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -131,7 +135,7 @@ foam.CLASS({
                     .tag(self.stack.BACK, {
                       buttonStyle: foam.u2.ButtonStyle.TERTIARY,
                       icon: 'images/back-icon.svg',
-                      label: `All ${config.of.name}s`
+                      label: `All ${backLabel}`
```
mcarcaso: Can you use foam.String.pluralize instead?

tharmaman: Done.

mcarcaso: I think you can remove the code above and simply do a `foam.String.pluralize(config$browseTitle)` here.
It'll look odd for "Liquidity Settings" because the result might be "Liquidity Settingses" but I don't think we should be accounting for that here.

PR #2422, opened 2019-07-18
Made delete in DAOSummary require confirmation and Removed border from buttons

tharmaman:


mcarcaso (approved):
No destructive colour on this guy huh? Logan ok with this?

tharmaman:
Yeah no destructive because eventually this one is supposed to have a modal

PR #2423, opened 2019-07-18
Reload NSpec service

lchanmann:
Reset service factory after the service setting is changed so that the nspec service is updated accordingly.

Addresses . https://nanopay.atlassian.net/browse/CPF-2149

### DIFF
--- a/src/foam/core/SingletonFactory.java
+++ b/src/foam/core/SingletonFactory.java
@@ -26,7 +26,7 @@ public class SingletonFactory
     return instance_;
   }
 
-  public void setDelegate(XFactory delegate) {
+  public synchronized void setDelegate(XFactory delegate) {
     delegate_ = delegate;
   }
 }
--- a/src/foam/core/SingletonFactory.java
+++ b/src/foam/core/SingletonFactory.java
@@ -19,10 +19,14 @@ public class SingletonFactory
   public synchronized Object create(X x) {
     if ( delegate_ != null ) {
       XFactory delegate = delegate_;
-      delegate_ = null;
+      setDelegate(null);
       instance_ = delegate.create(x);
     }
 
     return instance_;
+  }
+
+  public void setDelegate(XFactory delegate) {
+    delegate_ = delegate;
   }
 }
--- a/src/foam/nanos/boot/Boot.java
+++ b/src/foam/nanos/boot/Boot.java
@@ -10,8 +10,8 @@ import foam.core.*;
 import foam.dao.AbstractSink;
 import foam.dao.ArraySink;
 import foam.dao.DAO;
-import foam.dao.java.JDAO;
 import foam.dao.ProxyDAO;
+import foam.dao.java.JDAO;
 import foam.nanos.auth.Group;
 import foam.nanos.auth.User;
 import foam.nanos.logger.Logger;
@@ -27,12 +27,10 @@ import static foam.mlang.MLang.EQ;
 public class Boot {
   // Context key used to store the top-level root context in the context.
   public final static String ROOT = "_ROOT_";
-  public final static String RELOADED_SERVICES = "_RELOADED_SERVICES_";
 
   protected DAO serviceDAO_;
   protected X   root_ = new ProxyX();
   protected Map<String, NSpecFactory> factories_ = new HashMap<>();
-  protected Set<String> reloadedServices_ = new HashSet<>();
 
   public Boot() {
     this("");
@@ -67,16 +65,13 @@ public class Boot {
       root_.putFactory(sp.getName(), factory);
     }
 
-    root_.put(RELOADED_SERVICES, reloadedServices_);
     serviceDAO_.listen(new AbstractSink() {
       @Override
       public void put(Object obj, Detachable sub) {
         NSpec sp = (NSpec) obj;
 
         logger.info("Reload service factory:", sp.getName());
-        if ( factories_.get(sp.getName()).invalidate(sp) ) {
-          reloadedServices_.add(sp.getName());
-        }
+        factories_.get(sp.getName()).invalidate(sp);
       }
     }, null);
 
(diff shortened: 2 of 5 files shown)


jlhughes (approved):

src/foam/core/SingletonFactory.java
```diff
@@ -19,10 +19,14 @@ public SingletonFactory(XFactory delegate) {
...
       XFactory delegate = delegate_;
-      delegate_ = null;
+      setDelegate(null);
       instance_ = delegate.create(x);
     }
 
     return instance_;
   }
+
+  public void setDelegate(XFactory delegate) {
```
kgrgreer: Needs to be synchronized.

src/foam/nanos/boot/Boot.java
```diff
@@ -58,25 +61,21 @@ public Boot(String datadir) {
 
     for ( int i = 0 ; i < l.size() ; i++ ) {
       NSpec sp = (NSpec) l.get(i);
+      XFactory factory = new SingletonFactory(new NSpecFactory((ProxyX) root_, sp));
```
kgrgreer: Why do we need to decorate in a SingletonFactory? Doesn't NSpecFactory already cache the service?

kgrgreer: Just add an invalidate() method on the NSpecFactory and call it instead.

src/foam/nanos/boot/Boot.java
```diff
@@ -58,25 +61,21 @@ public Boot(String datadir) {
...
-          logger.info("Updating service configuration: ", sp.getName());
 
-          FObject service = (FObject) root_.get(sp.getName());
-          List<PropertyInfo> props = service.getClassInfo().getAxioms();
-          for (PropertyInfo prop : props) {
-            prop.set(service, prop.get(newService));
-          }
-        }
+        logger.info("Reload service factory:", sp.getName());
+        ((SingletonFactory) factories_.get(sp.getName()))
```
kgrgreer: Cast to an NSpecFactory and call invalidate() instead.

src/foam/nanos/boot/Boot.java
lchanmann: We are using SingletoneFactory when registering services adding invalidate() method on NSpecFactory would not help.

src/foam/core/SingletonFactory.java
lchanmann: Fixed.

src/foam/nanos/boot/Boot.java
lchanmann: Fixed.

src/foam/nanos/boot/Boot.java
lchanmann: Fixed.

lchanmann:
Addressed your comments. @kgrgreer

src/foam/nanos/boot/Boot.java
```diff
@@ -13,23 +13,25 @@
...
-import java.util.List;
+
+import java.util.*;
+
 import static foam.mlang.MLang.EQ;
 
 public class Boot {
   // Context key used to store the top-level root context in the context.
   public final static String ROOT = "_ROOT_";
+  private final static Set<String> RELOADED_SERVICES = new HashSet<>();
```
kgrgreer: Why is this static? This would prevent me from booting two NANOS servers in the same VM. Can't it just be an instance variable?

kgrgreer: What is this Set used for?

lchanmann: It needs a way to tell if a service is reloaded so that the already loaded web agents and nano services can be reloaded - https://github.com/foam-framework/foam2/pull/2423/commits/9649657c7c8f74325a77ce9d6ac498133088a18d.

> Why is this static? This would prevent me from booting two NANOS servers in the same VM.

It is not an instance variable because we do not pass the instance of Boot to NanoRouter/NanoServiceRouter.

I'm open to change. What are the other alternative you have in mind?

lchanmann: It's for storing the reloaded service name for use in reloading web agents and nano services.

lchanmann: Changed to `reloadedServices_` instance variable and store it in the context to be accessible in NanoRouter and NanoServiceRouter- https://github.com/foam-framework/foam2/pull/2423/commits/9a7e40c31279e26dad07658ca1101d65125d5150.

lchanmann:
Addressed your comments. @kgrgreer Please take another look, thanks.

jlhughes:
@lchanmann can you explain the reloadedServices_ , I'm still not clear on its purpose reading through the code.

lchanmann:
We need a way to communicate with NanoRouter/NanoServiceRouter that a web agent or nano service needs to be reloaded since these routers also have a cached version of it - https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/http/NanoRouter.java#L109.

So reloadedServices_ would store the service name that was invalidated - https://github.com/lchanmann/foam2/blob/ReloadNSpecService/src/foam/nanos/boot/Boot.java#L78 and NanoRouter/NanoServiceRouter can act accordingly.

jlhughes:
@lchanmann got it, thanks.

src/foam/nanos/boot/Boot.java
```diff
@@ -13,23 +13,26 @@
...
-import java.util.List;
+
+import java.util.*;
+
 import static foam.mlang.MLang.EQ;
 
 public class Boot {
   // Context key used to store the top-level root context in the context.
   public final static String ROOT = "_ROOT_";
+  public final static String RELOADED_SERVICES = "_RELOADED_SERVICES_";
```
kgrgreer: Have Nanorouter listen to changes in the NSpecDAO and get rid of this.

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2423/commits/e5fed73a234dbc4a10fdf6540fb95a181405f07a.

lchanmann:
Addressed your comments. @kgrgreer

lchanmann:
There is a [NanoServiceRouter.java](https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/box/NanoServiceRouter.java) but I'm not quite sure when it's being used so leaving it unchanged.

PR #2425, opened 2019-07-20
Switch to Agency

lchanmann:
~Stub thread pool would not execute context agent upon submission. To actually run the submitted agents, the client needs to call `invokeAll()` manually.~

## Changes

- Replace ThreadPool with Agency
- Rename FixedThreadPool to ThreadPoolAgency
- Add AgentAgency (it was StubThreadPool before)
- Update "threadPool" service setting


jlhughes:
Who (or will) calls invokeAll?

lchanmann:
> Who (or will) calls invokeAll?

I plan to use it for IdentityMind integration test. It allows us to invoke the real (async) actions before performing the assertions. For example:

```java
((IdentityMindService) getIdentityMindService()).setDefaultProfile("ACCEPT");
Transaction tx = newTransaction(1000);
tx = (Transaction) ((DAO) getTransactionDAO()).put(tx);

test(findTxApprovalRequest(tx) == null,
  "No approval request is created yet before the IdentityMind transaction rule is run");

// Invoke IdentityMind validator
((StubThreadPool) getThreadPool()).invokeAll();

// Assertions
ApprovalRequest found = findTxApprovalRequest(tx);
test(found != null && found.getApprover() == getIdentityMindUserId() && found.getStatus() == ApprovalStatus.APPROVED,
  "An approval request with status=APPROVED is created when the IdentityMind ACCEPT-ed the transaction");

tx = (Transaction) ((DAO) getTransactionDAO()).find(tx);
test(tx.getStatus() == TransactionStatus.COMPLETED, "Compliance transaction should be COMPLETED");
```

src/foam/nanos/pool/ThreadPool.java
```diff
@@ -10,5 +10,5 @@
 import foam.core.X;
 
 public interface ThreadPool {
```
kgrgreer: Isn't this just the same as Agency?

kgrgreer:
Isn't threadpool the same interface as Agency and FixedThreadPool should just be an Agency implementation?

lchanmann:
Added your suggestions. @kgrgreer

src/foam/nanos/pool/AgentAgency.java
```diff
@@ -0,0 +1,24 @@
...
+  Queue<Runnable> agents = new LinkedList<>();
+
+  @Override
+  public void submit(X x, ContextAgent agent, String description) {
+    agents.add(new ContextAgentRunnable(x, agent, description));
+  }
+
+  public void execute() {
+    while ( ! agents.isEmpty() ) {
+      agents.poll().run();
```
kgrgreer: Why poll?

lchanmann: Fixed.

PR #2427, opened 2019-07-22
Remove background colour so the error states are less "jarring"

mcarcaso:
image


adamvy:
I think we should go further, change the text colour too and remove the exclamation mark

src/foam/core/Validation.js
```diff
@@ -187,7 +187,7 @@ foam.CLASS({
             predicateFactory: function(e) {
               return e.GTE(self, self.min);
             },
-            errorString: `${self.label} must be at least ${self.min}.`
+            errorString: `Please enter a number greater than or equal to ${self.min}.`
```
mcarcaso: these should probably have the label in them so an error from the server is more actionable

src/foam/core/Validation.js
```diff
@@ -315,7 +315,7 @@ foam.CLASS({
             predicateFactory: function(e) {
               return e.REG_EXP(self, /^$|.+@.+\..+/);
             },
-            errorString: `${this.label} is not an email address`
+            errorString: 'Please enter a valid email address'
```
adamvy: "Please enter an email address in the style of person@example.com"

or just

"Please enter an email address"

Drop the work  "valid"

src/foam/nanos/auth/Phone.js
```diff
@@ -22,7 +22,7 @@ foam.CLASS({
               foam.nanos.auth.Phone.NUMBER,
               /^(?:\+?1[-.●]?)?\(?([0-9]{3})\)?[-.●]?([0-9]{3})[-.●]?([0-9]{4})$/);
           },
-          errorString: 'Invalid phone number.'
+          errorString: 'Please enter a valid phone number.'
```
adamvy: Oh dear god, what is our validation for phone numbers?  I would say

"Please enter a phone number in the style of 555-555-5555 or +1-555-555-5555"

Saying "Please enter a valid phone number" doesn't help anyone as we dont' tell them what makes it valid, I think give an useful example or drop the validation.

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -89,8 +88,6 @@ foam.CLASS({
     ^error .foam-u2-DateView,
     ^error .foam-u2-view-date-DateTimePicker .date-display-box
     {
-      color: /*%DESTRUCTIVE3%*/ #d9170e !important;
-      background-color: /*%DESTRUCTIVE5%*/ #fbedec !important;
       border-color: /*%DESTRUCTIVE3%*/ #d9170e !important;
```
adamvy, line 91: Should border-color still be red as well?

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -332,6 +329,7 @@ foam.CLASS({
                     .start(self.Cols)
                       .addClass(self.myClass('validation-container'))
                       .show(errorSlot)
+                      /*j
```
adamvy: j

src/foam/nanos/auth/Phone.js
jlhughes: This will have to be locale specific.  The current regex is only north america. 
(**but not today!**)  So for now, I would vote for 'Please enter a phone number'. 
Can we not handle with and without dashes/periods and add them in if not provided?

src/foam/core/Validation.js
mcarcaso: went with "Please enter an email address"

src/foam/nanos/auth/Phone.js
mcarcaso: Went with "Please enter a phone number". I think the regex handles lack of dashes/periods. I'm sure that adding them in after the fact if they're missing is doable but I'm not gonna do that in this PR.

src/foam/u2/detail/SectionedDetailPropertyView.js
mcarcaso: I'll leave the border color. I think it's valuable for scrolling through a long detail view and being able to catch them. If somebody disagrees it's an easy enough change to make.

mcarcaso:
> I think we should go further, change the text colour too and remove the exclamation mark

I'm going to leave this PR as is. If somebody wants to go further they're more than welcome to.

PR #2429, opened 2019-07-22
#comment add custom authorizable to notifications

noodlemoodle:
foam pr for added notification permissions https://github.com/nanoPayinc/NANOPAY/pull/7521


jlhughes (approved):
Perfect.

src/foam/nanos/notification/Notification.js
```diff
@@ -114,5 +120,52 @@ foam.CLASS({
...
+      ],
+      type: 'Boolean',
+      javaCode: `
+        foam.nanos.auth.User user = (foam.nanos.auth.User) x.get("user");
+        if( user == null ) return false;
+
+        Long userId = user.getId();
+        String groupId = user.getGroup();
+
+        return getUserId() != userId && getGroupId() != groupId;
```
smirnovaka: `grroupId` property is only used as a helper to define what users should receive notifications but it is reset to null prior being saved to journals. You should only check for userId, this line will always return false

src/foam/nanos/notification/Notification.js
```diff
@@ -114,5 +120,52 @@ foam.CLASS({
...
+      `
+    },
+    {
+      name: 'authorizeOnCreate',
+      javaCode: `
+      AuthService auth = (AuthService) x.get("auth");
+      if ( ! auth.check(x, "*") ) throw new AuthorizationException("You don't have permission to create this notification.");
```
smirnovaka: please create permissions for create/read rather than *

smirnovaka (changes requested):

smirnovaka:
Great job @noodlemoodle, requested some changes. Also still thinking about making it more generic and use it for other services too. @jlhughes?

src/foam/nanos/notification/Notification.js
noodlemoodle: added permissions https://github.com/nanoPayinc/NANOPAY/pull/7506

src/foam/nanos/notification/Notification.js
```diff
@@ -114,5 +120,49 @@ foam.CLASS({
...
+      javaCode: `
+      AuthService auth = (AuthService) x.get("auth");
+      if ( ! auth.check(x, "*") ) throw new AuthorizationException("You don't have permission to delete notifications you do not own.");
+      `
+    },
+    {
+      name: 'authorizeOnRead',
+      javaCode: `
+      AuthService auth = (AuthService) x.get("auth");
+      if ( ! checkOwnership(x) && ! auth.check(x, "notificationDAO.read.*") ) throw new AuthorizationException("You don't have permission to read notifications you do not own.");
```
jlhughes: @penzital @noodlemoodle Is authorizeOnRead called on every object put in the sink on select()?  Curious how checkOwnership works?

noodlemoodle: Yes, the authorizer containing this is passed into the authorizationDAO which adds HAS_PERMISSION as part of the predicate in select. HAS_PERMISSION calls the authorizeOnRead on every object before putting it into the sink I believe. this predicate was added in this pr https://github.com/foam-framework/foam2/pull/2395
so in this case a read on a notification would be granted if the notification belong to current user or current user has admin permission.

smirnovaka: authorizeOnRead is called on every object unless current user has global read permission

smirnovaka: you don't need to check for global read permission here, `AuthorizableAuthorizer.checkGlobal` will do it for you. This method won't be even called if user has `notification.read.*` in these methods you actually need to check for `read.id`

smirnovaka (changes requested):
this pr will be refactored anyway

src/foam/nanos/notification/Notification.js
noodlemoodle: oh true true

PR #2430, opened 2019-07-22
Added Support to JDBC DAO (only MySQL now)

skanderturki:
I have added support for JDBC DAO with put and update support only for now,
Only works with MySQL for now because of SQL specific syntax.


skanderturki:
I signed it!

jlhughes:
Looks great. 
A trivial comment - perhaps replace the 
`} catch (Throwable e) {
      e.printStackTrace();`
with 
`} catch (Throwable e) {
      foam.nanos.logger.Logger logger = (foam.nanos.logger.Logger) x.get("logger");
      logger.error(e); `

src/foam/dao/JDBCConnectionSpec.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2019

skanderturki:
Ok Joel Thanks

src/foam/dao/JDBCConnectionSpec.js
```diff
@@ -0,0 +1,33 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+/* A JDBC Connection Specification to add to the system context */
```
kgrgreer: Add comment inside the model instead of as a comment so that it's available to our tools.
Ex.:
documentation: 'A JDBC Connection Specification to add to the system context.'

src/foam/dao/JDBCConnectionSpec.js
```diff
@@ -0,0 +1,33 @@
...
+foam.CLASS({
+  package: 'foam.dao',
```
kgrgreer: Maybe a jdbc sub-package like foam.dao.jdbc instead. In time we could have a lot of JDBC related models and it would be better that they all be in one spot and not pollute the main dao package.

src/foam/dao/JDBCConnectionSpec.js
skanderturki: OK

src/foam/dao/JDBCConnectionSpec.js
skanderturki: OK

src/foam/dao/JDBCConnectionSpec.js
```diff
@@ -0,0 +1,33 @@
...
+    {
+      class: 'String',
+      name: 'databaseName'
+    },
+    {
+      class: 'String',
+      name: 'userName'
+    },
+    {
+      class: 'String',
```
kgrgreer: Make be a password so that it isn't stored in plain-text:

class: 'Password',

src/foam/dao/JDBCPooledDataSource.java
```diff
@@ -0,0 +1,96 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2019

src/foam/dao/JDBCPooledDataSource.java
```diff
@@ -0,0 +1,96 @@
...
+
+import javax.sql.DataSource;
+import java.sql.DriverManager;
+
+public class JDBCPooledDataSource {
+
+  protected static String SQL_POOL_NAME = "SQLPool";
+
+  // Holds a reference to the connection pool ( .getConnection() )
+  protected static DataSource dataSource_;
```
kgrgreer: Why is this static?

src/foam/dao/JDBCPooledDataSource.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private JDBCPooledDataSource(){
+    // Nothing
+  }
+
+  public static DataSource getDataSource(X x) throws ClassNotFoundException, java.sql.SQLException{
+    try {
+      if(dataSource_ == null){
```
kgrgreer: See the FOAM style guide for spacing rules:

src/foam/dao/JDBCPooledDataSource.java
skanderturki: the dataSource_ object holds a pool of connections so I made it static so that all the users of the class use the same pool

src/foam/dao/JDBCPooledDataSource.java
skanderturki: ok

src/foam/dao/JdbcDAO.java
```diff
@@ -0,0 +1,384 @@
...
+import java.util.*;
+
+/*
+  We assume that the database is created by a script when the system starts the first time.
+  When the system restarts at any point in time it should verify that the database is already created otherwise it should create it.
+  When creating the database, we only do a CREATE SCHEMA sql instruction, then we create the app user and password and grant him full privileges on this database.
+  Any other database objects (tables) will be created on the fly by the application (when methods of this class are called).
+
+ */
+public class JdbcDAO extends AbstractDAO{
```
kgrgreer: Jdbc -> JDBC, capitalize acronyms, like DAO

src/foam/dao/JdbcDAO.java
```diff
@@ -0,0 +1,384 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2019

src/foam/dao/JdbcDAO.java
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
+    Connection c = null;
+    IndexedPreparedStatement stmt = null;
+    ResultSet resultSet = null;
+
+    try {
+      c = dataSource_.getConnection();
```
kgrgreer: Should get dataSource and connection from Context 'X', rather than using statics. Static values restrict composability. What happens if we want to connect to multiple data-sources in the same system? With the context I can register multiple instances, but using static variables precludes me from doing this.

src/foam/dao/JdbcDAO.java
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    return obj;
+  }
+
+  /**
+   * Returns list of properties of a metaclass
+   * @param of ClassInfo
+   */
+  private void getObjectProperties(ClassInfo of){
```
kgrgreer: Generally, making methods or variables private doesn't provide extra protection and makes testing and debugging more difficult. Especially for a method like this which is just a helper method and isn't granting any private access to internal structures.

skanderturki:
Ok, all requested changes were resolved.

src/foam/core/AbstractArrayPropertyInfo.java
```diff
@@ -6,17 +6,14 @@
 
 package foam.core;
 
-import foam.dao.pg.IndexedPreparedStatement;
+import foam.dao.jdbc.IndexedPreparedStatement;
```
kgrgreer, line 9: Why is this import needed?

src/foam/core/AbstractPropertyInfo.java
```diff
@@ -6,8 +6,7 @@
 
 package foam.core;
 
-import foam.dao.pg.IndexedPreparedStatement;
-import foam.lib.xml.Outputter;
+import foam.dao.jdbc.IndexedPreparedStatement;
```
kgrgreer, line 9: Why is this needed?

src/foam/dao/DAO.js
```diff
@@ -83,6 +87,7 @@ foam.INTERFACE({
       async: true,
       type: 'FObject',
       swiftThrows: true,
+      documentation: 'Retrieves a single object from the DAO, whose id is equal to the parameter id. If the object is found, the promise resolves with the object. If the object is not found, it rejects with a foam.dao.ObjectNotFoundException.',
```
kgrgreer: I don't think this is true. If not found it just returns null rather than throw an Exception. Also, you can specify the id or a predicate.

src/foam/dao/DAO.js
```diff
@@ -95,6 +100,7 @@ foam.INTERFACE({
...
```
kgrgreer: No need to duplicate the description for find(), since then future updates will require updating in two spots and could get out of date. Better to just say this is the internal version of find().

src/foam/dao/jdbc/AbstractJDBCDAO.java
```diff
@@ -0,0 +1,145 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.dao.jdbc;
+
+import foam.dao.AbstractDAO;
+
```
kgrgreer: I prefer to just group all imports together without spaces between lines.

src/foam/dao/jdbc/AbstractJDBCDAO.java
```diff
@@ -0,0 +1,145 @@
...
+import foam.core.*;
+
+import javax.sql.DataSource;
+import java.sql.SQLException;
+import java.sql.*;
+
+import java.lang.Exception;
+import java.util.*;
+
+public abstract class AbstractJDBCDAO extends AbstractDAO{
```
kgrgreer, line 22: Add documentation.

src/foam/dao/jdbc/AbstractJDBCDAO.java
```diff
@@ -0,0 +1,145 @@
...
+  /** Holds a reference to the connection pool ( .getConnection() ) */
+  protected static DataSource dataSource_;
+
+  /**
+   * Create the table in the database and return true if it doesn't already exist otherwise it does nothing and returns false
+   * @param of ClassInfo
+   */
+  public abstract boolean createTable(X x, ClassInfo of);
+
+  public AbstractJDBCDAO(X x, ClassInfo of, String poolName) throws java.sql.SQLException, ClassNotFoundException {
```
kgrgreer: I don't think the DAO should need to know the poolName. Can this class get access the dataSource from the context and not know what pool it is using?

src/foam/core/AbstractPropertyInfo.java
skanderturki: It is not, I made a refactoring to put JDBC classes into foam.dao.jdbc instead of foam.dao.pg. I'll remove it

src/foam/dao/jdbc/AbstractJDBCDAO.java
```diff
@@ -0,0 +1,145 @@
...
+
+    // Get the system global dataSource with its system global pool
+    JDBCPooledDataSource jp = new JDBCPooledDataSource(x, poolName);
+    dataSource_ = jp.getDataSource(x, poolName);
+
+    tableName_ = of.getObjClass().getSimpleName().toLowerCase();
+
+    getObjectProperties(of);
+
+    if ( ! createTable(x, of) ) {
```
kgrgreer: No need to check for error, since it will usually already be created. Maybe better to rename createTable() to maybeCreateTable() to indicate that it may or may not create the table.

src/foam/core/AbstractArrayPropertyInfo.java
skanderturki: Not needed, I made a refactoring of packages

src/foam/dao/jdbc/AbstractJDBCDAO.java
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Returns list of properties of a metaclass
+   * @param of ClassInfo
+   */
+  protected void getObjectProperties(ClassInfo of){
+
+    if(properties_ == null) {
+      List<PropertyInfo> allProperties = of.getAxiomsByClass(PropertyInfo.class);
+      properties_ = new ArrayList<PropertyInfo>();
+      for (PropertyInfo prop : allProperties) {
```
kgrgreer: See: https://github.com/foam-framework/foam2/blob/master/doc/guides/StyleGuide.md

src/foam/dao/DAO.js
skanderturki: Ok, who should change this in the word documentation?

src/foam/dao/DAO.js
skanderturki: ok

src/foam/dao/jdbc/AbstractJDBCDAO.java
skanderturki: ok

src/foam/dao/jdbc/AbstractJDBCDAO.java
skanderturki: ok

src/foam/dao/jdbc/AbstractJDBCDAO.java
skanderturki: ok

src/foam/dao/jdbc/JDBCPooledDataSource.java
```diff
@@ -0,0 +1,100 @@
...
+
+    try {
+      //Add code to get JDBCConnectionSpec properties
+      Object spec = x.get("JDBCConnectionSpec");
+
+      if( spec == null ) {
+        throw new java.sql.SQLException("JDBC connection information not found");
+      }
+
+      if ( !(spec instanceof JDBCConnectionSpec) ) {
```
kgrgreer: Just cast to JDBCConnectionSpec on line 30 and the system will throw the ClassCastException for you and this block can be removed.

src/foam/dao/jdbc/JDBCPooledDataSource.java
```diff
@@ -0,0 +1,100 @@
...
+        throw new ClassCastException("Expecting a JDBCConnectionSpec but obtained instead: " + spec.toString());
+      }
+
+      JDBCConnectionSpec jdbcSpec = (JDBCConnectionSpec) spec;
+
+      //jdbc:mysql://localhost/db?useUnicode=true&useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=UTC
+      String connectionURI = "jdbc:" + jdbcSpec.getDatabaseServer() + "://" + jdbcSpec.getHostName() +
```
kgrgreer: Would make more sense for this to be a method on JDBCConnectionSpec.

src/foam/dao/jdbc/JDBCPooledDataSource.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.sql.DataSource;
+import java.sql.DriverManager;
+
+public class JDBCPooledDataSource {
+
+  // Holds a reference to the connection pool ( .getConnection() )
+  protected DataSource dataSource_;
+  protected String poolName_;
+
+  public JDBCPooledDataSource(X x, String poolName) {
```
kgrgreer: Why do we need to supply the poolName?

src/foam/dao/jdbc/JDBCPooledDataSource.java
```diff
@@ -0,0 +1,100 @@
...
+      Logger logger = (Logger) x.get("logger");
+      logger.error(e);
+    } catch (java.sql.SQLException e) {
+      Logger logger = (Logger) x.get("logger");
+      logger.error(e);
+    }
+
+  }
+
+  public DataSource getDataSource(X x, String poolName) {
```
kgrgreer: Why does the client need to supply the poolName? Wasn't this instance configured with a specific poolname? If we want to client (DAO) to have a different pool we can just put them in a different sub-context with a different DataSource.

src/foam/dao/jdbc/MySQLJDBCDAO.java
```diff
@@ -0,0 +1,333 @@
...
+
+ */
+public class MySQLJDBCDAO extends AbstractJDBCDAO{
+
+  public MySQLJDBCDAO(X x, ClassInfo of, String poolName) throws java.sql.SQLException, ClassNotFoundException {
+    super(x, of, poolName);
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
```
kgrgreer: Is there anything MySQL specific about this put_() method? Could it be moved to AbstractJDBDCDAO?

src/foam/dao/jdbc/MySQLJDBCDAO.java
```diff
@@ -0,0 +1,333 @@
...
+import java.util.*;
+
+/**
+  We assume that the database is created by a script when the system starts the first time.
+  When the system restarts at any point in time it should verify that the database is already created otherwise it should create it.
+  When creating the database, we only do a CREATE SCHEMA sql instruction, then we create the app user and password and grant him full privileges on this database.
+  Any other database objects (tables) will be created on the fly by the application (when methods of this class are called).
+
+ */
+public class MySQLJDBCDAO extends AbstractJDBCDAO{
```
kgrgreer: Could we just rename this to MySQLDAO?

src/foam/dao/jdbc/MySQLJDBCDAO.java
```diff
@@ -0,0 +1,333 @@
...
+    } catch (Throwable e) {
+      Logger logger = (Logger) x.get("logger");
+      logger.error(e);
+      return null;
+    } finally {
+      closeAllQuietly(null, stmt);
+    }
+  }
+
+  @Override
```
kgrgreer: Is this MySQL specific?

src/foam/dao/jdbc/MySQLJDBCDAO.java
```diff
@@ -0,0 +1,333 @@
...
+      Logger logger = (Logger) x.get("logger");
+      logger.error(e);
+      return null;
+    } finally {
+      closeAllQuietly(resultSet, stmt);
+    }
+  }
+
+  @Override
+  public FObject remove_(X x, FObject obj) {
```
kgrgreer: Is this MySQL specific?

src/foam/dao/jdbc/MySQLJDBCDAO.java
```diff
@@ -0,0 +1,333 @@
...
+    } finally {
+      closeAllQuietly(resultSet, stmt);
+    }
+  }
+
+  /**
+   * Prepare the formatted column names. Appends column names like: (c1,c2,c3)
+   * @param builder builder to append to
+   */
+  public void buildFormattedColumnNames(FObject obj, StringBuilder builder) {
```
kgrgreer: Is this MySQL specific?

src/foam/dao/jdbc/MySQLJDBCDAO.java
```diff
@@ -0,0 +1,333 @@
...
+      }
+    }
+    builder.append(")");
+  }
+
+  /**
+   * Prepare the formatted UPDATE string like :  description='etc', name='etc'
+   * @param builder builder to append to
+   */
+  public void buildUpdateFormattedColumnNames(FObject obj, StringBuilder builder) {
```
kgrgreer: Is this MySQL specific?

src/foam/dao/jdbc/MySQLJDBCDAO.java
```diff
@@ -0,0 +1,333 @@
...
+      }
+    }
+
+  }
+
+  /**
+   * Prepare the formatted value placeholders. Appends value placeholders like: (?,?,?)
+   * @param builder builder to append to
+   */
+  public void buildFormattedColumnPlaceholders(FObject obj, StringBuilder builder) {
```
kgrgreer: Is this MySQL specific?

src/foam/dao/jdbc/PostgresJDBCDAO.java
```diff
@@ -4,25 +4,29 @@
...
 import java.sql.*;
+
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
 // TODO: Create AbstractJDBCDAO baseclass
-public class PostgresDAO
-  extends AbstractDAO
+public class PostgresJDBCDAO
```
kgrgreer: Let's rename to just PostgresDAO.

src/foam/dao/jdbc/PostgresJDBCDAO.java
```diff
@@ -42,27 +46,8 @@ public StringBuilder get() {
...
-      if ( "".equals(prop.getSQLType()) )
-        continue;
-      props_.add(prop);
-    }
-
-    table_ = of.getObjClass().getSimpleName().toLowerCase();
-    if ( ! createTable() ) {
-      //throw new SQLException("Error creating table");
-    }
+  public PostgresJDBCDAO(X x, ClassInfo of, String poolName) throws java.sql.SQLException, ClassNotFoundException {
```
kgrgreer: I don't think it needs to know the poolName.

src/foam/java/Interface.js
```diff
@@ -99,6 +99,11 @@ foam.CLASS({
 
       o.out('\n');
 
+      if( this.documentation ){
```
kgrgreer: Fix spacing.

src/foam/java/Interface.js
```diff
@@ -128,6 +133,21 @@ foam.CLASS({
 
       o.decreaseIndent();
       o.out('}');
+
     }
   ]
 });
+
+
+function limitSplit(o, str, maxWords){
```
kgrgreer: Document.

src/foam/java/Interface.js
```diff
@@ -128,6 +133,21 @@ foam.CLASS({
...
+    var res = '';
+    var arr = str.split(' ', 150);
+    for( i = 0 ; i < arr.length ; i++){
```
kgrgreer: spacing

src/foam/java/Interface.js
```diff
@@ -128,6 +133,21 @@ foam.CLASS({
...
+      res += arr[i] + ' ';
+      if( i % maxWords == 0 && i > 0 ){
```
kgrgreer: spacing

src/foam/java/InterfaceMethod.js
```diff
@@ -71,6 +75,16 @@ foam.CLASS({
   methods: [
     function outputJava(o) {
       o.indent();
+
+      if( this.documentation ){
```
kgrgreer: spacing, here and elsewhere

src/foam/java/InterfaceMethod.js
```diff
@@ -90,6 +104,23 @@ foam.CLASS({
...
       o.out(';\n');
+
+      if( this.documentation ){
+        o.decreaseIndent();
+      }
     }
   ]
 });
+
+function limitSplit(o, str, maxWords){
```
kgrgreer: Is this method duplicated?

src/foam/java/Method.js
```diff
@@ -73,6 +83,25 @@ foam.CLASS({
...
+      if( this.documentation ){
+        o.decreaseIndent();
+      }
+
     }
   ]
 });
+
+
+function limitSplit(o, str, maxWords){
```
kgrgreer: third time we have this method defined

kgrgreer:
This is two independent PR's: one for JDBCDAO and one for documentation. In future, put separate functionality in separate PRs.

src/foam/core/AbstractArrayPropertyInfo.java
skanderturki: It is used by the setStatementValue method, I only made a refactoring of the packages

src/foam/core/AbstractPropertyInfo.java
skanderturki: It is used by the prepareStatement method, I only made a refactoring of the packages

src/foam/dao/DAO.js
skanderturki: Replaced by:
Retrieves a single object from the DAO, whose id is equal to the parameter id or if the given predicate evaluates to true.
If the object is found, the promise resolves with the object.
If the object is not found, it returns null.

src/foam/dao/jdbc/AbstractJDBCDAO.java
skanderturki: is "createTableIfNotCreated" too long?

src/foam/dao/jdbc/JDBCPooledDataSource.java
skanderturki: ok

src/foam/dao/jdbc/JDBCPooledDataSource.java
skanderturki: OK now it's:

JDBCConnectionSpec jdbcSpec = (JDBCConnectionSpec) spec;
String connectionURI = jdbcSpec.buildConnectionURI(); //(method declared in the model)

src/foam/dao/jdbc/JDBCPooledDataSource.java
skanderturki: So that the DAO chooses which pool to use, so we can have different DAOs use different pools if we want to have a DAO with high priority use its own exclusive pool

src/foam/dao/jdbc/MySQLJDBCDAO.java
skanderturki: yes, the select options are specific to MySQL ( when the object is already there, the update part that was wrote in postgresDAO does not comply with MySQL syntax)

src/foam/dao/jdbc/MySQLJDBCDAO.java
skanderturki: ok

src/foam/dao/jdbc/MySQLJDBCDAO.java
skanderturki: No, the find has nothing MySQL specific

src/foam/dao/jdbc/MySQLJDBCDAO.java
skanderturki: Apperently no, I will move them to the AbstractJDBCDAO until we find a MySQL specific detail

src/foam/dao/jdbc/MySQLJDBCDAO.java
skanderturki: no

src/foam/dao/jdbc/MySQLJDBCDAO.java
skanderturki: Yes this is the part used by the put that is specific to MySQL

src/foam/dao/jdbc/MySQLJDBCDAO.java
skanderturki: no

src/foam/dao/jdbc/PostgresJDBCDAO.java
skanderturki: ok

src/foam/java/Interface.js
skanderturki: ok

src/foam/java/Method.js
skanderturki: yes, just for testing, will be unified in a util file

src/foam/dao/jdbc/JDBCPooledDataSource.java
skanderturki: No need to provide the poolname, corrected this

src/files.js
```diff
@@ -159,6 +159,7 @@ FOAM_FILES([
   { name: "foam/json2/Outputter" },
   { name: "foam/json2/Serializer" },
   { name: "foam/json2/Deserializer" },
+  { name: "foam/java/Util" },
```
kgrgreer: change to:
{ name: "foam/java/Util", flags: ['java'] },

So that it is only included when building Java, not for the web.

skanderturki: OK

src/foam/dao/ORBitalDAO.java
```diff
@@ -0,0 +1,38 @@
+package foam.dao;
```
kgrgreer: Can you remove this file from this PR.

skanderturki: OK

src/foam/dao/jdbc/AbstractJDBCDAO.java
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  @Override
+  public FObject remove_(X x, FObject obj) {
+    Connection c = null;
+    IndexedPreparedStatement stmt = null;
+
+    try {
+      c = dataSource_.getConnection();
+      StringBuilder builder = threadLocalBuilder_.get()
```
kgrgreer: Why do we need to build this string every time? Isn't it always the same so we could reuse it? Same as for other prepared statements.

src/foam/dao/jdbc/AbstractJDBCDAO.java
```diff
@@ -0,0 +1,283 @@
...
+
+  }
+
+  /**
+   * Returns list of properties of a metaclass
+   * @param of ClassInfo
+   */
+  protected void getObjectProperties(ClassInfo of){
+
+    if( properties_ == null ) {
```
kgrgreer: space after 'if'. The name of this method is misleading since it isn't a getter and returns void. Maybe something like buildPropertyList().

src/foam/dao/jdbc/AbstractJDBCDAO.java
```diff
@@ -0,0 +1,283 @@
...
+      PropertyInfo prop = (PropertyInfo) i.next();
+      if ( "id".equals(prop.getName()) )
+        continue;
+
+      builder.append(prop.createStatement());
+      builder.append("='");
+      builder.append(prop.get(obj));  //add the new property value
+      builder.append("'");
+      if ( i.hasNext() ) {
+        builder.append(",");
```
kgrgreer: When adding a single character, I think it is faster to just append a character. Ex.: append(',');

src/foam/java/InterfaceMethod.js
kgrgreer: if ( this.documentation ) {

src/foam/java/InterfaceMethod.js
```diff
@@ -65,11 +71,30 @@ foam.CLASS({
...
 
   methods: [
     function outputJava(o) {
+
+      if( this.documentation ){
+       str = foam.java.Util.removeSpacing(this.documentation);
+       lines = foam.java.Util.limitSplit(str, 25);
+       o.indent();
+       o.out('/**\n');
+       for( i = 0 ; i < lines.length ; i++ ){
```
kgrgreer: fix spacing

src/foam/java/Method.js
```diff
@@ -38,6 +43,22 @@ foam.CLASS({
 
   methods: [
     function outputJava(o) {
+      o.out('\n');
+
+      if( this.documentation ){
```
kgrgreer: spacing, here and elsewhere

src/foam/dao/jdbc/AbstractJDBCDAO.java
skanderturki: Currently I do a builder.append("c") to append a character. Is there a faster way?

PR #2432, opened 2019-07-23
Make the ordering of axiom installation take the original order of ax…

mcarcaso:


adamvy:
I fixed this myself slightly differently.

PR #2435, opened 2019-07-24
Added in the ability to pass in a custom set of columns to the DAOBrowserView

tharmaman:


src/foam/comics/v2/DAOBrowserView.js
```diff
@@ -102,6 +102,7 @@ foam.CLASS({
         return this.DAOControllerConfig.create({ dao: this.data });
       }
     },
+    'columns',
```
mcarcaso: I think this should be part of the config. Make it a StringArray and call it defaultColumns. Use the model's tableColumns to prefill this.

tharmaman: But columns only really make sense depending on the view right? Like as the config doesn't make sense when its rendering a TreeView.

mcarcaso: Yeah TreeView doesn't get the config. This view does and already makes decisions based on it. I'm not sure why this would be any different.

tharmaman: Done.

src/foam/comics/v2/DAOControllerConfig.js
```diff
@@ -55,6 +55,16 @@ foam.CLASS({
...
       expression: function(of) { return foam.String.pluralize(of.model_.label); }
     },
+    {
+      class: 'StringArray',
+      name: 'defaultColumns',
+      factory: null,
+      expression: function(of) { 
+        return of.model_.tableColumns.length 
+                ? of.model_.tableColumns
+                : of.model_.properties;
```
mcarcaso: Sorry I may have mislead you on this a little cuz I think I showed you the model_ approach but I think we're better off doing:
```
var tableColumns = of.getAxiomByName('tableColumns');
return tableColumns ? tableColumns.columns :
  of.getAxiomsByClass(foam.core.Property).map(p => p.name);
```
Reason being, we'll get any inherited properties or table columns this way.

mcarcaso (approved):

src/foam/comics/v2/DAOControllerConfig.js
tharmaman: Done.

PR #2438, opened 2019-07-25
#CPF-1651 : action in label doesn't effect the checkbox

Nauna:
The **issue** arose when a labelFormatter was used that contained a link to open a new window.

The click on the link, toggled the checkbox. 

(**solution**)This PR separates the checkbox from the label, so an action taken on the label is seperate from the action on the checkbox.


Nauna:
We are going to do this another way.

PR #2439, opened 2019-07-25
#comment remove usages of authorizeRead, add GlobalReadAuthorizer tha…

noodlemoodle:
…t returns true on all authorization checks on read


src/foam/nanos/auth/GlobalReadAuthorizer.java
```diff
@@ -0,0 +1,33 @@
...
+  // this authorizer is an extension of the standardauthorizer that is meant
+  // to replace the authorizeReads flag on services using the standardAuthorizer.
+  // setAuthorize(true) => setAuthorizer(new foam.nanos.auth.GlobalReadAuthorizer(\"permissionPrefix\"))
+
+  public GlobalReadAuthorizer(String permissionPrefix) {
+    super(permissionPrefix);
+  }
+
+  @Override 
+  public void authorizeOnRead(X x, FObject obj) throws AuthorizationException {
```
smirnovaka: I think it's ok if class inherits this method from super since it's only called when checkGlobal returns false

src/foam/dao/EasyDAO.js
```diff
@@ -305,11 +305,6 @@ return delegate;
       name: 'authorize',
       value: false
     },
```
smirnovaka, line 299: can you add property so we can set authorizer on EasyDAO. If set then that's the authorizer we set to AuthorizationDAO, otherwise standard/authorizable

noodlemoodle: the authorizer property was added in previous authorization pr https://github.com/foam-framework/foam2/pull/2418/files

src/foam/nanos/auth/GlobalReadAuthorizer.java
noodlemoodle: makes sense will remove and add globalread check for find back to authorizationdao

smirnovaka (changes requested):
can you please add authorizer to EasyDAO and whatever services used to use authorizeRead should now be using GlobalReadAuthorizer

src/foam/dao/EasyDAO.js
```diff
@@ -202,12 +202,12 @@ if ( getOrder() != null &&
 }
 
 if ( getAuthorize() || getAuthorizer() != null ) {
```
smirnovaka: I think `getAuthorize()` should be separately

src/foam/dao/EasyDAO.js
```diff
@@ -201,13 +201,13 @@ if ( getOrder() != null &&
   }
 }
 
-if ( getAuthorize() || getAuthorizer() != null ) {
-  if ( getAuthorizer() != null ) delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate, getAuthorizer(), getAuthorizeReads());
+if ( getAuthorizer() != null ) delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate, getAuthorizer());
+else if ( getAuthorize() ) {
```
kgrgreer: Should this be 'else'? Could I do both?

smirnovaka: I think you need to check for getAuthorize first and if thats true, then you proceed otherwise do nothing

noodlemoodle: if you want to use the default authorizers you can extend them in the custom authorizer

smirnovaka:
lgtm.
@kgrgreer

PR #2440, opened 2019-07-25
#comment add capabilityauthservice to foam auth

noodlemoodle:


TW80000 (changes requested):
The `CapabilityAuthService` has to be outside of `CachingAuthService`, otherwise the cache might become invalid if a user's capabilities change during their session.

TW80000 (approved):
This is fine, but I want to keep an eye on performance. If I recall correctly, `CapabilityAuthService` had a significant negative impact on performance.

noodlemoodle:
Closing since not needed for now.

PR #2443, opened 2019-07-25
Cpf 2438/remove authenticated dao

noodlemoodle:
migrate services to use authorizationdao instead of authenticateddao and remove authenticateddao.
nanopay pr https://github.com/nanoPayinc/NANOPAY/pull/7542


src/services
```diff
@@ -161,11 +159,11 @@ p({"class":"foam.nanos.boot.NSpec", "name":"ruleDAO",                          "
...
 
-p({"class":"foam.nanos.boot.NSpec","name":"localGroupPermissionJunctionDAO","description":"The junction DAO for the many-to-many relationship between groups and permissions","serve":false,"serviceScript":"return new foam.dao.EasyDAO.Builder(x).setOf(foam.nanos.auth.GroupPermissionJunction.getOwnClassInfo()).setAuthenticate(false).setAuthorize(true).setJournalType(foam.dao.JournalType.SINGLE_JOURNAL).setJournalName(\"groupPermissionJunctions\").setPm(true).build()"})
+p({"class":"foam.nanos.boot.NSpec","name":"localGroupPermissionJunctionDAO","description":"The junction DAO for the many-to-many relationship between groups and permissions","serve":false,"serviceScript":"return new foam.dao.EasyDAO.Builder(x).setAuthorize(true).setOf(foam.nanos.auth.GroupPermissionJunction.getOwnClassInfo()).setJournalType(foam.dao.JournalType.SINGLE_JOURNAL).setJournalName(\"groupPermissionJunctions\").setPm(true).build()"})
```
smirnovaka: isn't `authorize` true by default?

src/services
```diff
@@ -177,6 +175,6 @@ p({"class":"foam.nanos.boot.NSpec", "name":"pmDailyReduceManager","lazy":true,"s
...
+p({"class":"foam.nanos.boot.NSpec","name":"userCapabilityJunctionDAO","lazy":false,"serve":true,"authenticate":false,"serviceScript":"dao = new foam.dao.EasyDAO.Builder(x).setAuthorize(false).setJournalType(foam.dao.JournalType.SINGLE_JOURNAL).setJournalName(\"userCapabilityJunctions\").setOf(foam.nanos.crunch.UserCapabilityJunction.getOwnClassInfo()).build(); userCap =  new foam.nanos.crunch.UserCapabilityJunctionDAO.Builder(x).setDelegate(dao).build(); ruler = new foam.nanos.ruler.RulerDAO(x,userCap, \"userCapabilityJunctionDAO\"); return ruler;","client":"{\"of\":\"foam.nanos.crunch.UserCapabilityJunction\", \"remoteListenerSupport\": false}"})
```
smirnovaka: can you please remove `authenticate:false`, almost none of the services should be `authenticate:false` except the ones that let you signup/login

smirnovaka: remove `lazy:false`

smirnovaka (changes requested):
can you please go over properties like `lazy`, `serve` `authenticate` ... and check if their default values are appropriate. If so, do not repeat them

TW80000 (commented):
I think it would still be valuable to have a test suite for the `StandardAuthorizer`, much like the current `AuthenticatedDAOTest`.

PR #2444, opened 2019-07-25
Edit table column changes

mcarcaso:
A few updates:
- Show the table columns editor in their own page rather than from a popover because those don't work right.
- Allow all columns to be selected when configuring columns to show.
- Allow the user to choose if they want to always show, or always hide a particular column.


src/foam/u2/view/EditColumnsView.js
```diff
@@ -4,84 +4,118 @@
...
+          sectionName: '_defaultSection'
+        },
+        mode: 'RO'
+      },
+      factory: function() {
+        return this.allColumns
+          .map(c => {
+            var a = this.of.getAxiomByName(c);
+            var v = localStorage.getItem(this.of.id + '.' + a.name);
+            v = this.ColumnVisibility.VALUES.find(e => e.name == v);
```
mcarcaso: Shouldn't need the find here.

src/foam/u2/view/EditColumnsView.js
```diff
@@ -4,84 +4,118 @@
...
+              label: a.label || foam.String.labelize(a.name),
+              visibility: v || 'DEFAULT'
+            })
+          })
+      }
     },
     {
-      name: 'columns_'
+      class: 'StringArray',
+      name: 'allColumns',
```
mcarcaso, line 93: I do this in UnstyledTableView too. Put somewhere common?

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -188,13 +222,14 @@ foam.CLASS({
...
     function createColumnSelection() {
-      var editor = this.EditColumnsView.create({
-        columns: this.columns,
-        columns_$: this.columns_$,
-        table: this.of
+      this.__context__.stack.push({
+        class: 'foam.u2.DetailView',
+        data: this.EditColumnsView.create({
+          of: this.of,
+          displayColumns$: this.columns$,
```
mcarcaso: Don't think anything other than `of` is needed.

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -188,13 +222,14 @@ foam.CLASS({
     },
 
     function createColumnSelection() {
-      var editor = this.EditColumnsView.create({
-        columns: this.columns,
-        columns_$: this.columns_$,
-        table: this.of
+      this.__context__.stack.push({
```
mcarcaso: Can probably remove this method and just inline this in the on click.

mcarcaso: Import stack rather than grabbing it from context directly

src/foam/u2/view/EditColumnsView.js
```diff
@@ -4,84 +4,118 @@
...
-          // Force the view to update.
-          this.columns_ = this.columns_.slice();
-        } else {
-          // The string isn't important, all that matters is that there's some
-          // value for this key.
-          localStorage.setItem(localStorageKey, 'Y');
-          this.columns_ = this.columns_.filter((col) => col.name !== axiom.name);
-        }
-      };
+  actions: [
```
mcarcaso, line 101: An action to set all to default might be useful. Also a cancel button (which shouldn't be needed once we give the stackview a back button)

src/foam/u2/view/EditColumnsView.js
```diff
@@ -4,84 +4,118 @@
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
+foam.ENUM({
+  package: 'foam.u2.view',
+  name: 'ColumnVisibility',
+  values: [
+    {
+      name: 'DEFAULT',
```
mcarcaso, line 12: Give these labels.

src/foam/u2/view/EditColumnsView.js
mcarcaso: Take permissions into account?

mcarcaso: Not gonna do permissions for now. The properties shouldn't be filled in anyway

src/foam/u2/view/EditColumnsView.js
```diff
@@ -4,84 +4,124 @@
...
+      name: 'visibility',
+      label: '',
+      gridColumns: 6
+    }
+  ]
+});
 
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'EditColumnsView',
```
TW80000, line 55: This isn't a view anymore, right? Might be better to rename it.

TW80000 (commented):
LGTM

src/foam/u2/view/EditColumnsView.js
mcarcaso: Good point. As of https://github.com/foam-framework/foam2/pull/2455/files it is toE-able which makes it more like a view so I think it's ok to keep calling it that.

PR #2449, opened 2019-07-26
FloatView + CurrencyView formatting

kennyqiyenkan:
Changes FloatView element type to be text/number depending on DisplayMode.

CurrencyView will display value according to Locale.

### DIFF
--- a/src/foam/u2/CurrencyView.js
+++ b/src/foam/u2/CurrencyView.js
@@ -19,7 +19,10 @@ foam.CLASS({
 
   methods: [
     function dataToText(val) {
-      return this.SUPER(val / 100);
+      val = Math.floor(val / 100);
+      return this.mode === foam.u2.DisplayMode.RO ?
+        val.toLocaleString(undefined, {minimumFractionDigits: 2}) :
+        this.SUPER(val);
     },
 
     function textToData(text) {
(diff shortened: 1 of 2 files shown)


src/foam/u2/FloatView.js
```diff
@@ -30,7 +30,12 @@ foam.CLASS({
   `,
 
   properties: [
-    ['type', 'number'],
+    {
+      name: 'type',
+      expression: function(mode) {
+        return mode === foam.u2.DisplayMode.RO ? 'text' : 'number';
```
mcarcaso, line 36: Probably good to give a little explanation of why we're doing this.

src/foam/u2/CurrencyView.js
```diff
@@ -19,7 +19,10 @@ foam.CLASS({
 
   methods: [
     function dataToText(val) {
-      return this.SUPER(val / 100);
+      val = Math.floor(val / 100);
```
mcarcaso: FloatView has a formatNumber method on it that is used by its dataToText method. You might want to override that here and leave this method as is.

src/foam/u2/FloatView.js
```diff
@@ -30,7 +30,12 @@ foam.CLASS({
   `,
 
   properties: [
-    ['type', 'number'],
+    {
```
mcarcaso, line 33: Consider updating the formatNumber method in here to use the toLocalString.

mcarcaso (approved):

mcarcaso (approved):

PR #2453, opened 2019-07-29
Document and comment on the AuthService interface

TW80000:
* I saw that `checkUser` was being called by some recently checked in code and thought that was strange. I wanted to find out what the difference between `checkUser`, `checkUserPermission`, and `check` was, so I read the implementations and filled in the documentation for those methods. I documented the other methods as well while I was at it.
* While I was documenting the methods, I also included comments from points brought up during meetings about methods we don't like and might want to refactor or deprecate. This way they're down in writing in a place that's hard to miss.


kgrgreer:
Please resolve conflicts.

TW80000:
> Please resolve conflicts.

Resolved.

PR #2455, opened 2019-07-29
Make TableView column properties axiom arrays instead of string arrays

mcarcaso:
This should allow you to pass in your own axioms as columns.


TW80000 (approved):
LGTM

PR #2456, opened 2019-07-30
DNM - DIG Sink Parser Integration

mcarcaso:


src/foam/lib/AndPropertyPredicate.js
```diff
@@ -23,9 +23,20 @@ foam.CLASS({
     }
   }
 
+  return true;
+`
+    },
+    {
+      name: 'propertyPredicateCheckForDig',
+      javaCode: `
+  for ( foam.lib.PropertyPredicate p : getDelegates() ) {
```
mcarcaso, line 32: Don't add anything dig specific to these predicates. It's unfortunate that we even need these predicates in the first place. We should think about what it would take to make our PropertyInfos on java just like any other FObject so we can query them with the predicate library we already have in mlang.js

src/foam/lib/PropertyPredicate.js
```diff
@@ -20,7 +20,28 @@ foam.INTERFACE({
           javaType: 'foam.core.PropertyInfo'
         }
       ]
+    },
+    {
+      name: 'propertyPredicateCheckForDig',
```
mcarcaso, line 25: We shouldn't need this.

src/foam/lib/json/Outputter.java
```diff
@@ -45,6 +45,8 @@ protected SimpleDateFormat initialValue() {
   protected boolean       outputReadableDates_ = true;
   protected PropertyPredicate propertyPredicate_;
   protected Map<String, List<PropertyInfo>> propertyMap_ = new HashMap<>();
+  protected String[]      digColumns_ = null;
```
mcarcaso, line 48: JSON outputter should have no awareness of dig

src/foam/nanos/dig/DIG.js
```diff
@@ -53,6 +53,25 @@ foam.CLASS({
     },
     'cmd',
     'format',
+{
+      name: 'digColumns',
```
mcarcaso, line 57: I would just call this `columns` or maybe `fields`
I'd also start with making it a StringArray. We might want to change that later to something else but I think StringArray is  a good place to start.

src/foam/nanos/dig/DIG.js
```diff
@@ -53,6 +53,25 @@ foam.CLASS({
...
+      expression: function(daoKey) {
```
mcarcaso, line 58: FYI if you make it a StringArray, you won't be able to use `expression` unless you null out the factory on it. i.e. add `factory: null` if you want the `expression` to work.

src/foam/nanos/dig/DIG.js
```diff
@@ -53,6 +53,25 @@ foam.CLASS({
...
+        //if ( ! this.of ) return [];
+
+        var dao = this.__context__[daoKey];
+
+        if (dao.of.getAxiomByName('digColumns')) {
+        //alert("digColumns : " + dao.of.getAxiomByName('digColumns').columns);
+
+
+//        return dao.of.getAxiomsByClass(foam.core.Property)
```
mcarcaso, line 67: Use this commented out piece of code instead of `digColumns`.

src/foam/nanos/dig/DigSelectPropertyView.js
```diff
@@ -0,0 +1,123 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.dig',
+  name: 'DigSelectPropertyView',
```
mcarcaso, line 9: We'll make a widget for this later. For this PR, we'll just use the default view that StringArray provides.

src/foam/u2/Element.js
```diff
@@ -2648,6 +2648,17 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
+  package: 'foam.u2',
+  name: 'digColumns',
```
mcarcaso, line 2653: Shouldn't need this yet and it shouldn't be part of u2.

PR #2457, opened 2019-07-30
Add loading indicator to ReciprocalSearch

TW80000:
## Changes

* Show "Loading..." instead of "0 of 0 selected" while we're waiting for the queries to finish


src/foam/u2/view/ReciprocalSearch.js
```diff
@@ -16,6 +16,18 @@ foam.CLASS({
     'foam.u2.view.SearchViewWrapper'
   ],
 
+  constants: [
+    {
+      name: 'FORMAT_REGEX',
```
mcarcaso: I don't think you need this. Numbers have a toLocalString method that does this for you.

src/foam/u2/view/ReciprocalSearch.js
```diff
@@ -99,11 +111,17 @@ foam.CLASS({
     },
     {
       class: 'Int',
-      name: 'selectedCount'
+      name: 'selectedCount',
+      postSet: function() {
```
mcarcaso: Can we get rid of these postSets and change `isLoading` boolean to a `loadingRequests` int.
The in updateTotalCount and updateSelectedAccount, they do a loadingRequests++ before their select and a loadingRequests-- after, then you check if loadingRequests == 0 when deciding what to show?

src/foam/u2/view/ReciprocalSearch.js
kgrgreer: Number(12234234).toLocaleString();   -> "12,234,234"

src/foam/u2/view/ReciprocalSearch.js
```diff
@@ -113,6 +131,31 @@ foam.CLASS({
           predicate$: this.data$
         });
       }
+    },
+    {
+      class: 'Boolean',
+      name: 'isLoading',
```
kgrgreer: You could also just set the count to Number.NaN when you don't know and have the countText say Loading... when it's that value.

src/foam/u2/view/ReciprocalSearch.js
TW80000: Resolved by 26c02f92336f9547fa1a094d0f70000521674d0c.

src/foam/u2/view/ReciprocalSearch.js
TW80000: Resolved by 24246fe6f5c09eb6b84f47ee279e11d36cfc8205.

src/foam/u2/view/ReciprocalSearch.js
TW80000: Resolved by 24246fe6f5c09eb6b84f47ee279e11d36cfc8205. I took Mike's approach instead since it shows "Loading..." while we wait for the table to populate as well. The count usually finishes before the real data loads so it's a bit weird to see the count show up before the data loads. By waiting for all async requests to finish, "Loading..." is shown until everything is visible, which makes more sense from a UX perspective, in my opinion.

src/foam/u2/view/ReciprocalSearch.js
```diff
@@ -99,11 +99,17 @@ foam.CLASS({
     },
     {
       class: 'Int',
-      name: 'selectedCount'
+      name: 'selectedCount',
+      postSet: function() {
```
mcarcaso: I think you forgot to ditch these postSets

mcarcaso (approved):

src/foam/u2/view/ReciprocalSearch.js
TW80000: Thanks. Fixed by 93b99e5.

PR #2458, opened 2019-07-30
Export f object fixes

tharmaman:


src/foam/lib/csv/CSVOutputterImpl.js
```diff
@@ -29,9 +29,9 @@ foam.CLASS({
     {
       class: 'StringArray',
       name: 'props',
-      factory: function(of) {
```
mcarcaso, line 32: Instead of this change, can you make this an expression? You'll need to null out the factory if you do but it feels a little more robust.

src/foam/nanos/export/CSVDriver.js
```diff
@@ -22,14 +22,14 @@ foam.CLASS({
       of: 'foam.lib.csv.CSVOutputter',
       name: 'outputter',
```
mcarcaso: We should nuke this property and just create a new one every time you run the exportFObject method. The reason being, when you re-use the outputter, it'll be tied to a particular type of FObject and it will not re-output the columns.

src/foam/u2/ExportModal.js
```diff
@@ -111,7 +111,7 @@ foam.CLASS({
       if ( this.exportData ) {
         this.note = await exportDriver
           .exportDAO(this.__context__, this.exportData);
-      } else {
+      } else if ( this.exportObj ) {
```
mcarcaso: Maybe add an else that logs a warning or something?

src/foam/u2/ExportModal.js
```diff
@@ -121,15 +121,25 @@ foam.CLASS({
...
-      exportDriver.exportDAO(this.__context__, this.exportData)
-        .then(function(result) {
-          result = 'data:text/csv;charset=utf-8,' + result;
-          var link = document.createElement('a');
-          link.setAttribute('href', encodeURI(result));
-          link.setAttribute('download', 'data.csv');
-          document.body.appendChild(link);
-          link.click();
-        });
+      if ( this.exportData ) {
```
mcarcaso: Can you do something like this instead?
```
var p = this.exportData ?
    exportDriver.exportDAO(this.__context__, this.exportData) :
    Promise.resolve(exportDriver.exportFObject(this.__context__, this.exportObj));
p.then(result => {
            result = 'data:text/csv;charset=utf-8,' + result;
            var link = document.createElement('a');
            link.setAttribute('href', encodeURI(result));
            link.setAttribute('download', 'data.csv');
            document.body.appendChild(link);
            link.click();
})
```

tharmaman: Done. Did the same for the above to convert to a ternary.

src/foam/lib/csv/CSVOutputterImpl.js
tharmaman: Done.

src/foam/nanos/export/CSVDriver.js
tharmaman: Done.

mcarcaso (approved):

PR #2461, opened 2019-07-31
Make DateView re-set its data on blur

tharmaman:


src/foam/u2/DateView.js
```diff
@@ -34,23 +34,39 @@ foam.CLASS({
     [ 'placeholder', 'yyyy-mm-dd' ]
   ],
 
+  listeners: [
+    {
+      name: 'onBlur',
```
mcarcaso, line 39: Let's try to explain why we need to do this. Here's a shot at it:

A slot's relateTo prevents feedback loops between the two slots that are related to eachother. This feedback prevention means that adapts and preSets on the Date properties won't be reflected in the UI. Consider the following example:
0. data has a preSet to ensure the date must be before today.
1. user selects a date in the future from DateView
2. relateTo's `inputToData` triggers
3. data's preSet changes the date from tomorrow to today
4. relateTo's `dataToInput` does NOT fire because `inputToData` is still in progress.

Triggering dataToInput onBlur ensures that the value being presented in the UI is the most up to date value because it avoids being boxed out by the feedback loop prevention.

mcarcaso (approved):

src/foam/u2/DateView.js
```diff
@@ -34,23 +34,44 @@ foam.CLASS({
...
+    },
+
+    function inputToData(input) {
+      return new Date(input);
+    },
+
+    function dataToInput(data) {
+      // Using our own formatter to keep the date in the format (yyyy-mm-dd) while maintaining the locale date
+      const year = data.getFullYear();
+      const month = data.getMonth() + 1 > 9 ? data.getMonth() + 1 : `0${data.getMonth() + 1}`;
```
mcarcaso, line 71: I just discovered js's padStart that'll make life a little easier here:
```
const month = (data.getMonth() + 1).toString().padStart(2, '0');
const day = data.getDate().toString().padStart(2, '0');
```

PR #2463, opened 2019-07-31
CPF-278 Creating Java classes for MethodInfo to be able to get a method from an object

skanderturki:
Note: This PR contains code that was not merged and is being reviewed in previous PR.

Modified the class hierarchy of Axioms to include a MethodInfo class


mcarcaso:
The MethodInfo stuff should be independent of the JDBC stuff. Please do that in its own PR.

skanderturki:
> The MethodInfo stuff should be independent of the JDBC stuff. Please do that in its own PR.

Yes they are in two different PRs but this PR contains code from previous PR also.

PR #2478, opened 2019-08-02
[Temp PR] SinkParser 

minsunny:
https://nanopay.atlassian.net/browse/CPF-2244

will be gradually extended one by one here

### DIFF
--- a/src/foam/nanos/dig/DIG.js
+++ b/src/foam/nanos/dig/DIG.js
@@ -153,9 +153,8 @@ foam.CLASS({
           query = true;
           url += "fields=";
 
-          for ( var i = 0; i < fields.length; i++ ) {
+          for ( var i = 0; i < fields.length; i++ )
             url+= encodeURIComponent(fields[i]) + encodeURIComponent(",");
-          }
         }
         this.postURL = url;
 
--- a/src/foam/nanos/dig/DIG.js
+++ b/src/foam/nanos/dig/DIG.js
@@ -56,11 +56,11 @@ foam.CLASS({
     {
       class: 'StringArray',
       name: 'fields',
+      factory: null,
       expression: function(daoKey) {
          var dao = this.__context__[daoKey];
 
-         if ( dao != null )
-          return dao ? dao.of.getAxiomsByClass(foam.core.Property).map(p => p.name) : [];
+         return dao ? dao.of.getAxiomsByClass(foam.core.Property).map(p => p.name) : [];
       }
     },
     {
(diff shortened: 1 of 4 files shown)


src/foam/u2/Element.js
```diff
@@ -2648,6 +2648,17 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
+  package: 'foam.u2',
+  name: 'digColumns',
```
mcarcaso: digColumns has nothing to do with u2. Undo these changes. We'll introduce digColumns to models when we have everything else working.

src/foam/nanos/dig/DIG.js
```diff
@@ -53,6 +53,18 @@ foam.CLASS({
     },
     'cmd',
     'format',
+    {
+      name: 'digColumns',
```
mcarcaso: Make this a `class: 'StringArray'`. Doing so means you won't be able to use an expression unless you do a
`factory: null`

src/foam/nanos/dig/DIG.js
```diff
@@ -53,6 +53,18 @@ foam.CLASS({
...
+      expression: function(daoKey) {
+         var dao = this.__context__[daoKey];
```
mcarcaso, line 61: You might want to ensure that `dao` isn't null before attempting to use it.

src/foam/nanos/dig/DIG.js
```diff
@@ -53,6 +53,18 @@ foam.CLASS({
...
+
+         if ( dao.of.getAxiomByName('digColumns') ) {
+          return dao.of.getAxiomByName('digColumns').columns.toString();
```
mcarcaso: Remove this for now. We'll introduce the `digColumns` axiom when we're more ready later.

src/foam/nanos/dig/DIG.js
```diff
@@ -53,6 +53,18 @@ foam.CLASS({
...
+        } else {
+          return dao.of.getAxiomsByClass(foam.core.Property).toString();
```
mcarcaso: this isn't right. You want an array of strings, not the string of an array.
`return dao.of.getAxiomsByClass(foam.core.Property).map(p => p.name);`

src/foam/nanos/dig/DIG.js
mcarcaso: I'd rename this to either `columns` or `fields`
I like `fields` better because they're not always rendered as columns, as is the case for a JSON.

src/foam/nanos/dig/DIG.js
```diff
@@ -53,6 +53,16 @@ foam.CLASS({
     },
     'cmd',
     'format',
+    {
+      class: 'StringArray',
+      name: 'fields',
+      expression: function(daoKey) {
+         var dao = this.__context__[daoKey];
+
+         if ( dao != null )
```
mcarcaso: You do not need this if statement now that you check if dao is null in the return statement.

src/foam/nanos/dig/DIG.js
```diff
@@ -138,6 +148,11 @@ foam.CLASS({
           query = true;
           url += "q=" + encodeURIComponent(q);
         }
+        if ( fields ) {
```
mcarcaso: fields will always be an array so this check will never be false. Maybe you want to check if fields.length > 0

minsunny:
close this and moved here https://github.com/foam-framework/foam2/pull/2480

PR #2479, opened 2019-08-02
Speed up authorization checks for superadmins

TW80000:
Results in an 88.24% shorter mean duration and 93.75% shorter median
duration of calls to AuthService.check for users that meet the
conditions.

This AuthService decorator was working already, but there are other
users that could also be benefiting from the speed increase.


jlhughes (approved):
This will work for now. In the future we see an admin viewing the system as a particular user via adding and removing privileges from self.

PR #2480, opened 2019-08-02
SinkParser on DIG

minsunny:
https://nanopay.atlassian.net/browse/CPF-2244

i.e : 
request with '*' : returning all the fields 
request with min/max/sum/avg(propertyName) : returning proper value
request with the certain property list : returning filtered fields

to filter by certain fields 
-  service/dig?dao=userDAO&cmd=select&format=json&**fields=firstName,lastName,id**

to get sum, max, min, avg values
-  service/dig?dao=transactionLimitDAO&cmd=select&format=xml&**fields=max(amount)**

Screen Shot 2019-10-29 at 2 14 48 PM

Screen Shot 2019-10-29 at 2 15 06 PM


src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -64,6 +65,8 @@ public void execute(X x) {
     boolean             emailSet = email != null && email.length > 0 && ! SafetyUtil.isEmpty(email[0]);
     String              subject  = p.getParameter("subject");
     String              fileAddress = p.getParameter("fileaddress");
+    String              fields   = p.getParameter("fields");
+    String[]            fieldList = fields.split(",");
```
mcarcaso: You don't need this line.

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -115,6 +118,16 @@ public void execute(X x) {
       logger.debug("predicate", pred.getClass(), pred.toString());
       dao = dao.where(pred);
 
+      SinkParser sinkParser = new SinkParser(cInfo);
+
+      for ( int i = 0; i < fieldList.length; i++ ) {
```
mcarcaso: My understanding is that the parser can handle the entire comma separated string so you don't need to loop over it here. I.e:
```
SinkParser sinkParser = new SinkParser(cInfo);
StringPStream ps = new StringPStream(fields);
ParserContextImpl x_ = new ParserContextImpl();
ps = (StringPStream) sinkParser.parse(ps, x_);
```

src/foam/parse/SinkParser.java
```diff
@@ -26,12 +26,12 @@
   protected ClassInfo info_;
 
   public SinkParser(ClassInfo info) {
+    info_ = info;
```
adamvy: We should re-write this to not extend Alt but to rather just compose an Alt object.

```
public class SinkParser extends ProxyParser
{
  public SinkParser() {
    super(new Alt(starParser(), expressionListParser()));
  }
}
```

src/foam/parse/SinkParser.java
```diff
@@ -55,13 +49,11 @@ public Parser expressionParser() {
     return new Alt(parsers);
   }
 
-  /*
   @Override
   public PStream parse(PStream ps, ParserContext x) {
```
mcarcaso: Remove this method completely.

src/foam/parse/SinkParser.java
```diff
@@ -21,28 +21,22 @@
  *   ???: GROUP BY
  **/
 public class SinkParser
-  extends Alt
+  extends ProxyParser
 {
-  protected ClassInfo info_;
+  static ClassInfo info_;
```
mcarcaso: Remove this

src/foam/parse/SinkParser.java
```diff
@@ -21,28 +21,22 @@
...
-      starParser(),
-      expressionListParser()
-    };
-
-    info_ = info;
-  }
-
-  public Parser starParser() {
-    return new Literal("*");
+  public SinkParser() {
```
mcarcaso: Remove this

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -115,6 +117,12 @@ public void execute(X x) {
       logger.debug("predicate", pred.getClass(), pred.toString());
       dao = dao.where(pred);
 
+      SinkParser sinkParser = new SinkParser(cInfo);
+
+      StringPStream ps = new StringPStream(fields);
+      ParserContextImpl x_ = new ParserContextImpl();
+      ps = (StringPStream) sinkParser.parse(ps, x_);
```
mcarcaso: Below this line, print out ps.value to see if the SinkParser parsed it as you expected.

src/foam/parse/SinkParser.java
```diff
@@ -21,47 +21,25 @@
...
-  }
-
-  public Parser expressionListParser() {
-    return new Repeat(expressionParser(), ",", 1);
-  }
-
-  public Parser expressionParser() {
-    List         properties = info_.getAxiomsByClass(PropertyInfo.class);
+  public static Parser expressionParser(ClassInfo info) {
+    ClassInfo    cInfo = info;
```
mcarcaso: Why redeclare cInfo? Why not just call the argument cInfo in the args?

minsunny:
@nanoscott @kgrgreer  just made this up-to-date, please review again when you have a time

minsunny:
but seems it working weirdly on some classes using 'extend' such as user, account

Screen Shot 2019-10-29 at 2 51 03 PM

hchoura:
> but seems it working weirdly on some classes using 'extend' such as user, account
> 
> Screen Shot 2019-10-29 at 2 51 03 PM

This problem is with the html outputter. I check the CSV outputter ant it work fine.

hchoura (approved):

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -339,11 +342,64 @@ public void execute(X x) {
...
+        ParserContextImpl x_ = new ParserContextImpl();
+        ps = (StringPStream) sinkParser.parse(ps, x_);
+
+        Object[] psArray = (Object[]) ps.value();
+        logger.debug("ps.value.length: " + psArray.length);
+
+        for ( Object objPs : psArray ) {
+          logger.debug("ps.value() : " + objPs);
+
+          if ( objPs.toString().startsWith("groupBy") || objPs.toString().startsWith("min") || objPs.toString().startsWith("max") || objPs.toString().startsWith("avg") ) {
```
nanoscott, line 363: The list of support aggregate operation should be stored in an array and you should match against the whole array. Matching should be done with a case insensitive startsWith instead of just startsWith.

nanoscott, line 363: Can you add Count() to the list of aggregate operations that is supported?

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -339,11 +342,64 @@ public void execute(X x) {
...
+
+            if ( pInfo != null) {
+              if ( objPs.toString().startsWith("groupBy") ) {
+                foam.mlang.sink.GroupBy groupBy = (foam.mlang.sink.GroupBy) dao.select(MLang.GROUP_BY(pInfo, new foam.mlang.sink.Count()));
+
+                for ( Object key : groupBy.getGroups().keySet() ) {
+                  foam.mlang.sink.Count count = ((foam.mlang.sink.Count) groupBy.getGroups().get(key));
+                  out.println(key + " : " + count.getValue());
+                }
+              } else if ( objPs.toString().startsWith("min") || objPs.toString().startsWith("max") || objPs.toString().startsWith("avg") ) {
```
nanoscott, line 375: why are you doing this check a second time? it was already done above and can be skipped here by just using an 'else' statement.

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -339,11 +342,64 @@ public void execute(X x) {
...
+                foam.dao.Sink[] seqSinkArray = {MLang.MIN(pInfo), MLang.MAX(pInfo), MLang.AVG(pInfo), MLang.SUM(pInfo)};
```
nanoscott, line 376: Any reason you are returning all operations instead of just the one requested?

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -355,10 +411,14 @@ public void execute(X x) {
           }
           logger.debug(this.getClass().getSimpleName(), "objects selected: " + sink.getArray().size());
 
+          String[] fieldsArray = fields.split(",");
+          if ( fieldsArray[0].equals("*") ) fieldsArray = null;
```
nanoscott, line 415: Can this be the default?

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -597,4 +657,13 @@ protected void outputException(X x, HttpServletResponse resp, Format format, Pri
       // TODO
     }
   }
+
+  public String getPropertyName(String field) {
```
nanoscott, line 661: This won't work for count since it doesn't necessarily have a field in it .. not sure if we are expecting be COUNT() or COUNT(*)?

src/foam/parse/SinkParser.java
```diff
@@ -21,47 +22,46 @@
...
-      PropertyInfo info = (PropertyInfo) prop;
+      PropertyInfo pInfo = (PropertyInfo) prop;
 
-      parsers.add(new LiteralIC(info.getName()));
+      parsers.add(new LiteralIC(pInfo.getName()));
+      parsers.add(new LiteralIC("min(" + pInfo.getName() + ")"));
+      parsers.add(new LiteralIC("avg(" + pInfo.getName() + ")"));
+      parsers.add(new LiteralIC("sum(" + pInfo.getName() + ")"));
+      parsers.add(new LiteralIC("max(" + pInfo.getName() + ")"));
+      parsers.add(new LiteralIC("count(" + pInfo.getName() + ")"));
```
nanoscott, line 46: you don't support count right now.

src/foam/parse/SinkParser.java
```diff
@@ -21,47 +22,46 @@
...
-  /*
-  @Override
-  public PStream parse(PStream ps, ParserContext x) {
-    x = x.sub();
-    x.set("classInfo", info_);
-
-    return super.parse(ps, x);
-  }
-  */
+//  @Override
```
nanoscott, line 53: remove dead code.

nanoscott (changes requested):

PR #2486, opened 2019-08-07
this allows csvSink to spit out headers even when nothing to select

Nauna:
During a test I found that the outputter was not displaying anything, returning an empty string.

It turns out simply that during a select on an empty dao (ie where the predicate has filtered everything out), the csvSink.put was never called and thus no headers and no columns.

This PR allows makes the csvSink to output headers at least once.

### DIFF
--- a/src/foam/dao/CSVSink.js
+++ b/src/foam/dao/CSVSink.js
@@ -89,7 +89,6 @@ foam.CLASS({
         this.csv = this.outputter.toString();
       },
       javaCode: `
-        if ( getOutputter().getIsFirstRow() ) getOutputter().outputHeader(getX());
         setCsv(getOutputter().toString());
       `
     },
--- a/src/foam/dao/CSVSink.js
+++ b/src/foam/dao/CSVSink.js
@@ -8,7 +8,7 @@ foam.CLASS({
   package: 'foam.dao',
   name: 'CSVSink',
   extends: 'foam.dao.AbstractSink',
-  implements: [ 
+  implements: [
     'foam.core.Serializable'
   ],
   javaImports: [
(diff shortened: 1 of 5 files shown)


src/foam/lib/csv/CSVOutputter.js
```diff
@@ -8,13 +8,26 @@ foam.INTERFACE({
   package: 'foam.lib.csv',
   name: 'CSVOutputter',
 
+  properties: [
+    {
+      class: 'Boolean',
+      name: 'isFirstRow'
+    }
+  ],
+
```
Nauna: If I didn't move this out, then I would have had to cast the outputter as CSVOutputterImpl - in csvSink. I didn't think that was a good idea, in the future if we change things around we would have to find the cast and adjust.

src/foam/nanos/dig/DigWebAgent.java
```diff
@@ -7,7 +7,6 @@
 package foam.nanos.dig;
 
 import foam.core.*;
-import foam.dao.CSVSink;
 import foam.dao.ArraySink;
```
Nauna, line 10: just a clean up. This doesn't use the sink only the CSVOutputter

src/foam/dao/CSVSink.js
```diff
@@ -85,9 +85,11 @@ foam.CLASS({
     {
       name: 'eof',
       code: function() {
+        if ( this.outputter.isFirstRow ) this.outputter.outputHeader(this.__context__);
```
mcarcaso: Rather than doing it here, what if you did it in the toString() method of CSVOutputterImpl? Then the interface doesn't need to change and it can deal with the business logic of checking `isFirstRow`

Nauna: 🥇 🎉

mcarcaso (approved):

PR #2491, opened 2019-08-08
[CPF-1957] Add relationship check to table row count

KernelDeimos:
This commit fixes incorrect row counts on the table header of `foam.u2.DetailView` affecting properties that are many-to-many relationships.

### DIFF
--- a/src/foam/u2/DetailView.js
+++ b/src/foam/u2/DetailView.js
@@ -215,12 +215,26 @@ foam.CLASS({
                 hasTabs = true;
                 var label = p.label;
                 let tab = self.Tab.create({ label: label });
-                var dao = p.cls_ == foam.dao.ManyToManyRelationshipProperty
-                  ? p.get(self.data).getJunctionDAO()
-                  : p.get(self.data);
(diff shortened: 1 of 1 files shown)


KernelDeimos:
~~Should I make this a StringArray with an ArrayView instead? Already have the commit ready to push if that's better @Nauna~~ (commented this on the wrong PR)

TW80000 (commented):
I see now, nice find. I think a simpler fix would be to just change

```
  ? p.get(self.data).getJunctionDAO()
```

to

```
  ? p.get(self.data).getDAO()
```

Can you try that and see if that also works? Doing that should do what you're doing here but it happens under the hood.

KernelDeimos:
That worked! PR has been updated

TW80000 (commented):
Thanks for making the change. Can you please update it again so it only changes line 218? I think the ternary expression is more concise than the proposed code in this PR.

KernelDeimos:
Done

TW80000 (approved):

PR #2492, opened 2019-08-08
Random Sample PM

KateMengjunXia:
https://nanopay.atlassian.net/browse/CPF-2234
Run PM on random number hitting 0 (0-99), instead of on every operation.

### DIFF
--- a/src/foam/nanos/boot/Boot.java
+++ b/src/foam/nanos/boot/Boot.java
@@ -30,6 +30,7 @@ import static foam.mlang.MLang.EQ;
 public class Boot {
   // Context key used to store the top-level root context in the context.
   public final static String ROOT = "_ROOT_";
+  public static PM nullPM_ = new NullPM();
 
   protected DAO serviceDAO_;
   protected X   root_ = new ProxyX();
@@ -85,7 +86,7 @@ public class Boot {
         if ( rand == 0 ) {
           return new PM();
         } else {
-          return new NullPM();
+          return nullPM_;
         }
       }
     });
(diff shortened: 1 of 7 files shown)


src/foam/nanos/boot/Boot.java
```diff
@@ -75,6 +78,18 @@ public void put(Object obj, Detachable sub) {
...
     }, null);
 
+    // PM factory
+    root_ = root_.putFactory("PM", new XFactory() {
+      public Object create(X x) {
+        int rand = ThreadLocalRandom.current().nextInt(0, 100);
+        if ( rand == 0 ) {
+          return new PM();
+        } else {
+          return new NullPM();
```
adamvy: Don't allocate a NullPM instead use a static single instance all the time, that way we save on object allocations too.

KateMengjunXia: Changed to single instance.

src/foam/nanos/pm/NullPM.js
```diff
@@ -0,0 +1,35 @@
...
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.pm',
+  name: 'NullPM',
+  extends: 'foam.nanos.pm.PM',
+
```
jlhughes, line 11: @adamvy can we simply mark this with axiom singleton?

jlhughes (approved):

PR #2497, opened 2019-08-12
BlobPStream and other fixes

mcarcaso:


src/foam/lib/parse/BlobPStream.js
```diff
@@ -0,0 +1,89 @@
...
+    },
+    {
+      class: 'Int',
+      name: 'pos_'
+    },
+    {
+      class: 'String',
+      name: 'head_',
+      javaFactory: `
+        java.io.OutputStream os = new java.io.ByteArrayOutputStream(1);
```
adamvy, line 26: There must be a faster way to do this, but I can't figure out what it is pre-breakfast.  Will think about it.

mcarcaso: Haha yeah we'll figure something out. Opportunity for a performance tweak for sure.

PR #2498, opened 2019-08-12
Add new parsers to foam.parse

KernelDeimos:
This PR adds the following new functions to `foam.parse`:

### `implied(:any)`
Implied makes no changes to the PStream object other than setting the value to the one specified.

### ~~`chars1(:string)`~~
<removed; `chars` already functions this way>

### `untilEscaped(escape :ParserProperty, term :ParserProperty)`
UntilEscaped matches any character until the pattern `term`, but will also consume with the pattern `escape` with priority over `term`. This is useful for parsing quoted strings and handling specific escape sequences.

### DIFF
--- a/src/foam/parse/parse.js
+++ b/src/foam/parse/parse.js
@@ -420,6 +420,27 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
+  package: 'foam.parse',
+  name: 'Implied',
+
+  documentation: `Returns a value without affecting the
+    parser state.`,
+
+  properties: [
+    {
+      name: 'value',
+    }
+  ],
+
+  methods: [
+    function parse(ps) {
+      return ps.setValue(this.value);
+    },
+
+    function toString() { return 'implied(' + this.SUPER() + ')'; }
+  ]
+});
 
 foam.CLASS({
   package: 'foam.parse',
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Tagging @adamvy for review

src/foam/parse/parse.js
```diff
@@ -484,6 +505,38 @@ foam.CLASS({
...
+      name: 'string',
+      final: true
+    }
+  ],
+
+  methods: [
+    function parse(ps) {
+      if (!ps.valid) return undefined;
+      var ind = this.string.indexOf(ps.head);
+      return ind !== -1 ? ps.tail.setValue(this.string.charAt(ind)) : undefined;
```
adamvy: I think this will be functionally identical to the exisitng Chars.  The implied `value` of any pstream is the `ps.head` of the pstream that came before.

if ps.head === 'c' then ps.tail.value === 'c'.

src/foam/parse/parse.js
```diff
@@ -602,6 +655,59 @@ foam.CLASS({
...
+    }
+  ],
+
+  methods: [
+    function parse(ps, obj) {
+      var ret = [];
+      var esc = this.esc;
+      var term = this.term;
+
+      while ( ps.valid ) {
```
adamvy, line 653: I think you could compose this from existing primitives like so:

`repeat(not(<terminator>, alt(seq1(2, <escape>, <terminator>), anyChar))`

Although a shortcut for this could be useful, but instead of implementing parse() again you could just extend ProxyParser and put an expression/factory on delegate based upon the "terminator" and "escape" properties.

adamvy: Althought in Java we have foam.lib.json.StringParser which does a similar job and this has a written by hand parser, but that was somewhat done for performance reasons.

KernelDeimos: Oh awesome, I was hoping there was a way to do that with the existing parsers - thanks!

KernelDeimos: I actually need to capture my escape symbol as well (I have an action that returns the character escaped for valid JSON escapes), so I've tried

`repeat(not('"', alt(sym('json-escape'), anyChar())))`

but this is causing my parser to stop early. Any ideas?

src/foam/parse/parse.js
KernelDeimos: Thanks! Somehow I missed that

src/foam/parse/parse.js
KernelDeimos: I just realized I don't need to add a parser at all - if I can construct untilEscaped with primitives then I can just add a method to foam.parse.Parsers that uses the primitives the same way as in a grammar, right?

PR #2500, opened 2019-08-13
Rudimentary MLang predicate builder GUI

TW80000:
### Main changes

* Use the new strategizer service to populate the list of choices in `FObjectView`
* Use `FObjectView` of `Predicate` as the view for `PredicateProperty`-ies
* Added a `PropertyExpr` that implements `foam.mlang.Expr` and can be used as a strategy for that interface

The three changes above give us basic support for building MLang predicates in the GUI

### Supplementary changes

* Hid the Google DataStore Key property so it doesn't show up in the GUI when editing predicates
* Added a search field to the RichChoiceView for the `daoKey` property on the `Rule` model to make it easier to find the DAO you're looking for
* Added an assert statement to make sure `DAO.where` isn't called with a non-predicate property
* Fixed some slot memory leaks caused by links that were never being broken
* Updated `AttrSlot` to remove event listeners on unsubscribe
* Fixed a race condition in `SectionedDetailView` and its child views between slots of 'data' that was causing console errors
* Fix `SubSlot` so that it detaches when the parent class changes

## Future work

* Polish the UI, right now it's not particularly nice to work with


lchanmann:
I tested the PR against nanopay development branch:

1. when a rule doesn't have predicate set, editing would also not allow to edit its value
Screen Shot 2019-08-27 at 4 30 01 PM

2. should the UI for adding condition to predicate be a dropdown?
Screen Shot 2019-08-27 at 4 33 31 PM

3. Detail view doesn't show top-level predicate name
Screen Shot 2019-08-27 at 4 40 24 PM

kgrgreer:
What's the status of this?

TW80000:
> What's the status of this?

I started working again on it this week locally. Work is in progress.

TW80000:
Replaced by https://github.com/foam-framework/foam2/pull/2727

PR #2501, opened 2019-08-13
Remove nanopay from user and hashable test

michaelconnor468:


src/foam/parse/TestUser.java
```diff
@@ -32,7 +32,7 @@ public static void main(String[] args) {
     u1.setId(6);
     u1.setFirstName("Simon");
```
jlhughes: Please replace the names as well.   The account does not exist in production.

jlhughes (approved):

PR #2505, opened 2019-08-14
DotF predicate

lchanmann:
DotF predicate evaluates arg1 as a predicate with arg2 as its argument.

Suppose there is a rule with predicate: `EQ(User.ID, 1)`, the below query would evaluate the predicate against the given `user` object and check if the user id=1 then put the rule in the result.

    // the result should contain the rule
    ruleDAO.where(
      DOT_F(Rule.PREDICATE, user)
    )

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -3154,7 +3154,7 @@ foam.CLASS({
   implements: [ 'foam.core.Serializable' ],
 
   documentation: `A binary predicate that evaluates arg1 as a predicate with
-    arg2 as its argument.',
+    arg2 as its argument.`,
 
   properties: [
     {
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -3148,6 +3148,41 @@ foam.CLASS({
 
 
 foam.CLASS({
+  package: 'foam.mlang.predicate',
+  name: 'DotF',
+  extends: 'foam.mlang.predicate.AbstractPredicate',
+  implements: [ 'foam.core.Serializable' ],
+
+  documentation: `A binary predicate that evaluates arg1 as a predicate with
+    arg2 as its argument.',
+
+  properties: [
+    {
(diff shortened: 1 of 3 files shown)


src/foam/mlang/mlang.js
```diff
@@ -3147,6 +3147,41 @@ foam.CLASS({
 });
 
 
+foam.CLASS({
+  package: 'foam.mlang.predicate',
+  name: 'DotF',
+  extends: 'foam.mlang.predicate.AbstractPredicate',
```
TW80000: I think this can just extend `Binary` directly. Then you don't need to define the properties since they'll be inherited.

TW80000 (commented):
Just going to walk through it here so I understand. If you're doing a `.where` on `ruleDAO`, it's going to be applying the predicate passed as an argument to `.where` to each `Rule` in `ruleDAO`.

In your example, the first arg to `DOT_F` is `Rule.PREDICATE`, which is a `PropertyInfo`, which implements the `foam.mlang.Expr` interface. I'm assuming the second arg is an instance of the `User` class, which will get converted by `DOT_F` to a `foam.mlang.Constant`, which also implements `foam.mlang.Expr`.

Calling `.f` on each rule in `ruleDAO`, we'll have:

```Java
Object predicate = getArg1().f(obj);                  // predicate is EQ(User.ID, 1)
if ( predicate instanceof Predicate ) {               // true
  return ((Predicate) predicate).f(getArg2().f(obj)); // getArg2().f(rule) will be the user, so EQ(User.ID, 1).f(user) is true or false.
}
return false;
```

So it will work as expected.

src/foam/mlang/mlang.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2505/commits/ec964d1df48bd66249da7df007d903d7a36e0823.

TW80000 (approved):

lchanmann:
@kgrgreer is there something else needs to be addressed in the PR?

PR #2512, opened 2019-08-14
Cpf 2572/TreeView Updates

tharmaman:
Stdlib changes in regards to mike's PR here:
https://github.com/foam-framework/foam2/pull/2511

Added automatic canvas resizing, optimized speeding, added zoom in and zoom out functionality and also binded the toggling of expanding/collapsing child nodes with the scale factor.


src/foam/graphics/TreeGraph.js
```diff
@@ -38,7 +38,7 @@
 
    methods: [
      function initCView() {
-       this.SUPER();
+      this.SUPER();
```
mcarcaso: Looks like indentation got thrown off here.

src/foam/graphics/TreeGraph.js
```diff
@@ -263,33 +270,70 @@
...
+        {
+          name: 'doLayout',
+          isFramed: true,
+          documentation: 'Animate layout until positions stabilize',
+          code: function() {
+            var needsLayout = false;
+            // Scale and translate the view to fit in the available window
+            var gw = this.graph.width-110;
+            var w  = this.maxRight - this.maxLeft + 55;
+            // if ( w > gw ) {
```
mcarcaso: Either explain why this commented out or remove it completely.

src/foam/graphics/TreeGraph.js
```diff
@@ -263,33 +270,70 @@
...
+
+          for ( var i = 0; i < root.children.length; i++ ){
+            traverseAndCompare(root.children[i], root.x);
+          }
+          return;
+        }
+
+        traverseAndCompare(this.root);
+
+        // 300 gives a clear view of the spacing
```
mcarcaso: Do you mean 400 in this comment? Best not to mention the magic number's value in the comment so they don't both have to stay updated. Just explain why we add an arbitrary amount to the width.

src/foam/core/stdlib.js
```diff
@@ -433,7 +433,8 @@ foam.LIB({
       var argNames = foam.Function.argNames(fn);
       var args = [];
       for ( var i = 0 ; i < argNames.length ; i++ ) {
-        var a = source[argNames[i]];
+        var a = argNames[i].indexOf('$') == -1 ? source[argNames[i]] :
```
mcarcaso: From: https://github.com/foam-framework/foam2/pull/2511
Undo these changes from your PR since they're kinda unrelated to what your PR is doing. I'll handle getting that PR in independently of this PR.

mcarcaso (approved):

src/foam/graphics/TreeGraph.js
```diff
@@ -48,9 +48,8 @@
 
        // List for 'click' events to expand/collapse Nodes.
        this.canvas.on('click', function(e) {
-
          var x = e.layerX+this.nodeWidth/2, y = e.layerY;
-         var c = this.root.findFirstChildAt(x, y);
+         var c = this.root.findFirstChildAt(x/this.canvas.cview.scaleX, y/this.canvas.cview.scaleY);
```
adamvy: Hmm, I'm suprised that findFirstChildAt doesn't account for the scale, it's supposed to.

src/foam/graphics/TreeGraph.js
```diff
@@ -59,13 +58,21 @@
...
+
+           
          }
          this.doLayout();
        }.bind(this));
      },
 
      function doLayout() {
-       if ( this.root ) { this.root.layout(); this.root.doLayout(); }
+       if ( this.root ) {
```
adamvy, line 69: Unrolling to two lines isn't bad, but all these extra blank lines can be removed.

src/foam/graphics/TreeGraph.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
tharmaman: Yeah but this space know aligns it up with the rest of the code.

src/foam/graphics/TreeGraph.js
tharmaman: Yeah had to add it otherwise it wouldn't work when zoomed in or zoomed out.

src/foam/graphics/TreeGraph.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
tharmaman: Done.

src/foam/core/stdlib.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
```diff
@@ -263,33 +267,67 @@
...
+          documentation: 'Animate layout until positions stabilize',
+          code: function() {
+            var needsLayout = false;
+            // Scale and translate the view to fit in the available window
+            var gw = this.graph.width-110;
+            var w  = this.maxRight - this.maxLeft + 55;
+
+            var x = (-this.maxLeft+25)/w * gw + 55;
+            needsLayout = this.convergeTo(this.x$, x) || needsLayout;
+            if ( this.layout() || needsLayout ) this.doLayout();
```
kgrgreer: I'm not too keen on this if/else layout. Format the if part the same as the else part.

tharmaman: Done.

src/foam/graphics/TreeGraph.js
kgrgreer: Couldn't/shouldn't the findFirstChildAt() method handle the scaling?

src/foam/graphics/TreeGraph.js
kgrgreer: Could you use braces and indentation for both parts. Make the if part the same as the else part not make the else part the same as the if part ;-).

src/foam/graphics/TreeGraph.js
```diff
@@ -263,33 +267,66 @@
...
+        };
+          
+        const traverseAndCompare = root  => {
+          if ( root.maxLeft < maxes.maxLeft ) maxes.maxLeft = root.maxLeft;
+          if ( root.maxRight > maxes.maxRight ) maxes.maxRight = root.maxRight;
+
+          for ( var i = 0; i < root.children.length; i++ ){
+            traverseAndCompare(root.children[i], root.x);
+          }
+          return;
```
kgrgreer: Is this 'return' needed?

src/foam/graphics/TreeGraph.js
```diff
@@ -263,33 +267,66 @@
...
+     {
+       name: 'updateCWidth',
+       isFramed: true,
+       code: function() {
+        const maxes = {
+          maxLeft: 0,
+          maxRight: 0
+        };
+          
+        const traverseAndCompare = root  => {
```
kgrgreer: What's the advantage of this syntax over just: function traverseAncCompare(root) { ?

src/foam/graphics/TreeGraph.js
```diff
@@ -263,33 +267,66 @@
...
      }
+   ],
+
+   listeners: [
+     {
+       name: 'updateCWidth',
+       isFramed: true,
+       code: function() {
+        const maxes = {
+          maxLeft: 0,
```
kgrgreer, line 309: Wouldn't the maxLeft at least be the width of this box? Same for right? Would this fix the need for the hard-coded 400 elsewhere in the code?

src/foam/graphics/TreeGraph.js
tharmaman: Doesn't appear as the X and Y values get directly manipulated upon changing the scale of the canvas and so when changing the scale, the X and Y values will be whatever it is at scale = 1, X won't become X * scaleX. And after inspecting the findFirstchildAt function there is no adjustment made for the scale at all as it uses the hard X and Y values:
Screen Shot 2019-08-15 at 8 27 59 PM

src/foam/graphics/TreeGraph.js
tharmaman: Nope will remove.

src/foam/graphics/TreeGraph.js
tharmaman: Was just using ES6 syntax which turns it into an anonymous arrow function where the keyword 'this' doesn't bind to the function. But it isn't necessary in this case as we don't even use this. Will change it back to function traverseAndCompare(root) for consistency.

src/foam/graphics/TreeGraph.js
tharmaman: My bad couldn't think today, was nearly homeless in Boston for the weekend. Just fixed it.

src/foam/graphics/TreeGraph.js
tharmaman: By box are you referring to the:
**Canvas**
- can't set it up using `this.canvas.cview.width` or `this.width` because otherwise will go in an infinite loop
- hard coded the initial canvas size in of 2000 

**Canvas Container**
- I set it up as 1614 after removing the hard-coded 400 which is the width of the container and it still cuts off:
Screen Shot 2019-08-16 at 10 46 35 AM

**Root**
- I set it up as `this.root.width`and it still cuts off:
Screen Shot 2019-08-16 at 10 50 09 AM

tharmaman: Found a solution, I set width as:
`var width = Math.abs(maxes.maxLeft - maxes.maxRight) + this.nodeWidth + this.padding * 2;`

This is because they always get seem to cut off halfway. So `this.nodeWidth` accounts for the 2 (1/2)nodeWidths on both ends of the tree, and the padding * 2 adds the padding to both ends of the tree.

Managed to get it fit almost perfectly, however there's no padding at the left and all the way at the right the lines start to fade.

Screen Shot 2019-08-16 at 10 57 27 AM
Screen Shot 2019-08-16 at 10 57 20 AM

So I then **increased the padding factor by 2 to 4.** And now it seems to fit better on both ends:
Screen Shot 2019-08-16 at 10 58 11 AM
Screen Shot 2019-08-16 at 10 58 06 AM

Final width adjustment is:
`var width = Math.abs(maxes.maxLeft - maxes.maxRight) + this.nodeWidth + this.padding * 4;`

src/foam/graphics/TreeGraph.js
mcarcaso: You sure about that? This looks like it's one off from the line below it. That said, you're right in that this whole file seems to be off by one but don't address that in this PR because it'll make reviewing it tough

src/foam/graphics/TreeGraph.js
```diff
@@ -263,33 +267,72 @@
...
+          maxRight: 0
+        };
+          
+        function traverseAndCompare(root) {
+          if ( root.maxLeft < maxes.maxLeft ) maxes.maxLeft = root.maxLeft;
+          if ( root.maxRight > maxes.maxRight ) maxes.maxRight = root.maxRight;
+
+          for ( var i = 0; i < root.children.length; i++ ){
+            traverseAndCompare(root.children[i], root.x);
```
mcarcaso: Doesn't look like you're using the second arg here anymore.

src/foam/graphics/TreeGraph.js
```diff
@@ -263,33 +267,72 @@
...
+          }
+        }
+
+        traverseAndCompare(this.root);
+
+        // needed to use the adjustments to width in order to account for proper fitting on the screen
+        // since by default with out it, the leftmost node and right most node get cutoff by half
+        // padding adjustments are there to proper spacing and also for the edge connectors to 
+        // fully render
+        var width = Math.abs(maxes.maxLeft - maxes.maxRight) + this.nodeWidth + this.padding * 4;
```
mcarcaso, line 328: It feels like there's something deeper we don't really understand when it comes to the sizing here. Adding `nodeWidth` and `padding * 4` feels super random. `padding * 2` would make some sense because it's like you're adding left padding and right padding to the width. Why does adding a `nodeWidth` to the width of the canvas help?

Also, when we were working on the home button, wasn't the `x` value the center of the node rather than the far left side. Is this a detail possibly being overlooked? Or does maxLeft/maxRight already handle that?

src/foam/graphics/TreeGraph.js
kgrgreer: I didn't mean that findFirstChildAt() does take the scale into account. I meant that it should. It would make a better interface if it just handled it for you.

src/foam/graphics/TreeGraph.js
tharmaman: True. fixed it.

src/foam/graphics/TreeGraph.js
tharmaman: So without the adjustments, the canvas cuts off halfway on the leftmost node and the right most node. Thats where the intution of adding a nodeWidth helps to adjust for that part because its technically adding in two node halves in.

src/foam/graphics/TreeGraph.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
tharmaman: Done, you were right. there was a curly bracket that threw me off by aligning with the current indentation, so I just aligned those to the rest of the code.

PR #2520, opened 2019-08-19
Remove nanopay model browser

michaelconnor468:


jlhughes:
Does this css need to move nanopay side or we are not using this at the moment?

PR #2513, opened 2019-08-15
Polish ArrayView and friends

TW80000:
## Description

* Add some UI/UX polish to the `ArrayView`, `CollapsableDetailView`, and `DraftDetailView`

#### Two items in list, both collapsed
Screen Shot 2019-08-15 at 11 26 15 AM

#### Two items in list, one expanded
Screen Shot 2019-08-15 at 11 26 33 AM

#### Two items in list, both collapsed
Screen Shot 2019-08-15 at 11 30 37 AM

#### Two items in list, one expanded
Screen Shot 2019-08-15 at 11 30 48 AM

PR #2523, opened 2019-08-20
revert change

nanoNeel:
https://nanopay.atlassian.net/browse/CPF-2673


TW80000 (approved):
I made this change in https://github.com/foam-framework/foam2/pull/2513 but it was a mistake. I thought `section$title` was a Slot but it's a string, so my change doesn't make sense. I was trying to hide an `h2` in the DOM that was causing unnecessary whitespace when there was no text content in it, but I think I modified the wrong h2.

PR #2524, opened 2019-08-20
Add SimpleSearch to DAOBrowserViews

kennyqiyenkan:
As title says.


TW80000 (approved):
My only suggestion is to get rid of the left and right padding in the query bar. You can see in the screenshot that the edge of the search box doesn't line up with the edge of the table and likewise for the export button on the right.

Screen Shot 2019-08-21 at 10 03 14 AM

PR #2525, opened 2019-08-20
Fixes for three bugs that blocked merge ( AbstractJDBCDAO + Makefile + SQLException uncaught)

skanderturki:


jlhughes (approved):
builds clean on both foam and nanopay.

PR #2527, opened 2019-08-21
Border Sectioned detail view

zahnno:


src/foam/u2/detail/SectionedDetailView.js
```diff
@@ -30,6 +30,13 @@ foam.CLASS({
     }
   `,
 
+  properties: [
+    {
+      name: 'border',
```
TW80000, line 36:
```suggestion
      class: 'foam.u2.ViewSpec',
      name: 'border',
```

src/foam/u2/detail/SectionedDetailView.js
```diff
@@ -51,7 +58,7 @@ foam.CLASS({
                       .add(s.title$)
                       .show(s.title$)
                     .end()
-                    .start(self.CardBorder)
+                    .start({ class: self.border })
```
TW80000:
```suggestion
                    .start(self.border)
```

src/foam/u2/detail/SectionedDetailView.js
```diff
@@ -30,6 +30,13 @@ foam.CLASS({
     }
   `,
 
+  properties: [
+    {
+      name: 'border',
+      value: 'foam.u2.borders.CardBorder'
```
TW80000:
```suggestion
      factory: function() {
        return this.CardBorder;
      }
```

TW80000 (commented):
This works, although it would be more flexible to let `border` be any ViewSpec instead of requiring that it be a class id. That way any ViewSpec can be used, which opens up the ability to set default values for properties on your borders, using functions as ViewSpecs so code can be executed before returning an object, etc.

TW80000 (approved):

PR #2529, opened 2019-08-21
dateview bug

noodlemoodle:
remove custom tableCellFormatters that were taking the first 10 chars of isoString and use the generic tableCellFormatter for Date and DateTime that supports up to 6-digit years 
in this pr
https://github.com/nanoPayinc/NANOPAY/pull/7767

### DIFF
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -98,6 +98,14 @@ foam.CLASS({
       value: function (_, d) {
         if ( typeof d === 'number' ) return new Date(d);
         if ( typeof d === 'string' ) {
+          
+          // TODO: Currently some DateViews are allowing users to enter 6-digit years 
+          // which needs to be fixed via setAttribute('max', '9999-12-31') on DateViews
+          // However, there are already existing journal entries that contain Dates
+          // with 6-digit years and they are not showing up in tables because the invalid
+          // date error below is being thrown. 
+          // This is a temporary fix to migrate journal entries with bad dates to use proper date format
+          if ( d.indexOf('-') > 4 ) d = '9999' + d.substring(d.indexOf('-'));
           var ret = new Date(d);
 
           if ( isNaN(ret.getTime()) ) throw 'Invalid Date: ' + d;
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -98,14 +98,20 @@ foam.CLASS({
       value: function (_, d) {
         if ( typeof d === 'number' ) return new Date(d);
         if ( typeof d === 'string' ) {
-          
-          // TODO: Currently some DateViews are allowing users to enter 6-digit years 
-          // which needs to be fixed via setAttribute('max', '9999-12-31') on DateViews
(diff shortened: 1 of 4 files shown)


noodlemoodle:
please review @Nauna @mcarcaso https://nanopay.atlassian.net/browse/CPF-2687

Nauna:
@noodlemoodle after talking with Joel, considering we aren't validating dates specifically the year "10000" is a valid date. And the solution to CPF-2687 would be better suited to just remove the max attribute. We can enforce standards on specific fields.

src/foam/core/types.js
```diff
@@ -98,6 +98,20 @@ foam.CLASS({
...
         if ( typeof d === 'number' ) return new Date(d);
         if ( typeof d === 'string' ) {
+
+          // Currently we are allowing users to input >4-digit years 
+          // in our DateViews, but the ISO string is not formatted correctly
+          // A valid ISO string for >4-digit year must be padded to 6 digits
+          // in the year field and prepended by '+'
+          // this fixes the problem of the invalid string causing error to be
+          // thrown below.
+          year = d.substring(0, d.indexOf('-'));
```
mcarcaso, line 108: We shouldn't need to do any of this. I was looking at this with Anna earlier and the server is sending us an invalid date so the fix should be there and not here. `src/foam/lib/json/Outputter.java` uses a hand crafted date formatter that attempts to do ISO format but it's incorrect. It should use a canned formatter that java already provides. One of these: https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -212,8 +215,11 @@ foam.CLASS({
       class: 'foam.u2.view.TableCellFormatter',
       name: 'tableCellFormatter',
       value: function(date) {
-        // allow the browser to deal with this since we are technically using the user's preference
-        if ( date ) this.add(date.toLocaleString());
```
mcarcaso, line 216: I don't think we want to change this. Let the user's browser do this work for us.

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -194,8 +194,11 @@ foam.CLASS({
       class: 'foam.u2.view.TableCellFormatter',
       name: 'tableCellFormatter',
       value: function(date) {
-        // allow the browser to deal with this since we are technically using the user's preference
-        if ( date ) this.add(date.toLocaleDateString());
```
mcarcaso, line 198: When using foam.core.Date, we should always be displaying the UTC value. I'd suggest making this method do something like:
```
var d = new Date();
d.setFullYear(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate());
this.add(d.toLocaleDateString());
```
This constructs a date that isn't necessarily equal to the original date given but it's strictly for display purposes which is fine. It allows us to render the date in the user's preferred locale format.

noodlemoodle:
Closing because I think this has already been fixed elsewhere

PR #2530, opened 2019-08-23
fix server/client timeZone differentiation

Nauna:
Yikes after a lot of investigating and hating of js dates, it turns out this is all that was needed for correct saving of date with client/server timezone differences.

The issue was the string input was the correct date but the 'new Date(input)' was being saved in the local timeZone - causing the date to be saved incorrectly.

specifying UTC on saving the new Date object - then stores the correct date.

FYI: To illustrate my frustration, the image below displays console cmds I inputted within seconds, while on a breakpoint.

image

why are the first 2 dates different by hours??

### DIFF
--- a/src/foam/u2/DateView.js
+++ b/src/foam/u2/DateView.js
@@ -63,12 +63,12 @@ foam.CLASS({
     },
 
     function inputToData(input) {
-      return new Date(input);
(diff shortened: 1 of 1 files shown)


Nauna:
doing in yet another way.

PR #2531, opened 2019-08-23
CPF-2636 add YoungerThan mlang to Date

Nauna:
- to filter anything over a certain date

needed for Expiry Date - validation = a date after today.


Nauna:
GOing with NEQ(OlderThan) instead of this logic. - NOTE: however with NEQ(OlderThan) an empty state is no-longer null - thus need to add a not null predicate to balance logic.

PR #2533, opened 2019-08-23
Fix date years over4 digits

Nauna:
Started with work that @noodlemoodle began. She did 90% of the work on this.
https://github.com/foam-framework/foam2/pull/2529/files

Then @mcarcaso made some suggestions for this fix. 

Im making the PR cause I tested.

### DIFF
--- a/src/foam/lib/json/Outputter.java
+++ b/src/foam/lib/json/Outputter.java
@@ -30,7 +30,7 @@ public class Outputter
   protected static ThreadLocal<SimpleDateFormat> sdf = new ThreadLocal<SimpleDateFormat>() {
     @Override
     protected SimpleDateFormat initialValue() {
-      SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+      SimpleDateFormat df = new SimpleDateFormat("+yyyyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
       df.setTimeZone(java.util.TimeZone.getTimeZone("UTC"));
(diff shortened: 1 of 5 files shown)


kgrgreer:
Has this been tested to be compatible with JS and Swift? Can we send JS->Java, Java->JS, Swift->Java and Java->Swift?

mcarcaso:
Not yet for swift and the hard coded “+” isn’t technically correct.

src/foam/lib/json/Outputter.java
```diff
@@ -30,7 +30,7 @@
   protected static ThreadLocal<SimpleDateFormat> sdf = new ThreadLocal<SimpleDateFormat>() {
     @Override
     protected SimpleDateFormat initialValue() {
-      SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+      SimpleDateFormat df = new SimpleDateFormat("+yyyyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
```
kgrgreer, line 33: I'm not too keen on this change because it outputs 2019 as 002019. Can't we make this output 4-digit dates as 4-digit dates only only use 6-digit dates for those beyond 9999, since below 999 9 is the common case?

mcarcaso: Yes. It will complicate the outputting logic a little but if you want to support dates further in the future or in the past, you'll need to have some conditions in the outputter. The simplest approach would be to ditch date formatting and just use epoch ms as the serialized format.

PR #2534, opened 2019-08-23
Deduplicate a potentially expensive expression & improve docs

TW80000:
The expression `( (TreeNode) state ).get(( (TreeNode) state ), key, prop_)` was duplicated in the code, so if the compiler doesn't perform common subexpression substitution, then that expression would be needlessly evaluated twice.

### DIFF
--- a/src/foam/dao/index/TreeLookupFindPlan.java
+++ b/src/foam/dao/index/TreeLookupFindPlan.java
@@ -20,16 +20,20 @@ public class TreeLookupFindPlan implements FindPlan {
   }
 
   /**
-   * The cost will claculate by the node size
+   * The cost is calculated based on the size of the tree.
    */
   public long cost() {
     return ((Double) Math.log(Long.valueOf(size_).doubleValue())).longValue();
   }
 
   public FObject find(Object state, Object key) {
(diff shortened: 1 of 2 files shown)


src/foam/dao/index/TreeLookupFindPlan.java
```diff
@@ -20,16 +20,20 @@ public TreeLookupFindPlan(PropertyInfo prop, long size) {
...
    */
   public long cost() {
     return ((Double) Math.log(Long.valueOf(size_).doubleValue())).longValue();
   }
 
   public FObject find(Object state, Object key) {
-    if ( state != null && state instanceof TreeNode ) {
-      // Sometimes the object is not exist, it will return a null value.
-      return ( (TreeNode) state ).get(( (TreeNode) state ), key, prop_) == null ? null : (FObject) ( (TreeNode) state ).get(( (TreeNode) state ), key, prop_).value;
+    if ( state instanceof TreeNode ) {
```
kgrgreer, line 30: Why remove the check for state == null? This will happen on an empty DAO.

TW80000:
It's unnecessary since the instanceof check will be false whenever it's
null anyway.

PR #2535, opened 2019-08-24
[CPF 2717] Live Script Bundler

KernelDeimos:
### What this PR Adds
- Watch source files for changes (using WatchService) and re-run the Javascript builder (`./tools/js_build/build.js`)
- Serve the file at `./tools/js_build/foam-bin.js` using a buffer so that the file can be rewritten without interrupting/pausing a request

### How to Use
To enable this on the frontend with the buffer, simply add `<script src="YOUR_HOST/service/liveScriptBundler"></script>`.

To enable this on the frontend without the buffer, set the url to `/tools/js_build/foam-bin.js` directly instead.

### Future Improvements
- Use `foam.lib.parser` to parse files.js properly
- Make LiveScriptBundler a model (although it will never be used on the client, this may still be useful for other reasons)


src/foam/nanos/http/LiveScriptBundler.java
```diff
@@ -0,0 +1,353 @@
+package foam.nanos.http;
```
kgrgreer, line 7: Add copyright header.

src/foam/nanos/http/LiveScriptBundler.java
```diff
@@ -0,0 +1,353 @@
...
+      line != null                    ;
+      line = reader.readLine()
+    ) {
+      line = line.trim();
+      if (line.length() < 1) continue;
+      if ( lineException ) {
+        line = "{" + line + "}";
+        lineException = false;
+      }
+      else if (line.charAt(0) != '{') continue;
```
kgrgreer: Please match the style of the 'if' and 'else if' parts, like this:

if ( lineException ) {
  line = "{" + line + "}";
  lineException = false;
} else if (line.charAt(0) != '{') {
  continue;
}

src/foam/nanos/http/LiveScriptBundler.java
```diff
@@ -0,0 +1,353 @@
...
+      }
+
+      name += ".js";
+      paths.add(name);
+    }
+
+    return paths;
+  }
+
+  private void log_(String evt, String msg) {
```
kgrgreer, line 360: Use the standard nanos logger.

src/foam/nanos/http/LiveScriptBundler.java
```diff
@@ -0,0 +1,353 @@
+package foam.nanos.http;
+
```
kgrgreer, line 8: Please follow the FOAM style guide:
https://github.com/foam-framework/foam/wiki/CodingStyle

src/foam/nanos/http/LiveScriptBundler.java
KernelDeimos: ~@kgrgreer I actually wanted to do that, but wasn't sure how to  call `x.get("logger")` outside of the evaluate method~ oh, that was rather simple

src/foam/nanos/http/LiveScriptBundler.java
KernelDeimos: Added

src/foam/nanos/http/LiveScriptBundler.java
KernelDeimos: Updated

src/foam/nanos/http/LiveScriptBundler.java
KernelDeimos: Updated

src/foam/nanos/http/LiveScriptBundler.java
KernelDeimos: Updated

PR #2538, opened 2019-08-26
Swift fix1

michaelconnor468:


src/foam/mlang/mlang.js
```diff
@@ -3371,6 +3371,7 @@ foam.CLASS({
     {
       type: 'Regex',
       javaInfoType: 'foam.core.AbstractObjectPropertyInfo',
+      swiftType: 'String',
```
mcarcaso: Don't fix this here. Instead fill it in here: https://github.com/foam-framework/foam2/blob/master/src/foam/core/Type.js#L358

src/foam/core/Type.js
```diff
@@ -344,7 +344,7 @@ foam.CLASS({
   axioms: [ { class: 'foam.pattern.Singleton' } ],
   properties: [
     ['java', 'double'],
-    ['swift', 'Double'],
+      ['swift', 'Double'],
```
mcarcaso: ??

mcarcaso (approved):

PR #2540, opened 2019-08-26
Fix ExprProperty adaptValue

lchanmann:
Allow adapting foam model class to expression property to support `EQ(Thing.OF, OtherThing)` where OtherThing is a model class.

Credit to: @TW80000.

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -205,7 +205,7 @@ foam.CLASS({
       if ( o.class && this.__context__.lookup(o.class, true) ) {
         return this.adaptValue(this.__context__.lookup(o.class).create(o, this));
       }
-      if ( typeof o.id === 'string' && foam.isRegistered(o.id) ) {
+      if ( foam.core.FObject.isSubClass(o) ) {
         return foam.mlang.Constant.create({ value: o });
       }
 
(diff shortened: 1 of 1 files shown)


TW80000 (approved):

src/foam/mlang/mlang.js
```diff
@@ -205,6 +205,9 @@ foam.CLASS({
       if ( o.class && this.__context__.lookup(o.class, true) ) {
         return this.adaptValue(this.__context__.lookup(o.class).create(o, this));
       }
+      if ( typeof o.id === 'string' && foam.isRegistered(o.id) ) {
```
mcarcaso: Why do you care if it's registered? Also, if the purpose of this PR is for handling foam.core.Models, shouldn't the check on line 198 be catching these?

lchanmann: Line 198 catches the instance of a model class but not the model class itself. We want to be able to do `EQ(Thing.OF, OtherThing)` and OtherThing is a class.

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2540/commits/b0f5d8866afd736cef2a9ad6edfd4210e1a6ba34.

PR #2542, opened 2019-08-26
journaled sessiondao

michaelconnor468:


TW80000 (approved):

jlhughes:
We can do it another PR, but we need to make a number of session properties storage transient so they don't cause updates such as lastModified or lastUsed, uses, ... 
We only want 'new' 'remove' to be journaled.

TW80000:
> We can do it another PR, but we need to make a number of session properties storage transient so they don't cause updates such as lastModified or lastUsed, uses, ...
> We only want 'new' 'remove' to be journaled.

If we make `lastUsed` transient then we might run into situations where a session's TTL is 8h, a session is created at 8:00am and used until 11:30am, server goes down at 11:45am and is restored at 11:50am. When it restores, it can't read `lastUsed` because it's transient and thus not journaled so `lastUsed` defaults to `null`. Then the cron job has to make a decision about whether to clear it or not. Since the TTL is 8h and it was really last used at 11:30, it shouldn't be cleared, but the server doesn't know that. And one can just as easily think of an example where the session _should_ be cleared but again, the cron doesn't have enough information to make that decision after a restart. We have no way of knowing if the session should be cleared or not if the TTL is finite unless we journal `lastUsed`, unless I'm missing something. I guess we could also journal `created` since that won't change anyway and use TTL + created to make an educated guess, but even then it's not completely correct without `lastUsed`.

jlhughes:
We simply cannot journal every session change. 
option: on restart/clearing - ignore when lastUsed is null.  On restart, we get the same behaviour as a new session.

jlhughes (approved):

PR #2547, opened 2019-08-27
Initial (2) addition of the ORBital DAO feature CPF-278

skanderturki:
Example of testing script:

orb = new foam.dao.ORBitalDAO();

acc1 = new net.nanopay.account.Account(x, 11, true, false, "MyAccount", "Personal account", true,
      true, "PersAcc", true, "Personal", 100, 150, new java.util.Date(2019, 12, 01), 1, 
      new java.util.Date(2019, 12, 01), 1, 0, 1);

print(acc1.getBalance());

Object[] args = new Object[1];
args[0] = x;

req = new foam.dao.ORBRequest(acc1, "findBalance", args);
print(orb.cmd_(x, req));


src/foam/dao/ORBitalDAO.java
```diff
@@ -0,0 +1,54 @@
...
+import foam.core.FObject;
+import foam.core.MethodInfo;
+import foam.core.X;
+import foam.mlang.order.Comparator;
+import foam.mlang.predicate.Predicate;
+
+/**
+ * DAO class used to delegate a method execution request to the specified object (specified by an ORBRequest object).
+ */
+public class ORBitalDAO extends foam.dao.AbstractDAO{
```
jlhughes: What is the expected ORBitalDAO setup? 
Example: I expect to create an ORBRequest on the client and call cmd(), then on the server the ORBitalDAO, somewhere in the DAO stack, tests the cmd object and if an ORBRequest, executes it. 
In this scenario the ORBitalDAO should extend ProxyDAO, and test if the cmd object is an ORBRequest.

skanderturki:
What is the addition of the ProxyDAO compared to the AbstractDAO? Currently
you can make something like:

orb = new foam.dao.ORBitalDAO();

acc1 = //some account object

Object[] args = { x };

req = new foam.dao.ORBRequest(acc1, "findBalance", args);
orb.cmd_(x, req);   // The OrbitalDAO converts the req to an ORBRequest
object

skanderturki:
Do you mean that the ORBitalDAO should be installed as a service?

skanderturki:
With the addition of an ORBitalDAO as a service :

p({"class":"foam.nanos.boot.NSpec", "name":"anORBitalDAO", "lazy":false, "serve":true, "serviceScript":"return new foam.dao.ORBitalDAO();",  .... })

-> we can do:

    orb = x.get("anORBitalDAO");

    acc1 = // some Account object

    Object[] args = {x};

    req = new foam.dao.ORBRequest(acc1, "findBalance", args);

    orb.cmd_(x, req)

jlhughes:
Thinking: 
on the javascript side: 
`req = await acc1.findBalance(x);`
calls something like: 
`accountDAO.cmd(x, ORBRequest.create({"receiverObjectId": this.id, "methodName", "findBalance", "args": args})`

on the server/java side:
An ORBitalDAO on the localAccountDAO delegate chain will intercept cmd(), find the object, call the method, return the result. 

@kgrgreer a model method marked as 'remote' where we would generate the client/server call - I can't think how we map   acc1.findBalance(x) to accountDAO.cmd(x, ORBRequest...  - the dao to call cmd on. 
EasyDAO can wrap client dao in contextualizing dao which adds a 'dao' key which returns the dao through which to send the cmd().

src/foam/core/AbstractFObject.js
```diff
@@ -279,6 +279,26 @@ foam.CLASS({
         return property != null && property.isDefaultValue(this);
       `
     },
+    {
+      name: 'getMethod',
```
mcarcaso: Do you really need this getter/setter here? The ClassInfo's getAxiomByClass should be all you need for getting a method and I don't know when anybody would ever want to set a method.

src/foam/core/ClassInfoImpl.java
```diff
@@ -72,6 +72,9 @@ public ClassInfo addAxiom(Axiom a) {
     if ( a instanceof PropertyInfo ) {
       ((PropertyInfo)a).setClassInfo(this);
     }
+    if ( a instanceof MethodInfo) {
```
mcarcaso, line 75: Rather than doing this, you might want to make an interface for something like ClassInfoAware and check if it's an instance of that. Looks like there's a TODO above saying exactly this. Otherwise, we're going to need to explicitly add each axiom here and that approach doesn't scale well.

src/foam/core/MethodInfo.java
```diff
@@ -0,0 +1,26 @@
...
+
+package foam.core;
+
+public interface MethodInfo extends Axiom {
+
+  public ClassInfo getClassInfo();
+  public MethodInfo setClassInfo(ClassInfo p);
+
+  public String getName();
+  public void setName(String name);
```
mcarcaso: When would you want to change the name of a method at runtime? I wouldn't put this on the Interface if you can help it.

src/foam/core/MethodInfo.java
```diff
@@ -0,0 +1,26 @@
...
+
+  public Object get(Object obj);
```
mcarcaso: get/set on a method doesn't really make sense. What would these methods do?

src/foam/core/MethodInfo.java
```diff
@@ -0,0 +1,26 @@
...
+  public void set(Object obj, Object value);
+
+  public boolean getPermissionRequired();
```
mcarcaso: This method seems like it should be part of some other interface that PropertyInfo also implements.

src/foam/core/MethodInfo.java
```diff
@@ -0,0 +1,26 @@
...
+  public void authorize(foam.core.X x);
```
mcarcaso: It's unclear what this method is expected to do. It doesn't return or throw anything?

src/foam/core/SimpleMethodInfo.java
```diff
@@ -0,0 +1,76 @@
...
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package foam.core;
+
+import foam.nanos.auth.AuthService;
+import foam.nanos.auth.AuthorizationException;
+
+public abstract class SimpleMethodInfo implements MethodInfo
```
mcarcaso: I'm not sure it's really worth having this class. You should be able to create these anonymously when generating the MethodInfos. You wouldn't need to even store a name as the getName() method could be hardcoded with the method's name.
If you do want to keep this class around, I'd still recommend removing the bulk of it. getClassInfo, setClassInfo,  authorize, and toString are the only methods I think you'll want to keep around. Also if you take this approach, make this class abstract and remove the constructor.

src/foam/dao/ORBRequest.java
```diff
@@ -0,0 +1,47 @@
+package foam.dao;
+
+import foam.core.FObject;
+
+/**
+ * ORBRequest objects are used to pass method call arguments to ORBitalDAOs request brokers.
+ */
+public class ORBRequest{
```
mcarcaso: Why not model this?

src/foam/dao/ORBitalDAO.java
mcarcaso: Model this and make it extend ProxyDAO. Then you'll just have to fill in the cmd_ method.

src/foam/dao/ORBitalDAO.java
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Method that delegates execution to method of object specified by {@param obj}
+   *
+   * @param obj argument must be an ORBRequest object
+   */
+  @Override
+  public Object cmd_(foam.core.X x, Object obj){
+    ORBRequest request = (ORBRequest) obj;
```
mcarcaso: Probably want to check if it's an instance of ORBRequest before blindly casting it.

src/foam/dao/ORBitalDAO.java
```diff
@@ -0,0 +1,54 @@
...
+    FObject receiverObj =  ( receiverID == null )? request.getReceiverObject() : (FObject) x.get(request.getReceiverObjectID());
+
+    ClassInfo classInfo = receiverObj.getClassInfo();
+    MethodInfo methodInfo = (MethodInfo) classInfo.getAxiomByName(request.getMethodName());
+
+    return methodInfo.call(x, receiverObj, request.getArgs());
+  }
+
+  @Override
+  public FObject put_(X x, FObject obj) {
```
mcarcaso: Remove all these other methods and let the ProxyDAO do the work.

src/foam/java/refinements.js
```diff
@@ -635,14 +635,49 @@ foam.CLASS({
...
-    {
-      class: 'Boolean',
-      name: 'remote'
     }
   ],
 
   methods: [
+    function buildMethodInfoInitializer(cls) {
+      // Add MethodInfo field for each method
+      initializerString = 'new foam.core.SimpleMethodInfo(){\n';
```
mcarcaso: Use template strings in js: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
It'll make this a lot easier to read.

src/foam/core/ClassInfoImpl.java
skanderturki: In my humble opinion, I am not sure that an interface is necessary! Apparently not all axioms are ClassInfoAware. So maybe it would be added complexity for not much if we added another level. To be discussed?!

src/foam/core/AbstractFObject.js
skanderturki: ok

src/foam/core/MethodInfo.java
skanderturki: ok

src/foam/core/MethodInfo.java
skanderturki: ok

src/foam/core/MethodInfo.java
skanderturki: I took these two methods from PropertyInfo as I thought that maybe we would like to constraint the call of the method remotely for authorized users. They are not used currently but I thought it would be useful to have them.

src/foam/dao/ORBRequest.java
skanderturki: ok

src/foam/dao/ORBitalDAO.java
skanderturki: ok

src/foam/dao/ORBitalDAO.java
skanderturki: yes

src/foam/dao/ORBitalDAO.java
skanderturki: ok

src/foam/java/refinements.js
skanderturki: done

src/foam/core/ClassInfoImpl.java
mcarcaso: >Apparently not all axioms are ClassInfoAware.

This is why we need the extra interface. Otherwise, we're going to have to add code here every time we create a new axiom type that needs the class info set. If all axioms were ClassInfoAware then we could blindly set the class info on all axioms that get added here but since that's not the case, we need some other way of identifying that an axiom can have the class info set on it.

src/foam/core/MethodInfo.java
mcarcaso: Yeah we might want them in the future but for now let's keep them out of this while you're still building this out. We need to be sure not to turn anything on until we've tightened up the security around but but for the time being, we'll focus on making the ORB work.

FYI the `authorize` method comes from the Expr interface which this Axiom probably won't implement.

src/foam/core/ClassInfoImpl.java
skanderturki: Ok done

src/foam/dao/ORBitalDAO.js
```diff
@@ -0,0 +1,61 @@
...
+    {
+      name: 'cmd_',
+      type: 'Object',
+
+      args: [
+        { name: 'x' , type: 'foam.core.X'}, { name: 'obj' , type: 'FObject' }
+      ],
+
+      javaCode: `
+        ORBRequest request = null;
```
jlhughes, line 36: Since the ORBitalDAO is a Proxy, rather than the try/catch, test if the obj is an instance of ORBRequest, if it is then process it, else pass it on to the delegate.

skanderturki: OK

jlhughes:
@skanderturki why did you close this?

skanderturki:
I made a mistake to commit to my main forked branch. So I needed to close
it to be able to work on other tasks. I will republish the code on another
branch

PR #2550, opened 2019-08-27
untransianted required properties

michaelconnor468:


src/foam/nanos/session/Session.js
```diff
@@ -51,7 +51,8 @@ foam.CLASS({
     },
     {
       class: 'Long',
-      name: 'uses'
+      name: 'uses',
+      transient: true
```
TW80000: Sorry, this should just be `storageTransient`, not `networkTransient` as well. If it were `networkTransient` then admins wouldn't be able to see how many uses sessions have.

TW80000 (changes requested):

TW80000 (approved):

jlhughes (commented):
I would expect lastUsed to also required storageTransient.  
See discussion in https://github.com/foam-framework/foam2/pull/2542

jlhughes (changes requested):
Make lastUsed storageTransient. 
On restart the session will effectively be reset.  We'll err with this rather than write every session update for now.

jlhughes (approved):

src/foam/nanos/session/Session.js
```diff
@@ -65,6 +66,7 @@ foam.CLASS({
     {
       class: 'Long',
       name: 'uses'
```
jlhughes:
```suggestion
      name: 'uses',
```

PR #2551, opened 2019-08-27
ScrollTableView updates

mcarcaso:
The original goal was to make scrolling by a large amount more performant. To do so, the idea of scrolling up or down a single page was removed and now the view is capable of figuring out what pages need to be rendered based on the current scroll position.


src/foam/u2/view/TableView.js
```diff
@@ -29,6 +29,7 @@ foam.CLASS({
       overflow: hidden;
       position: sticky;
       top: 0;
+      z-index: 1000;
```
adamvy: Is this necessary?  Or can we alternatively set the z-index of the entire tableview from the ScrollTableView ?

mcarcaso: Without this it gets covered by the contents below it that scroll but I can look into alternatives here.

mcarcaso: Updated.

src/foam/u2/view/ScrollTableView.js
```diff
@@ -218,45 +103,68 @@
...
+    },
+    {
+      class: 'Int',
+      name: 'currentTopPage_',
+      expression: function(numPages_, scrollPos_, scrollHeight) {
+        var scrollPercent = scrollPos_ / scrollHeight;
+        var topPage = Math.floor(scrollPercent * numPages_);
+        var topPageProgress = (scrollPercent * numPages_) % 1;
+        if ( topPageProgress < 0.5 ) topPage = topPage - 1;
+        return Math.min(Math.max(0, numPages_ - 3), Math.max(0, topPage));
```
TW80000: I assume the 3 here is because we show 3 pages at a time? If so it might make sense to move that number into a constant and document it.

mcarcaso:
Just noticed a regression in the width of the table. Will update...

TW80000 (commented):
Rows in the table can't be clicked.

mcarcaso:
PTAL. I'll merge this in the next day or so if there are no other comments/concerns.

PR #2555, opened 2019-08-28
Swift fix2

michaelconnor468:


src/foam/box/LocalBoxRegistry.js
```diff
@@ -93,18 +93,16 @@ return (foam.box.ExportBox)registration;
...
-])
-exportBox = service?.clientBox(exportBox) ?? exportBox
-
-let registration = Registration_create([
-  "exportBox": exportBox,
-  "localBox": service?.serverBox(box) ?? box
-])
-registry_[name] = registration
-return registration.exportBox
+var exportBox: foam_box_Box = foam_box_ExportBox()
```
mcarcaso: Since it's already being required, you should be able to do a `ExportBox_create` here which will create the object through the context and set the context properly. You can also set args when doing that so you can make it more similar to js. Something like:
```
ExportBox_create(args: [
  "localBox": box,
  "messengerBox": SubBox_create(args: [
    "name": name,
    "delegate": me
  ])
])
```

src/foam/box/LocalBoxRegistry.js
```diff
@@ -93,18 +93,13 @@ return (foam.box.ExportBox)registration;
...
-])
-exportBox = service?.clientBox(exportBox) ?? exportBox
-
-let registration = Registration_create([
-  "exportBox": exportBox,
-  "localBox": service?.serverBox(box) ?? box
-])
-registry_[name] = registration
-return registration.exportBox
+return foam_box_ExportBox([
```
mcarcaso: could you use ExportBox_create instead? Then you don't need to provide context because that'll be done for you.

src/foam/box/LocalBoxRegistry.js
```diff
@@ -93,18 +93,13 @@ return (foam.box.ExportBox)registration;
...
+  "localBox": box,
+  "messengerBox": foam_box_SubBox([
```
mcarcaso: same here. SubBox_create

PR #2556, opened 2019-08-28
Setting a Canvas Width Default To Center Everything

tharmaman:


src/foam/graphics/TreeGraph.js
```diff
@@ -329,7 +329,7 @@
         var delta = Math.abs(this.width - width) / width;
 
         if ( delta > 0.01 ) {
-          this.width = width;
+          this.width = width > 1600 ? width : 1600;
```
mcarcaso: `this.width = Math.min(width, 1600);`

mcarcaso (approved):

src/foam/graphics/TreeGraph.js
kgrgreer: Why 1600?

tharmaman: Just changed it to use the clientWidth of the canvasContainer instead which makes more sense than the hard coded 1600.

tharmaman: Also added in mike's suggestion of Math.min

PR #2557, opened 2019-08-28
DO_NOT_MERGE: Fix SafetyUtil.compare

lchanmann:
Some properties on a model might not be always Comparable e.g., Rule.action.

Add non-Comparable checks in SafetyUtil.compare to account for comparing these might-be-non-Comparable properties.

### DIFF
--- a/src/foam/util/SafetyUtil.java
+++ b/src/foam/util/SafetyUtil.java
@@ -18,6 +18,20 @@ public class SafetyUtil {
     if ( o2 == null ) return  1;
     if ( o1 == null ) return -1;
 
+    // Number subtypes (Long, Integer, etc.) are type-specific comparable and
+    // can only be compareTo the same type eg., Long can only compareTo Long.
+    //
+    // Converting o1 and o2 to double then comparing allows FOAM property types
+    // that are java.lang.Number compatible such as Long, Int, Short, Byte,
+    // Float and Double to be able to compareTo one another.
+    //
+    // Ex. The generated `isDefaultValue` of a Long property compares its value
+    // to 0, which is an integer.
+    //
+    // This also allows predicates such as EQ, GT, LT and friends comparison
+    // without explicit casting.
+    //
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

kgrgreer:
I'm hesitant to do this because this is the single most called method in the system. Or, at least, compartors in general are. Do the regular String and Number Property types call this method or some other comparator? How does this affect system performance?

lchanmann:
> Do the regular String and Number Property types call this method or some other comparator? How does this affect system performance?

String and Number property types would call this method when `compare(o1, o2)` is called. Number type comparison would be handled by the preceding if block lines 21-28 and the checks here would not affect its performance at all.

For String type, I run some benchmarks and one result is shown below.

- With the changes:
```
time SafetyUtil.compare(a, a) for 100000 calls:
 .  avg: 0.001551, min: 0.001450, max: 0.932794, total: 172.704258
time SafetyUtil.compare(b, a) for 100000 calls:
 .  avg: 0.001710, min: 0.001479, max: 1.813468, total: 173.594318
time SafetyUtil.compare(a, b) for 100000 calls:
 .  avg: 0.001678, min: 0.001457, max: 0.063818, total: 167.641391
```

- Without the changes:
```
time SafetyUtil.compare(a, a) for 100000 calls:
 .  avg: 0.001629, min: 0.001477, max: 1.670524, total: 172.419300
time SafetyUtil.compare(b, a) for 100000 calls:
 .  avg: 0.001686, min: 0.001489, max: 0.060309, total: 171.930042
time SafetyUtil.compare(a, b) for 100000 calls:
 .  avg: 0.001697, min: 0.001477, max: 0.078580, total: 172.618523
```

The average time difference for a single call is at sub-millisecond. For a total of 100k calls, the performance difference also seems to be not noticeable.

@kgrgreer @jlhughes

lchanmann:
@kgrgreer are we good with the changes?

PR #2559, opened 2019-08-29
expired sessions last used time + ttl should be before (<=) current time

noodlemoodle:


jlhughes (approved):
Ha, so much for code reviews.

PR #2561, opened 2019-08-29
[CPF-2717] [part 2] Remove javafx.util dependancy

KernelDeimos:


jlhughes (approved):

src/foam/nanos/http/LiveScriptBundler.java
```diff
@@ -384,4 +383,17 @@ private void log_(String evt, String msg) {
       logger.info(this.getClass().getSimpleName(), eventStr, msg);
     }
   }
+
+  private class Pair<K,V> {
```
kgrgreer, line 387: Very little code anyway, but could we just use the standard:

https://docs.oracle.com/javase/8/docs/api/java/util/AbstractMap.SimpleEntry.html

PR #2569, opened 2019-09-04
[CPF-2781] Add MenuDefinedViewList

KernelDeimos:
This allows separation of views that are often used together in a simple layout (i.e. rows or columns) without creating additional views to combine them.

For example, in a Capability info page there's a back button at the top, but it may be desirable to display just the capability info by itself. To navigate to the capability info view with a back button the views can be combined as follows:
```
self.stack.push({
  class: 'foam.nanos.menu.MenuDefinedViewList',
  containerView: 'foam.u2.layout.Rows',
  childViews: [
    {
      class: 'foam.nanos.crunch.ui.CapabilityPageMenu'
    },
    {
      class: 'foam.nanos.crunch.ui.CapabilityInfoView'
    }
  ]
});
```

This is added to the foam.nanos.menu since it may be useful elsewhere, including inside a menuDAO entry.

### DIFF
--- a/src/foam/nanos/menu/MenuDefinedViewList.js
+++ b/src/foam/nanos/menu/MenuDefinedViewList.js
@@ -1,0 +1,35 @@
+foam.CLASS({
+  package: 'foam.nanos.menu',
+  name: 'MenuDefinedViewList',
+  extends: 'foam.u2.View',
+
+  properties: [
+    {
+      name: 'containerView',
+      class: 'foam.u2.ViewSpec'
+    },
+    {
+      name: 'childViews',
+      class: 'FObjectArray',
+      of: 'foam.u2.View'
+    }
+  ],
+
+  methods: [
+    function initE() {
+      var self = this;
+      var view = this;
+
+      view = view
+        .start(self.containerView);
+      
+      for ( var i = 0; i < self.childViews.length; i++ ) {
+        view = view
+          .start(self.childViews[i]).end();
+      }
+
+      view = view
+        .end();
+    }
+  ]
+});


src/foam/nanos/menu/MenuDefinedViewList.js
```diff
@@ -0,0 +1,35 @@
...
+
+  properties: [
+    {
+      name: 'containerView',
+      class: 'foam.u2.ViewSpec'
+    },
+    {
+      name: 'childViews',
+      class: 'FObjectArray',
+      of: 'foam.u2.View'
```
TW80000, line 14:
```suggestion
      of: 'foam.u2.ViewSpec'
```

src/foam/nanos/menu/MenuDefinedViewList.js
```diff
@@ -0,0 +1,35 @@
...
+
+  methods: [
+    function initE() {
+      var self = this;
+      var view = this;
+
+      view = view
+        .start(self.containerView);
+      
+      for ( var i = 0; i < self.childViews.length; i++ ) {
```
TW80000, line 26: Could use `.forEach(self.childViews, function() { /* ... */ })` instead.

src/foam/nanos/menu/MenuDefinedViewList.js
```diff
@@ -0,0 +1,35 @@
...
+        view = view
+          .start(self.childViews[i]).end();
```
TW80000, line 28: When you have `.start(foo).end()` you can just do `.tag(foo)` instead. You only need start and end when you want to do other calls in between.

TW80000 (commented):
We already have [DualView](https://github.com/foam-framework/foam2/blob/master/src/foam/u2/view/DualView.js) for when you want a view of two other arbitrary views. Not sure if that satisfies your use case though.

We could also name this `CompositeView` or `MultiView` instead since it's quite general.

That being said, I can't imagine there'd be many places we'd want to use this rather than just creating a new view, but I could be wrong.

jlhughes:
Replaced by MultiView.

PR #2574, opened 2019-09-05
Create ZoomMapView

mcarcaso:
Demo: https://imgur.com/0QvHQTi
Can be accessed by starting a local server from the foam2 dir:
`python -m SimpleHTTPServer`
and going to:
http://localhost:8000/?model=foam.demos.graphics.ZoomMapViewDemo

### DIFF
--- a/src/foam/graphics/GreekView.js
+++ b/src/foam/graphics/GreekView.js
@@ -16,22 +16,30 @@ foam.CLASS({
     {
       class: 'Color',
       name: 'viewBorder',
-      value: 'white'
+      value: 'red'
     },
     {
       class: 'Int',
       name: 'viewBorderWidth',
-      value: 20
+      value: 5 
     },
     {
       class: 'Int',
       name: 'navBorderWidth',
-      value: 20
+      value: 10
     },
     {
       class: 'Map',
       name: 'viewPortPosition',
-      value: { x: 0, y: 0 }
+      preSet: function(_, n) {
+        if ( ! this.hasOwnProperty('navView_') || ! this.hasOwnProperty('viewPortView_') ) {
+          return { x: 0, y: 0 }
+        }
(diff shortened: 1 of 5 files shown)


src/foam/graphics/GreekView.js
```diff
@@ -0,0 +1,189 @@
+foam.CLASS({
+  package: 'foam.graphics',
+  name: 'GreekView',
```
kgrgreer: Let's rename this LegendView or MapView or ZoomMapView or PanView or something else. I can't find anywhere online that this is referred to as a GreekView. I have an encyclopedia of visual components at home that I'll check.

src/foam/graphics/GreekView.js
```diff
@@ -0,0 +1,189 @@
...
+  extends: 'foam.graphics.CView',
+  requires: [
+    'foam.graphics.CView',
```
kgrgreer: sort requires

src/foam/graphics/GreekView.js
```diff
@@ -0,0 +1,189 @@
+foam.CLASS({
```
kgrgreer: missing copyright

src/foam/graphics/GreekView.js
```diff
@@ -0,0 +1,189 @@
...
+      this.add(viewPort);
+      this.add(navContainer);
+      
+    }
+  ]
+});
+
+foam.CLASS({
+  package: 'foam.graphics',
+  name: 'GreekViewDemo',
```
kgrgreer: Move to own file (under different package).

src/foam/graphics/GreekView.js
mcarcaso: Went with ZoomMapView

PR #2575, opened 2019-09-05
Refactored FixedSizedDAO

jlhughes:
FixedSizeDAO which purges delegate MDAO based on Comparator, Predicate, Skip size.

Test:

of = foam.nanos.auth.User.getOwnClassInfo();

dao = new foam.dao.FixedSizeDAO2.Builder(x)
  .setDelegate(new foam.dao.MDAO(of))
  .setComparator(foam.mlang.MLang.DESC(of.getAxiomByName("lastModified")))
  .setPredicate(foam.mlang.MLang.TRUE)
  .setSize(10)
  .build();

long startTime = new java.util.Date().getTime();
for ( int i = 100 ; i <= 200 ; i++ ) {
    dao.put(new foam.nanos.auth.User.Builder(x)
      .setId(i)
      .setFirstName("Test " + i)
      .setLastModified(new java.util.Date(startTime + i))
      .build());
}

print(dao.select(new foam.mlang.sink.Count()).getValue());

print(dao.select(foam.mlang.MLang.MIN(of.getAxiomByName("firstName"))));
print(dao.select(foam.mlang.MLang.MAX(of.getAxiomByName("firstName"))));
Output:

10
value: Test 191, arg1: foam.nanos.auth.User.firstName
value: Test 200, arg1: foam.nanos.auth.User.firstName


src/foam/dao/EasyDAO.js
```diff
@@ -131,6 +132,13 @@ if ( head instanceof foam.dao.MDAO ) {
   }
 }
 
+if ( getFixedSizeDAO() &&
+     ( head instanceof foam.dao.MDAO || head instanceof foam.dao.NullDAO ) &&
```
TW80000: What's the idea behind these checks? If `fixedSizeDAO` is set to true and we're silently not adding that decorator in certain situations then it seems like it would be very easy to introduce unintended bugs where we expect the decorator to be there but it isn't.

src/foam/dao/EasyDAO.js
```diff
@@ -521,6 +529,16 @@ return delegate;
       name: 'lastModifiedByAware',
       class: 'Boolean',
       value: false
+    },
+    {
+      name: 'fixedSizeDAO',
```
TW80000: Alternative names: `fixedSize: true` and `size: 10000`. The other properties don't include "DAO" in the name so it would be nice to stay consistent.

src/foam/dao/FixedSizeDAO.js
```diff
@@ -16,6 +16,7 @@ foam.CLASS({
     prohibitively large. Instead, this DAO will store some of that data in
     memory, but it will start to replace old data after a certain point to
     prevent infinite growth.
+    NOTE: Install via EasyDAO. It is important that when removeFromDelegate is true (the default) that this DAO is placed between the JDAO and it's delegate which is typically a MDOA or NullDAO.
```
TW80000: Typos: a → an, MDOA → MDAO, it's → its

src/foam/dao/FixedSizeDAO.js
```diff
@@ -25,34 +26,44 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'Int',
-      name: 'nextIndex'
+      documentation: 'when removing from internal circular buffer also remove from delegate DAO.',
```
TW80000: Capitalization: when → When

src/foam/dao/FixedSizeDAO.js
```diff
@@ -82,7 +93,12 @@ foam.CLASS({
   } finally {
```
TW80000: There's an off-by-one bug in the few lines above where we won't replace the item at index 0 after wrapping around since we add 1 to `insertAt` after setting it to 0.

src/foam/dao/FixedSizeDAO.js
```diff
@@ -82,7 +93,12 @@ foam.CLASS({
...
     getLock().unlock();
   }
+  foam.core.FObject discard = getFixedSizeArray()[insertAt];
```
TW80000: It seems really wasteful of memory that we're storing the entire object in an array in memory as well as potentially storing the objects in the delegate MDAO. Unrelated to your changes here, but maybe we should rethink that. Maybe we could just be storing the ids?

src/foam/dao/FixedSizeDAO.js
TW80000: Why does this have to come after the JDAO? It shouldn't matter, should it? If we're recording removals in the JDAO, that's what we want anyway since when we read from the journal after a restart, the items that were removed before the restart shouldn't be accessible anyway.

jlhughes: We don't really want to delete anything - consider emails or logs.   We only want to remove them from memory.   Similar with Rule History. 
Archive servers for example give us access to this older data.

src/foam/dao/FixedSizeDAO.js
jlhughes: This would suggest a FixedSizeMDAO - if we are storing in memory. 
The FixedSizeDAO must know if the delegate DAO is actually storing, consider LoggingDAO and which is setup with an NullDAO rather than an MDAO.

PR #2576, opened 2019-09-06
Add Duration property type and use in Script and Session

TW80000:
The `Duration` property extends `Long` and is used when you want to store a value in milliseconds that represents an amount of elapsed time. A `tableCellFormatter` is included for the new property that displays the value in a human-readable format. For example, it will display '8h' instead of '28800000'. It also omits units for which the value is empty, so it will display '8h 1ms' instead of '28800001' or '8h 0m 0s 1ms'.

Fixed a bug in the `tableCellFormatter` function for `Script.lastDuration` that wasn't subtracting from the value when splitting the duration up into components.

### DIFF
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -249,6 +249,25 @@ foam.CLASS({
 
 foam.CLASS({
   package: 'foam.core',
+  name: 'Duration',
+  extends: 'Long',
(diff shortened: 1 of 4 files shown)


jlhughes (approved):
Nice.

PR #2578, opened 2019-09-06
[CPF-2694] Check cookies for sessionId in SessionWebAgent

KernelDeimos:
Adds a check for `sessionId` in request cookies (the cookie is sent to any `/services` url), making `req.getParameter`<sup>[1](#ftnt1)</sup> a fallback.

Additionally, adds a check for the query string containing `sessionId` and displays a warning if this is the case.

<a name="ftnt1">1</a>: Post data and querystring; order of precedence is not specified in Java docs.

### DIFF
--- a/src/foam/nanos/http/SessionWebAgent.java
+++ b/src/foam/nanos/http/SessionWebAgent.java
@@ -16,6 +16,7 @@ import foam.nanos.logger.Logger;
 import foam.nanos.session.Session;
 import foam.util.SafetyUtil;
 
+import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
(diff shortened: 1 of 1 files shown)


src/foam/nanos/http/SessionWebAgent.java
```diff
@@ -44,12 +45,34 @@ public void execute(X x) {
...
     try {
-      // check session id
-      String sessionId = req.getParameter("sessionId");
+      String sessionId = "";
+
+      // check for session id in cookie
+      Cookie[] cookies = req.getCookies();
+      for ( Cookie c : cookies ) {
+        if ( c.getName().equals("sessionId") ) {
+          sessionId = c.getValue();
```
TW80000, line 54: Super minor, but you could add a `break` after this line since we don't need to loop over the rest of the cookies.

TW80000 (approved):

src/foam/nanos/http/SessionWebAgent.java
KernelDeimos: Updated

PR #2582, opened 2019-09-09
removed select support

michaelconnor468:


KateMengjunXia (changes requested):
Why not just remove the whole overriding 'select_' method?

kgrgreer:
Do we still use the HidePropertiesSink anywhere? Can we remove it now?

michaelconnor468:
I don’t think so

michaelconnor468:
- Kate

PR #2584, opened 2019-09-09
Fixed choice with other

KateMengjunXia:
https://nanopay.atlassian.net/browse/CPF-2746

Fixed the issue with `<ChoiceView>` where `<choice>` doesn't get updated after `<choices>` are populated from `<dao>`.

Updated `<ChoiceWithOtherView>` to reset `<otherData_>` to default when non-other data is chosen.

### DIFF
--- a/src/foam/u2/view/ChoiceView.js
+++ b/src/foam/u2/view/ChoiceView.js
@@ -36,6 +36,12 @@ foam.CLASS({
   `,
 
   properties: [
+    'savedData',
+    {
+      class: 'Boolean',
+      name: 'savedDataSet',
+      value: false
+    },
     {
       class: 'String',
       name: 'name',
@@ -246,6 +252,10 @@ foam.CLASS({
       for ( var i = 0 ; i < choices.length ; i++ ) {
         if ( foam.util.equals(choices[i][0], data) ) return choices[i];
       }
+      if ( this.dao && ! this.savedDataSet ) {
+        this.savedData = data;
+        this.savedDataSet = true;
+      }
       return null;
     },
 
@@ -284,6 +294,7 @@ foam.CLASS({
 
         p.then(function(a) {
           this.choices = a.map(this.objToChoice);
+          if ( this.savedData !== undefined ) this.choice = this.findChoiceByData(this.savedData);
           if ( this.data == null && this.index === -1 ) this.index = this.placeholder ? -1 : 0;
         }.bind(this));
       }
--- a/src/foam/u2/view/ChoiceWithOtherView.js
+++ b/src/foam/u2/view/ChoiceWithOtherView.js
@@ -57,7 +57,11 @@ foam.CLASS({
       class: 'Boolean',
       name: 'showOther_',
       expression: function(choiceData_, otherKey) {
-        return foam.util.equals(choiceData_, otherKey);
+        if ( foam.util.equals(choiceData_, otherKey) ) {
+          return true;
+        }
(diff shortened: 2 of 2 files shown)


kgrgreer:
I don't understand. Can you explain this to me in person? Thx

src/foam/u2/view/ChoiceView.js
```diff
@@ -246,6 +252,10 @@ foam.CLASS({
       for ( var i = 0 ; i < choices.length ; i++ ) {
         if ( foam.util.equals(choices[i][0], data) ) return choices[i];
       }
+      if ( this.dao && ! this.savedDataSet ) {
```
adamvy, line 255: Presumably you'd need to do this in findChoiceByText as well?

KateMengjunXia: When `<choice>` is updated, its `<postSet>` will update `<text>` whose `<postSet>` will call `<findChoiceByText>` if necessary

adamvy:
@KateMengjunXia The root problem may actually be the postSet in 'text'.  In 'data' we have a postset that if findChoiceByData returns false, we just generate a choice to represent the current value.  But we don't have the same login in the postSet of 'text'.

```
    {
      name: 'text',
      postSet: function(o, n) {
        if ( o !== n ) this.choice = this.findChoiceByText(n);
      }
    },
```
vs
```
    {
      name: 'data',
      postSet: function(o, n) {
        if ( o !== n ) this.choice = this.findChoiceByData(n) || [n, n];
      }
    },
```

If we add the `|| [n, n]` case to the 'text' postSet, does that also fix the problem?

KateMengjunXia:
> @KateMengjunXia The root problem may actually be the postSet in 'text'. In 'data' we have a postset that if findChoiceByData returns false, we just generate a choice to represent the current value. But we don't have the same login in the postSet of 'text'.
> 
> ```
>     {
>       name: 'text',
>       postSet: function(o, n) {
>         if ( o !== n ) this.choice = this.findChoiceByText(n);
>       }
>     },
> ```
> 
> vs
> 
> ```
>     {
>       name: 'data',
>       postSet: function(o, n) {
>         if ( o !== n ) this.choice = this.findChoiceByData(n) || [n, n];
>       }
>     },
> ```
> 
> If we add the `|| [n, n]` case to the 'text' postSet, does that also fix the problem?

Yes
EDIT: actually that raises another problem, if we choose 'Other' and the data doesn't appear again later on

adamvy:
This class needs some better documentation about what's going on lol

mcarcaso:
I think a better fix is here:
https://github.com/foam-framework/foam2/pull/2586

jlhughes:
@TW80000 is this required after https://github.com/foam-framework/foam2/pull/2689

KateMengjunXia:
> @TW80000 is this required after #2689

I believe https://github.com/foam-framework/foam2/pull/2586 has replaced this PR

TW80000:
> > @TW80000 is this required after #2689
> 
> I believe #2586 has replaced this PR

@KateMengjunXia Can you please test the issue in the ticket again on the master branch to see if this PR can be closed?

KateMengjunXia:
> > > @TW80000 is this required after #2689
> > 
> > 
> > I believe #2586 has replaced this PR
> 
> @KateMengjunXia Can you please test the issue in the ticket again on the master branch to see if this PR can be closed?

Just double checked, the issue had been fixed. Will close this PR.

PR #2586, opened 2019-09-10
Make the ChoiceWithOtherView react to choiceView's choices changing.

mcarcaso:


src/foam/u2/view/ChoiceWithOtherView.js
```diff
@@ -66,7 +66,8 @@ foam.CLASS({
     ['', 'propertyChange.choiceData_', 'toData'],
     ['', 'propertyChange.otherData_', 'toData'],
     ['', 'propertyChange.data', 'fromData'],
-    ['', 'propertyChange.choiceView_', 'fromData']
+    ['', 'propertyChange.choiceView_', 'fromData'],
+    ['choiceView_', 'propertyChange.choices', 'fromData'],
```
TW80000, line 70: If `choiceView_` changes does this listener get attached to the new value? Is that even worth caring about?

TW80000 (approved):
Didn't test, but I have no issue with this. I'm also not very familiar with ChoiceWithOtherView though, so don't count my opinion on this very highly.

PR #2590, opened 2019-09-10
Max date and cron fix

mcarcaso:


jlhughes:
NeverSchedule has been removed - this is no longer required.

PR #2596, opened 2019-09-11
Handled apiColumns

minsunny:
https://nanopay.atlassian.net/browse/CPF-2243

Just as we have tableColumns in a model, we also need digColumns in a model.

The digColumns would specify the default columns which would be returned from a find, create, or update of an entity. If extra columns were requested in a find call, then they would be returned as long as the user had permissions to access those fields.

### DIFF
--- a/src/foam/nanos/dig/DIG.js
+++ b/src/foam/nanos/dig/DIG.js
@@ -54,6 +54,19 @@ foam.CLASS({
     'cmd',
     'format',
     {
+      class: 'StringArray',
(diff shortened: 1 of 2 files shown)


TW80000:
Shouldn't this PR also include the code that's going to use the new axiom?

jlhughes:
@minsunny were you planning model updates to use this feature in another PR?

jlhughes:
Replaced by PROJECTION.

PR #2605, opened 2019-09-12
DO_NOT_MERGE ROPE

noodlemoodle:
TODO add caching


michaelconnor468:
@googlebot I consent.

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,259 @@
...
+        class: 'Map',
+        javaType: 'java.util.Map<String, java.util.Map<String, java.util.List<String>>>'
+      },
+      {
+        name: 'relationshipMap',
+        class: 'Map',
+        javaType: 'java.util.Map<String, java.util.List<String>>'
+      },
+      {
+        name: 'isInverse',
```
TW80000: Is this necessary? If it was the inverse, wouldn't the source and target keys be switched and the relationship key be different, meaning it's a different ROPE entirely?

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,259 @@
...
+          {
+            name: 'propertyKey',
+            javaType: 'String'
+          }
+        ],
+        javaType: 'List<String>',
+        documentation: `
+        this function returns a list of relationshipKeys to filter the ropeDAO with in intermediate steps of the ropeSearch
+        args :
+          rope - the current rope
```
TW80000: This isn't an arg

src/foam/nanos/rope/CompositeROPE.js
```diff
@@ -0,0 +1,66 @@
+
```
TW80000, line 6: Missing license header.

src/foam/nanos/rope/CompositeROPE.js
```diff
@@ -0,0 +1,66 @@
...
+foam.CLASS({
+    package: 'foam.nanos.rope',
```
TW80000: Please fix indentation

src/foam/nanos/rope/CompositeROPE.js
```diff
@@ -0,0 +1,66 @@
...
+    extends: 'foam.nanos.rope.ROPE',
+  
+    properties: [
+      {
+        name: 'compositeRopes',
+        class: 'List',
+        javaType: 'java.util.List<ROPE>'
+      }
+    ]
+  })
```
TW80000: Missing semicolon

src/foam/nanos/rope/CompositeROPE.js
```diff
@@ -0,0 +1,66 @@
+
+foam.CLASS({
+    package: 'foam.nanos.rope',
+    name: 'CompositeROPE',
+    extends: 'foam.nanos.rope.ROPE',
+  
+    properties: [
+      {
+        name: 'compositeRopes',
```
TW80000: I think `children`, `objects` or `ropes` would be a better name. This model is a composite ROPE, but its children aren't necessarily composite ROPEs.

src/foam/nanos/rope/CompositeROPE.js
```diff
@@ -0,0 +1,66 @@
...
+    javaImports: [
+      'foam.core.FObject',
+      'foam.core.X',
+      'java.util.List'
+    ],
+  
+    methods: [
+      {
+        name: 'check',
+        javaCode: `
```
TW80000: This method can be simplified to:

```Java
          for ( ROPE rope : getCompositeRopes() ) {
            if ( ! rope.check(x, obj, relationshipKey, crudKey, propertyKey) ) return false;
          }
          return true;
```

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,259 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+    package: 'foam.nanos.rope',
```
TW80000: Please fix indentation

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,259 @@
...
+      
+            for ( FObject junctionObj : junctionObjs ) {
+              FObject sourceObj = (FObject) (((DAO) x.get(getSourceDAOKey())).find((retrieveProperty(junctionObj, junctionObj.getClass(), "get", sourcePropertyName))));
+              sourceObjs.add(sourceObj);
+            }
+            break;
+          case "1:*" :
+            if ( getIsInverse() ) {
+              DAO rDAO = retrieveProperty(obj, obj.getClass(), "get", getRelationshipKey(), x);
+              sourceObjs = ((ArraySink) rDAO.where(INSTANCE_OF(((DAO) x.get(getSourceDAOKey())).getOf())).select(new ArraySink())).getArray();
```
TW80000: What's the point of the INSTANCE_OF query here?

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,259 @@
...
+        // if a there are no results, return false
+        if ( nextRelationships == null || nextRelationships.size() == 0 ) return false;
+    
+        // get the list of sourceObjs that have a relationship with the targetObj
+        List<FObject> sourceObjs = getSourceObjects(x, obj);
+    
+        DAO ropeDAO = (DAO) x.get("ropeDAO");
+        for ( FObject sourceObj : sourceObjs ) {
+          if ( sourceObj == null ) continue;
+          if ( nextRelationships.contains("__terminate__") ) {
```
TW80000, line 104: Is it possible for `nextRelationships` to contain both `"__terminate__"` and something else? If so, the other thing(s) won't have a chance to be followed.

src/services
```diff
@@ -712,3 +712,20 @@ p({
   "serve": true,
   "serviceClass": "foam.nanos.session.SimpleSessionService"
 })
+
+p({
+  "class":"foam.nanos.boot.NSpec",
```
TW80000: Please put a space after the colon here and on the relevant lines below.

src/services
```diff
@@ -712,3 +712,20 @@ p({
...
+  "name":"ropeDAO",
+  "lazy":false,
+  "serve":true,
+  "serviceScript":"""
+    return new foam.dao.EasyDAO.Builder(x)
+      .setAuthorize(true)
```
TW80000: This is the default and can be therefore be omitted.

src/foam/nanos/rope/RopeDocumentation.md
```diff
@@ -0,0 +1,113 @@
+![ROPE Logo](rope.png)
+
+# TODO DOC IS OUTDATED
```
TW80000: Still TODO?

src/foam/nanos/rope/ROPEAuthorizer.js
```diff
@@ -0,0 +1,224 @@
...
+      javaType: 'boolean',
+      args: [
+        { name: 'x', javaType: 'X' },
+        { name: 'obj', javaType: 'FObject' },
+        { name: 'targetDAOKey', javaType: 'String' },
+        { name: 'crudKey', javaType: 'String' },
+        { name: 'propertyKey', javaType: 'String' }
+      ],
+      javaCode: `    
+      if ( getRopeDAO() == null ) setRopeDAO((DAO) x.get("ropeDAO"));
```
TW80000: Please indent all `javaCode` strings appropriately.

ie: 

```Java
javaCode: `
  foo();
`
```

instead of

```Java
javaCode: `
foo();
`
```

src/foam/nanos/rope/ROPEAuthorizer.js
```diff
@@ -0,0 +1,224 @@
...
+          name: 'x',
+          javaType: 'X'
+        },
+        {
+          name: 'obj',
+          javaType: 'FObject'
+        }
+      ],
+      javaCode: `
+      if ( ((User) x.get("user")).getId() == User.SYSTEM_USER_ID ) return;
```
TW80000: Note to self: Think of where this should go.

src/foam/nanos/rope/ROPEAuthorizer.js
```diff
@@ -0,0 +1,224 @@
...
+      while ( i.hasNext() ) {
+        String propName = (String) i.next();
+        propertyUpdates.add(propName);
+      }
+      return propertyUpdates;
+      `
+    },
+    {
+      name: 'authorizeOnCreate',
+      type: 'Void',
```
TW80000: `type`, `javaThrows`, and `args` are all defined on the interface and don't need to be included here.

src/foam/nanos/rope/ROPEAuthorizer.js
```diff
@@ -0,0 +1,224 @@
...
+    'java.util.ArrayList',
+    'java.util.List',
+    'java.util.Iterator',
+    'java.util.Map',
+    'static foam.mlang.MLang.*'
+  ],
+
+  properties: [
+    {
+      name: 'ropeDAO',
```
TW80000: Can't we do

```Java
imports: [
  'ropeDAO'
]
```

and use

```Java
DAO ropeDAO = (DAO) getRopeDAO();
```

when we need to access it instead?

src/foam/nanos/rope/ROPEAuthorizer.js
```diff
@@ -0,0 +1,224 @@
...
+    },
+    {
+      name: 'getPropertiesUpdated',
+      javaType: 'List<String>',
+      args: [
+        { name: 'oldObj', javaType: 'FObject' },
+        { name: 'obj', javaType: 'FObject' }
+      ],
+      documentation: `
+      this function returns a list of the names of properties changed on update or set on create
```
TW80000: Capitalize and indent.
```suggestion
        This function returns a list of the names of properties changed on update or set on create.
```

src/foam/nanos/rope/ROPEAuthorizer.js
```diff
@@ -0,0 +1,224 @@
...
+          name: 'x',
+          javaType: 'X'
+        }
+      ],
+      javaCode: `
+      return false;
+      `
+    },
+    {
+      name: 'getPermissionPrefix',
```
TW80000: This isn't a method on the `Authorizer` interface. Do we need it?

src/foam/nanos/rope/ROPEAuthorizer.js
```diff
@@ -0,0 +1,224 @@
...
+  properties: [
+    {
+      name: 'ropeDAO',
+      class: 'Object',
+      javaType: 'DAO'
+    },
+    {
+      name: 'targetDAOKey',
+      class: 'String',
+      javaType: 'String'
```
TW80000: Is this necessary?

TW80000 (commented):
Submitting first round of comments now, but I haven't reviewed the tests or documentation in detail yet.

src/foam/nanos/rope/ROPE.js
noodlemoodle: this is being used in the case of 1:* relationships where the the getter method for the source object(s) can either be find... or get... depending on which side it's on

src/foam/nanos/rope/ROPEAuthorizer.js
noodlemoodle: this got added into the authorizer interface a while ago, but its not really being used for anything. can i remove from interface?

src/foam/nanos/rope/RopeDocumentation.md
michaelconnor468: Yes, I will update this in a bit as I wrote the whole thing a little while ago before the addition of some new functionality

michaelconnor468: It's pretty dope though

michaelconnor468: 10/10 documentation

src/foam/nanos/rope/ROPE.js
noodlemoodle: yes it is possible, fixed

src/foam/nanos/rope/ROPEAuthorizer.js
noodlemoodle: yes, fixed

src/foam/nanos/rope/ROPEAuthorizer.js
TW80000: I see. Yeah, I don't think it makes sense there.

src/foam/nanos/auth/AuthorizableAuthorizer.java
```diff
@@ -17,11 +17,6 @@
   public AuthorizableAuthorizer(String permissionPrefix) {  
     permissionPrefix_ = permissionPrefix;
   }
-
-  public String getPermissionPrefix() {
-    return permissionPrefix_;
```
TW80000: We should get rid of this variable declaration as well and anywhere it's used.

src/foam/nanos/auth/StandardAuthorizer.java
```diff
@@ -30,10 +30,6 @@ public String createPermission(String op, Object id) {
     return permissionPrefix_ + "." + op + "." + id;
   }
 
-  public String getPermissionPrefix() {
-    return permissionPrefix_;
```
TW80000: We should get rid of this variable declaration as well and anywhere it's used.

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+&nbsp;
+
+## ROPE Description
+
+#### Idea behind it
+
+At an abstract level, ROPE utilizes the built in FOAM relationship framework to allow the user of the FOAM framework to perform authorization checks based on previously declared relationships and the level of authorization they are configured to be granted.
+
+The ROPE authorization system searches the tree formed by these relationships and their implied permissions to see if the object trying to perform an operation on another object is in some way connected to it through the relationship framework in a way that would imply it being able to have some desired permissions. This allows for the formation of very complex and flexible nets of permissions to be defined without the need for any of it to be hard coded into its corresponding locations in a hard to manage and modify mess; it is all defined by ROPEs.
```
TW80000: > tree

It's actually a graph, not a tree. Please update the docs to reflect this.

> relationships and their implied permissions

Relationships only imply permissions in the colloquial meaning of the words. `Permission` and `imply` are both words with concrete meanings in FOAM, and I think it's confusing to the reader to use these words in FOAM documentation when they don't refer to the FOAM concepts of permission and implies.

> the object trying to perform an operation on another object

It's always a user trying to perform an operation on an object. I think it would be clearer to state it that way as well.

> through the relationship framework

I think "through the relationship graph" is more correct.

> This allows for the formation of very complex and flexible nets of permissions to be defined without the need for any of it to be hard coded into its corresponding locations in a hard to manage and modify mess; it is all defined by ROPEs.

I think this can be stated in a simpler way. Maybe something like "ROPE can be reconfigured at runtime, which makes it more flexible than doing authorization checks in code, which can only be changed at compile time."

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
+![ROPE Logo](rope.png)
+
+# ROPE User Guide and Documentation
+
+&nbsp;
```
TW80000: No need for this

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+## ROPE Description
+
+#### Idea behind it
+
+At an abstract level, ROPE utilizes the built in FOAM relationship framework to allow the user of the FOAM framework to perform authorization checks based on previously declared relationships and the level of authorization they are configured to be granted.
```
TW80000:
```suggestion
Developers can use ROPE to perform authorization checks based on relationships between models. The idea is that we can determine whether or not a user is authorized to access a given object by following a chain of relationships between objects from the object being accessed back to the user. If a suitable chain is found, then the user is granted access. Each link in the chain (called a "Rope") can be configured to allow certain accesses via it and disallow others.
```

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
+![ROPE Logo](rope.png)
+
+# ROPE User Guide and Documentation
+
+&nbsp;
+
+## ROPE Description
+
+#### Idea behind it
```
TW80000: Not a useful heading imo

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+One of the key defining features that makes the ROPE algorithm's authorization so versatile and configurable is the transitivity it gains from its nature. An abstract example being some object ***A*** attempting to perform an operation on some other object ***C***. Although ***A*** may not be directly related to ***C***, it may be related to some intermediate object ***B*** which is itself related to object ***C***. Given the correct configuration of the ROPEs on these two relationships; object A can be granted certain permissions toward object ***C*** through its relationship to object ***B***.
+
+More generally this applies for properties themselves within the objects and the relationships between them. ROPE allows you to work at both the property level when it comes to authorization for much needed simplicity in acheiving fine-grained controll.
```
TW80000: Typos
```suggestion
More generally this applies for properties themselves within the objects and the relationships between them. ROPE allows you to work at both the property level when it comes to authorization for much needed simplicity in achieving fine-grained control.
```

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+More generally this applies for properties themselves within the objects and the relationships between them. ROPE allows you to work at both the property level when it comes to authorization for much needed simplicity in acheiving fine-grained controll. 
+
+&nbsp;
+
+#### Using ROPE with DAOs
+
+The ROPE authorization system can be utilized by the user of the framework by appending a ROPEAuthorizer decorator to any DAO object that requires authorization. This decorator follows the standard FOAM Authorizer interface and performs authorization checks dynamically as the dao is used using the ROPE relationship search algorithm under the hood.
```
TW80000:
```suggestion
You can use ROPE by adding an `AuthorizationDAO` decorator with a `ROPEAuthorizer` to any DAO.
```

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+      .setTargetDAOKey("transactionDAO")
+      .build()
+  )
+      .
+      . 
+      .
+  .build();
+```
+
+Permissions based on relationships can be configured by the user by creating a ROPE objects from the ROPE.js model and setting the properties accordingly and afterwards appending the object to the application's ropeDAO which will be utilized by the ROPE algorithm to perform authorization checks. Given a missing ROPE, the algorithm trivially assumes that all permissions are not granted on that object.
```
TW80000:
```suggestion
Access control based on relationships can be configured by the user by creating instances of the ROPE.js model, setting the properties appropriately, and putting the ROPE to `ropeDAO`, which is used by `ROPEAuthorizer` to perform authorization checks.
```

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+&nbsp;
```
TW80000: Unnecessary imo

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+#### Composition of ROPES
+
+There are a few helper ROPEs with which can be used to combine regular ROPEs to form more complex logical operations. There are known more formally as composite ropes. AND and OR ROPEs can be found in the compositeROPE.js file. These act as regular ROPEs except that under the hood they delegate their checks to other ropes composed within them. The OR composite authorizes if only one of the ROPEs it is composed with authorizes and the AND requires all composed ROPEs to authorize.
```
TW80000:
```suggestion
There are a few higher-order ROPEs that can be used to combine regular ROPEs to form more complex logical operations. They are known more formally as composite ropes. AND and OR ROPEs can be found in the `CompositeROPE.js` file. These act as regular ROPEs except that under the hood they delegate their checks to other ropes composed within them. The OR composite authorizes if only one of the ROPEs it is composed with authorizes and the AND requires all composed ROPEs to authorize. 
```

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+Permissions based on relationships can be configured by the user by creating a ROPE objects from the ROPE.js model and setting the properties accordingly and afterwards appending the object to the application's ropeDAO which will be utilized by the ROPE algorithm to perform authorization checks. Given a missing ROPE, the algorithm trivially assumes that all permissions are not granted on that object.
+
+&nbsp;
+
+#### Composition of ROPES
+
+There are a few helper ROPEs with which can be used to combine regular ROPEs to form more complex logical operations. There are known more formally as composite ropes. AND and OR ROPEs can be found in the compositeROPE.js file. These act as regular ROPEs except that under the hood they delegate their checks to other ropes composed within them. The OR composite authorizes if only one of the ROPEs it is composed with authorizes and the AND requires all composed ROPEs to authorize. 
+
+&nbsp;
```
TW80000: Also unnecessary imo

src/foam/nanos/rope/README.md
```diff
@@ -0,0 +1,175 @@
...
+&nbsp;
+
+## Technical Notes on the Proper Setup of ROPE Objects
+
+##### ROPEAuthorizer
+
+Akin to any standard foam authorizer, ROPEAuthorizer is an implementation of the corresponding interface and can be used interchangably on any DAO. Each ROPEAuthorizer has a corresponding targetDAOKey property which is utilized in finding the corresponding ROPEs to perform authorization checked with. The bulk of the algorithm is implemented within the ROPE model itself with an entrypoint in the check method to which the ROPEAuthorizer merely delegates.
```
TW80000:
```suggestion
Akin to any standard FOAM authorizer, `ROPEAuthorizer` is an implementation of the corresponding interface and can be used interchangeably on any DAO. Each `ROPEAuthorizer` has a corresponding `targetDAOKey` property which is utilized in finding the corresponding ROPEs to perform authorization checked with. The bulk of the algorithm is implemented within the ROPE model itself with an entry point in the check method to which the `ROPEAuthorizer` merely delegates.
```

TW80000 (commented):
Not finished, but submitting my comments so far now. Will continue later.

TW80000:
@noodlemoodle Is this still DO_NOT_MERGE? If not, can you please remove that from the title?

src/foam/nanos/auth/AuthorizableAuthorizer.java
noodlemoodle: this is still being used to create the permission string, which is used in `checkGlobalRead`/`checkGlobalRemove`

src/foam/nanos/auth/StandardAuthorizer.java
noodlemoodle: this is still being used to create the permission string, which is used in `checkGlobalRead`/`checkGlobalRemove`

src/foam/nanos/rope/README.md
noodlemoodle: removed

src/foam/nanos/rope/README.md
noodlemoodle: removed

src/foam/nanos/rope/README.md
noodlemoodle: updated

TW80000:
@noodlemoodle and I talked in person and think it's a good idea to model the maps used for `relationshipMap` and `crudMap`. I believe @noodlemoodle is working on that now. Writing this down to provide an update for anyone keeping track of this PR.

I'd also like it if we included a few examples of ROPE objects in their journalled representation (ie: what they look like in the journal file) in the documentation, since that's likely how developers are going to be creating/configuring ROPEs (at least until we're able to generate them from metadata on `foam.RELATIONSHIP` declarations).

src/foam/nanos/rope/README.md
```diff
@@ -34,14 +33,10 @@ transactionDAO = new foam.dao.EasyDAO.Builder(x)
 
 Access control based on relationships can be configured by the user by creating instances of the ROPE.js model, setting the properties appropriately, and putting the ROPE to `ropeDAO`, which is used by `ROPEAuthorizer` to perform authorization checks.
 
-&nbsp;
-
 #### Composition of ROPES
 
 There are a few higher-order ROPEs that can be used to combine regular ROPEs to form more complex logical operations. They are known more formally as composite ropes. AND and OR ROPEs can be found in the `CompositeROPE.js` file. These act as regular ROPEs except that under the hood they delegate their checks to other ropes composed within them. The OR composite authorizes if only one of the ROPEs it is composed with authorizes and the AND requires all composed ROPEs to authorize. 
 
-&nbsp;
```
michaelconnor468: &nbsp
&nbsp

src/foam/nanos/rope/README.md
```diff
@@ -73,16 +66,12 @@ One important method to note in the ROPE model is `check`, which handles the wor
...
-
 ##### To set up which permissions this ROPE will enable
 
 Both of the following methods of setting up a ROPE can be used in conjunction to achieve the desired functionality and are illustrated with some practical examples in the following section.
 
 ROPE works by checking which permissions are implied given any that a User might already have in a transitive fashion. The first thing that is checked whenever an authorization check takes place is the crud matrix. This relates an operation; create, read, update, or delete which maps to another mapping. This second mapping relates properties with lists of properties that one of which must be authorized to grant authorization to that property as a whole. Also contained within is a ***\_\_default\_\_*** property which can be searched to grant authorization after all other properties have been exhausted.
 
-&nbsp;
```
michaelconnor468: &nbsp
&nbsp

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,221 @@
...
+        } catch ( NoSuchMethodException | SecurityException | IllegalAccessException | IllegalArgumentException | InvocationTargetException e ) {
+          if ( e instanceof NoSuchMethodException && ((Class) objClass).getSuperclass() != null ) return retrieveProperty(obj, ((Class) objClass).getSuperclass(), prefix, propertyName, x);
+          System.err.println("ROPE ERROR: Attempted access on non-existant property ");
+          e.printStackTrace();
+        } 
+        return null;
+      `
+    },
+    {
+      name: 'getNextRelationships',
```
KernelDeimos: This looks like a property getter; could be renamed to avoid confusion

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,221 @@
...
+        relationshipKey - if the relationshipKey is specified, get list of keys in relationshipMap[relationshipKey]
+        crudKey - if the crudKey is specified, get list of keys in crudMap[crudKey]["__default__"], or
+        propertyKey - if the propertyKey is specified, get list of keys in crudMap[crudKey][propertyKey] or crudMap[crudKey]["__default__"], if the first returns null
+      `,
+      javaCode: `
+        List<String> next = new ArrayList<String>();
+    
+        if ( crudKey != null ) {
+          next = getCrudMap() == null ? null : getCrudMap().get(crudKey, propertyKey);
+        } else if ( relationshipKey != null && ! relationshipKey.equals("") ) {
```
KernelDeimos: `relationshipKey.equals("")` should be changed to `"".equals(relationshipKey)` (this is recommended in our code review guide to prevent NPEs)

src/foam/nanos/rope/ROPEAuthorizer.js
```diff
@@ -0,0 +1,152 @@
...
+        }
+        return propertyUpdates;
+      `
+    },
+    {
+      name: 'authorizeOnCreate',
+      javaCode: `
+        // TODO remove after SystemAuthorizer is created
+        User user = (User) x.get("user");
+        if ( user != null && ( user.getId() == User.SYSTEM_USER_ID || user.getGroup().equals("admin") || user.getGroup().equals("system") ) ) return;
```
KernelDeimos: `"system".equals(...` may be preferable for each of these

src/foam/nanos/rope/ROPEUtils.js
```diff
@@ -0,0 +1,167 @@
...
+    {
+      name: 'delete',
+      class: 'Map',
+      javaType: 'java.util.Map<String, NextRelationshipsList>',
+    }
+  ],
+  
+  axioms: [
+    {
+      name: 'javaExtras',
```
KernelDeimos, line 101: Could these be modelled as FOAM methods rather than a javaExtras axiom? Then documentation properties can be added to them, and other future FOAMy things will be compatible.

src/foam/nanos/rope/ROPEUtils.js
```diff
@@ -0,0 +1,167 @@
...
+           * Returns the list of relationshipKeys of ropes that can authorize a crud operation, depending on the
+           * crudKey and propertyKey.
+           * If the propertyKey is null, empty, or not found, return __default__ list.
+           * Otherwise, return the list mapped to by the property name.
+           * */
+          public List<String> get(CrudOperation crudKey, String propertyKey) {
+            Map<String, NextRelationshipsList> map = this.get(crudKey);
+            if ( map == null ) return null;
+    
+            if ( propertyKey == null || propertyKey.equals("") || ! map.containsKey(propertyKey) ) {
```
KernelDeimos: `propertyKey.equals("")` should be `"".equals(propertyKey)`

KernelDeimos:
Just an idea, what if instead of using strings as the value in CRUDMap, requiring special strings like `__terminate__`, a sort of `AbstractRopeCommand` FObject was used?

For example:
```
foam.CLASS({
  name: 'AllowPropertyROPECmd',
  extends: 'AbstractRopeCmd',
  properties: [ 'propertyName' ],
});

foam.CLASS({
  name: 'TerminateROPECmd',
  extends: 'AbstractRopeCmd',
});
```

@TW80000 is this related to what you mentioned above?

src/foam/nanos/rope/ROPEUtils.js
KernelDeimos: Oops, missed the comment

src/foam/nanos/rope/ROPE.js
noodlemoodle: 5/5 NPEs prevented

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,221 @@
...
+      required: true
+    },
+    {
+      name: 'relationshipKey',
+      class: 'String'
+    },
+    {
+      name: 'crudMap',
+      class: 'FObjectProperty',
+      of: 'foam.nanos.rope.CRUDMap'
```
noodlemoodle: add doc

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,221 @@
...
+    },
+    {
+      name: 'relationshipMap',
+      class: 'FObjectProperty',
+      of: 'foam.nanos.rope.RelationshipMap',
```
noodlemoodle, line 62: add doc

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,221 @@
...
+    },
+    {
+      name: 'isInverse',
+      class: 'Boolean',
+      value: false
```
noodlemoodle: default

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,221 @@
...
+    
+        DAO ropeDAO = (DAO) x.get("ropeDAO");
+        for ( FObject sourceObj : sourceObjs ) {
+          if ( sourceObj == null ) continue;
+          if ( nextRelationships.contains("__terminate__") ) {
+            if ( sourceObj instanceof User && ((User) sourceObj).getId() == ((User) x.get("user")).getId() ) return true;
+          }
+          for ( String nextRelationship : nextRelationships ) {
+            if ( nextRelationship.equals("__terminate__") ) continue;
+            List<ROPE> nextRopes = (List<ROPE>) ((ArraySink) ropeDAO.inX(x).where(AND(
```
noodlemoodle: put ropedao.inx outside loop

src/foam/nanos/rope/ROPE.js
```diff
@@ -0,0 +1,221 @@
...
+              EQ(ROPE.TARGET_DAOKEY, getSourceDAOKey()),
```
noodlemoodle: also can go outside loop and WHERE

TW80000:
> Just an idea, what if instead of using strings as the value in CRUDMap, requiring special strings like `__terminate__`, a sort of `AbstractRopeCommand` FObject was used?
> 
> For example:
> 
> ```
> foam.CLASS({
>   name: 'AllowPropertyROPECmd',
>   extends: 'AbstractRopeCmd',
>   properties: [ 'propertyName' ],
> });
> 
> foam.CLASS({
>   name: 'TerminateROPECmd',
>   extends: 'AbstractRopeCmd',
> });
> ```
> 
> @TW80000 is this related to what you mentioned above?

What would be the motivation for doing it this way? The only advantage I can think of is that it avoids potential naming conflicts, but the downside is that it makes it more cumbersome to work with ROPE, which is already relatively cumbersome. The type is already relatively complex:

`Map<String, Map<String, Array<String>>>`

Changing the `Array<String>` part to an object with an `Array<String>` property makes it even more complicated. 

Personally, I think properties named `__terminate__` are highly unlikely. Contrast this with creating new ROPE objects, which we need to do all the time to use ROPE. We'd be making a very rare case a slight bit easier while making the extremely common case more difficult. Therefore I think we should keep it as is and just continue to use strings.

If there are other advantages that I'm not thinking of or you disagree, please let me know.

PR #2608, opened 2019-09-17
[CPF-2854] Make foam.core.JSON more robust for missing classes

KernelDeimos:
This PR adds a check in case `( opt_ctx || foam ).lookup(cls)` reports an undefined value, making foam more robust in the case that such an error occurs.


KernelDeimos:
My latest change clarifies the warning message and reverts the change in behaviour. This new warning may have helped me determine the cause sooner.

PR #2609, opened 2019-09-17
Session logic consolidation

TW80000:
* Adds an `applyTo` method to `Session` that serves as a central way of appropriately updating a context based on a session. A number of places throughout the repository are updated to call this method to update the session context instead of doing so manually to ensure consistent behaviour across the board. This method also allows us to move some of the logic out of `SessionServerBox` and into the method, making `SessionServerBox` simpler.
* Updates the logger for sessions so that it shows both the agent and the user when a user is acting as another user. This will make things easier to audit.
* Sets `authenticate: true` on the `userDAO` service.
* Saves sessions when user id is set so it will persist across server restarts.

### DIFF
--- a/src/foam/box/SessionServerBox.java
+++ b/src/foam/box/SessionServerBox.java
@@ -9,20 +9,26 @@ package foam.box;
 import foam.core.X;
 import foam.dao.DAO;
 import foam.nanos.app.AppConfig;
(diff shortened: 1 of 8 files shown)


jlhughes (approved):
Verified with manual testing. Looks good.

PR #2610, opened 2019-09-17
[CPF-2909] Use java code for addressParser service

KernelDeimos:
@zahnno should review this first because I don't know where to test the service.

This PR fixes CPF-2909 and CPF-2854.

As an interesting side-note, this bug brought to light inconsistencies between the old development build and the foam-bin.js build. In this particular case, foam-bin.js build would try to load an object from a Context.__cache__ (using `lookup()`) before it has been registered, causing an error; the old development build also tried to do this but using a different caller (unsure which) that did not try to use the undefined object that was reported by lookup.

### DIFF
--- a/src/services
+++ b/src/services
@@ -587,9 +587,9 @@ p({
   "name":"addressParser",
   "lazy":false,
   "serve":false,
-  "service":{
-    "class":"foam.nanos.geocode.GoogleMapsAddressParser",
-  }
+  "serviceScript": """
+    new foam.nanos.geocode.GoogleMapsAddressParser.Builder().build();
+  """,
 })
 p({
   "class":"foam.nanos.boot.NSpec",
--- a/src/services
+++ b/src/services
@@ -588,7 +588,7 @@ p({
   "lazy":false,
   "serve":false,
   "serviceScript": """
-    new foam.nanos.geocode.GoogleMapsAddressParser.Builder().build();
+    return new foam.nanos.geocode.GoogleMapsAddressParser.Builder(x).build();
   """,
 })
 p({


TW80000 (approved):

src/services
```diff
@@ -587,9 +587,9 @@ p({
   "name":"addressParser",
   "lazy":false,
   "serve":false,
-  "service":{
-    "class":"foam.nanos.geocode.GoogleMapsAddressParser",
-  }
+  "serviceScript": """
+    new foam.nanos.geocode.GoogleMapsAddressParser.Builder().build();
```
zahnno: You're going to have to pass "x" into the Builder method.

zahnno: And return?? I think

TW80000: > And return?? I think

I thought so too, but looked it up and found this:

> Methods with unspecified return types may return any type of object (as in the previous example). Alternatively they may also simply issue a "return;" without a value, in which case the effective type of the method is "void" (no type). In either case, the return statement is optional. If the method does not perform an explicit "return" statement and the return type is not explicitly set to void, the value of the last statement or expression in the method body becomes the return value (and must adhere to any declared return typing). - http://www.beanshell.org/manual/methods.html

So BeanShell is funky and doesn't require an explicit return from methods. Although I don't know if a script counts as a method, so we should probably confirm that it works.

jlhughes:
```suggestion
    return new foam.nanos.geocode.GoogleMapsAddressParser.Builder(x).build();
```
This also fixed DIG - presently the dao selection list is broken.

KernelDeimos:
Updated - added return and x

jlhughes (approved):

PR #2612, opened 2019-09-18
Add missing test class

jlhughes:
CPF-2915 #time 30m

### DIFF
--- a/src/foam/nanos/nanos.js
+++ b/src/foam/nanos/nanos.js
@@ -122,6 +122,7 @@ FOAM_FILES([
   { name: "foam/nanos/test/Test" },
   { name: "foam/nanos/test/TestBorder" },
   { name: "foam/nanos/cron/Cron" },
+  { name: "foam/nanos/cron/test/IntervalScheduleTest" },
   { name: "foam/nanos/export/ExportDriverRegistry"},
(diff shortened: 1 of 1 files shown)


TW80000 (commented):
I tried to reproduce the issue and notice that I only get the console error when building using the liveScriptBundler. When I build with `./build.sh -w` then the tests table loads correctly with no errors. So we might consider the bug with the liveScriptBundler instead. That said, until we can fix the issue with the liveScriptBundler, we can merge this. Although I think `TimeOfDayScheduleTest` should be added as well.

jlhughes:
I'll get Eric to create tickets.

KernelDeimos:
I actually had IntervalScheduleTest and TimeOfDaySchedule test both added in their own tests.jrl already (src/foam/nanos/cron/tests.jrl) - do I need to add them to both places? @TW80000

TW80000:
> I actually had IntervalScheduleTest and TimeOfDaySchedule test both added in their own tests.jrl already (src/foam/nanos/cron/tests.jrl) - do I need to add them to both places? @TW80000

@KernelDeimos Yeah, you need to list both in `classes.js` if you want FOAM to generate Java code for them afaik.

PR #2614, opened 2019-09-18
[CPF-2929] Add sourceURL to foam-bin.js output

KernelDeimos:
Caveats:
- sourceURL requires code to be put in an eval() function. This adds escapes and makes the output source slightly larger, but as far as I can tell has not affected the performance of page loads or runtime.
- line numbers are inaccurate when javaCode and css are stripped out.
- full path does not display (but filename does) in Chrome if a file has more than one model. This is likely because the same sourceFile is specified multiple times and the contents of each model does not match the contents of the entire file.

### DIFF
--- a/src/foam/build/Builder.js
+++ b/src/foam/build/Builder.js
@@ -464,7 +464,7 @@ console.log("Done.");
             if ( typeof s.source !== 'undefined' ) {
               dataTmp += `\n//# sourceURL=/`+s.source;
             }
-            // NB: eval() is necessary for `//# sourceURL` to work
+            // eval() is necessary for `//# sourceURL` to work
             data += 'eval(' + JSON.stringify(dataTmp) + ');\n';
           },
           eof: function() { then(data); }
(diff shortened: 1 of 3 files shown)


src/foam/build/Builder.js
```diff
@@ -458,8 +458,14 @@ console.log("Done.");
...
 
-
-            data += `foam.${func}(${serializer.stringify(filterAxiomsByFlags(s))});\n`;
+            // Print FOAM model definition
+            dataTmp = `foam.${func}(${serializer.stringify(filterAxiomsByFlags(s))});\n`
+            // Print sourceURL annotation for browser inpector
+            if ( typeof s.source !== 'undefined' ) {
+              dataTmp += `\n//# sourceURL=/`+s.source;
+            }
+            // NB: eval() is necessary for `//# sourceURL` to work
```
kgrgreer: What does NB: mean?

KernelDeimos: "Note well" - I'll remove that from the comment so it's less confusing

PR #2617, opened 2019-09-18
Set default for invalid dates in javascript

KernelDeimos:
When an invalid date is provided for a Date property on the Javascript side, the current error handling throws an exception and breaks the UI. This PR changes the behaviour to set a default date of `258854-02-07T05:41:06.892Z` (close to the maximum Javascript can support, `MAX_SAFE_INTEGER * 0.9`). Depending on the browser's implementation this date may be different, but it should always be a date after 2038-01-19.

### DIFF
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -100,7 +100,9 @@ foam.CLASS({
         if ( typeof d === 'string' ) {
           var ret = new Date(d);
 
-          if ( isNaN(ret.getTime()) ) throw 'Invalid Date: ' + d;
+          if ( isNaN(ret.getTime()) ) {
+            ret = new Date(Number.MAX_SAFE_INTEGER || Number.MAX_VALUE);
+          }
 
           return ret;
         }
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -101,7 +101,8 @@ foam.CLASS({
           var ret = new Date(d);
 
           if ( isNaN(ret.getTime()) ) {
-            ret = new Date(Number.MAX_SAFE_INTEGER || Number.MAX_VALUE);
+            ret = new Date((Number.MAX_SAFE_INTEGER || Number.MAX_VALUE) * 0.9);
+            console.warn("Invalid date: " + d + "; assuming "+ret.toISOString()+".");
           }
 
           return ret;


jlhughes:
Can you fix the test case please: 
`Failures:
1) Date throws on invalid date strings
  Message:
    Expected function to throw an exception.
  Stack:
    Error: Expected function to throw an exception.
        at UserContext.<anonymous> (/home/travis/build/foam-framework/foam2/test/src/core/types.js:114:46)`

src/foam/core/types.js
```diff
@@ -100,7 +100,10 @@ foam.CLASS({
         if ( typeof d === 'string' ) {
           var ret = new Date(d);
 
-          if ( isNaN(ret.getTime()) ) throw 'Invalid Date: ' + d;
+          if ( isNaN(ret.getTime()) ) {
+            ret = new Date((Number.MAX_SAFE_INTEGER || Number.MAX_VALUE) * 0.9);
```
mcarcaso, line 104: What is this 0.9 for? Seems kinda arbitrary and I'm not sure it's the behaviour that would be expected. If I set the date to 'Mike', what would you expect to happen?

If the issue you're having is on the front-end, why not do some error handling in the views rendering it rather than the core where all dates are handled?

jlhughes: @mcarcaso I'm not keen on this either. it came up during our migration of old cron schedules to new.

KernelDeimos: Using MAX_SAFE_INTEGER without 0.9 caused the error to persist still. I tried subtracted 1 as well (and subtracting 9,000,000) and it didn't work. I couldn't find any documentation for the maximum value the Date constructor allows so I used `* 0.9` which worked.

I definitely agree that handling the error on a view is better. I haven't found a good way to do this yet since the exception is thrown the moment the property gets set. Maybe properties should have a sort of `isInvalid` flag so we can implement this.

KernelDeimos: Actually I just found the correct maximum value for Date; it's `8640000000000000`

mcarcaso:
Sounds to me like the server is sending bad dates to the client and it’s
struggling to parse them. What’s the root cause?

KernelDeimos:
On staging the jar was updated after Cron.Schedule was added but before NeverSchedule was removed, so an unmigrated task was but on the default of NeverSchedule and its scheduled time is now Java's Long.MAX_INT. When @mcarcaso told me this integer gets sent to the client I removed NeverSchedule. I think it would be good to add validation predicates to Date so that on the Java side it won't allow a bad date to enter in the first place.

KernelDeimos:
See #2619 for a preventative fix

jlhughes (approved):

src/foam/core/types.js
mcarcaso: Why can't you migrate the journals so they have valid dates rather than changing the behaviour for everybody? To check if a date is valid in JS you can do a `Number.isNaN(d.getTime())`.

PR #2618, opened 2019-09-19
Solid Q1 Monster Merge FOAM

tharmaman:
Includes:
1. Refactoring of TreeGraph to use the new tree graph algorithm and the creation of TreeNode rather than making it an inner class
2. Rounding of corners on the card border
3. localToGlobalCoordinates method on CView to enable proper clicking for the Tree
4. Double parser update to deal with nulls. 
5. make csvparser not return null when it can't parse.

@mcarcaso @nanoMichal @kennyqiyenkan @adamvy @kgrgreer


src/foam/core/Proxy.js
```diff
@@ -510,3 +510,78 @@ foam.CLASS({
     }
   ]
 });
+
+foam.CLASS({
+  package: 'foam.core',
+  name: 'RequestResponseCache',
```
mcarcaso: I was being lazy when I put this here. Probably deserves to be in its own file but you might want to check with @kgrgreer if this is worth submitting or not and if this is a good name for this axiom.

src/foam/dao/daoUtils.js
```diff
@@ -105,6 +105,25 @@ public ProxyDAO(foam.core.X x, foam.dao.DAO delegate) {
...
+    {
+      class: 'RequestResponseCache',
+      of: 'foam.dao.DAO',
+      name: 'cache',
+      methods: [ 'find_', 'select_' ],
+    },
+    {
+      class: 'Int',
+      name: 'ttl',
+      value: 60000
```
mcarcaso: This might be a bit aggressive of a TTL. This means you'll cache responses for find and select for a full minute. Making the TTL like 1 second might be good enough to really lighten the traffic load.

...actually you might wanna remove all this stuff from your PR completely since https://github.com/foam-framework/foam2/pull/2571/files does exactly what I just told you to do lol.

src/foam/graphics/TreeGraph.js
```diff
@@ -33,306 +36,76 @@
...
-            else {
-              this.graph.updateCWidth();
-            }
-          
-            this.graph.invalidate();
+          if ( this.root && this.root.layout() ) {
+            this.invalidate();
+            this.doLayout();
+          } else {
+            // console.log(Date.now() - this.time)
```
mcarcaso: Remove

src/foam/graphics/TreeGraph.js
```diff
@@ -33,306 +36,76 @@
      {
        name: 'formatNode',
        value: function() {}
-     }
+     },
+     'time'
```
mcarcaso: You still need this?

src/foam/graphics/TreeGraph.js
```diff
@@ -33,306 +36,76 @@
...
+   ],
+
+   topics: [
+    'layoutComplete'
```
mcarcaso: I was gonna say you should probably rename this to something like `onSizeChange` since that's where it's actually getting published but now that I look closer, you can probably get rid of this topic all together and just subscribe to the width and/or height if you care to know when the size changes.

src/foam/graphics/TreeGraph.js
```diff
@@ -33,306 +36,76 @@
...
    ],
 
    methods: [
      function initCView() {
        this.SUPER();
 
+       this.color = 'white';
```
mcarcaso: I'd make this configurable. In properties, you can add a property that's `['color', 'white']` which gives a default value to the color property and this could be overwritten if a user cares to do so.

src/foam/graphics/TreeGraph.js
```diff
@@ -33,306 +36,76 @@
...
-           var delta = Math.abs(slot.get() - newValue);
-           if ( delta < 0.001 ) {
-             slot.set(newValue);
-             return false;
-           }
-           slot.set((2*slot.get() + newValue)/3);
-           return true;
+       name: 'doLayout',
+       isMerged: true,
+       mergeDelay: 1,
```
mcarcaso: Looks like you can remove this. The default mergeDelay for a listener is 16ms which should be good enough for your purposes I think. Try it out.

src/foam/graphics/TreeGraph.js
```diff
@@ -33,306 +36,76 @@
...
-        var delta = Math.abs(this.width - width) / width;
+         var deltaX = Math.abs(this.width - width) / width;
+         var deltaY = Math.abs(this.height - height) / height;
 
-        if ( delta > 0.01 ) {
-          this.width = width;
-          this.doLayout();
-        }
-       }
+         if ( deltaX > 0.01 || deltaY > 0.01 ) {
```
mcarcaso: How important is checking the delta here? Why can't you just check if `this.width != width || this.height != height`? If you need to do this delta check, please explain in a comment why.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+    'relationship'
+  ],
+  exports: ['as parentNode'],
+
+  properties: [
+    'data',
+    {
+      name: 'outline',
+      expression: function (x, nodeWidth, expanded, childNodes, padding) {
+        var nodeLeftEdgePadded = x - nodeWidth / 2 - padding / 2;
```
mcarcaso: why have these variables declared here? could you just create the rootLevelOutline in place?
```
        var rootLevelOutline = [
          {
            left: x - nodeWidth / 2 - padding / 2,
            right: x + nodeWidth / 2 + padding / 2
          }
        ];
```

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+        var nodeRightEdgePadded = x + nodeWidth / 2 + padding / 2;
+        var rootLevelOutline = [
```
mcarcaso: Looks like you don't use this until the very end. Can you construct it on the fly when you actually need it rather than declaring it at the top? Seeing it up here makes me expect to see you using it throughout the method.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+          {
+            left: nodeLeftEdgePadded,
+            right: nodeRightEdgePadded
+          }
+        ];
+
+        var champion = [];
```
mcarcaso: This isn't a very descriptive name for somebody looking at this code for the first time. It's more like a `outlinePerLevel` or `levelEdges` or something. Or maybe even just `returnValue` or `outline`.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+    'padding',
+    'parentNode?',
+    'relationship'
+  ],
+  exports: ['as parentNode'],
+
+  properties: [
+    'data',
+    {
+      name: 'outline',
```
mcarcaso, line 36: This variable could use some documentation describing that it's an array containing the max left and right values per level of the tree.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+          {
+            left: nodeLeftEdgePadded,
+            right: nodeRightEdgePadded
+          }
+        ];
+
+        var champion = [];
+
+        for (let i = 0; i < childNodes.length && expanded; i++) {
+          // get child outline
```
mcarcaso: These comments are more "what" than "why". Comments that describe "what" is happening are often not useful and incur maintenance burden because they'll need to be updated when code is updated (which is often forgotten!) so I strongly recommend against doing this unless there's a good reason to do so. Explaining "why" the code is doing what it's doing is more useful because it provides context for why it was written in the first place.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+            champion[j] = champion[j] || {
+              left: Number.MAX_SAFE_INTEGER,
+              right: Number.MIN_SAFE_INTEGER
+            };
+            champion[j].left = Math.min(childOutline[j].left, champion[j].left);
+            champion[j].right = Math.max(childOutline[j].right, champion[j].right);
+          }
+        }
+
+        var totalOutline = rootLevelOutline.concat(champion);
```
mcarcaso: Probably don't need this variable. I'd probably have just renamed `champion` to `outline` and included the root in it so you simply do a `return outline;` as the very last statement.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+    'foam.graphics.Box',
+    'foam.graphics.CView',
+    'foam.graphics.Label',
+    'foam.graphics.Line'
+  ],
+
+  imports: [
+    'formatNode',
+    'graph',
+    'nodeHeight',
```
mcarcaso: Can you remove the nodeHeight and nodeWidth in favor of simply setting the height and width of the node when you're creating it?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+    { name: 'height', factory: function () { return this.nodeHeight; } },
+    { name: 'width', factory: function () { return this.nodeWidth; } },
+    ['border', 'gray'],
+    {
+      name: 'childNodes',
+      factory: function () { return []; }
+    },
+    ['left', 0],
+    ['right', 0],
+    ['maxLeft', 0],
```
mcarcaso: do you need any of these left/right/maxLeft/maxRight values?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+    ['maxRight', 0],
+    {
+      name: 'x',
+      expression: function(neighborX, centerX) {
+        return neighborX + centerX;
+      },
+      postSet: function() {
+        this.neighborX = 0;
```
mcarcaso: Why is this needed? Who is messing with the `x` value and why are you throwing that away?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+        this.centerX = 0;
+        this.x = undefined;
+      }
+    },
+    {
+      class: 'Float',
+      name: 'neighborX'
```
mcarcaso: What's the neighborX for? Is it avoidable? If not, can you please add documentation explaining what this value is for?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+    },
+    {
+      class: 'Float',
+      name: 'centerX'
```
mcarcaso: What's the centerX for? Is it avoidable? If not, can you please add documentation explaining what this value is for?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+      this.formatNode();
+
+      if (this.relationship) {
+        var data = this.data.clone(this.__subContext__);
+        try {
+          data[this.relationship.forwardName].select(data => {
+            this.addChildNode({ data: data });
+          });
+        } catch (x) { }
+        this.graph.doLayout();
```
mcarcaso: Are you sure you want to trigger a layout here or after the select is complete?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+  ],
+
+  methods: [
+    function initCView() {
+      this.SUPER();
+
+      this.formatNode();
+
+      if (this.relationship) {
+        var data = this.data.clone(this.__subContext__);
```
mcarcaso: Why do you need to clone? You're not manipulating the data so it's not clear why you'd want to clone it.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+        try {
```
mcarcaso: Do you need this try/catch?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+          if ( ! p.expanded ) return;
+          p = p.parent;
+        }
+
+        var point = this.globalToLocalCoordinates({
+          w: 1,
+          x: e.offsetX,
+          y: e.offsetY,
+        });
+        point.x += this.width / 2;
```
mcarcaso, line 152: Why do you need to do this? Might want to explain that the x value is actually the middle of the node and not the far left

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+
+        var hit = this.hitTest(point)
```
mcarcaso: Do you need to store this in a variable or can you use it directly in the if statement?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+        if ( hit ) {
+          this.expanded = ! this.expanded;
+        }
+      });
+    },
+
+    function paint(x) {
+      if (!this.parentNode || this.parentNode.expanded) this.SUPER(x);
```
mcarcaso: foam2 formatting should be :
`if ( ! this.parentNode || this.parentNode.expanded ) this.SUPER(x);`

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+    },
+
+    function paintSelf(x) {
+      x.save();
+
+      x.translate(-this.width / 2, 0);
+      this.SUPER(x);
+
+      // reset translate and shadow settings
```
mcarcaso: I'm not sure this comment provides any value.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+
+    function paintConnectors(x) {
+      function line(x1, y1, x2, y2) {
+        x.beginPath();
+        x.moveTo(x1, y1);
+        x.lineTo(x2, y2);
+        x.stroke();
+      }
+
+      x.lineWidth = 0.5; //this.borderWidth;
```
mcarcaso: What's with the commented out code?

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+      }
+    },
+
+    function addChildNode(args) {
+      var node = this.cls_.create(args, this);
+      node.y = this.height * 2;
+      this.add(node);
+      this.childNodes = this.childNodes.concat(node);
+      node.outline$.sub(() => {
+        this.outline = [];
```
mcarcaso, line 220: Explain why you need to do this. It's to force a recalculation of the outline because there's no way to subscribe to all children outlines via an expression.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+        this.outline = undefined
+      });
+      return this;
+    },
+
+    function distanceTo(node) {
+      var outlineA = this.outline;
```
mcarcaso: I'm not sure the value of declaring these up here. You can refer to this.outline and node.outline inline and it'll be a little clearer what's what.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+      var outlineB = node.outline;
+      minLevels = Math.min(outlineA.length, outlineB.length);
+
+      var champion = Number.MAX_SAFE_INTEGER;
```
mcarcaso: I'd rename this to `distance` or something a bit more descriptive.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+      for (var i = 0; i < minLevels; i++) {
+        // we only need to check if the right of outlineA overlaps with the left of outlineB
```
mcarcaso: I don't think you need this comment.

src/foam/graphics/TreeNode.js
```diff
@@ -0,0 +1,297 @@
...
+        var overlapDistance = outlineB[i].left - outlineA[i].right;
+
+        champion = Math.min(champion, overlapDistance);
+      }
+      return champion;
+    },
+
+    function layout() {
+      const { childNodes } = this;
```
mcarcaso: Why do you need this? Could you use this.childNodes everywhere instead?

src/foam/java/refinements.js
```diff
@@ -248,7 +248,10 @@ foam.CLASS({
         foam.core.PropertyInfo prop = this;
         map.put(getName(), new foam.lib.csv.FromCSVSetter() {
           public void set(foam.core.FObject obj, String str) {
+          try{
```
mcarcaso: Explain why this is needed and clean up the formatting.

mcarcaso:
What is "Liquid"? ;)
Might want to keep these code names a little more secret.

src/foam/graphics/TreeGraph.js
tharmaman: Nope, removed

src/foam/graphics/TreeGraph.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
tharmaman: Renamed to onSizeChange

src/foam/java/refinements.js
tharmaman: Done.

tharmaman:
> What is "Liquid"? ;)
> Might want to keep these code names a little more secret.

I just git inverted the name

src/foam/graphics/TreeNode.js
tharmaman: I just thought would make the could look prettier. Changed it all to this.childNodes then.

src/foam/graphics/TreeNode.js
tharmaman: Done. Renamed to minDistance.

src/foam/graphics/TreeNode.js
tharmaman: Done.

src/foam/graphics/TreeNode.js
tharmaman: Done.

src/foam/graphics/TreeNode.js
tharmaman: I think that comment is there to say that the lineWidth is equal to the borderWidth value of 0.5, but might as well set it to this.borderWidth instead or add a lineWidth property.

src/foam/graphics/TreeNode.js
tharmaman: Done.

src/foam/graphics/TreeNode.js
tharmaman: Done.

src/foam/graphics/TreeNode.js
tharmaman: Done.

src/foam/graphics/TreeGraph.js
tharmaman: Removed the mergeDelay

src/foam/graphics/TreeNode.js
tharmaman: Nope removed.

src/foam/dao/daoUtils.js
tharmaman: Cool just checked out src/foam/dao/daoUtils.js, src/foam/dao/EasyDAO.js and src/foam/core/Proxy.js from foam Master

src/foam/core/Proxy.js
tharmaman: See right below, just checked out the three files related to your own PR.

src/foam/graphics/TreeGraph.js
tharmaman: Think this was from the legacy code where we wanted to reduce the amount this.doLayout() was called to have the graph resize faster by having a delta instead of converging on the exact value. Changed it to `this.width != width || this.height` since the new graph resizes super fast anyways

src/foam/graphics/TreeNode.js
tharmaman: Added comment

src/foam/graphics/TreeNode.js
tharmaman: Removed the clone.

src/foam/graphics/TreeNode.js
tharmaman: Removed tryCatch block

src/foam/graphics/TreeNode.js
tharmaman: Removed the doLayout here as it doesn't make a difference.

src/foam/graphics/TreeNode.js
tharmaman: Done, also did the same for padding.

src/foam/graphics/TreeNode.js
tharmaman: Added.

src/foam/graphics/TreeNode.js
tharmaman: Moved it to the bottom right before we concat it, still kept the declaration of rootLevelOutline though so its clear to the reader what it is.

src/foam/graphics/TreeNode.js
tharmaman: Renamed to outlineBelowRoot as not to confuse with the childOutline variable.

src/foam/graphics/TreeNode.js
tharmaman: Yeah its not used anywhere else, moved it directly in the if statement

src/foam/graphics/TreeNode.js
tharmaman: Made a better comment

src/foam/graphics/TreeNode.js
tharmaman: Added comment

src/foam/graphics/TreeNode.js
tharmaman: removed it and just did `return rootLevelOutline.concat(outlineBelowRoot);` as I think its a little bit more clearer to what's happening

src/foam/graphics/TreeNode.js
tharmaman: Not sure lol, removing the whole postSet here works.

src/foam/graphics/TreeNode.js
tharmaman: K im just gunna create a property called lineWidth instead borderWidth doesn't even exist in treeNode

tharmaman: Yeah so when the lines are equal to borderWidth the line shows up a bit thicker, i'll expose a property called lineWidth to TreeNode and set that to 0.5. I noticed afterwards lineWidth gets reset to borderWidth in order to draw the collapse/expand arrows, ill leave that to be set to borderWidth.

jlhughes (approved):

src/foam/lib/parse/CSVParser.java
```diff
@@ -66,7 +66,7 @@ public PStream parse(PStream ps, ParserContext px) {
             }
             sink.put(obj, detach);
           } catch ( Exception e ) {
-            return null;
+           // return null;
```
mcarcaso: Don't just comment this out. Leave a comment as to why you want to do nothing when a value fails to parse and remove the `return null`

src/foam/graphics/TreeGraph.js
```diff
@@ -36,303 +38,68 @@
      }
    ],
 
+   topics: [
+    'onSizeChange'
```
mcarcaso, line 42: Sorry maybe my previous comment about this wasn't too clear. I'm pretty sure you can remove this topic completely and subscribe to the treegraph's width$ instead.

src/foam/lib/parse/CSVParser.java
tharmaman: Added a comment.

src/foam/graphics/TreeGraph.js
tharmaman: Replaced both subs by making a sub to width$ and the search feature doesn't seem to work anymore:
Screen Shot 2019-09-27 at 3 33 28 PM

src/foam/lib/parse/CSVParser.java
tharmaman: Done.

PR #2619, opened 2019-09-20
[CPF-2948] Add validation predicates to foam.core.Date

KernelDeimos:
FOAM relies on the `Date` object in Javascript as the value for properties of type `foam.core.Date`. According to the EMCAScript spec and my own testing, Date supports a range from `-8640000000000000` to `+8640000000000000`. If an input value exceeds this range, an error will be thrown, causing the UI to break.

Currently it is possible to store dates into journals that cannot be handled by the Javascript Date object. This PR adds validation predicates to `foam.core.Date`, making Javascript's limitation the lower and upper bounds.

Addition: The timestamp literals above correspond to the date range of `-271821-04-20` to `+275760-09-13` with zeroed time values.

### DIFF
--- a/src/foam/core/CoreTypesValidationTest.js
+++ b/src/foam/core/CoreTypesValidationTest.js
@@ -1,0 +1,56 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.core',
+  name: 'CoreTypesValidationTestModel',
+  properties: [
+    { name: 'id', class: 'String' },
+    {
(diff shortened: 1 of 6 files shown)


jlhughes (approved):

src/foam/core/CoreTypesValidationTest.js
```diff
@@ -0,0 +1,100 @@
...
+  properties: [
+    { name: 'id', class: 'String' },
+    {
+      name: 'testDate', class: 'DateTime'
+    }
+  ]
+});
+
+foam.CLASS({
+  package: 'foam.core',
```
kgrgreer, line 19: Tests are currently under /foam2/test/src/core not in the core itself. I would prefer to keep tests out of the main packages and in a parallel structure elsewhere.

PR #2620, opened 2019-09-20
Fix Java large negative numbers

KernelDeimos:


KernelDeimos:
Note for reviewers: are float/double values being sent here as well? If so there may be a second bug here, since doubles should have `d` appended, and values like `8589934592.01234L` will cause a compiler error

src/foam/java/refinements.js
```diff
@@ -35,7 +35,8 @@ foam.LIB({
           return b ? "true" : "false";
         },
         Number: function(n) {
-          return '' + n + (n > Math.pow(2, 31) ? 'L' : '');
+          return '' + n +
+            (n > Math.pow(2, 31) || n < -Math.pow(2,31) ? 'L' : '');
```
mcarcaso, line 39: You can use `Math.abs` to make this a bit shorter
`return '' + n + (Math.abs(n) > Math.pow(2, 31) ? 'L' : '');`

mcarcaso (approved):

PR #2622, opened 2019-09-23
DO NOT MERGE: Prevent session fixation attacks

TW80000:
## Behaviour changes
* Prevent session fixation attacks by changing the session token when a user authenticates
  * Since we change the session token used to access the server on the server side, we need to inform the client that it should start using the new token in future requests. I've implemented this behaviour in `SessionClientBox` and `SessionServerBox`.

## Refactoring
* Remove `AuthServiceClientBox`
  * From what I can tell, `AuthServiceClientBox` was created to solve the problem of error messages not showing up in the GUI when a user fails to sign in. This was happening because `SessionClientBox` was catching all `AuthenticationException`s being thrown by the server and showing the login screen when that happens. This was consuming the error messages that were meant to be displayed to the user, such as "User not found" or "Invalid credentials". Rather than introducing a new box that's very similar in nature to the existing one, a flag could have been added to `SessionClientBox` to signify whether the user should be prompted to login on an `AuthenticationException` or not. This is the approach I took in this PR, removing the need for `AuthServiceClientBox`, which simplifies things since there are fewer places a developer needs to look at to understand how sessions are handled on the client.
* Moved `SessionReplyBox` into its own file
* Formatting changes

Relevant ticket for nanopay developers: https://nanopay.atlassian.net/browse/CPF-2877


src/foam/nanos/session/Session.js
```diff
@@ -110,6 +110,23 @@ foam.CLASS({
       javaFactory: 'return applyTo(getX());',
       hidden: true,
       transient: true
+    },
+    {
+      class: 'String', // TODO: Create a GUID property type.
+      name: 'accessToken',
```
jlhughes: Can/should we use this as the bearer token as well?

TW80000: Definitely, I'll include that in this PR.

kgrgreer:
Did we agree not to do this?

TW80000:
> Did we agree not to do this?

We discussed it, but I'm still not convinced that we shouldn't be doing this. I think it's worthwhile to make a particular kind of attack impossible even though we have other measures in place to help prevent session fixation attacks anyway. I agree that theoretically shouldn't matter if we change the access token on login if we don't put session tokens in URLs, don't use insecure cookies, etc. However, there's always the possibility that someone will start putting a session id in a URL again, or someone will start using cookies in an insecure way, or a session will be leaked some other way. By changing the access token every time a user authenticates, we provide defence in depth and make session fixation attacks impossible by stopping the possibility at the root.

If there are _n_ known attack vectors for an attacker to pre-select a known session id for a victim, then I agree we should try to prevent all of them. But we should also just close up the root of the problem anyway so if we miss any of the attack vectors or new ones show up, they won't be exploitable anyway.

src/foam/box/SessionServerBox.java
```diff
@@ -123,6 +128,24 @@ public void send(Message msg) {
         return;
       }
 
+      // Decorate the reply box to add the sessionId as an attribute to the
```
TW80000, line 131: @KernelDeimos since we talked about you doing something similar, here's an example of adding a replyBox decorator on the server side.

PR #2623, opened 2019-09-23
Composite ruleAction

nanoMichal:
Allows the running of multiple rule actions from one rule.

### DIFF
--- a/src/foam/nanos/ruler/CompositeRuleAction.js
+++ b/src/foam/nanos/ruler/CompositeRuleAction.js
@@ -1,18 +1,7 @@
 /**
  * @license
- * Copyright 2019 Google Inc. All Rights Reserved.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
  */
 
 foam.CLASS({
--- a/src/foam/nanos/ruler/CompositeRuleAction.js
+++ b/src/foam/nanos/ruler/CompositeRuleAction.js
@@ -19,7 +19,7 @@ foam.CLASS({
   package: 'foam.nanos.ruler',
   name: 'CompositeRuleAction',
 
-  documentation: 'Runs multiple rule actions from one action.',
+  documentation: 'Runs multiple rule actions from one action. Completes actions sequentially in array, and actions stack.',
 
   implements: ['foam.nanos.ruler.RuleAction'],
 
@@ -39,9 +39,11 @@ foam.CLASS({
     {
       name: 'applyAction',
       javaCode: `
-        for ( RuleAction action : getRuleActions() ) {
-          action.applyAction(x,obj,oldObj,ruler,agency);
-        }
+        if ( getRuleActions() != null )
+          for ( RuleAction action : getRuleActions() ) {
+            if ( action != null )
+              action.applyAction(x,obj,oldObj,ruler,agency);
+          }
       `
     }
   ]
--- a/src/foam/nanos/ruler/CompositeRuleAction.js
+++ b/src/foam/nanos/ruler/CompositeRuleAction.js
@@ -39,11 +39,9 @@ foam.CLASS({
     {
       name: 'applyAction',
       javaCode: `
-        if ( getRuleActions() != null )
-          for ( RuleAction action : getRuleActions() ) {
-            if ( action != null )
-              action.applyAction(x,obj,oldObj,ruler,agency);
-          }
+        for ( RuleAction action : getRuleActions() ) {
+           action.applyAction(x,obj,oldObj,ruler,agency);
+        }
       `
     }
   ]
--- a/src/foam/nanos/ruler/CompositeRuleAction.js
+++ b/src/foam/nanos/ruler/CompositeRuleAction.js
@@ -40,7 +40,7 @@ foam.CLASS({
       name: 'applyAction',
       javaCode: `
         for ( RuleAction action : getRuleActions() ) {
-           action.applyAction(x,obj,oldObj,ruler,agency);
+          action.applyAction(x, obj, oldObj, ruler, agency);
         }
       `
     }
--- a/src/foam/nanos/ruler/CompositeRuleAction.js
+++ b/src/foam/nanos/ruler/CompositeRuleAction.js
@@ -1,8 +1,25 @@
+/**
+ * @license
+ * Copyright 2019 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


src/foam/nanos/ruler/CompositeRuleAction.js
```diff
@@ -0,0 +1,31 @@
...
+      of: 'foam.nanos.ruler.RuleAction',
+      name: 'ruleActions'
+    }
+  ],
+
+  methods: [
+    {
+      name: 'applyAction',
+      javaCode: `
+        for( RuleAction action : getRuleActions() ) {
```
TW80000: Shouldn't this be wrapped in something like the following?

```Java
agency.submit(x, new ContextAgent() {
  @Override
  public void execute(X x) {
    // Your code here
  }
}, "Executing multiple rule actions.");
```

TW80000: Missing space after `for`.

src/foam/nanos/ruler/CompositeRuleAction.js
```diff
@@ -0,0 +1,31 @@
+foam.CLASS({
+  package: 'foam.nanos.ruler',
+  name: 'CompositeRuleAction',
+
+  documentation: 'runs multiple rule actions from one action',
```
TW80000: Capitalization and punctuation

src/foam/nanos/ruler/CompositeRuleAction.js
```diff
@@ -0,0 +1,31 @@
+foam.CLASS({
```
TW80000, line 7: Missing license header

lchanmann:
Please add tests.

nanoMichal:
@googlebot I signed it!

src/foam/nanos/ruler/CompositeRuleAction.js
nanoMichal: Wouldn't each rule action use its own context agent if it requires one?

lchanmann: I guess agency.submit would be in the child action itself.

TW80000: 👍

src/foam/nanos/ruler/test/RulerDAOTest.java
```diff
@@ -87,6 +92,46 @@ public void testRuleHistory(X x) {
...
+    actions[0] = (x12, obj, oldObj, ruler, agent) -> {
+      User user = (User) obj;
+      user.setEmail("action1"+user.getEmail());
+    };
+    compositeAction.setRuleActions(actions);
+    rule10.setAction(compositeAction);
+    ruleDAO.put(rule10);
+    user1.setEmail("nanos@nanos.net");
+    user1 = (User) userDAO.put_(x, user1).fclone();
+    test(SafetyUtil.equals("action1nanos@nanos.net", user1.getEmail()),"one rule action changed user email as expected: "+ user1.getEmail());
```
lchanmann: user1.getEmail().equals("action1nanos@nanos.net") instead of SafetyUtil.equals?

src/foam/nanos/ruler/test/RulerDAOTest.java
```diff
@@ -87,6 +92,46 @@ public void testRuleHistory(X x) {
...
+    actions[1] = (x12, obj, oldObj, ruler, agent) -> {
+      User user = (User) obj;
+      user.setEmail("action2"+user.getEmail());
+    };
+    compositeAction.setRuleActions(actions);
+    rule10.setAction(compositeAction);
+    ruleDAO.put(rule10);
+    user1.setEmail("nanos@nanos.net");
+    user1 = (User) userDAO.put_(x, user1).fclone();
+    test(SafetyUtil.equals("action2action1nanos@nanos.net", user1.getEmail()),"Both rule actions changed user email as expected: "+user1.getEmail() );
```
lchanmann: same as above

src/foam/nanos/ruler/test/RulerDAOTest.java
```diff
@@ -34,6 +37,8 @@ public void runTest(X x) {
     testRuleHistory(x);
     testUpdatedRule(x);
     removeData(x);
+    testCompositeRuleAction(x);
+    removeData(x);
```
lchanmann, line 38: It seems like removeData doesn't remove rule10?

src/foam/nanos/ruler/test/RulerDAOTest.java
```diff
@@ -87,6 +92,46 @@ public void testRuleHistory(X x) {
...
+    //test null array of rule actions
+    CompositeRuleAction compositeAction = new CompositeRuleAction();
+    Predicate pred10 = EQ(DOT(NEW_OBJ, foam.nanos.auth.User.EMAIL), "nanos@nanos.net");
+    rule10.setPredicate(pred10);
+    rule10.setOperation(Operations.CREATE_OR_UPDATE);
+    rule10.setAction(compositeAction);
+    ruleDAO.put(rule10);
+    user1.setEmail("nanos@nanos.net");
+    user1 = (User) userDAO.put_(x, user1).fclone();
+    test(SafetyUtil.equals("nanos@nanos.net", user1.getEmail()),"no rule action resulted in no operation");
```
lchanmann: Please use "string".equals instead.

src/foam/nanos/ruler/CompositeRuleAction.js
```diff
@@ -0,0 +1,50 @@
...
+      of: 'foam.nanos.ruler.RuleAction',
+      name: 'ruleActions'
+    }
+  ],
+
+  methods: [
+    {
+      name: 'applyAction',
+      javaCode: `
+        if ( getRuleActions() != null )
```
mcarcaso: Why can't this simply be:
```
for ( RuleAction action : getRuleActions() ) {
  action.applyAction(x,obj,oldObj,ruler,agency);
}
```
FObjectArrays get a default empty array so you shouldn't need to be checking if that's null and I don't know why you'd want to handle an array with nulls in it.

nanoMichal: The user will need to ensure no null rule actions. this makes sense. thanks @mcarcaso

src/foam/nanos/ruler/CompositeRuleAction.js
```diff
@@ -0,0 +1,48 @@
...
+      name: 'ruleActions'
+    }
+  ],
+
+  methods: [
+    {
+      name: 'applyAction',
+      javaCode: `
+        for ( RuleAction action : getRuleActions() ) {
+           action.applyAction(x,obj,oldObj,ruler,agency);
```
mcarcaso: Indentation is 3 spaces from the for loop and not 2. Commas between args

src/foam/nanos/ruler/test/RulerDAOTest.java
TW80000: I think `SafetyUtil.equals` is safer because it will not throw an NPE if `user1.getEmail()` returns null.

mcarcaso: You can also avoid NPEs by reversing the order:
`"action1nanos@nanos.net".equals(user1.getEmail());`
I'm indifferent on the approach since it's in a test so an NPE would signal that something is broken anyway.

Do we define the `test` method somewhere? I wonder if it would be worth adding a `testEquals(Object, Object, String)` that compares the first two parameters and the third param is the error message. Not for this PR but could be useful to think about.

src/foam/nanos/ruler/CompositeRuleAction.js
```diff
@@ -0,0 +1,48 @@
+/**
```
kgrgreer, line 1: This is the correct license header, since you don't work for Google.

/**
 * @license
 */

PR #2625, opened 2019-09-24
Update session object-level access control logic and add SessionService

TW80000:
This commit makes it so that a user must pass one of three checks in
order to operate on a session. The checks are as follows:

  1. The user is the "session user", meaning the user's id matches the
     value of the userId property on the session instance.
  2. The user has a permission of the form "session.<operation>.<spid>"
     where operation is "create", "read", "update", or "delete". This
     permission allows them to operate on all sessions where the user
     in the session has a SPID that matches the SPID in the permission.
  3. The user has a permission of the form "session.<operation>.<id>"
     where operation is the same as above and id is the id of a session,
     which is a GUID.

The first rule allows users to modify their own session in so far as the
property-level access control checks allow them to. The second rule
allows "weak" admins - which are administrators of service providers - to
manage the sessions of users under their service. The third check allows
"super" admins to be able to do anything. Note that the second check
won't always pass for super admins because it will fail if the userId of
the session is not set.

### DIFF
--- a/src/foam/nanos/session/Session.js
+++ b/src/foam/nanos/session/Session.js
@@ -14,6 +14,11 @@ foam.CLASS({
     'foam.nanos.auth.CreatedByAware'
   ],
 
+  imports: [
+    'auth',
+    'localUserDAO'
+  ],
+
   javaImports: [
     'foam.core.X',
     'foam.dao.DAO',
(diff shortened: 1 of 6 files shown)


kgrgreer:
I would rather not see sessionDAO made available to users. I would prefer an authToken service with specific methods for desired use-cases, like:
SessionId createAuthToken(userid, ttl);

jlhughes (approved):
So does this still give us (an admin) to manage a long-term-session for client/customer?

TW80000:
@jlhughes @kgrgreer Updated. Now there's a `SessionService` interface with methods to create and destroy sessions. It includes an access control check that only lets one create or destroy sessions for users in a SPID if they have the `session.create.<spid>` or `session.delete.<spid>` permissions, respectively.

`sessionDAO` is still served but the authorization methods on `Session` were changed to only check for `session.create.*`, `session.read.*`, `session.update.*`, and `session.delete.*` permissions, which only admin will have.

TW80000:
Example requests:

* Create a session
  * `http://localhost:8080/service/sugar?service=sessionService&method=createSession&interfaceName=foam.nanos.session.SessionService&userId=1348`
  * Returns the created session's id (which should be used as the bearer token)
* Create a session with a custom TTL
  * `http://localhost:8080/service/sugar?service=sessionService&method=createSessionWithTTL&interfaceName=foam.nanos.session.SessionService&userId=1348&ttl=9223372036854775807`
  * Returns the created session's id (which should be used as the bearer token)

jlhughes (approved):

src/foam/nanos/session/Session.js
```diff
@@ -14,6 +14,11 @@ foam.CLASS({
     'foam.nanos.auth.CreatedByAware'
   ],
 
+  imports: [
```
kgrgreer: I guess it doesn't hurt, but are these checks still needed?

src/foam/nanos/session/SimpleSessionService.js
```diff
@@ -0,0 +1,110 @@
...
+
+        // TODO: Change to query by access token property instead of id when we
+        // support that.
+        Session session = (Session) localSessionDAO.find(accessToken);
+
+        if ( session == null ) {
+          throw new RuntimeException("Session not found.");
+        }
+
+        if (
```
kgrgreer: Is this check actually preventing anything? If I know the sessionID, I could just use it to logout the user anyway (or do anything else on behalf of the user).

src/foam/nanos/session/Session.js
TW80000: I'm not sure what you mean by "checks" here since you commented on the imports. Can you clarify?

If you mean the `authorizeOnXYZ` checks below, then yes, we do still need them because `AuthorizationDAO` calls the `authorizeOnXYZ` methods and we need a condition that will be true for superadmins and false for everyone else.

src/foam/nanos/session/SimpleSessionService.js
TW80000: Hmm, good point. I think you're right.

What if this method took a user id instead and deleted all sessions for that user? Then the check would make sense again since guessing or obtaining another user's id is relatively easy since they're sequence numbers and we don't make any effort to hide user ids, whereas guessing someone's session id (a GUID) is (intended to be) impossible.

src/foam/nanos/session/Session.js
kgrgreer: Yes, that's what I meant. But only superadmins would even have access to the DAO now that we don't need to make the DAO directly available now that we have the service. It should only be superadmin and system accessing the DAO.

TW80000: Are you suggesting that we only need service-level permission checks then? Like `service.sessionDAO` and only give that to superadmins? We still want to serve it because we'd like to be able to manage sessions through the GUI.

src/foam/nanos/session/SimpleSessionService.js
kgrgreer: I think we would only want to let you delete sessions you created. If the user logs into the system themself do we want to let you kick them off? If you want to disable them, and you have the access, then you could disable the User and the system could then decide if it wants to remove their active sessions or the Auth system would just stop let them doing things anyway.

TW80000: Fair enough. I'll just remove this method then.

TW80000: Resolved by e8a9568.

src/foam/nanos/session/Session.js
TW80000: Resolved by 25d353a.

PR #2626, opened 2019-09-24
Delete old build files

hchoura:


src/gen.sh
```diff
@@ -2,5 +2,9 @@
 #Concatenate files into one services file
 SRCDIR=$(dirname $0)
 cd "$SRCDIR"
+
+# Delete old build files
+rm -r ../build/
```
jlhughes, line 7: does 'make clean' not do this? or you want to run it each time?

hchoura: No. 
make clean do:
rm -f build/foam2.jar
rm -f build/foam2.classpath.txt
rm -rf .foam2-gensrcs

jlhughes: so why not add this to make clean?

jlhughes: with the clean change above, this is no longer needed correct?

jlhughes:
Superceded by https://github.com/foam-framework/foam2/pull/2639

PR #2627, opened 2019-09-24
Log errors with logger rather than standard out.

jlhughes:
Also, trap the NoSuchFieldException and log an error.

In development, scripts fail if email templates are not configured.

### DIFF
--- a/src/foam/nanos/notification/email/EmailTemplateApplyEmailPropertyService.js
+++ b/src/foam/nanos/notification/email/EmailTemplateApplyEmailPropertyService.js
@@ -37,7 +37,7 @@ foam.CLASS({
       try {
         emailMessage = emailTemplateObj.apply(x, group, emailMessage, templateArgs);
       } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


src/foam/nanos/notification/email/EmailTemplateApplyEmailPropertyService.js
```diff
@@ -37,7 +37,7 @@ foam.CLASS({
       try {
         emailMessage = emailTemplateObj.apply(x, group, emailMessage, templateArgs);
       } catch (Exception e) {
-        throw new NoSuchFieldException("@EmailTemplateApplyEmailPropertyService: emailTemplate.apply has failed. emailTemplate = {id:" + templateName + ", group:" + group + "}" + e);
+        logger.error(new NoSuchFieldException("@EmailTemplateApplyEmailPropertyService: emailTemplate.apply has failed. emailTemplate = {id:" + templateName + ", group:" + group + "}" + e));
```
TW80000: As a tiny optimization, we could move the logger declaration here inside the catch block. That way we don't get it out of the context unless we'll need it.

TW80000 (approved):
LGTM

PR #2628, opened 2019-09-25
Fix bug where editEnabled: false would still allow editing

TW80000:
The bug was happening because the `editEnabled` property wasn't being passed to `DAOUpdateControllerView`.

Ticket: https://nanopay.atlassian.net/browse/CPF-2995


jlhughes (approved):
How is Create controlled?

TW80000:
> How is Create controlled?

Through the `createEnabled` flag.

PR #2629, opened 2019-09-25
[CPF-824] Create AbstractFileJournal and make FileJournal a subclass

KernelDeimos:
This PR makes FileJournal a subclass of AbstractFileJournal. It also moves the `implements` for `foam.dao.Journal` up to FileJournal.

**After this**
A PR after this will rename RoutingJournal to RoutingFileJournal and make it a subclass of AbstractFileJournal which implements a new `foam.dao.RoutingJournal` interface.

### DIFF
--- a/src/foam/dao/AbstractFileJournal.js
+++ b/src/foam/dao/AbstractFileJournal.js
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2018 The FOAM Authors. All Rights Reserved.
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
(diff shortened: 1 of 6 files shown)


KernelDeimos:
I'm going to add a quick comment here on why this change is needed.

Currently FileJournal extends AbstractJournal (implements Journal), which meant the only way RoutingJournal could be given a service name is by adding it to its context. @adamvy suggested a better approach would be to change RoutingJournal's interface to accept service name for `put` and `remove`, then create an adapter (RoutingJournal->Journal) that holds the service name.

In order to change the interface of RoutingJournal it can no longer extend FileJournal (because FileJournal implements Journal), but it can extend this new AbstractFileJournal and implement its own interface.

src/foam/dao/AbstractFileJournal.js
```diff
@@ -0,0 +1,435 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
TW80000: 2019

jlhughes (approved):

TW80000 (commented):
I can't really comment anything useful on this PR since at the moment since I'm completely unfamiliar with our journal-related code. I'll have to read some code before I'm able to review this in any meaningful way.

PR #2630, opened 2019-09-25
Fix rules due to Rule.id change

lchanmann:
## Changes

- Update rule.id in rules.jrl files
- Update references to rule.id and rule.name


src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -161,7 +161,8 @@ foam.CLASS({
         Session session = x.get(Session.class);
         session.setUserId(user.getId());
         ((DAO) getLocalSessionDAO()).put(session);
-        session.setContext(session.applyTo(session.getContext()));
+        // QUESTION: can we applyTo current context instead session context as below?
+        session.setContext(session.applyTo(x));
```
lchanmann: @TW80000 can we applyTo the current context instead of the session context?

TW80000: I don't think so. The current context contains the request and response, as well as anything else that other decorators have added. The context also could have been swapped out for the admin context at any point prior.

It's safe to apply to the session context itself though, because we're careful about when and how we update the session context. The "effective context" used for each request has no such guarantees and can modified anywhere in a number of different ways. We don't want to risk all of those unknowns making it into the session context, because then they'll be there for all subsequent requests.

Why do you want to apply to the current context anyway? Maybe I can help think of a different way to accomplish what you're trying to do.

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2630/commits/88e19f9b8a40462409895bf289d35d7366fb828b.

lchanmann: Thanks @TW80000.

jlhughes (approved):

src/foam/nanos/auth/UserAndGroupAuthService.js
TW80000: No problem

PR #2632, opened 2019-09-26
Add ServiceProviderAware interface

jlhughes:
* ServiceProviderAwareDAO to enforce spid on new objects.  
* AppConfig provides a default spid - allowing default spid per application controller. 
* modelled AuthorizationDAO and Authorizer (initially used by ServiceProviderAwareDAO, but the dependency has been removed)
* EasyDAO test for Aware objects directly tests the assignability of the service rather than require the user to explicitly enable/disable the option.


src/foam/dao/EasyDAO.js
```diff
@@ -514,6 +521,11 @@ return delegate;
       name: 'testData',
       generateJava: false
     },
+    {
+      name: 'serviceProviderAware',
+      class: 'Boolean',
+      value: false
```
TW80000: It would be slightly nicer in my opinion if this property had a factory that returns `foam.nanos.auth.ServiceProviderAware.class.isAssignableFrom(getOf().getObjClass())`. Then you would avoid a situation in which someone sets this property to false and therefore expects there not to be a ServiceProviderAwareDAO when in fact there might still be one added. It gives the option of explicitly setting this to false to ignore the fact that `of` might implement ServiceProviderAware.

Not crucial though, just a suggestion.

src/foam/nanos/auth/ServiceProviderAware.js
```diff
@@ -0,0 +1,26 @@
...
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.INTERFACE({
+  package: 'foam.nanos.auth',
+  name: 'ServiceProviderAware',
+
+  methods: [
```
TW80000: Other interfaces I'd consider very similar to this one, like `LastModifiedAware`, `CreatedByAware`, etc. just define a property rather than getter and setter methods. Wouldn't we want `ServiceProviderAware` to be consistent with the other ones? Or is there a reason we want this to be different?

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'ServiceProviderAwareDAO',
+  extends: 'foam.nanos.auth.AuthorizationDAO',
```
TW80000: This doesn't seem right to me. A decorator that adds awareness of the SPID property shouldn't have anything to do with authorization checks.

Edit: See my comment below.

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -0,0 +1,67 @@
...
+    if ( isCreate ) {
+      User user = (User) x.get("user");
+      if ( user != null &&
+           ! SafetyUtil.isEmpty(user.getSpid()) ) {
+        sp.setSpid(user.getSpid());
+      } else {
+        sp.setSpid(getDefaultSpid());
+      }
+    } else if ( ! oldSp.getSpid().equals(sp.getSpid()) ) {
+      authorizer_.authorizeOnUpdate(x, (FObject) oldSp, obj);
```
TW80000: I think you could just replace this line with

```suggestion
      if ( ! ((AuthService) getAuth()).check(x, "spid.update.*") ) {
        throw new AuthorizationException("You do not have permission to update SPIDs.");
      }
```

and get rid of the `extends: AuthorizationDAO`.

jlhughes: Yes, I could but isn't that the point of AuthorizationDAO?

src/foam/nanos/auth/ServiceProviderAware.js
jlhughes: It's different because it already is different.  ServiceProvider.js exists and User already was using spid. 
And I don't understand the 'just add a property' - this is an interface. Interfaces don't support properties.

src/foam/dao/EasyDAO.js
jlhughes: I like that.

src/foam/nanos/auth/ServiceProviderAwareDAO.js
TW80000: No, AuthorizationDAO does much more. It's given an Authorizer and applies it to objects in the DAO during CRUD operations. The methods on `Authorizer` are `authorizeOnCreate(X, FObject)`, `authorizeOnRead(X, FObject)`, `authorizeOnUpdate(X, FObject)`, and `authorizeOnDelete(X, FObject)`. If the Authorizer is a `StandardAuthorizer`, it'll check permissions. If it's an `AuthorizableAuthorizer`, it'll call the `Authorizable` methods on the objects, which may or may not do permission checks. It depends on how the object chose to implement the `Authorizable` interface.

Here you're using an `AuthorizableAuthorizer`. I think you might have had `StandardAuthorizer` in mind. Using a `StandardAuthorizer` with "spid" as its permission prefix would check the "spid.update.<spid>" permission on update, "spid.create.<spid>" permission on create, etc. It would apply it for all CRUD operations on the DAO. However, that's not what we want here. We only care about protecting the SPID property from unauthorized changes, not stopping users from making any updates/creates/reads/deletes at all if they lack the spid permission.

We really only care about 1 specific case here: updating a SPID on a SpidAware FObject. Therefore it makes more sense just to use 1 specific permission check.

src/foam/nanos/auth/ServiceProviderAware.js
TW80000: > And I don't understand the 'just add a property' - this is an interface. Interfaces don't support properties.

Interfaces don't in general in the OOP sense, but they appear to in FOAM. Which makes them more like mix-ins or traits than an interface, but it's still a `foam.INTERFACE`.

See the definition of `CreatedByAware` (and the other ones I mentioned above):

```Java
foam.INTERFACE({
  package: 'foam.nanos.auth',
  name: 'CreatedByAware',

  properties: [
    {
      class: 'Reference',
      of: 'foam.nanos.auth.User',
      name: 'createdBy',
      documentation: 'User who created the entry'
    }
  ]
});
```

Although there's a bug right now where saying a model implements one of these traits/mix-ins/interfaces doesn't actually add the property to the model, that model still has to add the property to the list "manually".

jlhughes: Well that's handy. Thanks.

TW80000 (approved):
Didn't test but code looks reasonable to me.

PR #2633, opened 2019-09-26
[CPF-2616] Update capability model for capabilities dashboard

KernelDeimos:


src/foam/nanos/crunch/Capability.js
```diff
@@ -95,6 +102,14 @@ foam.CLASS({
       class: 'StringArray',
       documentation: `List of permissions granted by this capability`
     },
+    {
+      name: 'price',
```
TW80000, line 106: Is this supposed to represent a monetary value? If so, we need to also store a reference to the currency the price is in.

src/foam/nanos/crunch/Capability.js
```diff
@@ -53,7 +55,12 @@ foam.CLASS({
     {
       name: 'description',
       class: 'String',
-      documentation: `Description of capability`
+      documentation: `Brief description of capability`
+    },
+    {
+      name: 'longDescription',
```
TW80000, line 61: This strikes me as unnecessary. Can you give an example of when we'd want both a short and long description? Are they both user-facing?

KernelDeimos: My thinking is `description` would be a summary like "Make USD Payments", while `longDescription` would be a block of text explaining the capability

kgrgreer: This seems reasonable to me. Like an entry in an app store which has both a one line and much longer description.

src/foam/nanos/crunch/Capability.js
kgrgreer: Yes.

jlhughes: @KernelDeimos you'll need an additional property to hold the 'currency' value.

PR #2634, opened 2019-09-26
[CPF-2616] Add MenuDefinedViewList

KernelDeimos:
Currently it is difficult to specify children of a view for a menu item or navigation action. This can be useful to decouple navigation items from pages without the need to create a new model for every pairing.

Example use:
```javascript
self.stack.push({
  class: 'foam.nanos.menu.MenuDefinedViewList',
  containerView: 'foam.u2.layout.Rows',
  childViews: [
    {   
      class: 'foam.nanos.crunch.ui.CapabilityPageMenu'
    },  
    {   
      class: 'foam.nanos.crunch.ui.CapabilityInfoView'
    }   
  ]   
}); 
```

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -385,6 +385,7 @@ FOAM_FILES([
   { name: "foam/swift/refines/Remote", flags: ['swift'] },
 
   { name: "foam/nanos/menu/DAOMenu2" },
+  { name: "foam/nanos/menu/MenuDefinedViewList" },
 
   { name: "foam/box/LogBox" },
   { name: "foam/box/MultiDelegateBox" },
--- a/src/foam/nanos/menu/MenuDefinedViewList.js
+++ b/src/foam/nanos/menu/MenuDefinedViewList.js
@@ -1,0 +1,36 @@
+foam.CLASS({
+  package: 'foam.nanos.menu',
+  name: 'MenuDefinedViewList',
+  extends: 'foam.u2.View',
+
+  properties: [
+    {
+      name: 'containerView',
+      class: 'foam.u2.ViewSpec'
+    },
+    {
+      name: 'childViews',
+      class: 'FObjectArray',
+      of: 'foam.u2.View'
+    }
+  ],
+
+  methods: [
+    function initE() {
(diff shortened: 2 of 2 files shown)


KernelDeimos:
Note: This will also be useful to display capability selection using existing views

kgrgreer:
Couldn't you use a ViewMenu (that we already have) and then specify a configured Rows view for its view?

KernelDeimos:
@kgrgreer I tried doing this, but I ran into issues setting the `childNodes` property from a menu. I'll try again later so I can post a more specific comment on what happens.

KernelDeimos:
Setting either the `childNodes` or `children` properties in the following ways does not display the child nodes:
```
self.stack.push({ class: 'foam.u2.layout.Rows',
  childNodes: [
    { class: 'foam.nanos.crunch.ui.CapabilityPageMenu'},
    { class: 'foam.nanos.crunch.ui.CapabilityInfoView'} ] });
```
```
self.stack.push({ class: 'foam.u2.layout.Rows',
  children: [
    { class: 'foam.nanos.crunch.ui.CapabilityPageMenu'},
    { class: 'foam.nanos.crunch.ui.CapabilityInfoView'} ] });
```

Using literal views however does work. Unfortunately this can't be in a menu and causes other issues (breaks imports in the child view; not sure why):
```
self.stack.push({ class: 'foam.u2.layout.Rows',
  children: [
    foam.nanos.crunch.ui.CapabilityPageMenu.create(),
    foam.nanos.crunch.ui.CapabilityInfoView.create() ] });
```

TW80000 (commented):
How is this different from `foam.u2.MultiView`?

KernelDeimos:
MultiView does not let you set a container view, which complicates using it with layouts

KernelDeimos:
I think it would be better to change Element.js so that what Kevin described earlier is possible, but I'm unsure how that would be fixed. Could it require the addition of a ViewSpecArray property?

kgrgreer:
Let me experiment to see why this doesn't work and if I can fix it.

PR #2638, opened 2019-09-27
Added default tooltip

KateMengjunXia:
Related ticket: https://nanopay.atlassian.net/browse/CPF-2981
nanopay side of this PR: https://github.com/nanoPayinc/NANOPAY/pull/8003


TW80000 (commented):
What if we did something like this in `initE` and got rid of `initTooltip` entirely?

```JavaScript
var prevTooltipObj_ = null;
this.tooltip$.sub(() => {
  if ( prevTooltipObj_ ) prevTooltipObj_.remove();
  prevTooltipObj_ = this.Tooltip.create({
    target: this,
    text: this.tooltip || this.getAttribute('title')
  }, this);
});
```

KateMengjunXia:
> What if we did something like this in `initE` and got rid of `initTooltip` entirely?
> 
> ```js
> var prevTooltipObj_ = null;
> this.tooltip$.sub(() => {
>   if ( prevTooltipObj_ ) prevTooltipObj_.remove();
>   prevTooltipObj_ = this.Tooltip.create({
>     target: this,
>     text: this.tooltip || this.getAttribute('title')
>   }, this);
> });
> ```

The function inside `sub` never gets called if the element's `tooltip` is not explicitly initialized (the part where we want to use `this.getAttribute('title')` for). I'd like to change it to the following, which seems to work:
```js
if ( this.tooltip || this.getAttribute('title') != undefined ) {      
      var prevTooltipObj_ = this.Tooltip.create({
        target: this,
        text: this.tooltip || this.getAttribute('title')
      }, this);

      this.tooltip$.sub(() => {
        prevTooltipObj_.remove();
        prevTooltipObj_ = this.Tooltip.create({
          target: this,
          text: this.tooltip || this.getAttribute('title')
        }, this);
      });
     }
```

TW80000:
Like we talked about in person, I think this is the approach we want:

```JavaScript
    var prevTooltipObj_;

    if ( this.tooltip || this.getAttribute('title') != undefined ) {      
      prevTooltipObj_ = this.Tooltip.create({
        target: this,
        text: this.tooltip || this.getAttribute('title')
      }, this);
    }

    this.onDetach(this.tooltip$.sub(() => {
      if ( prevTooltipObj_ ) prevTooltipObj_.remove();
      var newTooltipText = this.tooltip || this.getAttribute('title');
      if ( newTooltipText ) {
        prevTooltipObj_ = this.Tooltip.create({
          target: this,
          text: newTooltipText
        }, this));
      }
    });
```

That way, if the tooltip isn't necessary when we initialize the element, then we don't create an unnecessary Tooltip element, but if it gets set later then we create the Tooltip element when it's needed.

PR #2641, opened 2019-09-30
Crunch expiry and daoKey

noodlemoodle:


src/foam/nanos/crunch/Capability.js
```diff
@@ -85,6 +85,12 @@ foam.CLASS({
       class: 'DateTime',
       documentation: `Datetime of when capability is no longer valid`
     },
+    {
+      name: 'duration',
+      class: 'Int',
```
TW80000, line 90: Can use `class: 'Duration'` instead if this value represents an amount of milliseconds. See https://github.com/foam-framework/foam2/blob/master/src/foam/core/types.js#L253-L269

src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
```diff
@@ -0,0 +1,41 @@
+package foam.nanos.session.cron;
```
TW80000: Missing license header.

src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
```diff
@@ -0,0 +1,41 @@
...
+    Date today = new Date();
+
+    // Only active, i.e., GRANTED, UserCapabilityJunctions should be checked
+    List<UserCapabilityJunction> activeJunctions = ((ArraySink) userCapabilityJunctionDAO
+      .where(EQ(UserCapabilityJunction.STATUS, CapabilityJunctionStatus.GRANTED))
+      .select(new ArraySink()))
+      .getArray();
+
+    for ( UserCapabilityJunction activeJunction : activeJunctions ) {
+      if ( (activeJunction.getExpiry()).before(today) ) {
```
TW80000: Do we know that `activeJunction.getExpiry()` is never going to be null? Can't they set duration instead of expiry?

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -56,7 +61,7 @@ foam.CLASS({
         User user = getUser(x);
         AuthService auth = (AuthService) x.get("auth");
         boolean isOwner = ((UserCapabilityJunction) obj).getSourceId() == user.getId();
-        boolean hasPermission = auth.check(x, "service.*");
+        boolean hasPermission = auth.check(x, "*");
```
TW80000: If we remove the `hasPermission` variable and just inline the expression below, then we'll avoid doing the auth check if it's not necessary (ie: when `isOwner` is true).

Also, the default message for `AuthenticationException` is "Permission denied", so on the line below we can just `throw new AuthorizationException();`.

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -99,16 +104,100 @@ foam.CLASS({
       javaCode: `
       checkOwnership(x, obj);
 
+      // if the junction is being uppdated from GRANTED to EXPIRED, put into junctionDAO without checking prereqs and data
```
TW80000: Typo: uppdated

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -99,16 +104,100 @@ foam.CLASS({
...
+      UserCapabilityJunction old = (UserCapabilityJunction) getDelegate().find_(x, ((UserCapabilityJunction) obj).getId());
+      if ( old.getStatus() == CapabilityJunctionStatus.GRANTED && ((UserCapabilityJunction) obj).getStatus() == CapabilityJunctionStatus.EXPIRED )
```
TW80000: What if `old` is null?

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -99,16 +104,100 @@ foam.CLASS({
...
+      UserCapabilityJunction old = (UserCapabilityJunction) getDelegate().find_(x, ((UserCapabilityJunction) obj).getId());
+      if ( old.getStatus() == CapabilityJunctionStatus.GRANTED && ((UserCapabilityJunction) obj).getStatus() == CapabilityJunctionStatus.EXPIRED ) 
+        return getDelegate().put_(x, obj);
+
       boolean prereq = checkPrereqs(x, obj);
       boolean data = validateData(x, obj);
 
-      if ( prereq && data ) ((UserCapabilityJunction) obj).setStatus(CapabilityJunctionStatus.GRANTED);
+      if ( prereq && data ) {
+        ((UserCapabilityJunction) obj).setStatus(CapabilityJunctionStatus.GRANTED);
```
TW80000: Should just do this cast once and put in a variable rather than inlining it so many times.

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -99,16 +104,100 @@ foam.CLASS({
...
+      // if the junction is being uppdated from GRANTED to EXPIRED, put into junctionDAO without checking prereqs and data
+      UserCapabilityJunction old = (UserCapabilityJunction) getDelegate().find_(x, ((UserCapabilityJunction) obj).getId());
+      if ( old.getStatus() == CapabilityJunctionStatus.GRANTED && ((UserCapabilityJunction) obj).getStatus() == CapabilityJunctionStatus.EXPIRED ) 
+        return getDelegate().put_(x, obj);
+
       boolean prereq = checkPrereqs(x, obj);
       boolean data = validateData(x, obj);
 
-      if ( prereq && data ) ((UserCapabilityJunction) obj).setStatus(CapabilityJunctionStatus.GRANTED);
+      if ( prereq && data ) {
```
TW80000: If we inline the checks rather than evaluating them both first and storing the results in variables, then we can avoid unnecessary work if the first one evaluates to false. We should put the one we suspect will be cheaper first as well. I'm assuming `validateData` is less expensive than `checkPrereqs`, so we should do:

```
if ( validateData(x, obj) && checkPrereqs(x, obj) ) { /* ... */ }

File: src/foam/nanos/crunch/UserCapabilityJunctionDAO.js [hunk 9]
```diff
@@ -99,16 +104,100 @@ foam.CLASS({
...
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'obj',
+          type: 'foam.nanos.crunch.UserCapabilityJunction'
+        }
+      ],
+      documentation: `save the UserCapabilityJunction data to its dao, if daoKey is provided in the Capability model`,
```
TW80000: Please provide more of an explanation here so that's it's clear why we need to do this. Maybe include an example as well. This is what Kevin, you and I talked about in person.

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -99,16 +104,100 @@ foam.CLASS({
...
+      javaCode: `
+      DAO capabilityDAO = (DAO) x.get("capabilityDAO");
+      Logger logger = (Logger) x.get("logger");
```
TW80000: Can move this above the place we use it so that we don't need to do the work of getting it out of the context when we don't need it.

src/services
```diff
@@ -536,10 +536,60 @@ p({
...
-p({"class":"foam.nanos.boot.NSpec","name":"prerequisiteCapabilityJunctionDAO","lazy":false,"serve":true,"authenticate":false,"serviceScript":"dao = new foam.dao.EasyDAO.Builder(x).setAuthorize(false).setJournalType(foam.dao.JournalType.SINGLE_JOURNAL).setJournalName(\"prerequisiteCapabilityJunctions\").setOf(foam.nanos.crunch.CapabilityCapabilityJunction.getOwnClassInfo()).build(); return new foam.nanos.auth.AuthorizationDAO(x, dao, new foam.nanos.auth.GlobalReadAuthorizer(\"prerequisiteCapabilityJunction\"));","client":"{\"of\":\"foam.nanos.crunch.CapabilityCapabilityJunction\", \"remoteListenerSupport\": false}"})
+
+p({
+  "class":"foam.nanos.boot.NSpec", 
+  "name":"capabilityDAO",                         
+  "lazy":true,  
+  "serve":true,  
+  "serviceScript":"""
+    dao = new foam.dao.PMDAO(x, new foam.dao.java.JDAO(x, foam.nanos.crunch.Capability.getOwnClassInfo(), \"capabilities\"));
```
TW80000: Escape characters aren't necessary in multi-line strings. Please remove them.

TW80000: Is there a reason we're not using `EasyDAO` here?

src/services
```diff
@@ -536,10 +536,60 @@ p({
...
+  """, 
+  "client":"{\"of\":\"foam.nanos.crunch.Capability\"}"
+})
+
+p({
+  "class":"foam.nanos.boot.NSpec",
+  "name":"userCapabilityJunctionDAO",
+  "lazy":false,
+  "serve":true,
+  "authenticate":false,
```
TW80000: This service is served and unauthenticated, and it also sets authorize to false. Is there any form of security on this DAO? If not then we need to add some.

src/services
```diff
@@ -536,10 +536,60 @@ p({
...
+  """,
+  "client":"{\"of\":\"foam.nanos.crunch.UserCapabilityJunction\", \"remoteListenerSupport\": false}"
+})
+
+p({
+  "class":"foam.nanos.boot.NSpec",
+  "name":"deprecatedCapabilityJunctionDAO",
+  "lazy":false,
+  "serve":true,
+  "authenticate":false,
```
TW80000: Same security-related comment as above.

src/services
```diff
@@ -536,10 +536,60 @@ p({
...
+  """,
+  "client":"{\"of\":\"foam.nanos.crunch.CapabilityCapabilityJunction\", \"remoteListenerSupport\": false}"
+})
+
+p({
+  "class":"foam.nanos.boot.NSpec",
+  "name":"prerequisiteCapabilityJunctionDAO",
+  "lazy":false,
+  "serve":true,
+  "authenticate":false,
```
TW80000: Same security-related comment as above.

TW80000 (changes requested):

src/foam/nanos/crunch/Capability.js
noodlemoodle: Sorry, should have commented in documentation. This represents the number of days, added in documentation

src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
noodlemoodle: added check for null expiry value, could you please explain what you mean in second sentence?

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
noodlemoodle: true, changed!

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
noodlemoodle: added check for null

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
noodlemoodle: true, changed!

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
noodlemoodle: yes, moved

src/services
noodlemoodle: set authenticate to true for all crunch services, in the case of usercapabilityjunctiondao, there is authorization on the UserCapabilityJunctionDAO decorator

src/services
noodlemoodle: no there isn't, migrated to easydao

src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
```diff
@@ -0,0 +1,47 @@
...
+    Date today = new Date();
+
+    // Only active, i.e., GRANTED, UserCapabilityJunctions should be checked
+    List<UserCapabilityJunction> activeJunctions = ((ArraySink) userCapabilityJunctionDAO
+      .where(EQ(UserCapabilityJunction.STATUS, CapabilityJunctionStatus.GRANTED))
+      .select(new ArraySink()))
+      .getArray();
+
+    for ( UserCapabilityJunction activeJunction : activeJunctions ) {
+      if ( activeJunction.getExpiry() != null && (activeJunction.getExpiry()).before(today) ) {
```
kgrgreer: Couldn't this have been included in the where() clause above?

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -97,18 +101,106 @@ foam.CLASS({
...
+      // Only update the expiry for non-active junctions, i.e., non-expired, non-pending, or granted junctions whose expiry is not yet set
+      if ( ( old != null && old.getStatus() == CapabilityJunctionStatus.GRANTED && old.getExpiry() != null ) || obj.getStatus() != CapabilityJunctionStatus.GRANTED ) 
+        return obj;
+
+      DAO capabilityDAO = (DAO) x.get("capabilityDAO");
+      Capability capability = (Capability) capabilityDAO.find((String) obj.getTargetId());
+      Date junctionExpiry = capability.getExpiry();
+      
+      if ( capability.getDuration() > 0 ) {
+        DateFormat format = new SimpleDateFormat("yyyy/MM/dd");
```
kgrgreer: Where is this used?

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -97,18 +101,106 @@ foam.CLASS({
...
+        Date today = new Date();
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(today);
+        calendar.add(Calendar.DATE, capability.getDuration());
+
+        if ( junctionExpiry == null ) junctionExpiry = calendar.getTime();
```
kgrgreer: Please don't ever format if/else's this way. If you have an else part then indent both parts and use {}'s.

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -195,7 +287,7 @@ foam.CLASS({
       name: 'find_',
       javaCode:`
       FObject result = super.find_(x, id);
-      if(result != null) checkOwnership(x, result);
+      if(result != null) checkOwnership(x, (UserCapabilityJunction) result);
```
kgrgreer: Fix spacing.

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
noodlemoodle: No, removed

src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
noodlemoodle: Yes, moved to where clause

PR #2645, opened 2019-09-30
easydao deletedawaredao decorator is using builder instead of constru…

noodlemoodle:
…ctor?


src/foam/dao/EasyDAO.js
```diff
@@ -171,7 +171,7 @@ if ( getServiceProviderAware() ) {
 }
 
 if ( getDeletedAware() ) {
-  delegate = new foam.nanos.auth.DeletedAwareDAO.Builder(getX()).setDelegate(delegate).build();
+  delegate = new foam.nanos.auth.DeletedAwareDAO(getX(), delegate);
```
TW80000, line 174: This works, but even better would be to model `DeletedAwareDAO`.

TW80000 (approved):

jlhughes:
What is the issue here?
And model it to do what?
Ok, forget that, I just checked - yes, it should have been modelled.

noodlemoodle:
> What is the issue here?
> And model it to do what?
> Ok, forget that, I just checked - yes, it should have been modelled.

Troy is working on modelling this, so I will be closing this PR

PR #2647, opened 2019-10-01
[CPF-3061] Add Java support for foam.core.Exception

KernelDeimos:
This PR adds support for generating Java exceptions with corresponding FOAM models.

### Creating an exception
```javascript
foam.EXCEPTION({
  package: 'foam.nanos.auth',
  name: 'AuthorizationException',
  javaExceptionExtends: 'SecurityException',
  javaSuperArgs: [ 'message' ],

  properties: [
    { class: 'String', name: 'message',
      value: 'Permission denied.'
    },
    { class: 'String', name: 'extra' }
  ]
});
```

### Throwing a native exception (edited)
```java
AuthorizationException ex = new AuthorizationException.Builder(x)
  .setExtra("Hello, World!")
  .build();
throw (AuthorizationNativeException) ex.newNative(null); // Uses default message
```

### Catching a native exception (edited)
```java
try {
  // ...
} catch (AuthorizationNativeException exJava) {
  System.out.println(exJava.getMessage()); // "Permission denied."

  AuthorizationException ex = (AuthorizationException) exJava.getFoamException();
  System.out.println(ex.getExtra()); // "Hello, World!"
}
```


KernelDeimos:
~~There's a slight problem right now where the message passed to `throwNative` is ignored. Attempting to fix it.~~ (fixed)

KernelDeimos:
In order to properly support an exception hierarchy I need to make some changes to the interfaces.

In particular, the getNativeException method's type is specific to the class, but this makes it impossible to extend, so really it should just be a `foam.core.NativeException`; also the getFoamException method of native exceptions should likewise be `foam.core.Exception`.

kgrgreer:
Can we not generate Builders for Exceptions and just use full-constructors instead?
Will review.

KernelDeimos:
The way I have it working right now is that two classes are generated:
- MyNameException - extends FObject, implements `foam.core.MessageException`
- MyNameNativeException - extends `java.lang.Exception`, can be thrown

MyNameException gets a builder because it's no different from other models. MyNameNativeException doesn't have a builder, and it's constructor takes `(foam.core.MessageException foamException, String message)` as its arguments.

KernelDeimos:
It may be worth mentioning here that last(x2) weekend I also made a script to migrate every `throw new AuthorizationException` for a modelled equivalent of the existing AuthorizationException. This script could be used for other exceptions too, but I wouldn't be surprised if AuthorizationException and AuthenticationException are the biggest ones.

KernelDeimos:
@kgrgreer The first part of the redesign (modelling an exception with properties with the model's name) is complete - next I need to add a function to generate an FObject (and a corresponding private class)

To do this I want to move most of the code in FObject.buildJavaClass to a new method of foam.java.Class, similar to fromFObjectClass in foam.java.Exception. Would this be okay?

nick-prat:
@kgrgreer We've updated this pr to now generate a native exception class alongside a foam exception class, with the names `someNativeException` and `someException` respectively.  

Both classes have the same properties, the only differences are that `someException` will have a `newNative()` public method that returns an instance of `someNativeException`. `someException` extends an `abstractException` while `someNativeException` extends `abstractNativeException` which extends `java.lang.Exception` to allow it to be thrown.  So for example, if you had a `UserFeebackException` instance you could throw it via `throw userFeedbackException.newNative()`.

If you would like to see an example of this, I've been using `foam.comics.v2.userfeedback.UserFeedbackException` for testing.  I would like a confirmation on this design before I further refine it and test it.

PR #2648, opened 2019-10-02
Delete empty spaces from the generated code.

hchoura:


src/foam/java/Field.js
```diff
@@ -41,9 +41,9 @@ foam.CLASS({
         this.final ? 'final ' : '',
         this.type, ' ', this.name);
       if ( this.initializer ) {
-        o.increaseIndent();
+        //o.increaseIndent();
```
kgrgreer: Delete these two lines rather than comment out.

src/foam/java/CodeProperty.js
```diff
@@ -23,6 +23,7 @@ foam.CLASS({
     {
       name: 'adapt',
       value: function(o, v) {
+        if ( v == 'false' || v == 'true') return v;
```
kgrgreer, line 26: What is this line for?

src/foam/java/PropertyInfo.js
```diff
@@ -148,21 +148,21 @@ foam.CLASS({
             type: this.propType,
             visibility: 'public',
             args: [{ name: 'o', type: 'Object' }],
-            body: 'return ((' + this.sourceCls.name + ') o).' + this.getterName + '();'
+            body: 'return ( (' + this.sourceCls.name + ') o ).' + this.getterName + '();'
```
kgrgreer: I don't think these spaces (and below) are required.

src/foam/java/refinements.js
```diff
@@ -952,9 +952,9 @@ foam.CLASS({
 
       var m = info.getMethod('cast');
       m.body = `return ( o instanceof Number ) ?
-        ((Number)o).intValue() :
+        ( (Number) o ).intValue() :
```
kgrgreer: The FOAM standard would be:
((Number) o)
So just one more space.

src/foam/java/CodeProperty.js
hchoura: if the code property is just a boolean, we need to output it rather than considering it as an entire instruction.

PR #2650, opened 2019-10-03
Fix bugs with DeletedAwareDAO

TW80000:
* Model `DeletedAwareDAO`
  * This fixes a bug where using the Builder for `DeletedAwareDAO` would return a ProxyDAO instead of a `DeletedAwareDAO` because `DeletedAwareDAO` was hand-written Java and inherited from `ProxyDAO`, which is modelled and thus generated and had a generated Builder. Since `EasyDAO` was using the Builder for `DeletedAwareDAO`, it wouldn't work as expected.
* Split `permissionRequired` into `readPermissionRequired` and `writePermissionRequired`
  * This lets us specify when a property requires permission to write but doesn't require permission to read. This is helpful for the `deleted` property, which we need users to be able to read because we need to use it in MLangs. The `removeAll_` implementation of `DeletedAwareDAO` does a select on the delegate with an `EQ(DELETED, false)`, which won't work if the user doesn't have permission to read the `deleted` property. Therefore we make the `deleted` property `writePermissionRequired` but not `readPermissionRequired`.


jlhughes:
@TW80000 regarding Phone, User, ... models.  The previous behaviour and intent was 'write' protection, so let's not add 'readPermissionRequired:true' on any of these models at this time.

TW80000:
> @TW80000 regarding Phone, User, ... models. The previous behaviour and intent was 'write' protection, so let's not add 'readPermissionRequired:true' on any of these models at this time.

Sounds good. I'd prefer to do that in a separate PR though just to keep things simple and to get this through.

jlhughes (approved):

PR #2652, opened 2019-10-03
Support higher-granularity property view modes & add sections for User properties

TW80000:
* Add `createMode`, `readMode`, and `updateMode` to `Property`
  * These new property hints can be used to specify the view mode of the
property under different controller modes. This opens up the possibility
to hide properties when creating an object but allow it to be edited
when the object is being edited, for example. It allows for more
fine-grained control than simply setting visibility, which should be
considered deprecated once this commit is merged.
* Set modes for a number of properties and add sections to User model
  * The administrative section requires permission to view
* Use `myClass` instead of `myCls` in Tooltip to fix console warning

### DIFF
--- a/src/foam/u2/detail/SectionedDetailPropertyView.js
+++ b/src/foam/u2/detail/SectionedDetailPropertyView.js
@@ -261,6 +261,8 @@ foam.CLASS({
     'foam.core.ProxySlot',
     'foam.u2.layout.Cols',
     'foam.u2.layout.Rows',
+    'foam.u2.ControllerMode',
+    'foam.u2.DisplayMode',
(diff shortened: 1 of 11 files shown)


src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -273,6 +275,19 @@ foam.CLASS({
       expression: function(prop) {
         return prop.createVisibilityFor(this.data$).map((m) => m !== this.Visibility.HIDDEN);
       }
+    },
+    {
+      name: 'mode',
```
kgrgreer, line 280: This is inelegant. Ask me how to do better.

PR #2654, opened 2019-10-04
Fix remove left margin for checkboxes

skanderturki:


src/foam/u2/CheckBox.js
```diff
@@ -75,7 +75,7 @@ foam.CLASS({
 
   css: `
     ^ {
-      margin: 4px;
+      margin: 4px 0px 4px 0px;
```
kgrgreer: Is this this same as just margin: 4px 0;  ?

skanderturki: yes

TW80000 (changes requested):
If this is to align the checkbox with the label in detail views, then we should also remove the top and bottom margin so it's consistent with the other inputs.

src/foam/u2/CheckBox.js
```diff
@@ -75,7 +75,7 @@ foam.CLASS({
 
   css: `
     ^ {
-      margin: 4px;
+      margin: 4px 0px;
```
TW80000:
```suggestion
      margin: 0;
```

skanderturki: I tried that but the checkbox becomes too close to the label

TW80000: That's not what I'm seeing. With no margin, it's exactly the same distance from the label as text inputs, whereas with 4px it's (obviously) 4px lower than it should be.

Screen Shot 2019-10-04 at 12 48 45 PM

skanderturki: You're right, in my pc (chrome or safari) there is absolutely no space for both input and checkboxes, so you're right it's more consistent to set all margins to 0


Screen Shot 2019-10-04 at 10 37 10 AM

TW80000 (approved):

PR #2659, opened 2019-10-04
Let SectionedDetailView support a property whitelist

TW80000:
`foam.u2.DetailView` supports this but lacks a number of features that `SectionedDetailView` has. In a few places we're extending `DetailView` so that we can provide a new detail view with a subset of properties that would normally be available. However, by doing so we miss out on all of the features that `SectionedDetailView` has but `DetailView` doesn't. By adding support for this feature to `SectionedDetailView`, we can extend `SectionedDetailView` instead and get the best of both worlds.

### DIFF
--- a/src/foam/u2/detail/AbstractSectionedDetailView.js
+++ b/src/foam/u2/detail/AbstractSectionedDetailView.js
@@ -42,6 +42,10 @@ foam.CLASS({
       class: 'FObjectArray',
       of: 'foam.core.Property',
       name: 'propertyWhiteList',
+      documentation: `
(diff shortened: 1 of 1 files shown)


src/foam/u2/detail/AbstractSectionedDetailView.js
```diff
@@ -38,6 +38,21 @@ foam.CLASS({
         return data && data.cls_;
       }
     },
+    {
+      class: 'FObjectArray',
+      of: 'foam.core.Property',
+      name: 'propertyWhiteList',
```
kgrgreer, line 44: Please document: what this is for.

PR #2660, opened 2019-10-04
Revert "fix safari RO css issue"

tharmaman:
Reverts foam-framework/foam2#2640

Border doesn't hide on read-only, reverting until new solution


kgrgreer:
FYI: @zahnno @yij793

PR #2665, opened 2019-10-07
Read-Only View Improvements

TW80000:
Create a separate read view for properties so that we can have more
appropriate views when a property is being read. This has a number of
advantages:

* Avoids needing to contort the "edit" view to look sensible in RO mode
by using CSS, which is often done imperfectly in practice
* Lets us do interesting things like displaying an address in one,
human-readable line instead of showing each field independently
* We can make Reference properties show as links in their RO view,
which would let users easily click around between objects in the system


src/foam/dao/DAOProperty.js
```diff
@@ -27,7 +27,18 @@ foam.CLASS({
   properties: [
     {
       name: 'view',
-      value: {class: 'foam.comics.InlineBrowserView'},
+      value: { class: 'foam.comics.InlineBrowserView' },
+    },
+    {
+      name: 'readView',
```
kgrgreer: I'm not too keen on having a readView on property because it means that I'll often have to set both view and readView at the same time. I would rather have the view decide. For convenience, we could have a ReadWriteView like:
view: { class: 'foam.u2.ReadWriteView', readView: ...,  writeView }
But if I have a custom view for colours, URLs, telephone numbers, etc. I shouldn't have to know to set two views, one for edit and one for view mode.

TW80000: > I'll often have to set both view and readView at the same time.

You only have to set it if you're not satisfied with the default values of those properties for your property type. Which is the same as it already is with just `view`. It's quite rare that you'd want to set both to the same view anyway.

> But if I have a custom view for colours, URLs, telephone numbers, etc. I shouldn't have to know to set two views, one for edit and one for view mode.

But with what you're suggesting, you'd still have to know to set `view` to `ReadWriteView` instead of your custom view directly. Either way you need to be aware of the distinction between the read view and the write view, so moving it doesn't buy you anything.

I don't particularly like the idea of having just `view` because almost every property's view will become `ReadWriteView`, and if you want to override it you'll need to copy the one of the two that you don't want to override.

For example, if we did what you're suggesting, the `view` of `FObjectProperty` would become

```
{
  class: 'foam.u2.view.ReadWriteView',
  readView: {
    class: 'foam.u2.view.CitationView'
  },
  writeView: {
    class: 'foam.u2.DetailView'
  }
}
```

and now if I want to override the read view for any FObjectProperty I need to set it to

```
{
  class: 'foam.u2.view.ReadWriteView',
  readView: {
    class: 'my.custom.thing.CustomView'
  },
  // Manually copy and paste this from the default because I don't want to change this
  writeView: {
    class: 'foam.u2.DetailView'
  }
}
```

rather than simply setting the `readView` directly to the custom view and leaving the write view as its default.

jlhughes: @kgrgreer Troy's argument is sound.

TW80000: @kgrgreer and I talked in person and decided to stick with just `view` so that the "API" of views doesn't change. Each property has a single view, and it's up to that view to decide what to show based on the mode of the view. Therefore we created `ModeAltView`, which shows different views based on the mode. A new view was created for each property type (or at least most of them) that extends `ModeAltView` and sets the `readView` and `writeView` to whatever is appropriate for that type. New views were created where there were no suitable ones already available.

src/foam/u2/view/CurrencyView.js
```diff
@@ -0,0 +1,29 @@
...
+
+  requires: [
+    'foam.u2.CurrencyView',
+    'foam.u2.view.ValueView'
+  ],
+
+  properties: [
+    {
+      name: 'readView',
+      value: { class: 'foam.u2.view.ValueView' }
```
kgrgreer, line 22: Do we want to format the currency here?

src/foam/u2/view/DateView.js
```diff
@@ -0,0 +1,29 @@
...
+
+  requires: [
+    'foam.u2.DateView',
+    'foam.u2.view.ValueView'
+  ],
+
+  properties: [
+    {
+      name: 'readView',
+      value: { class: 'foam.u2.view.ValueView' }
```
kgrgreer, line 22: Would it make sense for this to be the default value for ModeAltView.readView?

src/foam/u2/view/ReferencePropertyView.js
```diff
@@ -0,0 +1,29 @@
...
+
+  properties: [
+    {
+      name: 'readView',
+      value: { class: 'foam.u2.view.ReadReferenceView' }
+    },
+    {
+      name: 'writeView',
+      value: { class: 'foam.u2.view.ReferenceView' }
+    }
```
kgrgreer: I think a shorter version would work:
properties: [
  [ 'readView', 'foam.u2.view.ReadReferenceView' ],
  [ 'writeView', 'foam.u2.view.ReferenceView' ]
]

src/foam/u2/view/ValueView.js
```diff
@@ -0,0 +1,22 @@
...
+
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'ValueView',
+  extends: 'foam.u2.View',
+
+  documentation: 'Just shows the value of data as a string.',
+
+  methods: [
+    {
```
kgrgreer: Just:
methods: [
  function initE() {
        return this.add(this.data);
   }
]

would be shorter

src/foam/u2/view/DateView.js
TW80000: I would argue that there shouldn't be default views for `ModeAltView`. It's essentially a combinator, so any two views will work. To give one of them a default value related to primitive properties might give the impression that `ModeAltView` is in some sense specific to primitive properties, whereas without the defaults it's more clear that it's simply a combinator.

src/foam/u2/view/ValueView.js
TW80000: Resolved by a1584561a

src/foam/u2/view/ReferencePropertyView.js
TW80000: Resolved by a1584561a

src/foam/u2/view/CurrencyView.js
TW80000: Yes, but currently `foam.core.Currency` properties have no way of keeping track of which currency in particular the amount is for, so we have no way of knowing how to format it without making assumptions. I'd like to look into remedying this situation, but I think doing so falls outside of the scope of this PR.

src/foam/u2/view/DateView.js
kgrgreer: Then would something like an AnyReadView be a good idea? (for a future PR, possibly), which really only just handled primitives and FObjects (as Detail or CitationViews)?

TW80000: If all of the primitive types and FObjectProperty already have correct read views by default (as set by this PR), then is there any point in adding AnyReadView? I don't see the use case for it.

PR #2670, opened 2019-10-07
Add views to PropertyUpdate

KernelDeimos:


jlhughes (approved):

kgrgreer:
Couldn't we make AnyView just handle FObject instead?

PR #2671, opened 2019-10-07
Add 'enabled' flag and honour it during menu display.

jlhughes:


TW80000 (approved):
Didn't test but LGTM.

PR #2674, opened 2019-10-08
Update AnyView to render FObject

KernelDeimos:


kgrgreer:
Please resolve conflicts.

src/foam/u2/view/AnyView.js
```diff
@@ -54,6 +54,14 @@ foam.CLASS({
               return foam.Object.isInstance(o) ? o : {};
             }
           }),
+          foam.u2.view.AnyView.Choice.create({
+            label: 'FObject',
+            type: foam.core.FObject,
+            view: foam.u2.view.FObjectView,
+            toType: function(o) {
+              return foam.core.FObject.isInstance(o) ? o : {};
```
TW80000:
```suggestion
              return foam.core.FObject.isInstance(o) ? o : foam.core.FObject.create();
```

TW80000 (commented):
Not sure if this really matters, but if the idea of `toType` is to do a "best effort" conversion to that type, then an instance of `FObject` is closer to `FObject` than a plain old JavaScript object.

src/foam/u2/view/AnyView.js
KernelDeimos: That makes sense; I committed this change. I think right now this case will never happen, but it would be an awfully subtle bug if this changed in the future.

src/foam/u2/view/AutomaticView.js
```diff
@@ -1,25 +0,0 @@
-foam.CLASS({
-  package: 'foam.u2.view',
-  name: 'AutomaticView',
```
TW80000, line 3: Please remove `AutomaticView` from `files.js` as well.

TW80000 (changes requested):

TW80000:
Tested with the change to `files.js` and it works. Will approve once that change is made.

PR #2676, opened 2019-10-08
[CPF-3110] Make liveScriptBundler block requests while building

KernelDeimos:


src/foam/nanos/http/LiveScriptBundler.java
```diff
@@ -31,6 +32,7 @@
 
   // Caching
   protected String javascriptBuffer_;
+  protected ReentrantLock requestBlockingLock_;
```
kgrgreer: Is this needed? Is the doRebuildJavascript() method reentrant? Couldn't you just make that method synchronized and then to a synchronized (this) {} catch(InterruptedException e) {} in execute()?

KernelDeimos: Do synchronized methods and synchronized blocks interact that way, or would I need to add two synchronized blocks instead?

KernelDeimos: ~After doing some reading, I think that would still allow a request to go through during building. The reason is that it would behave equivalently to having two separate reentrant locks (one to each synchronized block)~ After reading something that was worded much less confusingly I'm going to push that change

jlhughes (approved):

PR #2677, opened 2019-10-08
Filter System v1

kennyqiyenkan:
New Filter System.

Aside from introducing a more user friendly design, it fundamentally changes our full search to use OR predicates rather than just the one selection. This should add some flexibility.

In regards to Relationships and Reference Filtering, a target/source DAO Key must be provided, and have permissions to, for it to work. (So this might break things) From my understanding, a reference with a provided `of` would figure out the targetDAOKey. However, for cases which require a specific DAO to be used for a Reference, a targetDAOKey needs to be defined.

Also note, this is v1.0. Hopefully, 2.0 will include a more dynamic property search rather than relying on searchColumns of a model.

**Notes for V2**:
Add a more option to allow users to filter properties not predefined by devs.
Add an advanced option to allow verbose search by power users


src/foam/u2/filter/BooleanFilterView.js
```diff
@@ -0,0 +1,133 @@
...
+    ^container .foam-u2-md-CheckBox:checked {
+      background-color: #406dea;
+      border-color: #406dea;
+    }
+  `,
+
+  properties: [
+    {
+      name: 'property',
+      documentation: `The property that this view is filtering by. Should be of
```
jlhughes, line 62: For this and the others, is this requirement enforced anywhere?  What does configuration look like for developers - what is required in the model or table views?

kennyqiyenkan: `property` right now is being populated by the property defined in `searchColumns` in a model. So like before, if you define a property in `searchColumns`, it will be used as a filterable parameter.

As far as my knowledge goes, it is not enforced atm

jlhughes:
With regard to breaking things. The logic uses the property.targetDAOKey which defaults to the daoize name which is correct for 95% of daos.  We'll just have to review views in ablii that use such things as publicBusinessDAO ... for example, but then ablii has it's own search at the moment, so doubt current ablii will be an issue.

deejes:
@googlebot I consent.

src/foam/u2/filter/BooleanFilterView.js
```diff
@@ -0,0 +1,133 @@
...
+      type Boolean.`,
+      required: true
+    },
+    {
+      class: 'Boolean',
+      name: 'bool1',
+      documentation: `Lets the user pick boolean they want to filter by`,
```
kgrgreer: What does this mean?

src/foam/u2/filter/BooleanFilterView.js
```diff
@@ -0,0 +1,133 @@
...
+    },
+    {
+      name: 'predicate',
+      documentation: `All SearchViews must have a predicate as required by the
+      SearchManager. The SearchManager will read this predicate and use it
+      to filter the dao being displayed in the view.`,
+      expression: function(bool1, bool2) {
+        if ( ( ! bool1 && ! bool2 ) || ( bool1 && bool2 )) return this.True.create();
+
+        return bool1 ?
```
kgrgreer: This block would be shorter as just:
foam.mlang.predicate.Eq.create({
            arg1: this.property,
            arg2: bool1
          })

src/foam/u2/filter/DateFilterView.js
```diff
@@ -0,0 +1,140 @@
...
+      documentation: `A date selectable by the user. Used as the second argument
+          to the predicate generated by this view.`,
+    },
+    {
+      name: 'predicate',
+      documentation: `All SearchViews must have a predicate as required by the
+          SearchManager. The SearchManager will read this predicate and use it
+          to filter the dao being displayed in the view.`,
+      expression: function(qualifier, date1, date2) {
+        if ( ! qualifier || ! date || ! isNaN(date.valueOf()) ) return this.True.create();
```
kgrgreer: If instead of importing True you implement: foam.mlang.Expressions, then you can just do this.TRUE.

src/foam/u2/filter/DateFilterView.js
```diff
@@ -0,0 +1,140 @@
...
+
+        if ( qualifier !== 'Bt' ) {
+          return foam.mlang.predicate[qualifier].create({
+            arg1: this.property,
+            arg2: new Date(date1)
```
kgrgreer: Isn't date1 already a Date?

src/foam/u2/filter/DateFilterView.js
```diff
@@ -0,0 +1,140 @@
...
+          });
+        }
+
+        return foam.mlang.predicate.And.create({
```
kgrgreer: Easier to use MLang.AND() and GT and LT.

src/foam/u2/filter/EnumFilterView.js
```diff
@@ -0,0 +1,248 @@
...
+        var ordinals = selectedOptions.map(option => option.ordinal);
+
+        if ( ordinals.length === 1) {
+          return foam.mlang.predicate.Eq.create({
+            arg1: this.property,
+            arg2: ordinals[0]
+          });
+        }
+
+        var pred = foam.mlang.predicate.Or.create({ args: [] });
```
kgrgreer: Easier to use IN.

src/foam/u2/filter/BooleanFilterView.js
kennyqiyenkan: Just documentation to explain the purpose of it. Would it be better to name them perhaps?

src/foam/u2/filter/DateFilterView.js
kennyqiyenkan: true. I was just doing a very mcarcaso thing to do. Haha.

src/foam/u2/filter/EnumFilterView.js
```diff
@@ -0,0 +1,248 @@
...
+          }))
+          .add(this.slot(function(selectedOptions, filteredOptions) {
+            var element = this.E();
+            return element
+              .start('p').addClass(self.myClass('label-section'))
+                .add(self.LABEL_FILTERED)
+              .end()
+              .call(function() {
+                self.filteredOptions.forEach(function(option, index) {
+                  const label = option.label ? option.label : option.name;
```
kgrgreer: shorter as:  option.label || option.name;

src/foam/u2/filter/FilterSearch.js
```diff
@@ -0,0 +1,118 @@
...
+* @license
+* Copyright 2019 The FOAM Authors. All Rights Reserved.
+* http://www.apache.org/licenses/LICENSE-2.0
+*/
+
+foam.CLASS({
+  package: 'foam.u2.filter',
+  name: 'FilterSearch',
+  extends: 'foam.u2.View',
+
```
kgrgreer, line 11: Add documentation:

src/foam/u2/filter/FilterSearch.js
```diff
@@ -0,0 +1,118 @@
...
+        if ( of.model_.tableColumns ) {
+          return of.model_.tableColumns.filter(function(c) {
+            var axiom = of.getAxiomByName(c);
+            return axiom && axiom.searchView;
+          });
+        }
+
+        return of.getAxiomsByClass(foam.core.Property)
+        .filter((prop) => prop.searchView && ! prop.hidden)
```
kgrgreer: indent next two lines

src/foam/u2/filter/FilterViewController.js
```diff
@@ -0,0 +1,163 @@
...
+
+foam.CLASS({
+  package: 'foam.u2.filter',
+  name: 'FilterViewController',
+  extends: 'foam.u2.View',
+
+  documentation: `Manages the FilterViews`,
+
+  requires: [
+    'foam.mlang.predicate.True'
```
kgrgreer: Easier to implements: MLang

src/foam/u2/filter/FilterViewController.js
```diff
@@ -0,0 +1,163 @@
...
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.filter',
+  name: 'FilterViewController',
+  extends: 'foam.u2.View',
+
+  documentation: `Manages the FilterViews`,
```
kgrgreer: No need to use backquote for a single-line string.

src/foam/u2/filter/IntegerFilterView.js
```diff
@@ -0,0 +1,162 @@
...
+          });
+        }
+
+        var pred1 = 'Gt';
+        var pred2 = 'Lt';
+        if ( qualifier === 'Bte' ) {
+          pred1 = 'Gte';
+          pred2 = 'Lte';
+        }
+        return foam.mlang.predicate.And.create({
```
kgrgreer: Easier to use MLang.AND

src/foam/u2/filter/BooleanFilterView.js
```diff
@@ -0,0 +1,128 @@
...
+      name: 'property',
+      documentation: `The property that this view is filtering by. Should be of
+      type Boolean.`,
+      required: true
+    },
+    {
+      class: 'Boolean',
+      name: 'bool1',
+      documentation: `Lets the user pick boolean they want to filter by`,
+      value: false
```
nanoscott, line 71: Shouldn't the value for bool1 be 'true'? Should you be adding labels and more explicit documentation to both bool1 and bool2 so it is clear what values they represent?

src/foam/u2/filter/DateFilterView.js
```diff
@@ -0,0 +1,129 @@
...
+      border: solid 1px #cbcfd4;
+      background-color: #ffffff;
+    }
+  `,
+
+  properties: [
+    {
+      name: 'property',
+      documentation: `The property that this view is filtering by. Should be of
+          type Date.`,
```
nanoscott, line 50: You say - 'should be of type date' .. do you want to add validation to enforce the 'should'?

kennyqiyenkan: The way it is currently implemented, the developer shouldn't be setting the property. The act of defining a `date` property in their models should be enough.

kgrgreer:
Is this okay to Merge?

kennyqiyenkan:
If there isn't any other issues or comments, it should be safe. Am trying to get more data to play with to test out the filters properly as well.

kgrgreer:
@kennyqiyenkan Please reply or resolve to my comments. Thx

PR #2683, opened 2019-10-09
Set contentType for xml on dig

minsunny:
After put done,  its returning message blocked but this can be a fix.


jlhughes (approved):
It looks like we flip between returning            
resp.setContentType("application/xml");
and 

for format XML.   
Are we returning 'xml' data or 'html' data?

minsunny:
@jlhughes 

for "Put" after put done, it returning a text type message,
but "Select" it returning a xml type.

PR #2685, opened 2019-10-09
Corrects alignment of text inputs

skanderturki:


TW80000:
@skanderturki can you give me an example of where in the UI the alignment is incorrect?

skanderturki:
In Operations->Users (On chrome) for example

skanderturki:
The id, legal name and email are in smaller font and with 8px additional left padding

kgrgreer:
@TW80000 Please review. @skanderturki remove rather than just comment out.

skanderturki:
ok (i'll remove the line)

TW80000:
The alignment isn't incorrect. There's supposed to be consistent horizontal and vertical padding on all inputs as determined by the theme. The default theme has a horizontal input padding of 8px, so the 8px padding is correct as is.

I see what you mean though. When in read-only mode, we hide the border of inputs but the padding stays, so it looks like the fields are misaligned:

Screen Shot 2019-10-09 at 3 30 55 PM

The issue here isn't that inputs have padding, it's that we're using inputs in read-only mode in the first place. I have a PR open that adds a read view for properties that's separate from the write view, so we will no longer use inputs for read-only views in that case. See https://github.com/foam-framework/foam2/pull/2665.

TW80000 (changes requested):
We shouldn't remove padding from inputs. They need to have padding that comes from the theme. Instead, we need to stop using inputs as views for properties in read-only mode.

skanderturki:
Yes, that's why I only commented out the line in the beginning but maybe for the demo they should use this fix until the problem is resolved in a better way

TW80000:
@kgrgreer If you want to merge this today for the sake of the demo, then we can revert it afterwards and move forward with #2665 for the proper fix.

TW80000:
@skanderturki https://github.com/foam-framework/foam2/pull/2665 has been merged and this issue should have been fixed by it. Can you please check and close this PR if the issue has been fixed?

skanderturki:
OK

skanderturki:
> @skanderturki #2665 has been merged and this issue should have been fixed by it. Can you please check and close this PR if the issue has been fixed?

It looks great now

PR #2686, opened 2019-10-10
DO_NOT_MERGE add invalidx to authorizationdao builder

noodlemoodle:
https://github.com/nanoPayinc/NANOPAY/pull/8113 nanopay pr

### DIFF
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -220,7 +220,7 @@ if ( getOrder() != null &&
   }
 }
 
-if( getAuthorize() ) {
+if ( getAuthorize() ) {
   delegate = new foam.nanos.auth.AuthorizationDAO.Builder(getX())
     .setDelegate(delegate)
     .setAuthorizer(getAuthorizer())
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -220,7 +220,12 @@ if ( getOrder() != null &&
   }
 }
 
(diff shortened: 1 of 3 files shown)


kgrgreer:
Why not to merge?

noodlemoodle:
> Why not to merge?


haven't tested it very well yet, and need to fix failing tests on nanopay side

src/foam/dao/EasyDAO.js
```diff
@@ -220,7 +220,12 @@ if ( getOrder() != null &&
   }
 }
 
-if( getAuthorize() ) delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate, getAuthorizer());
+if( getAuthorize() ) {
```
```suggestion
if ( getAuthorize() ) {
```

src/foam/nanos/auth/AuthorizationDAO.js
```diff
@@ -57,6 +57,17 @@ foam.CLASS({
   ],
 
   methods: [
+    {
+      name: 'init_',
+      javaCode: `
+      AuthorizationException exception = new AuthorizationException("When " +
```
TW80000: Please indent the whole code block one more level.

TW80000 (changes requested):

TW80000:
As discussed in person, we're going to do the following instead:

* Make `InvalidX` extend `ProxyX`
* In `Boot.java` when initializing the services, set the context of ContextAware DAOs to `InvalidX` which delegates to the system context
* Modify `InvalidX` to log a warning rather than throw an error

PR #2687, opened 2019-10-10
User label adjust

Nauna:
keep consistent cross platform & fix to conditions


jlhughes (approved):

src/foam/nanos/auth/User.js
```diff
@@ -435,13 +435,22 @@ foam.CLASS({
       name: 'label',
       type: 'String',
       code: function label() {
-        return this.organization || this.businessName || ( this.lastName ? this.firstName + ' ' + this.lastName : this.firstName );
+        if ( this.lastName && this.firstName ) return this.firstName + ' ' + this.lastName;
```
jlhughes, line 444: Isn't this just the legalName?

TW80000 (commented):
I think the label should just be the `legalName` for `User` and `Contact`, and `Business` should override it to return the `businessName` or `organization`.

jlhughes:
I concur with Troy.

Nauna:
> I think the label should just be the `legalName` for `User` and `Contact`, and `Business` should override it to return the `businessName` or `organization`.

External Contacts need organization/businessName logic.

@TW80000 @jlhughes My proposed logic makes this function defined once and conforms to everyone's needed logic. Overriding the function, would require this function to be placed on, User, Contact, and Business. And from what I understand the business logic is changing and we won't need a label() function for this soon.

Why is an override a preferred method here??

TW80000:
> External Contacts need organization/businessName logic.

Ok, then Contact would just do that instead.

> @TW80000 @jlhughes My proposed logic makes this function defined once and conforms to everyone's needed logic. Overriding the function, would require this function to be placed on, User, Contact, and Business. And from what I understand the business logic is changing and we won't need a label() function for this soon.
> 
> Why is an override a preferred method here??

That's the whole idea of polymorphism. All 3 can be treated as `User`s (at least for the time being), and all 3 have a `label` method. Subclasses can choose to override the methods they inherit. Just like Business and Contact override the `authorizeOnRead` method they inherit from user instead of adding a ton of conditionals or switch cases to check if instance of Business or Contact in the implementation of `authorizeOnRead` in User.

jlhughes:
We are trying to split User from Business. A user should not have a business property such as business name, but can have a relationship to a business.   
At the moment this is messy, and from javascript it would be messy to get the Contacts Business if it exists and get it's name for the label.  
For the time being, since everything descends from User, overriding label, in Contact and  Business,  to return the appropriate value is the way to go for now.

jlhughes:
I don't think organization should be used in any of this.   Organization is an association - like your employer.   businessName should be restricted to business or contact.
User - legalName | first/lastName
Contact - businessName |  super
Business - businessName | super

src/foam/nanos/auth/User.js
Nauna: yes but the expression||factory I once remember on legalName is no longer there. Can't find it defined for a user.

So this is just a fail safe.

jlhughes: @Nauna yes , you are correct, the expression for legalName has been removed. 
So legalName followed by first/last.

src/foam/nanos/auth/User.js
```diff
@@ -175,7 +175,7 @@ foam.CLASS({
       width: 100,
       tableWidth: 160,
       validateObj: function(organization) {
-        if (!(organization.trim())) {
+        if ( ! (organization.trim()) ) {
```
jlhughes: Side question, why do we say 'Company' name is required for 'Organization'? organization is perfectly correct term on it's own.  For an an employee 'user' the organization is the entity that employs them.  We have businessName for business/company.

jlhughes (approved):

jlhughes:
I'm good with this now.

src/foam/nanos/auth/User.js
Nauna: lol. Nice attention to detail. What you are saying makes sense to me. So looks like companyName is associated to the mapping to businessName. And somewhere along the way organization was used as businessName as well. So now they are the same :(

Yikes this should be cleaned up. If you give the thumbs up @jlhughes I'll make a ticket that will hopefully clean this all up.

I will change this right now.

jlhughes: re: Organization / BussinessName - yes, please create a ticket.  Suspect this will impact many views. as some onboarding views only test for organization and not businessname.

PR #2688, opened 2019-10-10
Tooltip position

KateMengjunXia:
https://nanopay.atlassian.net/browse/CPF-3055


kgrgreer:
What does this do? Does the tooltip move around with the mouse?

KateMengjunXia:
> What does this do? Does the tooltip move around with the mouse?

This makes tooltips move around with the mouse. Yes.

PR #2690, opened 2019-10-10
Different heights dropdown text fields

skanderturki:
In some user interfaces (example: Operations -> Users), drop down lists and text inputs have different hights.

### DIFF
--- a/src/foam/u2/DetailPropertyView.js
+++ b/src/foam/u2/DetailPropertyView.js
@@ -44,6 +44,7 @@ foam.CLASS({
     }
     .foam-u2-PropertyView-view {
       padding: 2px 8px 2px 6px;
+      height: 48px !important;
     }
     .foam-u2-PropertyView-units  {
       color: #444;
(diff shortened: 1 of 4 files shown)


kgrgreer:
LGTM

TW80000 (changes requested):
This doesn't look right to me. The way we make height consistent across inputs is by getting all inputs to respect the `inputHeight` value that comes from the `Theme` model.

Not only that, but not all views of properties are inputs, so forcing all of them to a fixed height doesn't make sense.

Could you provide a screenshot of where the heights look inconsistent? The users screen looks correct as is to me.

skanderturki:
Yes
Screen Shot 2019-10-10 at 3 09 20 PM

skanderturki:
Here status and compliance have a 35px height while name and id fore example have a 45px height

TW80000:
@skanderturki https://github.com/foam-framework/foam2/pull/2665 has been merged and this issue should have been fixed by it. Can you please check and close this PR if the issue has been fixed?

skanderturki:
ok

TW80000:
Fixed by #2665

PR #2693, opened 2019-10-15
Add default newItem value to ArrayView

KernelDeimos:
When creating an item in an ArrayView, an empty string value is added to the end of the array. In the case of an FObjectArrayView for an interface type, this causes an exception to be thrown when updating the parent slot.

This PR adds a `defaultNewItem` property to ArrayView, allowing the view of an FObjectProperty to specify a reasonable value for new items.

### DIFF
--- a/src/foam/u2/view/ArrayView.js
+++ b/src/foam/u2/view/ArrayView.js
@@ -10,6 +10,7 @@ foam.CLASS({
   extends: 'foam.u2.View',
 
   requires: [
+    'foam.core.FObject',
     'foam.u2.layout.Cols',
     'foam.u2.layout.Rows'
   ],
@@ -40,7 +41,7 @@ foam.CLASS({
       },
       code: function() {
         var newItem = this.defaultNewItem;
-        if ( foam.core.FObject.isInstance(newItem) ) {
+        if ( this.FObject.isInstance(newItem) ) {
           newItem = newItem.clone();
         }
         this.data[this.data.length] = newItem;
--- a/src/foam/u2/view/ArrayView.js
+++ b/src/foam/u2/view/ArrayView.js
@@ -24,6 +24,10 @@ foam.CLASS({
       class: 'foam.u2.ViewSpec',
       name: 'valueView',
       value: { class: 'foam.u2.view.AnyView' }
+    },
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/ArrayView.js
```diff
@@ -35,7 +39,11 @@ foam.CLASS({
         return mode === foam.u2.DisplayMode.RW;
       },
       code: function() {
-        this.data[this.data.length] = '';
+        var newItem = this.defaultNewItem;
+        if ( foam.core.FObject.isInstance(newItem) ) {
```
TW80000: Should add

```JavaScript
    'foam.core.FObject',
```

to the `requires` array and change this line to

```suggestion
        if ( this.FObject.isInstance(newItem) ) {
```

TW80000 (commented):
This works for arrays of Strings and FObjects now, but I wonder if there's a way to make it work for arbitrary types. Like if `ArrayView` had an `of :: Class` property and you could just add `of.defaultValue` to the array. Not sure if we have such a property on models though, although I think we do on the Java side.

If we do this, we could also add a `fromProperty` method to `ArrayView` that sets `of` based on the property's class.

src/foam/u2/view/ArrayView.js
KernelDeimos: Updated this

KernelDeimos:
@TW80000 I was thinking about the solution you mentioned earlier and thought of this:

- Every subclass of `foam.core.Property` has a default value
- Every array subclass of `foam.core.Property` has a default value of `[]`.

What if it was a requirement for any Array property have both a default value of `[]`, and a default value for new items?

So for `foam.core.StringArray` it would be `[]` and `""`;
for `foam.core.FObjectArray` it would be `[]` and `this.of.create()`

TW80000 (changes requested):
This looks good, but I'd also like to add the following to `FObjectArrayView` as part of this PR since it extends `ArrayView`:

```JavaScript
{
  name: 'defaultNewItem',
  expression: function(of) {
    return of.create();
  }
}
```

TW80000 (approved):

PR #2694, opened 2019-10-15
CPF-3149 Prevent broken slots due to Javascript errors

KernelDeimos:
It is possible for exceptions related to setting a property (setter, adapt, etc) to be thrown at the time that a subslot updates a parent slot. This causes the two slots to be broken in one direction, resulting in confusing behaviour.


TW80000 (approved):
This makes sense to me.

PR #2695, opened 2019-10-15
Update cron scheduledTime on schedule changed

lchanmann:
## Changes

- Add CronScheduleDAO to cronDAO

### DIFF
--- a/src/foam/nanos/cron/CronScheduleDAO.js
+++ b/src/foam/nanos/cron/CronScheduleDAO.js
@@ -1,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.cron',
+  name: 'CronScheduleDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  documentation: 'Update cron scheduledTime on schedule changed',
+
+  javaImports: [
+    'foam.nanos.cron.Cron',
+    'foam.util.SafetyUtil'
+  ],
+
+  methods: [
+    {
+      name: 'put_',
+      javaCode: `
+        Cron newCron = (Cron) obj;
+        Cron oldCron = (Cron) getDelegate().find_(x, obj);
+
+        if ( oldCron == null
+          || ! SafetyUtil.equals(oldCron.getSchedule(), newCron.getSchedule())
+        ) {
+          newCron.setScheduledTime(newCron.getNextScheduledTime());
+        }
+        return getDelegate().put_(x, obj);
+      `
+    }
+  ]
+});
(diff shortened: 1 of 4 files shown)


src/foam/nanos/cron/CronScheduleDAO.js
```diff
@@ -0,0 +1,35 @@
...
+
+  methods: [
+    {
+      name: 'put_',
+      javaCode: `
+        Cron newCron = (Cron) obj;
+        Cron oldCron = (Cron) getDelegate().find_(x, obj);
+
+        if ( oldCron == null
+          || ! SafetyUtil.equals(oldCron.getSchedule(), newCron.getSchedule())
```
KernelDeimos, line 27: Should this be `! SafetyUtil.equals(oldCron.getScheduledTime(), newCron.getScheduledTime()` instead?

lchanmann: No, I want to re-calculate the `scheduledTime` only when the cron.schedule is changed.

KernelDeimos: After a quick discussion this makes sense - the scheduledTime value was not being updated after a schedule is changed. This means that if the old schedule's next scheduled time was a date further in the future than the new schedule, CronScheduler.java would not update scheduledTime until the previously calculated time.

KernelDeimos:
After discussing this with Troy it sounds like there may be an even simpler solution. The `scheduledTime` property could have a getter that always returns `getNextScheduledTime()`. The caveat with this approach is that `scheduledTime` stored in the journals can have an incorrect value at the record before the schedule is changed. (edit: which currently is the case anyway, except we're also using that incorrect value)

@jlhughes do we need to store `scheduledTime` in journals? I think it can always be calculated by a human based on the record and the commit hash at run time. (also this just gave me a cool idea related to single-journal; have a foamCommitDAO store the current commit hash at boot)

jlhughes:
re: storing scheduledTime - I'm good with not storing anything that can be calculated.  and if we calculated on load/replay then no need to store.  
re: for human reading - just needed in the UI.  Do we think this is necessary via DIG?

lchanmann:
scheduledTime is re-calculated on script run then gets stored in the journal so that we don't have to re-calculate the scheduledTime for all cronjobs every time.

KernelDeimos:
Oh that's a good point. We would need to calculate the scheduled time for all cronjobs every second. @TW80000 what do you think?

Edit: just re-read Joel's comment about recalculating on replay. Do we already have a way to do this? (I think a factory works if the property is set not to be stored)

TW80000:
> Oh that's a good point. We would need to calculate the scheduled time for all cronjobs every second. @TW80000 what do you think?

That makes sense. I think this PR makes sense in light of @lchanmann's comment. It's more performant just to update when the schedule changes, and since Crons are read so frequently, we should take all the performance improvements we can get.

Like @jlhughes points out, it is indeed a space vs time tradeoff. Calculating it every time is more costly than storing it in terms of CPU, but storing it takes more memory than not storing it. Although presumably if you're going to be calculating the value then you're going to be recreating it in memory every second anyway, so there's really not much memory to be saved.

As for seeing it on the client, since the implementations of `Schedule.getNextScheduledTime()` are  in Java, we can't calculate them on the client, so we need to calculate it on the server and store it if we want to see it on the client anyway.

TW80000 (approved):

jlhughes (approved):

PR #2696, opened 2019-10-15
added validator to easydao

michaelconnor468:


src/foam/dao/EasyDAO.js
```diff
@@ -166,6 +165,10 @@ if ( getDecorator() != null ) {
   delegate = (ProxyDAO) getDecorator();
 }
 
+if ( getValidated() ) {
+  delegate = getValidator();
```
jlhughes: this doesn't make sense, the delegate is simply replaced, what happened to what the delegate previously delegated to?

jlhughes (approved):

jlhughes (approved):

PR #2697, opened 2019-10-15
Invalid x refactor

noodlemoodle:


src/foam/core/InvalidX.java
```diff
@@ -6,29 +6,43 @@
...
+    logger_.warning("Unsafe access to " + nspecName_ + ". Please use .inX() instead.");
+
+    return ret;
   }
 
+  @Override
   public X put(Object key, Object value) {
-    throw exception_;
-  }
+    setX(getX().put(key, value));
```
TW80000: Rather than copy this code from `ProxyX`, we should just change the line below from

```Java
  return (X) this;
```

to

```Java
  return super.put(key, value);
```

src/foam/core/InvalidX.java
```diff
@@ -6,29 +6,43 @@
...
+    super(delegate);
+    delegate_ = delegate;
+    nspecName_ = nspecName;
+    logger_ = logger;
   }
 
+  @Override
   public Object get(X x, Object key) {
-    throw exception_;
+    Object ret = getX().get(x, key);
```
TW80000: Rather than copy this code from `ProxyX`, we should just change line 30 below from

```Java
  return ret;
```

to

```Java
  return super.get(x, key);
```

src/foam/core/InvalidX.java
```diff
@@ -6,29 +6,43 @@
...
-    throw exception_;
+    return (X) this;
   }
 
   @Override
-  public String toString() {
-    throw exception_;
+  public X putFactory(Object key, XFactory factory) {
+    logger_.warning("Unsafe access to " + nspecName_ + ". Please use .inX() instead.");
+    setX(getX().putFactory(key, factory));
```
TW80000: Same as above.

src/foam/dao/EasyDAO.js
```diff
@@ -220,7 +220,12 @@ if ( getOrder() != null &&
   }
 }
 
-if( getAuthorize() ) delegate = new foam.nanos.auth.AuthorizationDAO(getX(), delegate, getAuthorizer());
+if( getAuthorize() ) {
```
```suggestion
if ( getAuthorize() ) {
```

src/foam/nanos/boot/NSpecFactory.java
```diff
@@ -45,10 +45,11 @@ public synchronized Object create(X x) {
 
     try {
       if ( logger != null ) logger.info("Creating Service", spec_.getName());
+      X serviceX = spec_.getName().endsWith("DAO") ? new InvalidX(x_.getX(), spec_.getName(), logger) : x_.getX();
```
TW80000: Now that I'm reviewing this, it would be better to move this below line 50 and change it like so:

```suggestion
      X serviceX = ns instanceof DAO ? new InvalidX(x_.getX(), spec_.getName(), logger) : x_.getX();
```

TW80000 (changes requested):

TW80000 (approved):

PR #2699, opened 2019-10-16
CPF-3169 Adding PasswordPolicy

nanoscott:
- Add PasswordPolicy to Group
- UpdatePassword method cleanup


nanoscott:
Requires: https://github.com/nanoPayinc/NANOPAY/pull/8148

nanoscott:
Note that test cases still need to be added for the PasswordPolicy class.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.auth.ruler',
+  name: 'AddPasswordHistoryAction',
+  documentation: 'Adds an entry into the status history of a transaction',
```
jlhughes: documentation almost matches.

jlhughes (commented):
Looks good so far. Haven't tested.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
nanoscott: done.

src/foam/nanos/auth/Group.js
```diff
@@ -98,6 +98,18 @@ foam.CLASS({
...
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.nanos.auth.PasswordPolicy',
+      name: 'passwordPolicy',
+      view: { class: 'foam.u2.view.FObjectView' },
+      documentation: 'Password policy for this group.',
+      javaFactory: `
+        foam.nanos.auth.PasswordPolicy passwordPolicy = new foam.nanos.auth.PasswordPolicy();
+        passwordPolicy.setEnabled(false); // disabled by default
```
TW80000: Wouldn't it be more appropriate to leave this up to the `enabled` property to default to false?

src/foam/nanos/auth/HistoricPassword.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'HistoricPassword',
+
+  documentation: 'Historic hashed password value for a user.',
+
+  javaImports: [
+    'java.util.Date'
```
TW80000: Is this used?

src/foam/nanos/auth/HistoricPassword.js
```diff
@@ -0,0 +1,30 @@
...
+
+  properties: [
+    {
+      class: 'DateTime',
+      name: 'timeStamp',
+      visibility: 'RO',
+      documentation: 'Time at which password entry was created'
+    },
+    {
+      class: 'String',
```
TW80000:
```suggestion
      class: 'Password',
```

src/foam/nanos/auth/HistoricPassword.js
```diff
@@ -0,0 +1,30 @@
...
+
+  javaImports: [
+    'java.util.Date'
+  ],
+
+  properties: [
+    {
+      class: 'DateTime',
+      name: 'timeStamp',
+      visibility: 'RO',
```
TW80000: `visibility` is deprecated. Please use `createMode`, `readMode`, and `updateMode` instead. In this case, you'd just want `updateMode: 'RO'`.

src/foam/nanos/auth/HistoricPassword.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
TW80000: 2019

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
TW80000: 2019

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+        }
+      ],
+      javaCode: `
+        // check if this policy is enabled
+        if ( ! this.getEnabled() ) {
+          return; // password is valid by default
+        }
+
+        // check minimum length
+        int minLength = (int)Math.max(this.getMinLength(), MIN_PASSWORD_LENGTH);
```
TW80000:
```suggestion
        int minLength = (int) Math.max(this.getMinLength(), MIN_PASSWORD_LENGTH);
```

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+        String minLengthRegex = "^.{" + minLength + ",}$";
+        if ( SafetyUtil.isEmpty(potentialPassword) || ! (Pattern.compile(minLengthRegex)).matcher(potentialPassword).matches() ) {
+          throw new RuntimeException("Password must be at least " + minLength + " characters long.");
+        }
+
+        // check password history
+        if ( this.getPreventHistoricPasswordCount() > 0 && user != null)
+        {
```
TW80000: Should be on previous line

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+          HistoricPassword[] historicPasswords = user.getPasswordHistory();
+          int maxCount = Math.min((int)this.getPreventHistoricPasswordCount(), historicPasswords.length);
+          for ( int i = 0; i < maxCount; i++ )
+          {
```
TW80000: Should be on previous line

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+
+        // check minimum length
+        int minLength = (int)Math.max(this.getMinLength(), MIN_PASSWORD_LENGTH);
+        String minLengthRegex = "^.{" + minLength + ",}$";
+        if ( SafetyUtil.isEmpty(potentialPassword) || ! (Pattern.compile(minLengthRegex)).matcher(potentialPassword).matches() ) {
+          throw new RuntimeException("Password must be at least " + minLength + " characters long.");
+        }
+
+        // check password history
+        if ( this.getPreventHistoricPasswordCount() > 0 && user != null)
```
TW80000:
```suggestion
        if ( this.getPreventHistoricPasswordCount() > 0 && user != null )
```

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+        {
+          HistoricPassword[] historicPasswords = user.getPasswordHistory();
+          int maxCount = Math.min((int)this.getPreventHistoricPasswordCount(), historicPasswords.length);
```
TW80000:
```suggestion
          int maxCount = Math.min((int) this.getPreventHistoricPasswordCount(), historicPasswords.length);
```

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+    'java.util.regex.Pattern'
+  ],
+
+  methods: [
+    {
+      name: 'validate',
+      async: true,
+      args: [
+        {
+          name: 'x',
```
TW80000: Is this used? It doesn't look like it. If not, we should remove it.

src/foam/nanos/auth/User.js
```diff
@@ -32,7 +33,8 @@ foam.CLASS({
     'foam.mlang.predicate.Predicate',
     'foam.nanos.auth.AuthService',
     'foam.util.SafetyUtil',
-    'static foam.mlang.MLang.EQ'
+    'static foam.mlang.MLang.EQ',
```
TW80000: Sort

src/foam/nanos/auth/User.js
```diff
@@ -19,7 +19,8 @@ foam.CLASS({
 
   requires: [
     'foam.nanos.auth.Address',
-    'foam.nanos.auth.Phone'
+    'foam.nanos.auth.Phone',
+    'foam.nanos.auth.HistoricPassword'
```
TW80000: Sort

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+        Setting preventHistoricalPasswordCount to a value greater than zero will enable a check on that many previous password entries
+        when resetting the users passwords to prevent them from using the same password again.
+        `
+    },
+  ],
+
+  javaImports: [
+    'foam.nanos.auth.User',
+    'foam.util.Password',
+    'foam.util.SafetyUtil',
```
TW80000: Sort

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+    {
+      class: 'Long',
+      name: 'minLength',
+      documentation: 'Minimum length of the password.',
+      min: 6,
+      value: 6
+    },
+    {
+      class: 'Long',
+      name: 'preventHistoricPasswordCount',
```
TW80000: Name suggestion: `numberOfPriorPasswordsToCheck`

src/foam/nanos/auth/HistoricPassword.js
```diff
@@ -0,0 +1,30 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'HistoricPassword',
```
TW80000: Name suggestion: `PriorPassword`. Historic means "famous or important in history, or potentially so", whereas prior means "existing or coming before in time, order, or importance".

src/foam/nanos/auth/HistoricPassword.js
```diff
@@ -0,0 +1,30 @@
...
+  documentation: 'Historic hashed password value for a user.',
+
+  javaImports: [
+    'java.util.Date'
+  ],
+
+  properties: [
+    {
+      class: 'DateTime',
+      name: 'timeStamp',
```
TW80000: Could make this `created` and take advantage of `CreatedAware` and associated machinery.

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+      ],
+      javaCode: `
+        // check if this policy is enabled
+        if ( ! this.getEnabled() ) {
+          return; // password is valid by default
+        }
+
+        // check minimum length
+        int minLength = (int)Math.max(this.getMinLength(), MIN_PASSWORD_LENGTH);
+        String minLengthRegex = "^.{" + minLength + ",}$";
```
TW80000, line 82: Wouldn't it be simpler just to check `potentialPassword.length >= minLength`?

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,105 @@
...
+
+        // check password history
+        if ( this.getPreventHistoricPasswordCount() > 0 && user != null)
+        {
+          HistoricPassword[] historicPasswords = user.getPasswordHistory();
+          int maxCount = Math.min((int)this.getPreventHistoricPasswordCount(), historicPasswords.length);
+          for ( int i = 0; i < maxCount; i++ )
+          {
+            if ( Password.verify(potentialPassword, historicPasswords[historicPasswords.length - (1 + i) ].getPassword()) ) {
+              throw new RuntimeException("Password must be different than previous " + this.getPreventHistoricPasswordCount() + " passwords");
```
TW80000:
```suggestion
              throw new RuntimeException("Password must be different from previous " + this.getPreventHistoricPasswordCount() + " passwords.");
```

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -258,9 +248,36 @@ foam.CLASS({
...
+        if ( user != null ) {
+          Group ancestor = user.findGroup(x);
+          if ( ancestor != null ) {
+            throw new AuthenticationException("Group not found");
+          }
+
+          // Check password policy
+          passwordPolicy = ancestor.getPasswordPolicy();
+          while ( passwordPolicy == null || ! passwordPolicy.getEnabled() )
+          {
```
TW80000: Should be on the previous line. Same below.

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -258,9 +248,36 @@ foam.CLASS({
...
+            passwordPolicy = ancestor.getPasswordPolicy();
+          }
         }
+
+        // Use the default password policy if nothing is found
+        if ( passwordPolicy == null || ! passwordPolicy.getEnabled() )
+        {
+          passwordPolicy = new PasswordPolicy();
+          passwordPolicy.setEnabled(true);
+          passwordPolicy.setMinLength(PasswordPolicy.MIN_PASSWORD_LENGTH);
```
TW80000: Rather than do this, the `minLength` property could just default to that constant by definition.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
TW80000: 2019

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
```diff
@@ -0,0 +1,61 @@
...
+foam.CLASS({
+  package: 'foam.nanos.auth.ruler',
+  name: 'AddPasswordHistoryAction',
+  documentation: 'Adds an entry into the password history of a user',
+
+  implements: ['foam.nanos.ruler.RuleAction'],
+
+  javaImports: [
+    'foam.nanos.logger.Logger',
+    'foam.nanos.auth.HistoricPassword',
```
TW80000: Sort

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
```diff
@@ -0,0 +1,61 @@
...
+        User user = (User) obj;
+        HistoricPassword[] historicPasswordsOld = user.getPasswordHistory();
+        
+        // compute how large the new array should be and where it should start copying from
+        int newArrayLength = historicPasswordsOld.length + 1;
+        int startCopyingIndex = 0;
+        int copyLength = historicPasswordsOld.length;
+        
+        // only store up to the maximum number of password history records
+        if (historicPasswordsOld.length >= MAXIMUM_HISTORIC_PASSWORDS) {
```
TW80000:
```suggestion
        if ( historicPasswordsOld.length >= MAXIMUM_HISTORIC_PASSWORDS ) {
```

src/foam/nanos/auth/test/PasswordPolicyTest.js
```diff
@@ -0,0 +1,178 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
TW80000: 2019

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
TW80000: Shouldn't this be

```suggestion
        if ( newArrayLength >= MAXIMUM_HISTORIC_PASSWORDS ) {
```
?

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
```diff
@@ -0,0 +1,61 @@
...
+          newArrayLength = (int) MAXIMUM_HISTORIC_PASSWORDS;
+          startCopyingIndex = 1;
```
TW80000: I don't get this. Say `MAXIMUM_HISTORIC_PASSWORDS` was 10, someone had an array with 8 values in it and we change it `MAXIMUM_HISTORIC_PASSWORDS` to 5.

Why would we start copying from index 1? Wouldn't that skip the 1st value of the 8 and only take the next 4?

Shouldn't the place we start copying from be `historicPasswordsOld.length - MAXIMUM_HISTORIC_PASSWORDS`? So in this case `8 - 5 = 3` so we copy the passwords at indices 3, 4, 5, 6, and 7, which are the most recent 5 of the 8 available prior passwords? Then the 3 oldest passwords get dropped.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
```diff
@@ -0,0 +1,61 @@
...
+          copyLength = (int) MAXIMUM_HISTORIC_PASSWORDS - 1;
+        }
+
+        // copy the old array into a new array
+        HistoricPassword[] historicPasswordNew = new HistoricPassword[newArrayLength];
+        System.arraycopy(historicPasswordsOld, startCopyingIndex, historicPasswordNew, 0, copyLength);
+
+        // add the new entry to the end of the array
+        HistoricPassword historicPassword = new HistoricPassword();
+        historicPassword.setPassword(user.getPassword());
```
TW80000: For this to be the hashed value, this rule needs to executed _after_ `UserPasswordHashingDAO`, otherwise `user.getPassword()` will be "" and `user.getDesiredPassword()` will be the cleartext password.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
```diff
@@ -0,0 +1,61 @@
...
+      name: 'MAXIMUM_HISTORIC_PASSWORDS',
+      type: 'Long',
+      value: 10
+    }
+  ],
+
+  methods: [
+    {
+      name: 'applyAction',
+      javaCode: `
```
TW80000, line 34: Does this code need to be in the agency thing so it doesn't execute on trace commands?

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
```diff
@@ -0,0 +1,61 @@
...
+        User user = (User) obj;
```
TW80000:
```suggestion
        User user = (User) oldObj;
```

You're using the new user object here instead of the old one, and since this rule only applies on update (not create), it would be adding the new password to the array instead of the old one. Alternatively, you could change the operation of this rule to apply on updates and creates and then I think it would be correct.

src/foam/nanos/ruler/predicate/PropertyChangePredicate.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+  ],
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: `
+        FObject nu  = (FObject) NEW_OBJ.f(obj);
+        FObject old = (FObject) OLD_OBJ.f(obj);
+        return old != null && nu != null &&
```
TW80000, line 27: This predicate will only ever apply on object updates if you keep these null checks in, not on creates, where `old` would be null but `nu` wouldn't be. You could argue either way as to whether creating an object is a property update or not.

src/foam/nanos/ruler/predicate/PropertyChangePredicate.js
```diff
@@ -0,0 +1,34 @@
...
+          ! foam.util.SafetyUtil.equals(
```
TW80000, line 28: `PropertyInfo.diff()` would be more appropriate since we're diffing two properties, wouldn't it?

src/foam/nanos/auth/Group.js
nanoscott: Yes. I'll remove the extra lines.

src/foam/nanos/auth/HistoricPassword.js
nanoscott: Not used yet, but I was going to replace some of the existing functionality which stores the 'previous password' with this code and the date is necessary for that as well. I think it is interesting to keep around as well in case context is needed for when the given value was used.

src/foam/nanos/auth/HistoricPassword.js
nanoscott: This is not an entity which is going to be stored in a DAO, it is a nested object within User ..  so CreatedAware and associated machinery is not applicable.

src/foam/nanos/auth/HistoricPassword.js
nanoscott: done.

src/foam/nanos/auth/PasswordPolicy.js
nanoscott: @TW80000  isn't it already sorted? which one is out of order?

src/foam/nanos/auth/PasswordPolicy.js
nanoscott: good catch. i'll remove it.

src/foam/nanos/auth/PasswordPolicy.js
nanoscott: trying to preserve the old functionality .. this is how it was checked before .. i'm wondering if this regex prevents blank passwords as well as ensuring the length is what it saws it is. I would have to test that theory out though. i think the regex is also valid for nulls too.

src/foam/nanos/auth/UserAndGroupAuthService.js
nanoscott: done.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
nanoscott: yeah .. i thought of this too, but i need to be checking the old array length, not the new one. The idea here is that if the old array length is 9 (hence newArrayLength == 10), then i want to copy all 9 values (index 0-8), not 8 of them (index 1-8) .. so we need to check the old length.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
nanoscott: The idea is that this array does contain the current password as the latest entry in the array .. so we prevent re-use of the same password starting from the current one not starting from the previous one.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
nanoscott: I was only considering the normal course of operations where we would hit the boundary and only want to copy the 9 latest entries instead of the general case where someone has added more than the maximum to the password history. I'll update it to work for this more generic case though.

src/foam/nanos/ruler/predicate/PropertyChangePredicate.js
nanoscott: Yeah .. i think this is meant to run only on update. I'd argue that during a create operation nothing has 'changed' per se .. just that it is being set .. i.e. initial value.

src/foam/nanos/ruler/predicate/PropertyChangePredicate.js
nanoscott: SafetyUtil is more convient here because it returns a boolean directly. Is there a functional difference between the two? I can't find proper implementations fo the diff methods to see whether it uses the compareTo call the same way that the SafetyUtil.equals method does. If it is more correct in more cases, then we should switch it over .. but I'm not sure whether that is true. If you have examples of the implementation of the diff() method in java for different property types, send them over ...

nanoscott (commented):
First half of comments addressed .. several big ones still to go.

src/foam/nanos/auth/ruler/AddPasswordHistoryAction.js
nanoscott: The ruler is not used in the FOAM framework for the localUserDAO. I will update the downstream project though.

src/foam/nanos/auth/PasswordPolicy.js
TW80000: Oh my bad, I missed that the line below started with java, not foam.

src/foam/nanos/auth/PasswordPolicy.js
TW80000: The `SafetyUtil.isEmpty(potentialPassword)` you're doing anyway catches nulls, so it'll be safe to do the length check after that one.

It was probably a more complicated regex before. Now that it's simply a length check it's simpler just to check the length of the string directly.

src/services
```diff
@@ -168,7 +168,15 @@ p({
 })
 
 p({"class":"foam.nanos.boot.NSpec", "name":"cronScheduler",                    "lazy":false, "serviceClass":"foam.nanos.cron.CronScheduler" })
-p({"class":"foam.nanos.boot.NSpec", "name":"localUserDAO",                       "serviceScript":"return new foam.dao.PMDAO(x, new foam.dao.SequenceNumberDAO (1000, new foam.nanos.auth.PreventDuplicateEmailDAO.Builder(x).setDelegate(new foam.dao.java.JDAO(x, foam.nanos.auth.User.getOwnClassInfo(), \"users\")).build()));"})
+p({
+  "class":"foam.nanos.boot.NSpec", "name":"localUserDAO",
+  "serviceScript":"""
+    return 
+      new foam.dao.PMDAO(x, 
+      new foam.dao.SequenceNumberDAO (1000,
```
TW80000:
```suggestion
      new foam.dao.SequenceNumberDAO(1000, 
```

src/services
```diff
@@ -168,7 +168,15 @@ p({
 })
 
 p({"class":"foam.nanos.boot.NSpec", "name":"cronScheduler",                    "lazy":false, "serviceClass":"foam.nanos.cron.CronScheduler" })
-p({"class":"foam.nanos.boot.NSpec", "name":"localUserDAO",                       "serviceScript":"return new foam.dao.PMDAO(x, new foam.dao.SequenceNumberDAO (1000, new foam.nanos.auth.PreventDuplicateEmailDAO.Builder(x).setDelegate(new foam.dao.java.JDAO(x, foam.nanos.auth.User.getOwnClassInfo(), \"users\")).build()));"})
+p({
+  "class":"foam.nanos.boot.NSpec", "name":"localUserDAO",
+  "serviceScript":"""
```
TW80000:
```suggestion
  "serviceScript": """
```

src/services
```diff
@@ -168,7 +168,15 @@ p({
 })
 
 p({"class":"foam.nanos.boot.NSpec", "name":"cronScheduler",                    "lazy":false, "serviceClass":"foam.nanos.cron.CronScheduler" })
-p({"class":"foam.nanos.boot.NSpec", "name":"localUserDAO",                       "serviceScript":"return new foam.dao.PMDAO(x, new foam.dao.SequenceNumberDAO (1000, new foam.nanos.auth.PreventDuplicateEmailDAO.Builder(x).setDelegate(new foam.dao.java.JDAO(x, foam.nanos.auth.User.getOwnClassInfo(), \"users\")).build()));"})
+p({
+  "class":"foam.nanos.boot.NSpec", "name":"localUserDAO",
```
TW80000:
```suggestion
  "class": "foam.nanos.boot.NSpec",
  "name": "localUserDAO",
```

src/services
```diff
@@ -168,7 +168,15 @@ p({
...
+  "serviceScript":"""
+    return 
+      new foam.dao.PMDAO(x, 
+      new foam.dao.SequenceNumberDAO (1000, 
+      new foam.nanos.auth.PreventDuplicateEmailDAO.Builder(x).setDelegate(
+        new foam.dao.java.JDAO(x, foam.nanos.auth.User.getOwnClassInfo(), \"users\")).build()));
```
TW80000:
```suggestion
        new foam.dao.java.JDAO(x, foam.nanos.auth.User.getOwnClassInfo(), "users")
      ).build()));
```

src/services
```diff
@@ -168,7 +168,15 @@ p({
...
+  """})
```
TW80000:
```suggestion
  """
})
```

src/foam/nanos/auth/User.js
```diff
@@ -347,6 +349,21 @@ foam.CLASS({
...
+      javaFactory: `
+        foam.nanos.auth.PriorPassword[] priorPasswords = new foam.nanos.auth.PriorPassword[1];
+        priorPasswords[0] = new foam.nanos.auth.PriorPassword();
+        priorPasswords[0].setPassword(this.getPassword());
+        priorPasswords[0].setTimeStamp(new Date());
+        return priorPasswords;
+      `,
+      hidden: true,
+      networkTransient: true,
+      section: 'administrative'
```
TW80000: This way this property won't be editable in the GUI. We don't want people changing history.

```suggestion
      section: 'administrative',
      createMode: 'HIDDEN',
      updateMode: 'RO'
```

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -258,9 +248,33 @@ foam.CLASS({
...
       name: 'validatePassword',
       javaCode: `
-        if ( SafetyUtil.isEmpty(potentialPassword) || ! (Pattern.compile(PASSWORD_VALIDATE_REGEX)).matcher(potentialPassword).matches() ) {
-          throw new RuntimeException(PASSWORD_VALIDATION_ERROR_MESSAGE);
+        PasswordPolicy passwordPolicy = null;
+
+        // Retrieve the password policy from the user and group when available
+        if ( user != null ) {
+          Group ancestor = user.findGroup(x);
+          if ( ancestor != null ) {
```
KernelDeimos: When can this happen? Are users allowed to not have groups, or is that an error in the system? If the latter then IMO this should be a check inside `findGroup` itself

TW80000 (approved):
LGTM. Thanks for writing such comprehensive tests.

src/services
```diff
@@ -168,7 +168,15 @@ p({
...
 
 p({"class":"foam.nanos.boot.NSpec", "name":"cronScheduler",                    "lazy":false, "serviceClass":"foam.nanos.cron.CronScheduler" })
-p({"class":"foam.nanos.boot.NSpec", "name":"localUserDAO",                       "serviceScript":"return new foam.dao.PMDAO(x, new foam.dao.SequenceNumberDAO (1000, new foam.nanos.auth.PreventDuplicateEmailDAO.Builder(x).setDelegate(new foam.dao.java.JDAO(x, foam.nanos.auth.User.getOwnClassInfo(), \"users\")).build()));"})
+p({
+  "class":"foam.nanos.boot.NSpec", "name":"localUserDAO",
+  "serviceScript":"""
+    return 
+      new foam.dao.PMDAO(x, 
+      new foam.dao.SequenceNumberDAO (1000, 
+      new foam.nanos.auth.PreventDuplicateEmailDAO.Builder(x).setDelegate(
```
KernelDeimos, line 184: Would it be reasonable to add PreventDuplicateEmailDAO to EasyDAO? @TW80000

src/foam/nanos/auth/UserAndGroupAuthService.js
KernelDeimos: When does this happen? Is it expected than some users won't have a group or is that a system error? If it's the latter I think it could be checked inside findGroup and maybe throw an exception which indicates more criticality than AuthenticationException

src/services
TW80000: I would say no, since it's only applicable to models with an Email property, and we don't have an interface like `EmailAware` like we do for other properties, like `CreatedAware`. EasyDAO is generic and therefore shouldn't have options for non-generic decorators.

src/foam/nanos/auth/UserAndGroupAuthService.js
TW80000:
```suggestion
          if ( ancestor == null ) {
```

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -258,9 +248,33 @@ foam.CLASS({
...
+            throw new AuthenticationException("Group not found");
```
TW80000: I think this should be a `RuntimeException` instead. `AuthenticationException` means the user needs to authenticate (sign in) to do what they're trying to do. In this case the data is corrupted, so signing in won't change anything.

We should also add a logger statement here before throwing and record the user's id and their group.

src/foam/nanos/auth/UserAndGroupAuthService.js
TW80000: This would happen if a user's group was deleted, or a user was created with their `group` property set to something that isn't the id of a real group. In both cases though, it's not something we intend to happen.

nanoscott: This code ensures that any execution path that calls this code should work. So if a test case creates a user and doesn't associate that user to a group, then the test doesn't fail with a NPE. The real system should never be configured this way, but if it ever gets into a state where this happens then this method won't be the cause of the failure. It is best practice to ensure that when calling a method that it returns a non-null value .. so if there are other places that it is assumed to *always* exists, then yes, you should definitely fix the other code. That is not, however, in scope for this change.

nanoscott: Plus, findGroup is an autogenerated method, so it would be no easy feat to add code there ..

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,103 @@
...
+      max: 10,
+      documentation: `
+        User passwords are hashed and the hash is kept for historical record up to a maximum of 10 previous entries per user. 
+        Setting priorPasswordsToCheckCount to a value greater than zero will enable a check on that many previous password entries
+        when resetting the users passwords to prevent them from using the same password again.
+        `
+    },
+  ],
+
+  javaImports: [
```
kgrgreer: Put javaImports higher up around implements:.

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -0,0 +1,103 @@
...
+    'foam.nanos.auth.User',
+    'foam.util.Password',
+    'foam.util.SafetyUtil'
+  ],
+
+  methods: [
+    {
+      name: 'validate',
+      async: true,
```
kgrgreer: Why is this async?

nanoscott: I think it was a copy paste error. I'll remove it.

PR #2700, opened 2019-10-16
[DO_NOT_MERGE] [CPF-3036] Implement capability intercepts with new RemoteException message

KernelDeimos:
This PR adds logic to handle capability intercepts to the `foam.box` package. This prevents CPF-3061 (modelled exceptions) from blocking CPF-3036. Ideally, once modelled exceptions are completed, this code will be updated to remove CRUNCH-specific code from `foam.box`.


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -377,6 +378,19 @@ foam.CLASS({
       });
     },
 
+    function requestCapability(capabilityInfo) {
+      var self = this;
+      console.log("CAPABILITY SCREEN REQUESTED");
```
TW80000: Please remove.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -377,6 +378,19 @@ foam.CLASS({
...
+      return new Promise(function(resolve, reject) {
+        self.stack.push({
+          class: 'foam.nanos.crunch.ui.CapabilityInterceptView',
+          data: self.__subContext__["capabilityDAO"],
```
TW80000: I think we can change this to

```suggestion
          data: self.capabilityDAO,
```

if we add `'capabilityDAO'` to the imports.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -377,6 +378,19 @@ foam.CLASS({
...
+          capabilityOptions: capabilityInfo.capabilityOptions
+        });
+        self.loginSuccess$.sub(resolve);
```
TW80000: I don't think `loginSuccess` is what we want here, is it? Shouldn't we have `capabilityAcquired` or something like that, which keeps track of whether they've gotten the capability or not?

src/foam/nanos/crunch/ui/CapabilityInterceptView.js
```diff
@@ -0,0 +1,77 @@
...
+      overflow-x: scroll;
+    }
+  `,
+
+  methods: [
+    function initE() {
+      this
+        .addClass(this.myClass())
+
+        // Container for the actions
```
TW80000: We don't need the container since you're not including another group of buttons on the right, like the place this was copied from. Might as well change lines 44-54 to the following:

```suggestion
        .startContext({ data: this })
          .tag(this.CANCEL, { buttonStyle: 'SECONDARY' })
        .endContext()
```

src/foam/nanos/crunch/ui/CapabilityInterceptView.js
```diff
@@ -0,0 +1,77 @@
...
+            .startContext({ data: this })
+              .tag(this.CANCEL, { buttonStyle: 'SECONDARY' })
+            .endContext()
+          .end()
+        .end()
+
+        // Container for the detailview
+        .start()
+          .addClass(this.myClass('detail-container'))
+          // TODO:
```
TW80000: What still needs to be done here?

src/foam/nanos/auth/AuthorizationException.java
```diff
@@ -5,15 +5,32 @@
...
 package foam.nanos.auth;
+import foam.box.CapabilityRequiredRemoteException;
 
 /**
  * Throw this when a user is logged in, but is not permitted to do something.
  */
 public class AuthorizationException extends SecurityException {
+  // TODO: This should be decoupled from AuthorizationException after
+  //       the modelled exception feature is merged.
+  private CapabilityRequiredRemoteException capabilityRequired_;
```
TW80000: Wouldn't it be better if `AuthorizationException` had a `permission` property instead, that was just the id of a permission? Then you could just do something like

```Java
throw new AuthorizationException("You don't have permission to update that User.", "user.update.8046");
```

Then in the box code you could check if the exception has a permission set and if so, look up the set of capabilities that satisfy that permission and attach a list of those ids to the message.

So almost what you have here, but it makes it much easier for people to throw `AuthorizationExceptions` because the way it's currently written, they'd have to do something like the following, unless I'm mistaken:

```Java
CapabilityRequiredRemoteException crre = new CapabilityRequiredRemoteException.Builder(x)
  .setCapabilityOptions(new String[]{ "user.update.8046" })
  .setExplanation("Blah blah blah")
  .build();
throw new AuthorizationException("You don't have permission to update that User.", crre);
```

src/foam/box/RemoteException.js
```diff
@@ -30,3 +30,39 @@ foam.CLASS({
...
+        List of capabilities which can be used to satisfy the
+        permission that caused this error. A capability will
+        only intercept a permission if itself and all of its
+        implied capabilities or permissions can grant the
+        requested action.
+      `
+    },
+    {
+      class: 'String',
+      name: 'explanation',
```
TW80000: This should either be used or removed. If it's not used by this PR we can add it later if/when we need it.

src/foam/box/RemoteException.js
```diff
@@ -30,3 +30,39 @@ foam.CLASS({
...
+// TODO: Once modelled exceptions are merged this can be removed in
+//       favour of an exception modelled in foam.nanos.cron
+foam.CLASS({
+  package: 'foam.box',
+  name: 'CapabilityRequiredRemoteException',
+  implements: ['foam.core.Exception'],
+  properties: [
+    {
+      class: 'String',
+      name: 'id',
```
TW80000: Is this necessary if we're not storing these in a DAO?

src/foam/box/Message.js
```diff
@@ -59,6 +67,26 @@ foam.CLASS({
         Message replyMessage = new Message();
         replyMessage.setObject(reply);
 
+        replyBox.send(replyMessage);
+      `
+    },
+    {
+      name: 'replyWithCapabilityRequired',
```
TW80000: This doesn't strike me as desirable at first glance, although I'm not thoroughly convinced it's not. I'd consider this behaviour just a specific case of `replyWithException`, and would argue against changing the interface of message because of that. I think it would make more sense to inline this code above rather than extracting it to a method.

src/foam/nanos/auth/AuthorizationException.java
KernelDeimos: I was thinking of having the server box set capabilityRequired_ if applicable (based on if the permission is intercepted by any capabilities) so that someone throwing an authorization exception doesn't have to worry about capabilities - what do you think of this solution?

src/foam/box/RemoteException.js
KernelDeimos: Yes, but wow I totally forgot about that; this field should be renamed. It's actually for the exception id.

src/foam/box/Message.js
KernelDeimos: That's what I tried to do at first, but it significantly complicates things because I want to use a different class in place of RemoteException. I didn't want to spend too much time on it because I'm hoping to remove `CapabilityRequiredRemoteException` alltogether when I add modelled exceptions.

src/foam/nanos/controller/ApplicationController.js
KernelDeimos: Huh, I guess I copied that there by accident but I'm not sure what I could replace it with. `capabilityAquired` would be ambiguous because there's no indication of what capability was acquired. It looks like loginSuccess is needed for some behaviours of ApplicationController though so it may be sufficient to just remove it here.

src/foam/box/RemoteException.js
KernelDeimos: Oh, actually it needs to be called `id` because that's what SessionClientBox expects. It's really just the name of the exception, meant to match `id` property from `RemoteException`

src/foam/nanos/crunch/ui/CapabilityInterceptView.js
KernelDeimos: Updated; guess I forgot to type the comment. Right now this uses DAOBrowserView but in a future PR I want to replace that with an appropriate view for selecting capabilities

src/foam/nanos/controller/ApplicationController.js
TW80000: > It looks like loginSuccess is needed for some behaviours of ApplicationController though so it may be sufficient to just remove it here.

I agree that `loginSuccess` isn't required for this feature, but we _do_ need something that serves a similar purpose for this feature. The point of `loginSuccess` is to block progress until the user logs in, then resume it. In the same way, we need to block progress until the user either gets a capability or decides they don't want to continue with whatever they were trying to do.

src/foam/box/RemoteException.js
```diff
@@ -30,3 +30,31 @@ foam.CLASS({
...
+      // NOTE: Good candidate for a foam.script.Return
+      factory: () => { return "foam.box.CapabilityRequiredRemoteException" },
```
TW80000: These two lines can be replaced with `value: 'foam.box.CapabilityRequiredRemoteException'`.

And the whole property can be replaced with: `[ 'id', 'foam.box.CapabilityRequiredRemoteException' ]`

src/foam/box/RemoteException.js
TW80000: Is there a reason this doesn't extend `RemoteException`? It seems to me that it should, especially considering it has `RemoteException` in the name.

src/foam/box/RemoteException.js
```diff
@@ -30,3 +30,31 @@ foam.CLASS({
...
+      javaFactory: `return "foam.box.CapabilityRequiredRemoteException";`
+    },
+    {
+      class: 'StringArray',
+      name: 'capabilityOptions',
+      documentation: `
+        List of capabilities which can be used to satisfy the
```
TW80000: Text should wrap at line 80 according to the style guide.

src/foam/nanos/auth/AuthorizationException.java
TW80000: Yeah, that's what I meant by

> Then in the box code you could check if the exception has a permission set and if so, look up the set of capabilities that satisfy that permission and attach a list of those ids to the message.

but I guess it was clearer in my head, haha.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -52,7 +52,8 @@ foam.CLASS({
   imports: [
     'installCSS',
     'sessionSuccess',
-    'window'
+    'window',
+    'capabilityDAO'
```
TW80000: Sort

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -69,6 +70,7 @@ foam.CLASS({
     'notify',
     'pushMenu',
     'requestLogin',
+    'requestCapability',
```
TW80000, line 74: Sort

src/foam/nanos/auth/AuthorizationException.java
KernelDeimos: Oh, I actually mis-remembered the way I did something. Yeah I'll for sure have to remove that property from AuthorizationException and put one for permission there instead

src/foam/nanos/crunch/ui/CapabilityInterceptView.js
```diff
@@ -0,0 +1,77 @@
+foam.CLASS({
```
kgrgreer, line 7: Missing copyright header.

src/foam/box/Message.js
```diff
@@ -49,6 +49,14 @@ foam.CLASS({
 
         if ( replyBox == null ) return;
 
+        if ( t.getClass().getName() == "foam.nanos.auth.AuthorizationException" ) {
```
kgrgreer: Could you just do t.getClass() == foma.nanos.auth.AuthorizationException.class?

src/foam/nanos/controller/ApplicationController.js
KernelDeimos: Ah, the `.sub(resolve)` ties loginSuccess to the promise. My understanding now is that we could use just one property for capabilityAquired and reset it to false at the beginning of every requestCapability call. (this property would be a signal rather than meaningful information - is there a different way to do that?)

TW80000: > My understanding now is that we could use just one property for capabilityAquired and reset it to false at the beginning of every requestCapability call.

Yeah, that's right.

> this property would be a signal rather than meaningful information - is there a different way to do that?

I'd argue that the signal is meaningful.

src/foam/box/RemoteException.js
KernelDeimos: Turns out I had to go with `value` because `[ name, value ]` properties seem to be ignored by genJava.

src/foam/box/CapabilityInterceptBox.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
TW80000: 2019

src/foam/box/CapabilityInterceptBox.js
```diff
@@ -0,0 +1,105 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.box',
+  name: "CapabilityInterceptReplyDecoratorBox",
```
TW80000: Should use single quotes.

src/foam/box/CapabilityInterceptBox.js
```diff
@@ -0,0 +1,105 @@
...
+          new CapabilityInterceptBox.Builder(x)
+            .setDelegate((Box) msg.getAttributes().get("replyBox"))
+            .build());
+        getDelegate().send(msg);
+      `
+    }
+  ]
+});
+
+foam.CLASS({
```
TW80000: Please put each class in its own file. Makes it easier to find it that way.

src/foam/box/CapabilityInterceptBox.js
```diff
@@ -0,0 +1,105 @@
...
+  package: 'foam.box',
+  name: 'CapabilityInterceptBox',
+  extends: 'foam.box.ProxyBox',
+
+  javaImports: [
+    'foam.box.RPCErrorMessage',
+    'foam.box.CapabilityRequiredRemoteException',
```
TW80000, line 13: Sort

src/foam/box/CapabilityInterceptBox.js
```diff
@@ -0,0 +1,105 @@
...
+          // Remove permission name from AuthorizationException
+          // (it will still be sent if no capabilities were found)
+          authorEx.clearPermission();
+
+          // Get services needed to detect capability intercepts
+          User user = (User) x.get("user");
+          DAO capabilityDAO = (DAO) x.get("capabilityDAO");
+
+          List<String> capabilityOptions = new ArrayList<String>();
+          {
```
TW80000: What is this for? Is this a syntax error or some language feature I'm not aware of?

src/foam/box/CapabilityInterceptBox.js
```diff
@@ -0,0 +1,105 @@
...
+    'static foam.mlang.MLang.*'
+  ],
+
+  methods: [
+    {
+      name: 'send',
+      javaCode: `
+        foam.core.X x = getX();
+
+        FindCapabilityIntercept: {
```
TW80000: Is this necessary? Couldn't we just change everywhere that says

```Java
break FindCapabilityIntercept;
```

below to 

```Java
return super.send(msg);
```

? That would be preferable in my opinion.

src/foam/box/Message.js
```diff
@@ -52,6 +52,7 @@ foam.CLASS({
         RemoteException wrapper = new RemoteException();
         wrapper.setId(t.getClass().getName());
         wrapper.setMessage(t.getMessage());
+        wrapper.setServerObject(t);
```
TW80000: Rather than store the exception on a new property on `RemoteException` so you can get it later and get the permission string being checked from it, I think it would be simpler to store the permission string in `Message.localAttributes` and retrieve it from there in `CapabilityInterceptBox`. Then we don't need the `serverObject` property.

src/foam/box/RemoteException.js
```diff
@@ -30,3 +34,28 @@ foam.CLASS({
     }
   ]
 });
+
+// TODO: Once modelled exceptions are merged this can be removed in
+//       favour of an exception modelled in foam.nanos.cron
+foam.CLASS({
```
TW80000: Please put this in its own file.

src/foam/nanos/auth/AuthorizationException.java
```diff
@@ -5,15 +5,32 @@
...
 public class AuthorizationException extends SecurityException {
+  // TODO: This should be decoupled from AuthorizationException after
+  //       the modelled exception feature is merged.
+  protected String permission_;
+
   public AuthorizationException() {
     super("Permission denied.");
   }
+  public AuthorizationException(String message, String permission) {
+    super(message + " (permission: '" + permission + "')");
```
TW80000: I don't think we should include the permission in the message since the end user will be seeing the message sometimes.

src/foam/nanos/crunch/ui/CapabilityInterceptView.js
```diff
@@ -0,0 +1,75 @@
...
+
+        .startContext({ data: this })
+          .tag(this.CANCEL, { buttonStyle: 'SECONDARY' })
+        .endContext()
+
+        // Container for the detailview
+        .start()
+          .addClass(this.myClass('detail-container'))
+          // TODO: Replace with Capability selection view appropriate for any user
+          .tag(foam.comics.BrowserView.create({
```
TW80000: It's preferable to add models to the `requires` list and then use `this.<model name>` instead. Eg: `this.BrowserView`. Please do that for BrowserView and SearchMode.

src/foam/nanos/crunch/ui/CapabilityInterceptView.js
```diff
@@ -0,0 +1,75 @@
...
+  requires: [
+    'foam.nanos.crunch.Capability'
+  ],
+
+  imports: [
+    'stack'
+  ],
+
+  properties: [
+    { name: 'data' },
```
TW80000: If just providing a name, you can just use a string directly.

```suggestion
    'data',
```

src/foam/nanos/crunch/ui/CapabilityInterceptView.js
```diff
@@ -0,0 +1,75 @@
...
+    { name: 'capabilityOptions', class: 'StringArray' }
```
TW80000:
```suggestion
    {
      class: 'StringArray',
      name: 'capabilityOptions'
    }
```

src/foam/nanos/crunch/ui/CapabilityInterceptView.js
```diff
@@ -0,0 +1,75 @@
...
+  ],
+
+  css: `
+    ^ {
+      width: 1024px;
+      margin: auto;
+    }
+    ^action-container {
```
TW80000: Please remove the styles for `action-container` since it has been removed.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -377,6 +379,17 @@ foam.CLASS({
       });
     },
 
+    function requestCapability(capabilityInfo) {
+      var self = this;
+      return new Promise(function(resolve, reject) {
```
TW80000, line 396: When does `resolve` get called? Don't we need something like `loginSuccess` for that?

src/foam/nanos/crunch/Capability.js
```diff
@@ -101,6 +101,11 @@ foam.CLASS({
       class: 'StringArray',
       documentation: `List of permissions granted by this capability`
     },
+    {
+      name: 'permissionsIntercepted',
+      class: 'StringArray',
```
TW80000, line 106: This works and it's consistent with `permissionsGranted`, but something makes me think that they should both be many-to-many relationships, not string arrays. Not in the scope of this PR though.

src/foam/nanos/controller/ApplicationController.js
KernelDeimos: Right now it doesn't. I'm thinking `resolve` and `reject` should be passed to CapabilityInterceptView. If you agree then I'll add the properties to CapabilityInterceptView in this PR

src/foam/nanos/crunch/Capability.js
KernelDeimos: That makes sense. Right now searching a capability that intercepts a permission requires selecting all capabilities and iterating over every permissionsIntercepted list.

src/foam/nanos/controller/ApplicationController.js
TW80000: That could work.

src/foam/nanos/crunch/Capability.js
TW80000: Yeah. There are pros and cons to each approach. Groups used to have string arrays of permissions they implied though and they can be annoying to work with in a lot of different ways. This PR is big enough as is though so let's leave this as a string array for now.

src/foam/box/CapabilityInterceptBox.js
KernelDeimos: That's a block statement - it follows the rules of a block without having a loop or condition, so it's equivalent to `if (true) { ... }` but useful to reduce the scope of variables.

I like to use them when I'm creating variables that are only used for the purpose of populating another variable, (in this case `capabilities` is only used to populate `capabilityOptions`) but I admit it's not very commonly used.

src/foam/box/CapabilityInterceptBox.js
KernelDeimos: It would have to be two lines because Java doesn't allow returning void types for some reason. This would also make it easier to modify that behaviour later.

Would `super.send` be preferred over `getDelegate().send`?

src/foam/box/Message.js
KernelDeimos: That makes sense, implementing this now

src/foam/box/CapabilityInterceptBox.js
KernelDeimos: Edit: after changes related to your other comment there won't be as many repetitions of this, so the labelled block isn't useful anymore

src/foam/box/Message.js
KernelDeimos: Passing the Throwable instead of permission string, because the latter requires special handling of AuthorizationException inside Message.replyWithException. This still makes CapabilityInterceptBox look a lot cleaner though

TW80000:
@kgrgreer The code here looks good to me and the ideas make sense to me. I haven't tested it, but under the assumption it works, this PR gets a thumbs up from me. I've asked Eric to demo it to you in person to show it in action. If you could check it out and review this PR when you have a chance, that would be appreciated, thanks.

KernelDeimos:
Wow, 4-months-ago me really wasn't thinking properly.

I added a `permission` property to AuthorizationException because there was no way to know what permission failed, then I created CapabilityRequiredRemoteException so I could send a list of capabilities from the server because I didn't want to send permission strings to the client .

Really I can just add a `capabilities` property to AuthorizationException. The CapabilityAuthService can check the permission and then if it can't pass it and its delegate can't pass it then do a final query to populate the `capabilities` property.

So if I change this, I can get rid of CapabilityRequiredRemoteException, CapabilityInterceptBox, and CapabilityInterceptReplyDecoratorBox, as well as avoid changing the constructor args of AuthorizationException everywhere which is apparently something I thought was okay at the time.

Adding DO_NOT_MERGE until I fix this horrible PR I made

PR #2707, opened 2019-10-18
Fix bug where ModeAltView was clobbering context data

TW80000:
This seemed like a better solution than extending `foam.u2.Element` instead since doing so would have required us to copy over all of the methods and properties of `foam.u2.View`.

### DIFF
--- a/src/foam/u2/view/ModeAltView.js
+++ b/src/foam/u2/view/ModeAltView.js
@@ -15,6 +15,14 @@ foam.CLASS({
     'foam.u2.DisplayMode'
   ],
 
+  exports: [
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/ModeAltView.js
```diff
@@ -24,7 +32,14 @@ foam.CLASS({
       class: 'foam.u2.ViewSpec',
       name: 'writeView',
     },
-    'prop'
+    'prop',
+    {
+      name: 'contextData',
+      documentation: "See the comment in 'exports' above as to why this is necessary.",
```
Nauna, line 38: Troy wow you of all people ... double quotes? LOL

PR #2708, opened 2019-10-18
Prerequisite capability fix

noodlemoodle:
Fix issue where capabilitycapabilityjunctions where actually being used incorrectly since I interpreted the meaning of inverseName and forwardName wrong for relationships at the time I wrote logic using these junctions.
and update tests to reflect changes.


src/foam/nanos/crunch/Capability.js
```diff
@@ -120,7 +120,7 @@ foam.CLASS({
       documentation: `Checks if a permission or capability string is implied by the current capability`,
       javaCode: `
         if ( ! this.getEnabled() ) return false;
-
+
```
kgrgreer: Remove trailing whitespace.

PR #2709, opened 2019-10-18
DO_NOT_MERGE Unsafe xdao

noodlemoodle:
This pr is meant to log a warning when developers do not use context-oriented methods when reading/writing from a dao. 
However this is still failing 2 tests, and in practice there are too many warnings logged to merge this.
Will come back to it when I have some extra time but it is low priority


michaelconnor468:
@googlebot I consent

PR #2711, opened 2019-10-22
Tooltip timer

KateMengjunXia:
Display tooltip when cursor stays at the same position for half a second.
Tooltip stays the same when cursor moves inside the target element.


src/foam/u2/Tooltip.js
```diff
@@ -46,8 +43,65 @@ foam.CLASS({
...
+              'z-index': '2000',
+              'max-width': (screenWidth / 4)+'px',
+              'top': top,
+              'left': left
+            });
+          }
+        }, 500);
+      });
+      
+      self.target.on('mousedown', function() {
```
kgrgreer: Better to make this function be a listener rather than repeating 4 times.

KateMengjunXia: Done

PR #2712, opened 2019-10-22
Improve Log-Related UI

TW80000:
## Changes

* Remove lastModifiedBy on LogMessage
  * It's always set to 1, the id of the system user, so it's not
useful.
* Update view modes and table columns widths for LogMessage
* Let actions have views
* Improve buttons in EditColumnsView
  * Changed the order of the buttons in EditColumnsView to put cancel on the left
  * Require confirmation on "Reset All", which adds a red background and requires the user to click twice instead of once
  * Change "Cancel" to secondary button style

### Before

Screen Shot 2019-10-22 at 9 42 47 AM
Screen Shot 2019-10-22 at 9 42 58 AM

### After

Screen Shot 2019-10-22 at 1 32 10 PM
Screen Shot 2019-10-22 at 1 32 58 PM
Screen Shot 2019-10-22 at 1 39 33 PM

### DIFF
--- a/src/foam/nanos/logger/LogMessage.js
+++ b/src/foam/nanos/logger/LogMessage.js
@@ -8,20 +8,15 @@ foam.CLASS({
   package: 'foam.nanos.logger',
   name: 'LogMessage',
 
-  documentation: `Modelled log output.
-Implement LastModifiedByAware to suppress 'modified by' comment in journal output.`,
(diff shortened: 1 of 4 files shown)


kgrgreer:
createdBy should be set, if it's always 1, then I think this is a bug.

PR #2713, opened 2019-10-22
[CPF-3206] Remove ANSI colour escapes and control characters prior to JSON output

KernelDeimos:
Outputting ANSI colour escapes through the logger causes invalid JSON to be sent when logs are requested by the browser. Additionally, this reveals that control characters are not being removed from strings when they're serialized.

This PR removes ANSI colour escapes, then remaining control characters, prior to calling the existing escape functionality.

### DIFF
--- a/src/foam/lib/json/Outputter.java
+++ b/src/foam/lib/json/Outputter.java
@@ -12,6 +12,7 @@ import foam.core.FObject;
 import foam.core.PropertyInfo;
 import foam.dao.AbstractSink;
 import foam.lib.PermissionedPropertyPredicate;
+import foam.lib.PropertyPredicate;
 import foam.util.SafetyUtil;
 import java.io.*;
 import java.lang.reflect.Array;
--- a/src/foam/lib/json/Outputter.java
+++ b/src/foam/lib/json/Outputter.java
@@ -11,17 +11,17 @@ import foam.core.Detachable;
 import foam.core.FObject;
 import foam.core.PropertyInfo;
 import foam.dao.AbstractSink;
-import foam.lib.PropertyPredicate;
 import foam.lib.PermissionedPropertyPredicate;
+import foam.lib.PropertyPredicate;
 import foam.util.SafetyUtil;
 import java.io.*;
 import java.lang.reflect.Array;
 import java.text.SimpleDateFormat;
+import java.util.*;
 import java.util.Iterator;
 import java.util.List;
-import java.util.*;
-
 import org.apache.commons.io.IOUtils;
+
 
 public class Outputter
   extends AbstractSink
@@ -66,6 +66,7 @@ public class Outputter
     this.writer_ = writer;
   }
 
+  @Override
   public String stringify(FObject obj) {
     initWriter();
     outputFObject(obj);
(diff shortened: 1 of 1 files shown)


TW80000 (changes requested):
This fixes the issue of ANSI colour codes breaking the Logs table on the web client, but it also has the side effect of stripping all newlines from scripts, tests, and cron jobs. I think we need to be more selective with the control characters that we remove.

KernelDeimos:
Oh shoot, I forgot that whitespace characters exist there too - thanks!

KernelDeimos:
Fixed - control character are now removed after whitespace replacements

src/foam/lib/json/Outputter.java
```diff
@@ -111,15 +111,53 @@ public void outputString(String s) {
...
   public String escapeMultiline(String s) {
-    return s.replace("\\", "\\\\");
+    s = removeColour(s);
+    s = s.replace("\\", "\\\\");
+    s = removeControlCharacters(s);
+    return s;
+  }
+
+  // TODO: This does not cover all possible ANSI escape codes yet
+  public String removeColour(String s) {
```
kgrgreer: I don't think we should remove colours. I should be able to output any valid string to JSON. We should escape codes like \u2605 instead of just stripping.
See: https://www.crockford.com/mckeeman.html

KernelDeimos: That makes sense, I can see how ANSI escapes shouldn't get special treatment here. That brings two questions to mind:

- Should logDAO allow ANSI escapes? (I vote yes because that opens cool possibilities)
- Should foam.core.String have meta properties to advise users (i.e. views) how to render and process them? I'm thinking like `{ class: 'String', name: 'message', meta: { ansicolors: true }`
  - Alternatively, maybe 'StringANSI' is an alias of 'String' which views can treat differently

KernelDeimos:
Something in vscode changed the imports around :/ I thought I disabled that

src/foam/lib/json/Outputter.java
```diff
@@ -111,15 +111,42 @@ public void outputString(String s) {
...
+    s = s.replace("\\", "\\\\");
+    s = escapeControlCharacters(s);
+    return s;
+  }
+
+  public String escapeControlCharacters(String s) {
+    int lastStart = 0;
+    String escapedString = "";
+    for ( int i=0; i < s.length(); i++ ) {
+      if ( s.charAt(i) >= ' ' ) continue;
```
kgrgreer: Better do var c = s.charAt(i) and reuse for this line and next since it creates a new object which needs to be garbage collected.

src/foam/lib/json/Outputter.java
```diff
@@ -111,15 +111,42 @@ public void outputString(String s) {
...
+      char c = s.charAt(i);
+      // Character to hex
+      char right = (char) (c & 0x0F);
+      char left = (char) ((c & 0xF0) >> 4);
+      right += '0'; if ( right > '9' ) right += 'A' - '9' - 1;
```
kgrgreer: Messy to combine lines like this.

src/foam/lib/json/Outputter.java
```diff
@@ -111,15 +111,42 @@ public void outputString(String s) {
...
-    return s.replace("\\", "\\\\");
+    s = s.replace("\\", "\\\\");
+    s = escapeControlCharacters(s);
+    return s;
+  }
+
+  public String escapeControlCharacters(String s) {
+    int lastStart = 0;
+    String escapedString = "";
+    for ( int i=0; i < s.length(); i++ ) {
```
kgrgreer: int i = 0

TW80000 (approved):
Tested and it works. My only suggestion is to possibly use a `StringBuilder` or `StringBuffer` to build up `escapedString`.

PR #2715, opened 2019-10-23
Format currency in tabelCellFormatter

KateMengjunXia:
nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/8189


kgrgreer:
This calls fxService, which means it should be in nanopay.

jlhughes:
@nanoMichal currency/unit move is merged, so please update against this. 
Also, yes, the tablecellformatter should be in nanopay, but also this looks like some generic tablecell formatter and is aware of currency.   The UnitValue formatter should be doing this.

PR #2719, opened 2019-10-23
Convert dug to rules

nanoNeel:
Nanopay side PR:  https://github.com/nanoPayinc/NANOPAY/compare/CPF-2955/convert-DUG-to-rules?expand=1


src/files.js
```diff
@@ -704,5 +704,7 @@ FOAM_FILES([
   { name: "foam/strategy/StrategyReference" },
   { name: "foam/strategy/StrategizerService" },
   { name: "foam/strategy/BasicStrategizer" },
-  { name: "foam/strategy/ClientStrategizerService" }
+  { name: "foam/strategy/ClientStrategizerService" },
+
+  { name: "foam/nanos/dig/DUGRule"}
```
lchanmann: Please move this line to nanos.js.

src/foam/dao/EasyDAO.js
```diff
@@ -150,6 +150,13 @@ if ( getFixedSize() != null ) {
   }
 }
 
+
+if ( getRuler() ) {
+  foam.dao.DAO temp = getFixedSize() != null ? fixedSizeDAO : head;
+  foam.dao.DAO ruler = new foam.nanos.ruler.RulerDAO(getX(), temp, this.getName() );
+  pxy.setDelegate(ruler);
+}
```
lchanmann: Please move rulerDAO decoration to be after DeletedAwareDAO since the DeletedAwareDAO would cut-off other decorators that override `remove_`.

src/foam/dao/EasyDAO.js
```diff
@@ -150,6 +150,13 @@ if ( getFixedSize() != null ) {
   }
 }
 
+
+if ( getRuler() ) {
+  foam.dao.DAO temp = getFixedSize() != null ? fixedSizeDAO : head;
+  foam.dao.DAO ruler = new foam.nanos.ruler.RulerDAO(getX(), temp, this.getName() );
```
lchanmann: It would be more flexible to let user `setRulerDaoKey` so that it wouldn't be limited to `this.getName()`. For example, we could have existing rules with daoKey='localUserDAO' while the actual ruler is on bareUserDAO which is the parent of localUserDAO.

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,103 @@
...
+  ],
+
+ tableColumns: [
+    'id',
+    'documentation',
+    'enabled',
+    'priority',
+    'daoKey',
+    'operation',
+  ],
```
lchanmann: Can we have the same table columns as previously?
Screen Shot 2019-10-28 at 9 48 42 AM

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,103 @@
...
+
+  searchColumns: [
+    'id',
+    'url',
+    'enabled',
+    'priority',
+    'daoKey',
+    'operation',
+    'after',
+    'validity'
```
lchanmann: priority, operation, after, and validity are not DUG specific shouldn't be searchable from here.

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,103 @@
...
+  ],
+
+  properties: [
+    {
+      name: 'id',
+      visibility: 'RW'
```
lchanmann: We would rather not exposing `id` to users who have access to DUG rule since ones can accidentally or intentionally overwrite existing rules.

Please create a new property called name and remove 'RW' access from `id`.

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,103 @@
...
+      of: 'foam.nanos.http.Format',
+      name: 'format',
+      view: {
+        class: 'foam.u2.view.ChoiceView',
+        choices: [
+          [2, 'JSON'],
+          [4, 'XML'],
+        ]
+      }
+    },
```
lchanmann: The first choice is 'JSON' but if I didn't touch it when creating new DUG rule it didn't get saved to the journal.

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,103 @@
...
+        `
+    },
+    {
+      name: 'after',
+      value: true
+    },
+    {
+      name: 'predicate',
+      hidden: true,
+      networkTransient: true
```
lchanmann, line 108: user would not be able to change predicate from the GUI.

src/foam/nanos/dig/DUGRuleAction.js
```diff
@@ -0,0 +1,48 @@
+foam.CLASS({
+  package: 'foam.nanos.dig',
+  name: 'DUGRuleAction',
+
+  documentation: `Rule action for DUG`,
```
lchanmann: single quote is sufficient

src/foam/nanos/dig/DUGRuleAction.js
```diff
@@ -0,0 +1,48 @@
...
+        @Override
+        public void execute(X x) {
+          HTTPSink sink = null;
+          try {
+            sink = new HTTPSink(getUrl(), getFormat());
+          } catch (Exception e) {
+          }
+          sink.put(obj, null);
+        }
+      }, "DUG Rule");
```
lchanmann: indentation

lchanmann (changes requested):

src/foam/dao/EasyDAO.js
jlhughes: @lchanmann @nanoNeel Keep James in your loop. I've asked him to add RulerDAO support to EasyDAO.

src/foam/nanos/dig/DUGDAO.java
```diff
@@ -1,62 +0,0 @@
...
-
-    dug.setOwner(user.getId());
-    executeDUG(x, dug);
-
-    return super.put_(x, obj);
-  }
-
-  protected void executeDUG(X x, DUG dug) {
-    try {
-      dug.execute(Auth.sudo(x, dug.getOwner()));
```
lchanmann, line 56: @jlhughes do we need to preserve the user context when dug is run.

lchanmann: One of the old behaviors of DUG is that it switches to dug.owner context before getting the target dao from that context then add an HTTPSink listener.

For DUG rule, we don't need to get the target dao since the rule engine already does that for us and hands back the `obj` and the thing left to be done is putting it to an HTTPSink as seen in DugRuleAction.

src/foam/dao/EasyDAO.js
```diff
@@ -137,11 +137,11 @@ if ( head instanceof foam.dao.MDAO ) {
     getMdao().addIndex(getIndex());
   }
 }
-
+foam.dao.ProxyDAO fixedSizeDAO = null;
 if ( getFixedSize() != null ) {
   if ( head instanceof foam.dao.MDAO &&
        pxy != null ) {
-    foam.dao.ProxyDAO fixedSizeDAO = (foam.dao.ProxyDAO) getFixedSize();
+     fixedSizeDAO = (foam.dao.ProxyDAO) getFixedSize();
```
lchanmann: we can revert fixedSizeDAO declaration to be inside this block.

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,102 @@
...
+    },
+    {
+      name: 'asyncAction',
+      view: { class: 'foam.u2.tag.TextArea' },
+      javaGetter: `
+          DUGRuleAction action = new DUGRuleAction();
+          action.setUrl(getUrl());
+          action.setFormat(getFormat());
+          return action;
+        `
```
lchanmann: indentation

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,102 @@
...
+    },
+    {
+      name: 'after',
+      value: true
+    },
+    {
+      name: 'predicate',
+      hidden: true,
+      // Currently the front end does not load this correctly, causing incorrect values to be saved with each update.
```
lchanmann: // FIXME: ...

src/foam/nanos/dig/DUGRuleAction.js
```diff
@@ -0,0 +1,48 @@
...
+          @Override
+          public void execute(X x) {
+            HTTPSink sink = null;
+            try {
+              sink = new HTTPSink(getUrl(), getFormat());
+            } catch (Exception e) {
+            }
+            sink.put(obj, null);
+          }
+        }, "DUG Rule");
```
lchanmann: It might be helpful to include url.
```suggestion
        }, "DUG Rule (url:" + getUrl() + ")");
```

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,102 @@
+foam.CLASS({
+  package: 'foam.nanos.dig',
+  name: 'DUGRule',
+  extends: 'foam.nanos.ruler.Rule',
+
+  requires: [
+    'foam.nanos.http.Format'
+  ],
+
+ tableColumns: [
```
lchanmann: indentation

src/foam/nanos/dig/DUGRule.js
```diff
@@ -0,0 +1,102 @@
...
+        placeholder: '--',
+      },
+    },
+    {
+      name: 'enabled',
+      value: true
+    },
+    {
+      name: 'action',
+      hidden: true
```
lchanmann: we might also need `transient: true`.

lchanmann (commented):
Can we hide Id and Documentation and set `section: 'basicInfo'` for other properties?

Screen Shot 2019-10-29 at 5 03 58 PM

lchanmann:
All my comments are addressed. @jlhughes @kgrgreer Please take another look.

jlhughes (approved):

src/foam/nanos/dig/DUGDAO.java
jlhughes: sudo in this case makes sense.  The object will be processed as per the rule owners privileges.

PR #2723, opened 2019-10-24
Add acceptFormat property to DigFileUploadView 

yij793:
changes: add one more property to DigFileUploadView to allow user choice the type of upload file type, such as image/png or application/vnd.ms-excel, if it is empty, will be no limit

### DIFF
--- a/src/foam/nanos/dig/DigFileUploadView.js
+++ b/src/foam/nanos/dig/DigFileUploadView.js
@@ -106,7 +106,6 @@ foam.CLASS({
     {
       class: 'String',
       name: 'acceptFormat',
-      value: false
     },
     [ 'uploadHidden', false ],
     [ 'removeHidden', false ],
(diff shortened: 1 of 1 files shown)


minsunny:
will update Dig.js too as this view is used Dig.js

src/foam/nanos/dig/DigFileUploadView.js
```diff
@@ -103,6 +103,11 @@ foam.CLASS({
       name: 'dragActive',
       value: false
     },
+    {
+      class: 'String',
+      name: 'acceptFormat',
+      value: false
```
TW80000: `false` doesn't make sense as a default value for a String property.

yij793: I removed the value, just wondering should we make this value to 'application/vnd.ms-excel, text/csv'?

TW80000: Yeah, otherwise anything that depended on the default value will break.

yij793: Ok, changed, please review

TW80000 (approved):
I have no issue with this.

PR #2724, opened 2019-10-24
Supporting Map Parameter on SUGAR

minsunny:
DO not merge, I am checking now

### DIFF
--- a/src/foam/nanos/dig/SugarWebAgent.java
+++ b/src/foam/nanos/dig/SugarWebAgent.java
@@ -27,23 +27,30 @@ import java.lang.Exception;
 import java.lang.reflect.*;
 import java.nio.CharBuffer;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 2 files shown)


src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -27,23 +27,30 @@
...
+  protected Method cur_method;
+
   public SugarWebAgent() {}
 
   public void execute(X x) {
     Logger              logger         = (Logger) x.get("logger");
     PrintWriter         out            = x.get(PrintWriter.class);
     HttpServletResponse resp           = x.get(HttpServletResponse.class);
-    HttpParameters      p              = x.get(HttpParameters.class);
+    HttpServletRequest  req            = x.get(HttpServletRequest.class);
```
jlhughes, line 47: Why are you not using HttpParameters?  Do we need to add map support to it?
The HttpParameters abstracts a number of parameter passing methods.

PR #2725, opened 2019-10-25
CPF-3137: Remove Phone model and use a phone number property on User

skanderturki:
CPF-3137: Remove "Phone" model and just use a phone number property on User instead

Removed Phone object and replaced it by two properties phoneNumber and phoneVerified for each phone (phone, businessPhone and mobile).

No data migration at this point.

Linked to: https://github.com/nanoPayinc/NANOPAY/pull/8206


skanderturki:
NOTE: In three different files, a "validationPredicate" is commented out for the moment as it raises an error in the compilation: (foam/nanos/auth/User.js,  nanopay/contacts/Contact.js and nanopay/model/Business.js)

/Users/skander/Desktop/CPF-3137-bis/NANOPAY/build/src/java/foam/nanos/auth/User.java:2008: error: package foam.mlang.ExpressionsSingleton does not exist
        new foam.mlang.ExpressionsSingleton.Builder(foam.core.EmptyX.instance())
                                           ^

src/foam/nanos/auth/User.js
```diff
@@ -229,36 +228,47 @@ foam.CLASS({
       section: 'administrative'
     },
     {
-      class: 'FObjectProperty',
-      of: 'foam.nanos.auth.Phone',
-      name: 'phone',
+      class: 'String',
```
TW80000: We have a `foam.core.PhoneNumber` property that we should be taking advantage of.

```suggestion
      class: 'PhoneNumber',
```

src/foam/nanos/auth/User.js
```diff
@@ -229,36 +228,47 @@ foam.CLASS({
...
-      class: 'String',
-      name: 'phoneNumber',
-      transient: true,
-      documentation: `Omits properties of the phone number object and returns
-        the phone number.`,
-      expression: function(phone) {
-        return phone.number;
-      },
+      class: 'Boolean',
+      name: 'phoneNumberVerified',
```
TW80000, line 238: We don't want users to be able to set this value.

```suggestion
      name: 'phoneNumberVerified',
      writePermissionRequired: true
```

src/foam/nanos/auth/User.js
```diff
@@ -229,36 +228,47 @@ foam.CLASS({
...
       section: 'personal'
     },
     {
-      class: 'FObjectProperty',
-      of: 'foam.nanos.auth.Phone',
+      class: 'String',
```
TW80000: Same as above.

```suggestion
      class: 'PhoneNumber',
```

src/foam/nanos/auth/User.js
```diff
@@ -229,36 +228,47 @@ foam.CLASS({
...
-      of: 'foam.nanos.auth.Phone',
-      name: 'phone',
+      class: 'String',
+      name: 'phoneNumber',
       documentation: 'Personal phone number.',
-      factory: function() {
-        return this.Phone.create();
-      },
-      view: { class: 'foam.u2.detail.VerticalDetailView' },
+//      validationPredicates: [
```
TW80000: We should just refine `foam.core.PhoneNumber` to add the validation predicates there, then we don't need to repeat it.

src/foam/test/CSVSinkDemo.js
```diff
@@ -81,8 +79,8 @@ foam.CLASS({
             str: 'Random data ' + i,
             bool: Math.random() < 0.5,
             arr: [Math.random()],
-            fop: this.Phone.create({ number: '111-111-1111' })
-          }));
+            fop: '111-111-1111' })
```
TW80000:
```suggestion
            fop: '111-111-1111'
          }));
```

and remove the line below.

skanderturki: Ok thanks

src/foam/nanos/auth/User.js
skanderturki: In some files (user.js, contact.js, business.js) you can find:
    {
      class: 'Boolean',
      name: 'emailVerified',
      hidden: true
    }
I think maybe your remark applies here also. No?!

src/foam/nanos/auth/User.js
skanderturki: Question, when you add a validationPredicate to a class, do you refer to it by 'this'?
Example:
foam.CLASS({
  package: 'foam.core',
  name: 'PhoneNumber',
  extends: 'String',
  label: 'Phone number'
});


foam.CLASS({
  package: 'foam.core',
  name: 'FormattedPhoneNumber',
  extends: 'PhoneNumber',
  label: 'Formatted Phone number',
  validationPredicates: [
    {
      args: ['this'],
        return e.OR(
                      e.REG_EXP(
                                    e,
                                    /^(?:\+?1[-.●]?)?\(?([0-9]{3})\)?[-.●]?([0-9]{3})[-.●]?([0-9]{4})$/),
                      foam.mlang.predicate.False.create()
                    );
      },
      errorString: 'Please enter a valid phone number'
    }
  ]
});

src/foam/nanos/auth/User.js
TW80000: You're right, it does. We need to do a survey/review at some point and make sure we're doing that wherever it makes sense to.

src/foam/nanos/auth/User.js
TW80000: Check out https://github.com/foam-framework/foam2/blob/master/src/foam/core/Validation.js#L309-L344 for an example.

src/foam/test/CSVSinkDemo.js
```diff
@@ -81,8 +79,9 @@ foam.CLASS({
             str: 'Random data ' + i,
             bool: Math.random() < 0.5,
             arr: [Math.random()],
-            fop: this.Phone.create({ number: '111-111-1111' })
+            fop: '111-111-1111'
           }));
+          );
```
TW80000, line 84: Please remove.

src/foam/nanos/auth/User.js
```diff
@@ -660,3 +650,44 @@ foam.CLASS({
     }
   ]
 });
+
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'FormattedPhoneNumber',
+  extends: 'foam.core.PhoneNumber',
```
TW80000: We should refine `foam.core.PhoneNumber` rather than extend it. I don't see any reason `PhoneNumber` itself shouldn't have these validation predicate.

```suggestion
  refines: 'foam.core.PhoneNumber',
```

This refinement should also move to `Validation.js` so it can be found with the rest of the validation refinements.

TW80000 (changes requested):

src/foam/nanos/auth/User.js
skanderturki: Yes, a refinement, sure.
I also thought integrating them directly to phone number but thought maybe other validation rules in other countries may exist, I don't know.

src/foam/nanos/auth/User.js
```diff
@@ -229,36 +228,27 @@ foam.CLASS({
       section: 'administrative'
     },
     {
-      class: 'FObjectProperty',
-      of: 'foam.nanos.auth.Phone',
-      name: 'phone',
+      class: 'foam.core.PhoneNumberPropertyValidationRefinement',
```
TW80000:
```suggestion
      class: 'PhoneNumber',
```

src/foam/nanos/auth/User.js
```diff
@@ -229,36 +228,27 @@ foam.CLASS({
...
-      },
+      class: 'Boolean',
+      name: 'phoneNumberVerified',
+      writePermissionRequired: true,
       section: 'personal'
     },
     {
-      class: 'FObjectProperty',
-      of: 'foam.nanos.auth.Phone',
+      class: 'foam.core.PhoneNumberPropertyValidationRefinement',
```
TW80000:
```suggestion
      class: 'PhoneNumber',
```

TW80000 (commented):
@skanderturki I think these changes should fix the error you posted about in Slack.

src/foam/nanos/auth/User.js
skanderturki: Do you mean that our refinement (PhoneNumberPropertyValidationRefinement) would overwrite the original definition of PhoneNumber? This means PhoneNumber for example can only be refined once?! The semantics of refinement in foam should be clarified, or maybe it is but I don't know where.

TW80000: No, but I see the misunderstanding now. Refinements in FOAM modify/update the model being refined, they don't replace it. So in this way you can define a model and refine it later to add features/enhancements to it.

That's what we're doing in this PR. We're refining the PhoneNumber model (which represents a property whose value is a phone number, not a phone number itself) to add validation information to it. Other parts of the code base do refinements to add other features, such as Element.js setting default views for properties.

PR #2726, opened 2019-10-25
Add currency model

nanoMichal:


src/foam/u2/Element.js
```diff
@@ -2370,7 +2370,7 @@ foam.CLASS({
 foam.CLASS({
   package: 'foam.u2',
   name: 'CurrencyViewRefinement',
-  refines: 'foam.core.Currency',
+  refines: 'foam.core.UnitValue',
```
lchanmann, line 2373: line 2372 should it be 'UnitValueViewRefinement'?

src/foam/u2/search/FilterController.js
```diff
@@ -506,7 +506,7 @@ foam.CLASS({
 foam.CLASS({
   package: 'foam.u2.search',
   name: 'CurrencyRefinement',
-  refines: 'foam.core.Currency',
+  refines: 'foam.core.UnitValue',
```
lchanmann, line 509: line 508, name: 'UnitValueRefinement'?

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -138,7 +138,7 @@ foam.CLASS({
 foam.CLASS({
   package: 'foam.u2.view',
   name: 'CurrencyTableCellFormatterRefinement',
-  refines: 'foam.core.Currency',
+  refines: 'foam.core.UnitValue',
```
lchanmann, line 141: line 140, name: 'UnitValueTableCellFormatterRefinement'

lchanmann (commented):
We might also need unitDAO.

src/foam/u2/Element.js
nanoMichal: Thanks! changed.

src/foam/u2/search/FilterController.js
nanoMichal: Thanks! changed.

src/foam/u2/view/TableCellFormatter.js
nanoMichal: Thanks! changed.

src/foam/core/Currency.js
```diff
@@ -0,0 +1,224 @@
...
+    },
+    {
+      class: 'Long',
+      name: 'numericCode',
+      documentation: 'The numeric code associated with a type of currency.',
+      required: true
+    },
+    {
+      class: 'Int',
+      name: 'precision',
```
KateMengjunXia: Why repeat these properties (name, alphabeticCode, precision) that are already in the base class?

KateMengjunXia (commented):
Also, currencyDAO needs to be replaced by unitDAO

src/foam/core/Currency.js
```diff
@@ -0,0 +1,224 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+ foam.CLASS({
+  package: 'foam.core',
+  name: 'Currency',
+  extends: 'foam.core.Unit',
+
+  documentation: `The base model for storing, using and managing currency information.
+    All class properties require a return of *true* in order to pass.`,
```
TW80000: What does this mean? Properties don't have return values.

src/foam/core/Unit.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+ foam.CLASS({
+  package: 'foam.core',
+  name: 'Unit',
+  //abstract: true,
```
TW80000: Remove

tools/classes.js
```diff
@@ -483,7 +483,8 @@ var classes = [
   'foam.nanos.crunch.RemoveJunctionsOnUserRemoval',
   //authservice
   'foam.nanos.auth.CapabilityAuthService',
-
+    'foam.core.Unit',
```
TW80000: Please fix indentation

src/foam/u2/view/CurrencyView.js
```diff
@@ -9,7 +9,7 @@ foam.CLASS({
   name: 'CurrencyView',
   extends: 'foam.u2.view.ModeAltView',
 
-  documentation: 'A view for foam.core.Currency properties.',
+  documentation: 'A view for foam.core.UnitValue properties.',
```
TW80000: Should change the name of the View and places it's used as well then.

src/foam/core/Unit.js
```diff
@@ -0,0 +1,46 @@
...
+
+  documentation: `The abstract model for fungible digitized assets`,
+
+  ids: [
+    'alphabeticCode'
```
TW80000: Why wouldn't we just rename `alphabeticCode` to `id`? I don't think `alphabeticCode` makes much sense as a name, whereas `id` is consistent with the rest of the system.

src/foam/core/Unit.js
```diff
@@ -0,0 +1,46 @@
...
+  ],
+
+  javaImports: [
+    'foam.util.SafetyUtil'
```
TW80000: This isn't used and should be removed.

src/foam/core/Unit.js
```diff
@@ -0,0 +1,46 @@
...
+  ],
+
+  properties: [
+    {
+      class: 'String',
+      name: 'name',
+      documentation: 'name of the asset',
```
TW80000:
```suggestion
      documentation: 'Name of the asset.',
```

src/foam/core/Unit.js
```diff
@@ -0,0 +1,46 @@
...
+      label: 'Code',
+      documentation: 'The alphabetic code associated with the asset. Used as an ID.',
+      required: true,
+      aliases: [
+        'id'
+      ]
+    },
+    {
+      class: 'Int',
+      name: 'precision',
```
TW80000, line 29: I see an issue with this.

Since `UnitValue` extends long, all values are integers. Therefore `precision` doesn't make sense on `Unit`.

Either that or `UnitValue` should extend `Float` or `Double` instead.

src/foam/core/Currency.js
nanoMichal: thanks Kate!

src/foam/core/Unit.js
nanoMichal: yes, we need to do this.

src/foam/core/Unit.js
nanoMichal: each value is stored in the system as a long, precision turns it into a decimal for view purposes. ie precision of two and UnitValue of 100 would display as 1.00

src/foam/u2/view/CurrencyView.js
nanoMichal: I think this might need to stay as currency view. cmd-shift-R error.

src/foam/core/Unit.js
TW80000: True, forgot it was just for display. Thanks.

src/foam/core/Unit.js
nanoMichal: changed this to id.

src/foam/core/Unit.js
```diff
@@ -0,0 +1,37 @@
...
+      documentation: 'Name of the asset.',
+      required: true
+    },
+    {
+      class: 'String',
+      name: 'id',
+      label: 'Code',
+      documentation: 'The alphabetic code associated with the asset. Used as an ID.',
+      required: true,
+      aliases: [
```
jlhughes: no longer required.

src/foam/core/Unit.js
```diff
@@ -0,0 +1,37 @@
...
+      class: 'String',
+      name: 'name',
+      documentation: 'Name of the asset.',
+      required: true
+    },
+    {
+      class: 'String',
+      name: 'id',
+      label: 'Code',
+      documentation: 'The alphabetic code associated with the asset. Used as an ID.',
```
jlhughes: update doc - alphabetic code could be an example

src/services
```diff
@@ -712,3 +712,27 @@ p({
   "serve": true,
   "serviceClass": "foam.nanos.session.SimpleSessionService"
 })
+p({
+  "class":"foam.nanos.boot.NSpec",
```
TW80000: There should be a space after the colon here and on relevant lines below.

src/services
```diff
@@ -712,3 +712,27 @@ p({
...
+  "name":"currencyDAO",
+  "serve":true,
+  "serviceScript":
+  """
+  return new foam.dao.EasyDAO.Builder(x)
```
TW80000: This block from lines 721-729 should be indented two more spaces. Same goes for lines 733-736 below.

src/foam/core/types.js
```diff
@@ -489,7 +489,7 @@ foam.CLASS({
 
 foam.CLASS({
   package: 'foam.core',
-  name: 'Currency',
+  name: 'UnitValue',
```
TW80000, line 492: The plan was that `UnitValue` properties would have a method named `findUnit :: FObject → Promise Unit` that takes an `FObject` and returns a `Promise` of a `Unit`. Then we would refine `UnitValue` to add a `tableCellFormatter` that calls `findUnit` to get the `Unit`, then use the `Unit`'s `format` method to get the value to display in the table cell.

Shouldn't we add the `findUnit` method here? And add the `tableCellFormatter` refinement in this PR as well?

src/foam/core/Unit.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+ foam.CLASS({
+  package: 'foam.core',
+  name: 'Unit',
```
TW80000, line 9: Related to another comment I made: `Unit` is supposed to have a `format :: Long → String` method. Shouldn't that be added on this model?

src/foam/core/types.js
nanoMichal: Yep! @KateMengjunXia is completing this

src/foam/core/Unit.js
nanoMichal: Yes it will be by @KateMengjunXia's formatting Pr

src/services
nanoMichal: thanks.

src/services
nanoMichal: addressed.

jlhughes (approved):

PR #2727, opened 2019-10-25
MLang Predicate GUI Support

TW80000:
### Main changes

* Use the new strategizer service to populate the list of choices in `FObjectView`
* Use `FObjectView` of `Predicate` as the view for `PredicateProperty`-ies
* Use `FObjectView` of `Expr` as the view for `ExprProperty`-ies
* Added a `PropertyExpr` that implements `foam.mlang.Expr` and can be used as a strategy for that interface
* Added most predicates and exprs to `strategyDAO` so they show up as options in the dropdown

The changes above give us basic support for building MLang predicates in the GUI

### Supplementary changes

* Hid the Google DataStore Key property so it doesn't show up in the GUI when editing predicates
* Added a search field to the RichChoiceView for the `daoKey` property on the `Rule` model to make it easier to find the DAO you're looking for
* Added an assert statement to make sure `DAO.where` isn't called with a non-predicate property
* Fixed some slot memory leaks caused by links that were never being broken
* Updated `AttrSlot` to remove event listeners on unsubscribe
* Fixed a race condition in `SectionedDetailView` and its child views between slots of 'data' that was causing console errors
* Fix `SubSlot` so that it detaches when the parent class changes

## Future work

* Polish the UI

## Demo

https://giant.gfycat.com/PoliteTornBluewhale.webm


src/foam/mlang/expr/PropertyExpr.js
```diff
@@ -0,0 +1,76 @@
...
+    An Expr whose value is a property. Even though Property implements Expr
+    directly, we want this so FObjectView can list this as an option for a
+    concrete implementation of Expr instead of listing every property of every
+    model.
+  `,
+
+  properties: [
+    {
+      class: 'Class',
+      name: 'model',
```
kgrgreer: Normally we name is 'of'.

src/foam/u2/view/FObjectView.js
```diff
@@ -52,40 +59,50 @@ foam.CLASS({
       name: 'of'
     },
     {
+      class: 'Array',
       name: 'choices',
```
hchoura, line 63: The actual version will show just two type (User and Group). Did this a desired behaviour? or we need to get the exhaustive list of model.

TW80000: We need to add all of the models we want to show up in the list to the `strategyReferences` file. I only added `Group` and `User` as part of this PR, but I can add anything else if it makes sense. Most of our models are on the nanopay side though, so I'd have to add them in a nanopay PR.

PR #2729, opened 2019-10-28
Password view refactor

Nauna:
Refactoring Password Views:

This is stage one:

Here I've created one view that has three things that adjust the view and functionality.

1) horizontal(boolean) toggles the way the view is displayed
2) token(String) toggles the view to either be an update password or change password
3) topBarShow toggles the top banner with logo on/off

There will be at least two more stages.
Stage two - will be to refactor the Resend/Email Views for passwords. (Currently in progress)
Stage three+ - will be to clean up anything remaining (ie SuccessPasswordView ... ), also as of right now I'm thinking I maybe able to create something better just considering the similarities in the views for stage one and two.


jlhughes (approved):

nanoscott (approved):

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -1,235 +1,231 @@
...
-        .end()
-        .start(this.UPDATE_PASSWORD).addClass('update-BTN').end()
-      .end()
-    }
+        .addClass(this.myClass())
+        .startContext({ data: this })
+          .start().addClass('top-bar').show(this.topBarShow)
+            .start('img')
+              .attr('src', this.theme.logo)
+              .callIfElse( this.horizontal, function() {
```
zahnno, line 171: Here's another way you could do this. I'll leave it up to you to decide if it's better.
```suggestion
.enableClass('logoHorizontal', this.horizontal$);
.enableClass('logoCenterVerticle', this.horizontal$, true);
```
The third argument in enableClass negates the evaluated boolean value from the second argument.

zahnno: This would also provide the view the ability toggle between the two states without reinitializing the view.

PR #2732, opened 2019-10-29
Benchmark performance improvements

jlhughes:
- BenchmarkRunner - support incrementing through threads. 
- No longer clone on find()
- No longer add all properties to digest/signature by default
- update tests.

### DIFF
--- a/src/foam/core/AbstractFObject.js
+++ b/src/foam/core/AbstractFObject.js
@@ -383,7 +383,7 @@ foam.CLASS({
         Iterator i = props.iterator();
         while ( i.hasNext() ) {
           PropertyInfo prop = (PropertyInfo) i.next();
-          if ( ! prop.includeInSignature() ) continue;
+          if ( ! prop.includeInDigest() ) continue;
           if ( ! prop.isSet(this) ) continue;
           if ( prop.isDefaultValue(this) ) continue;
           signer.update(prop.getNameAsByteArray());
--- a/src/foam/core/AbstractFObject.js
+++ b/src/foam/core/AbstractFObject.js
@@ -237,6 +237,7 @@ foam.CLASS({
             // nop - ignore - only copy common properties.
           }
         }
+        this.__frozen__ = false;
         return this;
       `
     },
--- a/src/foam/core/AbstractFObject.js
+++ b/src/foam/core/AbstractFObject.js
@@ -237,7 +237,6 @@ foam.CLASS({
             // nop - ignore - only copy common properties.
           }
         }
-        this.__frozen__ = false;
         return this;
       `
     },
--- a/src/foam/core/AbstractFObject.js
+++ b/src/foam/core/AbstractFObject.js
@@ -405,9 +405,9 @@ foam.CLASS({
         Iterator i = props.iterator();
         while ( i.hasNext() ) {
           PropertyInfo prop = (PropertyInfo) i.next();
-          if ( ! prop.includeInSignature() ) continue;
+          if ( ! prop.includeInDigest() ) continue;
           if ( ! prop.isSet(this) ) continue;
-          if ( prop.isDefaultValue(this) ) continue;
+          //if ( prop.isDefaultValue(this) ) continue;
           verifier.update(prop.getNameAsByteArray());
           prop.updateSignature(this, verifier);
         }
--- a/src/foam/core/AbstractFObject.js
+++ b/src/foam/core/AbstractFObject.js
@@ -361,7 +361,7 @@ foam.CLASS({
           PropertyInfo prop = (PropertyInfo) i.next();
           if ( ! prop.includeInDigest() ) continue;
           if ( ! prop.isSet(this) ) continue;
-          if ( prop.isDefaultValue(this) ) continue;
+          //if ( prop.isDefaultValue(this) ) continue;
           md.update(prop.getNameAsByteArray());
           prop.updateDigest(this, md);
         }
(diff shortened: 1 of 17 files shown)


src/foam/core/AbstractFObject.js
```diff
@@ -237,6 +237,7 @@ foam.CLASS({
             // nop - ignore - only copy common properties.
           }
         }
+        this.__frozen__ = false;
```
kgrgreer: Why are you doing this? Once an object is frozen it needs to be frozen forever.

src/foam/dao/MDAO.java
```diff
@@ -48,6 +48,7 @@ The base AltIndex has two complete subindexes (each holds the entire DAO).
   protected Object   state_ = null;
   protected Object   writeLock_ = new Object();
   protected Set      unindexed_ = new HashSet();
+  long count = 0;
```
kgrgreer: Stale. Remove.

src/foam/dao/MDAO.java
```diff
@@ -89,7 +90,9 @@ public FObject objIn(FObject obj) {
   }
 
   public FObject objOut(FObject obj) {
-    return AbstractFObject.maybeClone(obj);
+    //    System.err.println("objOut.maybeClone() "+getOf().getId()+" "+(count++));
```
kgrgreer: Remove.

src/foam/nanos/auth/CreatedAwareDAO.js
```diff
@@ -33,6 +33,7 @@ foam.CLASS({
       javaCode: `
         // only set created if object does not exist in DAO yet
         if ( obj instanceof CreatedAware && getDelegate().find_(x, obj) == null ) {
+          if ( obj.isFrozen() ) obj = obj.fclone();
```
kgrgreer: If we do this here, then all DAO's that modify the incoming object will need to do the same check. Better to not do it and just require the uses clone() before put()'ing themselves.

src/foam/nanos/auth/CreatedByAwareDAO.js
```diff
@@ -20,6 +20,7 @@ foam.CLASS({
         }
         return this.delegate.find_(x, obj).then(function(result) {
           if ( result == null ) {
+            obj = obj.fclone();
```
kgrgreer: I would assume the object is already not frozen.

src/foam/nanos/auth/CreatedByAwareDAO.js
```diff
@@ -29,6 +30,7 @@ foam.CLASS({
         // only set created by if object does not exist in DAO yet
         if ( obj instanceof CreatedByAware && getDelegate().find_(x, obj) == null ) {
           User user = (User) x.get("user");
+          if ( obj.isFrozen() ) obj = obj.fclone();
```
kgrgreer: Same.

src/foam/nanos/auth/LastModifiedAwareDAO.js
```diff
@@ -28,6 +28,7 @@ foam.CLASS({
       },
       javaCode: `
         if ( obj instanceof LastModifiedAware ) {
+          if ( obj.isFrozen() ) obj = obj.fclone();
```
kgrgreer: same

src/foam/nanos/auth/LastModifiedByAwareDAO.js
```diff
@@ -22,6 +22,7 @@ foam.CLASS({
         if ( obj instanceof LastModifiedByAware ) {
           User user = (User) x.get("user");
           User agent = (User) x.get("agent");
+          if ( obj.isFrozen() ) obj = obj.fclone();
```
kgrgreer: same

src/foam/nanos/script/Script.js
```diff
@@ -273,11 +273,11 @@ foam.CLASS({
           this.status = this.ScriptStatus.RUNNING;
           this.runScript().then(() => {
             this.status = this.ScriptStatus.UNSCHEDULED;
-            this.scriptDAO.put(this);
+            this.scriptDAO.put(this.fclone());
```
kgrgreer: How did we update the status on the previous line if 'this' was frozen?

src/foam/nanos/script/Script.js
```diff
@@ -273,11 +273,11 @@ foam.CLASS({
...
           }).catch((err) => {
             console.log(err);
             this.status = this.ScriptStatus.ERROR;
-            this.scriptDAO.put(this);
+            this.scriptDAO.put(this.fclone());
```
kgrgreer: same

src/foam/core/AbstractFObject.js
jlhughes: this was just testing.  Needs to be removed.

src/foam/nanos/auth/CreatedByAwareDAO.js
jlhughes: well you would think, but I ended up here, because if was frozen.

src/foam/nanos/script/Script.js
jlhughes: another entry from testing.  needs to be undone.

jlhughes: I've reverted this, but I don't know how to fix this, as copyFrom should always fail, unless 'this.frozen' is false.

src/foam/nanos/script/ScriptRunnerDAO.java
```diff
@@ -54,12 +56,13 @@ Script runScript(final X x, Script newScript) {
         public void execute(X y) {
           try {
             script.setStatus(ScriptStatus.RUNNING);
-            getDelegate().put_(x, script);
+            getDelegate().put_(x, script).fclone();
```
jlhughes:
```suggestion
            getDelegate().put_(x, script);
```

PR #2737, opened 2019-10-29
Notification menu item view

zahnno:
Created new notification menu item view to reuse in other navigation views.

Removed notification logic in current UserView and replaced with notification menu item.


src/foam/nanos/u2/navigation/NotificationMenuItem.js
```diff
@@ -0,0 +1,143 @@
+foam.CLASS({
```
kgrgreer, line 6: Add copyright header and documentation:.

PR #2740, opened 2019-10-29
Fix searchColumns refinements not being respected

TW80000:
Without this, if you refine a model and change `searchColumns`, the old value will still be used.


lchanmann:
❤️thanks @TW80000.

carl-zzz:
❤️thanks @TW80000.

jlhughes (approved):

PR #2741, opened 2019-10-29
fixing an issue with TopNav and a race condition

Nauna:
@kgrgreer fix, I made PR

This fixes an issue that was caused by a menu(TopNav) to not render, due to a race condition and the incorrect rendering of DOM that at first is undefined.

### DIFF
--- a/src/foam/nanos/controller/ApplicationController.js
+++ b/src/foam/nanos/controller/ApplicationController.js
@@ -202,11 +202,23 @@ foam.CLASS({
       of: 'foam.nanos.theme.Theme',
       name: 'theme'
     },
+    {
+      class: 'foam.u2.ViewSpec',
+      name: 'topNavigation_',
+      factory: function() {
+        return this.TopNavigation;
+      }
+    },
+    {
+      class: 'foam.u2.ViewSpec',
+      name: 'footerView_',
+      factory: function() {
+        return this.FooterView;
+      }
+    },
     'currentMenu',
     'lastMenuLaunched',
-    'webApp',
-    'topNavigation_',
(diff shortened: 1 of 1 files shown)


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -254,8 +254,9 @@ foam.CLASS({
         this.fetchTheme().then(() => {
           this
             .addClass(this.myClass())
-            .start('div', null, this.topNavigation_$)
-              .tag(this.TopNavigation)
+            .start()
+              .tag(this.topNavigation_$.map(
+                (s) => s ? foam.u2.ViewSpec.createView(s, {}, this, this.__subContext__) : 'NULL'))
```
TW80000: I think this should work. It keeps the existing behaviour of defaulting to the imported `TopNavigation`, and also removes the need to call ViewSpec.createView manually, since `tag` will do it for you.

```suggestion
                function(s) {
                  return this.E().tag(s || self.TopNavigation);
                }
```

Maybe we can get rid of the `map` altogether and just do something like

```
.tag(this.topNavigation_$)
```

but I'm not sure that'll work.

If `topNavigation_` isn't `class: 'foam.u2.ViewSpec'` then it should be updated to that as well. And it could have a factory that returns `this.TopNavigation`.

Nauna: So I've adjusted to use default. And made similar changes to the footerview.

But the suggestion doesn't work. Looks like because its an inner element, it doesn't size correctly in the ablii app, where the menu is suppose to vertically align.

image

PR #2747, opened 2019-10-30
[CPF-824] Implement Routing Journal (take 2)

KernelDeimos:
This PR adds the ability to create one or more runtime journals which are shared between a number of DAOs.

- Shared journal records are prefixed with the service name followed by a literal dot (`.`). This makes the new journal format consistent with the old in that it is close to valid Javascript syntax.
- Lazy-loaded services will be booted by RoutingJournal during replay. RoutingJournal will call `x.get` in a thread; this allows the service initialization to block until replay is completed.
- Shared journal services provide an MDAO with the repository journal already replayed on it to the RoutingJournal, then wait for RoutingJournal to finish replay.


src/foam/dao/EasyDAO.js
```diff
@@ -279,8 +282,31 @@ return delegate;
...
+          (SharedJournalFactorySingleton) getX().get("sharedJournalFactory");
+        foam.dao.Journal jrl =
+          sharedJournalFactory.getJournal(
+            getJournalName(), getNSpec().getName(), myMDAO);
+        return new foam.dao.java.JDAO.Builder(getX())
+          .setDelegate(myMDAO)
+          .setOf(getOf())
+          .setJournal(jrl)
+          .build();
+      }
```
KernelDeimos, line 309: Another PR (or this one if requested) could implement a facade (as per the comment in JDAO) to contain this code, and the similar code for SINGLE_JOURNAL that's inside JDAO right now.

src/foam/dao/RoutingJournal.js
```diff
@@ -38,68 +77,141 @@ try {
...
+            }
+          }
+          return getReplayDAO_(service);
+        }
+
+        waitObj = new AtomicBoolean();
+
+        synchronized ( waitObj ) {
+          getReplayDAOWaitObjects().put(service, waitObj);
+          getReplayDAOWaitObjectsLock().unlock();
```
KernelDeimos, line 145: put is done inside this synchronized block to prevent waitObj.notifyAll() (inside registerReplayDAO method) from being called before waiting starts (which would result in a deadlock if it happens during the while condition below)

src/foam/dao/EasyDAO.js
```diff
@@ -279,8 +282,31 @@ return delegate;
...
+      else if ( getJournalType().equals(JournalType.SHARED_JOURNAL) ) {
+
+        foam.dao.MDAO myMDAO = new foam.dao.MDAO(getOf());
+        X resourceStorageX = getX();
+        if ( System.getProperty("resource.journals.dir") != null ) {
+          resourceStorageX = getX().put(foam.nanos.fs.Storage.class,
+              new ResourceStorage(System.getProperty("resource.journals.dir")));
+        }
+        new foam.dao.FileJournal.Builder(resourceStorageX)
```
jlhughes, line 296: If order doesn't matter, we could concatenate all the repository journals into a single journal shipped with the jar.

jlhughes:
Replaced by CPF-157/Medusa

PR #2750, opened 2019-10-31
adjust style and url clear

Nauna:
main adjustment here was for the token.
Because of a double call to this view from app controller, the first call will delete the url search params.

The css adjustments are just to make this look a little nicer.

### DIFF
--- a/src/foam/nanos/auth/ChangePasswordView.js
+++ b/src/foam/nanos/auth/ChangePasswordView.js
@@ -50,7 +50,7 @@ foam.CLASS({
     }
 
     ^ .top-bar {
-      background: /*%PRIMARY1%*/
+      background: /*%PRIMARY1%*/ #202341;
       width: 100%;
       height: 8vh;
       border-bottom: solid 1px #e2e2e3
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -49,13 +50,14 @@ foam.CLASS({
     }
 
     ^ .top-bar {
+      background: /*%PRIMARY1%*/
```
TW80000: If FOAM is run in a non-nanos app, then views won't have theme and this won't be replaced properly, so it's best practice to write it like this if you want to use theme properties in CSS, which is recommended:

```suggestion
      background: /*%PRIMARY1%*/ #202341;
```

That way in non-nanos FOAM apps, the CSS falls back on the default value.

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -157,6 +158,7 @@ foam.CLASS({
         this.clearProperty('originalPassword');
         this.clearProperty('newPassword');
         this.clearProperty('confirmationPassword');
+        if ( this.token ) window.history.replaceState(null, null, window.location.origin+'/#reset');
```
TW80000:
```suggestion
        if ( this.token ) window.history.replaceState(null, null, window.location.origin + '/#reset');
```

TW80000 (approved):
I just reviewed `ChangePasswordView` and have a few issues with it in its current state, but those issues aren't within the scope of this PR. Approving for the token change, but would like to talk with you in person @Nauna to see if we can simplify `ChangePasswordView`.

PR #2753, opened 2019-10-31
Unit value formatter

KateMengjunXia:
nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/8189

### DIFF
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -494,7 +494,11 @@ foam.CLASS({
   properties: [
     {
       class: 'String',
-      name: 'unitPropName'
+      name: 'unitPropName',
+      value: 'denomination',
+      documentation: `
+        The name of the property of a model that contains the denomination String.
+      `
     }
   ]
 });
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -490,7 +490,13 @@ foam.CLASS({
 foam.CLASS({
   package: 'foam.core',
   name: 'UnitValue',
-  extends: 'Long'
+  extends: 'Long',
+  properties: [
+    {
+      class: 'String',
+      name: 'unitPropName'
+    }
+  ]
 });
 
 
--- a/src/foam/u2/view/TableCellFormatter.js
+++ b/src/foam/u2/view/TableCellFormatter.js
@@ -152,7 +152,7 @@ foam.CLASS({
           obj.__context__.currencyDAO.find(unitId).then((unit) => {
             var slot = obj.slot(axiom.name).map((propValue) => unit.format(propValue));
             this.add(slot);
-            this.tooltip$.follow(slot);
+            this.onDetach(this.tooltip$.follow(slot));
           });
         } else {
           console.warn(obj.cls_.name, ' does not have the property: ', axiom.unitPropName);
--- a/src/foam/u2/view/TableCellFormatter.js
+++ b/src/foam/u2/view/TableCellFormatter.js
@@ -155,7 +155,7 @@ foam.CLASS({
             this.tooltip$.follow(slot);
           });
         } else {
-          console.warn(obj.cls_.name, 'doen not have the property: ', axiom.unitPropName);
+          console.warn(obj.cls_.name, ' does not have the property: ', axiom.unitPropName);
           this.add(value);
         }    
       }
--- a/src/foam/u2/view/TableCellFormatter.js
+++ b/src/foam/u2/view/TableCellFormatter.js
@@ -145,9 +145,10 @@ foam.CLASS({
       class: 'foam.u2.view.TableCellFormatter',
       name: 'tableCellFormatter',
       value: function(value, obj, axiom) {
-        if ( axiom.unitPropName ) {
-          var unitProp = obj.cls_.getAxiomByName(axiom.unitPropName);
+        var unitProp = obj.cls_.getAxiomByName(axiom.unitPropName);
+        if ( unitProp ) {
           var unitId = unitProp.f(obj);
+          // TODO: replace currencyDAO with unitDAO
           obj.__context__.currencyDAO.find(unitId).then((unit) => {
             var formatted = unit.format(value);
             this.start()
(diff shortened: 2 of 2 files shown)


src/foam/core/types.js
```diff
@@ -490,7 +490,13 @@ foam.CLASS({
 foam.CLASS({
   package: 'foam.core',
   name: 'UnitValue',
-  extends: 'Long'
+  extends: 'Long',
+  properties: [
+    {
+      class: 'String',
+      name: 'unitPropName'
```
jlhughes: how does unitPropName work?

KateMengjunXia: The `tableCellFormatter` will use it find the property that has the Currency String
An example is here in `Invoice.js`: https://github.com/nanoPayinc/NANOPAY/pull/8189/files

TW80000: @jlhughes This is the idea that Kate and I came up with. The problem is that we still need to store the exact unit (eg: CAD vs USD) as data on each instance of an object with a `UnitValue` property. So we would either need to make `UnitValue` store both the long (quantity) and the string (unit id) as its value, or continue what we're currently doing and require a second property that accompanies the `UnitValue` property and that stores the unit id. Eg: `amount` and `denomination`. Then all we're missing is a way to relate the two properties, which `unitPropName` accomplishes.

It's not the only potential solution to the problem, but it's very close to what we have now so very few changes are required, and none of them should be breaking. And it still allows us to define the `tableCellFormatter` once in a generic way and avoid repeating ourselves.

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -144,11 +144,25 @@ foam.CLASS({
...
-      value: function(value) {
-        this.start()
-          .style({'text-align': 'left', 'padding-right': '20px'})
-          .add('$' + (value/100).toFixed(2).replace(/(\d)(?=(\d{3})+\.)/g, '$1,'))
-        .end();
+      value: function(value, obj, axiom) {
+        if ( axiom.unitPropName ) {
+          var unitProp = obj.cls_.getAxiomByName(axiom.unitPropName);
+          var unitId = unitProp.f(obj);
+          obj.__context__.currencyDAO.find(unitId).then((unit) => {
```
TW80000, line 152: This should be `unitDAO` since this refinement is for `UnitValue` properties, and currencies should be stored in `unitDAO`.

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -144,11 +144,25 @@ foam.CLASS({
     {
       class: 'foam.u2.view.TableCellFormatter',
       name: 'tableCellFormatter',
-      value: function(value) {
-        this.start()
-          .style({'text-align': 'left', 'padding-right': '20px'})
-          .add('$' + (value/100).toFixed(2).replace(/(\d)(?=(\d{3})+\.)/g, '$1,'))
-        .end();
+      value: function(value, obj, axiom) {
+        if ( axiom.unitPropName ) {
```
TW80000: Since we're refining `UnitValue`, `axiom` should always be a `UnitValue`, and `unitPropName` should always be set. I think it would be better to log a warning if it's not set and just output the unformatted value in that case.

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -144,11 +144,25 @@ foam.CLASS({
...
+          var unitId = unitProp.f(obj);
+          obj.__context__.currencyDAO.find(unitId).then((unit) => {
+            var formatted = unit.format(value);
+            this.start()
+              .add(formatted)
+            .end();
+            this.tooltip = formatted;
+          });
+        } else {
+          var formatted = '$' + (value/100).toFixed(2).replace(/(\d)(?=(\d{3})+\.)/g, '$1,');
```
TW80000: Again, we're refining `UnitValue` here, which isn't necessarily a currency. So this doesn't make sense here since it's specifically related to currency formatting.

TW80000 (changes requested):

src/foam/core/types.js
jlhughes: can the default 'unitPropName' be 'denomination' ?  Then it only needs to be specified if the model has more than one (or differs).

TW80000: That seems fine to me.
@KateMengjunXia could you please add a documentation comment to this property to explain what it's for?

src/foam/u2/view/TableCellFormatter.js
KateMengjunXia: We currently don't have a `unitDAO` yet. @nanoMichal has been working on splitting the DAO's, we can later on change all the `currencyDAO` in nanopay to `unitDAO` where applicable.

TW80000: Ok, then please add a TODO comment so we don't forget to change that afterwards.

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -144,11 +144,20 @@ foam.CLASS({
...
+        if ( unitProp ) {
+          var unitId = unitProp.f(obj);
+          // TODO: replace currencyDAO with unitDAO
+          obj.__context__.currencyDAO.find(unitId).then((unit) => {
+            var slot = obj.slot(axiom.name).map((propValue) => unit.format(propValue));
+            this.add(slot);
+            this.tooltip$.follow(slot);
+          });
+        } else {
+          console.warn(obj.cls_.name, 'doen not have the property: ', axiom.unitPropName);
```
TW80000:
```suggestion
          console.warn(obj.cls_.name, ' does not have the property: ', axiom.unitPropName);
```

src/foam/u2/view/TableCellFormatter.js
```diff
@@ -144,11 +144,20 @@ foam.CLASS({
...
-        .end();
+      value: function(value, obj, axiom) {
+        var unitProp = obj.cls_.getAxiomByName(axiom.unitPropName);
+        if ( unitProp ) {
+          var unitId = unitProp.f(obj);
+          // TODO: replace currencyDAO with unitDAO
+          obj.__context__.currencyDAO.find(unitId).then((unit) => {
+            var slot = obj.slot(axiom.name).map((propValue) => unit.format(propValue));
+            this.add(slot);
+            this.tooltip$.follow(slot);
```
TW80000:
```suggestion
            this.onDetach(this.tooltip$.follow(slot));
```

TW80000 (changes requested):

PR #2754, opened 2019-11-01
add flcone

nanoNeel:


src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -231,7 +231,7 @@ private void saveHistory(Rule rule, FObject obj) {
    * @return Reloaded object
    */
   private FObject reloadObject(FObject obj, FObject oldObj, FObject nu, boolean greedy) {
-    FObject old = oldObj;
+    FObject old = (FObject) oldObj.fclone();
```
kgrgreer, line 234: We'll get a NPE if oldObj is null. Better to call SafetyUtil.deepClone(oldObj);

PR #2755, opened 2019-11-01
RuleAction that throws an Exception

nanoMichal:
This rule action throws an exception with the indicated message. 
Useful for terminating execution of rule groups if the predicate evaluates to true.

### DIFF
--- a/src/foam/nanos/ruler/action/ThrowErrorAction.js
+++ b/src/foam/nanos/ruler/action/ThrowErrorAction.js
@@ -4,7 +4,7 @@
  * http://www.apache.org/licenses/LICENSE-2.0
  */
 
- foam.CLASS({
+foam.CLASS({
   package: 'foam.nanos.ruler.action',
   name: 'ThrowErrorAction',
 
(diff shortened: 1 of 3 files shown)


src/foam/nanos/ruler/action/ThrowErrorAction.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+ foam.CLASS({
```
TW80000: Extra space

TW80000 (approved):
Seems reasonable to me.

src/foam/nanos/ruler/action/ThrowErrorAction.js
nanoMichal: 🧐

lchanmann (commented):
Rule engine would behave differently if this action is used as an action or asyncAction.
The former would cause all rules in the same group to not run but for the latter (asyncAction) it would move on to the next rule.

nanoMichal:
@lchanmann My use case would be for stopping the rule group. What do you suggest, add a note in the documentation for the two results?

PR #2757, opened 2019-11-01
NO_MERGE_GIT_ISSUES_:( added aware toggle to easydao

michaelconnor468:


src/foam/dao/EasyDAO.js
```diff
@@ -566,35 +566,41 @@ return delegate;
       name: 'testData',
       generateJava: false
     },
+    {
+      documentation: 'Enables automated adding of aware doas to qualifying decorator chains',
```
```suggestion
      documentation: 'Enables automated adding of property-related DAO decorators to qualifying decorator chains',
```

src/foam/dao/EasyDAO.js
```diff
@@ -566,35 +566,41 @@ return delegate;
...
+      name: 'enableAware',
+      class: 'Boolean',
+      value: true
+    },
     {
       name: 'serviceProviderAware',
       class: 'Boolean',
-      javaFactory: 'return foam.nanos.auth.ServiceProviderAware.class.isAssignableFrom(getOf().getObjClass());'
+      javaFactory: 'return !getEnableAware() ? false : foam.nanos.auth.ServiceProviderAware.class.isAssignableFrom(getOf().getObjClass());'
```
```suggestion
      javaFactory: 'return getEnableAware() && foam.nanos.auth.ServiceProviderAware.class.isAssignableFrom(getOf().getObjClass());'
```

is simpler. Please change the others as well.

src/foam/dao/EasyDAO.js
```diff
@@ -566,35 +566,41 @@ return delegate;
       name: 'testData',
       generateJava: false
     },
+    {
+      documentation: 'Enables automated adding of aware doas to qualifying decorator chains',
+      name: 'enableAware',
```
TW80000: This is close enough to `EnabledAware` to be confusing. I think we should try to come up with a better name. `enablePropertyDecorators` maybe?

TW80000 (changes requested):

src/foam/dao/EasyDAO.js
michaelconnor468: yahh that's probably a good idea; I'm going to change it to enableAwareDecorators

jlhughes (approved):
Sure, this works.

PR #2758, opened 2019-11-01
Cpf 3147/easy dao aware3

michaelconnor468:


jlhughes (approved):

src/foam/dao/EasyDAO.js
```diff
@@ -566,35 +566,41 @@ return delegate;
       name: 'testData',
       generateJava: false
     },
+    {
+      documentation: 'Enables automated adding of property-related DAO decorators to qualifying decorator chains',
+      name: 'enableAwareDecorators',
```
TW80000: I still don't like this name. I think `enablePropertyDecorators` is better. "Aware" just happens to be in the name of the relevant decorators, but the english meaning of the word "aware" doesn't apply here.

jlhughes: @TW80000 @michaelconnor468 regarding the name, this flag is only affecting the 'interface' decorators - so perhaps named along that line.  enableInterfaceDecorators?  As in each we test if the model implements the interface.

TW80000 (approved):

jlhughes (approved):

PR #2759, opened 2019-11-01
user with no grp gets default passPolicy

Nauna:
issue was if a user does not have a group they would not be able to set a password. THus this will pick up the default passwordPolicy if the user does not have a group


src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -279,6 +275,7 @@ foam.CLASS({
         }
 
         // Validate the password against the password policy
+
```
nanoscott: remove extra line.

nanoscott (approved):

TW80000 (approved):

jlhughes (approved):
We'll need to consider how to impose a password policy of the group we know the user will belong to after signup.

PR #2765, opened 2019-11-04
Notification setting models

KateMengjunXia:
https://nanopay.atlassian.net/browse/CPF-2940
nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/8314


src/foam/nanos/notification/SendEmail.js
```diff
@@ -0,0 +1,19 @@
...
+        { name: 'notification', type: 'foam.nanos.notification.Notification' },
+      ],
+      javaCode: `
+      return;
+        
+      `
+    }
+  ]
+});
+
```
zahnno: Why is this model needed? Wouldn't the NotificationConfig implement the NotificationSetting?

jlhughes: This is one of the many NotificationSettings in the NotificationConfig.  The NotifcationConfig walks all settings and calls sentNotification()

jlhughes: @KateMengjunXia can this be named EmailNotificationSettings?
Also, where is the logic to walk through the settings and call sendNotification?

src/foam/nanos/notification/SendEmailSetting.js
```diff
@@ -0,0 +1,53 @@
+foam.CLASS({
+  package: 'foam.nanos.notification',
+  name: 'SendEmailSetting',
```
jlhughes: Why do we call these send...  why not just EmailSettings and NotificationSettings?
And the NotificationSettings has send(x, user, notification).

src/foam/nanos/auth/User.js
```diff
@@ -589,6 +593,29 @@ foam.CLASS({
...
+      name: 'notify',
+      javaCode: `
+        DAO agentJunctionDAO      = (DAO) x.get("agentJunctionDAO");
+        DAO notificationConfigDAO = (DAO) x.get("notificationConfigDAO");
+        
+        UserUserJunction junction = (UserUserJunction) agentJunctionDAO
+          .find(EQ(UserUserJunction.SOURCE_ID, getId()));
+        if ( junction == null ) {
+          throw new RuntimeException("A junction between the user and its entities cannot be found.");
+        }
```
zahnno: User's should be calling their own notificationConfig, not one on the agentJunction - I think it'd be okay to have it exist on the user itself.

jlhughes (approved):
Looks good

PR #2766, opened 2019-11-04
Add test data producers

KernelDeimos:
This PR adds models to aid in generating test data for models with validation predicates.

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -618,6 +618,8 @@ FOAM_FILES([
   { name: "foam/test/AllSerializableProperties" },
   { name: "foam/test/TestEnum" },
   { name: "foam/test/TestObj" },
+  { name: "foam/test/testdata/TestDataPopulator" },
+  { name: "foam/test/testdata/generators" },
   { name: "foam/core/FObjectTest" },
   { name: "com/google/foam/demos/bubbles/Bubbles" },
   { name: "com/google/foam/demos/u2/AllViews" },
--- a/src/foam/test/testdata/TestDataPopulator.js
+++ b/src/foam/test/testdata/TestDataPopulator.js
@@ -1,0 +1,70 @@
+
+foam.CLASS({
+  package: 'foam.test.testdata',
+  name: 'TestDataPopulator',
+
+  javaImports: [
+    'java.util.HashMap',
+    'java.util.List',
+    'java.util.Map'
+  ],
+
+  properties: [
+    {
+      name: 'stringProducers',
+      class: 'Map',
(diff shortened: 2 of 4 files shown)


TW80000:
I like the idea of being able to generate test data, but I don't think this is the best approach. I was imagining we could refine the core Property types to add new property properties that support random data generation. For example, we could refine all properties to add a `getRandomValue` method (or some other name) that returns an appropriate value for whatever the type is. For example, `String`s might return lorem ipsum, `Email` properties might return random email addresses, `Long`s return random numbers, etc.

These methods could be used for generating test data or for generating fake data for display/demo purposes.

This approach would take advantage of polymorphism, where each type of property provides its own implementation of the best way to generate example data. It's also more aligned with what we do elsewhere. For example, this is what we do to refine properties in u2 to give them appropriate base views based on their types. Longs get these views, Dates get different views, etc.

KernelDeimos:
That makes sense to me as long as a specific model with validation predicates is able to override this for a specific property (maybe by setting a property property to an implementor of an interface that has getRandomValue).

Another thing I was thinking of was having a boolean to allow occasional invalid data, which could be useful for tests regarding how that's handled, or how invalid data affects performance. I'm not sure how I'd do it with the property refinement approach but I'm sure there's a way.

kgrgreer:
One thing I would like to add to properties is sample values, which would be similar, or could be used for similar purposes. For example, a firstName property might have examples: ['Kevin', 'Elvis', 'Kate' , 'Madonna'] and a date field might have: examples: ['1970-11-19', '2019-02-28' ]. I'm interested in doing this primarily for documentation purposes. I mean, people generally know what a firstName is, but for things like bank transit numbers, it's helpful to have an example. I'm not sure this sort of features pays its weight. I sometimes write scripts to generate random values, but it isn't difficult, and I do it at most once a year,  so I'm not sure it's worthwhile given it would take me time to find the helper package that helps with this and then learn how it works . I think in less time I could just write the few lines necessary. This is a philosophical debate about where we draw the line about what sort of things FOAM provides and what it doesn't, or even, what is worth encoding as a library vs. just writing as required. Things we we could do is to add a pickRandom() method to foam.lib.Array or add probabilistic method execution support like FOAM1 had: https://github.com/foam-framework/foam/blob/master/core/stdlib.js#L197  But, I also agree with @Troy, in that if we were going to have this feature, it would be better to use Property polymorphism, which defaults to the type/Property classes by default. This would let properties with values like min, max, etc. to generate values in range.

PR #2771, opened 2019-11-05
Run OrderedSink before PredicatedSink to improve performance

KernelDeimos:
This change was benchmarked with contactDAO. contactDAO was populated with 50k entries, and then a 1000 iterations of the following query were run for each trial:
```
 contactDAO
      .orderBy(new foam.mlang.order.Desc(Contact.ID))
      .limit(500)
      .select(new ArraySink());
```

Results were as follows:

### Prior to the change
trial1: 7.8ms avg
trial2: 7.1ms avg
trial3: 8.8ms avg

### After the change
trial1: 6.7ms avg
trial2: 5.8ms avg
trial3: 6.7ms avg

Credit to @TW80000 for coming up with a theory that this will boost performance

### DIFF
--- a/src/foam/dao/AbstractDAO.js
+++ b/src/foam/dao/AbstractDAO.js
@@ -594,12 +594,12 @@ if ( ( skip > 0 ) && ( skip < AbstractDAO.MAX_SAFE_INTEGER ) ) {
   sink = new SkipSink(skip, 0, sink);
 }
 
-if ( order != null ) {
-  sink = new OrderedSink(order, null, sink);
-}
-
 if ( predicate != null && predicate.partialEval() != null && ! ( predicate instanceof foam.mlang.predicate.True) ) {
   sink = new PredicatedSink(predicate, sink);
+}
+
+if ( order != null ) {
+  sink = new OrderedSink(order, null, sink);
 }
 
 return sink;


KernelDeimos:
This needs more evaluation - it resulted in a performance increase when ordering by ID, but not when ordering by other fields

kgrgreer:
In your benchmark, is contactDAO predicated? Or is it authenticated, which is providing the predicate?

kgrgreer:
I think we can make this better by only reversing the order if there is a Limit. Otherwise, the old order will be more efficient.

KernelDeimos:
I'm actually not so sure there's a performance increase anymore. There should be, but I wasn't able to observe it again after I fixed a problem with my benchmarking.

I'm going to record what I learned in this comment so it doesn't get lost:

## Benchmarking
### Preconditions:
- contactDAO with 1000 records (close to the number in production)
- user is a non-admin Ablii user (so that authorization predicate is required)
- query specifies order for a property other than ID (FIRST_NAME was used)
- query has limit and skip (90 and 750 were used arbitrarily)

### Results after moving order (this PR)
- About 10-15% slower; haven't determined why yet

### Results after adding sink to create fallback sub when sub is null
- No significant difference observed. I want to double-check this result later.

## Manual testing
- Manually loading a contactDAO with a very high number of records (50,000 was used) results in a very slow pageload because getting the record count takes a long time. Getting the 90 records for the table takes at most a couple seconds.

kgrgreer:
If the predicate is expensive and you have a limit, then it makes sense to order first because then you won't need to apply the predicate to entries that appear after the limit.
However, if you don't have a limit, then you'll need to apply the predicate to all entries anyway, so it should be faster to predicate first so that you have fewer entries to sort.
So, the order should depend on if you have limit or not.
Also, it could depend on whether you have a "slow" or "fast" predicate, but that's probably not worth accounting for (in terms of code complexity).

TW80000:
I spent some time yesterday testing this out. I ran into an issue, unfortunately. When I switched the order of the `Sink` decorators so that the predicate gets applied after ordering, I got an error saying `User` couldn't be cast to `Contact` (which is a subclass of User and is stored in the same DAO). The cast was happening in a call to `.compare`, which was being called as part of the sort that `OrderedSink` does. This makes sense, since `contactDAO` is just a `FilteredDAO` on top of `userDAO`, which just adds a predicate. But since that predicate gets applied _after_ the order when we switch the order of the sink decorators, the order is still being applied to other objects in the DAO that aren't `Contact`s.

Due to this issue, I don't think it's possible to do the predicate after ordering without making additional changes.

The thought crossed my mind of supporting a way to flag predicates to denote whether they should apply before or after the sort, but it's a bit of an ugly idea. We'd have to walk the predicate if it contains `AND`s or `OR`s and look for instances of predicates that apply before or after, then split it up and apply some before and the rest after, while still keeping the desired behaviour and semantics of the original predicate.

Another thought is that it might be safe to say that `IsInstanceOf` and `IsClassOf` are special cases and should always be applied before the order. You'd still have to walk the predicate to extract these "special" predicates though. I'm not certain that these two predicates are always special cases though, or whether others are special cases as well.

PR #2773, opened 2019-11-05
Accounted for a case where remove doesn't actually remove the object

tharmaman:


src/foam/dao/CachingDAO.js
```diff
@@ -108,8 +109,12 @@ foam.CLASS({
       are up to date. */
     function remove_(x, o) {
       var self = this;
-      return self.src.remove(o).then(function() {
-        return self.delegate.remove_(x, o);
+      return self.src.remove(o).then(removedObj => {
+        // only remove from cache and delegate if the object was actually removed (the removed object is not null)
+        if ( removedObj ){
+          this.onSrcRemove(removedObj);
+          return self.delegate.remove_(x, o);
```
zahnno, line 116: I don't think you need to call this.onSrcRemove since we're returning self.delegate.remove_(x, o).

kgrgreer: I agree.

src/foam/dao/CachingDAO.js
```diff
@@ -89,7 +89,8 @@ foam.CLASS({
       var proxy = this.src$proxy;
       proxy.listen(this.QuickSink.create({
         putFn: this.onSrcPut,
-        removeFn: this.onSrcRemove,
+        // TODO: figure out if this part is important
+        // removeFn: this.onSrcRemove,
```
kgrgreer, line 93: Why is this being removed?

PR #2775, opened 2019-11-06
add support for Post method in SUGAR

hchoura:


src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -6,29 +6,37 @@
...
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Parameter;
+import java.nio.CharBuffer;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.gson.Gson;
```
kgrgreer: Why are we using this library?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -299,9 +321,40 @@ protected Object setParameterValue(X x, HttpServletResponse resp, PrintWriter ou
...
+          arg = ( (Double) object ).longValue();
+          break;
+        case "java.lang.String":
+          arg = object;
+          break;
+        //TODO add the case "String[]":
+        case "foam.core.X":
+          arg = x;
+          break;
+        default:
```
nanoscott: The whole reason for doing this is so that more complex objects can be passed into the SUGAR API. Can you figure out how to parse FObjects here. An example of a call that you should be able to parse is the AgentAuthService.canActAs where the second and third parameters are User objects and not simple string/int/bools .. you will need to figure out how to parse these here as well.

nanoscott (changes requested):

src/foam/nanos/dig/SugarWebAgent.java
nanoscott: @kgrgreer .. how do we parse JSON into a map of key value pairs?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -123,19 +136,28 @@ public void execute(X x) {
               }
 
               paramTypes[j] = pArray[j].getType();
-              arglist[j] = p.getParameter(pArray[j].getName());
+
+              if ( p.getParameter(pArray[j].getName()) != null ) {
```
nanoscott: I think we can simply remove the ability for users to pass in method parameters in the GET URL. This code should no longer be necessary. So it would be required that whenever a method has parameters, the caller uses a POST call with a message body in order to pass in the parameters.

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -123,19 +136,28 @@ public void execute(X x) {
...
               logger.debug(pArray[j].getName() + " :   " + p.getParameter(pArray[j].getName()));
 
               // casting and setting according to parameters type
               String typeName = pArray[j].getType().getCanonicalName();
-              Class paramObj_ = null;
-              Object obj_ = null;
 
               if ( typeName.equals("foam.core.X") ) {
                 arglist[j] = x;
               } else if ( ! SafetyUtil.isEmpty(p.getParameter(pArray[j].getName())) ) {
```
nanoscott: This should be able to be remove too.

src/foam/nanos/dig/SugarWebAgent.java
hchoura: I switch it to foam MapParser

src/foam/nanos/dig/SugarWebAgent.java
hchoura: I prefer to have the post and the get method for SUGAR.

src/foam/nanos/dig/SugarWebAgent.java
hchoura: We will create object based on the method type.

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -45,6 +56,12 @@ public void execute(X x) {
     String              methodName     = p.getParameter("method");
     String              interfaceName  = p.getParameter("interfaceName");
 
+    String data = p.getParameter("data");
```
nanoscott: what happens when data is not provided? i.e. data is null?

src/foam/nanos/dig/SugarWebAgent.java
nanoscott: GET still works as long as there are no method parameters in the URL. We can remove this code and force users to pass in method parameters through a POST body.

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -123,19 +141,59 @@ public void execute(X x) {
...
 
               if ( typeName.equals("foam.core.X") ) {
                 arglist[j] = x;
               } else if ( ! SafetyUtil.isEmpty(p.getParameter(pArray[j].getName())) ) {
+                //get method
                 arglist[j] = setParameterValue(x, resp, out, p, typeName, pArray[j]);
+              } else if ( paramName != null ) {
+                //post method
+                //TODO array
+                if ( typeName.equals("double") || typeName.equals("boolean") || typeName.equals("int") || typeName.equals("long") || typeName.equals("long") || typeName.equals("java.lang.String") ) {
```
nanoscott: You shouldn't do it this way. You can add all of the supported types to a Set (which could be made a constant), then do a Set.contains here instead of having this long if statement.

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -123,19 +141,59 @@ public void execute(X x) {
...
+                  arglist[j] = mapPostParam.get(paramName);
```
nanoscott: won't you still have the casting problem here where all numbers are parsed as Doubles? If not, how is the MapParser able to distinguish between ints and longs and doubles?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -123,19 +141,59 @@ public void execute(X x) {
...
+                } else {
+                  
+                  Object arg      = null;
+                  Class  objClass = null;
+                  try {
+                    objClass = Class.forName(typeName);
+                    // For Enum Parameter
+                    if ( objClass != null && objClass.isEnum() ) { 
+                      for ( int t = 0 ; t < objClass.getEnumConstants().length ; t++ ) {
+                        if ( objClass.getEnumConstants()[t].toString().equals(mapPostParam.get(paramName)) ) {
```
nanoscott: what if the enum value is passed in as an integer?

nanoscott: use SafetyUtil.equals here

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -123,19 +141,59 @@ public void execute(X x) {
...
+                          arg = objClass.getEnumConstants()[t];
+                        }
+                      }
+                    } else {
+                      FObject             obj = (FObject) objClass.newInstance();
```
nanoscott: can't objClass be null here?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -123,19 +141,59 @@ public void execute(X x) {
...
+                      Map<String, String> mParam   = (Map) mapPostParam.get(paramName);
+
+                      for ( Map.Entry<String, String> entry : mParam.entrySet() ) {
+                        obj.setProperty(entry.getKey(), entry.getValue());
```
nanoscott: so this works even if the property is not a string?

nanoscott: what about nested objects?

nanoscott (changes requested):

nanoscott:
Talked to @hchoura offline and discussed adding two things to this PR:
- Add a test to exercise the SUGAR API
- Update the SUGAR UI to allow for users to paste JSON into the Data field and send the POST properly

src/foam/nanos/dig/SUGAR.js
```diff
@@ -279,12 +242,12 @@ foam.CLASS({
       name: 'postButton',
       label: 'Send POST Request',
       code: async function() {
-        if ( this.postURL !== '' ) {
+        if ( this.sugarURL !== '') {
           var req = this.HTTPRequest.create({
-            url: window.location.protocol + '//' + window.location.hostname + ':' + window.location.port + this.postURL + this.postData+ "&sessionId=" + localStorage.defaultSession,
+            url: window.location.protocol + '//' + window.location.hostname + ':' + window.location.port + this.sugarURL + '?data=' + this.postData+ '&sessionId=' + localStorage.defaultSession,
```
nanoscott, line 247: postData should not be included twice in this .. if it is included in the URL ensure that it is encoding properly so that browsers can handle the request.

nanoscott (approved):
Just one minor comment. Please address ... otherwise it looks good.

src/foam/lib/json/FObjectParser.java
```diff
@@ -46,7 +46,13 @@ public PStream parse(PStream ps, ParserContext x) {
               subx.set("enum", c);
               subParser = EnumParserFactory.getInstance(c);
             } else {
-              Object obj = ((X) x.get("X")).create(c);
+              Object obj = null;
```
kgrgreer, line 49: Why do we do this rather than just always provide the X when using FObjectParser?

src/foam/nanos/dig/SUGAR.js
```diff
@@ -208,7 +208,7 @@ foam.CLASS({
       // TODO: appears not to work if named 'url', find out why.
       name: 'sugarURL',
       label: 'URL',
-      hidden: false,
+      hidden: true,
```
kgrgreer, line 211: Why do we hide this?

src/foam/nanos/dig/SUGAR.js
```diff
@@ -217,54 +217,17 @@ foam.CLASS({
...
-              }
-            }
-          }
-        }
-        if ( (url.length + this.postData.length) >= this.MAX_URL_SIZE ) {
-          this.postURL = url;
-        }
-
-        return encodeURI(url + this.postData);
+        this.postData=this.data.toString();
```
kgrgreer: Formatting:
this.postData = this.data.toString();

src/foam/nanos/dig/SUGAR.js
```diff
@@ -279,12 +242,12 @@ foam.CLASS({
       name: 'postButton',
       label: 'Send POST Request',
       code: async function() {
-        if ( this.postURL !== '' ) {
+        if ( this.sugarURL !== '') {
```
kgrgreer: Why was the space removed?

src/foam/nanos/dig/SugarWebAgent.java
```diff
@@ -40,20 +50,33 @@ public void execute(X x) {
...
     HttpServletResponse resp           = x.get(HttpServletResponse.class);
     HttpParameters      p              = x.get(HttpParameters.class);
-    CharBuffer          buffer_        = CharBuffer.allocate(65535);
-    String              serviceName    = p.getParameter("service");
-    String              methodName     = p.getParameter("method");
-    String              interfaceName  = p.getParameter("interfaceName");
+    String              data           = p.getParameter("data");
 
     try {
+      if ( SafetyUtil.isEmpty(data) ) {
```
kgrgreer, line 56: Is this an error if you call a method with no arguments?

hchoura: data contain the service, method, interfaceName and set of parameters. It should not be empty.

src/foam/nanos/dig/SUGAR.js
hchoura: Since we adopt the post method for all SUGAR service. The URL will be not useful. Instead, the bottom will send the request.

src/foam/lib/json/FObjectParser.java
hchoura: Can you explain more?

src/sugars
```diff
@@ -0,0 +1 @@
...": \"checkUser\",\n    \"X\": \"X\",\n\n    \"user\": {\n    \t\"class\":\"foam.nanos.auth.User\",\"id\":1348,\"enabled\":true,\"loginEnabled\":true,\"firstName\":\"admin\",\"middleName\":\"\",\"lastName\":\"admin\",\"organization\":\"\",\"department\":\"\",\"email\":\"admin@nanopay.net\",\"emailVerified\":true,\"type\":\"User\",\"language\":\"en\",\"timeZone\":\"\",\"desiredPassword\":\"\",\"passwordLastModified\":\"2017-12-03T18:02:50.968Z\",\"note\":\"\",\"businessName\":\"\",\"website\":\"\",\"lastModified\":\"2019-11-13T22:28:47.364Z\",\"invited\":false,\"invitedBy\":0,\"previousStatus\":0,\"status\":2,\"jobTitle\":\"\",\"welcomeEmailSent\":false,\"portalAdminCreated\":false,\"createdPwd\":false,\"inviteAttempts\":0,\"operatingBusinessName\":\"\",\"thirdParty\":false,\"PEPHIORelated\":false,\"signUpToken\":\"\",\"deleted\":false,\"deviceToken\":\"\",\"personalTheme\":0,\"twoFactorEnabled\":false,\"lastModifiedBy\":1348,\"createdBy\":0,\"spid\":\"nanopay\",\"group\":\"admin\",\"loginAttempts\":0,\"isPublic\":true,\"nextLoginAttemptAllowedAt\":\"2019-11-13T22:27:44.807Z\",\"integrationCode\":0,\"hasIntegrated\":false,\"compliance\":2},\n    \"permission\":\"menu.read.doc\"\n}"})
```
nanoscott: Please use the new multi-line format here so this is readable.

src/foam/nanos/dig/SUGAR.js
nanoscott: looks like you are using the URL below to perform the POST, no?

src/foam/lib/json/FObjectParser.java
nanoscott: Please follow-up with Kevin directly to address this comment.

PR #2776, opened 2019-11-06
start of crunch documentation

noodlemoodle:


src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,83 @@
+# CRUNCH User Guide and Documentation
+
+&nbsp;
+
+### CRUNCH - Continuous Reactive User Nano-Capability Hierarchy
+
+CRUNCH is a [Knowledge-Based Configuration System](https://en.wikipedia.org/wiki/Knowledge-based_configuration) that configures a user's account with various capabilities based upon the features and functionalities.
```
kgrgreer: Grammar for this sentence is odd at the end. The "based upon the features and functionalities" part.

src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,83 @@
...
+Using CRUNCH we model all the potential capabilities of a user along with various compliance, onboarding, security, and approval requirements of those capabilities. Those models form a knowledge-graph which allows CRUNCH to query for set of capabilities and receive sets of onboarding requirements.
```
kgrgreer: Second sentence is also odd and not clear.

src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,83 @@
...
+CRUNCH uses this knowledge graph to dynamically generate onboarding wizards for the particular configuration or set of capabilities that a user has requested, as well as dynamically add capabilities to a user's account at run-time and provide a gradual onboarding experience.
+For more information on CRUNCH as a solution to the onboarding problem, and some concrete use cases of CRUNCH, please visit the [CRUNCH wiki](https://github.com/nanoPayinc/NANOPAY/wiki/CRUNCH)
+
+#### Capability
+Capability is the core model of CRUNCH, and describes a task that the user can perform, or a permission they are granted. 
+In the liquid use case, they denote a set of actions that a user may or may not perform on an object. For more detail, please visit the README under `liquid/crunch/`
```
kgrgreer: I don't think we should be mentioning liquid here.

src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,83 @@
+# CRUNCH User Guide and Documentation
+
+&nbsp;
+
+### CRUNCH - Continuous Reactive User Nano-Capability Hierarchy
+
+CRUNCH is a [Knowledge-Based Configuration System](https://en.wikipedia.org/wiki/Knowledge-based_configuration) that configures a user's account with various capabilities based upon the features and functionalities.
+Using CRUNCH we model all the potential capabilities of a user along with various compliance, onboarding, security, and approval requirements of those capabilities. Those models form a knowledge-graph which allows CRUNCH to query for set of capabilities and receive sets of onboarding requirements.
+CRUNCH uses this knowledge graph to dynamically generate onboarding wizards for the particular configuration or set of capabilities that a user has requested, as well as dynamically add capabilities to a user's account at run-time and provide a gradual onboarding experience.
+For more information on CRUNCH as a solution to the onboarding problem, and some concrete use cases of CRUNCH, please visit the [CRUNCH wiki](https://github.com/nanoPayinc/NANOPAY/wiki/CRUNCH)
```
kgrgreer: This link is only available inside nanopay. If none of the material is proprietary, then maybe we should move to the FOAM repo.

src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+#### Relationships Between Capabilities and Other Entities
+
+##### Capability - (Prerequisite) Capability
+- Stored in the `prerequisiteCapabilityJunctionDAO`
+- If a Capability has a prerequisite Capability, the user must be GRANTED the prerequisite Capability before being eligible to apply for the Capability
+
+##### Capability - (Deprecated) Capability
+- stored in the `deprecatedCapabilityJunctionDAO`
+- If a Capability has been deprecated by another Capability, it is essentially no longer enabled, and ignored by the system
```
kgrgreer: Users can still have the capability, can't they? It just is no longer selectable/grantable as a new capability?

src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,83 @@
...
+1. Check whether `p` is the name of an enabled and non-deprecated Capability
+2. Try to find a GRANTED and non-EXPIRED UserCapabilityJunction such that `User == u && Capability == p`. If found, return true
+3. Find the non-EXPIRED UserCapabilityJunctions where `User == u`, for each UCJ, check if the Capability is non-DEPRECATED and implies `p`
+    - `implies` is a method on the Capability model. A Capability `c` implies a String `p` if :
+        - `c` has a prerequisite Capability which implies `p`
+        - `p` is in the set of `permissionsGranted` by `c`, or implied by the set of `permissionsGranted` by `c`*.
+        - the name of `c` implies `p`* 
+        - *e.g., `obj.*` implies `obj.read`
+
+A notable difference between this decorator and other AuthService decorators is that if at any point, the result of the check is found to be true, `true` is returned instead of moving on to the next decorators in the AuthService chain. And if at any point, the result of the check is found to be false, the check moves on to return the result of its delegate
```
kgrgreer, line 71: Is this really different?

kgrgreer: I would say that while this AuthService can grant permissions to a user, it won't remove permissions that they receive from other AuthServices in the chain.

src/foam/nanos/crunch/README.md
noodlemoodle: that's true. Sorry, I keep thinking deprecated means something else.. also had to updated some logic in capabilityauthservice to reflect this

src/foam/nanos/crunch/README.md
noodlemoodle: I think other decorators that override check does the opposite of this where it goes on to check the result of its delegates when the result of the current check is found to be true, and returns false when the result of current is found to be false. In this case, we are returning true on true and returning the delegate on false since the delegate will not return true for crunch-specific permission strings.

src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,82 @@
+# CRUNCH User Guide and Documentation
+
+&nbsp;
+
+### CRUNCH - Continuous Reactive User Nano-Capability Hierarchy
+
+CRUNCH is a [Knowledge-Based Configuration System](https://en.wikipedia.org/wiki/Knowledge-based_configuration) that configures a user's account with various capabilities based upon the features and functionalities.
```
noodlemoodle:
```suggestion
CRUNCH is a [Knowledge-Based Configuration System](https://en.wikipedia.org/wiki/Knowledge-based_configuration) that configures a user's account with various capabilities.
```

src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,82 @@
+# CRUNCH User Guide and Documentation
+
+&nbsp;
+
+### CRUNCH - Continuous Reactive User Nano-Capability Hierarchy
+
+CRUNCH is a [Knowledge-Based Configuration System](https://en.wikipedia.org/wiki/Knowledge-based_configuration) that configures a user's account with various capabilities.
+Using CRUNCH we model all the potential capabilities of a user along with various compliance, onboarding, security, and approval requirements of those capabilities. Those models form a knowledge-graph which allows CRUNCH to query for set of capabilities and receive sets of onboarding requirements.
```
noodlemoodle:
```suggestion
Using CRUNCH we model all the potential capabilities of a user along with various compliance, onboarding, security, and approval requirements of those capabilities. Those models form a knowledge-graph which allows CRUNCH to generate onboarding requirements based on the individual's needs and existing capabilities.
```

src/foam/nanos/crunch/README.md
noodlemoodle: I'm not sure that it's non-proprietary, will remove this for now

src/foam/nanos/crunch/README.md
```diff
@@ -0,0 +1,82 @@
+# CRUNCH User Guide and Documentation
+
+&nbsp;
+
+### CRUNCH - Continuous Reactive User Nano-Capability Hierarchy
+
+CRUNCH is a [Knowledge-Based Configuration System](https://en.wikipedia.org/wiki/Knowledge-based_configuration) that configures a user's account with various capabilities.
+Using CRUNCH we model all the potential capabilities of a user along with various compliance, onboarding, security, and approval requirements of those capabilities. Those models form a knowledge-graph which allows CRUNCH to generate onboarding requirements based on the individual's needs and existing capabilities.
+CRUNCH uses this knowledge graph to dynamically generate onboarding wizards for the particular configuration or set of capabilities that a user has requested, as well as dynamically add capabilities to a user's account at run-time and provide a gradual onboarding experience.
+For more information on CRUNCH as a solution to the onboarding problem, and some concrete use cases of CRUNCH, please visit the [CRUNCH wiki](https://github.com/nanoPayinc/NANOPAY/wiki/CRUNCH)
```
noodlemoodle:
```suggestion
```

PR #2778, opened 2019-11-07
Predicates

nanoMichal:


jlhughes (approved):

src/foam/nanos/ruler/predicate/PropertyEQProperty.js
```diff
@@ -0,0 +1,51 @@
...
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: `
+        if ( getIsNew() ) {
+          FObject nu  = (FObject) NEW_OBJ.f(obj);
+          return EQ(nu.getClassInfo().getAxiomByName(getProp1()), nu.getClassInfo().getAxiomByName(getProp2())).f(nu);
+        }
+        else {
```
kgrgreer: No need for the 'else' if you return'ed in the 'if'.

PR #2784, opened 2019-11-08
EasyDAO made Easy

michaelconnor468:


jlhughes (approved):

src/foam/dao/EasyDAO.js
```diff
@@ -862,18 +823,31 @@ return this;
...
         this.mdao && this.mdao.addIndex.apply(this.mdao, arguments);
         return this;
       },
       javaCode: `
-if ( getMdao() != null ) {
-  getMdao().addIndex(index);
-}
-return this;
-`
+          if ( getMdao() != null ) {
```
TW80000: These three lines are indented one level too deep.

src/foam/dao/EasyDAO.js
```diff
@@ -862,18 +823,31 @@ return this;
...
+            getMdao().addIndex(index);
+          }
+        return this;
+      `
     },
+    {
+      name: 'printDecorators',
```
TW80000, line 832: Where does this get called from? Note that we also have a script called "Show Decorators" that will do this.

src/foam/dao/EasyDAO.js
```diff
@@ -325,20 +301,20 @@ return delegate;
...
-      /** Enable local in-memory caching of the DAO. */
+      documentation: 'Enable local in-memory caching of the DAO',
       class: 'Boolean',
       name: 'cache',
       generateJava: false,
       value: false
     },
     {
-      /** Enable standard authorization. */
+      documentation: 'Enable standard authorization',
```
TW80000: It's not necessarily standard anymore, it might use `AuthorizableAuthorizer` if the model implements `Authorizable`. So this should just be `Enable authorization`.

src/foam/dao/EasyDAO.js
michaelconnor468: the dao itself; since EasyDAO is the end decorator for all daos once we migrate you just have to do a dao.printDecorators() either in a service script or anywhere in the source code for debugging

TW80000 (approved):

PR #2790, opened 2019-11-11
Add enabledaware to nspec

xuerongNanopay:
Add enabledaware to NSpec, which allow different instance to config its own services

### DIFF
--- a/src/foam/nanos/boot/NSpec.js
+++ b/src/foam/nanos/boot/NSpec.js
@@ -6,6 +6,9 @@
 foam.CLASS({
   package: 'foam.nanos.boot',
   name: 'NSpec',
+  javaImplements: [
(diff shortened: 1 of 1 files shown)


xuerongNanopay:
> @googlebot I signed it!

src/foam/nanos/boot/NSpec.js
```diff
@@ -160,6 +163,20 @@ foam.CLASS({
       },
       readPermissionRequired: true,
       writePermissionRequired: true
+    },
+    {
+      class: 'Boolean',
+      name: 'enabled',
+      value: true,
+      view: {
```
jlhughes: The view is not needed. We already have a view for Boolean.

jlhughes (approved):

PR #2792, opened 2019-11-11
Add Thread.currentThread id to Log messages.

jlhughes:


src/foam/nanos/logger/DAOLogger.js
```diff
@@ -76,7 +76,10 @@ if ( initializing.get() ) {
   System.out.println("DAOLogger initializing");
   return;
 }
-LogMessage lm = new LogMessage.Builder(getX()).setSeverity(severity).setMessage(message).build();
+LogMessage lm = new LogMessage.Builder(getX())
+                    .setThreadName(String.valueOf(Thread.currentThread().getId()))
```
TW80000: Is this necessary? Shouldn't the factory do this for you?

TW80000 (approved):

src/foam/nanos/logger/DAOLogger.js
jlhughes: yes. updated.

PR #2793, opened 2019-11-12
Fix bug where objects couldn't be updated via DAO Controller

TW80000:
## High Level

I had only tested create yesterday, not viewing or updating, and
because of that I didn't realize that my changes had broken viewing and
editing via the GUI.

This commit fixes the issue.

## Technical

Now we only use FObjectView in the create controller, not the update
controller. This allows users to select the subclass when creating a new
instance, but not when editing an existing object.

To respect the DAO Controller's feature of a custom detail view, this
commit updates FObjectView so that it supports changing the view used
for the FObject. In DAOCreateControllerView we set that view to the
custom detail view for the DAO controller if it's available.

### DIFF
--- a/src/foam/comics/BrowserView.js
+++ b/src/foam/comics/BrowserView.js
@@ -116,8 +116,8 @@ foam.CLASS({
         if ( searchMode )  config.searchMode  = searchMode;
         if ( subtitle )    config.subtitle    = subtitle;
         if ( title )       config.title       = title;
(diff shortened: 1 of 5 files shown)


KernelDeimos:
What should we do if we want to replace a property of an existing object with something else? For example, in the Cron UI I'd like to have it be possible to change IntervalSchedule to TimeOfDaySchedule without re-creating the Cron job.

PR #2794, opened 2019-11-12
when mode is RO or DISABLED, the label should not clickable

yij793:
origin ticket address:  https://nanopay.atlassian.net/browse/CPF-3393
change: add condition to the click event, when the mode for checkbox is RO or DISABLED, the label should unclickable.
@TW80000 
@jlhughes

### DIFF
--- a/src/foam/u2/CheckBox.js
+++ b/src/foam/u2/CheckBox.js
@@ -56,6 +56,7 @@ foam.CLASS({
                       this.labelFormatter,
                       function() { this.add(self.label$); })
           .on('click', function() {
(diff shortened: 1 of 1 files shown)


src/foam/u2/CheckBox.js
```diff
@@ -56,6 +56,7 @@ foam.CLASS({
                       this.labelFormatter,
                       function() { this.add(self.label$); })
           .on('click', function() {
+          if (self.mode === foam.u2.DisplayMode.RO || self.mode === foam.u2.DisplayMode.DISABLED) return;
```
kgrgreer: Please fix indentation.

TW80000:
```suggestion
            if ( self.getAttribute('disabled') ) return;
```

PR #2795, opened 2019-11-12
Add StrategyReferences for Schedule selection

jlhughes:

PR #2797, opened 2019-11-13
Cron UI and FObjectView improvements

TW80000:
* Add strategy references for cron schedules
* Fallback on 'of' when no strategies available
* Sort class dropdown in `FObjectView` alphabetically
* Add missing detach listener
* Use `FObjectView` as default write view for `FObjectProperty`-ies
* Set default choice in `FObjectView` to first choice
* Fix bug where default value of `ChoiceView` wasn't being respected
* Put `TimeHSM` UI inputs inline to save screen space


TW80000:
I think this covers https://github.com/foam-framework/foam2/pull/2795.

PR #2798, opened 2019-11-14
Sign up sign in

Nauna:


src/files.js
```diff
@@ -191,6 +191,7 @@ FOAM_FILES([
   { name: 'foam/u2/view/StringArrayRowView', flags: ['web'] },
   { name: 'foam/u2/borders/NullBorder', flags: ['web'] },
   { name: 'foam/u2/borders/CardBorder', flags: ['web'] },
+  { name: 'foam/u2/borders/SplitScreenBorder', flags: ['web'] },
```
Nauna, line 194: This was SplitBorder in business code side. Moved it to foam. Business code branch (https://github.com/nanoPayinc/NANOPAY/pull/8340) deletes legacy file

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -137,6 +137,16 @@ foam.CLASS({
...
   properties: [
+    {
+      name: 'loginVariables',
+      expression: function(client$userDAO) {
+        return {
+          dao_: client$userDAO || null,
+          imgPath: '',
+          group: 'system'
+        };
+      }
```
Nauna, line 150: This is used to generate the differences in SignIn/SignUp dynamically based on the application. 

group here uses 'system' which seems scary to let anyone sign up via a system access, but it's the only group in set in foam.

src/foam/u2/TextInputCSS.js
```diff
@@ -16,7 +16,7 @@ foam.CLASS({
       name: 'code',
       value: `
         ^ {
-          height: %INPUTHEIGHT%;
+          height: /*%INPUTHEIGHT%*/ 32px;
```
Nauna, line 19: fixes an issue with the country drop down not rendering to the same size as other fields.

src/menus
```diff
@@ -12,7 +12,7 @@ p({"class":"foam.nanos.menu.Menu","id":"admin.regions","label":"Regions","handle
 p({"class":"foam.nanos.menu.Menu","id":"logs","label":"Logs","handler":{"class":"foam.nanos.menu.DAOMenu","title":"Server Log","createEnabled":false,"daoKey":"logMessageDAO"},"order":1000,"parent":"admin"})
 p({"class":"foam.nanos.menu.Menu","id":"reset","label":"Reset Password","handler":{"class":"foam.nanos.menu.ViewMenu"},"order":1000,"parent":"hidden"})
 p({"class":"foam.nanos.menu.Menu","id":"admin.scripts","label":"Scripts","handler":{"class":"foam.nanos.menu.DAOMenu","daoKey":"scriptDAO"},"order":1000,"parent":"admin"})
-p({"class":"foam.nanos.menu.Menu","id":"sign-in","label":"Sign in","handler":{"class":"foam.nanos.menu.ViewMenu","view":{"class":"foam.nanos.auth.SignInView"}},"order":1000,"parent":"hidden"})
+p({"class":"foam.nanos.menu.Menu","id":"sign-in","label":"Sign in","handler":{"class":"foam.nanos.menu.ViewMenu","view":.stack.push({ "class": "foam.u2.view.LoginView", "mode_": \"SignIn\" },"order":1000,"parent":"hidden"})
```
zahnno: Why the ".stack.push"?

src/foam/u2/view/LoginView.js
```diff
@@ -0,0 +1,309 @@
...
+    -webkit-transform: scale(-1, 1);
+    -o-transform: scale(-1, 1);
+    -ms-transform: scale(-1, 1);
+    transform: scale(-1, 1);
+    margin-right: 10px;
+  }
+  /* *** ON SMESTYLES ??? */
+  .inline-block {
+    display: inline-block;
+  }
```
zahnno: Is there a reason why these styles aren't contained in a css class?

It might make more sense to place these in AppStyles.js if you want them applied throughout the application.

src/foam/u2/view/LoginView.js
```diff
@@ -0,0 +1,309 @@
...
+  properties: [
+    {
+      class: 'Boolean',
+      name: 'topBarShow_',
+      factory: function() {
+        if ( this.appConfig && this.appConfig.url ) {
+          return true;
+        }
+        return false;
+      },
```
zahnno, line 143: Why is this based on the appConfig being present?

src/foam/u2/view/LoginView.js
```diff
@@ -0,0 +1,309 @@
...
+  ],
+
+  messages: [
+    { name: 'GO_BACK', message: 'Go to ' },
+    { name: 'MODE1', message: 'SignUp' }
+  ],
+
+  methods: [
+    function init() {
+      if ( ! this.param ) {
```
zahnno: This could be moved to the `param` properties factory

src/foam/nanos/controller/ApplicationController.js
zahnno: Why are we containing this within a generic key value. It might be useful to model this. 

Also is there any reason why you're making its value change based on client$userDAO? If I've defined my own loginVariables when tagging or creating my applicationController, my imgPath and group would be overwritten when the userDAO is instantiated on the client. 

I think I know what you're trying to do here but I don't think this is the best place for it. It might make more sense adding these key values as props on the LoginView.

src/foam/u2/view/LoginView.js
```diff
@@ -0,0 +1,309 @@
...
+        this.param = {};
+      }
+      this.param.dao_ = ! this.param.dao_ ? this.loginVariables.dao_ : this.param.dao_;
+      if ( this.mode_ === this.MODE1 ) {
+        this.param.group_ = ! this.param.group_ ? this.loginVariables.group_ : this.param.group_;
+        this.param.countryChoices_ = this.param.countryChoices_ ? this.param.countryChoices_ : ['CA', 'US'];
+        this.model = this.SignUp.create(this.param, this);
+      } else {
+        this.model = this.SignIn.create(this.param, this);
+      }
```
zahnno, line 211: Why is this condition required? Couldn't we create the model with params passed into the model we're creating?

So { class: 'LoginView', model: 'foam.u2.SignUp', group: 'sme', countryChoices: ['ca','us'] }

src/foam/u2/view/LoginView.js
```diff
@@ -0,0 +1,309 @@
...
+      // second footer
+        .start().br()
+          .start('span').addClass('bold-text-with-pad').add(this.model.SUB_FOOTER_TXT).end()
+          .start('span').addClass('link')
+            .add(this.model.SUB_FOOTER_LINK)
+            .on('click', () => {
+              this.model.subfooterLink();
+            })
+          .end()
+        .end();
```
zahnno, line 250: I think we could support this by making footer link actions or subFooter properties with views for click/styling within the sign in/sign up model allowing us to removing this logic and allowing us to extend or replace footer views with whatever we'd like.

src/foam/u2/view/LoginView.js
```diff
@@ -0,0 +1,309 @@
...
+      this.addClass(this.myClass())
+      // full width bar with navigation to app landing page
+        .start().addClass('top-bar-nav').show(this.topBarShow_)
+          .start()
+            .start().addClass('topBar-txt-link')
+              .start('span')
+                .addClass('horizontal-flip')
+                .addClass('inline-block')
+                .add('➔')
+              .end()
```
zahnno, line 270: This allows us to show and hide the topbar you've created but what if I wanted to implement my own top bar?

src/foam/u2/view/LoginView.js
```diff
@@ -0,0 +1,309 @@
...
+          temp.substring(temp.indexOf('://') + 3) :
+          temp;
+      }
+    },
+    {
+      class: 'String',
+      name: 'backLink_',
+      factory: function() {
+        return this.model.backLink_ ? this.model.backLink_ : this.appConfig.url;
+      },
```
zahnno, line 190: Shouldn't this be an expression taking in `model` as an argument? Considering the idea to have the loginView be instantiated once and toggle between the two states, sign in and sign up.

src/foam/u2/view/LoginView.js
```diff
@@ -0,0 +1,309 @@
...
+        .start().show(this.topBarShow).addClass('topBar-logo-Back')
+          .start('img')
+            .attr('src', this.theme.logo)
+            .addClass('top-bar-img')
+          .end()
+      .end()
+      // Title txt and Model
+        .start().addClass('title-top').add(this.model.TITLE).end()
+        .startContext({ data: this })
+          .addClass('content-form').tag(this.MODEL).br()
```
zahnno, line 229: Why not make this into a slot so we don't have to push new loginView to update the page? It would make this view a lot more dynamic. I'd like to see that logic applied with the rest of this initE.

zahnno (changes requested):
I've made my initial review after reading the loginView to understand the overlying concept in this design and would like to further discuss how we can improve it. I've left initial comments indicating some concerns and ideas.

src/foam/nanos/controller/ApplicationController.js
Nauna: 1) hmmm ... model the loginVariables... I like it. Good call.
2) to your point on the second paragraph. group and imgPath don't change- hardcoded. But since the new plan is to model this, then moot point.
3) props on loginView, doesn't work, these values change based on application, thus need applicationControllers to define them. 

But with Model, I see this working out to be a model, a dao, and journals ... we could even permission the selects - potentially opening up switching views from one applicationController and/or using crunch. I'm even now thinking, move some of the disableFieldBooleans_ to this jrl and just do a select on this loginVarDao when handling tokens. ... hmmm

src/menus
Nauna: not suppose to be there! Need to push that change. SORRY.

src/foam/u2/view/LoginView.js
Nauna: What makes you think I want this throughout the application? This is very specific to this view, I do think we need to organize our css style. But didn't want to go down that rabbit hole. Considering this is used from FOAM, connect and Ablii ... I'd have to map all our css styles related to desired css. Function over Form on this.

src/foam/u2/view/LoginView.js
Nauna: Good catch, it should be this.backLink_, I made some adjustments while coding.

This top bar is from Ablii's design to return to landing page-  thus if no landing page defined, don't show topBar.

src/foam/u2/view/LoginView.js
Nauna: Interesting point!

Nauna: Didn't have the use case, but you are right. Thx

src/foam/u2/view/LoginView.js
Nauna: lol ... another good point. Thx

src/foam/u2/view/LoginView.js
Nauna: Need to test something before I say anything. Appreciate you drawing attention to this.

src/foam/u2/view/LoginView.js
Nauna: Not sure what you are implying here? this.model is a slot, and if we wanted to not call LoginView, then how would that work? I see each view as a page, calling LoginView for two separate scenarios and for two seperate pages. If we don't call LoginView then  ... ???

This does bring a point though, in the footerLink(), I wouldn't have to call LoginView, if I was aware of the calling view in the model ... food for thought... will consider this. But otherwise, don't see away to avoid calling LoginView.

src/foam/u2/view/LoginView.js
Nauna: This footer stuff is perhaps not named correctly. They are links that exist in signUp/SignIn they are base functions for this view. Considering all previous views shared these ... not sure I agree with switching these.

image


image

src/foam/u2/view/LoginView.js
Nauna: Currently connect does this:
image

Nauna: above img, shows connect with topBarShow_ = false
below img, shows Ablii with topBarShow_ = true

image

src/foam/nanos/controller/ApplicationController.js
zahnno: I think placing it on Theme.js would work out.

src/foam/u2/view/LoginView.js
zahnno: I'm assuming because if LoginInView is instantiated, it'll apply it's css onto the application. Since these style aren't contained they would affect all elements that have defined inline-block or horizontal-flip as a class.

src/foam/u2/view/LoginView.js
zahnno: Sure, but I wouldn't be able to create my own top bar and have it displayed. Implementing this can really help us customize our login view instead of strictly coding it in within the initE of the loginView.

src/foam/u2/view/LoginView.js
zahnno: How about defaulting the property to false and have the developer decide if they want it or not? Also, consider my top bar comment on allow customization.

src/foam/u2/view/LoginView.js
zahnno: Disregard my comment. I'm was thinking we'd have a constant reference to a loginView on the applicationController but that's not the best idea.

src/foam/u2/view/LoginView.js
zahnno: I'm suggesting this idea because this sub footer can be extensively different for each application. We shouldn't be limiting it to our use case. "Not a user yet?" might be a different phrase for another application OR an application might want to add an image instead of text.

Nauna: > "Not a user yet?" might be a different phrase for another application

can be set with the model's axiom(message) name: FOOTER_TXT

this model can be overridden with extending the base model and replacing the file in the applicationController that wants to use it instead EX) this.__subContext__.register(this.SignUp, 'foam.nanos.u2.navigation.SignUp');

Nauna: > OR an application might want to add an image instead of text.

can't account for every possibility. Will come to that when we come to that. This works, and follows our current design.

src/foam/u2/view/LoginView.js
Nauna: can't account for every possibility. Will come to that when we come to that. This works, and follows our current design.

src/foam/nanos/controller/ApplicationController.js
Nauna: ^ Can do later: https://nanopay.atlassian.net/browse/CPF-3588

src/foam/u2/view/LoginView.js
Nauna: What your commenting on is the default behavior. EX)In net.nanopay.ui.Controller line:308, I have explicitly set this to false.

src/foam/u2/view/LoginView.js
Nauna: Don't know what I was agreeing with. tbh. I don't see this working. 
The view is instantiated on each call to either signUp or SignIn.
To have one state stored between two models ... don't see any benefits with what is being suggested. I am reusing one view, but these are different things.

src/foam/u2/view/LoginView.js
Nauna: cleanup

jlhughes:
tested - looks good.  Also did a change password. 
combined it with the menu'ing system and all works well. 
Haven't done a code review yet.

PR #2799, opened 2019-11-14
Do not send email already in FAILED state.

jlhughes:


TW80000 (approved):
This seems fine to me. I think it would be valuable to track down what's trying to send a failed email in the first place though and fix that. Get at the root of the problem.

jlhughes:
Yes, there should be a cron job that should be controlling this - on it's select.

jlhughes:
cronjob is correct:

`    emailMessages = ((ArraySink)
      emailMessageDAO
        .where(EQ(EmailMessage.STATUS, Status.UNSENT))
`

Ok, would like this a a failsafe for the time being.

PR #2800, opened 2019-11-14
Add t to error log to generate stack trace in log files to help troub…

jlhughes:
…leshoot session server box issues.

### DIFF
--- a/src/foam/box/SessionServerBox.java
+++ b/src/foam/box/SessionServerBox.java
@@ -124,7 +124,7 @@ public class SessionServerBox
 
       msg.getLocalAttributes().put("x", effectiveContext);
     } catch (Throwable t) {
-      logger.error("Error throw in SessionServerBox: " + t, " ,service: " + spec.getName());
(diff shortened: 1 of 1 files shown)


src/foam/box/SessionServerBox.java
```diff
@@ -124,7 +124,7 @@ public void send(Message msg) {
 
       msg.getLocalAttributes().put("x", effectiveContext);
     } catch (Throwable t) {
-      logger.error("Error throw in SessionServerBox: " + t, " ,service: " + spec.getName());
+      logger.error("Error throw in SessionServerBox: " + t, " ,service: " + spec.getName(), t);
```
TW80000, line 127: Isn't `t` already included in the log earlier in this line?

jlhughes: Suspect the + t performs a + t.toString() . rather then the intended    , t ,   which would have generated the stack trace.

TW80000 (approved):

PR #2801, opened 2019-11-14
adding a background colour to be used with our logos

Nauna:
one logo we have has white font, that is intended to have a background colour.

so depending on the logo and theme, we need a background colour.

### DIFF
--- a/src/foam/nanos/theme/Theme.js
+++ b/src/foam/nanos/theme/Theme.js
@@ -93,7 +93,7 @@ foam.CLASS({
     },
     {
       class: 'Color',
-      name: 'logoBackgrounColour',
+      name: 'logoBackgroundColour',
       documentation: 'The logo background colour to display in the application.',
       section: 'colours'
     },
--- a/src/foam/nanos/theme/Theme.js
+++ b/src/foam/nanos/theme/Theme.js
@@ -92,6 +92,12 @@ foam.CLASS({
       documentation: 'Uses largeLogo image on various views instead of logo.'
     },
     {
+      class: 'Color',
+      name: 'backgroundLogo',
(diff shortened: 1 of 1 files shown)


src/foam/nanos/theme/Theme.js
```diff
@@ -91,6 +91,12 @@ foam.CLASS({
       name: 'largeLogoEnabled',
       documentation: 'Uses largeLogo image on various views instead of logo.'
     },
+    {
+      class: 'Color',
+      name: 'backgroundLogo',
```
jlhughes: Should it be called logoBackground or logoBackgrounColour?  else I would think this controls some 'background' logo.

Nauna: done

src/foam/nanos/theme/Theme.js
```diff
@@ -91,6 +91,12 @@ foam.CLASS({
       name: 'largeLogoEnabled',
       documentation: 'Uses largeLogo image on various views instead of logo.'
     },
+    {
+      class: 'Color',
+      name: 'logoBackgrounColour',
```
TW80000: Typo

```suggestion
      name: 'logoBackgroundColour',
```

TW80000 (changes requested):

TW80000:
Shouldn't `BusinessLogoView` be updated to respect this new property as well?

Nauna:
> Shouldn't `BusinessLogoView` be updated to respect this new property as well?

@TW80000 felt the response required some follow-up work so I moved my response to a ticket.
https://nanopay.atlassian.net/browse/CPF-3443

PR #2802, opened 2019-11-14
Improve Capability Detail View

TW80000:
* Groups inputs under two sections
* Made `id` RO in update mode
* Made `daoKey` editable
* Code formatting fixes

### DIFF
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -40,21 +40,35 @@ foam.CLASS({
     'daoKey'
   ],
 
+  sections: [
+    {
+      name: '_defaultSection',
+      title: 'Administrative'
+    },
+    {
+      name: 'uiSettings',
+      title: 'UI Settings',
+      help: 'These properties are used to control how this capability appears in the GUI.'
+    }
+  ],
+
   properties: [
-    
     {
       name: 'id',
-      class: 'String'
+      class: 'String',
+      updateMode: 'RO'
     }, 
     {
       name: 'icon',
       class: 'Image',
-      documentation: `Path to capability icon`
+      documentation: `Path to capability icon`,
+      section: 'uiSettings'
     },
     {
       name: 'description',
       class: 'String',
-      documentation: `Description of capability`
+      documentation: `Description of capability`,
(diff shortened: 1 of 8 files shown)


jlhughes (approved):

src/foam/nanos/crunch/Capability.js
```diff
@@ -39,21 +40,35 @@ foam.CLASS({
...
+    }
+  ],
+
   properties: [
-    
     {
       name: 'id',
-      class: 'String'
+      class: 'String',
+      updateMode: 'RO'
```
KernelDeimos, line 59: @TW80000 I've seen a lot of `id` properties without `updateMode: 'RO'`, but I also understand why it makes sense to do this. When I look for this in other code reviews I'll be doing, are there any situations you can think of where `id` shouldn't be read-only?

src/foam/nanos/crunch/crunchtest/CapabilityTest.java
```diff
@@ -80,27 +73,31 @@ public void runTest(X x) {
     testCapability(x);
   }
 
+  // TODO: Actually test something here.
```
KernelDeimos, line 76: This comment could be more specific so the call to `test` below doesn't cause someone to mistakenly think it's outdated.

KernelDeimos (changes requested):

src/foam/nanos/crunch/Capability.js
kgrgreer: In lots of places (scripts, menus, etc.) you get to select your id.

TW80000: It's fine for `id` to have `createMode: 'RW'` (which is the default), but it doesn't really ever make sense for id to be editable in update mode.

Now that I think about it, for sequence number or GUID ids, `createMode: 'HIDDEN'` would actually be more useful since we don't want to let users set the id to something unexpected and the id will be set for them by the backend anyway.

Arguably it does make sense in certain places to let the user specify the id though, usually where ids are strings. In that case you may want to change the id of a Script or a Group. Note that this will create a new object rather than updating an existing one though.

PR #2803, opened 2019-11-14
Add permissions for strategyDAO and strategyreference.read

jlhughes:


TW80000 (commented):
The services are already defined in `src/services`. I put them there because Mike told me that `.jrl` files in other directories aren't picked up when you build FOAM alone (ie: not using the nanopay build script).

jlhughes:
we should fix foam.  I'll rename them.

TW80000 (approved):

PR #2805, opened 2019-11-15
removed permission prefix from standard/authorizable authorizers and …

noodlemoodle:
…authorizer interface

### DIFF
--- a/src/foam/nanos/auth/AuthorizableAuthorizer.java
+++ b/src/foam/nanos/auth/AuthorizableAuthorizer.java
@@ -18,10 +18,6 @@ public class AuthorizableAuthorizer implements Authorizer {
     permissionPrefix_ = permissionPrefix;
   }
 
-  public String getPermissionPrefix() {
-    return permissionPrefix_;
-  }
-
   public void authorizeOnCreate(X x, FObject obj) throws AuthorizationException {
     if ( obj instanceof Authorizable ) {
       ((Authorizable) obj).authorizeOnCreate(x);
--- a/src/foam/nanos/auth/Authorizer.js
+++ b/src/foam/nanos/auth/Authorizer.js
@@ -99,10 +99,6 @@ access to an FObject under different circumstances.`,
         },
       ],
       javaType: 'boolean'
-    },
-    {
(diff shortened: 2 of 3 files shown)


src/foam/nanos/auth/AuthorizableAuthorizer.java
```diff
@@ -18,10 +18,6 @@ public AuthorizableAuthorizer(String permissionPrefix) {
     permissionPrefix_ = permissionPrefix;
   }
 
-  public String getPermissionPrefix() {
-    return permissionPrefix_;
```
TW80000, line 22: I'm thinking it makes sense to get rid of `permissionPrefix_` altogether here. However, that would break the current implementation of `checkGlobalRead` and `checkGlobalRemove`.

So the question then becomes whether `checkGlobalRead` and `checkGlobalRemove` should be part of the `Authorizer` interface.

I think the answer is "no" from a logical perspective, since you can always get equivalent behaviour from `authorizeOnRead` and `authorizeOnDelete`, making the `checkGlobal*` methods redundant from a logical perspective. However, the performance impact of removing them is huge, so while the program logic doesn't change, real-world performance changes significantly, which cannot be ignored.

So either we need to keep the logically redundant methods and try and think of a meaningful implementation of them for different `Authorizer`s, or we need to think of a different way to keep the performance speed up without including those methods on the interface.

Another piece of the puzzle is that we have `GlobalReadAuthorizer`, which always returns true for `checkGlobalRead`, and we want to keep that behaviour.

So I think we need to keep the `checkGlobal*` methods as part of the interface. The current implementation of those methods for `AuthorizableAuthorizer` doesn't seem right to me though. Permissions are checked by `StandardAuthorizer` and `GlobalReadAuthorizer`, whereas `AuthorizableAuthorizer` should delegate to the `Authorizable` object. That would mean that `Authorizable` would need the `checkGlobal*` methods as well though, which I'm not sure makes sense either.

So, even though I'd like to remove `permissionPrefix_` here, I can't think of an alternative good enough to justify the effort it would take to refactor things right now, so we'll leave it for now.

TW80000 (approved):

PR #2806, opened 2019-11-15
using style width and height instead of displayHeight and displayWidt…

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3446
Note: the reason causing this issue is the `displayHeight` and `displayWidth` are responsive in safari and will change the size of error icon depending on the length of text.  this change will fix the width and height for error icon which is 16px.


yij793:
I closed this since this issue only happen in staging. need wait for comfirm

PR #2808, opened 2019-11-15
finished method addDecorateAfter

michaelconnor468:


src/foam/dao/EasyDAO.js
```diff
@@ -828,6 +828,57 @@ foam.CLASS({
...
+          name: 'decorator',
+          javaType: 'foam.dao.ProxyDAO'
+        },
+        {
+          documentation: 'Decorator in the EasyDAO chain to place in relation to',
+          name: 'location',
+          javaType: 'foam.core.ClassInfo'
+        },
+        {
+          documentation: 'If true, decorator chain placed before the dao instead of inbetween the supplied dao and mdao',
```
jlhughes, line 847: Assuming this property came later during development.  Can we rename class to addDecorator and then the 'reverse' property becomes 'before' or 'after'.

michaelconnor468: sure

michaelconnor468: "Can we rename class to addDecorator" can you elaborate a bit on this though; I'm not sure which class to rename

jlhughes: Sorry, that should have read 'method', so addDecoratorAfter to addDecorator

jlhughes (approved):

PR #2809, opened 2019-11-18
remove unused method

michaelconnor468:


TW80000 (commented):
Looks like the JavaScript implementation gets called here: https://github.com/foam-framework/foam2/blob/master/test/src/lib/dao.js#L1012

I don't know if `EasyDAO` is even in a working state on the JavaScript side anyway. @kgrgreer, @jlhughes any thoughts on whether to remove this or not? We have both this `addIndex` method and an `index` property which appear to have the same purpose.

jlhughes (approved):

jlhughes:
Travis CI failure:
`1) EasyDAO-permutations forwards addPropertyIndex
  Message:
    TypeError: dao.addIndex is not a function
  Stack:
    TypeError: dao.addIndex is not a function
        at UserContext.<anonymous> (/home/travis/build/foam-framework/foam2/test/src/lib/dao.js:1012:9)
Pending:
`

jlhughes:
1. setIndex - takes an array of FObjectProperty, creates an index for each via MDAO.addIndex(FObjectProperty) - **only for Java**
2. addPropertyIndex - takes var arg of PropertyInfos and creates index for each via MDAO.addIndex(PropertyInfo) - for both javascript and Java
3. addIndex - takes a single 'Index' - via MDAO.addIndex(Index) - for both javascript and java. 

From an nspec configuration perspective, all are useful and 'addIndex' is available if you really need complete control over the Index creation.   So regardless of the state of javascript, no need to remove this.

PR #2810, opened 2019-11-19
Added phone number property for data migration from phone object

skanderturki:
Added duplicate phone number property for data migration

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -241,13 +241,25 @@ foam.CLASS({
     },
     {
       class: 'String',
-      name: 'phoneNumber',
+      name: 'phoneNumberOld',
       transient: true,
       documentation: `Omits properties of the phone number object and returns
         the phone number.`,
       expression: function(phone) {
         return phone.number;
       },
+      section: 'personal'
+    },
+    {
+      class: 'PhoneNumber',
+      name: 'phoneNumber',
+      documentation: 'Personal phone number.',
+      section: 'personal'
+    },
+    {
+      class: 'Boolean',
+      name: 'phoneNumberVerified',
+      writePermissionRequired: true,
       section: 'personal'
     },
     {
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/User.js
```diff
@@ -261,6 +273,18 @@ foam.CLASS({
       view: { class: 'foam.u2.detail.VerticalDetailView' },
       section: 'personal'
     },
+    {
+      class: 'PhoneNumber',
+      name: 'mobile',
```
TW80000: This name conflicts with the other property named "mobile".

src/foam/nanos/auth/User.js
```diff
@@ -241,7 +241,7 @@ foam.CLASS({
     },
     {
       class: 'String',
-      name: 'phoneNumber',
+      name: 'phoneNumberOld',
```
TW80000: This isn't going to work because when we restart the server after applying the changes, JDAO won't know that the values previously in `phoneNumber` now belong in `phoneNumberOld`. Instead of changing the existing name, we need to leave the existing name as is and add a new one to copy the data to. So the steps should look like this:

## 1. Current state

* `phone :: FObjectProperty of Phone`
* `phoneNumber :: String`
  * `transient: true`
* `mobile :: FObjectProperty of Phone`

## 2. After first set of changes (first PR merged)

* `phone :: FObjectProperty of Phone` (**unchanged**)
* `phoneNumber :: PhoneNumber` (**changed**)
  * `transient: false` (this is the default, so rather than changing to false, can just remove the line that sets transient to true)
  * We didn't need a new property in this case, we can just modify this one to suit our needs since it already shares the same name
  * This is where we'll copy the existing data to from `phone.number`
* `mobile :: FObjectProperty of Phone` (**unchanged**)
* `mobileNumber :: PhoneNumber` (**new**)
  * This is where we'll copy the existing data to from `mobile.number`

A new Script will be included in `scripts.jrl` that loops over all existing users and copies the phone and mobile data over to the new properties and saves the changes.

## 3. After the second set of changes (second PR merged)

* ~`phone :: FObjectProperty of Phone`~ (**removed, all references updated**)
* `phoneNumber :: PhoneNumber` (**unchanged**)
* ~`mobile :: FObjectProperty of Phone`~ (**removed, all references updated**)
* `mobileNumber :: PhoneNumber` (**unchanged**)

---

I didn't include the `*Verified` properties in the above explanation, but they should be self-explanatory if you understand the above.

TW80000 (commented):
Please see my comments. This PR should also include the migration script. If you want an explanation on how to write a migration script let me know.

src/foam/nanos/auth/User.js
skanderturki: ok thanks

TW80000:
This PR LGTM. Had a few comments on the nanopay PR though: https://github.com/nanoPayinc/NANOPAY/pull/8377#pullrequestreview-321144518

PR #2812, opened 2019-11-19
CPF-3177 Create FObjectPropertyFilterView

carl-zzz:
# JIRA Ticket: 
- https://nanopay.atlassian.net/browse/CPF-3177
- We need a FObjectPropertyFilterView so that we can search by toSummary of FObjectProperty

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -498,6 +498,7 @@ FOAM_FILES([
   { name: 'foam/u2/filter/FilterSearch', flags: ['web'] },
   { name: 'foam/u2/filter/FilterViewController', flags: ['web'] },
   { name: 'foam/u2/filter/StringFilterView', flags: ['web'] },
+  { name: 'foam/u2/filter/FObjectPropertyFilterView', flags: ['web'] },
   { name: 'foam/u2/filter/IntegerFilterView', flags: ['web'] },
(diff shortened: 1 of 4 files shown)


jlhughes:
I'm really hesitant to support this; operations will grow to depend on it, but it's not scalable as it only operates on the client side. 
If we do choose to support this perhaps we can move it nanopay side so not to pollute the foam waters with this hack of a search feature.

TW80000:
I agree with @jlhughes. I don't think loading the entire DAO on the client and searching it there is feasible for large DAOs. It will put significant strain on the server to serialize all of that data and the network response would be huge (several GBs for some DAOs).

Another reason I don't think this approach is desirable is because predicates, limit, skip, and order all get applied on the server-side, and in the correct order. Not only that, but searching is efficient on the server due to the way data is stored in memory, _and_ it can take advantage of indexes. Separating this part of predicate out to the client-side doesn't fit in.

KernelDeimos:
Commenting here for later so I don't forget: If this kind of functionality was added to searching (I'm guessing requiring changes to TreeIndex?) it may be useful if other values from properties (i.e. not just `toSummary`) are searchable as well. (with a generic solution rather than multiple cases)

jlhughes:
Closing - will not implement.

PR #2813, opened 2019-11-20
Fix GUI for editing appConfig NSpec

TW80000:


jlhughes (dismissed):

jlhughes:
Testing - still required a refresh before the properties would show.

PR #2814, opened 2019-11-20
Support creating agency session using sessionService

lchanmann:
Added agentId parameter to sessionService and validation.

### DIFF
--- a/src/foam/nanos/session/Session.js
+++ b/src/foam/nanos/session/Session.js
@@ -21,7 +21,8 @@ foam.CLASS({
     'foam.nanos.logger.Logger',
     'foam.nanos.logger.PrefixLogger',
     'foam.util.SafetyUtil',
-    'java.util.Date'
+    'java.util.Date',
+    'static foam.mlang.MLang.*'
   ],
 
   tableColumns: [
@@ -217,6 +218,9 @@ foam.CLASS({
 
         if ( getUserId() == 0 ) return rtn;
 
+        // Validate
+        validate(x);
+
         DAO localUserDAO  = (DAO) x.get("localUserDAO");
         DAO localGroupDAO = (DAO) x.get("localGroupDAO");
         AuthService auth  = (AuthService) x.get("auth");
@@ -242,6 +246,42 @@ foam.CLASS({
         return rtn
           .put("group", group)
           .put("appConfig", group.getAppConfig(rtn));
+      `
+    },
+    {
+      name: 'validate',
+      args: [
+        { name: 'x', type: 'Context' }
+      ],
+      javaCode: `
+        if ( getUserId() < 0 ) {
+          throw new IllegalStateException("User id is invalid.");
+        }
+
+        if ( getAgentId() < 0 ) {
+          throw new IllegalStateException("Agent id is invalid.");
+        }
+
+        if ( getAgentId() > 0 ) {
+          DAO localUserDAO = (DAO) x.get("localUserDAO");
+          User sessionAgent = (User) localUserDAO.inX(x).find(getAgentId());
+
+          if ( sessionAgent == null ) {
+            throw new RuntimeException(String.format("Agent with id '%d' not found.", getAgentId()));
+          }
+
+          DAO agentJunctionDAO = (DAO) x.get("agentJunctionDAO");
+          UserUserJunction junction = (UserUserJunction) agentJunctionDAO.find(
+            AND(
+              EQ(UserUserJunction.SOURCE_ID, getAgentId()),
+              EQ(UserUserJunction.TARGET_ID, getUserId())
+            )
+          );
+
+          if ( junction == null ) {
+            throw new RuntimeException("The junction between user and agent was not found.");
+          }
+        }
       `
     }
   ]
--- a/src/foam/nanos/session/Session.js
+++ b/src/foam/nanos/session/Session.js
@@ -21,7 +21,8 @@ foam.CLASS({
     'foam.nanos.logger.Logger',
     'foam.nanos.logger.PrefixLogger',
     'foam.util.SafetyUtil',
-    'java.util.Date'
+    'java.util.Date',
+    'static foam.mlang.MLang.*'
   ],
 
   tableColumns: [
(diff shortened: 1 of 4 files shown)


jlhughes:
In the GD case, if the session is requested for the 'business' is any of the agency setup required?

lchanmann:
In that case, no agency setup is required.

lchanmann:
This is just an extension on sessionService in general to allow creating agency session.

src/foam/nanos/session/SimpleSessionService.js
```diff
@@ -45,19 +47,45 @@ foam.CLASS({
           throw new IllegalArgumentException("User id must be a positive integer.");
         }
 
-        if ( ! hasSPIDPermission(x, "create", userId) ) {
+        if ( ! hasSPIDPermission(x, "create", userId ) ) {
```
TW80000:
```suggestion
        if ( ! hasSPIDPermission(x, "create", userId) ) {
```

src/foam/nanos/session/SimpleSessionService.js
```diff
@@ -72,11 +100,11 @@ foam.CLASS({
       ],
       type: 'Boolean',
       javaCode: `
-        AuthService auth         = (AuthService) getAuth();
-        DAO         localUserDAO = (DAO) getLocalUserDAO();
+        AuthService auth         = (AuthService) x.get("auth");
```
TW80000: Any reason you changed this? Taking advantage of `imports` seems foamier to me than getting the services we need in code.

src/foam/nanos/session/SimpleSessionService.js
```diff
@@ -45,19 +47,45 @@ foam.CLASS({
...
-        if ( ! hasSPIDPermission(x, "create", userId) ) {
+        if ( ! hasSPIDPermission(x, "create", userId ) ) {
           throw new AuthorizationException("You don't have permission to create a session for that user.");
         }
 
+        if ( agentId < 0 ) {
+          throw new IllegalArgumentException("Agent id is invalid.");
+        }
+
+        if ( agentId > 0 ) {
```
TW80000: I can't decide whether I think this is a good place to do checks or not. Part of me thinks we should move all checks to `Session.applyTo`.

In either case, there are a number of other things we could be checking as well. For example, whether the junction is disabled or not, whether the agent or user have `loginEnabled`, were `deleted`, and potentially other checks as well.

I guess the question is: Should sessions perform all of these checks whenever `applyTo` is called? Or should it be up to the rest of the system to do the appropriate checks in advance? The more I think about it the more I think all of the validation should happen in `applyTo`.

jlhughes: With regard where checks should occur, I would side with applyTo.  One place were we can 'easily' see and manage these important validations.

TW80000: For posterity, @lchanmann and I talked in person and agree that validation checks should be moved to `applyTo`.

src/foam/nanos/session/SimpleSessionService.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2814/commits/3fe6e5951e7cec2fb236a2bf0801a7cc5b119954.

src/foam/nanos/session/SimpleSessionService.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2814/commits/3fe6e5951e7cec2fb236a2bf0801a7cc5b119954.

src/foam/nanos/session/SimpleSessionService.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2814/commits/3fe6e5951e7cec2fb236a2bf0801a7cc5b119954 and https://github.com/foam-framework/foam2/pull/2814/commits/78d32c8ad6ff062017be23c6e7f519588ce73b44.

src/foam/nanos/session/Session.js
```diff
@@ -243,6 +247,59 @@ foam.CLASS({
...
+      name: 'checkUserEnabled',
+      args: [
+        { name: 'x', type: 'Context' },
+        { name: 'userId', type: 'Long' }
+      ],
+      javaCode: `
+        User user = (User) ((DAO) x.get("localUserDAO")).inX(x).find(userId);
+
+        if ( user == null || user.getDeleted() ) {
+          throw new RuntimeException(String.format("User with id '%d' is not found.", userId));
```
TW80000:
```suggestion
          throw new RuntimeException(String.format("User with id '%d' not found.", userId));
```

src/foam/nanos/session/Session.js
```diff
@@ -243,6 +247,59 @@ foam.CLASS({
...
+        }
+
+        if ( ! user.getEnabled() ) {
+          throw new RuntimeException(String.format("User with id '%d' is disabled.", userId));
```
TW80000:
```suggestion
          throw new RuntimeException(String.format("The user with id '%d' has been disabled.", userId));
```

src/foam/nanos/session/Session.js
```diff
@@ -243,6 +247,59 @@ foam.CLASS({
...
+      `
+    },
+    {
+      name: 'checkUserEnabled',
+      args: [
+        { name: 'x', type: 'Context' },
+        { name: 'userId', type: 'Long' }
+      ],
+      javaCode: `
+        User user = (User) ((DAO) x.get("localUserDAO")).inX(x).find(userId);
```
TW80000: I think it makes more sense to use the system context here because we want to make sure we will be allowed to find the user.

```suggestion
        User user = (User) ((DAO) x.get("localUserDAO")).find(userId);
```

src/foam/nanos/session/Session.js
```diff
@@ -243,6 +247,59 @@ foam.CLASS({
...
+        }
+
+        if ( getAgentId() > 0 ) {
+          checkUserEnabled(x, getAgentId());
+
+          DAO agentJunctionDAO = (DAO) x.get("agentJunctionDAO");
+          UserUserJunction junction = (UserUserJunction) agentJunctionDAO.find(
+            AND(
+              EQ(UserUserJunction.SOURCE_ID, getAgentId()),
+              EQ(UserUserJunction.TARGET_ID, getUserId())
```
TW80000, line 276:
```suggestion
              EQ(UserUserJunction.TARGET_ID, getUserId()),
              EQ(UserUserJunction.STATUS, net.nanopay.auth.AgentJunctionStatus.ACTIVE)
```

src/foam/nanos/session/Session.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2814/commits/af11f002f478b55b6dc63691f4b907f00858349f.

src/foam/nanos/session/Session.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2814/commits/af11f002f478b55b6dc63691f4b907f00858349f.

src/foam/nanos/session/Session.js
lchanmann: net.nanopay.auth.AgentJunctionStatus is currently on nanopay side.

lchanmann:
Thanks @TW80000 for the thorough review.

src/foam/nanos/session/Session.js
TW80000: I took note of this and we'll handle this separately. Might move that feature to FOAM or drop it.

TW80000 (approved):

PR #2817, opened 2019-11-21
add error style for RichChoiceView-selection-view, remove selection-v…

yij793:
address:  https://nanopay.atlassian.net/browse/CPF-2255
Notes: I didn't remove the RichChoiceView style in `SectionedDetailPropertyView` since it may affect other view which using RichChoiceView.


TW80000 (approved):
This works, but I'd still prefer if we removed most of the RichChoiceView CSS overrides here.

jlhughes (approved):

PR #2826, opened 2019-11-22
[CPF-3429] Misc. modelDAO-related fixes

KernelDeimos:
This PR makes the following changes

- Property `predicate` of `foam.core.ValidationPredicate` is now transient. This property is generated from validationPredicates, and may include circular references.
- A nully installCSS has been added to WindowNodeJSRefinement
- Java refinements.js reports `null` for the value of a `foam.core.Function`

This is an intermediate step to get DocBrowser working again, and to provide a modelDAO development server for future use.

### DIFF
--- a/src/foam/core/Validation.js
+++ b/src/foam/core/Validation.js
@@ -26,6 +26,7 @@ foam.CLASS({
       class: 'FObjectProperty',
       of: 'foam.mlang.predicate.Predicate',
       name: 'predicate',
+      transient: true,
       expression: function(predicateFactory) {
         return predicateFactory ?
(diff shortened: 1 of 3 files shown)


src/foam/core/Validation.js
```diff
@@ -26,6 +26,7 @@ foam.CLASS({
       class: 'FObjectProperty',
       of: 'foam.mlang.predicate.Predicate',
       name: 'predicate',
+      transient: true,
```
kgrgreer, line 29: Why is this needed? Add a comment to explain.

PR #2827, opened 2019-11-22
[DO_NOT_MERGE] [CPF-3403] Update DocBrowser to use a modelDAO property

KernelDeimos:
Changes DocBrowser to use a modelDAO property. The `expression` for modelDAO is taken from foam.doc.ModelBrowser but it only has a couple hundred entries (really there are over 2000 models). Ideally this will eventually use a modelDAO hosted by a development server containing both server-side and client-side models. 

- Added DO_NOT_MERGE tag because I mistakenly added unrelated files.js changes to this branch.


KernelDeimos:
This may cause a merge conflict with changes @hchoura may make a PR for to make DocBrowser functional again.

This PR will be needed in the future when a `modelDAO` is served (development server in Node.js - I'll work on this after the server for build statistics server [or as a service of] if I have the opportunity). The reason for this would be to make models that aren't available to the client in production (i.e. models for backend logic) available to the model documentation browser in development with the production Javascript build (`foam-bin.js`)

PR #2830, opened 2019-11-22
move ApiTestBase to foam

hchoura:


nanoscott:
@hchoura Move the tests file over as well .. currently they are in nanopay/src/net/nanopay/test/tests.jrl but you should move them to a FOAM tests file.

nanoscott:
Also move the 'Development Portal' menu item order to the end of the list .. so increase the order on the Development Portal menu item to be 1000 instead of 100

PR #2831, opened 2019-11-22
Fix notification bell count and UserInfoNavigation

zahnno:


jlhughes (commented):
Do we have any other references to **agent** on the foam side?

zahnno:
@jlhughes the AgentAuthService logic exists on the foam side, along with the junctions and relationships that pertain to the functionality.

src/foam/nanos/u2/navigation/UserInfoNavigationView.js
```diff
@@ -71,7 +71,7 @@ foam.CLASS({
             .add(this.slot((agent, user) => {
               return this.E().addClass('name-container')
                   .start('span').addClass(this.myClass('userName'))
-                    .add( agent ? user.businessName() : user.label())
+                    .add( agent.firstName ? user.businessName : user.label())
```
jlhughes: User.businessName is not required so perhaps the test and output should be: 
```suggestion
                    .add( agent.businessName ? agent.businessName : user.label())
```

zahnno: I don't think agent.businessName makes much sense to check here since agents could be apart of multiple users also the purpose of this condition is if agent exists/has a displayable value(firstName), we use user(Business) business name, instead of first and last name. We account for both states. One displaying a business name the other a users first name and last name.

Then the slot on ln 77 accounts for the agent and displays the appropriate label depending on it state.

Here's the design I'm referencing https://app.zeplin.io/project/5cf963f36af3e11bb4bedf5f/screen/5cf9674986ef376f6070fe8e

TW80000: Why aren't we just using `label()`?
```suggestion
                    .add(user.label())
```

If it's the portal, then `agent` isn't there and `user` is a `User`, so `user.label()` will give us the legal name (human name), which is what we want.

If it's Ablii, then `agent` _is_ there and `user` is a `Business`, so `user.label()` will give us the business name, which is what we want.

jlhughes: Deferring to Troy below.

src/foam/nanos/u2/navigation/UserInfoNavigationView.js
```diff
@@ -71,7 +71,7 @@ foam.CLASS({
             .add(this.slot((agent, user) => {
```
zahnno:
```suggestion
             .add(this.slot((user) => {
```

TW80000 (approved):

PR #2832, opened 2019-11-24
Extract outputException to DigUtil

lchanmann:
Extract outputException in DigWebAgent and SugarWebAgent to DigUtil.

### DIFF
--- a/src/foam/nanos/dig/DigUtil.java
+++ b/src/foam/nanos/dig/DigUtil.java
@@ -6,6 +6,7 @@
 
 package foam.nanos.dig;
 
+import foam.core.FObject;
 import foam.core.X;
 import foam.lib.*;
 import foam.lib.csv.CSVOutputter;
@@ -30,10 +31,16 @@ public class DigUtil {
    * @param format The format of the output.
    */
   public static void outputException(X x, DigErrorMessage error, Format format) {
+    HttpServletResponse resp = x.get(HttpServletResponse.class);
+
+    resp.setStatus(Integer.parseInt(error.getStatus()));
+    outputFObject(x, error, format);
+  }
+
+  public static void outputFObject(X x, FObject object, Format format) {
     HttpServletResponse resp  = x.get(HttpServletResponse.class);
     PrintWriter         out   = x.get(PrintWriter.class);
 
-    resp.setStatus(Integer.parseInt(error.getStatus()));
     if ( format == Format.JSON ) {
       //output error in json format
       resp.setContentType("application/json");
@@ -43,7 +50,7 @@ public class DigUtil {
       Outputter outputterJson = new Outputter(x).setPropertyPredicate(new AndPropertyPredicate(x, new PropertyPredicate[] {new NetworkPropertyPredicate(), new PermissionedPropertyPredicate()}));
       outputterJson.setOutputDefaultValues(true);
       outputterJson.setOutputClassNames(true);
-      outputterJson.output(error);
+      outputterJson.output(object);
       out.println(outputterJson.toString());
 
     } else if ( format == Format.XML )  {
@@ -51,7 +58,7 @@ public class DigUtil {
       resp.setContentType("application/xml");
 
       foam.lib.xml.Outputter outputterXml = new foam.lib.xml.Outputter(OutputterMode.NETWORK);
-      outputterXml.output(error);
+      outputterXml.output(object);
       out.println(outputterXml.toString());
 
     } else if ( format == Format.CSV )  {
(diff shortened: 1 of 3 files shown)


src/foam/nanos/dig/DigUtil.java
```diff
@@ -0,0 +1,77 @@
...
+      JSONParser jsonParser = new JSONParser();
+      jsonParser.setX(x);
+      Outputter outputterJson = new Outputter(x).setPropertyPredicate(new AndPropertyPredicate(new PropertyPredicate[] {new StoragePropertyPredicate(), new PermissionedPropertyPredicate()}));
+      outputterJson.setOutputDefaultValues(true);
+      outputterJson.setOutputClassNames(true);
+      outputterJson.outputJSONJFObject(error);
+      out.println(outputterJson.toString());
+
+    } else {
+      // TODO
```
jlhughes: Can you generate a warning or log message please.  Or throw a java.lang.UnsupportedOperationException.

jlhughes (approved):

jlhughes (approved):

src/foam/nanos/dig/DigUtil.java
```diff
@@ -0,0 +1,78 @@
+package foam.nanos.dig;
```
jlhughes, line 7: License header please.

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2832/commits/d03029976dee996b104fa23834a95c877202baed.

src/foam/nanos/dig/DigUtil.java
```diff
@@ -0,0 +1,84 @@
...
+import foam.lib.json.JSONParser;
+import foam.lib.json.Outputter;
+import foam.lib.json.OutputterMode;
+import foam.nanos.dig.exception.DigErrorMessage;
+import foam.nanos.http.Format;
+
+import javax.servlet.http.HttpServletResponse;
+import java.io.PrintWriter;
+
+public class DigUtil {
```
kgrgreer, line 25: Document what this class/file is for.

src/foam/nanos/dig/DigUtil.java
```diff
@@ -0,0 +1,84 @@
...
+  public static void outputException(X x, HttpServletResponse resp, Format format, PrintWriter out, DigErrorMessage error) {
+    if ( resp == null ) resp = x.get(HttpServletResponse.class);
```
kgrgreer: Why would resp be null? If it is, isn't that because of a bug somewhere else that we should fix?

src/foam/nanos/dig/DigUtil.java
```diff
@@ -0,0 +1,84 @@
...
+    if ( out == null ) out = x.get(PrintWriter.class);
```
kgrgreer: Why would out be null?

src/foam/nanos/dig/DigUtil.java
kgrgreer, line 25: If the only thing this class does is output exceptions, then wouldn't a more specific name be better? Something like DigExceptionOutputter?

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2832/commits/340b4db558bd19d16930d65f2f1707e5ff057123.

src/foam/nanos/dig/DigUtil.java
lchanmann: The null resp is for falling back to the response in the context so that it doesn't need to be passed in explicitly.

src/foam/nanos/dig/DigUtil.java
lchanmann: The null out is for falling back to the PrintWriter in the context so that it doesn't need to be passed in explicitly.

src/foam/nanos/dig/DigUtil.java
lchanmann: We plan to use it as a container for Dig related static utility methods.

src/foam/nanos/dig/DigUtil.java
kgrgreer: But why wouldn't you pass it in explicitly? Either pass in the context or as an argument, but doing both doesn't make sense.

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2832/commits/54f8f146104f33c4fa1457457b6160af32926819.

src/foam/nanos/dig/DigUtil.java
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/2832/commits/54f8f146104f33c4fa1457457b6160af32926819.

PR #2833, opened 2019-11-25
Assembly models

nanoMichal:


src/foam/util/concurrent/Assembly.js
```diff
@@ -0,0 +1,55 @@
...
+      documentation: `
+        Mark this Assembly as having completed.
+        Will notify to unblock thread blocked by waitToComplete().
+      `
+    },
+    {
+      name: 'waitToComplete',
+      documentation: `
+        Block until the Assembly becomes complete.
+        Calling lets Assembly know it isn't last in queue.
```
KernelDeimos, line 51: Does waitToComplete get called by AssemblyLine?

KernelDeimos (approved):

src/foam/util/concurrent/AbstractAssembly.js
```diff
@@ -0,0 +1,87 @@
...
+   * Provides concurrency support.
+   **/
+
+  implements: [
+    'foam.util.concurrent.Assembly'
+  ],
+
+  properties: [
+    {
+      name: 'isCompleted',
```
kgrgreer, line 24: Isn't synchronized will cause a bug.

nanoMichal: I guess I can't add "synchronized: true" to the property but I can make the setter/getter synchronized by overwriting it, and then the "complete" method wouldn't need to be synchronized. Or I can add synchronized to property?

PR #2834, opened 2019-11-25
Create a new object when parsing an FObject

hchoura:
When we parse an object, the context may return null. (in case that the
object is send from the client)

ProxyParser jsonP =  new MapParser();
Map mapPostParam = (Map) jsonP.value();
jsonP.setX(x);

Even setting the context don't fix the issue.


hchoura:
@googlebot I fixed it.

hchoura:
@googlebot I fixed it.

hchoura:
@googlebot I fixed it.

PR #2836, opened 2019-11-26
decorator injection searches for last ProxyDAO, it failed if the last…

jlhughes:
… ProxyDAO did have a non ProxyDAO delegate.

### DIFF
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -155,8 +155,10 @@ foam.CLASS({
           }
           // The decorator dao may be a proxy chain
           ProxyDAO proxy = (ProxyDAO) getDecorator();
-          while ( proxy.getDelegate() != null ) 
+          while ( proxy.getDelegate() != null &&
+                  proxy.getDelegate() instanceof ProxyDAO ) {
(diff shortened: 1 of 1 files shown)


src/foam/dao/EasyDAO.js
```diff
@@ -155,8 +155,10 @@ foam.CLASS({
           }
           // The decorator dao may be a proxy chain
           ProxyDAO proxy = (ProxyDAO) getDecorator();
-          while ( proxy.getDelegate() != null ) 
+          while ( proxy.getDelegate() != null &&
+                  proxy.getDelegate() instanceof ProxyDAO ) {
```
michaelconnor468, line 159: This is a good catch; I feel as if it's going to be a source of a few developer errors in the future. What do you think about adding a warning message to this as well to alert whoever is using it incorrectly? I thought of making it something along the lines of,

if ( proxy.getDelegate() instanceof ProxyDAO ) {
logger.error(this.getClass().getSimpleName(), "NSpec.name", (getNSpec() != null ) ? getNSpec().getName() : null,  "Decorator chain used in setDecorator must be null ending");
break;
}

and putting it within the original while condition

PR #2838, opened 2019-11-26
Fix bug when changing class in create controller

TW80000:
When creating a new object, the user might change between different subclasses.
The current behaviour of FObjectView is to copyFrom the old object to the
new one when the class changes, but this causes an often unwanted side
effect where default values of the first class chosen are copied to the
new class chosen instead of using the new class' default values.

This commit gets rid of that behaviour and simply creates a new instance
of the new subclass when chosen.

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -39,9 +39,7 @@ foam.CLASS({
         if ( newValue !== oldValue ) {
           var m = this.__context__.lookup(newValue, true);
           if ( m ) {
-            var n = m.create(null, this);
-            n.copyFrom(this.data);
-            this.data = n;
+            this.data = m.create(null, this);
           }
         }
       }


jlhughes (approved):

mcarcaso:
I’m on phone so I can’t see too much but this change will lose whatever
data was which is probably fine for create controller but for an update it
could clobber/clear the data. Is that desirable?

TW80000:
> I’m on phone so I can’t see too much but this change will lose whatever data was which is probably fine for create controller but for an update it could clobber/clear the data. Is that desirable?

@mcarcaso You would be right, but we're not using `FObjectView` in `DAOUpdateControllerView`, just in `DAOCreateControllerView`. So you can't change the subclass after creation anyway (at least via the GUI (using those views)). So we don't come across that situation anyway.

That being said, I guess your question still applies to when `FObjectView` is used as the view of a property. In that case, I still think its desirable to drop changes instead of copying them over. It's kind of a toss-up whether the properties that are common between the different classes are ones you want to copy over or not. In the case of something like 'name', you would want to copy because that's not going to change. Anything with a default value (like CAD bank account has certain defaults that are different than the defaults of a USD bank account) is less likely to make sense to copy.

So the question is whether it's worse to have to re-input some data or lose the "correct" defaults. I think losing the correct defaults is worse because it's not clear to the user which properties have meaningful default values, and therefore they have no way of knowing what they need to change.

When I'm creating an account and I choose CAD bank account, a bunch of properties default to CAD-related values. When I switch the class to USD bank account, the CAD-related properties are copied over to the USD bank account object instead of using the USD bank account defaults. That seems worse to me than getting the right defaults and having to fill in a few fields again.

mcarcaso:
Nice description of the issues :)

I'm mostly indifferent on the behaviour. There are pros/cons to each
approach that you laid out real nice and the desired behaviour is pretty
context dependent. A flag on the view to let the developer choose the
behaviour could be useful? Another option would be to prompt the user and
have them decide if they want changes to be copied over when the class
changes?

Anyway, I'm ok with the changes you made and will leave it up to you if you
want to pursue something that can handle both scenarios but if there's no
need for it now, no sense doing it right away.

PR #2840, opened 2019-11-27
add mode to intview

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3507
note: the intview class visibility RO isn't working. please check the change I made.
@TW80000

### DIFF
--- a/src/foam/u2/IntView.js
+++ b/src/foam/u2/IntView.js
@@ -44,6 +44,7 @@ foam.CLASS({
       this.SUPER(p);
       this.min = p.min;
       this.max = p.max;
+      this.mode = p.mode;
     }
   ]
 });


src/foam/u2/IntView.js
```diff
@@ -44,6 +44,7 @@ foam.CLASS({
       this.SUPER(p);
       this.min = p.min;
       this.max = p.max;
+      this.mode = p.mode;
```
TW80000, line 47: This doesn't make sense to me. `p` is a `Property` and afaik, `Property`-ies don't have a `mode` property.

yij793: > This doesn't make sense to me. `p` is a `Property` and afaik, `Property`-ies don't have a `mode` property.

We have `mode` property at `View` class in Element.js line: 2699
`    class: 'Enum',
      of: 'foam.u2.DisplayMode',
      name: 'mode',`
and the intView is extend from View

TW80000: Right, so `this.mode` makes sense, but it's `p.mode` that doesn't make sense to me.

KernelDeimos (commented):
There is another IntView at `foam.u2.view.IntView` rather than `foam.u2.IntView`. The one in the `view` package extends `foam.u2.view.ModeAltView` which has this value for documentation:
> A view that shows one of several views, depending on the display mode.

Does foam.u2.view.IntView have the same behaviour as this change, or is this something different?

PR #2842, opened 2019-11-27
Add sugar data and update the Post request.

hchoura:


nanoscott (approved):

src/foam/nanos/dig/SUGAR.js
```diff
@@ -208,7 +208,7 @@ foam.CLASS({
       // TODO: appears not to work if named 'url', find out why.
       name: 'sugarURL',
       label: 'URL',
-      hidden: false,
+      hidden: true,
```
kgrgreer, line 211: Why are we hiding?

PR #2845, opened 2019-11-28
stop hidden HTML generation

nanoNeel:


kgrgreer:
Brilliant!

src/foam/layout/SectionAxiom.js
```diff
@@ -52,17 +52,20 @@ foam.CLASS({
...
 
       if ( this.permissionRequired ) {
-        var permSlot = data$.map(data => {
-          if ( ! data || ! data.__subContext__.auth ) return false;
-          return data.__subContext__.auth.check(null,
-            `${data.cls_.id.toLowerCase()}.section.${this.name}`);
-        });
+        var permSlot = foam.core.SimpleSlot.create({value: false});
+        var data = data$.get();
+        if ( data && data.__subContext__.auth ) {
```
TW80000, line 57: This block won't update when data changes now though. If `data` is null at first and gets set later, the permission check won't be performed.

kgrgreer:
The old code didn't work because you can't return a promise from
slot.map(). When you're available Neel, I'll show you how to easily fix.

PR #2848, opened 2019-11-28
Cpf 278 ORBitalDAO feature

skanderturki:
Previous deiscussions and corrections requests may be found here:
https://github.com/foam-framework/foam2/pull/2547


src/foam/dao/ORBitalDAO.js
```diff
@@ -0,0 +1,58 @@
...
+      name: 'cmd_',
+      type: 'Object',
+
+      args: [
+        { name: 'x' , type: 'foam.core.X'}, { name: 'obj' , type: 'Object' }
+      ],
+
+      javaCode: `
+        ORBRequest request = null;
+        if ( obj instanceof ORBRequest ) request = (ORBRequest) obj;
```
kgrgreer: Please don't format if/else this way in general, but in this case, this would be easier to read:

 if ( ! ( obj instanceof ORBRequest )  ) return getDelegate().cmd_(x, obj);

request = (ORBRequest) obj;

src/foam/dao/ORBitalDAO.js
```diff
@@ -0,0 +1,58 @@
...
+
+        // Get the receiving object
+        FObject receiverFObj = request.getReceiverObject();
+
+        receiverFObj =  ( receiverFObj == null && receiverID != null )? (FObject) x.get(receiverID) : receiverFObj;
+        if ( receiverFObj != null ) {
+          ClassInfo classInfo = receiverFObj.getClassInfo();
+          MethodInfo methodInfo = (MethodInfo) classInfo.getAxiomByName(request.getMethodName());
+          return methodInfo.call(x, receiverFObj, request.getArgs());
+        } else {
```
kgrgreer: No need for an 'else' if you've already returned in the if. Also, code would read better if you negated the if condition. ie.

if ( receiverFObj == null )  return null;

ClassInfo classInfo = receiverFObj.getClassInfo();
MethodInfo methodInfo = (MethodInfo) classInfo.getAxiomByName(request.getMethodName());
return methodInfo.call(x, receiverFObj, request.getArgs());

src/foam/java/refinements.js
```diff
@@ -675,6 +675,56 @@ foam.CLASS({
   ],
 
   methods: [
+    function buildMethodInfoInitializer(cls) {
+      // Add MethodInfo field for each method
+      initializerString = `new foam.core.MethodInfo(){
+        @Override
+        public String getName(){
+          return \"${this.name}\";
```
kgrgreer: Are the \'s required?

src/foam/java/refinements.js
```diff
@@ -675,6 +675,56 @@ foam.CLASS({
...
+        else if ( this.args[i].type )
+          argsString += '(' + this.args[i].type.replace('...', '[]').replace('final ', '') + `)(args[${ i }])`;
+        else if ( this.args[i].class )
+          argsString += '(' + this.args[i].class.replace('...', '[]').replace('final ', '') + `)(args[${ i }])`;
+        else
+          continue;
+        if ( i != this.args.length - 1 ) argsString += ', ';
+      }
+      initializerString += argsString + ');\n';
+      if( this.javaType == 'void' ) initializerString += '            return null;\n';
```
kgrgreer: space after "if"

src/foam/java/refinements.js
```diff
@@ -675,6 +675,56 @@ foam.CLASS({
...
+
+      // Close try block
+      if ( exceptions ) { initializerString += `          }
+         catch (java.lang.Exception e) {
```
kgrgreer: "java.lang." not required.

src/foam/java/refinements.js
kgrgreer: Are the backslashes required?

src/foam/dao/ORBitalDAO.js
skanderturki: ok

src/foam/dao/ORBitalDAO.js
skanderturki: ok

src/foam/java/refinements.js
skanderturki: it's a string so it should be surrounded by double quotes. And the backslashes are there to escape the "

src/foam/java/refinements.js
skanderturki: ok

src/foam/java/refinements.js
skanderturki: ok

src/foam/java/refinements.js
kgrgreer: But it isn't inside a " quoted string, it's inside a ` quoted string, so the " has no special meaning and doesn't need to be escaped. As far as I know.

skanderturki: ok

PR #2849, opened 2019-11-29
Notification settings fixes

zahnno:
- Removed unnecessary imports
- Removed references to Business Class and business vocabulary.

### DIFF
--- a/src/foam/nanos/auth/Relationships.js
+++ b/src/foam/nanos/auth/Relationships.js
@@ -48,8 +48,8 @@ foam.RELATIONSHIP({
   cardinality: '1:*',
   sourceModel: 'foam.nanos.auth.UserUserJunction',
   targetModel: 'foam.nanos.notification.NotificationSetting',
-  forwardName: 'notificationSettingsForBusinessUsers',
-  inverseName: 'businessUser',
+  forwardName: 'notificationSettingsForUserUsers',
+  inverseName: 'userJunction',
   sourceProperty: {
     hidden: true
   }
--- a/src/foam/nanos/notification/NotificationSetting.js
+++ b/src/foam/nanos/notification/NotificationSetting.js
@@ -62,6 +62,7 @@ foam.CLASS({
         notification.setUserId(user.getId());
         notification.setBroadcasted(false);
         notification.setGroupId(null);
+        notification.setEmailIsEnabled(false);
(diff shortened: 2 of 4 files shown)


src/services
```diff
@@ -709,6 +709,8 @@ p({
 p({
   "class": "foam.nanos.boot.NSpec",
   "name": "notificationSettingDAO",
+  "lazy": false,
```
kgrgreer: Why is it not lazy?

zahnno: Reverted, there was no reason.

src/foam/nanos/auth/Relationships.js
```diff
@@ -48,8 +48,8 @@ foam.RELATIONSHIP({
   cardinality: '1:*',
   sourceModel: 'foam.nanos.auth.UserUserJunction',
   targetModel: 'foam.nanos.notification.NotificationSetting',
-  forwardName: 'notificationSettingsForBusinessUsers',
-  inverseName: 'businessUser',
+  forwardName: 'notificationSettingsForUserUsers',
```
KernelDeimos, line 51: Does this break calls to `businessUserJunction.getNotificationSettingsForBusinessUsers(x)`?

src/foam/nanos/notification/NotificationSetting.js
```diff
@@ -70,6 +62,7 @@ foam.CLASS({
         notification.setUserId(user.getId());
         notification.setBroadcasted(false);
         notification.setGroupId(null);
+        notification.setEmailIsEnabled(false);
```
KernelDeimos, line 65: How does this become true when sending an email is desired?

src/foam/nanos/auth/Relationships.js
zahnno: Yes it will.

zahnno: I renamed this because references to businesses shouldn't exist in foam unless it's explicitly required.

src/foam/nanos/notification/NotificationSetting.js
zahnno: Another settings is responsible for sending emails - foam.nanos.notification.EmailSetting.

jlhughes (approved):

PR #2856, opened 2019-12-03
Phone number validator

zahnno:
Fix phone number validation predicate to dismiss predicates if phone is not required.

### DIFF
--- a/src/foam/core/Validation.js
+++ b/src/foam/core/Validation.js
@@ -363,16 +363,9 @@ foam.CLASS({
                   self,
                   /^(?:\+?1[-.●]?)?\(?([0-9]{3})\)?[-.●]?([0-9]{3})[-.●]?([0-9]{4})$/
                  );
-              },
-              errorString: 'Please enter a valid phone number'
             },
-            {
-              args: [this.name],
-              predicateFactory: function(e) {
-                return e.NEQ(self, '');
-              },
-              errorString: 'Please enter a phone number'
-            }
+            errorString: 'Please enter a valid phone number'
+          }
         ];
       }
     }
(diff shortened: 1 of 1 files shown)


TW80000 (approved):
This works, but we can probably just get rid of the second predicate altogether since the first one will always be invalid when the second one is invalid anyway.

zahnno:
@TW80000 yah I agree with that

src/foam/core/Validation.js
```diff
@@ -354,30 +354,19 @@ foam.CLASS({
       name: 'validationPredicates',
       factory: function() {
         var self = this;
-        var ret = [
+        if ( ! this.required ) return;
```
TW80000:
```suggestion
        if ( ! this.required ) return [];
```

I think we need to return an empty array instead of undefined in this case. The code that uses the validationPredicates expects it to be an array. That's what the other validationPredicates definitions in this file do as well. Sorry I didn't catch this the first time.

TW80000 (changes requested):

src/foam/core/Validation.js
```diff
@@ -354,30 +354,19 @@ foam.CLASS({
...
             predicateFactory: function(e) {
                 return e.REG_EXP(
                   self,
                   /^(?:\+?1[-.●]?)?\(?([0-9]{3})\)?[-.●]?([0-9]{3})[-.●]?([0-9]{4})$/
                  );
-              },
-              errorString: 'Please enter a valid phone number'
-            }
+            },
+            errorString: 'Please enter a valid phone number'
```
kgrgreer, line 367: Better to use a Message so we can internationalize in the future. Not going to block PR, just saying.

PR #2859, opened 2019-12-04
Omits updating objectClass when initializating FObjectView

zahnno:
Data being set on FObjectView causes the postSet condition which checks for updates on objectClass to always pass due to the initial value always consisting an empty string. Added an additional condition of an empty oldValue to dismiss this initialization issue.

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -36,7 +36,7 @@ foam.CLASS({
         };
       },
       postSet: function(oldValue, newValue) {
-        if ( newValue !== oldValue ) {
+        if ( newValue !== oldValue && oldValue !== '' ) {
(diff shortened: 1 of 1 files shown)


TW80000 (approved):
The only thing I can think of that might break because of this is if you did something like this:

```JavaScript
.tag({
  class: 'foam.u2.view.FObjectView',
  objectClass: 'foam. nanos.auth.User'
})
```

By setting `objectClass` but not data, the `postSet` on `objectClass` would be triggered, but this condition would prevent it from setting `data`. I didn't test this though. I did look through everywhere we use FObjectView though and we don't do the above anywhere. We usually set `of`, not `objectClass`, and sometimes bind `data` to something.

Approving since it fixes the bug and shouldn't break any existing code, but I'll try to think of a better way we could be doing this.

PR #2860, opened 2019-12-05
[DO_NOT_MERGE] Implement ServiceProviderAware filtering on find and select

jlhughes:
Provide filtering by spid on find and select. 
Until ROPE is fully integrated. 

NOTE: uses reflection to craft the Reference findFoo(x) calls.  

Running test cases:
`$ rm -rf /tmp/nanopay; ./build.sh -TUserAndGroupPermissionTest,ServiceProviderAwareTest`

Outstanding
* consideration of remove and removeAll.


src/foam/nanos/script/TestRunnerScript.js
```diff
@@ -12,6 +12,7 @@ foam.CLASS({
   javaImports: [
     'foam.dao.DAO',
     'foam.dao.ArraySink',
+    'foam.nanos.app.AppConfig',
```
lchanmann: AppConfig is unused.

src/foam/nanos/auth/test/ServiceProviderAwareTest.js
```diff
@@ -0,0 +1,250 @@
...
+        List users = sink.getArray();
+        test (users.size() == 3, "["+(idx++)+"]  ServiceProvicerAwareDAO select matched on spid. expected: 3, found: "+users.size());
+
+        ctxUser = new User.Builder(x)
+          .setId(99996)
+          .setEmail("test@example.com")
+          .setSpid("fail")
+          .setGroup("admin")
+          .build();
+        ctxUser = (User) delegate.put_(x, ctxUser);
```
lchanmann: Seems like the context user (admin) who belongs to "test" spid can add new user for other spid "fail". Is it intended or I misunderstand the flow here?

src/foam/nanos/auth/ServiceProviderAwareSupport.js
```diff
@@ -0,0 +1,169 @@
...
+          name: 'obj',
+          type: 'Object'
+        }
+      ],
+      type: 'java.lang.reflect.Method',
+      javaThrows: ['Exception'],
+      javaCode: `
+    java.lang.reflect.Method method = (java.lang.reflect.Method) getFindMethods().get(name);
+    if ( method == null ) {
+      String methodName = "find" + name.substring(0,1).toUpperCase() + name.substring(1);
```
lchanmann, line 184: we would also need to handle capitalization for multi-word property name such as userId - https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/notification/Notification.js#L79.

src/foam/nanos/auth/test/ServiceProviderAwareTest.js
jlhughes: The delegate is an explicit MDAO - not going through the normal DAO stack.  Just for test setup.

src/foam/nanos/auth/ServiceProviderAwareSupport.js
jlhughes: Yep, the real solution is to refine PropertyInfo with the Reference methods, but have no idea how to yet.

jlhughes: multi-word should work.  If the name is userId,  find becomes findUserId, so just the first letter needs to be uppercased.

lchanmann: that's right, it should work fine.

src/foam/dao/EasyDAO.js
```diff
@@ -542,9 +546,16 @@ foam.CLASS({
...
+      // note: want this decorator on the non-local/served stack. 
       name: 'serviceProviderAware',
       class: 'Boolean',
-      javaFactory: 'return getEnableInterfaceDecorators() && foam.nanos.auth.ServiceProviderAware.class.isAssignableFrom(getOf().getObjClass());'
+      javaFactory: `return getNSpec().getServe() &&
+                            ! getEnableInterfaceDecorators() &&
+                            foam.nanos.auth.ServiceProviderAware.class.isAssignableFrom(getOf().getObjClass());`
+    },
+    {
+      name: 'serviceProviderAwarePropertyInfos',
```
TW80000, line 586: What is this? A `documentation` comment would be useful.

src/foam/nanos/auth/test/DummySp.js
```diff
@@ -0,0 +1,18 @@
+foam.CLASS({
```
TW80000, line 7: Missing license header

src/foam/util/Auth.java
```diff
@@ -33,6 +33,7 @@ public static X sudo(X x, User user) {
     Session session = new Session();
     x = x.put(Session.class, session);
     x = x.put("user", user);
+    x = x.put("spid", user.getSpid());
```
TW80000: I wonder if we should be using `applyTo` here. Then we could put the SPID in the context during `applyTo`.

src/foam/nanos/auth/User.js
```diff
@@ -525,17 +525,14 @@ foam.CLASS({
...
         boolean findSelf = SafetyUtil.equals(this.getId(), user.getId()) ||
           (
             agent != null &&
             SafetyUtil.equals(agent.getId(), this.getId())
           );
 
-        if (
-          ! findSelf &&
-          ! auth.check(x, "user.read." + this.getId()) &&
-          ! auth.check(x, "spid.read." + this.getSpid())
```
TW80000, line 548: We do similar checks in `authorizeOnUpdate` and `authorizeOnDelete`. Should we remove those as well?

src/foam/dao/EasyDAO.js
```diff
@@ -542,9 +546,16 @@ foam.CLASS({
       value: true
     },
     {
+      // note: want this decorator on the non-local/served stack. 
       name: 'serviceProviderAware',
       class: 'Boolean',
-      javaFactory: 'return getEnableInterfaceDecorators() && foam.nanos.auth.ServiceProviderAware.class.isAssignableFrom(getOf().getObjClass());'
+      javaFactory: `return getNSpec().getServe() &&
+                            ! getEnableInterfaceDecorators() &&
```
TW80000: Why was the `!` added? Wouldn't you consider `ServiceProviderAware` an interface decorator?

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -39,7 +56,7 @@ foam.CLASS({
     if ( isCreate ) {
       if ( SafetyUtil.isEmpty(sp.getSpid()) ||
            ( ! SafetyUtil.isEmpty(sp.getSpid()) &&
-             ! ((AuthService) x.get("auth")).check(x, "spid.create.*") ) ) {
+             ! ((AuthService) x.get("auth")).check(x, "spid.create."+sp.getSpid()) ) ) {
```
TW80000: Should create a variable for `((AuthService) x.get("auth"))` since we use it several times.

TW80000:
```suggestion
             ! ((AuthService) x.get("auth")).check(x, "spid.create." + sp.getSpid()) ) ) {
```

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -49,13 +66,48 @@ foam.CLASS({
         }
       }
     } else if ( ! sp.getSpid().equals(oldSp.getSpid()) &&
-                ! ((AuthService) x.get("auth")).check(x, "spid.update.*") ) {
+                ! (((AuthService) x.get("auth")).check(x, "spid.update."+oldSp.getSpid()) &&
```
TW80000:
```suggestion
                ! (((AuthService) x.get("auth")).check(x, "spid.update." + oldSp.getSpid()) &&
```

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -49,13 +66,48 @@ foam.CLASS({
...
+                   ((AuthService) x.get("auth")).check(x, "spid.update."+sp.getSpid())) ) {
```
TW80000:
```suggestion
                   ((AuthService) x.get("auth")).check(x, "spid.update." + sp.getSpid())) ) {
```

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -9,15 +9,32 @@ foam.CLASS({
   name: 'ServiceProviderAwareDAO',
   extends: 'foam.dao.ProxyDAO',
 
-  documentation: 'Enforce Spid to that of context user or AppConfig default.',
+  documentation: `Enforce Spid to that of context user or AppConfig default on put,
```
TW80000:
```suggestion
  documentation: `Set the SPID of the object being created to that of the context user or the AppConfig default,
```

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -49,13 +66,48 @@ foam.CLASS({
...
+    {
+      name: 'find_',
+      javaCode: `
+    FObject result = getDelegate().find_(x, id);
+    if ( result == null ||
+         ((AuthService) x.get("auth")).check(x, "*") ) {
+      return result;
+    }
+
+    if ( new ServiceProviderAwareSupport().match(x, getPropertyInfos(), result) ) {
```
TW80000, line 94: Do we need to create a new instance of `ServiceProviderAwareSupport` every time? Couldn't we make it a singleton or static class instead?

src/foam/nanos/auth/ServiceProviderAwareSupport.js
```diff
@@ -0,0 +1,178 @@
...
+      if ( user == null ) {
+        // TODO/REVIEW: occurs during login. See AuthenticationApiTest
+        return true;
+      }
+
+      if ( obj != null &&
+           obj instanceof ServiceProviderAware ) {
+        ServiceProviderAware sp = (ServiceProviderAware) obj;
+        return user.getSpid().equals(sp.getSpid()) ||
+          ((AuthService) x.get("auth")).check(x, "spid.read."+sp.getSpid());
```
TW80000:
```suggestion
          ((AuthService) x.get("auth")).check(x, "spid.read." + sp.getSpid());
```

src/foam/nanos/auth/ServiceProviderAwareSupport.js
```diff
@@ -0,0 +1,178 @@
...
+      }
+
+      Object result = obj;
+      while ( result != null ) {
+        PropertyInfo pInfos[] = (PropertyInfo[]) properties.get(result.getClass().getName());
+        if ( pInfos == null ) {
+          return false;
+        }
+        for ( int i = 0; i < pInfos.length; i++) {
```
TW80000:
```suggestion
        for ( int i = 0; i < pInfos.length; i++ ) {
```

src/foam/nanos/auth/ServiceProviderAwareSupport.js
```diff
@@ -0,0 +1,178 @@
...
+          foam.core.PropertyInfo pInfo = pInfos[i];
+          try {
+            java.lang.reflect.Method method = getFindMethod(x, pInfo.getName(), result);
+            result = invokeMethod(x, method, result);
+            if ( result != null &&
+                 result instanceof ServiceProviderAware ) {
+              ServiceProviderAware sp = (ServiceProviderAware) result;
+              return user.getSpid().equals(sp.getSpid()) ||
+                ((AuthService) x.get("auth")).check(x, "spid.read."+sp.getSpid());
```
TW80000:
```suggestion
                ((AuthService) x.get("auth")).check(x, "spid.read." + sp.getSpid());
```

src/foam/nanos/auth/ServiceProviderAwareSupport.js
```diff
@@ -0,0 +1,178 @@
...
+        }
+        for ( int i = 0; i < pInfos.length; i++) {
+          foam.core.PropertyInfo pInfo = pInfos[i];
+          try {
+            java.lang.reflect.Method method = getFindMethod(x, pInfo.getName(), result);
+            result = invokeMethod(x, method, result);
+            if ( result != null &&
+                 result instanceof ServiceProviderAware ) {
+              ServiceProviderAware sp = (ServiceProviderAware) result;
+              return user.getSpid().equals(sp.getSpid()) ||
```
TW80000: Isn't this always going to return in the first iteration of the loop? It will if `result` isn't null anyway. Don't we want to loop over all properties and return if at least one of them is the user we're looking for? Right now this will return on the first user it finds whether it's the right user or not I think.

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -49,13 +66,48 @@ foam.CLASS({
...
+    {
+      name: 'select_',
+      javaCode: `
+    if ( ((AuthService) x.get("auth")).check(x, "*") ) {
+      return super.select_(x, sink, skip, limit, order, predicate);
+    }
+
+    ProxySink proxy = (ProxySink) getDelegate().select_(
+      x,
+      new ServiceProviderAwareSink(x, sink, getPropertyInfos()),
```
TW80000: Hmm, I don't think this plays nicely with the other sink decorators. When `SkipSink`, `LimitSink`, `PredicatedSink`, and `OrderedSink` are added, they'll be _outside_ of this sink if I'm not mistaken. This is an issue because if we apply skip and limit before this sink gets hit, then we'll only be filtering out the limited number. What we really want is for the SPID filtering to occur before or as part of the predicate stage of the sink decorator chain, which happens before skip, limit, and order.

TW80000 (commented):
Haven't reviewed everything yet. Just submitting what I have so far.

src/foam/nanos/auth/User.js
jlhughes: The authorizeOnUpdate are correct, or at least work with ServiceProviderAwareDAO.put which tests for 'update' or 'create' permissions,  there are some test cases for this. 

Remove/Delete I haven't yet considered.

src/foam/dao/EasyDAO.js
jlhughes: Presently we have local and non-local daos. In most cases the **non-local** is served and interface decorators such as created, createdBy are on the **local** stack, and we (should) disable them on the **non-local** stack - as not needed in both places.  
This particular decorator I 'think' should be on the **non-local** stack as we don't it need on the **local** stacks where most access is by the system user which this decorator short circuits. 
I supposed just testing 'served' would be sufficient.  And of course this is just the 'default', as we can explicitly set the property to true or false.

src/foam/util/Auth.java
jlhughes: added to Session.applyTo

src/foam/nanos/auth/ServiceProviderAwareDAO.js
jlhughes: The propertyInfos are dynamic, but for a given application this filtering will only be applied to a handful of DAOs so this support class could be static/singleton with a map of the map of class/propertyinfos keyed on map hashcode, and also storing the calculated findFoo(x) methods.  On further thought we can provide a property for Multiton - so could use the propertyInfos and that would take care of managing a support instance per propertyinfo set.

jlhughes: Update: Multiton is not implemented on the java side - we'd have to replace 'new'.   
The support class can manage the map map lookups.  Let's leave this as an optimization exercise if/when needed.

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -49,13 +81,67 @@ foam.CLASS({
...
+      javaCode: `
+    if ( ((AuthService) x.get("auth")).check(x, "*") ) {
+      return super.select_(x, sink, skip, limit, order, predicate);
+    }
+
+    Predicate spidPredicate = predicate;
+
+    String spid = (String) x.get("spid");
+    if ( spid != null ) { // spid may be null during account creation.
+      if ( ServiceProviderAware.class.isAssignableFrom(getOf().getObjClass()) ) {
```
lchanmann, line 112:
```suggestion
      if ( obj instanceof ServiceProviderAware ) {
```

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -49,13 +81,67 @@ foam.CLASS({
...
+    {
+      name: 'select_',
+      javaCode: `
+    if ( ((AuthService) x.get("auth")).check(x, "*") ) {
+      return super.select_(x, sink, skip, limit, order, predicate);
+    }
+
+    Predicate spidPredicate = predicate;
+
+    String spid = (String) x.get("spid");
```
lchanmann, line 110: this is the only place that used the spid stored in the context and other places the spid is from the current user, are they different?

src/foam/nanos/ticket/services.jrl
```diff
@@ -37,7 +37,8 @@ p({
       .setDecorator(
         new foam.dao.history.HistoryDAO(x, "ticketHistoryDAO",
           new foam.nanos.ticket.TicketAddCommentDAO.Builder(x).setDelegate(
-            new foam.nanos.ticket.TicketOwnerDAO.Builder(x).build()).build())
+            new foam.nanos.ticket.TicketOwnerDAO.Builder(x).setDelegate(
+              new foam.dao.NullDAO.Builder(x).setOf(foam.nanos.ticket.Ticket.getOwnClassInfo()).build()).build()).build())
```
lchanmann, line 41: just curious, what is the effect of having a NullDAO in the decorator here?

src/foam/nanos/auth/ServiceProviderAwareDAO.js
jlhughes: See SessionServerBox.applyTo - spid is now added to the context so either will work, but going forward getting the spid from the context is the correct approach.

src/foam/nanos/ticket/services.jrl
jlhughes: On this branch all EasyDAO decorators must be terminate by a NullDAO rather than null.  EasyDAO initialization fails with NPE if any ProxyDAO.init is called which access delegate.of.  With null, the last delegate is null, with NullDAO a valid 'of' is available.

src/foam/nanos/auth/ServiceProviderAwareDAO.js
jlhughes: No 'obj' to test in 'select'.

lchanmann: Oh, that's right.

src/foam/nanos/auth/User.js
jlhughes: Track via CPF-4364

PR #2861, opened 2019-12-05
Map email args to user - EmailSettings

zahnno:
Replaces argument map values that reference User propertyInfo' with user values in email settings. This satisfies the condition where a notification references a single user and an extended or replaced doNotify interface method implies logic where multiple users notification settings are called. 

Each value referencing a User PropertyInfo is replaced with user values that pertain to the user receiving the notification, not the originating 'send to' user which the doNotify method is called on.

Example: 
```
args.put("account", accountNumber);
args.put("firstName", User.FIRST_NAME);

mapNotificationArguments will resolve argument list to:

args.put("account", accountNumber);
args.put("account", "userFirstName");
```

### DIFF
--- a/src/foam/nanos/notification/EmailSetting.js
+++ b/src/foam/nanos/notification/EmailSetting.js
@@ -11,14 +11,47 @@ foam.CLASS({
   extends: 'foam.nanos.notification.NotificationSetting',
 
   javaImports: [
+    'foam.core.PropertyInfo',
+    'foam.dao.DAO',
+    'foam.nanos.auth.User',
     'foam.nanos.logger.Logger',
     'foam.nanos.notification.email.EmailMessage',
     'foam.util.Emails.EmailsUtility',
     'java.util.Arrays',
+    'java.util.Iterator',
     'java.util.List',
+    'java.util.Map',
+    'static foam.mlang.MLang.EQ'
   ],
 
   methods: [
+    {
+      name: 'mapNotificationArguments',
+      type: 'Map',
+      documentation: `Iterate through arguments to replace propertyInfo values with the notified user' values.`,
+      args: [
+        { name: 'x', type: 'Context' },
+        { name: 'arguments', type: 'Map' },
+        { name: 'user', type: 'User' }
+      ],
+      javaCode: `
+        Logger logger = (Logger) x.get("logger");
+
+        Iterator entries = arguments.entrySet().iterator();
+        while (entries.hasNext()) {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/notification/EmailSetting.js
```diff
@@ -11,14 +11,47 @@ foam.CLASS({
...
+
+        Iterator entries = arguments.entrySet().iterator();
+        while (entries.hasNext()) {
+          Map.Entry entry = (Map.Entry) entries.next();
+          if ( entry.getValue() instanceof PropertyInfo ) {
+            if ( ! ((PropertyInfo) entry.getValue()).getClassInfo().equals(User.getOwnClassInfo()) ) {
+              logger.error("Cannot set a none User PropertyInfo as an email argument value");
+              continue;
+            }
+            String newValue = (String) ((PropertyInfo) entry.getValue()).get(user);
```
jlhughes: Does this throw an exception if get( argument ) is the wrong class? Thinking if that was the case then you could just try/catch and forgo the explicit test above. 
Also, in the test above does it support Contact and Business?

src/foam/nanos/notification/EmailSetting.js
```diff
@@ -11,14 +11,47 @@ foam.CLASS({
...
+            arguments.replace(entry.getKey(), newValue);
```
lchanmann: I wonder if setValue would work.
```suggestion
            entry.setValue(newValue);
```
If it does, we might be able to use for-each at line 40 as well.

src/foam/nanos/notification/EmailSetting.js
```diff
@@ -11,14 +11,47 @@ foam.CLASS({
...
     'java.util.Arrays',
+    'java.util.Iterator',
     'java.util.List',
+    'java.util.Map',
+    'static foam.mlang.MLang.EQ'
   ],
 
   methods: [
+    {
+      name: 'mapNotificationArguments',
```
lchanmann: looks like it is resolving the notification args instead of mapping :)

src/foam/nanos/notification/EmailSetting.js
zahnno: For each enforces generics too strictly for this. However, entry.setValue() does the same as replace so I'll go ahead and change that.

In regards to iterating over maps I found this to be useful https://www.testingexcellence.com/4-different-ways-iterate-map-java/

jlhughes (approved):

PR #2862, opened 2019-12-05
Rule Created Aware

blakegreen32:
Rule now implements these 4 classes:

'foam.nanos.auth.CreatedAware',
'foam.nanos.auth.CreatedByAware',
'foam.nanos.auth.LastModifiedAware',
'foam.nanos.auth.LastModifiedByAware'


src/foam/nanos/ruler/Rule.js
```diff
@@ -10,6 +10,13 @@
 
   documentation: 'Rule model represents rules(actions) that need to be applied in case passed object satisfies provided predicate.',
 
+  implements: [
+    'foam.nanos.auth.CreatedAware',
```
jlhughes, line 14: So you also need to added properties:
created, createdBy, .... 
Find another model with these are they are tableViews for the createdBy and LastModifiedBy to show the user. 
type of created and lastModified is DateTime.

jlhughes (approved):

PR #2866, opened 2019-12-06
Lifecycle aware

tharmaman:
Implemented LifecycleAware and attached it to the easyDAO.

Lifecycles consist of four stages: PENDING, ACTIVE, REJECTED and DELETED.

Viewing REJECTED and DELETED objects will require special permissions and under the LifecycleAwareDAO, a remove_ gets overridden with a put_ that sets the objects lifecycleState to DELETED similar to DeletedAware,


jlhughes (approved):

src/foam/dao/EasyDAO.js
```diff
@@ -194,6 +194,13 @@ foam.CLASS({
             .build();
         }
 
+        if ( getLifecycleAware() ) {
```
lchanmann: Can we have lifecycleAware before deletedAware? and log a deprecated warning in the deletedAware block.

nanoscott: Can both be enabled at the same time? If not, we should throw an error if both are set to true ..

nanoscott (approved):

src/foam/dao/EasyDAO.js
tharmaman: Done to both.

lchanmann:
LGTM, thanks @tharmaman.

PR #2867, opened 2019-12-07
[DO_NOT_MERGE] [CPF-3723] LogLevelTest

jlhughes:
Add Logging level TEST, only output during test runs. 
A replacement for System.out. 

**npm test** presently failing with:
`Failures:
1) ConsoleLogger should construct in default context
  Message:
    Expected function not to throw, but it threw [ false, 'Missing required import: testing in foam.log.ConsoleLogger' ].
  Stack:
    Error: Expected function not to throw, but it threw [ false, 'Missing required import: testing in foam.log.ConsoleLogger' ].
        at UserContext.<anonymous> (/Users/jhughes/workspace/repos/NANOPAY-logging/foam2/test/src/log/ConsoleLogger.js:82:12)
Pending:`

Originally 'testing' was named 'test', but changing to 'testing' and then to anything else results in the same failure. 
I've run out of ideas on how to fix the issue.


jlhughes:
Not needed

PR #2869, opened 2019-12-09
overwrite the validation in order to change the errorString set

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3728
note: currently the siginUp  view job title section error message is `Please enter a Your Job Title`. I added 
validationPredicates for this section in order to change the errorString to `Please select a Job Title`.  please let me know if there is a better way to do this.

### DIFF
--- a/src/foam/nanos/u2/navigation/SignUp.js
+++ b/src/foam/nanos/u2/navigation/SignUp.js
@@ -95,6 +95,15 @@ foam.CLASS({
         class: 'foam.u2.TextField',
         placeholder: 'Staff accountant'
       },
+      validationPredicates: [
+        {
+          args: ['department'],
+          predicateFactory: function(e) {
+            return e.NEQ(foam.nanos.u2.navigation.SignUp.DEPARTMENT, '');
+          },
+          errorString: 'Please select a Job Title.'
+        }
+      ],
       required: true
     },
     {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/u2/navigation/SignUp.js
```diff
@@ -95,6 +95,15 @@ foam.CLASS({
         class: 'foam.u2.TextField',
         placeholder: 'Staff accountant'
       },
+      validationPredicates: [
+        {
+          args: ['department'],
+          predicateFactory: function(e) {
+            return e.NEQ(foam.nanos.u2.navigation.SignUp.DEPARTMENT, '');
+          },
+          errorString: 'Please select a Job Title'
```
jlhughes: Is this a selection/drop-down?  as that is what 'select' suggests, while 'enter' suggest free form entry.

yij793: this isn't  a selection/drop-down. just an text input that only check if this section is empty or not

jlhughes: Then 'Please enter a Job Title' would be more appropriate.

TW80000: ^ I agree about 'enter' instead of 'select', but want to add that "job title" shouldn't be capitalized

yij793: Ok, chang made

jlhughes (approved):

TW80000 (approved):

PR #2870, opened 2019-12-09
add rw check as well when checking for read permission on AbstractPro…

noodlemoodle:
…pertyInfo

### DIFF
--- a/src/foam/core/AbstractPropertyInfo.java
+++ b/src/foam/core/AbstractPropertyInfo.java
@@ -157,7 +157,10 @@ public abstract class AbstractPropertyInfo
       AuthService auth = (AuthService) x.get("auth");
       String simpleName = this.getClassInfo().getObjClass().getSimpleName();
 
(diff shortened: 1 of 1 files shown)


TW80000 (approved):

src/foam/core/AbstractPropertyInfo.java
```diff
@@ -157,7 +157,7 @@ public void authorize(X x) {
       AuthService auth = (AuthService) x.get("auth");
       String simpleName = this.getClassInfo().getObjClass().getSimpleName();
 
-      if ( ! auth.check(x, String.format("%s.ro.%s", simpleName.toLowerCase(), this.getName().toLowerCase())) ) {
+      if ( ! auth.check(x, String.format("%s.ro.%s", simpleName.toLowerCase(), this.getName().toLowerCase())) && ! auth.check(x, String.format("%s.rw.%s", simpleName.toLowerCase(), this.getName().toLowerCase()))) {
```
kgrgreer: I think just concatenating the strings with + is much faster than using String.format().

kgrgreer: Don't call toLowerCase() twice on the same values. Store in variables.

PR #2871, opened 2019-12-09
CPF-3923/adding a generic delete modal to DAOSummaryView

Nauna:
#addresses: https://nanopay.atlassian.net/browse/CPF-3923
Simple but better User experience. Displays a simple white/red modal to confirm delete or cancel. 

notify() -> displays error or success

success goes back on stack, error stays on detailView

### DIFF
--- a/src/foam/comics/v2/DAOSummaryView.js
+++ b/src/foam/comics/v2/DAOSummaryView.js
@@ -116,7 +116,7 @@ foam.CLASS({
             this.finished.pub();
             this.stack.back();
           },
-          obj: this.data
+          data: this.data
         }));
       }
     }
--- a/src/foam/comics/v2/DAOSummaryView.js
+++ b/src/foam/comics/v2/DAOSummaryView.js
@@ -111,8 +111,11 @@ foam.CLASS({
       code: function() {
         this.add(this.Popup.create().tag({
           class: 'foam.u2.DeleteModal',
-          dao_: this.config.dao,
-          data: this,
+          dao: this.config.dao,
+          onDelete: () => {
+            this.finished.pub();
+            this.stack.back();
+          },
           obj: this.data
         }));
       }
--- a/src/foam/comics/v2/DAOSummaryView.js
+++ b/src/foam/comics/v2/DAOSummaryView.js
@@ -52,7 +52,8 @@ foam.CLASS({
     'foam.u2.layout.Cols',
     'foam.u2.layout.Rows',
     'foam.u2.ControllerMode',
-    'foam.u2.dialog.NotificationMessage'
+    'foam.u2.dialog.NotificationMessage',
+    'foam.u2.dialog.Popup',
   ],
   imports: [
     'stack'
@@ -107,18 +108,13 @@ foam.CLASS({
     },
     {
       name: 'delete',
-      confirmationRequired: true,
       code: function() {
-        this.config.dao.remove(this.data).then(o => {
-          this.finished.pub();
-          this.stack.back();
-        }, e => {
-          this.throwError.pub(e);
-          this.add(this.NotificationMessage.create({
-            message: e.message,
-            type: 'error'
-          }));
-        });
+        this.add(this.Popup.create().tag({
+          class: 'foam.u2.DeleteModal',
+          dao_: this.config.dao,
+          data: this,
+          obj: this.data
+        }));
       }
     }
   ],
--- a/src/foam/u2/DeleteModal.js
+++ b/src/foam/u2/DeleteModal.js
@@ -51,7 +51,7 @@ foam.CLASS({
   ],
 
   properties: [
-    'dao', 'onDelete', 'obj'
+    'dao', 'onDelete', 'data'
   ],
 
   methods: [
@@ -62,10 +62,10 @@ foam.CLASS({
         .start()
           .addClass(this.myClass('main'))
           .start('h2')
-            .add(this.TITLE).add(this.obj.cls_.name).add('?')
+            .add(this.TITLE).add(this.data.cls_.name).add('?')
           .end()
           .start('p')
-            .add(`${this.CONFIRM_DELETE_1} '${this.obj.toSummary()}'?`)
+            .add(`${this.CONFIRM_DELETE_1} '${this.data.toSummary()}'?`)
           .end()
         .end()
         .start()
@@ -83,7 +83,7 @@ foam.CLASS({
       name: 'delete',
       label: 'Delete',
       code: function(X) {
-        this.dao.remove(this.obj).then((_) => {
+        this.dao.remove(this.data).then((_) => {
           this.onDelete();
           this.notify(this.SUCCESS_MSG);
         }).catch((err) => {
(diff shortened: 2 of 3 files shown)


src/foam/u2/DeleteModal.js
```diff
@@ -0,0 +1,111 @@
...
+      color: /*%GREY1%*/ #525455;
+      border: none;
+      box-shadow: none;
+    }
+  `,
+
+  messages: [
+    { name: 'TITLE', message: 'Delete ' },
+    { name: 'CONFIRM_DELETE_1', message: 'Are you sure you want to delete ' },
+    { name: 'CONFIRM_DELETE_2', message: ' from your list?' },
```
jlhughes: Does this work in all cases? Is it that this delete used just in tableView?

src/foam/u2/DeleteModal.js
```diff
@@ -0,0 +1,111 @@
...
+      this.SUPER();
+      this
+        .addClass(this.myClass())
+        .start()
+          .addClass(this.myClass('main'))
+          .start('h2')
+            .add(this.TITLE).add(this.obj.cls_.name).add('?')
+          .end()
+          .start('p')
+            .add(`${this.CONFIRM_DELETE_1} '${this.obj.name}' ${this.CONFIRM_DELETE_2}`)
```
TW80000: Not all `FObject`s have a `name` property. I think `toSummary()` is safer.

src/foam/u2/DeleteModal.js
TW80000: I like the message more without this part. I think `Are you sure you want to delete Foo?` is simpler.

Nauna: So far tested with 2 models, but is built to be very generic. It is only used with DAOSummaryView, right now. So it's not apart of our more commonly used DAOController-which sets the tableView and such.

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -107,18 +108,13 @@ foam.CLASS({
...
-        }, e => {
-          this.throwError.pub(e);
-          this.add(this.NotificationMessage.create({
-            message: e.message,
-            type: 'error'
-          }));
-        });
+        this.add(this.Popup.create().tag({
+          class: 'foam.u2.DeleteModal',
+          dao_: this.config.dao,
```
TW80000: I'm pretty sure that conventionally an underscore at the end of a property name means it's "private" in scope, or intended to be used only internally by the view, not set or read by others. Please remove the underscore.

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -107,18 +108,13 @@ foam.CLASS({
...
+          data: this,
```
TW80000: I would have set `data` to the object since views are usually of FObjects, not other views. 

I also think it would be more flexible to give `DeleteModal` a callback function that gets called on successful delete instead of passing the view in. Passing the view like you do now tightly couples `DeleteModal` to the `DAOSummaryView`, whereas passing a callback method keeps the two separate. This way other places can use `DeleteModal` as well and provide different callbacks. Eg:

```suggestion
          data: this.data,
          onDelete: () => {
            this.finished.pub();
            this.stack.back();
          }
```

Then in `DeleteModal` when you call `remove` you can just call `onDelete` if it succeeds and print the success message.

TW80000 (changes requested):

src/foam/comics/v2/DAOSummaryView.js
Nauna: yup like this!

src/foam/u2/DeleteModal.js
Nauna: tested works

src/foam/comics/v2/DAOSummaryView.js
Nauna: it is private - in my mind- no one but the view is suppose to use it from here.

nonetheless, i see your point as well -not a biggy. Removed

src/foam/u2/DeleteModal.js
Nauna: One less message: I'm for it.

jlhughes (approved):

src/foam/u2/DeleteModal.js
```diff
@@ -0,0 +1,103 @@
...
+      border: none;
+      box-shadow: none;
+    }
+  `,
+
+  messages: [
+    { name: 'TITLE', message: 'Delete ' },
+    { name: 'CONFIRM_DELETE_1', message: 'Are you sure you want to delete' },
+    { name: 'SUCCESS_MSG', message: 'Successfully deleted' },
+    { name: 'FAIL_MSG', message: 'Failed to delete.' }
```
TW80000:
```suggestion
    { name: 'FAIL_MSG', message: 'Failed to delete' }
```

to be consistent with the success message.

src/foam/u2/DeleteModal.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+  ],
+
+  actions: [
+    {
+      name: 'delete',
+      label: 'Delete',
+      code: function(X) {
+        this.dao.remove(this.data).then((_) => {
+          this.onDelete();
```
TW80000: I see one potential issue the way it's currently written. If an error is thrown in the `onDelete` method, then the success notification won't be shown. However, we did indeed already delete the object since the call to the backend succeeded, so we should be telling the user that the delete succeeded.

Please move `this.notify(this.SUCCESS_MSG);` above the line this comment is on.

src/foam/u2/DeleteModal.js
```diff
@@ -0,0 +1,103 @@
...
+      this.SUPER();
+      this
+        .addClass(this.myClass())
+        .start()
+          .addClass(this.myClass('main'))
+          .start('h2')
+            .add(this.TITLE).add(this.data.cls_.name).add('?')
+          .end()
+          .start('p')
+            .add(`${this.CONFIRM_DELETE_1} '${this.data.toSummary()}'?`)
```
TW80000: Just a suggestion, but wouldn't it be more natural without the single quotes?

```suggestion
            .add(`${this.CONFIRM_DELETE_1} ${this.data.toSummary()}?`)
```

src/foam/u2/DeleteModal.js
```diff
@@ -0,0 +1,103 @@
...
+
+  messages: [
+    { name: 'TITLE', message: 'Delete ' },
+    { name: 'CONFIRM_DELETE_1', message: 'Are you sure you want to delete' },
+    { name: 'SUCCESS_MSG', message: 'Successfully deleted' },
+    { name: 'FAIL_MSG', message: 'Failed to delete.' }
+  ],
+
+  properties: [
+    'dao', 'onDelete', 'data'
```
TW80000: Optional: You could give these proper types. DAO could be a `foam.dao.DAOProperty`, `onDelete` can be a `Function`, and `data` can be an `FObjectProperty`.

TW80000 (approved):

src/foam/u2/DeleteModal.js
```diff
@@ -0,0 +1,115 @@
...
+
+  properties: [
+    {
+      class: 'foam.dao.DAOProperty',
+      name: 'dao'
+    },
+    {
+      class: 'Function',
+      name: 'onDelete',
+      value: () => {}
```
kgrgreer: This is already the default value.

PR #2872, opened 2019-12-10
Automatic set to pending for lifecycle aware create

tharmaman:


src/foam/comics/v2/DAOCreateView.js
```diff
@@ -82,7 +86,13 @@ foam.CLASS({
     {
       name: 'save',
       code: function() {
-        this.config.dao.put(this.data).then((o) => {
+        var clonedData = this.data.clone();
```
nanoscott: Shouldn't the clone be inside the if statement, since that's the only case it is actually necessary? So you do:
var cdata = this.data; 
if (LcAware.isInstance(cdata)) { 
  cdata = cdata.clone(); 
  ...

src/foam/comics/v2/DAOCreateView.js
```diff
@@ -82,7 +86,13 @@ foam.CLASS({
...
+
+        if ( foam.nanos.auth.LifecycleAware.isInstance(clonedData) ) {
```
TW80000: Can `require` this and change to:
```suggestion
        if ( this.LifecycleAware.isInstance(clonedData) ) {
```
Same for `LifecycleState` below.

src/foam/comics/v2/DAOCreateView.js
tharmaman: sounds good will account for it so theres no unnecessary clones

src/foam/comics/v2/DAOCreateView.js
tharmaman: thanks troykashi

jlhughes (approved):

PR #2876, opened 2019-12-10
Check if model has more than one section

lchanmann:
Check sections on the model and its ancestors.


TW80000 (approved):
As a tiny space optimization, you can just increment a count rather than use a Set since we only care about the size of the set anyway.

lchanmann:
Updated.

PR #2878, opened 2019-12-11
fix rule view

carl-zzz:
RuleDAO view has "Cannot read property 'find' of undefined" error. This will fix the error. 

```
Rule.js:215 Uncaught (in promise) TypeError: Cannot read property 'find' of undefined
    at Object.tableCellFormatter (Rule.js:215)
    at Object.format (FnFormatter.js:19)
    at Object.callOn (Element.js:1759)
    at Object.<anonymous> (UnstyledTableView.js:431)
    at Array.forEach (<anonymous>)
    at Object.forEach (Element.js:1782)
    at Object.<anonymous> (UnstyledTableView.js:427)
    at Object.addRow (Element.js:1713)
    at Element.js:591
```


jlhughes:
Is this because Rule.js does not import userDAO?

carl-zzz:
> Is this because Rule.js does not import userDAO?

@jlhughes Both work... Which way is better?

src/foam/nanos/ruler/Rule.js
```diff
@@ -212,7 +212,7 @@
       name: 'createdBy',
       visibility: 'RO',
       tableCellFormatter: function(value, obj) {
-        obj.userDAO.find(value).then(function(user) {
+        this.__subSubContext__.userDAO.find(value).then(function(user) {
```
kgrgreer: Best not to mess around with contexts directly. They're meant to be an internal thing and only used when you have no other option. Easier and better to just declare that you import userDAO.

kgrgreer:
Please resolve merge conflict.

PR #2879, opened 2019-12-11
Cpf 3645/resend email is not work after sigin up

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3645
note:
`signUp` after use sigin up they will log out and stoping them resend verification email.
`signIn` if unverified user try to login they will login both `verifyEmail` screen and `dashboard` screen. (
(
the way to recreate: 
0.must be unverified use
1.login
2.go back to ablii
3.login again
)
^
Screen Shot 2019-12-11 at 11 57 08 AM

### DIFF
--- a/src/foam/nanos/u2/navigation/SignUp.js
+++ b/src/foam/nanos/u2/navigation/SignUp.js
@@ -208,7 +208,6 @@ foam.CLASS({
           location.reload();
         } else {
           // logout once we have finished updating documents.
-          this.auth.logout();
           this.stack.push({
             class: 'foam.nanos.auth.ResendVerificationEmail'
           });
--- a/src/foam/nanos/u2/navigation/SignUp.js
+++ b/src/foam/nanos/u2/navigation/SignUp.js
@@ -207,7 +207,6 @@ foam.CLASS({
           window.history.replaceState(null, null, window.location.origin);
           location.reload();
         } else {
-          // logout once we have finished updating documents.
           this.stack.push({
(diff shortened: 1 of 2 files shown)


src/foam/nanos/u2/navigation/SignUp.js
```diff
@@ -208,7 +208,6 @@ foam.CLASS({
           location.reload();
         } else {
           // logout once we have finished updating documents.
-          this.auth.logout();
```
kgrgreer, line 211: You should remove the comment on the previous line that says you're logging out, if you're no longer logging out.

src/foam/nanos/u2/navigation/SignUp.js
```diff
@@ -208,7 +208,6 @@ foam.CLASS({
           location.reload();
         } else {
           // logout once we have finished updating documents.
```
jlhughes: Perhaps the comment should also be removed.

Nauna:
@yij793 did you test the context at the point where you think you are logged in to see if the user is indeed logged in? Believe the SideNav - menu is toggled with loginSuccess, so it's possible this is not getting set correctly, considering you moved this, isn't that the problem??

and further, the change  you made by deleting the logout, you are virtually keeping the user logged in - in the context even though you are just redirecting the user to another view.

The user gets logged in to save the acceptanceDocuments. They should be logged out after this happens. 

This is a potential security risk.

@TW80000 can you give your opinion

PR #2880, opened 2019-12-11
add one more string validation for minlength and maxlength are not given

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3649
note: currently validation for String class only trigger once if no given minlength or maxlength, add one more condition for the `StringPropertyValidationRefinement`
Screen Shot 2019-12-11 at 3 27 13 PM

### DIFF
--- a/src/foam/core/Validation.js
+++ b/src/foam/core/Validation.js
@@ -141,7 +141,7 @@ foam.CLASS({
             errorString: `Please enter a ${this.label} with at most ${this.maxLength} character${this.maxLength>1?'s':''}`
           });
         }
-        if ( this.required && ! foam.Number.isInstance(this.maxLength) && ! foam.Number.isInstance(this.minLength) ) {
+        if ( this.required && ! foam.Number.isInstance(this.minLength) ) {
           a.push({
             args: [this.name],
             predicateFactory: function(e) {
--- a/src/foam/core/Validation.js
+++ b/src/foam/core/Validation.js
@@ -141,6 +141,15 @@ foam.CLASS({
             errorString: `Please enter a ${this.label} with at most ${this.maxLength} character${this.maxLength>1?'s':''}`
           });
         }
+        if ( this.required && ! foam.Number.isInstance(this.maxLength || this.minLength) ) {
+          a.push({
+            args: [this.name],
+            predicateFactory: function(e) {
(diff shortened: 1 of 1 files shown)


src/foam/core/Validation.js
```diff
@@ -141,6 +141,15 @@ foam.CLASS({
             errorString: `Please enter a ${this.label} with at most ${this.maxLength} character${this.maxLength>1?'s':''}`
           });
         }
+        if ( this.required && ! foam.Number.isInstance(this.maxLength || this.minLength) ) {
```
TW80000:
```suggestion
        if ( this.required && ! foam.Number.isInstance(this.maxLength) && ! foam.Number.isInstance(this.minLength) ) {
```

Right now you're or-ing two values that may or may not be numbers. Which I guess works due to short-circuiting, but `0 || undefined` evaluates to `undefined`, so it wouldn't work for that case. But then again, I don't see why maxLength would ever be 0.

src/foam/core/Validation.js
```diff
@@ -141,6 +141,15 @@ foam.CLASS({
...
+          a.push({
+            args: [this.name],
+            predicateFactory: function(e) {
+              return e.GTE(foam.mlang.StringLength.create({ arg1: self }), 1);
+            },
+            errorString: `Please enter ${this.label}`
```
TW80000:
```suggestion
            errorString: `Please enter a ${this.label}`
```

to be consistent with the other error messages.

src/foam/core/Validation.js
```diff
@@ -141,6 +141,15 @@ foam.CLASS({
             errorString: `Please enter a ${this.label} with at most ${this.maxLength} character${this.maxLength>1?'s':''}`
           });
         }
+        if ( this.required && ! foam.Number.isInstance(this.maxLength) && ! foam.Number.isInstance(this.minLength) ) {
```
TW80000: What if `maxLength: 10` and `required: true` are set, but `minLength` isn't? Then the `required` predicate won't apply, even though we want it to. So I think this condition needs to be updated to:

```suggestion
        if ( this.required && ! foam.Number.isInstance(this.minLength) ) {
```

Sorry I didn't catch this the first time.

TW80000 (changes requested):

TW80000 (approved):

PR #2881, opened 2019-12-12
Fix capitalization in generated error messages

TW80000:
Eg: "Please enter a Phone Number" becomes "Please enter a phone number"

### DIFF
--- a/src/foam/core/Validation.js
+++ b/src/foam/core/Validation.js
@@ -102,7 +102,7 @@ foam.CLASS({
         }
         return !required ? null : [[name],
           function() {
-            return !this.hasOwnProperty(name) && (`Please enter a ${label}`);
+            return !this.hasOwnProperty(name) && (`Please enter a ${label.toLowerCase()}`);
           }]
       },
     },
@@ -129,7 +129,7 @@ foam.CLASS({
             predicateFactory: function(e) {
               return e.GTE(foam.mlang.StringLength.create({ arg1: self }), self.minLength);
             },
-            errorString: `Please enter a ${this.label} with least ${this.minLength} character${this.minLength>1?'s':''}`
+            errorString: `Please enter a ${this.label.toLowerCase()} with least ${this.minLength} character${this.minLength>1?'s':''}`
           });
         }
         if ( foam.Number.isInstance(this.maxLength) ) {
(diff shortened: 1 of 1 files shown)


jlhughes (approved):
Of course just works for the general case.  
Perhaps a ticket to capture future work to only lower case if only first letter is capitalized.

TW80000:
Yeah, and we could be smarter about using 'an' instead of 'a' for labels that start with a vowel. But that opens up a whole new can of worms if we start thinking about i18n as well.

kgrgreer:
You don't use "a" vs. "an" if the next word starts with a vowel or not, but if it sounds like a vowel.
Ex. An honour or A university.
Which means this would be very difficult to determine programmatically, and could even be different depending on the locale (not just language) of the user, depending on whether they pronounce things like leading H's or not.

src/foam/core/Validation.js
```diff
@@ -102,7 +102,7 @@ foam.CLASS({
         }
         return !required ? null : [[name],
           function() {
-            return !this.hasOwnProperty(name) && (`Please enter a ${label}`);
+            return !this.hasOwnProperty(name) && (`Please enter a ${label.toLowerCase()}`);
```
kgrgreer, line 105: `Please provide a value for ${label.toLowerCase()}` avoids the a vs an problem.

src/foam/core/Validation.js
```diff
@@ -129,7 +129,7 @@ foam.CLASS({
             predicateFactory: function(e) {
               return e.GTE(foam.mlang.StringLength.create({ arg1: self }), self.minLength);
             },
-            errorString: `Please enter a ${this.label} with least ${this.minLength} character${this.minLength>1?'s':''}`
+            errorString: `Please enter a ${this.label.toLowerCase()} with least ${this.minLength} character${this.minLength>1?'s':''}`
```
kgrgreer, line 132: Same as above, here and below.

src/foam/core/Validation.js
TW80000: Even simpler is "This field cannot be blank" or "This field is required". What do you think about that instead?

jlhughes: required fields will have a red boarder and red star underneath with this text. 

> * required
 
I think is sufficient. No issues with locale, property names. ... 
The placeholder text can handle provide a template/default/example.

jlhughes: required fields will have a red boarder and red star underneath with this text. 

> * required
 
I think this is sufficient. No issues with locale, property names. ... 
The placeholder text can handle providing a template/default/example.

src/foam/core/Validation.js
jlhughes: also if sticking with the remainder then 
with least -> with at least

TW80000:
This requires more thought, closing.

PR #2882, opened 2019-12-12
Fix a few issues with section-aware detail views

TW80000:
There are a few issues with the current implementation of sections and
detail views that extend AbstractSectionedDetailView that this commit
seeks to remedy.

The first issue is that sections are being rendered even if there are no
visible properties or actions in that section.

The second issue is that, due to a recent change in the way
SectionedDetailView is implemented, sections would be rendered side by
side when they should have been taking up the full width of the parent.
This issue is caused by an extra div being inserted when we return an
element from a slot in detail views. This extra div breaks the CSS grid
and thus the positioning.

The first approach I took to fix the second issue was to use an
ArraySlot of all of the sections and do the forEach inside of the slot's
return value instead of doing a slot of each section individually inside
of the forEach like it currently does (where is where the extra div is
being introduced). This approach had a downside, however. The detail
view would visibly 'flicker' while the slots were re-rendering. This
secondary issue was happening because slots can be hidden if the user
doesn't have permission to see them. Either we hide them by default and
show them if the user does end up having permission, or we show them by
default and hide if the user doesn't end up having permission. Either
way, the view changes. When several sections are doing this at roughly
the same time, the effect is jarring.

The approach taken to fix this secondary issue is to calculate the
visibility of all sections up front while the view is hidden, then show
the view when we're done waiting for all of the async permission checks.
This approach has a few other desirable properties, listed below.

* We can keep track of the properties and actions in each section that
  are visible based on the current 'data' and 'controllerMode' and get
  the detail views to only try to render those properties and actions.
* We also fix the first issue, because while we're calculating which
  sections are visible, we can check if there is at least one visible
  property or action in that section and not include it in the list if
  there isn't one.
* The "interface" that views that extend AbstractSectionedDetailView
  need to work will becomes simpler. In initE, all you need to worry
  about is putting everything inside a slot of `sections`.
  AbstractSectionedDetailView will make sure that `sections` only
  contains visible sections, and will set the properties and actions
  arrays for those sections to be only those that are visible. It does
  this properly by recalculating whenever `data` or `of` changes. This
  removes the need for the views extending AbstractSectionedDetailView
  to use slots of `data`. The slot of `sections` is a slot on `data`
  implicitly.

### DIFF
--- a/src/foam/core/lib.js
+++ b/src/foam/core/lib.js
@@ -84,6 +84,7 @@ Object.defineProperty(
 foam.assert = function assert(cond) {
   if ( ! cond ) {
     console.assert(false, Array.from(arguments).slice(1).join(' '));
+    if ( foam.isServer ) console.trace();
   }
 
   return cond;
--- a/src/foam/core/lib.js
+++ b/src/foam/core/lib.js
@@ -84,7 +84,6 @@ Object.defineProperty(
 foam.assert = function assert(cond) {
   if ( ! cond ) {
     console.assert(false, Array.from(arguments).slice(1).join(' '));
-    console.trace();
   }
 
   return cond;
--- a/src/foam/u2/detail/AbstractSectionedDetailView.js
+++ b/src/foam/u2/detail/AbstractSectionedDetailView.js
@@ -13,6 +13,10 @@ foam.CLASS({
     The abstract for property-sheet style Views with sections for editing an FObject.
   `,
 
+  imports: [
+    'auth'
+  ],
+
   requires: [
     'foam.core.Action',
     'foam.core.Property',
@@ -60,7 +64,7 @@ foam.CLASS({
     {
       class: 'FObjectArray',
       of: 'foam.layout.Section',
-      name: 'sections',
+      name: 'sections_',
       factory: null,
       expression: function(of) {
         if ( ! of ) return [];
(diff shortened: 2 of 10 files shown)


src/foam/core/lib.js
```diff
@@ -84,7 +84,6 @@ Object.defineProperty(
 foam.assert = function assert(cond) {
   if ( ! cond ) {
     console.assert(false, Array.from(arguments).slice(1).join(' '));
-    console.trace();
```
KernelDeimos, line 87: I really like when these show up - does the trace get printed somewhere else?

src/foam/u2/detail/AbstractSectionedDetailView.js
```diff
@@ -102,6 +106,214 @@ foam.CLASS({
...
+       *      'isAvailable' method if there is one.
+       *   3. That the section has at least one visible property or action,
+       *      which is based on:
+       *        (a) Whether or not the property or action requires permission to
+       *            be seen.
+       *        (b) Whether or not the action is visible based on the
+       *            'isAvailable' method of the action.
+       *        (c) Whether or not the property is visible based on the
+       *            controllerMode, visibility, or visibilityExpression.
+       */
```
KernelDeimos: Can listeners be FOAM method objects with a `documentation` property?

src/foam/u2/detail/AbstractSectionedDetailView.js
```diff
@@ -102,6 +106,214 @@ foam.CLASS({
...
+            });
+
+            // Filter out actions that the user doesn't have permission to see.
+            var actionPermissionPromises = visibleActions.map(action => {
+              return Promise.all(action.availablePermissions.map(permission => {
+                return this.auth.check(null, permission);
+              }));
+            });
+
+            return Promise.all([
```
KernelDeimos: Do these need `catch` callbacks or is it okay like this?

src/foam/core/lib.js
KernelDeimos: Oh lol, turns out I added that line myself and forgot. I think I remember adding it because there wasn't a trace printed for this anywhere, but is there perhaps a better way to do this?

jlhughes (approved):
Read through it, this is impressive. 
Looks good.

src/foam/core/lib.js
TW80000: I removed this out of personal preference, but if people prefer to keep it then I can revert this change. Personally I find the stack trace takes up too much space in the console.

src/foam/u2/detail/AbstractSectionedDetailView.js
```diff
@@ -102,6 +106,214 @@ foam.CLASS({
...
+      name: 'sections',
+      factory: function() {
+        return this.sections_;
+      }
+    },
+    {
+      class: 'Long',
+      name: 'outstandingCalculations_',
+      documentation: 'Used alongside `loading` to hide the view while it is loading',
+      factory: function() {
```
kgrgreer: Why is this factory needed?

src/foam/u2/detail/AbstractSectionedDetailView.js
kgrgreer: You can put listeners in long-form with documentation:
{
  name: 'updateSections',
  documentation: '...',
  code: function() {...}
}

src/foam/u2/detail/AbstractSectionedDetailView.js
TW80000: I don't think it is. Was just sanity checking while debugging and forgot to remove this afterwards. Will remove.

src/foam/core/lib.js
TW80000: Addressed by 1d8e5c759.

src/foam/u2/detail/AbstractSectionedDetailView.js
TW80000: Addressed by 1d8e5c759.

src/foam/u2/detail/AbstractSectionedDetailView.js
TW80000: Addressed by 1d8e5c759.

src/foam/u2/detail/AbstractSectionedDetailView.js
TW80000: Addressed by 1d8e5c759.

KernelDeimos (approved):

PR #2883, opened 2019-12-12
fix freezing issue - infinite loop issue in browser

Nauna:


src/foam/comics/DAOController.js
```diff
@@ -186,7 +186,7 @@ foam.CLASS({
         var sectionName = undefined;
         var hasManySections = false;
         var model = this.data.of.model_;
-        while ( model && model !== foam.core.FObject ) {
+        while ( model && model === foam.core.FObject ) {
```
zahnno, line 189: I think this defeats the purpose of this model parent iteration and looking for any associated sections. I think this might be something we're looking for:
```suggestion
        while ( model && model.id != 'foam.core.FObject' ) {
```

TW80000, line 189: @lchanmann does this make sense to you? Why was it `!==` originally? Why can't we just do `while ( model ) {` ?

lchanmann: Line 201, we look up the parent of the model then call `.model_`. I think we can instead compare `model && model.id !== 'foam.core.FObject'` to fix the issue.

The logic here is to traverse the model ancestors until we can determine if the model has many sections. And FObject is the root of all models so we want to stop there.

kgrgreer: This doesn't make sense to me. It looks like if it is any model other than FObject it will just not work. Also, easier to traverse up the classes by starting with this.data.of and then keep calling getSuperClass(). Also, 'model' is a Model but foam.core.FObject is a Class, which is probably the problem. Then from the class get the sections as cls.model_.sections.

zahnno: Thought I submitted this comment yesterday :/

lchanmann:
I made another PR https://github.com/foam-framework/foam2/pull/2885 for this in case we cannot wait for @Nauna back to address the comment. @kgrgreer @TW80000 @zahnno @jlhughes

jlhughes:
Replaced by https://github.com/foam-framework/foam2/pull/2885

PR #2884, opened 2019-12-13
[CPF-3729] Partial builds for genJava - cut 70% off builds for small changes

KernelDeimos:
This PR would have been difficult to split into multiple parts, so bear with me;

## Story Time

Enter `genJava`

- The genJava script was unaware of which files have been modified
- Even if `genjava.js` knew what files were changed, it operates on class names from `classes.js` and has never needed to care about file paths.
- The `loadServer` function in foam.js does add the source file to each model, but 163 classes were still not filtered because they were loaded from the ClassLoader.

Enter `foam.apploader.ClassLoader`

- When a model is loaded using the ClassLoader, the result is not consistent with files.js. The reason for this is ClassLoader finds files based on the classname, and therefore doesn't support having multiple models in a file.

Enter "walking all the files and generating a map of model IDs to filenames"

- Okay so this was complicated. Without invoking the classloader (that would be a bit of a catch 22) or depending on files.js (wouldn't it be nice if that was generated for us?), I saw two options for processing the javascript files:
  - Using a javascript AST parser
  - Run javascript files in an environment where `foam.CLASS` (and family) just reads the name of a model to store it in a map

I went the second option because I figured it would be the least horrible. It was more complicated than expected (classpaths like `foam.u2.CSS` being referenced outside of any functions, and similar issues), so I used a javascript native Proxy object that recurses to trick source files into thinking these variables and functions exist (so that an error doesn't terminate processing). It worked great! The only caveat is each "root variable" (ex: `foam`, `net`, `com`) needs to be specified. This is easy to maintain; I was able to get full coverage with just a few.

## FOAMLink

I wanted the map of classes to be output to a JSON file so it could be re-used by other build tools in the future (like the javascript builder), since the file-walking and model processing adds about 2 seconds.  This meant I needed to make a node script which also meant I had to name something. I decided to call it foamlink.

### `foamlink.js` Manifest Files

The foamlink script walks the file tree and keeps an eye out for `foamlink.js` files. A foamlink.js manifest contains extra information for edge-cases or files and directories that need to be ignored. Luckily these files need just a few entries to get everything working.

## `files.js`

While `files.js` is still needed for the javascript builder and projects not using liveScriptBundler, it could be generated with the help of FOAMLink. This will require some additional work to figure out dependancies, for which foamlink.js files could perhaps provide manual overrides.


src/foam/foamlink/FoamlinkData.js
```diff
@@ -0,0 +1,45 @@
+foam.CLASS({
```
TW80000, line 7: Missing license header for most new files in this PR.

src/foam/foamlink/FoamlinkData.js
```diff
@@ -0,0 +1,45 @@
...
+  ],
+  methods: [
+    function add(fileName, idsPresent) {
+      this.filesToClasses[fileName] = idsPresent;
+      // idsPresent could be an array if it was set using the MANUAL function
+      // in a foamlink.js file; it's usually a map with model ids as keys
+      if ( ! Array.isArray(idsPresent) ) {
+        idsPresent = Object.keys(idsPresent);
+      }
+      for ( var i=0; i < idsPresent.length; i++ ) {
```
TW80000:
```suggestion
      for ( var i = 0; i < idsPresent.length; i++ ) {
```

src/foam/foamlink/FoamlinkNodeModelFileDAO.js
```diff
@@ -0,0 +1,26 @@
...
+    'foam.apploader.ModelFileDAO',
+    'foam.foamlink.FoamlinkNodeModelFileFetcher',
+  ],
+  properties: [
+    'root',
+    {
+      name: 'delegate',
+      factory: function() {
+        return this.ModelFileDAO.create({
+          fetcher: this.FoamlinkNodeModelFileFetcher.create({root: this.root}), });
```
TW80000:
```suggestion
          fetcher: this.FoamlinkNodeModelFileFetcher.create({ root: this.root })
        });
```

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+      });
+    },
+    function exec(path) {
+      var self = this;
+      path = path || process.cwd();
+
+      return new Promise((resolve, reject) => {
+        this.processFoamlinkFile(path, true);
+
+        var walkPromises = [];
```
TW80000: Alternatively:

```suggestion
        Promise.all(this.repositories_.map(repoPath => this.startWalking(repoPath, reject))).then(resolve);
```

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+        for ( repoPath in this.repositories_ ) {
+          walkPromises.push(this.startWalking(repoPath, reject));
+        }
+
+        Promise.all(walkPromises).then(resolve);
```
TW80000: Do we need a `return` here?

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+      });
+    },
+    function startWalking(repoPath, reject) {
+      var self = this;
+
+      try {
+        this.processFoamlinkFile(repoPath, false, true);
+      } catch (e) {
+        reject(e)
```
TW80000:
```suggestion
        reject(e);
```

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+      }
+
+      var walker = foam.util.filesystem.FileWalker.create();
+      walker.files.sub((_1, _2, info) => {
+        try {
+          self.handleDirectory(repoPath, info);
+        } catch (e) {
+          console.error(e);
+          reject(e)
```
TW80000:
```suggestion
          reject(e);
```

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+      return walker.walk(repoPath);
+    },
+    function handleDirectory(repoPath, info) { // throws io
+      // These will be removed when a logger is used
+      var tag = "\033[36;1m<<<FOAMLINK>>>\033[0m ";
+      var g = (s) => "\033[32;1m" + s + "\033[0m ";
+      var w = "\033[33;1m[WARN]\033[0m ";
+
+      for ( k in this.ignorePaths_ ) {
+        if ( info.directory === k || info.directory.startsWith(k+'/') )
```
TW80000:
```suggestion
        if ( info.directory === k || info.directory.startsWith(k + '/') )
```

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+          || this.ignorePaths_[fileInfo.fullPath]
+        ) continue;
+
+        // console.log('   - >' + fileInfo.fullPath);
+        let text = this.nodejs_.fs.readFileSync(fileInfo.fullPath)
+        text = text.toString();
+
+        // Add any code specified by 'INJECT' in a foamlink file
+        if ( this.injectFiles_[fileInfo.fullPath] ) {
+          injectCode = this.injectFiles_[fileInfo.fullPath].join('\n')+'\n';
```
TW80000:
```suggestion
          injectCode = this.injectFiles_[fileInfo.fullPath].join('\n') + '\n';
```

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+          throw e;
+        }
+
+        for ( id in idsPresent ) {
+          if (
+            !(id === expectedPackage || id.startsWith(expectedPackage+'.'))
+          ) {
+            console.warn(tag + w +
+              'Package name does not match file path: ' +
+              '"'+id+'" is outside of "'+expectedPackage+'".'
```
TW80000:
```suggestion
              '"' + id + '" is outside of "' + expectedPackage + '".'
```

src/foam/foamlink/FoamlinkNodeModelFileFetcher.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
TW80000:
```suggestion
 * Copyright 2019 The FOAM Authors. All Rights Reserved.
```

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+        let fileInfo = info.files[name];
+        
+        if ( // These are the conditions in which a file is skipped
+          fileInfo.name === 'foamlink.js'
+          || ! fileInfo.name.endsWith('.js')
+          || this.reservedNames_.includes(fileInfo.name)
+          || this.ignorePaths_[fileInfo.fullPath]
+        ) continue;
+
+        // console.log('   - >' + fileInfo.fullPath);
```
TW80000: Remove

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+        }
+      }
+      __foamlink__.MANUAL = function(filePath, classes) {
+        let fullPath = self.nodejs_.path.join(path, filePath);
+        // This is the sad part of foamlink for when a file can't be processed
+        self.results.add(fullPath, classes);
+        self.ignorePaths_[fullPath] = {}
+      }
+      __foamlink__.ROOTS = function() {
+        for ( var i=0; i < arguments.length; i++ ) {
```
TW80000:
```suggestion
        for ( var i = 0; i < arguments.length; i++ ) {
```

src/foam/foamlink/FoamlinkExec.js
```diff
@@ -0,0 +1,217 @@
...
+          foam.foamlink.lib.addRoot(arguments[i]);
+        }
+      }
+
+      // Process the foamlink.js file
+      with ( __foamlink__ ) { eval(text); }
+    },
```
TW80000:
```suggestion
    }
```

src/foam/foamlink/FoamlinkNodeModelFileFetcher.js
```diff
@@ -0,0 +1,43 @@
...
+          ret(['withsource', {
+            text: js,
+            source: path
+          }]);
+        } catch(e) {
+          console.error(e);
+          ret(null);
+        }
+      });
+    },
```
TW80000:
```suggestion
    }
```

src/foam/util/filesystem/FileWalker.js
```diff
@@ -0,0 +1,120 @@
+foam.CLASS({
+  package: 'foam.util.filesystem',
+  name: 'FileWalker',
+  topics: ['files', 'directoryStart', 'directoryEnd', 'error'],
+  documentation: `
```
TW80000, line 7: Just a suggestion/personal preference, but I find it easier to scan a model file when there are empty lines between axioms (except for package, name and extends).

src/foam/util/filesystem/FileWalker.js
```diff
@@ -0,0 +1,120 @@
...
+      });
+    },
+    function walk(path, depth) {
+      if ( typeof depth === 'undefined' ) {
+        depth = 0;
+      }
+
+      var self = this;
+      return new Promise(function(resolve, reject) {
+        self.nodejs_.fs.readdir(path, function (err, files) {
```
TW80000:
```suggestion
        self.nodejs_.fs.readdir(path, function(err, files) {
```

src/foam/util/filesystem/FileWalker.js
```diff
@@ -0,0 +1,120 @@
...
+            return;
+          }
+
+          filesMessage = {
+            directory: path,
+            files: {}
+          };
+          directoriesToWalk = [];
+
+          files.forEach(function (file) {
```
TW80000:
```suggestion
          files.forEach(function(file) {
```

src/foam/util/filesystem/FileWalker.js
```diff
@@ -0,0 +1,120 @@
...
+              self.error.pub('stat', e);
+              return;
+            }
+            if ( stats.isDirectory() ) {
+              directoriesToWalk.push(fullPath);
+            } else {
+              filesMessage.files[file] = {
+                name: file,
+                fullPath: fullPath,
+                stats: stats,
```
TW80000:
```suggestion
                stats: stats
```

src/foam/util/filesystem/FileWalker.js
```diff
@@ -0,0 +1,120 @@
...
+              };
+            }
+          });
+
+          let recursePromises = [];
+          switch ( self.order ) {
+          case 'BFS':
```
TW80000, line 95: Indentation

src/foam/util/filesystem/FileWalker.js
```diff
@@ -0,0 +1,120 @@
...
+            self.directoryStart.pub(path);
+            self.files.pub(filesMessage);
+            directoriesToWalk.forEach((fullPath) => {
```
TW80000: Another personal preference kind of thing, but any time you create an empty array and loop over something else to populate the new array you can usually just map the old array (and filter if need be (or use `reduce` instead of `map` and `filter` to do both in one go)). Personally I find it easier to read that way since it's more declarative that way.

Eg:

```JavaScript
let ys = [];

for ( let i = 0; i < xs.length; i++ ) {
  ys.push(foo(xs[x]));
}
```

can be written as

```
let ys = xs.map(foo);
```

Eg with filtering:

```JavaScript
let ys = [];

for ( let i = 0; i < xs.length; i++ ) {
  if ( bar(xs[x]) ) {
    ys.push(foo(xs[x]));
  }
}
```

could be written as

```JavaScript
let ys = xs.filter(bar).map(foo);
```

or

```JavaScript
let ys = xs.reduce((acc, x) => bar(x) ? acc.concat([x]) : acc, []);
```

src/foam/util/filesystem/FileWalker.js
```diff
@@ -0,0 +1,120 @@
...
+              recursePromises.push(
+                self.walk(fullPath, depth+1)
```
TW80000:
```suggestion
                self.walk(fullPath, depth + 1)
```

src/foam/util/filesystem/FileWalker.js
```diff
@@ -0,0 +1,120 @@
...
+            Promise.all(recursePromises).then(() => {
+              self.directoryEnd.pub(path);
+              resolve();
+            });
+            break;
+          case 'DFS':
+            self.directoryStart.pub(path);
+            directoriesToWalk.forEach((fullPath) => {
+              recursePromises.push(
+                self.walk(fullPath, depth+1)
```
TW80000:
```suggestion
                self.walk(fullPath, depth + 1)
```

src/foam/util/filesystem/doc.flow
```diff
@@ -0,0 +1,3 @@
+<title>Package foam.util.filesystem</title>
+
+This package contains utilities for interacting with the filesystem.
```
TW80000, line 3: Could add a little more here if you don't mind. Even pasting the description of this PR in could be helpful.

tools/foamlink/foamlink.js
```diff
@@ -0,0 +1,39 @@
+if ( typeof RUNNING_IN_FOAMLINK !== 'undefined' )
+  ABORT("this file is not a foamlink manifest");
```
TW80000:
```suggestion
  ABORT("This file is not a foamlink manifest");
```

tools/genjava.js
```diff
@@ -47,6 +58,16 @@ externalFile.blacklist.forEach(function(cls) {
   blacklist[cls] = true;
 });
 
+var fileWhitelist = null;
+var classesNotFound = {};
+var classesFound = {};
+if ( incrementalMeta !== null ) {
+  fileWhitelist = {}; // set
+  for ( var i=0; i < incrementalMeta.modified.length; i++ ) {
```
TW80000:
```suggestion
  for ( var i = 0; i < incrementalMeta.modified.length; i++ ) {
```

tools/genjava.js
```diff
@@ -289,4 +330,20 @@ addDepsToClasses().then(function() {
...
   proxies.forEach(generateProxy);
+}).then(function () {
+  var notFound = Object.keys(classesNotFound).length;
+  var found = Object.keys(classesFound).length;
+
+  if ( notFound > 0 ) {
+    var allKeys = {};
+    for ( k in classesNotFound ) allKeys[k] = true;
+    for ( k in classesFound ) allKeys[k] = true;
+    console.log(''+found+'/'+Object.keys(allKeys).length+' '+
```
TW80000:
```suggestion
    console.log(`${found}/${Object.keys(allKeys).length} sources found (${notFound} missing));
```

TW80000 (commented):
First round of comments, all very superficial. Will do a more meaningful review after.

src/foam/foamlink/FoamlinkExec.js
KernelDeimos: This will be okay because calling `resolve` is equivalent to returning a promise

KernelDeimos:
Right now genjava uses foamlink mode by default which could cause problems if a project's build isn't configured for it. I think I'll fix this by setting `global.FOAMLINK_DATA` based on the environment variable `FOAMLINK_DATA` rather than hardcoding a path

KernelDeimos:
I fixed the genjava regression I mentioned earlier (foamlink enabled by default without build configuration). This PR will no longer affect any existing project using genjava until a path to the foamlink data file is specified with an environment variable.

Edit: oops, this is true after `c1e0a10`

src/foam/util/filesystem/doc.flow
KernelDeimos: This is inside `foam/util/filesystem` where I put the file walker, so I'll add a new flow doc in `foam/foamlink` with that description

KernelDeimos:
I'll remove `DO_NOT_MERGE` once I fix some ~stupid~ unexpected problem with relative vs absolute paths

KernelDeimos:
Finally got a chance to revisit this. I know what the problem is now and have a hack-fix for it, tomorrow I'll implement a proper fix and push some more code

KernelDeimos:
This has been updated @jlhughes @kgrgreer

I also did some more testing on the change. It looks like 70% build increase happens when modifying a Java file inside the `foam2` submodule of a large application using foam2, because only FOAM's Java source gets compiled.

I think this indicates that still too many Java files are being compiled when just one package with few dependencies is modified, which means it should be possible to get a similar speed boost for all cases but it will require some investigation.

src/foam/util/filesystem/doc.flow
KernelDeimos: I added earlier but in the wrong branch - should be there now

jlhughes (approved):

PR #2885, opened 2019-12-13
Fix infinite loop when opening DAOController detail view

lchanmann:
Traversing up the superclass chain instead.


TW80000 (approved):
Tested

PR #2886, opened 2019-12-13
Clone notification and args email settings

zahnno:
Had an issue where arguments being passed into email settings were updating due to logic in emailUtility (template argument being added to map). This problem was reflected on the notifications argument list as well as the resolved PropertyInfo which mapped back to the initial notification arguments. Cloning notification before resolving propertyinfo and setting the the cloned notification into the emailUtilities fixed the issue.

### DIFF
--- a/src/foam/nanos/notification/EmailSetting.js
+++ b/src/foam/nanos/notification/EmailSetting.js
@@ -21,6 +21,7 @@ foam.CLASS({
     'java.util.Iterator',
     'java.util.List',
     'java.util.Map',
+    'java.util.HashMap',
     'static foam.mlang.MLang.EQ'
   ],
 
@@ -40,7 +41,8 @@ foam.CLASS({
       javaCode: `
         Logger logger = (Logger) x.get("logger");
 
-        Iterator entries = arguments.entrySet().iterator();
+        Map<String, Object> args = new HashMap<String, Object>(arguments);
+        Iterator entries = args.entrySet().iterator();
         while (entries.hasNext()) {
           Map.Entry entry = (Map.Entry) entries.next();
           if ( entry.getValue() instanceof PropertyInfo ) {
@@ -52,7 +54,7 @@ foam.CLASS({
             entry.setValue(((PropertyInfo) entry.getValue()).get(user));
           }
         }
-        return arguments;
+        return args;
       `
     },
     {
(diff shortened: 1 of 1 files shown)


jlhughes:
@zahnno The resolution method should be notification wide, not just on EmailSettings. Is there a global place to put it? Or perhaps on the NotificationSettings base class - if we have one, where it clones self, resolves, and returns the copy.  Then extend classes can call super and do any extra Settings specific translation they need.

src/foam/nanos/notification/EmailSetting.js
```diff
@@ -63,19 +65,20 @@ foam.CLASS({
...
+            if ( clone.getEmailArgs() != null ) {
+              Map<String, Object> emailArgs = resolveNotificationArguments(x, clone.getEmailArgs(), user);
+              clone.setEmailArgs(emailArgs);
             }
 
             try {
-              if ( foam.util.SafetyUtil.isEmpty(notification.getEmailName()) ) {
-                message.setSubject(notification.getTemplate());
-                message.setBody(notification.getBody());
+              if ( foam.util.SafetyUtil.isEmpty(clone.getEmailName()) ) {
```
Nauna, line 76: sorry didn't catch this on previous reviews but,
this whole segment seems unnecessary:
 'if ( foam.util.SafetyUtil.isEmpty(clone.getEmailName()) '
This check is done in the EmailsUtility.
so passing clone.getEmailName() is fine no matter what.

The user, being passed in is for group usage. Don't see why you would want to pass in 'null'
Further if there is no template, then the args being passed in are ignored - or at least should be.

Nauna:
so, as I understand the email system is reusing the same map object and this is causing issues.

@zahnno in this PR you suggest the separation of the map in emailsSettings, 
and @jlhughes is suggesting a more global approach - as in the NotificationSettings base class.

But the issue starts with the usage of Model.PROPERTY(java-PropertyInfo) to dynamically set things on the emails?? 
If so then theoretically when you do the call to set these PROPERY's - which happens prior to sending the email ... what is the problem?
If false, @zahnno can you explain the actual problem a little more. Is the key value pairs getting confused?

src/foam/nanos/notification/EmailSetting.js
zahnno: To give some context on Notification Settings as an implementation. Users will be able to receive notifications based on what settings they own. As a developer I might create a notification and pass it into a users doNotify method. When that method is called it will attempt to send notifications based on all notification settings the user may have/opted into. For example (Email, SMS or Push).

Looks like the condition in EmailsUtility prevents this logic from happening if no email template is passed in. Which we'll have to allow.

zahnno:
@Nauna That was a little tough to decipher lol but yah, I think you understand what's happening. Just incase I'll reiterate through the logic here.

Notification objects being passed into the EmailSetting.sendNotification method may have a Map object set on its emailArgs property where one or many of its value reference a PropertyInfo object. 

In the sendNotification method we resolve the emailArgs by iterating through the key values of the map and replacing each value that reference a PropertyInfo with the users value related to the PropertyInfo.

The problem this PR sets out to fix is that updates made in EmailsUtility were mapping back to the Notification objects emailArgs which is used as a reference point of the original argument map (A Map prior to resolving to user values).

zahnno:
https://github.com/foam-framework/foam2/pull/2890 updated PR

PR #2887, opened 2019-12-16
DO_NOT_MERGE System Authorizer

noodlemoodle:
Added model for system authorizer, so that we can avoid repeating this check in every other Authorizer and instead use this in conjunction with other authorizers by using OrAuthorizer

SystemAuthorizer = Authorizer to be used for authorization on DAOs meant to be accessed by only the following entities: 
    1. System user
    2. Users of the "admin" group
    3. Users of the "system" group


kgrgreer:
I'm not sure this is a good idea. I would like to meet to discuss.

kgrgreer (changes requested):
I have concerns about how this is going to interact with ROPE. Lets meet to discuss.

PR #2893, opened 2019-12-17
Related to onboarding changes on SignUp and validation wordings

minsunny:
For nanopay side onboarding changes >>>> https://github.com/nanoPayinc/NANOPAY/pull/8550
this PR is needed to be merged.

1. added job title, removed department on SignUp view
2. Some wording changes on validations


src/foam/core/Validation.js
```diff
@@ -129,7 +129,7 @@ foam.CLASS({
             predicateFactory: function(e) {
               return e.GTE(foam.mlang.StringLength.create({ arg1: self }), self.minLength);
             },
-            errorString: `Please enter a ${this.label} with least ${this.minLength} character${this.minLength>1?'s':''}`
+            errorString: `Please enter ${this.label.toLowerCase()} with least ${this.minLength} character${this.minLength>1?'s':''}`
```
jlhughes:
```suggestion
            errorString: `Please enter ${this.label.toLowerCase()} with at least ${this.minLength} character${this.minLength>1?'s':''}`
```

src/foam/nanos/auth/Phone.js
```diff
@@ -22,7 +22,7 @@ foam.CLASS({
               foam.nanos.auth.Phone.NUMBER,
               /^(?:\+?1[-.●]?)?\(?([0-9]{3})\)?[-.●]?([0-9]{3})[-.●]?([0-9]{4})$/);
           },
-          errorString: 'Please enter a phone number'
+          errorString: 'Please enter phone number'
```
jlhughes, line 25: The above include 'valid' - which one is correct?   should they be the same.

src/foam/nanos/auth/User.js
```diff
@@ -180,6 +180,25 @@ foam.CLASS({
       updateMode: 'RO',
       section: 'personal'
     },
+   {
+      class: 'String',
+      name: 'jobTitle',
```
jlhughes, line 185: Really don't like this here, but have no other place for it at the moment. 
We are working through payroll, and HR app, and would like to move a number of properties off of user: department, organization, ....   to an 'employee' object of some type which is associated with a business and a user or with an organization if we build that out.     Anyway just future thoughts for now.

jlhughes (approved):

PR #2894, opened 2019-12-17
[CPF-3664] COUNTRY FIELD IN SIGNUP FIX

jinnnn9:
Fixes a bug where the css properties in themes are not applied to country field in sign up pages


jinnnn9:
@googlebot I signed it!

TW80000 (approved):

PR #2895, opened 2019-12-17
Add lastModifiedBy property to Script

KolombetM:
Add LastModifiedBy to Scripts to track who ran them last.

Unfortunately it works only wen script changed but not run.


jlhughes:
@Mykola97 this is actually working.  When writing to the journal we only output fields which have changed.  If you scan back in your journals/scripts you should find the first time you tested this and a lastModifiedBy the admin user 1348.  Thereafter this field did not change in your testing.

jlhughes (approved):

PR #2899, opened 2019-12-19
add HTML iframe support

KolombetM:
Added extra view class with css to remove margins


KolombetM:
Wrong author settings

PR #2900, opened 2019-12-19
encode hastags with %23

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3596
note: the `#` has special mean when it in the URI. `encodeURI` will not encode the # which will casuing missing the data after the `#`. When using encodeURIComponent it will break and failed to download file.

### DIFF
--- a/src/foam/comics/DAOController.js
+++ b/src/foam/comics/DAOController.js
@@ -266,7 +266,7 @@ foam.CLASS({
           result = 'data:text/csv;charset=utf-8,' + result;
           var encodedUri = encodeURI(result);
           var link = document.createElement('a');
-          link.setAttribute('href', encodedUri);
+          link.setAttribute('href', encodedUri.replace(/#/g, '%23'));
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

src/foam/comics/DAOController.js
```diff
@@ -266,7 +266,7 @@ foam.CLASS({
           result = 'data:text/csv;charset=utf-8,' + result;
           var encodedUri = encodeURI(result);
           var link = document.createElement('a');
-          link.setAttribute('href', encodedUri);
+          link.setAttribute('href', encodedUri.replace(/#/g, '%23'));
```
jlhughes, line 269: why not fix this in encodeURI?

yij793: encodeURI is a build-in function. I can fix this on line 267 with 
var encodedUri = encodeURI(result).replace(/#/g, '%23')

PR #2904, opened 2019-12-20
User create view improvements

TW80000:
### Main
* Hide a bunch of properties when creating a User
* Update client-side validation of a bunch of User properties
  * I used `validateObj` directly instead of `validationPredicates` because that way the back end doesn't do any validation checks. I don't want the back end to check validation because it breaks a large number of tests. I know this is bad, but we don't currently have a good solution for application-specific validation. I plan on thinking about this next, but consider it outside the scope of this PR.
* Made `PasswordView` the default view for `Password` properties
* Disable "Save" button in DAO Controller v2 if data is invalid
* Use properties supplied by `propertyWhitelist` directly so cloned properties with overrides are respected

### Other
* Formatting
* Hid a few props when creating Rules that don't make sense to set, such as `created` and `lastModified`

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -203,15 +203,8 @@ foam.CLASS({
       class: 'String',
       name: 'organization',
       documentation: 'The organization/business associated with the User.',
-      displayWidth: 80,
-      width: 100,
       tableWidth: 160,
-      // TODO: Use validatationPredicates instead.
-      validateObj: function(organization) {
-        if ( ! organization.trim() ) {
-          return 'Organization Required.';
-        }
-      },
+      required: true,
       section: 'business'
     },
     {
@@ -220,6 +213,7 @@ foam.CLASS({
       documentation: `The department associated with the organization/business
         of the User.`,
       width: 50,
+      createMode: 'HIDDEN',
       section: 'business'
     },
     {
@@ -266,6 +260,7 @@ foam.CLASS({
         return this.Phone.create();
       },
       view: { class: 'foam.u2.detail.VerticalDetailView' },
+      createMode: 'HIDDEN',
       section: 'personal'
     },
     {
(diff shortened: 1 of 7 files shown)


src/foam/nanos/auth/User.js
```diff
@@ -184,6 +184,11 @@ foam.CLASS({
       class: 'String',
       name: 'jobTitle',
       section: 'personal',
+      validateObj: function(jobTitle) {
```
kgrgreer: What if we set this property through java or DIG?

src/foam/nanos/auth/User.js
```diff
@@ -373,14 +380,17 @@ foam.CLASS({
...
       width: 100,
       storageTransient: true,
-      validateObj: function (password) {
-        var re = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9]).{7,32}$/;
-
-        if ( password.length > 0 && ! re.test(password) ) {
-          return 'Password must contain one lowercase letter, one uppercase letter, one digit, and be between 7 and 32 characters in length.';
+      // TODO: Validation that's consistent with the PasswordPolicy feature but
+      // also only applies on create.
+      validateObj: function(desiredPassword, passwordLastModified) {
```
kgrgreer: What about Java/Swift/DIG?

TW80000:
@kgrgreer please review again.

src/foam/nanos/auth/email/EmailVerificationWebAgent.java
```diff
@@ -62,7 +63,10 @@ public void execute(X x) {
       user = (User) user.fclone();
       emailToken.processToken(x, user, token);
     } catch (Throwable t) {
-      message = "Problem verifying your email.<br>" + t.getMessage();
+      String msg = t.getMessage();
+      ((Logger) x.get("logger")).error(msg);
```
jlhughes, line 67:
```suggestion
      ((Logger) x.get("logger")).error(msg, t);
```
wil generate a stack trace in the log file.

TW80000: @jlhughes do you want me to revert this then?

kgrgreer:
LGTM

PR #2905, opened 2019-12-20
move cmd_ and listen_ to easyDAO

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3258
Note: function cmd_ and listen_ will not get hit in `AbstractDAO` when making changes in DAO. Moving both function to RequestResponseClientDAO


TW80000:
To echo what I said in person, my suspicion is that there must be a bug in `ProxyDAO.listen_` or `ProxyDAO.cmd_` if this fixes the issue, since `EasyDAO` extends `ProxyDAO`, which extends `AbstractDAO`.

jlhughes:
I don't like the fix in EasyDAO.  Let's find the real problem.

jlhughes (approved):

kgrgreer:
> function cmd_ and listen_ will not get hit in AbstractDAO when making changes in DAO.
Why not?

It seems to me that this will only work in Client/Server mode but will break client-side only DAOs.

PR #2907, opened 2019-12-20
Performance

jlhughes:


src/foam/dao/AbstractFileJournal.js
```diff
@@ -58,28 +58,26 @@ foam.CLASS({
...
     }
   ],
 
   properties: [
+    {
+      class: 'Object',
+      name: 'line',
+      javaType: 'foam.util.concurrent.AssemblyLine',
+      javaFactory: 'return new foam.util.concurrent.SyncAssemblyLine();'
+//      javaFactory: 'return new foam.util.concurrent.AsyncAssemblyLine(getX());'
```
nanoscott: remove dead code (or explain why it's there)

src/services
```diff
@@ -525,7 +525,23 @@ p({
...
+  "class":"foam.nanos.boot.NSpec", 
+  "name":"ruleHistoryDAO",
+  "serve":true,  
+  "serviceScript":"""
+    return new foam.dao.EasyDAO.Builder(x)
+      .setPm(true)
+      .setSeqNo(true)
+      .setOf(foam.nanos.ruler.RuleHistory.getOwnClassInfo())
+      .setNullify(true)
+      //.setJournalType(foam.dao.JournalType.SINGLE_JOURNAL)
```
nanoscott: do we want to keep this in?

jlhughes: Yes, this needs to be undone, and kept.  
But also need to address it , as it is as expensive as the transactionDAO.

xuerongNanopay (approved):

PR #2909, opened 2019-12-21
Remove Session test of User.deleted as deleted refinement was added o…

jlhughes:
…utside of foam.


src/foam/nanos/session/Session.js
```diff
@@ -290,7 +290,7 @@ foam.CLASS({
       javaCode: `
         User user = (User) ((DAO) x.get("localUserDAO")).find(userId);
 
-        if ( user == null || user.getDeleted() ) {
+        if ( user == null ) {
```
lchanmann:
```suggestion
       if ( user == null
         || (user instanceof DeletedAware && ((DeletedAware)user).getDeleted())
       ) {
```

PR #2910, opened 2019-12-24
LastModifiedAware to check if object is modified

lchanmann:
Don't update lastModified/By when the object being put is exactly the same as the existing object already in the DAO.

### DIFF
--- a/src/foam/nanos/auth/LastModifiedByAwareDAO.js
+++ b/src/foam/nanos/auth/LastModifiedByAwareDAO.js
@@ -13,13 +13,18 @@ foam.CLASS({
     {
       name: 'put_',
       code: function(x, obj) {
-        if ( foam.nanos.auth.LastModifiedByAware.isInstance(obj) ) {
-          obj.lastModifiedBy = x.user.id;
+        if ( ! foam.nanos.auth.LastModifiedByAware.isInstance(obj) ) {
+          return this.delegate.put_(x, obj);
         }
-        return this.SUPER(x, obj);
+        return this.delegate.find_(x, obj).then(function(old) {
+          if ( ! obj.equals(old) ) {
+            obj.lastModifiedBy = x.user.id;
+          }
+          return this.delegate.put_(x, obj);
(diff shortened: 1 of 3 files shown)


src/foam/nanos/auth/LastModifiedByAwareDAO.js
```diff
@@ -13,13 +13,18 @@ foam.CLASS({
...
+        return this.delegate.find_(x, obj).then(function(old) {
+          if ( ! obj.equals(old) ) {
+            obj.lastModifiedBy = x.user.id;
+          }
+          return this.delegate.put_(x, obj);
+        }.bind(this));
       },
       javaCode: `
-        if ( obj instanceof LastModifiedByAware ) {
+        if ( obj instanceof LastModifiedByAware && ! obj.equals(getDelegate().find_(x, obj)) ) {
```
jlhughes, line 27: Thinking about performance and if find_(x, obj) is valid. 
find_(x, predicate) becomes dao.limit(1).select(x, ... predicate).
Alternative is   find(x, ((FObject) obj).getProperty("id")) 
@kgrgreer do we transform   find_(x, obj) to find_(x, getProperty("id") . ?   Couldn't find it searching. Did find that transform in Sink.js to test if entry already exists in the sink.

lchanmann: For MDAO, `find_` will get the primary key of `obj` https://github.com/foam-framework/foam2/blob/master/src/foam/dao/MDAO.java#L144, not going through dao.limit(1).select(x, ... predicate).

jlhughes (approved):
Please double check EasyDAO to ensure we install this very close to the MDAO - not before the 'Decorators' for example which could update the object.  Thanks.

src/foam/nanos/auth/LastModifiedByAwareDAO.js
kgrgreer: find(obj) is efficient. It doesn't do a select.

PR #2911, opened 2019-12-28
With try/finally ensure final synchronized block is called, which set…

jlhughes:
…s q_ to null when last/only job.

Both SyncAssemblyLine and AsyncAssemblyLine updated.  Sync is straightforward, but not sure about Async.  Both need review, but Async more so, and didn't have a test to exercise it.

### DIFF
--- a/src/foam/util/concurrent/AsyncAssemblyLine.java
+++ b/src/foam/util/concurrent/AsyncAssemblyLine.java
@@ -28,20 +28,16 @@ public class AsyncAssemblyLine
  public void enqueue(Assembly job) {
    final Assembly previous;
 
-   try {
-     synchronized ( startLock_ ) {
+   synchronized ( startLock_ ) {
+     try {
        previous = q_;
        q_ = job;
        job.executeUnderLock();
        job.startJob();
+     } catch (Throwable t) {
+       q_ = null;
+       throw t;
      }
-   } catch (Throwable t) {
-     // Isn't required, but helps GC last entry.
-     synchronized ( startLock_ ) {
-       // If I'm still the only job in the queue, then remove me
-       if ( q_ == job ) q_ = null;
-     }
-     throw t;
    }
 
    pool_.submit(x_, new ContextAgent() { public void execute(X x) {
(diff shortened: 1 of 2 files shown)


src/foam/util/concurrent/AsyncAssemblyLine.java
```diff
@@ -28,27 +28,39 @@ public AsyncAssemblyLine(X x) {
...
-     if ( previous != null ) previous.waitToComplete();
-
-     synchronized ( endLock_ ) {
-       job.endJob();
-       job.complete();
+   try {
+     synchronized ( startLock_ ) {
+       previous = q_;
+       q_ = job;
+       job.executeUnderLock();
```
jlhughes, line 35: @kgrgreer executeUnderLock previously wasn't in Async.  Was that intentional? 
Also looking at this now, if we try/catch here it can occur inside the sync and just set q_ = null on catch.

src/foam/util/concurrent/AsyncAssemblyLine.java
```diff
@@ -28,27 +28,39 @@ public AsyncAssemblyLine(X x) {
...
+       job.startJob();
      }
-
+   } catch (Throwable t) {
      // Isn't required, but helps GC last entry.
      synchronized ( startLock_ ) {
```
kgrgreer: I think this block should be moved outside of the catch, or put in and outside of the catch, or just removed altogether.

src/foam/util/concurrent/AsyncAssemblyLine.java
kgrgreer: Forgot to add.

src/foam/util/concurrent/AsyncAssemblyLine.java
```diff
@@ -29,25 +29,33 @@ public void enqueue(Assembly job) {
...
-     previous = q_;
-     q_ = job;
-     job.startJob();
+     try {
+       previous = q_;
+       q_ = job;
+       job.executeUnderLock();
+       job.startJob();
+     } catch (Throwable t) {
+       q_ = null;
```
kgrgreer, line 38: This looks like a bug. You should put q_ back to previous.

PR #2915, opened 2019-12-30
Add setting to hide the "Create" button in DAO Browser

TW80000:


src/foam/comics/v2/DAOControllerConfig.js
```diff
@@ -112,6 +112,12 @@ foam.CLASS({
         // for it which doesn't jive with the AbstractFObjectPropertyInfo.
         return { class: 'foam.u2.borders.NullBorder' };
       }
+    },
+    {
+      class: 'Boolean',
```
jlhughes, line 117: Why would we not make this permission based?

TW80000: We could, but not all DAOs/models use permissions for authorization.

jlhughes: Are we planning to copy all the comics/DAOController enableFoo properties?

jlhughes (approved):

src/foam/comics/v2/DAOControllerConfig.js
TW80000: We could. I wasn't planning on doing so until I had a concrete need though.

PR #2918, opened 2020-01-01
[CPF-3147] EasyDAO-TransactionDAO

jlhughes:


src/foam/nanos/ruler/predicate/PropertyEQValue.js
```diff
@@ -40,11 +40,12 @@ foam.CLASS({
     {
       name: 'f',
       javaCode: `
-        if ( getIsNew() ) {
+        FObject old = (FObject) OLD_OBJ.f(obj);
+        if ( getIsNew() ||
+             old == null ) {
```
lchanmann: when isNew=false, we want to compare the oldObj.prop to the value and if the oldObj is null should we just return false? @nanoMichal 

(similar thing for ropertyNEQValue.js and PropertyEQProperty.js)

src/foam/nanos/ruler/Rule.js
```diff
@@ -275,9 +275,8 @@
           try {
             return getPredicate().f(obj);
           } catch ( Throwable th ) { }
-
-          ((Logger) x.get("logger")).error(
-            "Failed to evaluate predicate of rule: " + getId(), t);
+          // ((Logger) x.get("logger")).debug(this.getClass().getSimpleName(), "id", getId(), "\\nrule", this, "\\nobj", obj, "\\nold", oldObj, "\\n", t);
```
lchanmann, line 278: do we want to leave the commented code here?

src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -57,21 +57,27 @@ public void setDelegate(DAO delegate) {
...
+        if ( ! isRuleApplicable(rule, obj, oldObj)) continue;
+        PM pm = (PM) x_.get("PM");
+        pm.setClassType(RulerDAO.getOwnClassInfo());
+        pm.setName(rule.getDaoKey() + ": " + rule.getId());
+        pm.init_();
+        applyRule(rule, obj, oldObj, agency);
+        pm.log(x_);
+        agency.submit(x_, x -> saveHistory(rule, obj), "Save history. Rule id:" + rule.getId());
+      } catch ( Exception e ) {
+        // logger.debug(this.getClass().getSimpleName(), "id", rule.getId(), "\\nrule", rule, "\\nobj", obj, "\\nold", oldObj, "\\n", e);
```
lchanmann, line 73: same as above

lchanmann (changes requested):

src/foam/nanos/ruler/Rule.js
jlhughes: yes. debugging rules is tricky, and keep enabling these to see what is going on.

src/foam/nanos/ruler/predicate/PropertyEQValue.js
jlhughes: without the check for old == null we have NPE because the old tests are blindly executed.

lchanmann: we can add the null check before the last return statement.
```
return old != null && EQ(old.getClassInfo().getAxiomByName ...
```

lchanmann (approved):

PR #2920, opened 2020-01-02
add scroll to in tabrow

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3762
Currently the tab is overlap with contents since there are too many tab in one row. removed the wrap and add scroll in css fix the issue
Screen Shot 2020-01-02 at 10 26 31 AM

### DIFF
--- a/src/foam/u2/DeleteModal.js
+++ b/src/foam/u2/DeleteModal.js
@@ -23,24 +23,6 @@ foam.CLASS({
       display: flex;
       justify-content: flex-end;
     }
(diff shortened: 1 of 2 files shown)


TW80000 (approved):

TW80000:
We still need this, in my opinion. The `Tabs` Element is still used outside of `TabbedDetailView` and thus will still be buggy if there are too many tabs. We need to fix it anyway, so I'd still like to merge this. Or change it to the behaviour where tabs wrap instead of having a single row that must be scrolled.

PR #2924, opened 2020-01-02
Avoid naming conflict between permissions

TW80000:
A group with "user.read.*" will be able to read all users, as intended,
but that permission also implies "user.read.lifecyclestate.deleted" (and
rejected), which is unintentional. We need to be able to give a group
access to read all users without letting them read those that have been
marked as deleted or rejected.


jlhughes (approved):
We really need to work through edge cases.

PR #2925, opened 2020-01-02
Don't hide date of birth field by default

TW80000:
We'll override it elsewhere instead, since other applications depend on
the default being not hidden.

FYI @minsunny


minsunny:
Thank you so much !

PR #2928, opened 2020-01-03
Add permissions, permission junctions and change test for create from…

jlhughes:
… create.id to create

### DIFF
--- a/src/foam/nanos/notification/NotificationSetting.js
+++ b/src/foam/nanos/notification/NotificationSetting.js
@@ -120,7 +120,7 @@ foam.CLASS({
       ],
       type: 'String',
       javaCode: `
-        return "notificationsetting." + operation + getId();
+        return "notificationsetting." + operation + "." + getId();
       `
     }
   ]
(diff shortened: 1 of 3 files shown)


src/foam/nanos/notification/NotificationSetting.js
```diff
@@ -120,7 +120,7 @@ foam.CLASS({
       ],
       type: 'String',
       javaCode: `
-        return "notificationsetting." + operation + "." + getId();
+        return "notificationsetting." + operation + getId();
```
TW80000: I think this was correct before. Without the dot we end up with permissions like "notificationsetting.read1496"

TW80000 (changes requested):

src/foam/nanos/notification/NotificationSetting.js
jlhughes:
```suggestion
        return "notificationsetting." + operation + "." + getId();
```

TW80000 (approved):

PR #2943, opened 2020-01-08
Wired up conditional proper feedback to CUD which use Approval system

tharmaman:
@jlhughes @kgrgreer @Nauna @TW80000

### DIFF
--- a/src/foam/comics/v2/DAOCreateView.js
+++ b/src/foam/comics/v2/DAOCreateView.js
@@ -49,7 +49,8 @@ foam.CLASS({
     'foam.u2.dialog.NotificationMessage'
   ],
   imports: [
-    'stack'
+    'stack',
+    'ctrl'
   ],
   exports: [
     'controllerMode'
@@ -78,6 +79,16 @@ foam.CLASS({
       }
     }
   ],
+  messages: [
+    {
+      name: 'SUCCESS_REQUEST_MESSAGE',
+      message: 'An approval request has been created.'
+    },
+    {
+      name: 'SUCCESS_CREATED_MESSAGE',
+      message: 'Successfully created.'
+    }
+  ],
   actions: [
     {
       name: 'save',
@@ -86,15 +97,22 @@ foam.CLASS({
       },
       code: function() {
         var cData = this.data;
+        var message = this.SUCCESS_CREATED_MESSAGE;
 
         if ( foam.nanos.auth.LifecycleAware.isInstance(cData) ) {
           cData = cData.clone();
           cData.lifecycleState = foam.nanos.auth.LifecycleState.PENDING;
+
+          // TODO: Make this check foam.nanos.auth.ApprovableAware.isInstance(cData) once ApprovableAware moves to foam
+          message = this.SUCCESS_REQUEST_MESSAGE;
         }
         
         this.config.dao.put(cData).then((o) => {
           this.data = o;
           this.finished.pub();
+          this.ctrl.add(this.NotificationMessage.create({
+            message: message
+          }));
           this.stack.back();
         }, (e) => {
           this.throwError.pub(e);
--- a/src/foam/comics/v2/DAOUpdateView.js
+++ b/src/foam/comics/v2/DAOUpdateView.js
@@ -96,6 +96,13 @@ foam.CLASS({
     }
   ],
 
+  messages: [
+    {
+      name: 'SUCCESS_REQUEST_MESSAGE',
+      message: 'An approval request has been created.'
+    }
+  ],
+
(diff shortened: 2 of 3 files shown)


src/foam/comics/v2/DAOCreateView.js
```diff
@@ -78,6 +79,16 @@ foam.CLASS({
       }
     }
   ],
+  messages: [
+    {
+      name: 'SUCCESS_REQUEST_MESSAGE',
+      message: 'An approval request has been created.'
```
jlhughes, line 85: 1. Why something so specific in a generic class? 
2. We can't have reference to Approvals in foam as they are on the nanopay side at the moment.

src/foam/comics/v2/DAOUpdateView.js
```diff
@@ -96,20 +96,34 @@ foam.CLASS({
     }
   ],
 
+  messages: [
+    {
+      name: 'SUCCESS_REQUEST_MESSAGE',
+      message: 'An approval request has been created.'
```
jlhughes, line 102: Similar.

jlhughes (changes requested):

src/foam/u2/DeleteModal.js
```diff
@@ -29,6 +29,7 @@ foam.CLASS({
     { name: 'TITLE', message: 'Delete ' },
     { name: 'CONFIRM_DELETE_1', message: 'Are you sure you want to delete' },
     { name: 'SUCCESS_MSG', message: ' deleted.' },
+    { name: 'SUCCESS_MSG_REQUEST', message: 'An approval request has been created.' },
```
jlhughes, line 32: What happens when approvals are not configured/enabled (via rules) - delete would be immediate would it not?

TW80000 (approved):
LGTM.

@jlhughes, did you see that they use the approval request message when the object is `LifecycleAware` and a normal message when it's not? Also the comments say we plan on checking `ApprovableAware` instead of `LifecycleAware` once it makes it into FOAM. `LifecycleAware` is already in FOAM, however.

src/foam/comics/v2/DAOCreateView.js
kgrgreer: I agree.

src/foam/comics/v2/DAOCreateView.js
```diff
@@ -86,15 +97,22 @@ foam.CLASS({
       },
       code: function() {
         var cData = this.data;
+        var message = this.SUCCESS_CREATED_MESSAGE;
 
         if ( foam.nanos.auth.LifecycleAware.isInstance(cData) ) {
           cData = cData.clone();
           cData.lifecycleState = foam.nanos.auth.LifecycleState.PENDING;
```
kgrgreer, line 104: What happens if the user creates the data through DIG instead? Then do they avoid going through lifecycle?

kgrgreer:
DNLGTM, let's discuss the whole design in person.

src/foam/comics/v2/DAOCreateView.js
tharmaman: The default lifecycleState property value for User is ACTIVE and so they should be able to skip through the approval process as I just accounted for it in the approvableAwareDAO decorator.

Otherwise if you wanted to change the default value to PENDING, we would have to make a DIG user create be able to set LifecycleState to ACTIVE before putting it into the DAO.

src/foam/comics/v2/DAOCreateView.js
tharmaman: Yeah so it's a kind of a problem because the majority of our CRUD views use the generic foam screens and so unfortunately we need a way to specify this feedback message otherwise the user will think that the object actually got created when it actually is in a pending state waiting to be approved. We can talk more tomorrow.

src/foam/u2/DeleteModal.js
tharmaman: Approvals are configured through the ApprovableAwareDAO which we said would eventually end up in the EasyDAO not through rules.

But if you see below, it would check if the object is LifecycleAware (approvableAware later on when we migrate) and if it is, it will show the approval request related message otherwise it would show the complete delete message. 

We can talk more tmrw but I have an idea about configuring the success messages in the DAOControllerConfig @kgrgreer .

src/foam/comics/v2/DAOCreateView.js
tharmaman: RE below:

> We can talk more tmrw but I have an idea about configuring the success messages in the DAOControllerConfig @kgrgreer .

PR #2944, opened 2020-01-08
asserting with less restrictions

Nauna:
https://github.com/nanoPayinc/NANOPAY/pull/8674#issuecomment-572074650

### DIFF
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -2176,9 +2176,6 @@ foam.CLASS({
       of: 'foam.u2.DisplayMode',
       name: 'createMode',
       documentation: 'The display mode for this property when the controller mode is CREATE.',
-      assertValue: function(newValue) {
-        foam.assert(newValue === foam.u2.DisplayMode.RW || newValue === foam.u2.DisplayMode.HIDDEN, `Create mode must be RW or HIDDEN. Property '${this.of ? this.of.id + '.' : ''}${this.name}' was '${newValue.label}' and must be fixed.`);
-      },
       value: 'RW'
     },
     {
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -2187,7 +2187,7 @@ foam.CLASS({
       name: 'readMode',
       documentation: 'The display mode for this property when the controller mode is VIEW.',
       assertValue: function(newValue) {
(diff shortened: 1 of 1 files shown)


src/foam/u2/Element.js
```diff
@@ -2177,7 +2177,7 @@ foam.CLASS({
       name: 'createMode',
       documentation: 'The display mode for this property when the controller mode is CREATE.',
       assertValue: function(newValue) {
-        foam.assert(newValue === foam.u2.DisplayMode.RW || newValue === foam.u2.DisplayMode.HIDDEN, `Create mode must be RW or HIDDEN. Property '${this.of ? this.of.id + '.' : ''}${this.name}' was '${newValue.label}' and must be fixed.`);
+        foam.assert(newValue === foam.u2.DisplayMode.RW || newValue === foam.u2.DisplayMode.HIDDEN || newValue === foam.u2.DisplayMode.RO, `Create mode must be RW or HIDDEN or RO. Property '${this.of ? this.of.id + '.' : ''}${this.name}' was '${newValue.label}' and must be fixed.`);
```
kgrgreer: Wouldn't it be simpler to just say that we don't support DISABLED?

kgrgreer: Also, add a comment explaining why DISABLED isn't supported for create mode.

Nauna: Ya that makes sense 🤔

TW80000: Just talked with Anna, and I think it makes sense to just remove the `assertValue` for `createMode`. I can see why RO could make sense if you want to show a dynamically updating field in CREATE mode based on another property, so that makes sense. I still can't think of a reason someone would want a DISABLED input in create mode, but I don't think there's a good reason for the framework to prevent anyone from doing that if they want to. I think the `assertValue` on `readMode` still makes sense though.

kgrgreer: I agree with @TW80000. Let's just remove.

src/foam/u2/Element.js
```diff
@@ -2187,7 +2187,7 @@ foam.CLASS({
       name: 'readMode',
       documentation: 'The display mode for this property when the controller mode is VIEW.',
       assertValue: function(newValue) {
-        foam.assert(newValue === foam.u2.DisplayMode.RO || newValue === foam.u2.DisplayMode.HIDDEN, `Read mode must be RO or HIDDEN. Property '${this.of ? this.of.id + '.' : ''}${this.name}' was '${newValue.label}' and must be fixed.`);
+        // no restrictions on DisplayMode
```
kgrgreer: Just remove the whole assertValue:.

PR #2945, opened 2020-01-08
Move all notification nspecs to package local services file

jlhughes:
Add localNotificationSettings nspec

### DIFF
--- a/src/foam/nanos/notification/services
+++ b/src/foam/nanos/notification/services
@@ -1,0 +1,89 @@
+p({
+  "class": "foam.nanos.boot.NSpec",
+  "name": "localNotificationDAO",
+  "lazy": false,
+  "serve": false,
+  "serviceScript": """
+    return new foam.dao.EasyDAO.Builder(x)
+      .setOf(foam.nanos.notification.Notification.getOwnClassInfo())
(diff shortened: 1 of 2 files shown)


TW80000:
If I recall correctly, @mcarcaso told me that doing this in foam breaks the foam build script/tool. Apparently it's only the nanopay build script that looks in subdirectories for now. I haven't tested this myself though, and I guess we could always just improve the foam build tool, but I thought I'd mention it.

kgrgreer:
Good to know, but I don't think anybody is building FOAM alone anyway, and
if that is broken, then we should fix it. Can you assign a low-priority
ticket to Eric to test/fix?

TW80000:
> Good to know, but I don't think anybody is building FOAM alone anyway, and if that is broken, then we should fix it. Can you assign a low-priority ticket to Eric to test/fix?

@kgrgreer will do.

TW80000:
https://nanopay.atlassian.net/browse/CPF-3954

mcarcaso:
Sorry I don’t really recall this but before I lost access to your repo you
were also using the JS build tool to bundle your app into a single payload.
If that still works then you should be good here. I can’t think of why this
change would break that off the top of my head but anything is possible.

PR #2950, opened 2020-01-08
Fix visual flickering when searching in RichChoiceView

TW80000:
It was flickering because we were re-rendering the sections every time a user adds or removes a character in the search field. Specifically, there was a small amount of time where the view was removed but the promises were still pending while the new contents were being looked up. When the promises resolve, the dropdown gets populated again. Since this happens quite quickly, there's a flickering effect where the dropdown briefly disappears and then reappears.

This PR changes that behaviour by only updating the content only when the promise resolves, which gets rid of the flickering.


jlhughes (approved):

Nauna:
tested and works. not only for his desired fix but also for a bug I encountered while using sections as a slot, and the search not working with the postSet on the filtered_ property.

thx TROY :D

PR #2954, opened 2020-01-09
Cpf 3657/add external url prop to app config

yij793:
address:  https://nanopay.atlassian.net/browse/CPF-3657
note: add new property to appConfig in order to add external Url. Also change loginView backLink_ since it should not redirect to staging.ablii.com

### DIFF
--- a/src/foam/u2/view/LoginView.js
+++ b/src/foam/u2/view/LoginView.js
@@ -191,7 +191,7 @@ foam.CLASS({
       class: 'String',
       name: 'backLink_',
       factory: function() {
-        return this.model.backLink_ ? this.model.backLink_ : this.appConfig.externalUrl || undefined;
+        return this.model.backLink_ || this.appConfig.externalUrl || undefined;
       },
       hidden: true
     }
--- a/src/foam/u2/view/LoginView.js
+++ b/src/foam/u2/view/LoginView.js
@@ -191,9 +191,7 @@ foam.CLASS({
       class: 'String',
       name: 'backLink_',
       factory: function() {
-        return !! this.model.backLink_ ?
-          this.model.backLink_ :
-          !! this.appConfig.externalUrl ? this.appConfig.externalUrl : undefined;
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/LoginView.js
```diff
@@ -193,7 +193,7 @@ foam.CLASS({
       factory: function() {
         return !! this.model.backLink_ ?
           this.model.backLink_ :
-          !! this.appConfig.url ? this.appConfig.url : undefined;
+          !! this.appConfig.externalUrl ? this.appConfig.externalUrl : undefined;
```
jlhughes: does this not work without the double negation !! ?  What bug are you protecting against with this?

yij793: > does this not work without the double negation !! ? What bug are you protecting against with this?

I'm not sure. I did not write those code. I only changed url to externalUrl. I can rewrite the logic if you suggest.

kgrgreer: You could also just do: this.appConfig.externalUrl || undefined;

jlhughes: re: rewrite - yes please.  Update to Kevin's recommendation.  Thanks.

jlhughes:
```suggestion
          this.appConfig.externalUrl : undefined;
```

src/foam/u2/view/LoginView.js
```diff
@@ -191,9 +191,7 @@ foam.CLASS({
       class: 'String',
       name: 'backLink_',
       factory: function() {
-        return !! this.model.backLink_ ?
-          this.model.backLink_ :
-          !! this.appConfig.url ? this.appConfig.url : undefined;
+        return this.model.backLink_ ? this.model.backLink_ : this.appConfig.externalUrl || undefined;
```
kgrgreer: Could be simplified as just:
this.model.backLink_ || this.appConfig.externalUrl || undefined;

PR #2956, opened 2020-01-10
fix logs not loading

nanoNeel:
https://nanopay.atlassian.net/browse/CPF-3928
credits to @lchanmann for finding and fixing the problem.

### DIFF
--- a/src/foam/dao/WriteOnlyFileJournal.js
+++ b/src/foam/dao/WriteOnlyFileJournal.js
@@ -9,9 +9,6 @@ foam.CLASS({
   name: 'WriteOnlyFileJournal',
   extends: 'foam.dao.FileJournal',
 
-  javaImports: [
-    'foam.core.FObject',
-  ],
   properties: [
     {
       name: 'outputClassNames',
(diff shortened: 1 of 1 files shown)


src/foam/dao/WriteOnlyFileJournal.js
```diff
@@ -32,8 +35,18 @@ foam.CLASS({
         final Object id = obj.getProperty("id");
 
         getLine().enqueue(new foam.util.concurrent.AbstractAssembly() {
+          FObject old;
           String record_ = null;
 
+          public Object[] requestLocks() {
+            return new Object[] { id };
```
kgrgreer: Why is this needed?

kgrgreer: requestLocks support isn't implemented yet, and if it were, this wouldn't work because you aren't guaranteed to get the exact same object every time for 'id', even if they do have the same value.

src/foam/dao/WriteOnlyFileJournal.js
```diff
@@ -32,8 +35,18 @@ foam.CLASS({
...
+          }
+
+          public void executeUnderLock() {
+            old = dao.find_(x, id);
```
jlhughes: 'old' is never used.

PR #2958, opened 2020-01-10
Server Side DAO Filtering

blakegreen32:
Currently, the DAO filtering loads the whole DAO on the client side in order to populate the filters. The client will not be able to handle this with any large DAO, so I have updated the logic to optimize it on the client and offload as much work (and memory) as possible to the server by calling GroupBy on the DAO select which in turn filters out all the duplicates and displays the necessary DAO content during filtration.

Ticket: https://nanopay.atlassian.net/browse/CPF-3792


nanoscott (approved):

src/foam/u2/filter/StringFilterView.js
```diff
@@ -113,30 +118,19 @@ foam.CLASS({
...
-        }, { temp: [], out: [] }).out;
-      }
+      name: 'daoContents'
     },
     {
       class: 'String',
-      name: 'search'
+      name: 'search',
+      postSet: function(_, n) {
+        this.dao.where(this.CONTAINS_IC(this.property, n)).limit(100).select(this.GroupBy.create({
```
kgrgreer, line 127: What is the limit(100) for? You know that this will not give you all values in most cases?

src/foam/u2/filter/StringFilterView.js
```diff
@@ -113,30 +118,19 @@ foam.CLASS({
...
+          arg1: this.property,
+          arg2: this.ArraySink.create()
```
kgrgreer: You don't actually use the return values? Just the keys? If so, this could just be a this.COUNT(), which would be faster and send less data across the network.

src/foam/u2/filter/StringFilterView.js
```diff
@@ -198,9 +184,13 @@ foam.CLASS({
 
   methods: [
     function initE() {
-      this.dao.select().then((results) => {
-        this.daoContents = results.array;
+      this.dao.limit(100).select(this.GroupBy.create({
```
kgrgreer, line 187: Same two comments as above.

src/foam/u2/filter/StringFilterView.js
nanoscott: This is supposed to be a set of possible values to filter by, so it seems reasonable to limit the set of values to filter by to a reasonable number of results that the user is likely to scroll through without causing performance problems on the client .. if they want more specific results, they should enter a search string to isolate on that result. I don't think an exhaustive list is the way to go here, as a set of unique values in a given column would just return the whole DAO and we'd be back in the same position as before.

blakegreen32: See Ticket Discussion Comments: https://nanopay.atlassian.net/browse/CPF-3792

kgrgreer: But the limit() is on the records in the DAO, not the resulting keys, so you might just get the same key 100 times.

kgrgreer: The MDAO can short-circuit GROUP_BY(COUNT)s.

PR #2961, opened 2020-01-11
User feedback

tharmaman:
Wired up UserFeedback and made it a storage transient property on AbstractFObject to temporarily store messages as an object on a put travels to the DAO and then back to the user.

Wired up UserFeedbackException to handle removes in cases where we don't want to return anything.

@kgrgreer @jlhughes @TW80000

### DIFF
--- a/src/foam/comics/v2/userfeedback/UserFeedback.js
+++ b/src/foam/comics/v2/userfeedback/UserFeedback.js
@@ -10,7 +10,7 @@ foam.CLASS({
   documentation: `
     A model to keep track of feedback messages as an object travels 
     through multiple decorators on the back-end and 
-    inevitably making its way back to the client who views the feedback
+    inevitably making its way back to the client who views the feedback.
   `,
 
   properties: [
--- a/src/foam/comics/v2/userfeedback/UserFeedback.js
+++ b/src/foam/comics/v2/userfeedback/UserFeedback.js
@@ -8,7 +8,7 @@ foam.CLASS({
   package: 'foam.comics.v2.userfeedback',
   name: 'UserFeedback',
   documentation: `
-    A model track of feedback messages as an object travels 
+    A model to keep track of feedback messages as an object travels 
     through multiple decorators on the back-end and 
     inevitably making its way back to the client who views the feedback
   `,
--- a/src/foam/comics/v2/userfeedback/UserFeedbackException.js
+++ b/src/foam/comics/v2/userfeedback/UserFeedbackException.js
@@ -12,7 +12,7 @@ foam.CLASS({
     In cases where the object is not returned to client user after a request,
     a UserFeedbackException will be thrown which will chain the message at the time
     of the exception to the other feedback messages the object has collected on its
-    path to the DAO
+    path to the DAO.
   `,
 
   properties: [
--- a/src/foam/comics/v2/userfeedback/UserFeedbackStatus.js
+++ b/src/foam/comics/v2/userfeedback/UserFeedbackStatus.js
@@ -9,7 +9,7 @@ foam.ENUM({
   name: 'UserFeedbackStatus',
   documentation: `
     Functional statuses which highlight different types of feedback
-    the client user is interested in
+    the client user is interested in.
   `,
 
   values: [
--- a/src/foam/core/AbstractFObject.js
+++ b/src/foam/core/AbstractFObject.js
@@ -96,6 +96,16 @@ foam.CLASS({
       name: 'userFeedback',
       class: 'FObjectProperty',
       of: 'foam.comics.v2.userfeedback.UserFeedback',
+      documentation: `
+        DAO decorators can attach messages to objects being being created,
+        updated or removed. These messages will be available on the client where
+        they can be displayed for the end user to read.
+        
+        Such feedback messages should be added to the FObject in question by
+        setting this property to an instance of the UserFeedback model. Note
+        that UserFeedback instances can be chained by setting the property named
+        'next'.
+      `
       storageTransient: true
     }
   ],


src/foam/comics/v2/userFeedback/UserFeedback.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.comics.v2.userFeedback',
```
kgrgreer: From: https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html
"Naming Conventions
Package names are written in all lower case to avoid conflict with the names of classes or interfaces."

src/foam/comics/v2/userFeedback/UserFeedbackException.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.comics.v2.userFeedback',
+  name: 'UserFeedbackException',
```
kgrgreer: add documentation:, to this model and others

src/foam/comics/v2/userFeedback/UserFeedback.js
```diff
@@ -0,0 +1,27 @@
...
+      of: 'foam.comics.v2.userFeedback.UserFeedbackStatus'
+    },
+    {
+      class: 'String',
+      name: 'message'
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.comics.v2.userFeedback.UserFeedback',
+      name: 'delegate'
```
kgrgreer: "delegate", or just "next"?

src/foam/comics/v2/userFeedback/UserFeedback.js
tharmaman: Done.

src/foam/comics/v2/userFeedback/UserFeedback.js
tharmaman: Changed to next

src/foam/comics/v2/userFeedback/UserFeedbackException.js
tharmaman: Done.

src/foam/comics/v2/userfeedback/UserFeedback.js
```diff
@@ -0,0 +1,32 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.comics.v2.userfeedback',
+  name: 'UserFeedback',
+  documentation: `
+    A model track of feedback messages as an object travels 
+    through multiple decorators on the back-end and 
+    inevitably making its way back to the client who view the feedback
```
kgrgreer: s/view/views/

src/foam/comics/v2/userfeedback/UserFeedbackException.js
```diff
@@ -0,0 +1,25 @@
...
+
+foam.CLASS({
+  package: 'foam.comics.v2.userfeedback',
+  name: 'UserFeedbackException',
+  implements: ['foam.core.Exception'],
+  documentation: `
+    In cases where the object is not returned back to client user after a request,
+    a UserFeedbackException will be thrown which will chain the message at the time
+    of the exception to the other feedback messages the object has collected on its
+    path to the dao
```
kgrgreer: s/dao/DAO/

kgrgreer: When is the object not returned?

src/foam/comics/v2/userfeedback/UserFeedbackException.js
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.comics.v2.userfeedback',
+  name: 'UserFeedbackException',
+  implements: ['foam.core.Exception'],
+  documentation: `
+    In cases where the object is not returned back to client user after a request,
```
kgrgreer: s/returned back/returned/

src/foam/core/AbstractFObject.js
```diff
@@ -91,6 +91,15 @@ foam.CLASS({
     }
   ],
 
+  properties: [
```
kgrgreer: Does this work? Does this property get included on the JS side? Shouldn't you add it to FObject.js instead?

src/foam/comics/v2/userfeedback/UserFeedbackException.js
tharmaman: In the case of remove, if the call is being intercepted by an approval request it will throw the exception.

But now that im thinking about it since removes do return an FObject, it could instead return the currentObj in the DAO with all the userFeedback from the initial obj that was passed through. Lemme know which one you think works better for remove however, since I remember you said its fine to return the object for put but for removes if I didn't remove it then it should technically be an error.

src/foam/comics/v2/userfeedback/UserFeedbackException.js
tharmaman: done.

src/foam/comics/v2/userfeedback/UserFeedback.js
tharmaman: Done.

src/foam/core/AbstractFObject.js
tharmaman: Tried it with both and also only on FObject.js, just adding the property doesn't work for some reason the console throws errors saying the setters and getters aren't found. We can talk about it more on Monday.

src/foam/comics/v2/userfeedback/UserFeedbackException.js
jlhughes: also on an intercepted remove - the exception allows/causes the cache to not remove - which is behaviour you want.

tharmaman: Oh yeah this was the original reason ^ so throwing an exception is the intended behaviour for removes

src/foam/comics/v2/userfeedback/UserFeedback.js
```diff
@@ -0,0 +1,32 @@
...
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.comics.v2.userfeedback',
+  name: 'UserFeedback',
+  documentation: `
+    A model track of feedback messages as an object travels
```
TW80000:
```suggestion
    A model to keep track of feedback messages as an object travels 
```

src/foam/comics/v2/userfeedback/UserFeedback.js
```diff
@@ -0,0 +1,32 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.comics.v2.userfeedback',
+  name: 'UserFeedback',
+  documentation: `
+    A model track of feedback messages as an object travels 
+    through multiple decorators on the back-end and 
+    inevitably making its way back to the client who views the feedback
```
TW80000:
```suggestion
    inevitably making its way back to the client who views the feedback.
```

src/foam/comics/v2/userfeedback/UserFeedbackStatus.js
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.ENUM({
+  package: 'foam.comics.v2.userfeedback',
+  name: 'UserFeedbackStatus',
+  documentation: `
+    Functional statuses which highlight different types of feedback
+    the client user is interested in
```
TW80000:
```suggestion
    the client user is interested in.
```

src/foam/comics/v2/userfeedback/UserFeedbackException.js
```diff
@@ -0,0 +1,25 @@
...
+
+foam.CLASS({
+  package: 'foam.comics.v2.userfeedback',
+  name: 'UserFeedbackException',
+  implements: ['foam.core.Exception'],
+  documentation: `
+    In cases where the object is not returned to client user after a request,
+    a UserFeedbackException will be thrown which will chain the message at the time
+    of the exception to the other feedback messages the object has collected on its
+    path to the DAO
```
TW80000:
```suggestion
    path to the DAO.
```

src/foam/comics/v2/userfeedback/UserFeedbackStatus.js
```diff
@@ -0,0 +1,20 @@
...
+  `,
+
+  values: [
+    { name: 'SUCCESS',   label: 'Success' },
```
TW80000: Should `INFO` be the default instead? (The first value is the default, so enum properties of this type will default to SUCCESS the way you have it currently)

src/foam/comics/v2/userfeedback/UserFeedbackStatus.js
```diff
@@ -0,0 +1,20 @@
...
+
+foam.ENUM({
+  package: 'foam.comics.v2.userfeedback',
+  name: 'UserFeedbackStatus',
+  documentation: `
+    Functional statuses which highlight different types of feedback
+    the client user is interested in
+  `,
+
+  values: [
```
TW80000, line 15: I think WARNING would be useful as well.

src/foam/core/AbstractFObject.js
```diff
@@ -91,6 +91,15 @@ foam.CLASS({
...
+    {
+      name: 'userFeedback',
+      class: 'FObjectProperty',
+      of: 'foam.comics.v2.userfeedback.UserFeedback',
```
TW80000, line 98:
```suggestion
      of: 'foam.comics.v2.userfeedback.UserFeedback',
      documentation: `
        DAO decorators can attach messages to objects being being created,
        updated or removed. These messages will be available on the client where
        they can be displayed for the end user to read.
        
        Such feedback messages should be added to the FObject in question by
        setting this property to an instance of the UserFeedback model. Note
        that UserFeedback instances can be chained by setting the property named
        'next'.
      `
```

TW80000 (commented):
Added a few documentation comments, but LGTM.

src/foam/core/AbstractFObject.js
kgrgreer: Add to FObject and include a javaType, otherwise it won't get generated on the Java side. You might have to add it as a refinement to FObject since type support may not yet be present when FObject is first loaded.

src/foam/comics/v2/userfeedback/UserFeedbackStatus.js
tharmaman: Done.

src/foam/comics/v2/userfeedback/UserFeedbackStatus.js
tharmaman: Done.

nanoscott:
@tharmaman .. i think you can close this, right?

PR #2965, opened 2020-01-13
Fixed tree search omg

tharmaman:
Basically the onSizeChangeComplete publish should fire off once the canvas finishes changing rather than anytime a change is done.

### DIFF
--- a/src/foam/graphics/TreeGraph.js
+++ b/src/foam/graphics/TreeGraph.js
@@ -51,7 +51,7 @@
    ],
 
    topics: [
-    'onSizeChange'
+    'onSizeChangeComplete'
    ],
 
    methods: [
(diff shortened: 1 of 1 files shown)


src/foam/graphics/TreeGraph.js
```diff
@@ -115,7 +115,8 @@
            this.invalidate();
            this.doLayout();
 
-           this.onSizeChange.pub();
+         } else {
+          this.onSizeChangeComplete.pub();
```
kgrgreer, line 119: Who is listening to this?

tharmaman: The AccountTreeView search feature on when to scroll. Basically the search feature only should scroll once the canvas has finished changing size and the nodes have been fully laid out which wasn't the case before, the scroll was triggering only when the canvas changes.

PR #2970, opened 2020-01-13
moving z-index to fix render

Nauna:
address: https://nanopay.atlassian.net/browse/CPF-3896

### DIFF
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -92,6 +92,9 @@ foam.CLASS({
     ^ {
       display: inline-block;
       position: relative;
+    }
+
+    ^setAbove {
       z-index: 1;
     }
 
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/RichChoiceView.js
```diff
@@ -374,7 +377,7 @@ foam.CLASS({
                   return Promise.all(promiseArray).then((resp) => {
                     var index = 0;
                     return this.E().forEach(sections, function(section) {
-                      this
+                      this.addClass('setAbove')
```
kgrgreer: I don't think this works. addClass 'setAbove' will set the class to just 'setAbove', but the css above is for ^setAbove which gets expanded to include the package and class name.

TW80000:
```suggestion
                      this.addClass(this.myClass('setAbove'))
```

TW80000 (changes requested):

src/foam/u2/view/RichChoiceView.js
```diff
@@ -374,7 +377,7 @@ foam.CLASS({
                   return Promise.all(promiseArray).then((resp) => {
                     var index = 0;
                     return this.E().forEach(sections, function(section) {
-                      this
+                      this.addClass(this.myClass('setAbove'))
```
TW80000:
```suggestion
                      this.addClass(self.myClass('setAbove'))
```

whoops, should probably be `self`.

TW80000 (approved):

PR #2971, opened 2020-01-13
TabChoiceView CSS Fix

kennyqiyenkan:
Modified line-height and padding css of TabChoiceView's label to display properly when the view does not have a lot of space.

### DIFF
--- a/src/foam/u2/view/TabChoiceView.js
+++ b/src/foam/u2/view/TabChoiceView.js
@@ -34,8 +34,8 @@ foam.CLASS({
 
     ^ label {
       cursor: pointer;
-      line-height: 48px;
-      padding: 0 32px;
+      line-height: 1.4px;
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/TabChoiceView.js
```diff
@@ -34,8 +34,8 @@ foam.CLASS({
 
     ^ label {
       cursor: pointer;
-      line-height: 48px;
-      padding: 0 32px;
+      line-height: 1.4px;
```
TW80000: This seems really small. Readable text isn't less than 1.4px tall, so are you depending on some side effect of a tiny line-height? How should I test this?

kennyqiyenkan: oops. thanks for the catch. there wasn't supposed to be a px.

PR #2972, opened 2020-01-13
Support outputting and replaying new class info id

lchanmann:
Liquid transaction approval flow saves generic transaction object to the journal and send out approval requests, after the transaction is approved it then goes through the normal transaction DAO where the transaction is quoted and the type is changed.

The journal entry should reflect that and be able set the correct type on replay.

### Screenshot
Screen Shot 2020-01-13 at 4 44 17 PM

### DIFF
--- a/src/foam/dao/AbstractFileJournal.js
+++ b/src/foam/dao/AbstractFileJournal.js
@@ -454,7 +454,7 @@ try {
           PropertyInfo prop = (PropertyInfo) e.next();
           mergeProperty(oldFObject, diffFObject, prop);
         }
-        return oldFObject;
+        return diffFObject.copyFrom(oldFObject);
       `
     },
     {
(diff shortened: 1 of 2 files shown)


src/foam/dao/AbstractFileJournal.js
```diff
@@ -454,7 +454,7 @@ try {
           PropertyInfo prop = (PropertyInfo) e.next();
           mergeProperty(oldFObject, diffFObject, prop);
         }
-        return oldFObject;
+        return diffFObject.copyFrom(oldFObject);
```
jlhughes, line 457: will this not overwrite the above mergeProperty?

lchanmann: No, the mergeProperty copies the new data to oldFObject then we want to return the diffFObject instead because it is the new type to write to the journal.

jlhughes: This would suggest we are not writing deltas to the journal, but that's not the case.

jlhughes: Ok, looking in FileJournal - yes this is correct. surprising we didn't catch this sooner.

jlhughes (approved):

nanoscott (approved):

PR #2973, opened 2020-01-14
[CPF-4026] Menu Authorizer and Predicate

jlhughes:
Manual tested by changing a Menu readPredicate to False and observing that menu no longer accessible from non-admin user which previously had menu access.

### DIFF
--- a/src/foam/nanos/menu/Menu.js
+++ b/src/foam/nanos/menu/Menu.js
@@ -91,7 +91,7 @@ foam.CLASS({
         class: 'foam.u2.view.JSONTextView'
       },
       javaFactory: `
-      return foam.mlang.MLang.TRUE;
+        return foam.mlang.MLang.TRUE;
       `,
     }
   ],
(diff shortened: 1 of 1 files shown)


src/foam/nanos/menu/Menu.js
```diff
@@ -70,13 +82,78 @@ foam.CLASS({
       name: 'icon',
       documentation: 'Icon associated to the menu item.',
       displayWidth: 80
+    },
+    {
+      documentation: 'Predicate providing arbitrary checks, in addition to the regular menu auth checks.',
+      class: 'FObjectProperty',
+      of: 'foam.mlang.predicate.Predicate',
```
lchanmann: We can use PredicateProperty

    class: 'foam.mlang.predicate.PredicateProperty',

src/foam/nanos/menu/Menu.js
```diff
@@ -70,13 +82,77 @@ foam.CLASS({
...
+      if ( ! auth.check(x, "menu.create") ) {
+        throw new AuthorizationException("You do not have permission to create menus.");
+      }
+      `
+    },
+    {
+      name: 'authorizeOnUpdate',
+      javaCode: `
+      AuthService auth = (AuthService) x.get("auth");
+      if ( ! auth.check(x, "menu.update."+getId()) ) {
```
TW80000:
```suggestion
      if ( ! auth.check(x, "menu.update." + getId()) ) {
```

src/foam/nanos/menu/Menu.js
```diff
@@ -70,13 +82,77 @@ foam.CLASS({
...
+        throw new AuthorizationException("You do not have permission to update this menu.");
+      }
+      `
+    },
+    {
+      name: 'authorizeOnDelete',
+      javaCode: `
+      AuthService auth = (AuthService) x.get("auth");
+      if ( ! auth.check(x, "menu.delete") ) {
```
TW80000: To stay consistent with the rest of the system, this should be "remove", not "delete" and should also include the id at the end.

```suggestion
      if ( ! auth.check(x, "menu.remove." + getId()) ) {
```

src/foam/nanos/menu/Menu.js
```diff
@@ -70,13 +82,77 @@ foam.CLASS({
...
+        throw new AuthorizationException("You do not have permission to delete menus.");
+      }
+      `
+    },
+    {
+      name: 'authorizeOnRead',
+      javaCode: `
+      AuthService auth = (AuthService) x.get("auth");
+      if ( ! ( auth.check(x, "menu.read."+getId()) &&
```
TW80000:
```suggestion
      if ( ! ( auth.check(x, "menu.read." + getId()) &&
```

src/foam/nanos/menu/Menu.js
```diff
@@ -70,13 +82,77 @@ foam.CLASS({
...
+      name: 'f',
+      type: 'Boolean',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        }
+      ],
+      javaCode: `
+      return getReadPredicate().f(
```
TW80000: [**Optional**]

Personally, I'd prefer that we indent javaCode one level beyond the `javaCode:` line to visually differentiate it from the property properties. Right now if I scan down the left side to see the property property keys, I see the following:

* documentation
* name
* type
* args
* javaCode
* return
* )

Obviously the last two don't belong because they're part of the value, not keys. Just my opinion though.

TW80000 (changes requested):
Only one important change, the remove permission.

src/foam/nanos/menu/Menu.js
jlhughes: I'm been the other way with indenting so that the resulting generated java looks correct, but as you prefer the javascript/java to be more readable.  We should add some standard/guide to our wiki

jlhughes: All formatting, spacing issues addressed.

src/foam/nanos/menu/Menu.js
```diff
@@ -70,13 +82,77 @@ foam.CLASS({
...
+        if ( ! auth.check(x, "menu.remove." + getId()) ) {
+          throw new AuthorizationException("You do not have permission to delete menus.");
+        }
+      `
+    },
+    {
+      name: 'authorizeOnRead',
+      javaCode: `
+        AuthService auth = (AuthService) x.get("auth");
+        if ( ! ( auth.check(x, "menu.read." + getId()) &&
```
TW80000: Potential optimization: if we switch the order so that it checks the predicate first, then we can short circuit and avoid the relatively expensive permission check when the predicate is false. Chances are that the predicate is almost always going to be cheaper than the permission check.

```Java
! ( false && x ) // always true, don't need to evaluate x
```

TW80000 (approved):
LGTM. Thought of one potential optimization you might consider.

src/foam/nanos/menu/Menu.js
jlhughes: Incorporated

PR #2974, opened 2020-01-14
CreatedByAgent

zahnno:
Objects implementing the createdByAwareDAO should consider acting as agents. Current implementation breaks previous logic where created by references user not agent.

Added CreatedByAgent properties to exisiting models implementing the createdByAwareDAO.


jlhughes (approved):

kgrgreer:
I'm okay with this, but we did talk about making the user/agent be something like a path.
So it could be somethig like: kgr/patrick, kgr as patrick, cron as system, cron as kgr as patrick.

PR #2975, opened 2020-01-14
Fix bug where SubSlot was detaching unnecessarily

TW80000:
SubSlots were detaching whenever the class of the object in the parent
slot changed. The reason for this, according to a comment in the code,
is that the new class will have "different SubSlots", which I assume
means different axioms.

An unfortunate side effect of this "feature" is that when the class of
the object in the parent slot changes to another class with the same
axiom, the SubSlot detaches anyway. It would be preferable, in my
opinion, if the SubSlot stays alive as long as the new class has the
same axiom the subslot was set up for. That's what this commit
implements.

### DIFF
--- a/src/foam/core/Slot.js
+++ b/src/foam/core/Slot.js
@@ -270,7 +270,6 @@ foam.CLASS({
       'For internal use only. Is used to implement the Slot.dot() method.',
 
   properties: [
-    'of',
     'parent', // parent slot, not parent object
     'name',
     'value',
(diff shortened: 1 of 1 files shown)


src/foam/core/Slot.js
```diff
@@ -334,10 +334,14 @@ foam.CLASS({
 
       // If the parent object changes class, then don't update
       // because a new class will have different sub-slots.
-      if ( o && ( this.of !== o.cls_.id || o.cls_.getAxiomByName(this.name) == null ) ) {
+      // If the new class has the same axiom as the old class, then we keep this
+      // SubSlot attached instead of detaching it.
+      if ( o && ( this.of !== o.cls_.id && o.cls_.getAxiomByName(this.name) == null ) ) {
```
kgrgreer: Is this part redundant: this.of !== o.cls_.id &&

TW80000: Resolved by 7c8235d

PR #2976, opened 2020-01-14
[DO_NOT_MERGE] Reorder NSpec order resolve startup failure - still don't know the root cause.

jlhughes:
When localLogMessageDAO is first in it's services file nspec loading fails with parse errors.  NOTE: it also is first in services.0 generation.   
When replaced by logLevelFilterLogger, the issue goes away.


jlhughes:
resolved.

PR #2977, opened 2020-01-14
Cpf 4032/setinner dao will overwirite inner dao java factory

yij793:
address: https://nanopay.atlassian.net/browse/CPF-4032
note: setInnerDAO in `localNotificationDAO`   will overwrite the `javaFactory` code due to 
```
  public foam.dao.DAO getInnerDAO() {
    if ( ! innerDAOIsSet_ ) {
      setInnerDAO(InnerDAOFactory_());
    }
    return innerDAO_;
  }
```
so the data will never gose in JDAO

@carl-zzz && @nanoNeel  find the issue
@xuerongCode found the way to fixed it

### DIFF
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -277,12 +277,8 @@ foam.CLASS({
         .setOf(getOf())
         .build();
       }
-      if ( getJournalType().equals(JournalType.SINGLE_JOURNAL) ) {
-        if ( getMDAO() != null )
-          return new foam.dao.java.JDAO(getX(), getMDAO(), getJournalName());
-        else
+      if ( getJournalType().equals(JournalType.SINGLE_JOURNAL) )
         return new foam.dao.java.JDAO(getX(), getOf(), getJournalName());
-      }
       return new foam.dao.MDAO(getOf());
       `
     },
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -270,11 +270,6 @@ foam.CLASS({
     {
       class: 'Object',
       type: 'foam.dao.DAO',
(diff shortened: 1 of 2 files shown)


src/foam/dao/EasyDAO.js
```diff
@@ -277,8 +282,12 @@ foam.CLASS({
         .setOf(getOf())
         .build();
       }
-      if ( getJournalType().equals(JournalType.SINGLE_JOURNAL) )
+      if ( getJournalType().equals(JournalType.SINGLE_JOURNAL) ) {
+        if ( getMDAO() != null )
+          return new foam.dao.java.JDAO(getX(), getMDAO(), getJournalName());
+        else
```
kgrgreer: No need for else if you've already returned in the if.

src/foam/nanos/notification/services
```diff
@@ -21,7 +21,7 @@ p({
         new foam.nanos.notification.SendSlackNotificationDAO(x)))).build())
       .setJournalType(foam.dao.JournalType.SINGLE_JOURNAL)
       .setJournalName("notifications")
-      .setInnerDAO(dao)
+      .setMDAO(dao)
```
jlhughes: Just replace the index setup with 
setIndex( new foam.core.PropertyInfo[] {
...
}
on the EasyDAO, and don't explicitly set the MDAO.

jlhughes (changes requested):

src/foam/nanos/notification/services
xuerongNanopay: I don’t think setIndex can set multiple index

jlhughes: yes it can.  it takes a PropertyInfo[]

jlhughes (approved):

PR #2978, opened 2020-01-14
Fixing LifecycleAwareDAO

nanoscott:
For some DAOs (e.g. ruleDAO), the base class in the DAO does not implement LifecycleAware .. but the inheriting classes do. This PR allows for find and select on the entities that do not implement LifecycleAware.


nanoscott:
@tharmaman .. some updates to LifecycleAwareDAO to handle rules and business rules.

PR #2979, opened 2020-01-14
Call super equals explicitly

KernelDeimos:


KernelDeimos:
oops hang on this is broken

KernelDeimos:
okay I'm gonna fix this and do a new PR

PR #2980, opened 2020-01-14
[CPF-4069] Call super equals explicitly

KernelDeimos:


src/foam/java/refinements.js
```diff
@@ -556,6 +556,14 @@ foam.LIB({
               return `get${foam.String.capitalize(p.name)}();`
             }).join('\n')
         });
+
+        cls.method({
+          name: 'equals',
+          type: 'boolean',
+          args: [ { name: 'o', type: 'Object' } ],
+          visibility: 'public',
+          body: `return super.equals(o);`
```
kgrgreer, line 565: I don't like this solution. Better to just re-insert super code here to be sure it's correct.

kgrgreer: This is just cheating.

KernelDeimos: You mean move the method out of AbstractFObject and put it in refinements? I would've done that but I wasn't sure if some models manually override equals

kgrgreer: No, just copy the code out of the super class and put it here in body:

KernelDeimos:
Closing this because it can be fixed better by moving `hashCode` instead

PR #2987, opened 2020-01-15
Add logging on async rule retry

lchanmann:
Log warning on exception before entering retry and log error when max retry has been reached.

### DIFF
--- a/src/foam/nanos/ruler/RetryManager.java
+++ b/src/foam/nanos/ruler/RetryManager.java
@@ -31,9 +31,7 @@ public class RetryManager {
         retry();
       } else {
         ((Logger) x_.get("logger")).error(
-          String.format("RetryManager(%s) max retry reached.", description_)
-          , exception_
-        );
+          "RetryManager(" + description_ + ") max retry reached.", exception_);
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/foam/nanos/ruler/RetryManager.java
```diff
@@ -39,6 +46,7 @@ public void run() {
               latch_.countDown();
             }
           } catch (Exception ex) {
+            exception_ = ex;
```
nanoscott, line 49: you should only keep the first exception .. so how about: 
```suggestion
            exception_ = (exception_ == null) ? ex : exception_;
```

jlhughes: yes, agree.

lchanmann: The first exception is already log in the RuleEngine class. This is the last exception when we stop retrying.

jlhughes: Then I'm ok with it.

jlhughes (approved):

src/foam/nanos/ruler/RetryManager.java
nanoscott: I'm ok with it too.

nanoscott (approved):

PR #2989, opened 2020-01-15
[DO_NOT_MERGE] Remove Login Using User Id And Add Login Using UserName

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/CPF-4063


src/foam/nanos/auth/PasswordExpiryAuthService.js
```diff
@@ -22,34 +22,39 @@ foam.CLASS({
...
-if ( isPasswordExpired(user) ) {
-  throw new AuthenticationException("Password expired");
-}
-return getDelegate().loginByEmail(x, email, password);`
+        `User user = (User) ((DAO) getLocalUserDAO())
+          .inX(x)
+          .find(
+            AND(
+              OR(
+                EQ(foam.nanos.auth.User.EMAIL, id.toLowerCase()),
```
TW80000: Pretty sure we don't need to include the package for `User` here and on lines 48 and 50.

src/foam/nanos/auth/PasswordExpiryAuthService.js
```diff
@@ -22,34 +22,39 @@ foam.CLASS({
...
+                EQ(foam.nanos.auth.User.USER_NAME, id)
+              ),
+              EQ(foam.nanos.auth.User.LOGIN_ENABLED, true)
```
TW80000: Should we be checking any of these properties on User as well?

* enabled (deprecated)
* emailVerified
* status
* deleted (will be deprecated soon)
* lifecycleState (nanopay side)

src/foam/nanos/u2/navigation/SignIn.js
```diff
@@ -98,27 +100,32 @@ foam.CLASS({
...
-        ).catch(
-          (err) => {
-            this.notify(err.message || 'There was a problem logging in.', 'error');
-        });
+          ).catch(
+            (err) => {
+              this.notify(err.message || 'There was a problem logging in.', 'error');
+          });
+        } else {
+          //TODO: change to 'Please enter email or username' when integratting
```
TW80000:
```suggestion
          //TODO: change to 'Please enter email or username' when integrating
```

and several places elsewhere. Not very important though.

src/foam/nanos/auth/PasswordExpiryAuthService.js
jlhughes: emailVerified: AND ( EQ(email, id), EQ(emailverified, true)
status: this is compliance - don't know if we block login on compliance issues - no way for them to correct anything - please inquire with Nadir. 
lifecycle: EQ (active)

nanopay-arthur: enabled: I will remove this.

emailVerified: We currently allow users with an unverified email to login to Ablii. They just get stuck at the "We sent you and email, please verify you email" page, so I don't think we should add the emailVerified condition. 

status: Checked in with Nadir and he said that we should prevent a user from logging in if their status is set to "Disabled" or "Revoked". Check already exists on NanopayUserAndGroupAuthService decorator.

lifecycle: don't add this. (doesn't exist on dev yet)

PR #2994, opened 2020-01-15
printStackTrace removed

olhabn:


src/foam/dao/jdbc/ConnectionPool.js
```diff
@@ -83,9 +83,7 @@ foam.CLASS({
   pool.setUrl(getConnectionString());
   pool.setInitialSize(getPoolSize());
   setPool(pool);
-} catch (Exception e) {
-  e.printStackTrace();
-}`
+} catch (Exception e) {}`
```
jlhughes, line 86: Can we not log these?  I don't like exceptions that are just eaten.

olhabn: Should I log it with logger.error() method or System.out.println?

PR #2995, opened 2020-01-15
Show property when readPermissionRequired=false

lchanmann:
If a property has `readPermissionRequired = false` then it should be shown in the detail view.

## Screenshot
- Before
Screen Shot 2020-01-15 at 5 34 07 PM

- After
Screen Shot 2020-01-15 at 5 34 42 PM

### DIFF
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -2257,10 +2257,12 @@ foam.CLASS({
 
           var propName = this.name.toLowerCase();
           var clsName  = this.forClass_.substring(this.forClass_.lastIndexOf('.') + 1).toLowerCase();
+          var canRead  = this.readPermissionRequired === false;
 
           return auth.check(null, `${clsName}.rw.${propName}`)
               .then(function(rw) {
                 if ( rw ) return Visibility.RW;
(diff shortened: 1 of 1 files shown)


src/foam/u2/Element.js
```diff
@@ -2257,10 +2257,12 @@ foam.CLASS({
 
           var propName = this.name.toLowerCase();
           var clsName  = this.forClass_.substring(this.forClass_.lastIndexOf('.') + 1).toLowerCase();
+          var canRead  = this.readPermissionRequired === false;
```
jlhughes, line 2260: Feel this should be !== true.   Thoughts?

TW80000 (approved):

src/foam/u2/Element.js
lchanmann: https://github.com/foam-framework/foam2/blob/master/src/foam/core/Property.js#L171-L175, I think it should be fine with === false.

jlhughes: Property.js
`  {
      class: 'Boolean',
      name: 'readPermissionRequired',
      value: false
    },`

jlhughes (approved):

PR #3000, opened 2020-01-16
Dao controller fixes plus feedback wiring TEMPORARY FIX

tharmaman:
As I talked with @jlhughes, for handling UserFeedbackExceptions for now we will just check if the string is about an approval request being sent out.


src/foam/u2/DeleteModal.js
```diff
@@ -86,10 +86,41 @@ foam.CLASS({
...
+
+          //     currentFeedback = currentFeedback.next;
+          //   }
+          // } else {
+          //   this.notify(err.message || this.FAIL_MSG, 'error');
+          // }
+
+          if ( err.message === "An approval request has been sent out."  ){
+            this.notify(err.message);
+
```
TW80000, line 118:
```suggestion
```

src/foam/comics/v2/DAOUpdateView.js
```diff
@@ -103,20 +103,59 @@ foam.CLASS({
...
       code: function() {
-        this.data.copyFrom(this.workingData);
-        this.config.dao.put(this.data).then(o => {
+        this.config.dao.put(this.workingData).then(o => {
           this.data = o;
           this.finished.pub();
+
+          if ( foam.comics.v2.userfeedback.UserFeedbackAware.isInstance(o) && o.userFeedback ){
+            var currentFeedback = o.userFeedback;
+            while ( currentFeedback ){
```
TW80000, line 112: The current behaviour of `NotificationMessage`s is to stack on top of each other if I'm not mistaken, so you'll likely only see the last message in the chain as a result of this. We should fix that separately, just wanted to point that out in case you didn't know.

PR #3003, opened 2020-01-16
[CPF-4032] Notification nspec - Replace setIndex with addPropertyIndex

jlhughes:
presently not compiling.


nanoscott (approved):
Ah, that's how you do it properly!

PR #3010, opened 2020-01-16
Improve efficiency by using entrySet over keySet

KernelDeimos:
This is more efficient because the map gets queried once per key rather than twice.

The change turned out to be much easier than I expected because it exists in code generation.

### DIFF
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -1538,7 +1538,7 @@ foam.CLASS({
                 if ( prop.get(obj) == null ) prop.set(obj, prop.of().newInstance());
                 entry.getValue().set((foam.core.FObject) prop.get(obj), str);
               } catch ( Throwable t ) {
-                // ???
+                t.printStackTrace() // cannot use logging from logging. 
               }
             }
           });
--- a/src/foam/java/refinements.js
+++ b/src/foam/java/refinements.js
@@ -1531,12 +1531,12 @@ foam.CLASS({
           p.fromCSVLabelMapping(map2);
         });
 
-        for ( String key : map2.keySet() ) {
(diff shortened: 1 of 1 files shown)


src/foam/java/refinements.js
```diff
@@ -1531,12 +1531,12 @@ foam.CLASS({
...
-          map.put(getName() + "." + key, new foam.lib.csv.FromCSVSetter() {
+        for ( java.util.Map.Entry<String, foam.lib.csv.FromCSVSetter> entry : map2.entrySet() ) {
+          map.put(getName() + "." + entry.getKey(), new foam.lib.csv.FromCSVSetter() {
             public void set(foam.core.FObject obj, String str) {
               try {
                 if ( prop.get(obj) == null ) prop.set(obj, prop.of().newInstance());
-                map2.get(key).set((foam.core.FObject) prop.get(obj), str);
+                entry.getValue().set((foam.core.FObject) prop.get(obj), str);
               } catch ( Throwable t ) {
                 // ???
```
jlhughes: It would be nice to log this 'debug' if possible.

KernelDeimos: Oh wow, my eyes skipped right over that thing. I can see why someone put question marks there though - PropertyInfo isn't ContextAware, but it does have access to ClassInfo but that's not ContextAware either... maybe conditionally log if the class has a Property axiom `x` of type `Context`?

jlhughes: Well think only a System.err makes sense, as this code will be called during log message creation.

jlhughes:
```suggestion
                t.printStackTrace() // cannot use logging from logging. 
```

jlhughes (approved):

PR #3012, opened 2020-01-16
create PURGE cmd_ to clear the cache

yij793:
address: https://nanopay.atlassian.net/browse/CPF-3754
issue : login with unverified account or 2FA enabled account, refresh the page, then login.
guessing: cache block getting data from menuDAO
fix: remove cacheType in menuDAO
Screen Shot 2020-01-16 at 1 58 23 PM

### DIFF
--- a/src/foam/nanos/u2/navigation/SignIn.js
+++ b/src/foam/nanos/u2/navigation/SignIn.js
@@ -86,7 +86,7 @@ foam.CLASS({
               class: 'foam.nanos.auth.ResendVerificationEmail'
             });
           } else {
-            this.menuDAO.cmd_(X, foam.dao.CachingDAO);
+            this.menuDAO.cmd_(X, foam.dao.CachingDAO.PURGE);
             window.location.hash = '';
             this.loginSuccess = !! this.user;
           }
--- a/src/foam/nanos/u2/navigation/SignIn.js
+++ b/src/foam/nanos/u2/navigation/SignIn.js
@@ -10,7 +10,8 @@ foam.CLASS({
     'loginSuccess',
     'notify',
     'stack',
-    'user'
+    'user',
+    'menuDAO'
   ],
 
   messages: [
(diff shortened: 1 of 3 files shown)


src/services
```diff
@@ -343,7 +343,7 @@ p({
       .setOrder(new foam.core.PropertyInfo[] {foam.nanos.menu.Menu.ORDER, foam.nanos.menu.Menu.LABEL})
       .build();
   """,
-  "client": "{\"of\":\"foam.nanos.menu.Menu\",\"cacheType\":\"FULL\"}"
+  "client": "{\"of\":\"foam.nanos.menu.Menu\"}"
```
jlhughes: Can you try this with TTL.  The timeout/delay is 15s.

yij793: Changed `\"cacheType\":\"FULL\"` to `\"cacheType\":\"TTL\"`. issue still present

jlhughes:
Solution that needs some investigation.  On successful login we want to send a 'purge' cache CMD on the menuDAO which the cachingDAO will handle.   This is all client side. 
Need to a PURGE constance to the CachingDAO and then send this on the cmd() method.  both need to be implemented. 
* add a constants: [ 'PURGE'], 
* implement cmd_(x, obj)  - test if obj equals PURGE, true clear the cache.  else pass it on. 
Then login can call dao.cmd(CachingDAO.PURGE)

src/foam/nanos/u2/navigation/SignIn.js
```diff
@@ -85,6 +86,7 @@ foam.CLASS({
               class: 'foam.nanos.auth.ResendVerificationEmail'
             });
           } else {
+            this.menuDAO.cmd_(X, foam.dao.CachingDAO);
```
jlhughes:
```suggestion
            this.menuDAO.cmd_(X, foam.dao.CachingDAO.PURGE);
```

PR #3018, opened 2020-01-16
Error fixed

olhabn:


jlhughes:
Resolved already. should use x, rather than getX().

PR #3024, opened 2020-01-17
Correct syntax

jlhughes:


src/foam/dao/TTLCachingDAO.js
```diff
@@ -24,7 +24,8 @@ foam.CLASS({
   properties: [
     {
       /** The cache to read items quickly. */
-      name: 'cache'
+      name: 'cache',
+      value: {}
```
kgrgreer, line 28: This will share the same cache between instances. I'll fix.

PR #3027, opened 2020-01-17
Fix bug related to cloning actions

TW80000:
FOAM currently doesn't handle cloning WeakMaps. When it tries to do so,
it ends up creating a POJO instead, which breaks downstream code. As a
workaround, we revert to using a normal object for runningMap instead of
a WeakMap.

@tharmaman

### DIFF
--- a/src/foam/core/Action.js
+++ b/src/foam/core/Action.js
@@ -123,15 +123,13 @@ If empty than no permissions are required.`
 If empty than no permissions are required.`,
     },
     {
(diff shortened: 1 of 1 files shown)


kgrgreer:
I reject this code. Alternative approach suggested to Troy.

kgrgreer (changes requested):
Put back to a WeakMap and define the property clone property to just return a new WeakMap. Avoids creating a memory leak.

PR #3030, opened 2020-01-17
Don't show submenu dropdown now that side nav handles it

TW80000:


jlhughes (approved):
great, was thinking the same.

zahnno:
I was investigating this as well. https://github.com/nanoPayinc/NANOPAY/pull/8860 consider this change considering it changes the behaviour for our platform. @jlhughes @kgrgreer @TW80000

src/foam/nanos/menu/SubMenu.js
```diff
@@ -17,8 +17,7 @@ foam.CLASS({
     },
 
     function launch(X, menu, parent) {
-      var view = this.createView(X, menu, parent);
-      view.open();
+      // noop
```
kgrgreer, line 20: Will this break the old style menus? Do we plan on supporting both styles?

TW80000: Yes, and I don't know. Probably safer to go with @zahnno's fix.

TW80000:
Closing in favour of https://github.com/nanoPayinc/NANOPAY/pull/8860.

PR #3031, opened 2020-01-17
Make ChoiceView use default value when user selects placeholder

TW80000:
Otherwise the user doesn't have any way of explicitly choosing the
default value.

The important part is that it's an _explicit_ action taken by the user.
When we were setting `this.data` to `undefined` before, we were telling
the data to go back to an "untouched" state. This has the undesirable
side effect of copyFrom ignoring that property since it's in an
untouched state. Instead, we want to explicitly set that property to a
default value (such as 0 for a long property, or "" for a string
property) so that copyFrom will copy that default value over.

### DIFF
--- a/src/foam/u2/view/ChoiceView.js
+++ b/src/foam/u2/view/ChoiceView.js
@@ -60,7 +60,7 @@ foam.CLASS({
 
         try {
           if ( ! n && this.placeholder ) {
-            this.data  = undefined;
+            this.data  = this.defaultValue !== undefined ? this.defaultValue : undefined;
             this.text  = this.placeholder;
             this.index = -1;
           } else {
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/ChoiceView.js
```diff
@@ -60,7 +60,7 @@ foam.CLASS({
 
         try {
           if ( ! n && this.placeholder ) {
-            this.data  = undefined;
+            this.data  = this.defaultValue !== undefined ? this.defaultValue : undefined;
```
TW80000: Ha, I guess this could just be 
```suggestion
            this.data  = this.defaultValue;
```

kgrgreer: How is this different than this.defaultValue?

PR #3033, opened 2020-01-18
dead code

hchoura:


src/foam/lib/json/UnknownReferenceParser.java
```diff
@@ -17,7 +17,7 @@ public UnknownReferenceParser() {
 
       public PStream parse(PStream ps, ParserContext x) {
         ps = ps.apply(delegate, x);
-        if ( ps == null && ! ps.valid() ) {
+        if ( ps == null ) {
```
jlhughes, line 20:
```suggestion
        if ( ps == null || ! ps.valid() ) {
```

jlhughes:
Replaced by https://github.com/foam-framework/foam2/pull/3035

PR #3036, opened 2020-01-18
delete dead code

hchoura:


src/foam/mlang/mlang.js
```diff
@@ -893,7 +893,6 @@ return stmt.toString();`
         `java.util.List<Predicate> args = new java.util.ArrayList<>();
 boolean update = false;
 True TRUE = new True();
-False FALSE = new False();
```
kgrgreer: I think we can remove the previous line also, since the TRUE constant is already defined on line 3344.

hchoura: but, this will be generated in Or.java that don't access to at that scoop.

PR #3043, opened 2020-01-20
[Do Not Merge] Catch AuthorizationException

nanoscott:
We don't want the authorization exception propogated to the FrontEnd, so I'm catching it and returning null.


src/foam/nanos/auth/AuthorizationDAO.js
```diff
@@ -91,8 +91,14 @@ foam.CLASS({
...
 
     FObject obj = super.find_(x, id);
-    if ( obj != null ) getAuthorizer().authorizeOnRead(x, obj);
-    return obj;
+    try 
+    {
+      if ( obj != null ) 
+        getAuthorizer().authorizeOnRead(x, obj);
+      return obj;
+    } catch (Exception e) {
```
TW80000:
```suggestion
    } catch (AuthorizationException e) {
```

src/foam/nanos/auth/AuthorizationDAO.js
```diff
@@ -91,8 +91,14 @@ foam.CLASS({
     if ( getAuthorizer().checkGlobalRead(x) ) return super.find_(x, id);
 
     FObject obj = super.find_(x, id);
-    if ( obj != null ) getAuthorizer().authorizeOnRead(x, obj);
-    return obj;
+    try 
+    {
```
TW80000: This should at the end of the line above.

TW80000 (approved):
I agree, this should have been the behaviour all along. I have a feeling this PR is likely to break some existing code though, so we should be careful when merging.

nanoscott:
I'm getting 2 new unit test failures, so marking as DNM for now.

image

jlhughes (approved):

PR #3051, opened 2020-01-20
Aligned primary-destructive ActionView

blakegreen32:
Added margin to align primary-destructive ActionView.

Ticket: https://nanopay.atlassian.net/browse/CPF-4221

### DIFF
--- a/src/foam/u2/ActionView.js
+++ b/src/foam/u2/ActionView.js
@@ -99,6 +99,7 @@ foam.CLASS({
       border-color: /*%DESTRUCTIVE3%*/ #d9170e;
       background-color: /*%DESTRUCTIVE3%*/ #d9170e;
       color: white;
+      margin-top: 12px;
     }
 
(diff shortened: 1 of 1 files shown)


TW80000 (commented):
I disagree with this. The ticket is referring to a specific place where the button is not aligned. The fix shouldn't be to add a margin to all primary-destructive buttons because doing so makes the margins inconsistent among all buttons, which is likely to cause a number of other alignment issues.

Instead, whatever specific view has an issue should add CSS to align its child elements. We should update the CSS of the child element to fix one case since the side effects of that are far-reaching.

In general, my philosophy regarding CSS for layout is that individual elements/components (like buttons) should almost never include any padding or margin. Instead, it should be up to the parent element to decide how to arrange its child elements by setting their margin, width, and other layout-related properties.

So, since the ticket was regarding MapView, I suggest that we update the CSS in MapView instead.

PR #3066, opened 2020-01-21
txt change - more appropriate msg

Nauna:
addresses: https://nanopay.atlassian.net/browse/CPF-4233

at least in part


src/foam/comics/v2/DAOBrowseControllerView.js
```diff
@@ -57,7 +57,7 @@ foam.CLASS({
   `,
 
   messages: [
-    { name: 'REFRESH_MSG', message: 'Table Refreshed' }
+    { name: 'REFRESH_MSG', message: 'Table Refreshing. Please be patient, as some networks might take longer.' }
```
nanoscott:
```suggestion
    { name: 'REFRESH_MSG', message: 'Refresh Requested. Please be patient, refresh may take up to a minute.' }
```
Not sure if we want to add a specific period of time though.

PR #3067, opened 2020-01-21
Some of the try-catch blocks were changed

olhabn:
Change this "try" to a try-with-resources

### DIFF
--- a/src/foam/blob/Blob.js
+++ b/src/foam/blob/Blob.js
@@ -715,7 +715,7 @@ this.setup(x);
 long size = blob.getSize();
 File tmp = allocateTmp(x, 1);
 
-try(HashingOutputStream os = new HashingOutputStream(new FileOutputStream(tmp))) {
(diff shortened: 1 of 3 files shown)


src/foam/blob/Blob.js
```diff
@@ -712,12 +712,10 @@ return file;`
...
 this.setup(x);
-HashingOutputStream os = null;
+long size = blob.getSize();
+File tmp = allocateTmp(x, 1);
 
-try {
-  long size = blob.getSize();
-  File tmp = allocateTmp(x, 1);
-  os = new HashingOutputStream(new FileOutputStream(tmp));
+try(HashingOutputStream os = new HashingOutputStream(new FileOutputStream(tmp))) {
```
kgrgreer: Please follow the FOAM style guidelines:
https://github.com/foam-framework/foam/wiki/FOAM2-Style-Guidelines

olhabn: Fixed

PR #3068, opened 2020-01-21
Updates to ApplicationLogoView

TW80000:
* Stop loading both large and normal logos but only showing one of them. Instead, only load the one we're going to display.
* Stop showing the app name beside the logo.

FYI @zahnno


zahnno:
You're removing the versatility of the ApplicationLogoView with this implementation.

TW80000:
> You're removing the versatility of the ApplicationLogoView with this implementation.

@zahnno it'll still swap out the logos dynamically if either of them change or if `largeLogoEnabled` changes. Is that what you mean?

PR #3071, opened 2020-01-21
Make table refresh button less prominent

TW80000:
It's not destructive, so it shouldn't be styled as destructive, and it's
not the primary action of the DAO controller, so it shouldn't occupy the
same space as the create button. This commit moves it alongside the
export button as a secondary action. I think it's more appropriate there
since it's available for any table and is always a secondary action,
just like exporting.

## Before
Screen Shot 2020-01-21 at 3 23 45 PM

## After
Screen Shot 2020-01-21 at 3 23 11 PM

FYI: @Nauna


Nauna:
said it on slack - but just for the record - very pretty.

My one concern is that users won't know what it does.

PR #3080, opened 2020-01-22
Fix bug where tableCellFormatter overrides were ignored

TW80000:
The issue was that the axiom that UnstyledTableView was taking the axiom
from was the axiom defined on `of`, which is likely the base class. But
if one of our objects in the table is a subclass of `of` that overrides
`tableCellFormatter` for one of the table columns, then the
`tableCellFormatter` was being ignored because we were looking up the
axiom on the base class and using its `tableCellFormatter`.

This commit fixes that by changing the way we specify which columns to
use. The old way looks like this:

```
columns: [
  'foo',
  my.model.path.MyModel.BAR.clone().copyFrom({
    baz: 'woozle'
  }),
  foam.core.Property.create({ /* ... */ })
]
```

There are three possibilities:

1. A string. This specifies the name of the axiom to use.
2. An axiom. You can also specify an axiom directly and override it if
you want by cloning it and modifying the clone.
3. A new axiom. A variant of #2, but this axiom doesn't actually exist
on the model at all, we're just making one up.

The table view was converting the strings into the axioms by looking
them up on `of`, so we end up with `columns_` being an array of axioms.

The new representation is to make `columns_` an array of tuples instead,
where the first item in the tuple is either an axiom name or case #3, a
made up property, and the other (second) item in the tuple is an object
of overrides, or null if no overrides are desired. Existing code still
works out of the box except for case #2. The new way looks like this:

```
columns: [
  'foo', // same
  ['bar', { baz: 'woozle' }],
  foam.core.Property.create({ /* ... */ }) // same
]
```

By using this representation, we can look up the appropriate axiom on the
subclass using the axiom name for cases #1 and #2, and for #3 we just
use that axiom directly and don't perform a lookup.

FYI: @nanoMichal

### DIFF
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -59,21 +59,20 @@ foam.CLASS({
       name: 'columns_',
       expression: function(columns, of, allColumns, editColumnsEnabled) {
         if ( ! of ) return [];
-        columns = columns.map(c => foam.String.isInstance(c) ? this.of.getAxiomByName(c) : c);
         if ( ! editColumnsEnabled ) return columns;
 
         // Reorder allColumns to respect the order of columns first followed by
         // the order of allColumns.
         allColumns = columns.concat(allColumns);
         allColumns = allColumns.filter((c, i) => {
-          return allColumns.findIndex(a => a.name == c.name) == i;
+          return allColumns.findIndex(a => a[0] == c[0]) == i;
         });
 
         return allColumns.filter(c => {
-          var v = this.ColumnConfig.create({ of: of, axiom : c }).visibility;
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -199,6 +203,7 @@ foam.CLASS({
       var columnSelectionE;
 
       if ( this.filteredTableColumns$ ) {
+        // TODO
```
kgrgreer, line 206: TODO what?

TW80000: Oh shoot, that was to remind myself to make sure this didn't break the feature where you can adjust the table columns. I'll check that.

TW80000: Fixed here: https://github.com/foam-framework/foam2/pull/3110

PR #3081, opened 2020-01-23
Add TTLSelectCachingDAO to EasyDAO

lchanmann:


src/foam/dao/CacheType.js
```diff
@@ -24,6 +24,10 @@ foam.ENUM({
     {
       name: 'TTL',
       documentation: "Using a TTL (Time to Live) cache"
+    },
```
kgrgreer: I think we should change the approach, since you might want both TTL and TTL_SELECT. Let's discuss.

lchanmann: Updated according to our discussion - https://github.com/foam-framework/foam2/pull/3081/commits/b374a9001e495232aa53ba72fdd142c4d0eb9111.

nanoscott (approved):
I think this will break the build for liquidPOC because of the removal of CacheType, so a follow-on PR will need to be created before we can update to FOAM master.

PR #3083, opened 2020-01-23
[CPF-4327] Add Cache-Control header to ResourceImageServlet

KernelDeimos:
~This doesn't fix anything, but the header probably should be there at least.~ Edit: this works fine but most of us have "Disable cache" turned on all the time

### DIFF
--- a/src/foam/nanos/servlet/ResourceImageServlet.java
+++ b/src/foam/nanos/servlet/ResourceImageServlet.java
@@ -49,7 +49,6 @@ public class ResourceImageServlet
       String ext = EXTS.get(request.substring(request.lastIndexOf('.') +1));
       try ( BufferedInputStream bis = new BufferedInputStream(is) ) {
         resp.setContentType(!SafetyUtil.isEmpty(ext) ? ext : DEFAULT_EXT);
-        // resp.setHeader("Content-Disposition", "filename=\"" + StringEscapeUtils.escapeHtml4(request) + "\"");
         resp.setHeader("Cache-Control", "public, max-age=3600"); // cache for 1 hour
 
         IOUtils.copy(bis, resp.getOutputStream());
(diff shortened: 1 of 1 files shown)


src/foam/nanos/servlet/ResourceImageServlet.java
```diff
@@ -49,7 +49,8 @@ protected void service(HttpServletRequest req, HttpServletResponse resp)
       String ext = EXTS.get(request.substring(request.lastIndexOf('.') +1));
       try ( BufferedInputStream bis = new BufferedInputStream(is) ) {
         resp.setContentType(!SafetyUtil.isEmpty(ext) ? ext : DEFAULT_EXT);
-        resp.setHeader("Content-Disposition", "filename=\"" + StringEscapeUtils.escapeHtml4(request) + "\"");
+        // resp.setHeader("Content-Disposition", "filename=\"" + StringEscapeUtils.escapeHtml4(request) + "\"");
```
jlhughes: Why did you remove this, not that I remember what is was for.

KernelDeimos: Oh, I forgot to mention that in my PR description. As far as I can tell this header is incorrect so I decided to comment it out while I was testing the OverlayActionListView icons. According to the MDN page the Content-Disposition header should begin with `inline` or `attachment` and filename should be just a name for the file if it's downloaded (right now it's a path)

kgrgreer: Yes, the header format is incorrect, but also, I don't think it makes sense. Look at the commit for when this was introduced to see why the developer added it. If no good reason is presented, then remove in a separate PR which explains why.

jlhughes: It was there from the inception of the class, so simply originally wrong .
@KernelDeimos just remove the line.

kgrgreer: @KernelDeimos Please remove this line and I'll merge.

PR #3084, opened 2020-01-23
[CPF-4327] Fix repeat icon requests with ClientResourceCache

KernelDeimos:
This fixes repeat requests for icon images from OverlayActionListView. Because of [`oops, no reason was found`] adding a Cache-Control header to ResourceImageServlet doesn't prevent the browser from downloading hundreds of copies of the same icon. To fix this issue, a ClientResourceCache was added that stores the icon on the client as a base64 string.

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -183,6 +183,7 @@ FOAM_FILES([
   { name: "foam/u2/Visibility"},
   { name: "foam/u2/RowFormatter" },
   { name: "foam/u2/WeakMap", flags: ['web'] },
+  { name: "foam/u2/ClientResourceCache", flags: ['js'] },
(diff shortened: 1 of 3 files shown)


jlhughes (approved):

kgrgreer:
I would prefer a more general server-side solution.

KernelDeimos:
It turns out I was wrong about #3083 not working; testing browser caching works much better after unchecking the "Disable cache" option :/

This PR is convenient for developers because it still reloads images when the page is reloaded (and the cache control PR only works for JAR builds). That makes me wonder if should should do both or have ClientResourceCache only enable for developers =

PR #3085, opened 2020-01-23
commenting out exports for demo - test and reimplement later

Nauna:
https://nanopay.atlassian.net/browse/CPF-4315


kgrgreer:
Can't we just fix the z-index so solve the actual problem rather than working around the problem by disabling export?

Nauna:
> Can't we just fix the z-index so solve the actual problem rather than working around the problem by disabling export?

The z-index is fixed. But it was asked of me to remove for liquid - since the export functionality was not and will not be tested comprehensively.

kgrgreer:
Who asked?

PR #3086, opened 2020-01-24
Fix bug where the table refresh button wouldn't work sometimes

TW80000:
The issue was that it was expecting `this.data` to be set to a DAO, but really the DAO it uses comes from the DAOControllerConfig.

### DIFF
--- a/src/foam/comics/v2/DAOBrowserView.js
+++ b/src/foam/comics/v2/DAOBrowserView.js
@@ -191,7 +191,7 @@ foam.CLASS({
       toolTip: 'Refresh Table',
       icon: 'images/refresh-icon-black.svg',
       code: function(X) {
-        this.data.cmd_(X, foam.dao.CachingDAO.PURGE);
+        this.config.dao.cmd_(X, foam.dao.CachingDAO.PURGE);
         this.add(foam.u2.dialog.NotificationMessage.create({
           message: this.REFRESH_MSG
         }));


Nauna:
tbh I didn't see a difference in the network call made before and after your change. That being said what you're implying does have merit.

Although I have to wonder since,
    {
      name: 'data'
    },

do we really want to limit this to ```this.config.dao```

perhaps a factory on data? like as a default ... just thoughts... I would not block this pr with this comment.

TW80000:
> tbh I didn't see a difference in the network call made before and after your change.

This change doesn't affect that. It fixes an issue where a console error is thrown if `this.data` is undefined. Sorry, I probably should have made that clearer in the description.

> do we really want to limit this to `this.config.dao`

The DAO that's actually used for the table is a filtered version of `this.config.dao`, see https://github.com/foam-framework/foam2/blob/master/src/foam/comics/v2/DAOBrowserView.js#L267 and https://github.com/foam-framework/foam2/blob/master/src/foam/comics/v2/DAOBrowserView.js#L159-L164. So I think it makes sense to use that same dao when purging the cache.

Nauna:
> > tbh I didn't see a difference in the network call made before and after your change.
> 
> This change doesn't affect that. It fixes an issue where a console error is thrown if `this.data` is undefined. Sorry, I probably should have made that clearer in the description.
> 
> > do we really want to limit this to `this.config.dao`
> 
> The DAO that's actually used for the table is a filtered version of `this.config.dao`, see https://github.com/foam-framework/foam2/blob/master/src/foam/comics/v2/DAOBrowserView.js#L267 and https://github.com/foam-framework/foam2/blob/master/src/foam/comics/v2/DAOBrowserView.js#L159-L164. So I think it makes sense to use that same dao when purging the cache.

Yup yup, ok I see ... config is built around data and config.dao is what we are using.
If I could I would approve! lol

PR #3088, opened 2020-01-24
add placeholder to RichChoiceView

hchoura:


Nauna:
Instead of adding a placeholder property on RichChoiceView what do you think about reusing the choosePlaceholder property. It's a more consistent msg then having the dashes as default:

image

But otherwise, nicely done. Code functions well.

src/foam/u2/view/RichChoiceView.js
```diff
@@ -298,7 +298,7 @@ foam.CLASS({
       documentation: 'Replaces choose from placeholder with passed in string.',
       expression: function(of) {
         var plural = of.model_.plural.toLowerCase();
-        return this.CHOOSE_FROM + plural + '...';
+        return this.CHOOSE_FROM + plural + ' ...';
```
TW80000, line 301: I disagree with this change.

src/foam/u2/view/RichChoiceView.js
```diff
@@ -343,13 +349,15 @@ foam.CLASS({
...
                     self.isOpen_ = ! self.isOpen_;
+                    this.placeholder = self.placeholder;
                   }
                 })
                 .start()
                   .addClass(this.myClass('custom-selection-view'))
-                  .add(this.slot((data) => {
-                    return this.E().tag(self.selectionView, {
+                  .add(this.slot((data, isOpen_) => {
+                   return this.E().tag(self.selectionView, {
```
TW80000:
```suggestion
                    return this.E().tag(self.selectionView, {
```

src/foam/u2/view/RichChoiceView.js
```diff
@@ -343,13 +349,15 @@ foam.CLASS({
...
                       data: data,
+                      defaultSelectionPrompt: self.placeholder,
```
TW80000: We're already binding `defaultSelectionPrompt` two lines below.

src/foam/u2/view/RichChoiceView.js
```diff
@@ -392,7 +400,19 @@ foam.CLASS({
                         .start().hide(!! section.hideIfEmpty && resp[index].value <= 0 || ! section.heading)
                           .addClass(self.myClass('heading'))
                           .add(section.heading)
-                        .end()
+                          .end()
```
TW80000: Fix indentation. This and the code you added below should be indented one level less than it is currently.

src/foam/u2/view/RichChoiceView.js
```diff
@@ -343,13 +349,15 @@ foam.CLASS({
                 .on('click', function() {
                   if ( self.mode === foam.u2.DisplayMode.RW ) {
                     self.isOpen_ = ! self.isOpen_;
+                    this.placeholder = self.placeholder;
```
TW80000: What does this do?

src/foam/u2/view/RichChoiceView.js
```diff
@@ -392,7 +400,19 @@ foam.CLASS({
...
+                          .end()
+                          .start()
+                            .addClass(self.myClass('selection-view'))
+                            .add('--------------------------')
+                                .callIf(! section.disabled, function() {
+                                  this.on('click', () => {
+                                    self.fullObject_ = undefined;
+                                    self.data = 0;
+                                    self.isOpen_ = false;
+                                    self.cross=true;
```
TW80000: Please remove.

TW80000 (changes requested):
Rather than add an option to the first section in the dropdown, I think we should either add a button in the dropdown that shows above the first section or include a button to clear in the button part itself.

src/foam/u2/view/RichChoiceView.js
hchoura: we need to add space between word and Ellipsis.

src/foam/u2/view/RichChoiceView.js
hchoura: we need to close the close the option list

TW80000:
Closing since #3090 was merged.

PR #3089, opened 2020-01-24
Close the RichChoiceView dropdown when user clicks away

TW80000:
This makes it behave more like native dropdowns. It also fixes an issue where two open `RichChoiceView`s would overlap if they were close enough to each other and one was above the other.

### DIFF
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -328,6 +328,26 @@ foam.CLASS({
       this.onDetach(this.data$.sub(this.onDataUpdate));
       this.onDataUpdate();
 
+      // Set up an event listener on the window so we can close the dropdown
+      // when the user clicks somewhere else.
+      var containerElement;
+      const fn = function(evt) {
+        var selfRect = (document.getElementById(self.id)).getClientRects()[0];
+        var containerRect = (document.getElementById(containerElement.id)).getClientRects()[0];
+        if (
+          ! (
+              evt.clientX >= selfRect.x &&
+              evt.clientX <= selfRect.x + selfRect.width &&
+              evt.clientY >= selfRect.y &&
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/RichChoiceView.js
```diff
@@ -328,6 +328,26 @@ foam.CLASS({
       this.onDetach(this.data$.sub(this.onDataUpdate));
       this.onDataUpdate();
 
+      // Set up an event listener on the window so we can close the dropdown
+      // when the user clicks somewhere else.
+      var containerElement;
+      const fn = function(evt) {
+        var selfRect = (document.getElementById(self.id)).getClientRects()[0];
```
kgrgreer: This is the same as this.el().getClientRects()[0]

src/foam/u2/view/RichChoiceView.js
```diff
@@ -328,6 +328,26 @@ foam.CLASS({
...
+        var containerRect = (document.getElementById(containerElement.id)).getClientRects()[0];
```
kgrgreer: see above

src/foam/u2/view/RichChoiceView.js
```diff
@@ -328,6 +328,26 @@ foam.CLASS({
...
+              evt.clientX >= selfRect.x &&
+              evt.clientX <= selfRect.x + selfRect.width &&
+              evt.clientY >= selfRect.y &&
+              evt.clientY <= containerRect.y + containerRect.height
+            )
+        ) {
+          self.isOpen_ = false;
+        }
+      };
+      window.addEventListener('click', fn);
```
kgrgreer: Importing 'window' rather than using the global 'window' allows your app/view to work with multiple windows.

TW80000:
@kgrgreer, all comments addressed, thanks for the review.

kgrgreer:
Very nice.

PR #3093, opened 2020-01-27
refine EMail regex and clean up email property in User.js

jinnnn9:


kgrgreer:
Please add notes to describe the reasoning for this PR. Do we no longer need to trim() emails?

kgrgreer:
Removes email trimming. Was that intentional? Can it be put back?

jinnnn9:
EMail class already had validation so the validation for email property in User.js seemed unnecessary which was the reason why I removed validateObj property . If we want trimming, we can use preSet property in place of validateObj property to trim the email.

src/foam/nanos/auth/User.js
```diff
@@ -218,24 +218,12 @@ foam.CLASS({
       displayWidth: 80,
       width: 100,
       preSet: function(_, val) {
-        return val.toLowerCase();
+        return val.toLowerCase().trim();
```
kgrgreer: Isn't it better if we add this to 'foam.core.EMail' instead, so that it affects all email addresses, and not just this one instance?

kgrgreer: Which appears in types.js

close001: Yeah I agree. I made a change accordingly and pushed it. Also, I changed the class of email property in signUp.js from String to EMail so that it does not need to use preSet anymore.

jinnnn9: Yeah I agree. I made a change accordingly and pushed it. Also, I changed the class of email property in signUp.js from String to EMail so that it does not need to use preSet anymore.

src/foam/core/Validation.js
```diff
@@ -331,7 +331,7 @@ foam.CLASS({
           {
             args: [this.name],
             predicateFactory: function(e) {
-              return e.REG_EXP(self, /^$|.+@.+\..+/);
+              return e.REG_EXP(self, /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/);
```
jlhughes, line 334: Why the test for @###.###.### ?

src/foam/core/types.js
```diff
@@ -431,7 +431,15 @@ foam.CLASS({
   name: 'EMail',
   extends: 'String',
   // FUTURE: verify
-  label: 'Email address'
+  label: 'Email address',
+  properties: [
+    [
+      'adapt',
+      function(_, v) {
+        return v.toLowerCase().trim();
```
TW80000, line 439: I'm pretty sure we (actually Kate) added `trim` as a Property property on `foam.core.String` a while back. So you might just be able to do something like this:

```JavaScript
{
  name: 'EMail',
  /* ... */
  trim: true
}
```

src/foam/core/Validation.js
jinnnn9: How about `/^[A-Za-z0-9._%+-]+@[a-zA-Z0-9]+\.[a-zA-Z]{2,6}$/` for the email regex?

jinnnn9:
https://github.com/foam-framework/foam2/pull/3104

PR #3096, opened 2020-01-27
Fix refresh button for some tables

TW80000:
I'm not exactly why, but the refresh button for tables wasn't working
for a certain DAO. Adding this command fixes it.


jlhughes (commented):
Yes, I remembering we couldn't decide where best to place this.

jlhughes (approved):

PR #3104, opened 2020-01-29
change email regex

jinnnn9:
- fix email regex so it does not break build

### DIFF
--- a/src/foam/nanos/u2/navigation/SignIn.js
+++ b/src/foam/nanos/u2/navigation/SignIn.js
@@ -27,7 +27,7 @@ foam.CLASS({
       hidden: true
     },
     {
-      class: 'EMail',
+      class: 'String',
       name: 'email',
       view: {
         class: 'foam.u2.TextField',
--- a/src/foam/nanos/u2/navigation/SignIn.js
+++ b/src/foam/nanos/u2/navigation/SignIn.js
@@ -27,19 +27,17 @@ foam.CLASS({
       hidden: true
     },
     {
-      class: 'String',
+      class: 'EMail',
       name: 'email',
       view: {
         class: 'foam.u2.TextField',
         focused: true
       },
-      preSet: function(o, n) {
-        return n.trim();
-      },
       visibilityExpression: function(disableEmail_) {
         return disableEmail_ ?
           foam.u2.Visibility.DISABLED : foam.u2.Visibility.RW;
-      }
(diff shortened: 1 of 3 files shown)


src/foam/nanos/u2/navigation/SignIn.js
```diff
@@ -27,7 +27,7 @@ foam.CLASS({
       hidden: true
     },
     {
-      class: 'EMail',
+      class: 'String',
```
jlhughes: Why the change back to String?

jinnnn9: I remembered someone had mentioned not wanting to show validation when users sign up. If we set class to EMail, it will display validation text so I reverted it back to how it had been.

TW80000: Could you leave it as Email but override the view for it in the sign in form instead?

jinnnn9: I think this is the reason why someone who wrote this code set class to String in the first place.

kgrgreer: Yes, why?

jinnnn9: I think you wouldn't usually have validation text display on login page (I think it was the writer's intention). @TW80000 I will try your approach.

src/foam/nanos/u2/navigation/SignIn.js
```diff
@@ -33,13 +33,11 @@ foam.CLASS({
...
       },
-      preSet: function(o, n) {
-        return n.trim();
-      },
       visibilityExpression: function(disableEmail_) {
         return disableEmail_ ?
           foam.u2.Visibility.DISABLED : foam.u2.Visibility.RW;
-      }
+      },
+      validationTextVisible: false
```
jlhughes, line 40: So this applies to sign-up as well?

jinnnn9: I haven't looked into it. I will change signUp as well if you want me to. I can also check any other places that use EMail to make sure they work well.

jlhughes: got it, thought signup and signin were the same model. So this is fine.  Thanks.

jlhughes (approved):

PR #3105, opened 2020-01-29
dao summary view back label

noodlemoodle:
in the dao summary view of obj A, if you click and go to some referenced object B, the back button says 'All {classB}' when it actually goes back to daosummary view of objA

### DIFF
--- a/src/foam/comics/v2/DAOSummaryView.js
+++ b/src/foam/comics/v2/DAOSummaryView.js
@@ -94,6 +94,10 @@ foam.CLASS({
       expression: function() {
         return foam.u2.detail.SectionedDetailView;
       }
+    },
+    {
+      class: 'String',
+      name: 'backLabel'
     }
   ],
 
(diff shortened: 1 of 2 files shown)


src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -169,6 +173,8 @@ foam.CLASS({
             });
           }
 
+          var label = self.backLabel || `All ${config$browseTitle}`;
```
TW80000, line 176: It would be foamier to make `backLabel`an expression of `config$browseTitle`.

```JavaScript
    {
      class: 'String',
      name: 'backLabel',
      expression: function(config$browseTitle) {
        return 'All ' + config$browseTitle;
      }
    }
```

noodlemoodle: thanks @TW80000 addressed in https://github.com/foam-framework/foam2/pull/3106

PR #3109, opened 2020-01-30
[CPF-4382] Add debug property

KolombetM:
https://nanopay.atlassian.net/browse/CPF-4382
"do not generate rule history for rules not chosen" and "support control over generating rule history during rule execution" were already implemented

### DIFF
--- a/src/foam/nanos/ruler/Rule.js
+++ b/src/foam/nanos/ruler/Rule.js
@@ -173,6 +173,11 @@
       }
     },
     {
+      class: 'Boolean',
+      name: 'debug',
+      documentation: 'Determines if logger uses debug'
+    },
+    {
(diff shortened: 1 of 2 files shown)


src/foam/nanos/ruler/Rule.js
```diff
@@ -172,6 +172,11 @@
         return nu || this.validity > 0;
       }
     },
+    {
+      class: 'Boolean',
+      name: 'debug',
+      documentation: 'Determines if logger uses debug'
```
jlhughes:
```suggestion
      documentation: 'Test this boolean before generating expensive logger.debug calls, to speed rule execution.'
```

jlhughes (approved):

PR #3111, opened 2020-01-30
Log hostname to support centralized log collection - NOC.

jlhughes:
Also replace using Builder for LogMessage creation with explicit constructor


TW80000 (approved):
LGTM

PR #3115, opened 2020-01-31
user created with only country

Nauna:
The one thing we are consistently collecting on User.address is the country.

So for display, changing the Address property to be more reflective of the information that is present.

The toolTip will show the address fields present.

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -326,8 +326,8 @@ foam.CLASS({
       tableCellFormatter: function(value, obj, axiom) {
         let addressString = '';
         for ( prop in value.instance_ ) if ( prop ) addressString += ` ${value[prop]}`;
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/User.js
```diff
@@ -322,6 +322,15 @@ foam.CLASS({
       factory: function() {
         return this.Address.create();
       },
+      label: 'Country',
+      tableCellFormatter: function(value, obj, axiom) {
+        let addressString = '';
+        for ( prop in value.instance_ ) if ( prop ) addressString += ` ${value[prop]}`;
+        return this.__subContext__.countryDAO.find(value.countryId).then((cobj) => {
+          if ( addressString ) this.setAttribute('title', addressString);
```
kgrgreer: This line doesn't need to be in the callback. It could be above with the 'for'. Also, you could do addressString.trim() to get rid of the extra space at the beginning.

PR #3120, opened 2020-02-03
wordBreak long names to fit on pg

Nauna:
address: https://nanopay.atlassian.net/browse/CPF-4313

also some syntax doesn't follow our style format ... fixed those ... makes code a little cleaner

image

### DIFF
--- a/src/foam/comics/v2/DAOSummaryView.js
+++ b/src/foam/comics/v2/DAOSummaryView.js
@@ -22,7 +22,7 @@ foam.CLASS({
     foam.pattern.Faceted.create()
   ],
 
-  css:`
+  css: `
     ^ {
       padding: 32px
     }
@@ -45,6 +45,12 @@ foam.CLASS({
 
     ^view-container {
       margin: auto;
+    }
+
+    ^truncate {
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
     }
   `,
 
@@ -145,7 +151,7 @@ foam.CLASS({
       this.SUPER();
       this
         .addClass(this.myClass())
-        .add(self.slot(function(data, data$id, config$CRUDActionsAuth$update, config$CRUDActionsAuth$delete,config$browseTitle, config$viewBorder, viewView) {
+        .add(self.slot(function(data, data$id, config$CRUDActionsAuth$update, config$CRUDActionsAuth$delete, config$browseTitle, config$viewBorder, viewView) {
 
           // iterate through permissions and replace % with data$id
           var editAction = self.EDIT;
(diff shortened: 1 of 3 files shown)


Nauna:
and changing update views ...

FROM:
image


TO:
image

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -154,9 +154,7 @@ foam.CLASS({
           if ( config$CRUDActionsAuth$update ) {
             var editArray = config$CRUDActionsAuth$update;
 
-            editArray = editArray.map(permission => {
-              return permission.replace("%", data$id);
-            });
+            editArray = editArray.map((permission) => permission.replace('%', data$id) );
```
TW80000:
```suggestion
            editArray = editArray.map((permission) => permission.replace('%', data$id));
```

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -166,10 +164,7 @@ foam.CLASS({
           if ( config$CRUDActionsAuth$delete ) {
             var deleteArray = config$CRUDActionsAuth$delete;
 
-
-            deleteArray = deleteArray.map(permission => {
-              return permission.replace("%", data$id);
-            });
+            deleteArray = deleteArray.map((permission) => permission.replace('%', data$id) );
```
TW80000:
```suggestion
            deleteArray = deleteArray.map((permission) => permission.replace('%', data$id));
```

TW80000 (commented):
I like the formatting fixes, but I actually don't think I agree with truncating long titles. I think it makes more sense for all of the information to remain visible than to hide part of it. Long titles will just wrap, won't they? In the example shown in the screenshots you posted, the title contains a very long "word" with no spaces, which means it won't wrap properly. I think it's more realistic to test with something like lorem ipsum:
Screen Shot 2020-02-03 at 1 16 48 PM

Personally I think it looks fine when it wraps and the user doesn't lose any information.

Nauna:
image


@TW80000  it does word wrap if multiple words.
But I would say if a user is getting to the point of a really long name it's probably because they are not writing in sentence structured names. I would imagine there might just be a really long name - consider LastNames ... but no research behind this. ... @kgrgreer @jlhughes either of you have a preference?

jlhughes:
The ellipse truncation is an issue, especially if the most significant info is at the end.  
Either ellipse in the middle or wrap is my preference, with wrapping over ellipse, unless both are possible.

Nauna:
have a solution one minute

Nauna:
Ok guys I think this will work?

It breaks up long words based on container, even if no spaces.

DAOSummaryView:
image
DAOUpdateView
image

Nauna:
and ... just to make sure I'm not breaking the current word-wrap with word spacing...
image

kgrgreer:
I'm fine either way. If someone enters such long names, then I don't really care if they get '...' or multiple lines. I doubt this will ever be an issue in practice.

jlhughes (approved):

PR #3122, opened 2020-02-03
Refresh page when session is destroyed on the server

TW80000:


src/foam/box/SessionClientBox.js
```diff
@@ -14,9 +14,19 @@ foam.CLASS({
...
     'requestLogin',
     'sessionTimer',
-    'group'
+    'window'
+  ],
+
+  messages: [
+    {
+      name: 'REFRESH_MSG',
+      message: 'Your session was deleted. The page will now be refreshed so that you can log in again.',
```
jlhughes: Wording 'deleted' -> 'expired' - I can see the user being concerned with 'deleted', but not 'expired'.  regardless of what we really do. 
```suggestion
      message: 'Your session has expired. The page will now be refreshed so that you can log in again.',
```

PR #3125, opened 2020-02-04
add auth to file service

nanoNeel:


src/foam/nanos/blob/FileService.java
```diff
@@ -61,11 +64,12 @@ protected void download(X x) {
...
       // IDs are unguessable cryptographically strong UUIDs, so no
       // permission check is really necessary.
-      
-// if ( userDAO_.find_(x, file.getOwner()) == null ) {
-//  resp.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
-//  return;
-// }
+      User owner = (User) userDAO_.find_(x, file.getOwner());
+      User user = (User) x.get("user");
+      if ( user == null || owner == null || (user.getId()!= owner.getId() && ! auth.check(x,"file.read.*")) ) {
```
jlhughes:
```suggestion
      if ( user == null || owner == null || (user.getId() != owner.getId() && ! auth.check(x, "file.read."+file.getd())) ) {
```

kgrgreer: Please follow FOAM style guide for spacing.

kgrgreer: Better to check file.read.<owner>, so you could have finer control. An admin with the file.read.* permission would still work.

TW80000 (approved):
LGTM

jlhughes (approved):

PR #3129, opened 2020-02-05
Adjusted delay for update history wait

KolombetM:
One of the async methods did not have enough time to save results into ruleHistory

### DIFF
--- a/src/foam/nanos/ruler/test/RulerDAOTest.java
+++ b/src/foam/nanos/ruler/test/RulerDAOTest.java
@@ -88,9 +88,16 @@ public class RulerDAOTest extends Test {
 
     // wait for async
     try {
-      Thread.sleep(asyncWait + 2000);
+      for (int i = 0; i < 2000; i++) {
+        Thread.sleep(asyncWait + 100);
+        ruleHistory = (RuleHistory) ruleHistoryDAO.find(ruleHistory.getId());
+        if ( ruleHistory.getResult().equals("Pending") )
+          continue;
+        else
+          break;
+      }
     } catch (InterruptedException e) { }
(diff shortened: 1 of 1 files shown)


src/foam/nanos/ruler/test/RulerDAOTest.java
```diff
@@ -88,7 +88,7 @@ public void testRuleHistory(X x) {
 
     // wait for async
     try {
-      Thread.sleep(asyncWait + 100);
+      Thread.sleep(asyncWait + 2000);
```
jlhughes: Thinking perhaps a better option is to loop for max amount of time.   so loop, delay 100, for a max of 2000 for example.

KolombetM: You don't seem to get out of the loop unless you finish all the iterations.
Screen Shot 2020-02-05 at 2 27 45 PM

jlhughes: move the find into the loop, exit early if not null.

src/foam/nanos/ruler/test/RulerDAOTest.java
```diff
@@ -88,9 +88,15 @@ public void testRuleHistory(X x) {
 
     // wait for async
     try {
-      Thread.sleep(asyncWait + 100);
+      for (int i = 0; i < 2000; i++) {
+        Thread.sleep(asyncWait + 100);
+        ruleHistory = (RuleHistory) ruleHistoryDAO.find(ruleHistory.getId());
+        if ( ruleHistory.getResult().equals("Pending") )
```
KolombetM: Since ruleHistory in the pending state, until it is done, we wait for the state to be changed.

jlhughes (approved):

PR #3136, opened 2020-02-06
minor fixes

Nauna:
moving 'complianceHistoryDAO' property which is business service out of framework. Already merged business logic with https://github.com/nanoPayinc/NANOPAY/pull/9254


jlhughes (approved):

src/foam/nanos/ruler/UserRefines.js
```diff
@@ -9,7 +9,7 @@ foam.CLASS({
     name: 'UserRefines',
     refines: 'foam.nanos.auth.User',
     imports: [
-      'complianceHistoryDAO'
+      'complianceHistoryDAO?'
```
jlhughes: We can't have this on the foam side. This needs to be a nanopay userRefinement.

jlhughes (changes requested):

src/foam/nanos/ruler/UserRefines.js
Nauna: removed this from foam and placed in business logic
https://github.com/nanoPayinc/NANOPAY/pull/9254

TW80000 (approved):

PR #3143, opened 2020-02-07
Alert when results exceed maximum download size.

jlhughes:
The results are stored in the href which has a limit of 524288 characters.
This will need refactoring to download as a file.


kgrgreer:
Brilliant!

PR #3144, opened 2020-02-07
add link to email temp

yij793:
address: https://nanopay.atlassian.net/browse/CPF-4551
notes: the button not working since  {{link}} is not being set.

### DIFF
--- a/src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js
+++ b/src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js
@@ -107,10 +107,13 @@ return true;`
 // context. Therefore we put the system user in the context here so that
 // decorators down the line won't throw NPEs when trying to access the user in
 // the context.
+AppConfig appConfig = (AppConfig) x.get("appConfig");
 User systemUser = (User) getX().get("user");
 x = x.put("user", systemUser);
 
+
 String newPassword = user.getDesiredPassword();
+String url = appConfig.getUrl().replaceAll("/$", "");
 
 DAO userDAO = (DAO) getLocalUserDAO();
 DAO tokenDAO = (DAO) getTokenDAO();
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

src/foam/nanos/auth/resetPassword/ResetPasswordTokenService.js
```diff
@@ -107,10 +107,13 @@ return true;`
 // context. Therefore we put the system user in the context here so that
 // decorators down the line won't throw NPEs when trying to access the user in
 // the context.
+AppConfig appConfig = (AppConfig) x.get("appConfig");
 User systemUser = (User) getX().get("user");
 x = x.put("user", systemUser);
 
+
 String newPassword = user.getDesiredPassword();
+String url = appConfig.getUrl().replaceAll("/$", "");
```
Nauna, line 116: can you make this group.appConfig pls.
like: https://github.com/nanoPayinc/NANOPAY/pull/9243/files

kgrgreer: FYI: replaceAll() is expensive because it compiles the regex every time, but I guess it's fine here since this isn't a high-frequency operation.

jlhughes: re: appconfig - Session.applyTo has been updated to do this for us, so pulling appConfig from the context is fine.  Thanks for catching that Anna.

jlhughes: actually we should just do this when setting up AppConfig.  I'll do that.

jlhughes: not as straightforward as I thought, we'll leave this for now.

PR #3147, opened 2020-02-10
Add `$push` and `$remove` to Array properties

KernelDeimos:
This PR adds `someName$push` and `someName$remove` to supplement a property `someName` of property type `Array` or `FObjectArray`.

For example:
```
foam.CLASS({
  package: 'com.example',
  name: 'CountArrayChanges',

  properties: [
    {
      name: 'values',
      class: 'Array',
      postSet: function () { this.changeCount++; }
    },
    {
      name: 'changeCount',
      class: 'Int',
      value: 0
    }
  ]
});

let subject = com.example.CountArrayChanges.create();
subject.values = [];
console.log(subject.changeCount); // 1
subject.values.push(1);
console.log(subject.changeCount); // 1
subject.values$push(1);
console.log(subject.changeCount); // 2
```


src/foam/core/FObjectArray.js
```diff
@@ -64,5 +64,39 @@ foam.CLASS({
...
+    function installInProto(proto) {
+      this.SUPER(proto);
+      var self = this;
+      Object.defineProperty(proto, self.name + '$push', {
+        get: function classGetter() {
+          return function (v) {
+            // Push value
+            this[self.name].push(v);
+            // Force property update
+            this[self.name] = this[self.name];
```
kgrgreer: Does this work? I thought if the property doesn't change then we don't fire the property change event.

src/foam/core/FObjectArray.js
```diff
@@ -64,5 +64,39 @@ foam.CLASS({
...
+        },
+        configurable: true
+      });
+      Object.defineProperty(proto, self.name + '$remove', {
+        get: function classGetter() {
+          return function (predicate) {
+            // Faster than splice or filter as of the time this was added
+            let oldArry = this[self.name];
+            let newArry = [];
+            for ( let i=0; i < oldArry.length; i++ ) {
```
kgrgreer: Please follow FOAM style guideline for spacing.

src/foam/core/FObjectArray.js
```diff
@@ -64,5 +64,39 @@ foam.CLASS({
...
       }
     }
+  ],
+
+  methods: [
+    function installInProto(proto) {
+      this.SUPER(proto);
+      var self = this;
+      Object.defineProperty(proto, self.name + '$push', {
+        get: function classGetter() {
```
kgrgreer: Why is this named classGetter?

src/foam/core/FObjectArray.js
```diff
@@ -64,5 +64,39 @@ foam.CLASS({
...
+            // Push value
+            this[self.name].push(v);
+            // Force property update
+            this[self.name] = this[self.name];
+          }
+        },
+        configurable: true
+      });
+      Object.defineProperty(proto, self.name + '$remove', {
+        get: function classGetter() {
```
kgrgreer: Why is this named classGetter?

kgrgreer:
Could FObjectArray extend Array and share some of the method definitions?

src/foam/core/FObjectArray.js
KernelDeimos: That's what I thought too, but when I added this it changed the behaviour to fire every time.

src/foam/core/FObjectArray.js
KernelDeimos: I did this to be consistent with types.js. I don't believe it changes the behaviour of anything so I assumed it was done for clarity. If you want this removed I can make another PR after to remove it from all of them.

kgrgreer: The name is given so that stack traces have meaningful names, but it was called classGetter when it was the getter, the name doensn't matter but it should be something meaningful or left out.

PR #3148, opened 2020-02-11
Ensure AppConfig is retrieved via user.group.

jlhughes:
In context setup, only over write AppConfig if user/group not available.


src/foam/box/SessionServerBox.java
```diff
@@ -94,7 +94,8 @@ public void send(Message msg) {
       session.touch();
 
       // TODO: Shouldn't this go somewhere else?
-      if ( req != null && ! SafetyUtil.isEmpty(req.getRequestURI()) ) {
+      if ( effectiveContext.get("group") == null &&
```
kgrgreer: How about some comments to explain what this is doing and why?

src/foam/nanos/auth/email/EmailTokenService.js
```diff
@@ -34,9 +34,8 @@ foam.CLASS({
 `try {
 DAO tokenDAO = (DAO) getX().get("localTokenDAO");
 DAO userDAO  = (DAO) getX().get("localUserDAO");
-AppConfig appConfig = (AppConfig) getX().get("appConfig");
-String url = appConfig.getUrl()
-    .replaceAll("/$", "");
+AppConfig appConfig = user.findGroup(x).getAppConfig(x);
```
TW80000, line 37:
```suggestion
AppConfig appConfig = (AppConfig) x.get("appConfig");
```

Does this work instead? The app config should already be set based on the group (if there's a group available). This happens in `Session.applyTo`, which gets called for every request to create the user's server-side request context. Not sure why this was using `getX()` instead of `x` here.

src/foam/box/SessionServerBox.java
TW80000: Why do you want to make this change? Wouldn't it be appropriate to do it even when the appConfig has been determined by the group?

src/foam/nanos/auth/email/EmailTokenService.js
jlhughes: The AppConfig is set based on the group when the user is logged in. But when the user is not logged in we still need to go through the user's group to get group appconfig overrides - such as sme url. 
yes, not sure about getx().

src/foam/box/SessionServerBox.java
jlhughes: I'll add some comments, but when the user was logged in, the group refined appConfig was being replaced.

src/foam/nanos/auth/email/EmailTokenService.js
TW80000: Ah, thanks, that makes sense. And I guess the user in `x` when there is one isn't necessarily the one we want to use anyway, it makes more sense to use the one passed as an argument to `generateTokenWithParameters` because that's who's getting the email.

src/foam/box/SessionServerBox.java
jlhughes: Moved to Session where it's obvious.

PR #3150, opened 2020-02-11
changes to validation for accounting for autocomplete

Nauna:
addresses: https://nanopay.atlassian.net/browse/CPF-4571

Went through options with design to handle auto-validation, and so:

decided that it's best to ignore that the button is reading an error, since selecting 'Sign In' still functions as a user action, and works correctly as soon as that happens.


Nauna:
This worked while demoing to design, but didn't when demoing to Kevin. SO it was decided to remove the validation, and revert validation to its original form.

PR #3151, opened 2020-02-11
Remove Login Using User Id And Add Login Using UserName

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/CPF-4063

#New branch PR extending from https://github.com/foam-framework/foam2/pull/2989


src/foam/nanos/auth/AuthService.js
```diff
@@ -77,31 +77,7 @@ foam.INTERFACE({
...
-      javaThrows: ['foam.nanos.auth.AuthenticationException'],
-      swiftThrows: true,
-      args: [
-        {
-          name: 'x',
-          type: 'Context'
-        },
-        {
-          name: 'email',
+          name: 'id',
```
jlhughes: perhaps 'identifier' to distinguish if from model 'id'.

jlhughes (approved):

PR #3154, opened 2020-02-12
Table Filter Loading Fix

blakegreen32:
If a table property filter returns an empty array in the filter view, a message is displayed "- NO OPTIONS AVAILABLE -" instead of always showing "- LOADING OPTIONS -" even though no data is returned.


kennyqiyenkan (approved):
LGTM. Though, maybe the use of `===` might be better. @jlhughes, wanna chime in?

jlhughes:
In this particular either (==, ===) is correct.

PR #3157, opened 2020-02-13
Refactor logging to improve performance:

jlhughes:
- use foam.util.SyncFastTimestamper to generate timestamps
- explicitly add createdBy and createdByAgent rather than use decorators


TW80000 (approved):
Can't say I know much about this part of the code, but the changes here look fine to me. Why did you want to remove the decorators for `createdBy` and `createdByAgent`?

jlhughes:
> Can't say I know much about this part of the code, but the changes here look fine to me. Why did you want to remove the decorators for `createdBy` and `createdByAgent`?

They don't work LogMessage.  The find(obj) always returns non-null.  Could be FixedSizeDAO or that id is transient.   But in any case this also speeds things up the tiniest bit as they are always 'create'.

PR #3159, opened 2020-02-13
[CPF-4574] Add assertion to 'of' of DAOProperty

KernelDeimos:
Adds an assertion in case `of` of DAOProperty has the wrong value.

Example - The following code looks okay because for DAOs `of` is supposed to the the class stored by the DAO:
```
  properties: [
    {
      name: 'featuredCapabilities',
      of: 'foam.nanos.crunch.Capability',
      documentation: `
        DAO Property to find four capabilities to feature.
      `,
      factory: function () {
        return this.capabilityDAO;
      }
    },
  ],
```

This is incorrect because DAOProperty extends FObjectProperty, and it follows that `of` is the class of the DAO and should be `foam.dao.DAO` literally or a subclass. The error currently displayed when this happens is `Cannot instantiate an Interface`, which does not indicate the real problem.

This is the new error message (assertion error) added by this PR:
```
Assertion failed: DAOProperty's type (specified by 'of') must be a subclass of DAO, but the type for 'featuredCapabilities' is 'foam.nanos.crunch.Capability'. (that's not a DAO)
```

### DIFF
--- a/src/foam/dao/DAOProperty.js
+++ b/src/foam/dao/DAOProperty.js
@@ -41,7 +41,19 @@ foam.CLASS({
       value: 'HIDDEN'
     },
     ['transient', true],
-    ['of', 'foam.dao.DAO'],
+    {
+      class: 'Class',
+      name: 'of',
+      value: 'foam.dao.DAO',
+      assertValue: function(ofValue) {
+        var type = foam.lookup(ofValue);
+        foam.assert(foam.dao.DAO.isSubClass(type),
+          `DAOProperty's type (specified by 'of') must be a ` +
+          `subclass of DAO, but the ` +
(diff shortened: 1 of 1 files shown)


src/foam/dao/DAOProperty.js
```diff
@@ -41,7 +41,19 @@ foam.CLASS({
       value: 'HIDDEN'
     },
     ['transient', true],
-    ['of', 'foam.dao.DAO'],
+    {
```
kgrgreer, line 44: Why would we even let you set the 'of' of a DAOProperty? Shouldn't it always just be 'DAO'? Couldn't we just override the setter to provide a warning that you can't set it?

PR #3163, opened 2020-02-13
Make classReferenceParser singletons to save memory.

hchoura:


src/foam/lib/json/ClassReferenceParser.java
```diff
@@ -12,6 +12,10 @@
 public class ClassReferenceParser
   extends ProxyParser
 {
+  private final static Parser instance__ = new ClassReferenceParser();
+
+  public static Parser instance() { return instance__; }
+
   public ClassReferenceParser() {
```
kgrgreer: Make the constructor private so nobody tries to skip the instance() method.

jlhughes (approved):

PR #3185, opened 2020-02-18
Add label to street number field in Address.js

jinnnn9:


jinnnn9:
The default label is too long and misaligns input fields in address view.

PR #3189, opened 2020-02-18
Rule ID change to GUID

noodlemoodle:
migrate rules to have long ids generated by guid instead of user-defined strs

nano side https://github.com/nanoPayinc/NANOPAY/pull/9354


src/foam/nanos/ruler/Rule.js
```diff
@@ -380,8 +386,8 @@
       name: 'javaExtras',
       buildJavaClass: function(cls) {
         cls.extras.push(`
-        public static Rule findById(Collection<Rule> listRule, String passedId) {
-          return listRule.stream().filter(rule -> passedId.equals(rule.getId())).findFirst().orElse(null);
+        public static Rule findById(Collection<Rule> listRule, Long id) {
```
kgrgreer: Why Long instead of long?

src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -150,7 +150,7 @@ public void putResult(Object result) {
     results_.put(currentRule_.getId(), result);
   }
 
-  public Object getResult(String ruleId) {
+  public Object getResult(Long ruleId) {
```
kgrgreer: Long -> long ?

src/foam/nanos/ruler/Rule.js
kgrgreer: Where is this called from? Is it performance sensitive? If so, could you use a Map instead?

noodlemoodle: i think this is called from a listener on ruledao in the rulerdao decorator to add to or remove from the rules list of the target dao whenever the ruledao changes

src/foam/nanos/ruler/RuleEngine.java
noodlemoodle: sorry, typo

kgrgreer:
LGTM

jlhughes (changes requested):
Ok, miscommunication here. 
Id should be class 'String' and set by GUID. 
We changed then from 'long' to String, as it was impossible to add them from the repository without conflict.

src/foam/nanos/ruler/Rule.js
```diff
@@ -63,7 +65,11 @@
     {
       class: 'String',
       name: 'id',
-      updateVisibility: 'RO',
+      hidden: true
```
jlhughes:
```suggestion
      visibility: 'HIDDEN'
```

jlhughes (approved):

PR #3191, opened 2020-02-18
generate language per location

hchoura:
to run the test use:
node demos/android/gen_xml.js

@nanoscott

### DIFF
--- a/demos/android/gen_xml.js
+++ b/demos/android/gen_xml.js
@@ -1,5 +1,12 @@
 #!/usr/bin/env node
 
+global.FOAM_FLAGS = {
+    web: true,
+    js: true,
+    debug: true,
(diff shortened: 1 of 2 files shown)


nanoscott:
@hchoura .. does this only work for Android code? What does it actually do?

hchoura:
> @hchoura .. does this only work for Android code? What does it actually do?

It work for other platform. it is the same concept.
It will generate in foam2/demos/android/gen 2 files. one for en, and another for French. it is based on the test example in demos/android/js/nodetooldemo/Test.js

kgrgreer:
Is this compatible with: https://github.com/foam-framework/foam2/pull/3202 ? Does anything need to be done because of that commit?

PR #3199, opened 2020-02-20
Adding support for test suites

nanoscott:


src/foam/nanos/test/TestConfig.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
```
jlhughes: What is the point of this model?  It's not used in Test.js. 
I see it on the nanopay side, but how is it used? 
Assuming this is associated with CI, why can't we set this as a System.property from build.sh?  
Or how did you envision it working with the build?

src/foam/nanos/script/TestRunnerScript.js
```diff
@@ -76,8 +77,13 @@ foam.CLASS({
         loggerFilter.setLogInfo(false);
         loggerFilter.setLogWarning(false);
 
+        TestConfig testConfig = (TestConfig) x.get("testConfig");
```
nanoscott: @jlhughes TestConfig is used here.

src/foam/nanos/test/TestConfig.js
nanoscott: @jlhughes Here is my vision .. we have separate deployment directories for benchmark and liquid (already exists) .. in those deployment directories we have a services.jrl file with a TestConfig NSpec that has the test suite specified. This test suite is used to filter the tests that are run for that deployment. So if you tried to run ./build.sh -cQt .. all the tests would be filtered by the 'liquid' test suite .. or if we had a separate benchmark deployment, ./build.sh -ctJbenchmark would run the 'benchmark' test suite.

src/foam/nanos/script/TestRunnerScript.js
nanoscott: I could rename it TestRunnerConfig instead of just TestConfig

jlhughes: yes please.

jlhughes (approved):
Approved - with one request to rename TestConfig to TestRunnerConfig.

src/foam/nanos/script/TestRunnerScript.js
nanoscott: rename affected ..

PR #3200, opened 2020-02-20
[DO_NOT_MERGE] [CPF-4715] Add ViewCapability interface for visible capabilities

KernelDeimos:
This PR adds view information to capabilities
- Capabilities that are visible to the user need additional properties such as a label, `isFeatured`, `isNew`, and possibly others in the future.
- It is not ideal to store these properties for all capabilities as more granular capabilities in the future will result in many hidden capabilities (for which these properties don't apply)
- https://nanopay.atlassian.net/browse/CPF-4715

The proposed solution is to have a ViewCapability interface. Any `visible` capability should also implement `ViewCapability` which will add this meta information.

### DIFF
--- a/src/foam/nanos/crunch/types/ViewCapability.js
+++ b/src/foam/nanos/crunch/types/ViewCapability.js
@@ -1,0 +1,50 @@
+foam.CLASS({
+  package: 'foam.nanos.crunch.types',
+  name: 'CapabilityViewMeta',
+
+  /*
+    Interesting note: since this class is only used by the
+      interface below, this would never fail in the
+      classloader even if one-model-per-file was implemented.
+  */
+
+  properties: [
+    {
+      name: 'label',
+      class: 'String',
+      documentation: `
+        A label for this capability when it is displayed to a user.
+      `
+    },
+    {
+      name: 'isFeatured',
+      class: 'Boolean'
+    },
+    {
+      name: 'isNew',
+      class: 'Date',
+      documentation: `
+        Date before which this capability should be displayed as
+        "new", or null for no indicator.
+      `
+    }
+  ]
+});
+
+foam.INTERFACE({
+  package: 'foam.nanos.crunch.types',
+  name: 'ViewCapability',
+
+  properties: [
+    {
+      name: 'viewMeta',
+      class: 'FObjectProperty',
+      of: 'CapabilityViewMeta',
+      documentation: `
+        To prevent cluttering the axiom namespace of a Capability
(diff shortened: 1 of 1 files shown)


Nauna:
ANNA
1)how do you invision a payable capability being defined with CapabilityViewMeta?
2) what is the benefit of an interface over a property on Capability?

ERIC -- sorry buddy just coping from slack:
So the payable capability would have to be on Nanopay side I think, and it would be another interface, like foam.INTERFACE({ id: 'net.nanopay.tx.crunch.PayToGetCapability' }) or something similar

interface over property means that a capability doesn’t have to have CapabilityViewMeta (if not visible for instance), and doesn’t have to have CapabilityPaymentMeta

ANNA
If we are defining meta data I don’t see why a capability can’t be payable on foam. Implementing the payments will be on nanopay, but a capabilityType doesn’t need to be defined on nanopay. …
A property can be null, and tbh (sorry) still see no benefit to this way of doing things, actually from what I’m understanding we would need an interface for each type and then I’d ask why then have a modeled property.
Why not create a property on Capability that is an Array of capabilityMetaEnum. This way we get our functionality and are only adding one enum class and one property. We can expand the enum at anytime and add functionality to each capability at anytime???

KernelDeimos:
@Nauna We could implement a `PayToGetCapability` on FOAM and then a refinement could be used to change the type for `price` - or maybe FOAM itself could have a `Price` type (which now that I think about it, I like better because it makes FOAM more useful for ecommerce applications and the like) (@kgrgreer what do you think?)

I think an interface for each type is only necessary for very general types of capabilities. My main reason for not adding properties to Capability is that I don't want to clutter backend functionality. ViewCapabilityMeta (or PaymentCapabilityMeta) should be separate from the Capability interface that CapabilityAuthService can see, otherwise I can imagine a scenario where a hack in CapabilityAuthService checks for a payment on a payment-required capability rather than the state of the capability being properly updated.

Regarding an `FObjectyArray of: CapabilityMetaEnum`, I'm not sure how this would fit things like `label` and `price` (but it would definitely work for flags).

Nauna:
@KernelDeimos understand your reasoning now.

So:
 payable capability - needs information like price
 capability - needs information (view data) like label

Good thinking. Ok so lets see what the bosses say, but I now agree with this approach.

Would still suggest:
payable interface with price on foam,
and capabilities to have property label, if we need it for all displayable capabilities,
and we should perhaps talk to product/design about the necessity of 'isNew'

PR #3203, opened 2020-02-20
Changed Script.js run action to save run details into dao which calls the action

KolombetM:


src/foam/nanos/script/Script.js
```diff
@@ -166,6 +167,12 @@ foam.CLASS({
       of: 'foam.nanos.auth.User',
       name: 'lastModifiedBy',
       documentation: 'User who last modified script'
+    },
+    {
+      class: 'String',
```
jlhughes, line 172: Please documentation on why this is necessary and how to set it from Reports for example. Thanks.

jlhughes (approved):

PR #3207, opened 2020-02-20
add cache from liquidCapabilityAuthService to capabilityAuthService

noodlemoodle:
https://github.com/nanoPayinc/NANOPAY/pull/9242


nanoscott (approved):

jlhughes (approved):

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -17,20 +17,83 @@ foam.CLASS({
...
+        },
+      ],
+      javaCode: `
+        if ( getInitialized() )
+          return;
+
+        DAO userCapabilityJunctionDAO = (DAO) x.get("userCapabilityJunctionDAO");
+        DAO capabilityDAO = (x.get("capabilityDAO") == null ) ? (DAO) x.get("capabilityDAO") : (DAO) x.get("localCapabilityDAO");
+        if ( capabilityDAO == null || userCapabilityJunctionDAO == null )
+          return;
```
kgrgreer, line 73: Should we still mark it as initialized?

kgrgreer: Would it be easier to just put this in the Object's init() method, which gets called when it is constructred?

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -55,12 +118,21 @@ foam.CLASS({
         if ( x.get(Session.class) == null ) return false;
         if ( user == null || ! user.getEnabled() ) return false;
 
-        // Check whether user has permission to check user permissions.
-        if ( ! getDelegate().check(x, "service.auth.checkUser") ) return false;
+        this.initialize(x);
+
+        String key = user.getId() + permission;
+        if ( (( Map<String, Boolean> )getCache()).containsKey(key) ) {
```
kgrgreer: Could you do this with only one lookup instead of two?

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -71,32 +143,61 @@ foam.CLASS({
...
+          // Add the result to the cache
+          (( Map<String, Boolean> ) getCache()).put(key, result);
+
         } catch (Exception e) {
           Logger logger = (Logger) x.get("logger");
           logger.error("check", permission, e);
         }
 
-        return getDelegate().checkUser(x, user, permission);
+        return result ? true : getDelegate().checkUser(x, user, permission);
```
kgrgreer: Simpler: return result || getDelegate().checkUser(x, user, permission);

src/foam/nanos/auth/CapabilityAuthService.js
noodlemoodle: re: mark as initialized 
I think if either capabilityDAO or userCapabilityJunctionDAO is null, this shouldn't be initialized. if the daos are null then we cant attach the listeners on them in the initialization code, so it wouldneed to be re-initialized later on anyway

re: init
I tried to put this in the init method, but I think this is causing recursion and npe problems in boot since the serviceDAO_ used to create the context during boot uses the permissionedPropertyDAO which uses auth chain

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -71,32 +142,61 @@ foam.CLASS({
...
+            // if the user has the permission, store this in the cache and return the result
+            // otherwise, move on to the 2nd part of the check
+            if ( result ) {
+              ((Map<String, Boolean>) getCache()).put(key, result);
+              return result;
+            }
           }
+
+          // 2. check if the user has a capability that grants the permission
+          ProxySink proxy = new ProxySink(x, new LimitedSink(x, 1, 0, new ArraySink())) {
```
kgrgreer: What is the ProxySink for?

kgrgreer: Couldn't you just write a Predicate and use where() instead?

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -71,32 +142,61 @@ foam.CLASS({
...
+            int count = 0;
+            @Override
+            public void put(Object o, Detachable sub) {
+              UserCapabilityJunction ucj = (UserCapabilityJunction) ((UserCapabilityJunction) o).deepClone();
```
kgrgreer: Why do you need to deepClone() the UCJ?

noodlemoodle: I got these 2 comments via email but they are not showing up on the page
> What is the ProxySink for?
> Couldn't you just write a Predicate and use where() instead?
ProxySink is used to override put and only put the ucjs where the capability's implies method returns true for the permission. Do we have a way to do this with predicate?

changed to predicate

> Why do you need to deepClone() the UCJ?
Sorry I can't remember why either.. but I don't think this is needed, will remove

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -55,12 +119,20 @@ foam.CLASS({
         if ( x.get(Session.class) == null ) return false;
         if ( user == null || ! user.getEnabled() ) return false;
 
-        // Check whether user has permission to check user permissions.
-        if ( ! getDelegate().check(x, "service.auth.checkUser") ) return false;
+        this.initialize(x);
+
+        String key = user.getId() + permission;
+        Boolean result = ( (Map<String, Boolean>) getCache() ).get(key);
+        if ( result != null ) return result;
```
nanoscott: should this be checking the delegate too?

noodlemoodle: Yes added now, thanks

nanoscott (approved):

PR #3211, opened 2020-02-21
Add support for multi languages

hchoura:


kgrgreer:
Is this breaking some tests?

nanoscott (approved):

nanoscott:
@hchoura @kgrgreer @jlhughes .. we need a way to control which 'language' is used per application .. so 'en-liquid' needs to be able to be shown, or 'fr-liquid', etc. So how can we control the 'language' through configuration here?

PR #3212, opened 2020-02-21
multi language example

hchoura:


src/com/foamdev/demos/helloLocale/Controller.js
```diff
@@ -0,0 +1,50 @@
...
+        'en'   : 'Hello ',
+        'en-US': 'Howdy ',
+        'fr'   : 'Bonjour ',
+      },
+    },
+  ],
+
+  methods: [
+    function initE() {
+      this.start('div').add(this.yourName$.prop.label).end().start('div').add(this.YOUR_NAME).end();
```
kgrgreer: Why not just this.YOUR_NAME.label?

PR #3215, opened 2020-02-24
Add EmailTemplate Property to EmailMessage

nanopay-arthur:
Will help debugging email related bugs on production.

### DIFF
--- a/src/foam/nanos/notification/email/EmailMessage.js
+++ b/src/foam/nanos/notification/email/EmailMessage.js
@@ -21,7 +21,8 @@ foam.CLASS({
     'subject',
     'to',
     'from',
-    'status'
+    'status',
+    
   ],
 
   properties: [
@@ -105,9 +106,9 @@ foam.CLASS({
       tableWidth: 100
     },
     {
-      class: 'String',
-      name: 'templateName',
-      value: null
+      class: 'Map',
+      name: 'templateArguments',
+      view: { class: 'foam.u2.view.MapView' }
     }
   ]
 });
--- a/src/foam/nanos/notification/email/EmailMessage.js
+++ b/src/foam/nanos/notification/email/EmailMessage.js
@@ -104,5 +104,10 @@ foam.CLASS({
       name: 'status',
       tableWidth: 100
     },
+    {
+      class: 'String',
+      name: 'templateName',
+      value: null
+    }
   ]
 });
--- a/src/foam/nanos/notification/email/EmailMessage.js
+++ b/src/foam/nanos/notification/email/EmailMessage.js
@@ -21,8 +21,7 @@ foam.CLASS({
     'subject',
     'to',
     'from',
-    'status',
-    
+    'status'
   ],
 
   properties: [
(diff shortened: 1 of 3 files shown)


src/foam/util/Emails/EmailsUtility.java
```diff
@@ -47,6 +47,7 @@ public static void sendEmailFromTemplate(X x, User user, EmailMessage emailMessa
     // Add template name to templateArgs, to avoid extra parameter passing
     if ( ! SafetyUtil.isEmpty(templateName) ) {
       if ( templateArgs != null ) {
+        emailMessage.setTemplateName(templateName);
```
jlhughes: Can you explain more how this is helpful?  Is it that the templateArgs are not visible anywhere? 
If the templateArgs were added to the email message then it would take care of itself.   The emailMessage could format itself, and also when viewing the emailMessage as a model you could see the templateArgs (key/values pairs).

kgrgreer: The message says it is to help with debugging.
If we stored the template name and args instead of the body, then the email messages could be much smaller.

nanoscott: The basic idea here is that we have a recurring error message in the logs that says we are trying to send an email message that has no content, but we aren't able to tell which email is actually failing because we can't see the code path that it's coming from .. the easiest way to discern the code path would be to see the email template that is being used. In order to figure out the email template, we can either just log it .. which we could do here .. or we could add it to the email message model, so that emails that are stuck in the 'UNSENT' state will have more information on where they came from .. because right now we aren't seeing anything in production for the unsent messages (i.e. the message content + subject are completely empty). So which do you prefer?

src/foam/nanos/notification/email/EmailMessage.js
```diff
@@ -21,7 +21,8 @@ foam.CLASS({
     'subject',
     'to',
     'from',
-    'status'
+    'status',
```
kgrgreer: Why add a trailing comma and blank line?

src/foam/nanos/notification/email/EmailMessage.js
```diff
@@ -105,9 +106,9 @@ foam.CLASS({
       tableWidth: 100
     },
     {
-      class: 'String',
-      name: 'templateName',
-      value: null
+      class: 'Map',
```
kgrgreer, line 108: Do you not need templateName?

nanopay-arthur: One of the key/value pairs in templateArguments is [emailTemplate, emailTemplateName].

src/foam/util/Emails/EmailsUtility.java
nanoscott: Looks like the template args has the template name in it .. so that's the easiest way plus we get more information about what we are trying to populate .. better solution all around @jlhughes .. I didn't see that the templateName was in there, sorry.

nanoscott (approved):

jlhughes (approved):

PR #3217, opened 2020-02-24
Check for null value in adapt email type

zahnno:
toLowerCase throws an exception on adapt if no email is provided

### DIFF
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -435,7 +435,7 @@ foam.CLASS({
     [
       'adapt',
       function(_, v) {
-        return v.toLowerCase().trim();
+        return v || v === '' ? v : v.toLowerCase().trim();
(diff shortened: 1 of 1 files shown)


src/foam/core/types.js
```diff
@@ -435,7 +435,7 @@ foam.CLASS({
     [
       'adapt',
       function(_, v) {
-        return v.toLowerCase().trim();
+        return v || v === '' ? v : v.toLowerCase().trim();
```
kgrgreer, line 438: This doesn't make sense to me.

zahnno: If v is null, toLowerCase throws an uncaught type error

zahnno:
Closing due to adapt being reverted back to preSet resolving the uncaught type error.

PR #3220, opened 2020-02-25
Cpf 4666/permission menu dao controller for table view create and detail view edit and delete

olhabn:
https://nanopay.atlassian.net/browse/CPF-4666


src/foam/comics/v2/DAOControllerConfig.js
```diff
@@ -138,8 +138,11 @@ foam.CLASS({
     },
     {
       class: 'FObjectProperty',
-      of: 'foam.comics.v2.CRUDActionsAuth',
-      name: 'CRUDActionsAuth'
+      of: 'foam.comics.v2.CRUDActionsPredicate',
```
kgrgreer: What advantage do we get by making this be a nested Object rather than just adding three Predicate properties here? Are we using the CRUDActionsPredicate somewhere else?

src/foam/lib/PermissionPredicate.js
```diff
@@ -0,0 +1,47 @@
+foam.CLASS({
```
kgrgreer, line 7: Add copyright header.

src/foam/comics/v2/CRUDActionsPredicate.js
```diff
@@ -0,0 +1,18 @@
+foam.CLASS({
```
jlhughes: license header, please - and any other files it's missing.

src/foam/comics/v2/CRUDActionsPredicate.js
```diff
@@ -0,0 +1,18 @@
...
+  package: 'foam.comics.v2',
+  name: 'CRUDActionsPredicate',
+  properties: [
+    {
```
jlhughes: why not 'read'? I may find further on but how do we hide actions for example?

src/foam/comics/v2/DAOBrowseControllerView.js
```diff
@@ -98,12 +98,16 @@ foam.CLASS({
...
 
       this.addClass(this.myClass())
-      .add(this.slot(function(data, config, config$CRUDActionsAuth$create, config$browseBorder, config$browseViews, config$browseTitle) {
-        var createAction = config$CRUDActionsAuth$create
-            ? self.CREATE.clone().copyFrom({
-              availablePermissions: self.CREATE.availablePermissions.concat(config$CRUDActionsAuth$create)
-            })
-            : self.CREATE;
+      .add(this.slot(function(data, config, config$browseBorder, config$browseViews, config$browseTitle) {
+        if ( ! config.CRUDPermission || ! config.CRUDPermission.create || ! config.CRUDPermission.create.f ) {
```
jlhughes: perhaps simplified with: 
`
  config.createEnabled = false;
  config.createEnabled = config.CRUDPermission && config.CRUDPermission.create && config.CRUDPermission.create.f ? config.CRUDPermission.f() || true; 
`

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -150,30 +150,31 @@ foam.CLASS({
...
-            deleteAction = self.DELETE.clone().copyFrom({
-              availablePermissions: self.DELETE.availablePermissions.concat(deleteArray)
-            });
+        .add(self.slot(function(data, data$id, config$viewBorder, viewView) {
+
+          if ( ! self.config.CRUDPermission) {
+            self.config.editEnabled = true;
+            self.config.deleteEnabled = true;
+          } else {
+            if ( ! self.config.CRUDPermission.update || ! self.config.CRUDPermission.update.f ) {
```
jlhughes: Simplify if you can, similar to above.

src/foam/lib/PermissionPredicate.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+  ],
+  javaImports: [
+    'foam.nanos.auth.AuthService'
+  ],
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: `
```
jlhughes, line 29: The PermissionPredicates above, when is x passed in? 
Can we not pass 'x' as the 'obj' in f(obj)?  On the java side will this not be the system context?

src/foam/lib/PermissionPredicate.js
jlhughes: @olhabn the current header. 
`/**
 * @license
 */
`

src/foam/lib/PermissionPredicate.js
olhabn: Fixed

src/foam/comics/v2/DAOControllerConfig.js
olhabn: Fixed

jlhughes (approved):

src/foam/comics/v2/DAOControllerConfig.js
```diff
@@ -138,8 +138,11 @@ foam.CLASS({
...
+      name: 'CRUDPermission',
```
jlhughes: property names should not begin with upper case. 
```suggestion
      name: 'crudPermission',  
```

src/foam/comics/v2/DAOControllerConfig.js
jlhughes: Why do we have this permission and still have 'created/edit/deletedEnabled' above?

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -150,30 +150,31 @@ foam.CLASS({
...
-
-            deleteArray = deleteArray.map((permission) => permission.replace('%', data$id));
-
-            deleteAction = self.DELETE.clone().copyFrom({
-              availablePermissions: self.DELETE.availablePermissions.concat(deleteArray)
-            });
+        .add(self.slot(function(data, data$id, config$viewBorder, viewView) {
+
+          if ( ! self.config.CRUDPermission) {
+            self.config.editEnabled = true;
```
jlhughes: If we are retaining editEnabled (and others above) for backwards compatibility (I presume), then self.config.editEnabled is either set or already has a default.

jlhughes (approved):

src/foam/comics/v2/DAOBrowseControllerView.js
```diff
@@ -77,8 +77,12 @@ foam.CLASS({
   actions: [
     {
       name: 'create',
-      isAvailable: function(config$createEnabled) {
-        return config$createEnabled;
+      isAvailable: function(config) {
+        try {
+          return ! config.create || ! config.create.f ? true : config.create.f();
```
kgrgreer, line 82: Why is this needed?

src/foam/comics/v2/DAOControllerConfig.js
```diff
@@ -119,27 +119,16 @@ foam.CLASS({
       }
     },
     {
-      class: 'Boolean',
-      name: 'createEnabled',
-      documentation: 'If set to false, the "Create" button will not be visible.',
-      value: true
+      class: 'foam.mlang.predicate.PredicateProperty',
+      name: 'create'
```
kgrgreer, line 123: Doesn't this cause a name conflict?

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -107,8 +107,12 @@ foam.CLASS({
   actions: [
     {
       name: 'edit',
-      isAvailable: function(config$editEnabled) {
-        return config$editEnabled;
+      isAvailable: function(config) {
+        try {
+          return ! config.update || ! config.update.f  ? true : config.update.f();
```
kgrgreer, line 112: Why is this needed? Why can't you just always call config.update.f()?

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -122,8 +126,12 @@ foam.CLASS({
     },
     {
       name: 'delete',
-      isAvailable: function(config$deleteEnabled) {
-        return config$deleteEnabled;
+      isAvailable: function(config) {
+        try {
+          return ! config.delete || ! config.delete.f ? true : config.delete.f();
```
kgrgreer, line 131: same

src/foam/lib/PermissionPredicate.js
```diff
@@ -0,0 +1,55 @@
...
+  name: 'PermissionPredicate',
+  extends: 'foam.mlang.predicate.AbstractPredicate',
+  implements: ['foam.core.Serializable'],
+  imports: [
+    'auth',
+  ],
+  properties: [
+    {
+      name: 'args',
+      class: 'StringArray'
```
kgrgreer, line 18: Better to have only one permission and then AND or OR them together if that's what you need. Otherwise, it isn't clear what it means when you have multiple permissions.

src/foam/lib/PermissionPredicate.js
```diff
@@ -0,0 +1,55 @@
...
+    'foam.nanos.auth.AuthService'
+  ],
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: `
+        X x = (X) obj;
+        AuthService auth = (AuthService) x.get("auth");
+        if ( args_ == null || args_.length <= 0 ) {
```
kgrgreer, line 32: If no authservice, then we're probably running in a demo or outside of nanos, in which case I think we should avoid a NPE and still allow.

src/foam/lib/PermissionPredicate.js
```diff
@@ -0,0 +1,55 @@
...
+          return true;
+        }
+
+        for ( String permission : args_ ) {
+          if ( auth.check(x, permission) )
+            return true;
+        }
+        return false;
+      `,
+      code: async function() {
```
kgrgreer, line 42: Does this work? Don't predicates need to return synchronously?

src/foam/comics/v2/DAOSummaryView.js
jlhughes: yes, this was before each method was changed to a Predicate.  And the issue is no default value of a TRUE predicate. 
return config.update.f()

src/foam/comics/v2/DAOControllerConfig.js
jlhughes: @olhabn revert to create/edit/deletedEnabled. 
and add default Predicate of TRUE.

src/foam/lib/PermissionPredicate.js
jlhughes: so need to use slots, default to false, then update when the call returns.

olhabn: > Does this work? Don't predicates need to return synchronously?

I believe it does. I'm not sure where I can find information about that

olhabn: > so need to use slots, default to false, then update when the call returns.

But other predicates return boolean

src/foam/lib/PermissionPredicate.js
olhabn: Could you please tell me what is NPE?

src/foam/lib/PermissionPredicate.js
olhabn: fixed https://github.com/foam-framework/foam2/pull/3234

src/foam/comics/v2/DAOSummaryView.js
olhabn: fixed https://github.com/foam-framework/foam2/pull/3234

src/foam/comics/v2/DAOSummaryView.js
olhabn: fixed https://github.com/foam-framework/foam2/pull/3234

src/foam/comics/v2/DAOControllerConfig.js
olhabn: fixed https://github.com/foam-framework/foam2/pull/3234

src/foam/comics/v2/DAOBrowseControllerView.js
olhabn: fixed https://github.com/foam-framework/foam2/pull/3234

PR #3228, opened 2020-02-26
Generate compareTo for each FObject

hchoura:


src/foam/core/AbstractFObject.js
```diff
@@ -107,31 +107,6 @@ foam.CLASS({
   // ],
 
   methods: [
-    {
```
kgrgreer: I would rather leave this here. For Android clients code size is more important than performance so we might like to disable creating the specific methods in that case.

src/foam/java/refinements.js
```diff
@@ -595,6 +595,19 @@ foam.LIB({
           args: [{ type: 'foam.core.X', name: 'x' }],
           body: 'setX(x);'
         });
+        
+        cls.method({
+          visibility: 'public',
+          name: 'compareTo',
+          type: 'int',
+          args:[{ name: 'o', type: 'Object' }], 
+          body: [cls.name+' o2 = ('+ cls.name + ') o;\n'
```
kgrgreer: I think a multi-line string with interpolation might be a little cleaner here.

src/foam/util/SafetyUtil.java
```diff
@@ -59,6 +60,42 @@ public static int compare(Object o1, Object o2) {
...
+    
+    return o1.compareTo(o2);
+  }
+
+  public static int compare(boolean o1, boolean o2) {
+    return o1 == o2 ? 0 : o1 ? 1 : 0;
+  }
+
+  public static int compare(String o1, String o2) {
+    return o1.equals(o2) ? 0 : o1 == null ? -1 : o2 == null ? 1 : o1.compareTo(o2);
```
kgrgreer: Check if o1 is null first. You check later, but you would have already have had a NPE by then.

src/foam/core/AbstractFObject.js
hchoura: but this will create reference to compare is ambiguous

PR #3230, opened 2020-02-26
Adding tracking request support

tharmaman:
Tracking Request is an ApprovalRequest with the approve and reject actions hidden. So basically if we want a user to be able to have an approval request for tracking purposes but not be able to action on them, we would set the tracking request property to true.

Also include an null case handler for grabbing a fresh copy of the data on the DAOSummaryView.

### DIFF
--- a/src/foam/nanos/approval/ApprovableAwareDAO.js
+++ b/src/foam/nanos/approval/ApprovableAwareDAO.js
@@ -135,8 +135,12 @@ foam.CLASS({
       }
 
       if ( getIsTrackingRequestSent() ){
-        sendSingleRequest(x, request, request.getInitiatingUser());
-        approverIds.remove(request.getInitiatingUser());
+        RoleApprovalRequest trackingRequest = (RoleApprovalRequest) request.fclone();
+        trackingRequest.setIsTrackingRequest(true);
+
+        sendSingleRequest(x, trackingRequest, trackingRequest.getInitiatingUser());
+
+        approverIds.remove(trackingRequest.getInitiatingUser());
       }
 
       for ( int i = 0; i < approverIds.size(); i++ ) {
--- a/src/foam/nanos/approval/ApprovableAwareDAO.js
+++ b/src/foam/nanos/approval/ApprovableAwareDAO.js
@@ -134,7 +134,7 @@ foam.CLASS({
         throw new RuntimeException("The only approver of " + modelName + " is the maker of this request!");
       }
 
-      if ( getIsTrackingRequestSent() ){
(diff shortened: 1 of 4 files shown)


kgrgreer:
LGTM

src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -135,8 +135,12 @@ foam.CLASS({
       }
 
       if ( getIsTrackingRequestSent() ){
```
jlhughes: formatting:
```suggestion
      if ( getIsTrackingRequestSent() ) {
```

src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -135,8 +135,12 @@ foam.CLASS({
...
-        sendSingleRequest(x, request, request.getInitiatingUser());
-        approverIds.remove(request.getInitiatingUser());
+        RoleApprovalRequest trackingRequest = (RoleApprovalRequest) request.fclone();
```
jlhughes, line 138: Below ApprovalRequest tests tracking, but here we assume RoleApprovalRequest.

src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -134,9 +134,13 @@ foam.CLASS({
         throw new RuntimeException("The only approver of " + modelName + " is the maker of this request!");
       }
 
-      if ( getIsTrackingRequestSent() ){
-        sendSingleRequest(x, request, request.getInitiatingUser());
-        approverIds.remove(request.getInitiatingUser());
+      if ( getIsTrackingRequestSent() ) {
```
nanoscott, line 137: question: how is this ever false? It looks like it's never set anywhere .. so why even bother having this check - why not just send it all the time?

tharmaman: Property on ApprovableAware 'isTrackingRequestSent', user can configure it there will make more sense once we transform this into a rule.

src/foam/nanos/approval/ApprovableAwareDAO.js
tharmaman: RE: my comment about moving the extra properties isFulfilled and Operation to approvalRequest and changing initiatingUser to createdBy.

tharmaman: We can merge this in now so it fixes and ill do the migration in another PR @jlhughes ?

PR #3231, opened 2020-02-27
[CPF-2616] Add capabilityCategoryDAO

KernelDeimos:
This PR adds the `CapabilityCategory` model, `capabilityCategoryDAO`, and `capabilityCategoryCapabilityJunctionDAO`. Capability categories will be displayed in the capability store after a future PR.

### DIFF
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -89,6 +89,10 @@ foam.CLASS({
       class: 'StringArray'
     },
     {
+      name: 'category',
+      class: 'String'
+    },
+    {
       name: 'version',
       class: 'String'
     },
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -89,10 +89,6 @@ foam.CLASS({
       class: 'StringArray'
     },
     {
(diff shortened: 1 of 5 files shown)


src/foam/nanos/crunch/Capability.js
```diff
@@ -88,6 +88,10 @@ foam.CLASS({
       name: 'keywords',
       class: 'StringArray'
     },
+    {
```
kgrgreer: Why is this property needed if there's a *:* relationship between Capability and CapabilityCategory?

src/services
```diff
@@ -725,6 +725,22 @@ p({
   "client":"{\"of\":\"foam.nanos.pm.PM\", \"remoteListenerSupport\": false}"
 })
 
+p({
+  "class":"foam.nanos.boot.NSpec",
```
kgrgreer, line 729: Is this line needed?

src/foam/nanos/crunch/Capability.js
KernelDeimos: Oops, forgot to remove that

src/services
KernelDeimos: Seems to be - it disappears from services if I remove this line

PR #3233, opened 2020-02-27
add hashcode in each FObject.

hchoura:


src/foam/java/refinements.js
```diff
@@ -595,6 +595,19 @@ foam.LIB({
...
         });
+        
+        if ( cls.name != 'AbstractFObject' ) {
+          cls.method({
+            visibility: 'public',
+            name: 'hashCode',
+            type: 'int',
+            body: 
+              ['int hash = 1'].concat(props.map(function(f) {
+                return 'hash += hash * 31 + java.util.Objects.hash('+foam.String.constantize(f.name)+'.get(this))';
```
kgrgreer: This will convert primitive types like numbers and booleans into objects, generating garbage, and will make this much slower than it needs to be. This is what we're trying to avoid by generating a specific method. Instead, add type specific hash methods to SafetyUtil so we can generate hashCode without converting to Object.

PR #3235, opened 2020-02-27
Detailed Action Boolean View

zahnno:
Detailed Action Boolean View provides boolean values a view that dynamically changes labels depending on the boolean state. An additional action is provided to allow updates to the data provided and put to the defined targetDAO. Boolean value labels and actions labels are provided when defining the view at a property level.

Ex: 
```javascript
{
  class: 'Boolean',
  name: 'myBoolean',
  view: {
    class: 'foam.u2.DetailedActionBooleanView',
    trueLabel: 'this label will display when prop value is true',
    falseLabel: 'this label will display when prop value is false',
    trueActionLabel: 'switch to false',
    falseActionLabel: 'switch to true',
    targetDAOKey: 'myDAO'
  }
}
``` 
To omit the action and display a checkbox that doesn't not update the data object, you can define showAction as false. This will update the property value but will require a put to a desired DAO.

Below is a screenshot of the view (Ablii  theme applied):

Screen Shot 2020-02-27 at 2 58 29 PM

Action is aware of display mode 'RW' (Connect theme applied):

Screen Shot 2020-02-27 at 2 57 19 PM

Setting autoSave to true on view will put the object to the targetDAO on action click. Otherwise the action will set the property value to a new value.

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -428,6 +428,7 @@ FOAM_FILES([
   { name: "foam/u2/IntView", flags: ['web'] },
   { name: "foam/u2/FloatView", flags: ['web'] },
   { name: "foam/u2/CurrencyView", flags: ['web'] },
+  { name: "foam/u2/DetailedActionBooleanView", flags: ['web'] },
   { name: "foam/u2/CheckBox", flags: ['web'] },
   { name: "foam/u2/md/CheckBox", flags: ['web'] },
   { name: "foam/u2/CitationView", flags: ['web'] },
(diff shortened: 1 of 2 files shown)


kgrgreer:
@jlhughes Do you think this should be in FOAM or maybe a u2 package inside nanopay?

jlhughes:
I see this as nanopay specific UI behaviour.

jlhughes:
replaced by https://github.com/nanoPayinc/NANOPAY/pull/9447

PR #3241, opened 2020-02-29
PMs to FileJournal to capture replay time.

jlhughes:
Also 
* clean up duplicate PM service entries.

### DIFF
--- a/src/foam/nanos/boot/Boot.java
+++ b/src/foam/nanos/boot/Boot.java
@@ -17,10 +17,8 @@ import foam.nanos.auth.User;
 import foam.nanos.logger.Logger;
 import foam.nanos.logger.ProxyLogger;
 import foam.nanos.logger.StdoutLogger;
-import foam.nanos.pm.DAOPMLogger;
 import foam.nanos.pm.NullPM;
 import foam.nanos.pm.PM;
-import foam.nanos.pm.PMLogger;
 import foam.nanos.script.Script;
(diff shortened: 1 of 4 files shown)


src/foam/nanos/boot/Boot.java
```diff
@@ -42,6 +44,8 @@ public Boot() {
   public Boot(String datadir) {
     Logger logger = new ProxyLogger(new StdoutLogger());
     root_.put("logger", logger);
+    root_.put("pmInfoDAO", new foam.nanos.pm.PMInfoMDAO());
```
kgrgreer: This shouldn't be necessary.  I would rather find out why this isn't working and fix it than hard-code PM's into Boot. Maybe we need a localPMInfoDAO which isn't authenticated or served, which is used by the PMLogger. Would be faster as well.

kgrgreer (changes requested):
Mostly good, but try creating a localPMInfoDAO and don't hard-code PM's into Boot.

src/foam/nanos/boot/Boot.java
jlhughes: The real fix, which was hidden my other troubleshooting was explicitly creating the PM rather than rely on the PM factory which just generates 1% of non null PMs.

jlhughes:
Issue was the PM factory.  Using explicit PM creation resolved the replay capture.

PR #3248, opened 2020-03-02
DO NOT MERGE - Fixing local DAO lookup on the client

nanoscott:


nanoscott:
@tharmaman .. found this bug .. can you review?

nanoscott:
Not needed.

PR #3249, opened 2020-03-02
Adding a specific hash methods to SafetyUtil.

hchoura:
The main reason to avoid converting object.


src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
     return o1 == o2 ? 0 : o1 < o2 ? -1 : 1;
   }
 
+  public static int hash(Object o1) {
+    return java.util.Objects.hash(o1);
+  }
+
+  public static int hash(FObject o1) {
```
kgrgreer: Is this correct? Don't we just want to call the hashCode() method on the object that we generated?

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
...
+    return java.util.Objects.hash(o1);
+  }
+
+  public static int hash(boolean o1) {
+    return java.util.Objects.hash(o1);
```
kgrgreer: This will convert it to an Object, which is what we were trying to avoid. Something like this:
return o1 ? 1 : 0;

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
...
+  }
+
+  public static int hash(String o1) {
+    return java.util.Objects.hash(o1);
```
kgrgreer: Just return o1.hashCode(); or to be safe:
return o1 == null ? 0 : o1.hashCode();

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
...
+  }
+
+  public static int hash(short o1) {
+    return java.util.Objects.hash(o1);
```
kgrgreer: return o1;

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
...
+  public static int hash(int o1) {
+    return java.util.Objects.hash(o1);
+  }
+
+  public static int hash(long o1) {
+    return java.util.Objects.hash(o1);
+  }
+
+  public static int hash(float o1) {
+    return java.util.Objects.hash(o1);
```
kgrgreer: Something like
return Float.toIntBits(o1);

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
...
+  }
+
+  public static int hash(double o1) {
+    return java.util.Objects.hash(o1);
```
kgrgreer: Use Double.doubleToRawLongBits(double value)

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
...
+  public static int hashCode(int o1) {
+    return o1;
+  }
+
+  public static int hashCode(long o1) {
+    return java.util.Objects.hashCode(o1);
+  }
+
+  public static int hashCode(float o1) {
+    return Float.floatToIntBits(o1);
```
kgrgreer, line 148: Something like
return Float.toIntBits(o1);

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
     return o1 == o2 ? 0 : o1 < o2 ? -1 : 1;
   }
 
+  public static int hashCode(Object o1) {
+    return java.util.Objects.hash(o1);
```
kgrgreer: Faster to do:
return o1 == null ? 0 : o1.hashCode();

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
...
+  }
+
+  public static int hashCode(FObject o1) {
+    return java.util.Objects.hash(o1);
```
kgrgreer: Faster to do:
return o1 == null ? 0 : o1.hashCode();

src/foam/util/SafetyUtil.java
```diff
@@ -100,6 +100,42 @@ public static int compare(double o1, double o2) {
...
+  public static int hashCode(short o1) {
+    return o1;
+  }
+
+  public static int hashCode(int o1) {
+    return o1;
+  }
+
+  public static int hashCode(long o1) {
+    return java.util.Objects.hashCode(o1);
```
kgrgreer: Something like this would be faster:
(int)(o1()^(o1>>>32))

kgrgreer: Not resolved!

src/foam/util/SafetyUtil.java
kgrgreer: Objects.hash() creates an array which generates garbage and is slower.

src/foam/util/SafetyUtil.java
kgrgreer: Objects.hash() creates an array which generates garbage and is slower.

PR #3253, opened 2020-03-03
Added object id specific permissions for permission predicate

tharmaman:
Basically allow handling a permission defined in the PermissionPredicate such as the following example.

If I want the permission checked for each id individually for each object such as:
"account.make.1234",
"account.make.1",
"account.make.456"
...
"account.make.921"

All I have to pass into the PermissionPredicate is just the first two portions of the permission string with a '%' character at the end instead of the actual id
"account.make.%"

### DIFF
--- a/src/foam/lib/PermissionPredicate.js
+++ b/src/foam/lib/PermissionPredicate.js
@@ -27,9 +27,8 @@ foam.CLASS({
     {
       name: 'f',
       javaCode: `
-        X x = (X) obj;
-        AuthService auth = (AuthService) x.get("auth");
-        return auth.check(x, permission_);
+        // java side is only needed for rules which is handled in the RulePermissionPredicate
+        throw new RuntimeException( new UnsupportedOperationException() );
       `,
       code: async function(objId) {    
         /**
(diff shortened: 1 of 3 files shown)


src/foam/lib/PermissionPredicate.js
```diff
@@ -31,7 +31,15 @@ foam.CLASS({
         AuthService auth = (AuthService) x.get("auth");
         return auth.check(x, permission_);
       `,
-      code: async function() {
+      code: async function(objId) {
```
jlhughes, line 33: @tharmaman only client side?

src/foam/lib/PermissionPredicate.js
```diff
@@ -27,11 +27,18 @@ foam.CLASS({
...
       name: 'f',
       javaCode: `
-        X x = (X) obj;
-        AuthService auth = (AuthService) x.get("auth");
-        return auth.check(x, permission_);
+        // java side is only needed for rules which is handled in the RulePermissionPredicate
+        throw new RuntimeException( new UnsupportedOperationException() );
       `,
-      code: async function() {
+      code: async function(objId) {
```
kgrgreer, line 33: Does this work? Can predicates be async?

tharmaman: Yeah it works. However, I can change it to a simple thenner instead since theres no looping of promises, don't really need async/await in this case

PR #3257, opened 2020-03-04
adding a redundancy to setting csv string

Nauna:


mcarcaso:
Why is this necessary? It's going to cause a StringBuilder to get built on every single put on the backend. I'm not sure you want this. My guess is that something isn't calling `eof`. You might want to hunt down that instead of doing this.

PR #3260, opened 2020-03-04
PropertyHasValue rule predicate

zahnno:
Implemented a PropertyHasValue rule predicate to make defining has predicates on rules easier.
Predicate will check if prop has a value.

Example of use within rule entry:
```javascript
  "predicate": {
    "class":"foam.mlang.predicate.And",
    "args":[
      {
        "class":"foam.nanos.ruler.predicate.PropertyHasValue",
        "propName":"paymentId"
      },
      {
        "class":"foam.mlang.predicate.Not",
        "arg1": {
          "class":"foam.nanos.ruler.predicate.PropertyHasValue",
          "propName":"paymentId",
          "isNew":false
        }
      }
    ]
  },
```

### DIFF
--- a/src/foam/nanos/nanos.js
+++ b/src/foam/nanos/nanos.js
@@ -249,6 +249,7 @@ FOAM_FILES([
   { name: "foam/nanos/ruler/predicate/PropertyEQProperty" },
   { name: "foam/nanos/ruler/Relationships" },
   { name: "foam/nanos/ruler/action/SendNotification" },
+  { name: "foam/nanos/ruler/predicate/PropertyHasValue" },
   { name: "foam/nanos/ruler/predicate/PropertyEQValue" },
   { name: "foam/nanos/ruler/predicate/PropertyNEQValue" },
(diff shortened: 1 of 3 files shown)


nanoMichal:
You can achieve the same thing with the PropertyNEQValue predicate, although you would need separate checks for null, '',[], undefined. based on what type of property you are checking it shouldn't be an issue though.
` {
 "class":"foam.nanos.ruler.predicate.PropertyNEQValue",
 "propName":"paymentId",
 "propValue":null,
 "isNew":false,
      },`

zahnno:
@nanoMichal I considered using that method but figured having to assume the empty value of the property seemed like a bit more effort. The Has predicate takes care of comparing all empty values.

https://github.com/foam-framework/foam2/blob/b4a7fe5ef9b8355dd7be6af7854ac195abbd7243/src/foam/mlang/mlang.js#L2071-L2090

jlhughes (approved):

PR #3263, opened 2020-03-05
Checking for lifecycle aware in ApprovableAwareDAO

nanoscott:


src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -154,6 +154,11 @@ foam.CLASS({
       User user = (User) x.get("user");
       Logger logger = (Logger) x.get("logger");
 
+      if ( ! ( obj instanceof LifecycleAware ) )
```
jlhughes, line 154: Can you move this before user and logger?

PR #3265, opened 2020-03-06
Approvable Aware easyDAO

tharmaman:


src/foam/dao/EasyDAO.js
```diff
@@ -228,6 +228,23 @@ foam.CLASS({
...
+          .setDelegate(delegate)
+          .build();
+
+        if ( getApprovableAware() && ! getApprovableAwareEnabled() )
+          delegate = new foam.nanos.approval.ApprovableAwareDAO
+          .Builder(getX())
+          .setDaoKey(getName())
+          .setOf(getOf())
+          .setDelegate(delegate)
+          .setIsEnabled(false)
```
jlhughes: Can you combine these and just set isEnabled(getApprovableAwareEnabled())?

src/foam/dao/EasyDAO.js
```diff
@@ -228,6 +228,23 @@ foam.CLASS({
         if ( getLastModifiedByAware() )
           delegate = new foam.nanos.auth.LastModifiedByAwareDAO.Builder(getX()).setDelegate(delegate).build();
 
+        if ( getApprovableAware() && getApprovableAwareEnabled() )
```
jlhughes: The logic below, if getApprovableAwareEnabled is false because interface decorators are false then getApprovableAware will also be false.  So the second setup below doesn't make sense.  
I added the comment below before this one.

tharmaman: Done.

src/foam/dao/EasyDAO.js
tharmaman: Done.

nanoscott (approved):

jlhughes (approved):

PR #3274, opened 2020-03-09
easter-egg: Add Y-Combinator to stdlib

KernelDeimos:


kgrgreer (changes requested):
I reject this change.

KernelDeimos:
It was worth a shot

PR #3288, opened 2020-03-13
[DO_NOT_MERGE] [CPF-3038] Add require() to AuthService

KernelDeimos:
In development of Capability intercepts for CRUNCH, I realized that there is currently no way for an AuthService in the chain to add new information about the failed authorization check. This is because `check()` returns a boolean rather than throwing an exception.

If AuthService throws an exception instead of returning a boolean, an AuthService in the chain can add additional information to the AuthorizationException as it falls down. In the case of CRUNCH, a list of capabilities that can grant the permission can be added to AuthorizationException.

### DIFF
--- a/src/foam/nanos/auth/AbstractAuthService.js
+++ b/src/foam/nanos/auth/AbstractAuthService.js
@@ -1,0 +1,21 @@
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'AbstractAuthService',
+  abstract: true,
+
+  implements: [
+    'foam.nanos.auth.AuthService'
+  ],
+  
+  methods: [
+    {
+      name: 'require',
+      javaCode: `
+        if ( ! check(x, permission) ) {
+          if ( exception != null ) throw exception;
+          throw new AuthorizationException();
+        }
+      `
+    }
+  ]
+});
--- a/src/foam/nanos/auth/ProxyAuthService.js
+++ b/src/foam/nanos/auth/ProxyAuthService.js
@@ -28,5 +28,17 @@ foam.CLASS({
       of: 'foam.nanos.auth.AuthService',
       name: 'delegate'
     }
+  ],
+
+  methods: [
+    {
+      name: 'require',
+      javaCode: `
+        if ( ! check(x, permission) ) {
+          if ( exception != null ) throw exception;
+          throw new AuthorizationException();
+        }
+      `
+    }
   ]
 });
--- a/src/foam/nanos/auth/AuthService.js
+++ b/src/foam/nanos/auth/AuthService.js
@@ -151,6 +151,8 @@ foam.INTERFACE({
       name: 'check',
       documentation: `
         Check if a user in the given context has the given permission.
+        The require method should be used instead if the intent is to throw an
+        exception.
       `,
       async: true,
       type: 'Boolean',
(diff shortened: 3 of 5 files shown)


src/foam/nanos/auth/AbstractAuthService.js
```diff
@@ -0,0 +1,21 @@
...
+    'foam.nanos.auth.AuthService'
+  ],
+  
+  methods: [
+    {
+      name: 'require',
+      javaCode: `
+        if ( ! check(x, permission) ) {
+          if ( exception != null ) throw exception;
+          throw new AuthorizationException();
```
nanoscott, line 16: Why not add the permission that was checked into the exception? It doesn't have to be in the message, but if we know the failing permission it may help debug things in the future.

src/foam/nanos/auth/ProxyAuthService.js
```diff
@@ -28,5 +28,17 @@ foam.CLASS({
...
     }
+  ],
+
+  methods: [
+    {
+      name: 'require',
+      javaCode: `
+        if ( ! check(x, permission) ) {
+          if ( exception != null ) throw exception;
+          throw new AuthorizationException();
```
nanoscott, line 39: add permission?

jlhughes (commented):
Can we subclass of AuthenticationException which can carry this additional data:
* permission
* any other crunch/capability data we need - even modeled

src/foam/nanos/auth/AbstractAuthService.js
KernelDeimos: I think this is a good idea. I don't think it should be sent to the client, but if the property is transient then it doesn't matter.

The same thing could be accomplished with an AuthService decorator at the end of the chain that adds the permission too - what do you think of this approach?

nanoscott: Sounds good to me .. @jlhughes added some more comments on this, so make sure to sync up on those, which sounds like it will incorporate this into it as well.

KernelDeimos:
> Can we subclass of AuthenticationException which can carry this additional data:
> 
> * permission
> * any other crunch/capability data we need - even modeled

In this design I would say an AuthService in the chain (delegate of CapabilityAuthService) would turn AuthorizationException into PermissionAuthorizationException, then CapabilityAuthService may replace the PermissionAuthorizationException with CapabilityOptionsAuthorizationException.

This makes sense - if you have a list of capabilities, you no longer have the permission; the permission doesn't matter anymore.

Although if we did want the permission to still be known even if there are corresponding capabilities the property `permission` might need to be in AuthorizationException itself (either that or CapabillityOptionsAuthorizationException extends PermissionAuthorizationException)

jlhughes:
AuthorizationException is ours, so we can modify/add to it.
we call check(permission) then generate the AuthorizationException.
AuthorizationException should have a new property (java) 'permission' which carries the permission that  'check' checked.   Then the current call flow doesn't need to change - to use 'required'.  
Additionally can have composite or compound Exceptions allow us to capture more than one Auth exception.  
And yes, the Auth exceptions can be caught and transformed into Capability exceptions 

The required permission should not be the 'message' of the AuthorizationException - as this may be reported.

KernelDeimos:
Added `DO_NOT_MERGE` because it actually _does_ need to be `check()` that throws an exception, so a migration of every line of code calling `auth.check` can't be avoided this way.

Joël suggested a solution that simplifies capability intercepts for now and doesn't stop us from doing this migration in the future - the PR for that should be ready soon.

PR #3287, opened 2020-03-13
Cpf 4805/add google sheets export

olhabn:
@jlhughes @zahnno 

Google Sheets export added. Please take a look

PR #3291, opened 2020-03-15
Np 46/implement pdf export support

olhabn:
@kgrgreer @zahnno @jlhughes 
GoogleSheets and PDFGoogleSheets added
Please ignore https://github.com/foam-framework/foam2/pull/3287


src/foam/nanos/export/ExportDriverRegistry.js
```diff
@@ -26,6 +26,18 @@ foam.CLASS({
       class: 'String',
       name: 'mimeType',
       value: 'text/csv;charset=utf-8'
+    },
+    {
+      class: 'Boolean',
+      name: 'canBeConverted'
```
jlhughes: I would like to go with isConvertible, isDownloadable, and isOpenable. 
Unless it causes complications in the ExportModel below.

src/foam/nanos/export/GoogleApiAuthService.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.api.client.util.store.FileDataStoreFactory;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.List;
+
+public class GoogleApiAuthService {
+  private static final String CREDENTIALS_FOLDER = "NANOPAY_HOME";
+  private static final String CREDENTIALS_FILE = "credentials.json";
```
jlhughes: How do the credentials work?

jlhughes (commented):
Looks good.

src/foam/nanos/export/GoogleApiAuthService.java
olhabn: We download file with credentials from google console, and then use them for retrieving access token for api, which we registered with google console.

jlhughes: 'download' - It looks like it downloads into NANOPAY_HOME - does this work when running the -u option? running from jar file?

olhabn: No, I did it manually. I wasn't sure what's the best way to automate it. Any suggestions?

src/foam/nanos/export/ExportDriverRegistry.js
olhabn: Fixed.
Thank you

src/foam/nanos/export/GoogleApiAuthService.java
olhabn: @jlhughes Never mind, I need a couple hours to figure it out

jlhughes: so /tmp is available everywhere. but I it is possible to use the browser db or file system?

olhabn: @jlhughes 
NANOPAY_HOME changed to /tmp.
Hm, but browser db is used for storing client data if I'm not mistaken, while credentials.json is a server side data. Please let me know if I misunderstood.
I've read https://cloud.google.com/docs/authentication/production, and I still believe we need to store credentials somewhere locally

jlhughes: ok, you can read/write to the 'var' directory.  If you look at the BMO planners/processors you'll see them managing temporary files in opt/nanopay/var.    
But in general we need a way to configure credentials on a running app - and then you can write the credentials.json locally if needed so it can be picked up the google library.

src/foam/nanos/export/GoogleApiCredentials.js
```diff
@@ -0,0 +1,46 @@
+foam.CLASS({
```
jlhughes, line 7: license header.

jlhughes (commented):
re: credentials - excellent.

src/foam/nanos/export/GoogleApiCredentials.js
olhabn: Sorry about that.
Fixed

jlhughes (approved):

PR #3296, opened 2020-03-18
Fix approval request setIsFulfilled update

nanoscott:
Approval requests can't be approved because the code to setIsFulfilled is causing an authorization exception. We need to elevate to the system context in order to complete this update.

### DIFF
--- a/src/foam/nanos/approval/ApprovableAwareDAO.js
+++ b/src/foam/nanos/approval/ApprovableAwareDAO.js
@@ -207,7 +207,7 @@ foam.CLASS({
           ApprovalRequest fulfilledRequest = (ApprovalRequest) approvedObjRemoveRequests.get(0);
           fulfilledRequest.setIsFulfilled(true);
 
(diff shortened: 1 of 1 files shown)


src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -207,7 +207,7 @@ foam.CLASS({
           ApprovalRequest fulfilledRequest = (ApprovalRequest) approvedObjRemoveRequests.get(0);
           fulfilledRequest.setIsFulfilled(true);
 
-          approvalRequestDAO.put_(x, fulfilledRequest);
+          approvalRequestDAO.put_(getX(), fulfilledRequest);
```
jlhughes: @nanoscott does this affect lastModifiedBy? 
I don't see away around it if it does. 
The change of context is appropriate.

noodlemoodle:
lgtm

nanoscott:
@jlhughes .. updated to set the proper user on the context.

jlhughes (approved):

PR #3297, opened 2020-03-19
Cpf 4219/change addcontactsteaptwo choiceview to country instead bankaccount

yij793:
address: https://nanopay.atlassian.net/browse/CPF-4818
other Nanopay PR:  https://github.com/nanoPayinc/NANOPAY/pull/9533

notes:  we want custom the model label but we don't want the origin model label.  We can't change the label since we using strategizer(which will taking model directly). So I made this changes can custom the label for those labels that are taking from model directly

### DIFF
--- a/src/foam/strategy/StrategyReference.js
+++ b/src/foam/strategy/StrategyReference.js
@@ -70,7 +70,7 @@ foam.CLASS({
     {
       class: 'String',
       name: 'label',
-      documentation: 'allow we set the StrategyReference display'
+      documentation: 'Override the default model label for Strategizer Choice selections.'
     }
   ]
 });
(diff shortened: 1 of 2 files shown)


jlhughes (commented):
Can we not add this support in the stategizer on the StrategyReference, then the 
tested and used as:   sr.strategy.label || sr.strategy.model_.label

src/foam/strategy/StrategyReference.js
```diff
@@ -66,6 +66,11 @@ foam.CLASS({
         'Outputter' and target = 'Transaction', then the special outputter would
         be included in the results.
       `
+    },
+    {
+      class: 'String',
+      name: 'label',
+      documentation: 'allow we set the StrategyReference display'
```
jlhughes:
```suggestion
      documentation: 'Override the default model label for Strategizer Choice selections.'
```

jlhughes (approved):

PR #3299, opened 2020-03-19
Validation export

zahnno:
Interface providing model an action to export validation on its properties.

Validation outputs:
- Property
- ValidationPredicate Error Strings
- Min and Max string length
- Min and Max num length


jlhughes:
Can we not do this with Axiom inspection?

kgrgreer (changes requested):
Let's meet to discuss.

PR #3301, opened 2020-03-19
[NP-107] Add CheckUserAuthService

KernelDeimos:
This fixes a vulnerability that becomes active when an AuthService implementor other than UserAndGroupAuthService grants permissions.

If a user doesn't have permission to check permissions for other users (`service.auth.checkUser`), they are still able to provided the permission was granted by an AuthService implementor other than UserAndGroupAuthService.

### DIFF
--- a/src/foam/nanos/auth/CheckUserAuthService.js
+++ b/src/foam/nanos/auth/CheckUserAuthService.js
@@ -1,0 +1,27 @@
+foam.CLASS({
+  package: 'foam.nanos.auth',
+  name: 'CheckUserAuthService',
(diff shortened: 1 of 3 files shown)


kgrgreer:
What is this needed for?

src/foam/nanos/auth/CheckUserAuthService.js
```diff
@@ -0,0 +1,27 @@
+foam.CLASS({
```
jlhughes, line 1: header.

lchanmann:
Would checking user permission via crunch be more costly than via user and group, why don't we inquire via crunch last?

PR #3303, opened 2020-03-20
Fixing outputter

olhabn:


src/foam/nanos/export/GoogleSheetsOutputter.js
```diff
@@ -89,8 +89,14 @@ foam.CLASS({
               propValues.push(obj[columnMethadata[i].columnName].toString().substring(0, 8));
               columnMethadata[i].perValuePatternSpecificValues.push(obj[columnMethadata[i].columnName].toString().substring(8));
             }
-            else
-              propValues.push(obj[columnMethadata[i].columnName].toString());
+            else {
```
kgrgreer: Why not and 'else if' instead of an 'else' and then an 'if'?

src/foam/nanos/export/GoogleSheetsOutputter.js
```diff
@@ -89,8 +89,14 @@ foam.CLASS({
...
+              if( typeof obj[columnMethadata[i].columnName] === "function" )
```
kgrgreer: space after 'if'.
If this is the case, won't it always be the case and then couldn't we just not include this column?

src/foam/nanos/export/GoogleSheetsOutputter.js
olhabn: Fixed

src/foam/nanos/export/GoogleSheetsOutputter.js
olhabn: Fixed

PR #3306, opened 2020-03-23
Fix import not working

nanoscott:


src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -553,9 +553,9 @@
              return false;
         }
         
-        var objId = self.ctrl.__subContext__[self.daoKey_].of.ID.type === 'Long' ? parseInt(this.objId) : this.objId;
+        var objId = ctrl.__subContext__[self.daoKey_].of.ID.type === 'Long' ? parseInt(self.objId) : self.objId;
```
kgrgreer: This tightly couples this class to the controller, which could change. Isn't the context passed as an argument to this method? Use that instead. Also, this seems fragile, checking for the type to be Long. What about other number types? A better approach would be to try to call the .adapt() method on the ID property, then it would work no matter the type.

nanoscott: @kgrgreer .. X is not a valid parameter for isAvailable .. what I really need here is for the import of ctrl to work. Do you know why the import of ctrl doesn't work for detailed view but works for the table view?

nanoscott: we actually need the opposite of adapt here .. adapt makes thing all the same type, what we want here is to make either a Long or a String for the two different primary key types that we use. If we were to use adapt, we would be getting the same object type back.

PR #3309, opened 2020-03-23
Add $set and $remove for Map properties, and fix similar methods for Array

KernelDeimos:


src/foam/core/types.js
```diff
@@ -596,6 +596,34 @@ foam.CLASS({
...
+  ],
+
+  methods: [
+    function installInProto(proto) {
+      this.SUPER(proto);
+      var self = this;
+      Object.defineProperty(proto, self.name + '$set', {
+        get: function mapSet() {
+          return function (k, v) {
+            console.log('uhh', this);
```
kgrgreer: remove

PR #3311, opened 2020-03-23
Np 94/adding drivers configuration

olhabn:
@kgrgreer @jlhughes 
Please take a look.
Thank you

Ticket:
https://nanopay.atlassian.net/browse/NP-94


olhabn:
fixing it

PR #3312, opened 2020-03-23
Fix map property methods

KernelDeimos:
Fixes genjava error caused by redefining a non-configurable property.

The `installInProto` method for properties is called more than once on the same `proto`. This caused genjava to fail because I didn't add `configurable: true` to the `$set` and `$remove` property methods for `Map`.


kgrgreer:
Did you fix the root cause, or are you just working around it?

kgrgreer:
I've merged this temporarily, but we should get to the root cause instead.

PR #3317, opened 2020-03-24
combined axioms

KolombetM:
https://github.com/nanoPayinc/NANOPAY/pull/9551 is failing because of ``Axiom name conflict``

### DIFF
--- a/src/foam/blob/Blob.js
+++ b/src/foam/blob/Blob.js
@@ -377,15 +377,12 @@ foam.CLASS({
   ],
 
   methods: [
-    function compareTo(other) {
-      return foam.blob.IdentifiedBlob.isInstance(other) && other.id == this.id;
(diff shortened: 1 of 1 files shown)


src/foam/blob/Blob.js
```diff
@@ -401,7 +398,10 @@ foam.CLASS({
         if ( o2 == null ) return 1;
         if ( o2 == this ) return 0;
         return foam.util.SafetyUtil.compare(getId(), o2.getId());
-      `
+      `,
+      code: function compareTo(other) {
```
jlhughes: believe this should be 
```suggestion
      code: function(other) {
```

jlhughes (changes requested):

jlhughes (approved):

PR #3318, opened 2020-03-24
Fixing compareTo method

nanoscott:


jlhughes:
See - already fixed https://github.com/foam-framework/foam2/pull/3317

src/foam/blob/Blob.js
```diff
@@ -396,6 +392,9 @@ foam.CLASS({
             type: 'Object',
           }
         ],
+      code: function compareTo(other) {
+        return foam.blob.IdentifiedBlob.isInstance(other) && other.id == this.id;
```
kgrgreer, line 396: How can this be correct? compareTo should return -1, 0, or 1, but this is returning a boolean.

PR #3320, opened 2020-03-24
[NP-368] Fix genjava paths and add better logs

KernelDeimos:
A PR that makes it possible to build without `-c` again

### DIFF
--- a/tools/genjava.js
+++ b/tools/genjava.js
@@ -21,7 +21,7 @@ if ( foamlinkMode ) {
 }
 
 var logger = {};
-logger.debug = (...args) => {};
+logger.debug = () => {};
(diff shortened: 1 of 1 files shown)


tools/genjava.js
```diff
@@ -18,10 +20,18 @@ if ( foamlinkMode ) {
   global.FOAMLINK_DATA = process.env['FOAMLINK_DATA'];
 }
 
+var logger = {};
+logger.debug = (...args) => {};
```
kgrgreer: What is the "...args" needed for?

KernelDeimos: It looks pretty. PR updated

PR #3322, opened 2020-03-25
[CPF-157] Medusa

jlhughes:
Notes on use are on the matching nanopay branch PR

Non-Medusa Changes
* boot script "main" moved into own journal so it can be instance specific. 
...


src/foam/box/MessageReplyBox.js
```diff
@@ -0,0 +1,35 @@
+/**
```
jlhughes: To be removed.

src/foam/box/SessionServerBox.java
```diff
@@ -42,29 +45,82 @@ public SessionServerBox(X x, Box delegate, boolean authenticate) {
...
+          StringTokenizer st = new StringTokenizer(authorization);
+          if ( st.hasMoreTokens() ) {
+            String authType = st.nextToken();
+            if ( HTTPAuthorizationType.BEARER.getName().equalsIgnoreCase(authType) ) {
+              if ( st.hasMoreTokens() ) {
+                sessionID = st.nextToken();
+                if ( sessionID != null ) {
+                  session = (Session) sessionDAO.find(sessionID);
+                  if ( session == null ) {
+                    // test and use non-clustered medusa sessions
```
jlhughes: NOTE: logging in with Load-Balancers has not been tested yet.

src/foam/box/socket/BoxHolder.js
```diff
@@ -0,0 +1,22 @@
+/**
```
jlhughes, line 1: Used by SocketConnectionBox

src/foam/box/SessionServerBox.java
```diff
@@ -42,29 +45,82 @@ public SessionServerBox(X x, Box delegate, boolean authenticate) {
...
+                    DAO internalSessionDAO = (DAO) getX().get("internalSessionDAO");
+                    if ( internalSessionDAO != null ) {
+                      session = (Session) internalSessionDAO.find(sessionID);
```
nanoscott: question: how big of a security hole is this? if an external actor got ahold of these internal session IDs, what permissions would they have?

follow-up question: if they could get access to these internal sesssion IDs, then it probably means they already have access to the system in some deep way .. so does the above question matter? Probably only if the internal session IDs are constant or otherwise discoverable.

src/foam/box/socket/SocketConnectionBox.js
```diff
@@ -0,0 +1,272 @@
...
+    }
+  ],
+
+  methods: [
+    {
+      name: 'send',
+      javaCode: `
+      PM pm = PM.create(getX(), this.getClass().getSimpleName(), getHost()+":"+getPort()+":send");
+      Box replyBox = (Box) msg.getAttributes().get("replyBox");
+      if ( replyBox != null ) {
```
nanoscott, line 166: why does it seem like the code below should only be executed if replyBox == null, not replyBox != null? what is happening when the replyBox is null?

src/foam/box/socket/SocketServerProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+          p.setStartTime(new java.util.Date(sent));
+          p.log(x);
+
+          pm = PM.create(x, pmKey, pmName+":execute");
+
+          int length = in_.readInt();
+          byte[] bytes = new byte[length];
+          StringBuilder data = new StringBuilder();
+          int total = 0;
+          while ( true ) {
```
nanoscott, line 93: is this the same code as in the SocketConnectionBox?

src/foam/box/SessionServerBox.java
jlhughes: In general, yes, more discussion needed here.  
The above is just so I can login before the cluster is running.
The real session/token issue at the socket connections between mediators and nodes. 
And this where we need to make use of non-static, rolling, hsm, ....  identifiers. 
For testing they all have known GUIDs.

src/foam/box/socket/SocketConnectionBox.js
jlhughes: When replyBox is null, it's like send and forget. 
When replyBox is not null, we need to keep track of it, as the SocketConnectionBox is replacing the ReplyBox with its own.  When the response is received, we lookup the original reply box and call it.

src/foam/box/socket/SocketServerProcessor.java
jlhughes: yes, the connection box is waiting on a reply just like the server for a request. 
We could abstract into a InputSocketReader or something.

src/foam/nanos/crunch/cronjobs
```diff
@@ -16,7 +16,7 @@ p({
     cron.execute(x);
     print("done");
   """,
-  "enabled":true
+  "enabled":false
```
jlhughes: Disable during medusa testing as we have a midnight cron bug - run repeated in the midnight hour.

nanoscott (approved):
LGTM .. there are some TODOs in the code and some minor formatting errors, but of all the code it seems sound. There was some stuff that I didn't follow, but I went in depth in the consensus algorithm, the voting algorithm and the ClusterConfigSupport code. I reviewed the Script/ScriptEvent updates as well.

src/foam/nanos/alarming/AlarmLoggingDAO.js
```diff
@@ -0,0 +1,45 @@
...
+  ],
+
+  methods: [
+    {
+      name: 'put_',
+      javaCode: `
+      Alarm alarm = (Alarm) getDelegate().put_(x, obj);
+      Logger logger = (Logger) x.get("logger");
+      switch ( alarm.getSeverity().getName() ) {
+        case "DEBUG":
```
lchanmann: Can the switch case be on the enum instead of the name?
```suggestion
      switch ( alarm.getSeverity() ) {
        case DEBUG:
```

src/foam/nanos/auth/AuthorizationDAO.js
```diff
@@ -59,7 +58,7 @@ foam.CLASS({
    if ( obj == null ) throw new RuntimeException("Cannot put null.");
 
     Object id = obj.getProperty("id");
-    FObject oldObj = getDelegate().inX(x).find(id);
+    FObject oldObj = id == null ? null : getDelegate().inX(x).find(id);
```
lchanmann: Just curious, do we check for id = null to avoid going through `find`? for a model using Long for the id property it won't be null.

src/foam/nanos/medusa/ClusterClientDAO.js
```diff
@@ -0,0 +1,247 @@
...
+      of: 'foam.nanos.boot.NSpec'
+    },
+    {
+      name: 'serviceName',
+      class: 'String',
+      javaFactory: `
+      if ( getNSpec() != null ) {
+        return getNSpec().getName();
+      }
+      return serviceName_;
```
lchanmann: String property factory in FOAM falls back to empty string, `return serviceName_` could make it null, is it intended?

src/foam/box/SessionServerBox.java
jlhughes: admin login is only supported on explicit instance.

src/foam/nanos/medusa/ClusterConfigSupport.js
```diff
@@ -0,0 +1,1051 @@
...
+          dao = dao.orderBy(ClusterConfig.PING_TIME);
+        }
+        List<ClusterConfig> configs = ((ArraySink) dao.select(new ArraySink())).getArray();
+        if ( configs.size() > 0 ) {
+          // return configs.get(0);
+          // return configs.get(configs.size() -1);
+          ClusterConfig cfg = configs.get(0);
+          getLogger().info("nextZone", "configs", configs.size(), "selected", cfg.getId(), cfg.getZone(), cfg.getIsPrimary(), cfg.getPingTime());
+          for ( ClusterConfig c : configs ) {
+            getLogger().info("nextZone", "other", cfg.getId(), cfg.getZone(), cfg.getIsPrimary(), cfg.getPingTime());
```
lchanmann: wondering should it log `c` instead of `cfg`?
```suggestion
          for ( ClusterConfig c : configs ) {
            getLogger().info("nextZone", "other", c.getId(), c.getZone(), c.getIsPrimary(), c.getPingTime());
```

src/foam/nanos/medusa/ClusterConfigMonitor.js
```diff
@@ -0,0 +1,210 @@
...
+          timer.purge();
+          getLogger().debug("timer", "cancel");
+          return;
+        }
+
+        // TODO: Non-Mediators just useful for reporting and network graph - the ping time could be reduced - see mn/services.jrl
+
+        // update ReplayingInfo for self.
+        DAO dao = (DAO) x.get("clusterConfigDAO");
+        dao.find_(x, config.getId());
```
lchanmann:
```suggestion
        DAO dao = (DAO) x.get("clusterConfigDAO");
        config = (ClusterConfig) dao.find_(x, config.getId());
```
Might not need to reload the config in case the id and realm were unchanging.

src/foam/nanos/medusa/ClusterConfigMonitor.js
```diff
@@ -0,0 +1,210 @@
...
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.medusa',
+  name: 'ClusterConfigMonitor',
+
+  documentation: 'NOTE: do not start with cronjob. This process starts the ClusterConfigPingSing which polls the Mediators and Nodes and will initiate Replay, and Elections.',
```
lchanmann:
```suggestion
  documentation: 'NOTE: do not start with cronjob. This process starts the ClusterConfigMonitorAgents which polls the Mediators and Nodes and will initiate Replay, and Elections.',
```

src/foam/nanos/medusa/ClusterConfigReplayDAO.js
```diff
@@ -0,0 +1,188 @@
...
+        ClusterConfig myConfig = support.getConfig(x, support.getConfigId());
+
+        if ( support.getStandAlone() &&
+             nu.getType() == MedusaType.NODE ) {
+          getLogger().debug("standalone");
+          ReplayingInfo replaying = (ReplayingInfo) x.get("replayingInfo");
+          replaying.setStartTime(new java.util.Date());
+
+          DAO dao = (DAO) x.get("medusaNodeDAO");
+          Min min = (Min) MIN(MedusaEntry.INDEX);
```
lchanmann, line 75: min is unused?

src/foam/nanos/alarming/AlarmLoggingDAO.js
jlhughes: yes

src/foam/nanos/auth/AuthorizationDAO.js
jlhughes:
```suggestion
    FObject oldObj = getDelegate().inX(x).find(id);
```

src/foam/nanos/medusa/ClusterClientDAO.js
jlhughes: serviceName is no longer required. It was earlier on.  ClusterClientDAOs are created by MedusaAdapterDAOs and some time ago they didn't have NSpec.  And I was flipping between using an NSpec directly, or NSpec.name.

src/foam/nanos/medusa/ClusterConfigMonitor.js
jlhughes: See ClusterConfigReplayingInfoDAO - the find on self, is just to invoke this decorator which updates the config for network graph.

src/foam/nanos/medusa/ClusterConfigReplayDAO.js
jlhughes: Yes.  The incomplete intent is make a more focused replay request to the nodes. Presently max is ignored as well, and the nodes just replay everything.

src/foam/box/socket/SocketConnectionBox.js
```diff
@@ -0,0 +1,314 @@
...
+        formatter.setX(getX());
+        formatter.output(msg);
+        message = formatter.builder().toString();
+        byte[] messageBytes = message.getBytes(StandardCharsets.UTF_8);
+        Socket socket = (Socket) getSocket();
+        if ( socket.isClosed() ||
+             ! socket.isConnected() ) {
+          throw new SocketException("Socket not connected.");
+        }
+        // omLogger.log(this.getClass().getSimpleName(), getId(), "pending");
```
nanoscott: several log messages are commented out in this class .. either remove them or uncomment them

src/foam/box/socket/SocketConnectionBoxManager.js
```diff
@@ -0,0 +1,218 @@
...
+      name: 'remove',
+      args: [
+        {
+          name: 'box',
+          type: 'SocketConnectionBox'
+        }
+      ],
+      javaCode: `
+      if ( box != null ) {
+//        getBoxes().remove(makeKey(box.getHost(), box.getPort()));
```
nanoscott: remove dead code

src/foam/crypto/hash/HashableTest.js
```diff
@@ -28,15 +28,15 @@ foam.CLASS({
 
         // non-chained hash with correct digest
         Hashable_HashWithValidAlgorithm(input, "MD5",
-            "47f68c52aaae1c2cdad696c59a248752");
+            "d41d8cd98f00b204e9800998ecf8427e");
```
nanoscott: I just had to change these too, so they may need to be updated again.

src/foam/dao/EasyDAO.js
```diff
@@ -606,8 +612,17 @@ foam.CLASS({
       }
     },
     {
-      /** Simpler alternative than providing serverBox. */
+      documentation: 'Cluster this DAO',
+      name: 'cluster',
+      class: 'Boolean',
+      javaFactory: `
+    return "true".equals(System.getProperty("CLUSTER"));
```
nanoscott:
```suggestion
    return SafetyUtil.equals("true", System.getProperty("CLUSTER"));
```

src/foam/dao/EasyDAO.js
```diff
@@ -208,6 +233,12 @@ foam.CLASS({
           delegate = (ProxyDAO) getDecorator();
         }
 
+        // set inner delegate_ to handle reentrant
+        // DelegateFactory calls from subsequent DAOs which may
+        // have init_ methods which in turn call getDelegate().
+        delegateIsSet_ = true;
+        delegate_ = new ProxyDAO.Builder(getX()).setDelegate(delegate).build();
```
nanoscott, line 240: this seems a bit concerning that the code within the factory calls getDelegate on EasyDAO before the factory is finished ... is the reentrance unavoidable?

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -144,6 +144,9 @@ foam.CLASS({
         if ( x.get(Session.class) == null ) return false;
         if ( user == null || ! user.getEnabled() ) return false;
         User realUser = ((Subject) x.get("subject")).getRealUser();
+        if ( realUser == null ) {
```
nanoscott: might be a merging artifact .. doesn't seem like a clustering change to me.

src/foam/dao/EasyDAO.js
jlhughes: The reentrant call would need to receive something else we'll deadlock.  Not sure how to avoid this.

src/foam/box/socket/SocketConnectionBoxManager.js
```diff
@@ -0,0 +1,218 @@
...
+          return box;
+        }
+
+        try {
+          Socket socket = new Socket();
+          socket.setSoTimeout(getSoTimeout());
+          SocketAddress address = new InetSocketAddress(host, port);
+          socket.connect(address, getConnectTimeout());
+          box = new SocketConnectionBox(x, key, socket, host, port);
+          add(box);
```
kgrgreer, line 158: The 'boxes' collection is synchronized but there's no sync at this level to prevent two from being created. Is that going to be a problem?

src/foam/box/socket/SocketServer.js
```diff
@@ -0,0 +1,111 @@
...
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.box.socket',
+  name: 'SocketServer',
+
+  documentation: `Waits on the socket connection for requests, passing them off to a SocketServerProcessor.`,
```
kgrgreer: no need for backquote.

src/foam/nanos/medusa/CheckWebAgent.java
```diff
@@ -0,0 +1,61 @@
...
+import foam.nanos.http.WebAgent;
+import javax.servlet.http.HttpServletResponse;
+import java.io.PrintWriter;
+
+/**
+   Health check web agent, intented to be called by load balancers
+   to add/remove targets from target group.
+   Responses from haproxy.org
+ */
+public class CheckWebAgent
```
kgrgreer: Perhaps a more descriptive name would be helpful. 'Check' is kind of vague.

src/foam/box/socket/SocketConnectionBoxManager.js
jlhughes: The model method is synchronized: 
`    {
      name: 'get',
      synchronized: true,
      type: 'foam.box.Box',
`

src/foam/nanos/medusa/CheckWebAgent.java
jlhughes: Renamed to HealthCheckWebAgent.

PR #3323, opened 2020-03-25
Basically moved the approvableAwareDAO after GUID and SEQNO

tharmaman:
Basically moved the approvableAwareDAO after GUID and SEQNO instead of just the GUID


nanoscott:
Isn't this the same as my change? https://github.com/foam-framework/foam2/pull/3321/files

nanoscott:
Ah, looks like I didn't move it back far enough.

nanoscott (approved):

PR #3326, opened 2020-03-26
Added a delete modal and redirecting to error screen with invalid token

minsunny:
this needs to be merged with https://github.com/nanoPayinc/NANOPAY/pull/9573

https://nanopay.atlassian.net/browse/NP-194
https://nanopay.atlassian.net/browse/NP-195#
Changed to delete both an invitation and its associated token not just set a new status

Screen Shot 2020-03-26 at 12 10 36 PM
Screen Shot 2020-03-26 at 12 35 30 PM


zahnno (changes requested):
This has references to nanopay. Why aren't we passing in the businessInvitationDAO into this modal on nanopay side?

minsunny:
closed this since no necessary need a change on foam side

PR #3331, opened 2020-03-27
Add Contact Redesign Section Changes

nanopay-arthur:
Relating to nanoPayinc/NANOPAY#9463

SectionDetailPropertView.js Change: Allows visibility: 'HIDDEN' sectioned properties to still be accessible.

Section.js Change: Will support filtration of unwanted inherited sections. Ex. this.sections = this.sections.filter((section) => section.fromClass === 'Contact');

### DIFF
--- a/src/foam/layout/Section.js
+++ b/src/foam/layout/Section.js
@@ -48,7 +48,8 @@ foam.CLASS({
     },
     {
       class: 'String',
-      name: 'fromClass'
+      name: 'fromClass',
+      documentation: 'The class name to which the section belongs to.'
     }
   ],
 
--- a/src/foam/layout/Section.js
+++ b/src/foam/layout/Section.js
@@ -45,6 +45,10 @@ foam.CLASS({
       value: function(data$) {
         return foam.core.ConstantSlot.create({value: true});
       }
+    },
+    {
+      class: 'String',
+      name: 'fromClass'
     }
   ],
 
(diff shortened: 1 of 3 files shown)


src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -226,7 +226,6 @@ foam.CLASS({
       this
         .addClass(this.myClass())
         .add(this.slot(function(mode, prop, prop$label) {
-          if ( mode === self.DisplayMode.HIDDEN ) return null;
```
zahnno, line 229: Won't removing this cause the labels and tooltips of the property to appear?

src/foam/layout/Section.js
```diff
@@ -45,6 +45,10 @@ foam.CLASS({
       value: function(data$) {
         return foam.core.ConstantSlot.create({value: true});
       }
+    },
+    {
+      class: 'String',
+      name: 'fromClass'
```
kgrgreer: Please add documentation:.

src/foam/u2/detail/SectionedDetailPropertyView.js
nanopay-arthur: I just tested out adding label and help to a sectioned property that has visibility: 'HIDDEN' set, and nothing from the property is displayed.

zahnno (approved):

PR #3333, opened 2020-03-27
add permission in UML diagram

hchoura:


src/foam/doc/DocBrowser.js
```diff
@@ -801,6 +807,10 @@ foam.CLASS({
       var nbrOfPropInNonConventionalDiag = 5;
       var propertyHeight = 20;
       this.className  = this.data.name;
+
+      if ( Array.isArray(this.allowedModels) && this.allowedModels.length && this.allowedModels.includes(this.className))
```
kgrgreer, line 811: Would be faster if these were maps instead of arrays. The adapt methods on the properties could convert from arrays to maps internally.

src/foam/doc/DocBrowser.js
```diff
@@ -793,6 +793,12 @@ foam.CLASS({
       value: 0,
       documentation: 'the y of the last required to element draw in the canvas.',
     },
+    {
+      name: 'allowedModels',
```
kgrgreer: No trailing commas.

kgrgreer:
Should this PR closed? Did you redo in another already merged PR?

PR #3336, opened 2020-03-30
Avoid to decorate the sink if it's a Count or GroupBy.

hchoura:


src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -56,8 +56,13 @@ foam.CLASS({
       javaCode: `
       if ( x.get("auth") != null ) {
         if ( predicate != null ) predicate.authorize(x);
-        foam.dao.Sink sink2 = ( sink != null ) ? new HidePropertiesSink(x, sink, this) : sink;
-        super.select_(x, sink2, skip, limit, order, predicate);
+          // don't decorate the sink if it's a Count or GroupBy
+          if ( ! ( sink instanceof foam.mlang.sink.Count || sink instanceof foam.mlang.sink.GroupBy) ) {
```
kgrgreer: Only skip if its a GroupBy(Count()), if it were a GroupBy() of something else, then we still decorate.

src/foam/nanos/auth/PermissionedPropertyDAO.js
```diff
@@ -56,8 +56,16 @@ foam.CLASS({
       javaCode: `
       if ( x.get("auth") != null ) {
         if ( predicate != null ) predicate.authorize(x);
-        foam.dao.Sink sink2 = ( sink != null ) ? new HidePropertiesSink(x, sink, this) : sink;
-        super.select_(x, sink2, skip, limit, order, predicate);
+        // don't decorate the sink if it's a Count or GroupBy(Count())
+        if ( ! ( sink instanceof foam.mlang.sink.Count || 
+            ((sink instanceof foam.mlang.sink.GroupBy) && 
+            ((foam.mlang.sink.GroupBy)sink).getArg2() instanceof foam.mlang.sink.Count && 
+            ((foam.mlang.sink.Count) ((foam.mlang.sink.GroupBy)sink).getArg2()).getLabel().equals("Count") )) ) {
```
kgrgreer: What is this line for?

hchoura: if it is count or GroupBy(exp,sink) and the sink 'arg2' is Count, we Don't need to decorate the DAO.

kgrgreer: Why do we do getLabel().equals("Count") here but do  instanceof foam.mlang.sink.Count above? Couldn't you use an instanceof in both cases? Also, import you did javaImports: for these classes it would make this much shorter and more readable.

PR #3346, opened 2020-04-03
Setting ports

olhabn:
@jlhughes @kgrgreer 
Please review

### DIFF
--- a/src/foam/nanos/export/GoogleApiCredentials.js
+++ b/src/foam/nanos/export/GoogleApiCredentials.js
@@ -42,7 +42,8 @@ foam.CLASS({
     },
     {
       class: 'Int',
-      name: 'port'
+      name: 'port',
+      visibility: 'RO'
     },
     {
       class: 'String',
--- a/src/foam/nanos/export/GoogleApiCredentials.js
+++ b/src/foam/nanos/export/GoogleApiCredentials.js
@@ -43,7 +43,8 @@ foam.CLASS({
     {
       class: 'Int',
       name: 'port',
-      visibility: 'RO'
+      hidden: true,
+      documentation: 'Port server uses for communication with authorization service'
     },
     {
       class: 'String',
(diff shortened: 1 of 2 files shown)


src/foam/nanos/export/GoogleApiCredentials.js
```diff
@@ -42,7 +42,8 @@ foam.CLASS({
     },
     {
       class: 'Int',
-      name: 'port'
+      name: 'port',
+      visibility: 'RO'
```
kgrgreer: Why is this RO?

src/services
```diff
@@ -1015,7 +1015,9 @@ p({
   "name":"googleApiCredentialsConfig",
   "lazy":false,
   "service": {
-    "class":"foam.nanos.export.GoogleApiCredentials"
+    "class":"foam.nanos.export.GoogleApiCredentials",
+    "port":64349,
```
kgrgreer, line 1019: Where does this number come from?

olhabn: random number we're agreed on

src/foam/nanos/export/GoogleApiCredentials.js
olhabn: I thought that's a good idea to make sure that no one will change port

kgrgreer: Do we need to even show it then? Could we make it hidden: true. Does anyone need to know this number? Could you add documentation:?

src/services
kgrgreer: Who is "we"? How did we agree on it? Can we document?

olhabn: I talked to Joel when I tried to make changes for staging so I think we agreed to use this one

src/foam/nanos/export/GoogleApiCredentials.js
olhabn: Fixed

PR #3348, opened 2020-04-06
add object[] compare

nanoNeel:
https://nanopay.atlassian.net/browse/NP-551
Transaction line items are not being written out to the journal if they are the only property updated.
Before this change it would go to `compare(Object o1, Object o2)` and always return 0.

### DIFF
--- a/src/foam/util/SafetyUtil.java
+++ b/src/foam/util/SafetyUtil.java
@@ -60,6 +60,12 @@ public class SafetyUtil {
     return ((Comparable) o1).compareTo(o2);
   }
 
+  public static int compare(Object[] o1, Object[] o2) {
+    if ( o2 == null ) return  1;
+    if ( o1 == null ) return -1;
+    return compare(o1.length, o2.length);
+  }
+
   public static int compare(FObject o1, FObject o2) {
     if ( o1 == o2   ) return  0;
     if ( o2 == null ) return  1;


src/foam/util/SafetyUtil.java
```diff
@@ -60,6 +60,12 @@ public static int compare(Object o1, Object o2) {
     return ((Comparable) o1).compareTo(o2);
   }
 
+  public static int compare(Object[] o1, Object[] o2) {
+    if ( o2 == null ) return  1;
+    if ( o1 == null ) return -1;
+    return compare(o1.length, o2.length);
```
kgrgreer, line 66: Shouldn't we also compare their contents?

jlhughes: @kgrgreer comparing contents only works if the order is guaranteed.

kgrgreer:
I would think that [1,2,3] is different than [3,2,1]. Comparing the
contents isn't going to break your transfers use-case.

jlhughes:
@kgrgreer The missing part of the conversation was
- the safetyutil will compare length
- and the model can implement compare and do any additional model specific comparisons that it requires. 
so @nanoNeel updated compare and has another PR for Transaction specific compare.  Since in most cases you'll need model specific compare to intelligently deal with arrays, this SafetyUtil change is not critical if we choose not to add it.

PR #3353, opened 2020-04-06
[NP-515] DocuSign Integration

KernelDeimos:
Currently this is in a state of partial completion. This is reflected in the documentation and there is a note for contributors.

What works:
- OAuth2 flow for DocuSign
- Storing DocuSign session keys
- Generating sanitized documents in DocuSign "HTML"
- Sending envelope requests for embedded signing flow
  (to the user in context, from a designated user)

What's missing
- Sending documents from context user to another user on the system
- DocuSign Connect handler to verify signing has completed
- Scheduled task for refreshing DocuSignSession tokens
- Allow user to switch from their default DocuSign account


KernelDeimos:
Oh, almost forgot to mention, this also adds alias support to PropertyParser.

jlhughes:
Closing until needed.

PR #3360, opened 2020-04-08
Enum Permissioning

blakegreen32:
Implemented permissioning of Enums through the EnumView. Added a 'permissioned' flag to the EnumView, if set to true the EnumView will perform permission checks on all the enum values returning a filtered ChoiceView. To setup permissioning just create enum.read.value permissions, assign them to the groups in the groupPermissionJunctions.jrl and make a custom view for the enum property being rendered setting 'permissioned: true'


src/foam/u2/EnumView.js
```diff
@@ -9,24 +9,80 @@ foam.CLASS({
...
+              hash[array[i]] = i;
+            }
+            this.of.VALUES.map((v) => {
+              var value = [v.label, true];
+              if ( hash.hasOwnProperty(value) ) {
+                values.push([v, v.label]);
+              }
+            });
+            return values;
+          } else {
```
kgrgreer: No need for an 'else' if you're returned in the 'if'.

src/foam/u2/EnumView.js
```diff
@@ -9,24 +9,80 @@ foam.CLASS({
...
+       value: false
+    },
+    {
+      name: 'permissionResults',
+      expression: async function(of, permissioned) {
+        if ( of ) {
+          if ( permissioned ) {
+            var results = [];
+            var model = of.name.toLowerCase();
+            for ( var i = 0; i < of.VALUES.length; i++ ) {
```
kgrgreer: spaces fore ;'s

src/foam/u2/EnumView.js
```diff
@@ -9,24 +9,80 @@ foam.CLASS({
...
+              var readPermission = model + '.read.' + of.VALUES[i].label.toLowerCase();
+              var permResult = await this.auth.check(null, readPermission);
+              results.push([of.VALUES[i].label, permResult]);
+            }
+            return results;
+          } else {
```
kgrgreer: No need for 'else'.

src/foam/u2/EnumView.js
```diff
@@ -9,24 +9,80 @@ foam.CLASS({
   name: 'EnumView',
   extends: 'foam.u2.view.ChoiceView',
 
+  imports: [
```
kgrgreer, line 12: Why is this an imports rather than just a property?

src/foam/u2/EnumView.js
```diff
@@ -9,24 +9,80 @@ foam.CLASS({
...
       name: 'of',
       required: true
     },
     {
-      name: 'choices',
-      expression: function(of) {
-        return of ? of.VALUES.map(function(v) { return [ v, v.label ]; }) : [];
+       class: 'Boolean',
+       name: 'permissioned',
+       value: false
```
kgrgreer: Redundant. False is already the default.

src/foam/u2/EnumView.js
```diff
@@ -9,24 +9,80 @@ foam.CLASS({
...
       if ( ! this.of ) this.of = p.of;
+    },
+
+    function permissionedChoices() {
+      this.permissionResults.then((array) => {
+        var values = [];
+        if ( this.of ) {
+          if ( this.permissioned ) {
+            var hash = {};
+            for ( var i = 0; i < array.length; i+=1 ) {
```
kgrgreer: spaces before ;'s

src/foam/u2/EnumView.js
blakegreen32: Was just following ESLint, do we usually have spaces before semicolons in for loops within foam2?

src/foam/u2/EnumView.js
```diff
@@ -9,24 +9,79 @@ foam.CLASS({
...
     },
+    {
+       class: 'Boolean',
+       name: 'permissioned'
+    },
+    {
+      name: 'permissionResults',
+      expression: async function(of, permissioned) {
+        var results = [];
+        if ( of ) {
```
kgrgreer: You could just do: if ( of && permissioned ) {

PR #3361, opened 2020-04-08
Np 561/write permission required lifecycle state foam

tharmaman:
As a cause of adding writePermissionRequired to lifecycleState properties, system or people with explicit lifecycleState write permissions should be able to change lifecycleState on objects.

This alleviates the need for the approver to have write permission to status on every model they can approve.

### DIFF
--- a/src/foam/nanos/approval/ApprovableAwareDAO.js
+++ b/src/foam/nanos/approval/ApprovableAwareDAO.js
@@ -173,7 +173,8 @@ foam.CLASS({
         if ( lifecycleObj.getLifecycleState() == LifecycleState.PENDING && currentObjectInDAO == null ){
           lifecycleObj.setLifecycleState(LifecycleState.ACTIVE);
         }
-        return super.put_(x,obj);
+        // only system/admin should have permission to change lifecycle state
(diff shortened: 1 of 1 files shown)


jlhughes:
Can you clarify.  The approver would have write permission, no? And if so why the need to change to system context?

jlhughes (approved):

src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -173,7 +173,8 @@ foam.CLASS({
         if ( lifecycleObj.getLifecycleState() == LifecycleState.PENDING && currentObjectInDAO == null ){
           lifecycleObj.setLifecycleState(LifecycleState.ACTIVE);
         }
-        return super.put_(x,obj);
+        // only system/admin should have permission to change lifecycle state
+        return super.put_(getX(), obj);
```
nanoscott, line 177: I don't think you want to change the context right in the middle of a put operation to the system context .. won't this affect all fo the operations that require the "user" or "agent" from the context, like CreatedBy/ModifiedBy?

PR #3365, opened 2020-04-11
Fixing thenby

olhabn:
@kgrgreer @jlhughes @zahnno 
Please review

https://nanopay.atlassian.net/browse/NP-638

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -2880,7 +2880,10 @@ foam.CLASS({
         // an equals of arg1.compare is falsy, which will then hit arg2
         return this.head.compare(o1, o2) || this.tail.compare(o1, o2);
       },
-      javaCode: `return getHead().compare(o1, o2) == 0 ? getTail().compare(o1, o2) : getHead().compare(o1, o2);`
+      javaCode: `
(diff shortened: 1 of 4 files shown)


src/foam/mlang/mlang.js
```diff
@@ -2867,28 +2868,37 @@ foam.CLASS({
...
-    function compare_(o1, o2) {
-      // an equals of arg1.compare is falsy, which will then hit arg2
-      return this.head.compare(o1, o2) || this.tail.compare(o1, o2);
+    {
+      name: 'compare',
+      code: function(o1, o2) {
+        // an equals of arg1.compare is falsy, which will then hit arg2
+        return this.head.compare(o1, o2) || this.tail.compare(o1, o2);
+      },
+      javaCode: `return getHead().compare(o1, o2) == 0 ? getTail().compare(o1, o2) : getHead().compare(o1, o2);`
```
kgrgreer: This is inefficient because it might have to call getHead().compare(o1, o2) twice. Better:
int ret = getHead().compare(o1, o2);
return ret == 0 ? getTail().compare(o1, o2) : ret;

olhabn: @kgrgreer fixed

PR #3367, opened 2020-04-13
implement authorizable

nanoNeel:
While fixing https://nanopay.atlassian.net/browse/NP-671, noticed that file owners could not view their own files.

### DIFF
--- a/src/foam/nanos/fs/File.js
+++ b/src/foam/nanos/fs/File.js
@@ -10,6 +10,16 @@ foam.CLASS({
 
   documentation: 'Represents a file',
 
+  implements: [
+    'foam.nanos.auth.Authorizable'
+  ],
+
+  javaImports: [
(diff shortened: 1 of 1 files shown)


src/foam/nanos/fs/File.js
```diff
@@ -59,5 +69,48 @@ foam.CLASS({
         return url;
       }
     }
+  ],
+
+  methods: [
+    {
+      name: 'authorizeOnCreate',
+      javaCode: `
+        AuthService auth = (AuthService) x.get("auth");
```
kgrgreer, line 78: Not really this Pr's issue, but wouldn't it make more sense if all of these authorizeOnXYZ methods were just given the AuthService as the second arg, since they always have to look it up as the first thing anyway?

kgrgreer:
LGTM

PR #3370, opened 2020-04-14
Approval Request View Reference Fix

blakegreen32:
In the Approval Request model we use a propety in the viewReference action called 'daoKey_' which is a formatted string that converts a daoKey string from 'localMyDAO' to 'myDAO' if the 'daoKey' property does not exist in the context. The previous implementation was breaking as it only took into account 'local' DAO strings. In the case of a user the DAO was 'bareUserDAO' and was being formatted to 'serDAO' which was breaking the viewReference action. I have now implemented functionality to format the string by removing the initial substring before the first capital letter in the daoKey. Which in turn would return the correct daoKey if it does not exist in the context.

Ticket: https://nanopay.atlassian.net/browse/NP-560

### DIFF
--- a/src/foam/nanos/approval/ApprovalRequest.js
+++ b/src/foam/nanos/approval/ApprovalRequest.js
@@ -390,15 +390,10 @@
         var X = this.ctrl.__subContext__;
 
         if ( ! X[key] ) {
-          // if DAO doesn't exist in context, change daoKey from localMyDAO
-          // (server-side) to myDAO (accessible on front-end)
-          for ( var i = 0 ; i < key.length ; i++ ) {
-            if ( key.charAt(i) == key.charAt(i).toUpperCase() ) {
-              key = key.substring(i);
-              break;
-            }
+          if ( key.startsWith('local') ) {
+            key = key.replace('local', '');
+            key = key.charAt(0).toLowerCase() + key.slice(1);
           }
-          key = key.charAt(0).toLowerCase() + key.slice(1);
         }
         return key;
       }
--- a/src/foam/nanos/approval/ApprovalRequest.js
+++ b/src/foam/nanos/approval/ApprovalRequest.js
@@ -388,11 +388,18 @@
       factory: function(o, n) {
         var key = this.daoKey;
         var X = this.ctrl.__subContext__;
+        
         // FIXME: change to a better implementation
         if ( ! X[key] ) {
           // if DAO doesn't exist in context, change daoKey from localMyDAO
           // (server-side) to myDAO (accessible on front-end)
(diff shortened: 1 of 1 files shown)


nanoscott (approved):

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -388,11 +388,17 @@
...
           // if DAO doesn't exist in context, change daoKey from localMyDAO
           // (server-side) to myDAO (accessible on front-end)
-          key = key.substring(5, 6).toLowerCase() + key.substring(6);
+          for ( var i = 0 ; i < key.length ; i++ ) {
+            if ( key.charAt(i) == key.charAt(i).toUpperCase() ) {
+              key = key.substring(i);
+              break;
+            }
+          }
+          key = key.charAt(0).toLowerCase() + key.slice(1);
```
lchanmann: if we have "afexBusinessDAO", would it become "businessDAO"?

blakegreen32: Thats a good point, if this dao did not exist in the context with the previous implementation it would return 'usinessDAO' which would be broken entirely. It was basically taking out the 'local' out of any dao not in the context by removing the first 5 characters from the string. I think when this was originally implemented the daoKey not in the context was supposed to always be a local dao, which in the case of the user model it was actually bareUserDAO. I'm open to other suggestions if someone has one.

blakegreen32: What I could do is check if the daoKey contains 'local' in the string and then perform the formatting, that still would not solve the issue of 'bareUserDAO' not being in the context and having to use 'userDAO' instead. I'm assuming afexBusinessDAO would be in the context, and i'm also assuming 'localUserDAO' is what was expected instead of 'bareUserDAO'.

lchanmann: The fact that we have "bareUserDAO" as daoKey in the approval request might because EasyDAO add ApprovableAwareDAO to it. @jlhughes @nanoscott can we skip ApprovableAware on bareUserDAO?

@blakegreen32 yes, you can check for startsWith 'local' then strip it out if still not found in the context then throw.

jlhughes: I would prefer to restrict to local (non-local) so test for local. Anything else we'll deal with explicitly.  We are slowly adding local for all daos, so they will be there going forward.

jlhughes: and regarding the usersDAOs. localUserDAO should have remained the base, and everything built on top of it, but it's a little late for that.

jlhughes (approved):

PR #3372, opened 2020-04-15
User Require Email Client Side

blakegreen32:
Added validation client side to the User model to ensure the user is prompted to enter an email address.

Ticket: https://nanopay.atlassian.net/browse/NP-663


jlhughes (changes requested):
This should be a nanopay requirement, not foam.

PR #3373, opened 2020-04-15
Rule Migration (Rule ApprovableAware + LocalRuleDAO)

tharmaman:


src/services
```diff
@@ -724,17 +724,45 @@ p({
...
+    {
+      "of": "foam.nanos.ruler.Rule"
+    }
+  """
+})
+
+p({
+  "class": "foam.nanos.boot.NSpec",
+  "name": "localRuleDAO",
+  "serve": true,
```
jlhughes: local daos should not be served.

jlhughes (changes requested):

noodlemoodle:
lgtm

src/services
tharmaman: fixed changed some other local daos to serve false too which were previously serve true

src/services
```diff
@@ -724,17 +724,44 @@ p({
...
   "serviceScript": """
+    dao = new foam.dao.EasyDAO.Builder(x)
+      .setEnableInterfaceDecorators(false)
+      .setOf(foam.nanos.ruler.Rule.getOwnClassInfo())
+      .setRuler(false)
+      .setInnerDAO(x.get("localRuleDAO"))
+      .build();
+
+    dao.addPropertyIndex(new foam.core.PropertyInfo[] { foam.nanos.ruler.Rule.NAME });
+    dao = new foam.dao.ValidatingDAO(x, dao);
```
lchanmann, line 735: Why not setValidated(true) on the easydao? or maybe not needed as the local one already has it.

src/services
```diff
@@ -724,17 +724,44 @@ p({
...
     dao = new foam.dao.EasyDAO.Builder(x)
       .setPm(true)
       .setGuid(true)
       .setOf(foam.nanos.ruler.Rule.getOwnClassInfo())
       .setJournalType(foam.dao.JournalType.SINGLE_JOURNAL)
       .setJournalName("rules")
+      .setContextualize(true)
+      .setValidated(true)
       .setRuler(false)
+      .setAuthorize(false)
```
lchanmann, line 760: Why do we not authorize the dao? Imagine, if you give me access to scriptDAO I can get localRuleDAO and I can change any rules.

kgrgreer: If we give you access to scriptDAO you can already do whatever you want.

kgrgreer: The point of localDAO's is that they're unauthorized to improve performance and data accessibility by the system.

PR #3375, opened 2020-04-15
Move User Validations to nanopay

blakegreen32:
Moved validateObj methods from foam.nanos.auth.User for properties firstName, lastName and email to the nanopay side.


jlhughes (approved):
@kgrgreer this was a result of discussion regarding email verification which I felt should be nanopay side.  And these follow.  The verification is application dependent.

PR #3379, opened 2020-04-16
Add Name Attributes to MenuToolBar and IconTextFieldView

nanopay-arthur:
Request from QA


src/foam/nanos/menu/MenuToolBar.js
```diff
@@ -127,11 +128,12 @@ foam.CLASS({
                 .addClass(self.myClass('option-icon-container'))
                 .start('img')
                   .addClass(self.myClass('option-icon'))
-                  .attr('src', menu.icon)
+                  .attrs({ 'src': menu.icon, 'name': self.myClass('option-icon') })
```
zahnno: Aren't we looking for the icon and labels to be unique to the menu items?

nanopay-arthur: @zahnno I was thinking about setting name to the menu label i.e. "label": "Search by Business Name", but it wouldn't be in camel case. Would that be an issue?

zahnno: No since QA needs a unique identifier for these inputs. I think menu id would be more appropriate.

zahnno (approved):

PR #3382, opened 2020-04-16
Improve mutations not allowed error

KernelDeimos:


src/foam/u2/Element.js
```diff
@@ -369,13 +376,34 @@ foam.CLASS({
...
     },
-    function onSetClass(cls, enabled) { throw new Error('Mutations not allowed in OUTPUT state.'); },
-    function onFocus(cls, enabled) { throw new Error('Mutations not allowed in OUTPUT state.'); },
-    function onAddListener(topic, listener) { throw new Error('Mutations not allowed in OUTPUT state.'); },
-    function onRemoveListener(topic, listener) { throw new Error('Mutations not allowed in OUTPUT state.'); },
-    function onSetStyle(key, value) { throw new Error('Mutations not allowed in OUTPUT state.'); },
-    function onSetAttr(key, value) { throw new Error('Mutations not allowed in OUTPUT state.'); },
-    function onRemoveAttr(key) { throw new Error('Mutations not allowed in OUTPUT state.'); },
+    function onSetClass(cls, enabled) {
+      foam.error.thro('Mutations not allowed in OUTPUT state.',
```
kgrgreer: This seems odd to me. Firstly, it hurts my sensibilities that throw is spelled wrong, and secondly, why couldn't we just create our own Error/Exception class so we could use regular syntax:
  throw foam.Error('Mutations not allowed in OUTPUT state.', {cls:cls,enabled:enabled});
And to toString on our Error class does the formatting. Then, if nobody ever does a toString() on the exception, there's no overhead. So, if we want this capability, I think it is better if we extend Error rather than throw.

KernelDeimos: That makes sense to me, I'll update this PR later to fix this. I used `thro` just so that the syntax highlighter wouldn't get confused, but I like the foam.Error solution better

kgrgreer:
Let's wait for new modelled exception support.

PR #3385, opened 2020-04-19
Correct schedule for expireUserCapabilityJunctionsCron - should be da…

jlhughes:
…ily, but running every second in the 0 hour.


KernelDeimos (approved):
looks like it'll work to me

PR #3386, opened 2020-04-20
fixed noDuplicateApprovalRequests logic for Update

noodlemoodle:
- fix duplicate approvalrequests still being created for UPDATE
- rename interface methods:
      - getApprovableCreateKey -> getApprovableHashKey
      - getApprovableKey -> getStringId
- deprecated hashId calculated in approvableAwareDAO for UPDATE and used ApprovableHashKey instead
- changed approvableHashKey from hash of the map containing all props changed to a string containing the hash of each prop changed separated by colons

nano side pr : https://github.com/nanoPayinc/NANOPAY/pull/9713


src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -403,7 +403,10 @@ foam.CLASS({
           .where(
             foam.mlang.MLang.AND(
               foam.mlang.MLang.EQ(ApprovalRequest.DAO_KEY, "approvableDAO"),
-              foam.mlang.MLang.EQ(ApprovalRequest.OBJ_ID, hashedId),
+              foam.mlang.MLang.OR(
+                foam.mlang.MLang.EQ(ApprovalRequest.OBJ_ID, hashedId),
```
jlhughes: Why is the hashedId not modeled?

kgrgreer: Why does it need to be?

kgrgreer: On line 397 above, it is better to separate the hashId components with some symbol (like ':').

noodlemoodle: @jlhughes making some more changes to this after some discussion, so hashId will be deprecated by getApprovableCreateKey().
@kgrgreer , will add ":" to the approvablecreatekey hash

src/foam/nanos/approval/ApprovableAware.js
```diff
@@ -55,27 +55,30 @@
...
-                  if ( next.getValue() instanceof Object[] ) {
-                    arrayProps.add((String) next.getKey());
+                  Object nextValue = next.getValue();
+                  if ( nextValue instanceof Object[] ) {
+                    next.setValue(Arrays.asList((Object[]) nextValue));
                   }
-                }
-                for ( String prop : arrayProps ) {
-                  diff.put(prop, Arrays.asList((Object[]) diff.get(prop)).hashCode());
+                  hash += ":" + String.valueOf(next.hashCode());
```
kgrgreer: This is very expensive because it creates a new StringBuilder and String on each iteration. Use a StringBuilder instead.

kgrgreer (approved):

src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -382,28 +382,14 @@ foam.CLASS({
...
-        // change last modified by to be the user
-        updatedProperties.put("lastModifiedBy", user.getId());
-
-        // remove lastmodified prop to exclude it from hash
-        Object lastModified = updatedProperties.remove("lastModified");
-
+        String approvableHashKey = ApprovableAware.getApprovableHashKey(x, obj);
         DAO approvableDAO = (DAO) x.get("approvableDAO");
 
-        String daoKey = "d" + getDaoKey();
```
nanoscott, line 393: this is going to break the approval unit tests which rely on this key structure.

noodlemoodle: what's the name of the tests relying on this? I ran tests just now but didnt see anything break

nanoscott: ./build.sh -cjQT LiquidApprovalTest

nanoscott: it's this query .. the CONTAINS_IC (there are two of them):
```
    approvalRequests = (ArraySink) this.getApprovalRequestDAO(x).inX(this.getSystemX()).where(AND(
      CONTAINS_IC(ApprovalRequest.OBJ_ID, String.valueOf(user.getId())),
      EQ(ApprovalRequest.CLASSIFICATION, "User"),
      EQ(ApprovalRequest.OPERATION, operation),
      EQ(ApprovalRequest.IS_FULFILLED, false)
    )).select(new ArraySink());```

File: src/foam/nanos/approval/ApprovableAware.js [hunk 4]
```diff
@@ -55,27 +55,30 @@
                 logger.error("Error instantiating : ", obj.getClass().getSimpleName(), e);
               }
               Map diff = oldObj == null ? null : oldObj.diff(obj);
+              StringBuilder hash_sb = new StringBuilder(obj.getClass().getSimpleName());
```
nanoscott, line 60: you need to include the StringId here as well .. if you don't then the same change to two different records will look the same .. so say the same comment was added to two different records, then this would hash that the same and they would overwrite one another.

nanoscott (changes requested):

src/foam/nanos/approval/ApprovableAwareDAO.js
noodlemoodle: added back a simplified hashedid to account for this

src/foam/nanos/approval/ApprovableAware.js
noodlemoodle: Thanks, included for the update case since create will always have different id

src/foam/nanos/approval/ApprovableAware.js
```diff
@@ -55,27 +57,41 @@
...
-                    arrayProps.add((String) next.getKey());
+                  Object nextValue = next.getValue();
+                  if ( nextValue instanceof Object[] ) {
+                    next.setValue(Arrays.asList((Object[]) nextValue));
                   }
-                }
-                for ( String prop : arrayProps ) {
-                  diff.put(prop, Arrays.asList((Object[]) diff.get(prop)).hashCode());
+                  if ( nextValue instanceof FObject ) {
+                    String hashedKey = getApprovableHashKey(x, (FObject) nextValue, Operations.CREATE);
```
nanoscott: Why is this hardcoded to Operation.CREATE?
```suggestion
                    String hashedKey = getApprovableHashKey(x, (FObject) nextValue, operation);
```

nanoscott (approved):

PR #3388, opened 2020-04-20
[DO_NOT_MERGE] User View Reference Quick Fix

blakegreen32:
Added check for bareUserDAO in daoKey property of ApprovalRequest model to shorten it to the client side userDAO for the View Reference action to work.

This check can be removed when bug is fixed to remove ‘bare’ userDAO and keep ‘local’ as the lowest level.

### DIFF
--- a/src/foam/nanos/approval/ApprovalRequest.js
+++ b/src/foam/nanos/approval/ApprovalRequest.js
@@ -390,6 +390,9 @@
         var X = this.ctrl.__subContext__;
 
         if ( ! X[key] ) {
+          if ( key == 'bareUserDAO' ) {
+            key = 'userDAO';
+          }
           if ( key.startsWith('local') ) {
             key = key.replace('local', '');
             key = key.charAt(0).toLowerCase() + key.slice(1);
--- a/src/foam/nanos/approval/ApprovalRequest.js
+++ b/src/foam/nanos/approval/ApprovalRequest.js
@@ -390,8 +390,10 @@
         var X = this.ctrl.__subContext__;
 
         if ( ! X[key] ) {
-          if ( key == 'bareUserDAO' ) {
-            key = 'userDAO';
+          // TODO: Remove check for bareUserDAO when it is removed and replaced with localUserDAO
+          if ( key.startsWith('bare') ) {
+            key = key.replace('bare', '');
+            key = key.charAt(0).toLowerCase() + key.slice(1);
           }
           if ( key.startsWith('local') ) {
             key = key.replace('local', '');


src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -390,6 +390,9 @@
         var X = this.ctrl.__subContext__;
 
         if ( ! X[key] ) {
+          if ( key == 'bareUserDAO' ) {
+            key = 'userDAO';
```
jlhughes: Why can't the rule that generates the approvalRequest do this?

nanoscott: We use bareUserDAO on the server and have to keep it. We need to use userDAO on the client for proper auth and permissioning, hence the bug.

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -390,6 +390,9 @@
         var X = this.ctrl.__subContext__;
 
         if ( ! X[key] ) {
+          if ( key == 'bareUserDAO' ) {
```
nanoscott: can we replace this with the same logic as the 'local' removal below... so something like:
```suggestion
          if ( key.startsWith('bare') ) {
            key = key.replace('bare', '');
            key = key.charAt(0).toLowerCase() + key.slice(1);
```

nanoscott (changes requested):

src/foam/nanos/approval/ApprovalRequest.js
nanoscott: the other instance of the 'bare' DAO is the bareTransactionDAO .. so we could make this a pattern just like the 'local' pattern.

blakegreen32: Ya I can do that

jlhughes: bareTransactionDAO - where @$#@#$ did that come from?

src/foam/nanos/approval/ApprovalRequest.js
jlhughes: so the client uses the daokey to find the object for the reference view?

jlhughes:
This is such a huge hack. There has to be a better way.

nanoscott:
@jlhughes I don't get how the 'local' change is OK, but the 'bare' change is a huge hack. So it looks like the way that approvals work is that whichever DAO the approvals are enabled on is set as the daoKey in the ApprovalRequest entity. This is done automatically to allow us to resolve the entity that we need to approve and the DAO we need to approve it on. 

We could try to implement some sort of method on the ApprovalRequest which indicates the client-side DAO, but that would just be moving the logic here somewhere else ... I don't see a way around it unless we build it into the nSpec service definition or the EasyDAO itself. But this also increases coupling, though it makes the relationship explicit instead of implicit.

Presently, we already have the 'local' convention, and this is just an extension of that convention. We need to come up with a proper way to map these DAOs to each other. But I think that's a larger conversation and in the meantime, we need to fix this bug.

jlhughes:
see: https://nanopay.atlassian.net/browse/NP-183
the 'local' change will be reverted when NP-183 is complete.  

Until recently we only had local and non-local, and in this context, the test for 'local' seemed reasonable.  but these prefixes are proliferating (which I don't agree with) and any logic to deal with them.  So I'm drawing the line.   Any system that requires explicit knowledge of client/server conventions can model it and provide support in it's runtime configuration.

blakegreen32:
This was merged into nanopay/foam2: https://github.com/nanoPayinc/foam2/pull/13

PR #3391, opened 2020-04-21
NP-208 FIx permission name check

carl-zzz:
https://nanopay.atlassian.net/browse/NP-208

### DIFF
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -2333,7 +2333,7 @@ foam.CLASS({
           var auth = data.__subContext__.auth;
 
           var propName = this.name.toLowerCase();
-          var clsName  = this.forClass_.substring(this.forClass_.lastIndexOf('.') + 1).toLowerCase();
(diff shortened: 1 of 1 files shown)


src/foam/u2/Element.js
```diff
@@ -2333,7 +2333,7 @@ foam.CLASS({
           var auth = data.__subContext__.auth;
 
           var propName = this.name.toLowerCase();
-          var clsName  = this.forClass_.substring(this.forClass_.lastIndexOf('.') + 1).toLowerCase();
+          var clsName  = data.cls_.id.substring(data.cls_.id.lastIndexOf('.') + 1).toLowerCase();
```
lchanmann: Should we fallback to forClass_ in case data.cls_ is undefined or null?

kgrgreer: You don't have to do this string manipulation if you already have the cls_, you can just do data.cls_.name;

carl-zzz: @kgrgreer please confirm

PR #3392, opened 2020-04-21
Add receiveOnly property in Signup.js

Henry0422:


jlhughes (commented):
Anyway we can do this on the nanopay side? The concept of receiveOnly is specific to nanopay.

PR #3393, opened 2020-04-21
Support phone to Appconfig & Refactor

zahnno:
Add support phone number to group and appconfig.
Refactor and include support phone on getAppConfig on group.


jlhughes (approved):
This works, but was just thinking we could associate a userid with appconfig, and that user defines email, phone, address (if you need to post something), ...

jlhughes (approved):

PR #3394, opened 2020-04-21
Fix the UML diagram

hchoura:
add a limit for the models


src/foam/doc/ModelBrowser.js
```diff
@@ -39,7 +50,7 @@ foam.CLASS({
               a.array.map(function(nspec) {
                 return self.parseClientModel(nspec)
               }).filter(function(cls) {
-                return !!cls;
+                return !!cls && (allowedModels.size == undefined? true : allowedModels[cls.id]);
```
kgrgreer: The && will perform the same task as the !!, so those can be removed.

PR #3395, opened 2020-04-22
[CPF-3313] Refactor change password view

jinnnn9:
- addresses https://nanopay.atlassian.net/browse/NP-651 and https://nanopay.atlassian.net/browse/CPF-3313

- nano pr: https://github.com/nanoPayinc/NANOPAY/pull/9739


src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
```
kgrgreer: According to what I learned from Google lawyers, you shouldn't update copyright dates. The date should be from when the file was first created and then never updated. So, "Copyright 2019" really means "Copyright 2019 and beyond".

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,170 @@ foam.CLASS({
...
+    ^ ^section h2 {
+      margin-top: 0;
+      margin-bottom: 4vh;
+      font-size: 2.5rem;
     }
-    ^ .subtitle {
-      color: #525455;
-      font-size: 1em;
+    /* subtitle */
+    ^ ^section .subtitle {
```
kgrgreer, line 57: What does ^ ^section mean? Is the first ^ necessary?

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,170 @@ foam.CLASS({
...
-      class: 'String',
-      name: 'token',
-      documentation: `This property toggles the view from updating a user password to resetting a user password.`,
       factory: function() {
-        var searchParams = new URLSearchParams(location.search);
-        return searchParams.get('token');
-      },
-      hidden: true
+        // implement a logic to hide logo in certain places
+        return true;
```
kgrgreer: Is this in the factory? Couldn't you just use value:?

src/foam/nanos/auth/ChangePasswordView.js
jinnnn9: I wanted to give a higher specificity to this property so that it is not overwritten by other 'subtitle' property defined elsewhere. Without ^, its specificity would be lower than 'subtitle' property defined elsewhere.

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,170 @@ foam.CLASS({
...
+      class: 'String',
+      name: 'mode',
+      documentation: 'name of password model used for this view',
+      factory: function() {
+        if ( window.location.hash.includes('personal-settings') ||
+             window.location.hash.includes('set-personal') ) {
+          return 'update';
+        }
+        const token = new URLSearchParams(location.search).get('token');
+        if ( token ) {
```
kgrgreer: Better: return token ? 'reset' : 'send';

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,170 @@ foam.CLASS({
...
-      view: {
-        class: 'foam.u2.view.PasswordView',
-        passwordIcon: true
-      },
-      minLength: 6
+      class: 'String',
+      name: 'mode',
+      documentation: 'name of password model used for this view',
+      factory: function() {
+        if ( window.location.hash.includes('personal-settings') ||
```
kgrgreer: Why do we do this?

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,170 @@ foam.CLASS({
...
-              foam.nanos.auth.ChangePasswordView.NEW_PASSWORD,
-              foam.nanos.auth.ChangePasswordView.CONFIRMATION_PASSWORD);
-          },
-          errorString: 'Passwords do not match.'
+      name: 'passwordModel',
+      documentation: 'password model used for this view',
+      factory: function() {
+        if ( this.mode === 'update' ) {
+          return this.UpdatePassword.create();
+        } else if ( this.mode === 'reset' ) {
```
kgrgreer: No need for 'else' if you've already returned in the 'if'.

src/foam/nanos/auth/ChangePasswordView.js
jinn9: ^ ^section .subtitle would be like 'changePasswordView changePasswordview-section subtitle'.

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,170 @@ foam.CLASS({
...
-      code: function(value) {
+      name: 'resizeInputFields',
+      code: function(isHorizontal) {
         // evaluation of this gridColumns on a property does not handle an evaluation, only an int, thus function here managing this.
-        this.cls_.getAxiomByName('originalPassword').gridColumns = value ? ( this.token ? 0 : 4 ) : 12;
-        this.cls_.getAxiomByName('newPassword').gridColumns = value ? ( this.token ? 6 : 4 ) : 12;
-        this.cls_.getAxiomByName('confirmationPassword').gridColumns = value ? ( this.token ? 6 : 4 ) : 12;
+        if ( isHorizontal ) {
+          if ( this.mode === 'update' ) {
+            this.UpdatePassword.getAxiomByName('originalPassword').gridColumns = 4;
```
kgrgreer: Is this the same as this.UpdatePassword.ORIGINAL_PASSWORD?

src/foam/nanos/auth/ChangePasswordView.js
jinnnn9: Yeah, I could. I made a factory so that when we actually want to add implementation to hide a logo, we can just write the logic here.

jinnnn9: Should I change it to value?

src/foam/nanos/auth/ResetPassword.js
```diff
@@ -0,0 +1,119 @@
...
+  methods: [
+    {
+      name: 'reset_',
+      code: function() {
+        this.clearProperty('originalPassword');
+        this.clearProperty('newPassword');
+        this.clearProperty('confirmationPassword');
+        if ( this.token ) window.history.replaceState(null, null, window.location.origin + '/#reset');
+      }
+    },
```
kgrgreer: No trailing commas.

src/foam/nanos/auth/ResetPassword.js
```diff
@@ -0,0 +1,119 @@
...
+  ],
+
+  methods: [
+    {
+      name: 'reset_',
+      code: function() {
+        this.clearProperty('originalPassword');
+        this.clearProperty('newPassword');
+        this.clearProperty('confirmationPassword');
+        if ( this.token ) window.history.replaceState(null, null, window.location.origin + '/#reset');
```
kgrgreer: What is this for?

src/foam/nanos/auth/ChangePasswordView.js
jinnnn9: I am using the url of page the user is currently viewing to determine which password model to use. For instance, the url for updating password page would be https://ablii-domain/personal-setting, and the url for sending password page would be https://abli-domain

src/foam/nanos/auth/ResetPassword.js
jinnnn9: This is the old code someone else wrote, but it looks like it clears values assigned to the three properties when resetting.

src/foam/nanos/auth/UpdatePassword.js
```diff
@@ -0,0 +1,114 @@
...
+        }
+      ]
+    }
+  ],
+
+  methods: [
+    {
+      name: 'reset_',
+      code: function() {
+        this.clearProperty('originalPassword');
```
kgrgreer, line 99: FYI: You can also do this.originalPassword = undefined;

src/foam/nanos/auth/UpdatePassword.js
```diff
@@ -0,0 +1,114 @@
...
+        this.clearProperty('newPassword');
+        this.clearProperty('confirmationPassword');
+        if ( this.token ) window.history.replaceState(null, null, window.location.origin + '/#reset');
+      }
+    },
```
kgrgreer: No trailing commas.

src/foam/nanos/auth/ChangePasswordView.js
jinnnn9: If you are asking why we need 'mode' property, I am using it to determine which password model to use for this view. I could've just done everything inside 'passwordModel' property, but it looked more intuitive and cleaner.

zahnno: I'd prefer if we pass in the model then have the changePasswordView define what model to use - especially if it's checking a url hash - which limits our reusability. Let's say for example I wanted to use UpdatePassword outside of the personal-settings, I wouldn't be able to do that without extending this class and rewriting this factory function.

I think having something like: 
```javascript
...{ class: 'foam.nanos.auth.ChangePasswordView', passwordModel: 'foam.nanos.auth.UpdatePassword' }
```
When creating the view would make things a bit easier and flexible to use.

zahnno (changes requested):

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,166 @@ foam.CLASS({
...
-        this.cls_.getAxiomByName('newPassword').gridColumns = value ? ( this.token ? 6 : 4 ) : 12;
-        this.cls_.getAxiomByName('confirmationPassword').gridColumns = value ? ( this.token ? 6 : 4 ) : 12;
+        if ( isHorizontal ) {
+          if ( this.mode === 'update' ) {
+            this.UpdatePassword.ORIGINAL_PASSWORD.gridColumns = 4;
+            this.UpdatePassword.NEW_PASSWORD.gridColumns = 4;
+            this.UpdatePassword.CONFIRMATION_PASSWORD.gridColumns = 4;
+          } else if ( this.mode === 'reset' ) {
+            this.ResetPassword.NEW_PASSWORD.gridColumns = 6;
+            this.ResetPassword.CONFIRMATION_PASSWORD.gridColumns = 6;
```
zahnno: I think you'll be able to use a ternary operation here if you have the password model. 
```javascript
passwordModel.getAxiomByName('newPassword').gridColumns = passwordModel.cls_.name == 'UpdatePassword' ?  4 : 6;
```
This might not be completely correct but I hope it conveys the idea.

jinnnn9: For scalability, wouldn't it be better to separate the implementation for each mode? Currently, we only need to consider two modes (reset and update) so tertiary operators work better, but later on if we decide to add more modes that use password fields, the expression might not be so clear at first sight as we will need to chain the expression. 

What do you think?

Nauna: So now that we have the separate models for passwords, the gridColumn logic should be on the models. 

this view really should not care what model it is being given. It should be able to handle the models as they are.

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,166 @@ foam.CLASS({
...
-      validationPredicates: [
-        {
-          args: ['newPassword', 'confirmationPassword'],
-          predicateFactory: function(e) {
-            return e.EQ(
-              foam.nanos.auth.ChangePasswordView.NEW_PASSWORD,
-              foam.nanos.auth.ChangePasswordView.CONFIRMATION_PASSWORD);
-          },
-          errorString: 'Passwords do not match.'
+      name: 'passwordModel',
```
Nauna: add a view to this property - { class: 'foam.u2.detail.SectionedDetailView' } should do it, may need to add the of: this.passwordModel.cls_ ... and get rid of the sectionView in initE. 

Also don't think we need to add anything to the SectionView and should be centering things as needed here. We just have used and will continue to use the SectionView is various ways, the fact that it has been maintained in its simplicity is something I'd very much like to keep. And don't really think its necessary to adjust at this time.

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,166 @@ foam.CLASS({
...
+          .start().addClass(this.myClass('section'))
+            .start(this.SectionView, {
+              data: this.passwordModel,
+              sectionName: this.passwordModel.cls_.model_.sections[0].name,
+              isCentered: true
+            }).end()
+          .end()
+          // link
+          .callIf(this.mode === 'send', function() {
+            this.start().addClass(self.myClass('link'))
```
Nauna, line 144: couldn't we just make them all go back to login? and get rid of this callif??

Then I just remembered this is also used in the user management section of connect and ablii so don't really want to necessarily go back to the login. SOO what if you made something on the model of sendPassword ... that was called here and would be null in other models thus rendering nothing.

Thinking something like

SendPassword back function. So here you'd have
.callIf(this.passwordModel.backFunction, function() {
            this.start().addClass(self.myClass('link')).add(this.passwordModel.BACK_TXT).on(click,function() { this.passwordModel.backFunction();}


then in documents add that this view supports models with backFunction and BACK_TXT ... then if need to be adjusted in the future, the data being rendered on this view will all be stored on the model and this view only knows that it is receiving a model.

jinn9: For sending email and resetting password, it should be fine. When the user is updating their password , however, they need to be logged in so I thought it wouldn't be a good idea to redirect them to the login page after updating the page.

src/foam/nanos/auth/ChangePasswordView.js
```diff
@@ -9,267 +9,166 @@ foam.CLASS({
...
     'stack',
     'theme',
     'user'
   ],
 
   requires: [
-    'foam.nanos.auth.User',
-    'foam.u2.detail.SectionView'
+    'foam.u2.detail.SectionView',
+    'foam.nanos.auth.SendPassword',
```
Nauna: I'd like to see this get to a point where it doesn't need to know anything that is currently required. Challenge to you Jin

jinn9: you mean like we should not require 'foam.u2.detail.SectionView', 'foam.nanos.auth.SendPassword', etc., and make it work?

src/foam/nanos/auth/ResetPassword.js
Nauna: can you see if it works without this?

src/foam/nanos/u2/navigation/SignIn.js
```diff
@@ -71,7 +71,7 @@ foam.CLASS({
       name: 'subfooterLink',
       code: function() {
         this.stack.push({
-          class: 'foam.nanos.auth.resetPassword.ForgotPasswordView'
+          class: 'foam.nanos.auth.ChangePasswordView'
```
Nauna: So based off of Patrick comments which I agree with this should pass in the SendPassword model here.

and I do think there are more calls to the passwordView which will need to updated. SORRY ... but also thank you!!

src/foam/u2/detail/SectionView.js
```diff
@@ -25,11 +25,14 @@ foam.CLASS({
 
   css: `
     .subtitle {
-      color: /*%GREY2%*/ #8e9090';
+      color: /*%GREY2%*/ #8e9090;
```
Nauna, line 28: as mentioned in previous comments ... not happy about changing this file in any way. tbh i don't think we need to, despite my emotional attachment to keeping this file as is.

Nauna (commented):
Jin pat yourself on the back you did a good job. There are few comments to hopefully simplify some of the code here, but overall looks good. Thanks a million for taking this on.

src/foam/u2/detail/SectionView.js
jinn9: I didn't add this property here. I believe this css property is used elsewhere.
@nanopay-arthur might be able be able to address this issue!

src/foam/nanos/auth/ResetPassword.js
jinnnn9: @Nauna It works well without calling reset_ function in ResetPassword model.

jinnnn9:
Addressed all the comments except for the comment about using SectionDetailView instead of SectionView

src/foam/u2/detail/SectionView.js
nanopay-arthur: @Nauna Adding the subtitle feature was super useful for the new Add Contact Redesigns since most flows had required one. The other way I initially thought of implementing sub-titles was have them be the first section property, but it ended up looking kinda off. @minsunny suggested this approach and I thought it was a good solution because it's opt-in and re-usable. If you're worried about the css, we can inline it to make sure it doesn't get applied where we don't want it.

jinnnn9: @nanopay-arthur may be able to address this issue.

jinnnn9:
@Nauna @zahnno Addressed all the comments

Nauna:
Looking wonderful! seriously but after testing I have a few comments:

- this is going to need a nanopay side pr, that will need to adjust for these changes.
ChangePasswordView is called 12 times in our platform and all that needs to be changed and tested. Luckily I've done some ...
  -- PersonalProfileView.js line 418:
> .start().tag( { class: foam.nanos.auth.ChangePasswordView, model: foam.nanos.auth.UpdatePassword.create({ ishorizontal: true }, this), modelOf: 'foam.nanos.auth.UpdatePassword', showHeader: false } ).end()

  -- SMECOntroller.js line 596:
> view = { class: 'foam.nanos.auth.ChangePasswordView', modelOf: 'foam.nanos.auth.ResetPassword' };

  -- PersonalSettingsView.js line 126:
> .start().tag( { class: foam.nanos.auth.ChangePasswordView, model: foam.nanos.auth.UpdatePassword.create({ ishorizontal: true }, this), modelOf: 'foam.nanos.auth.UpdatePassword', showHeader: false } ).end()

_if you add & test those changes you'll understand the following comments:_
- add isHorizontal to ChangePasswordView and remove from password models.
- the CSS ^content limits the sizing of the fields in horizontal mode and so once in ChangePasswordView I'd recommend toggling the CSS class 
> ^content-withHorizontal  {
      padding: 4vh;
      margin: 0 auto;
    }
    ^
    ^content{
      max-width: 30vw;
      padding-top: 4vh;
      margin: 0 auto;
    }

- another suggested css change:
  >  ^ ^section h2 {
      margin-top: 0;
      margin-bottom: 4vh;
      font-size: 2.5rem;
      **text-align: center;**
    }

- in ChangePasswordView ... property model ... recommend
 >   {
      class: 'FObjectProperty',
      of: this.modelOf,
      name: 'model',
      documentation: 'instance of password model used for this view',
      factory: function() {
        return foam.lookup(this.modelOf).create({}, this);
      },
      view: { class: 'foam.u2.detail.VerticalDetailView' }
    }

A factory instead of an expression is more appropriate here. Only need to build the model once. The view on the property is cleaner then what is currently set in the initE() ... after this change you can replace the model call in initE with **.start(this.MODEL).end()**

jinnnn9:
New update password view:

Screen Shot 2020-04-24 at 12 11 56 PM

zahnno (approved):

src/foam/nanos/auth/ChangePasswordView.js
kgrgreer: Okay, good technique. Maybe add a comment.

src/foam/u2/detail/SectionView.js
jinnnn9: @nanopay-arthur It would be very appreciated if you could give me a heads up if you decide to make changes to 'subtitle' css class. I am also using this class to select div element for subtitle in SectionView.

PR #3396, opened 2020-04-23
do not use cache if current context user is not session user

noodlemoodle:
When ApprovalRequest has been approved, the context user is switched to the person who made the ApprovalRequest so that authorization checks will be done on the correct user.
but the CachingAuthService permission map belongs to the user in the session, so the result of this step was incorrect.


src/foam/nanos/auth/CachingAuthService.java
```diff
@@ -60,7 +61,10 @@ public Object create(X x) {
 
   protected static Map<String,Boolean> getPermissionMap(final X x) {
     Session             session = x.get(Session.class);
-    Map<String,Boolean> map     = (Map) session.getContext().get(CACHE_KEY);
+    Map<String,Boolean> map = session.getUserId() == ((User) x.get("user")).getId() ? 
+      (Map) session.getContext().get(CACHE_KEY) :
+      null;
+
```
noodlemoodle:
```suggestion
```

nanoscott (approved):

src/foam/nanos/auth/CachingAuthService.java
```diff
@@ -60,7 +61,9 @@ public Object create(X x) {
 
   protected static Map<String,Boolean> getPermissionMap(final X x) {
     Session             session = x.get(Session.class);
-    Map<String,Boolean> map     = (Map) session.getContext().get(CACHE_KEY);
+    Map<String,Boolean> map = session.getUserId() == ((User) x.get("user")).getId() ?
```
kgrgreer: Document why you're doing this.
But if you don't use the cache map this will create a new one and override the one in the cache. Is that what you want?

jlhughes (approved):

PR #3400, opened 2020-04-24
[NP-869] - Set Ticket Owner Relationship Visibility To RO

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-869
#TimeSpent: 3h

### DIFF
--- a/src/foam/nanos/ticket/Relationships.js
+++ b/src/foam/nanos/ticket/Relationships.js
@@ -10,7 +10,7 @@ foam.RELATIONSHIP({
   unauthorizedTargetDAOKey: 'localTicketDAO',
   // REVIEW: unable to get owner to display in any section on Ticket.
   targetProperty: {
-    readVisibility: 'RO',
+    visibility: 'RO',
     section: 'infoSection',
     tableCellFormatter: function(value) {
       this.add(this.__subSubContext__.userDAO.find(value)


src/foam/nanos/ticket/Relationships.js
```diff
@@ -10,7 +10,7 @@ foam.RELATIONSHIP({
   unauthorizedTargetDAOKey: 'localTicketDAO',
   // REVIEW: unable to get owner to display in any section on Ticket.
   targetProperty: {
-    readVisibility: 'RO',
+    visibility: 'RO',
```
zahnno, line 13: Would this not prevent us from updating or setting ticket owners? I have a suspicion the RichChoiceView is not aware of visibility related modes if this being set enables us to make changes to the owner on edit or create.

nanopay-arthur: @zahnno This prevents the owner ChoiceView from being rendered as can be seen below.
Screen Shot 2020-04-24 at 12 04 50 PM
In regards to the property itself, once we create the ticket, the owner property gets set just fine as can be seen bellow. 
Screen Shot 2020-04-24 at 1 08 35 PM
In regards to update/edit ticket, you are right this will prevent us from editing the owner on a ticket. Is this functionality that we need?

zahnno: hmm, I expected the visibility function to apply to create read and update operations but I guess not. Seems kinda odd.

"In regards to update/edit ticket, you are right this will prevent us from editing the owner on a ticket. Is this functionality that we need?"

Looking at the property label it leads me to think that I'd be granting a user to sudo as the user selected under "sudo as User" but I don't see where we select the user sudoing into the user. If it's to select the user to sudo as admin or another user with greater permissions then we probably wouldn't.

zahnno: @jlhughes

PR #3402, opened 2020-04-27
[NP-972] - Add strategizerDisabled Property To FObjectView

nanopay-arthur:
Refs: https://nanopay.atlassian.net/browse/NP-972
#TimeSpent: 4h

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -79,6 +79,12 @@ foam.CLASS({
         model in the 'of' property. The user can choose to create an instance
         of one of the models in this list.
       `
+    },
+    {
+      class: 'Boolean',
+      name: 'strategizerDisabled',
+      documentation: 'Set this to true to disable the rendering of the strategizer.',
+      value: false
     }
   ],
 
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/FObjectView.js
```diff
@@ -79,6 +79,12 @@ foam.CLASS({
         model in the 'of' property. The user can choose to create an instance
         of one of the models in this list.
       `
+    },
+    {
+      class: 'Boolean',
+      name: 'strategizerDisabled',
```
kgrgreer: Can we make this more positive and reverse the meaning? Like 'enableStrategizer' and then make the default true.

zahnno (approved):

src/foam/u2/view/FObjectView.js
```diff
@@ -89,7 +95,7 @@ foam.CLASS({
     },
 
     function updateChoices() {
-      if ( this.of == null ) {
+      if ( this.of == null || ! this.enableStrategizer ) {
```
kgrgreer, line 98: Is || correct?

PR #3405, opened 2020-04-28
add support for localhost

hchoura:


src/foam/nanos/session/Session.js
```diff
@@ -243,6 +243,8 @@ foam.CLASS({
             } else {
               configUrl = "https://" + configUrl;
             }
+          } else {
+            configUrl = "http://localhost:8080";
```
jlhughes, line 247: This doesn't make sense.  This will override the req url to the localhost:8080 (which already the appConfig default) when getForceHttps is false.  
What does this fix?

jlhughes (changes requested):

src/foam/nanos/session/Session.js
kgrgreer: What is this for?

hchoura: I remark that the configUrl may equal to "" (empty string); and this will create an exception.
I add it as a default value.

jlhughes: @hchoura you are suggesting that          
String configUrl = ((Request) req).getRootURL().toString();
can produce an empty string?

PR #3410, opened 2020-04-29
skip permission check if the user is the file owner

nanoNeel:
https://nanopay.atlassian.net/browse/NP-1013


jlhughes:
Accepting for now. We'll address not updating files in the documentDAO.

PR #3412, opened 2020-04-29
Allow Sectioned Detail View to define sections

zahnno:
Enables WizardSectionsView to define sections when created.


jlhughes:
Can the section model not handle this? section enabled / predicate which the invoker of the detail view could configure - as the sections themselves should be available via axioms.

src/foam/u2/detail/AbstractSectionedDetailView.js
```diff
@@ -75,7 +81,9 @@ foam.CLASS({
       expression: function(of) {
         if ( ! of ) return [];
 
-        sections = of.getAxiomsByClass(this.SectionAxiom)
+        var selectSections = of.getAxiomsByClass(this.SectionAxiom);
+        if ( this.useSections.length ) selectSections = selectSections.filter((a) => this.useSections.includes(a.name));
+        sections = Array.from(selectSections)
```
kgrgreer: What does this line do?

zahnno: Copies selectSections to sections.

PR #3413, opened 2020-04-30
[NP-65] Support mapping Themes to URL domain.

jlhughes:
ThemeDomainsDAO manages a reverse lookup of domains to themes.
Theme now includes AppConfig which is applied before Group AppConfig settings.

NP-65 #time 4h


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -483,24 +484,32 @@ foam.CLASS({
...
-          this.theme = await dao.find(predicate);
+          predicates = [];
+          var domain = window.location.hostname;
+          var themeDomain = await themeDomainDAO.find(this.domain);
+          predicates.push(
+            this.AND(
+              this.EQ(this.Theme.ID, this.themeDomain.theme),
+              this.EQ(this.Theme.ENABLED, true)
+            ));
+          this.theme = await themeDAO.find(predicates);
```
zahnno: If theme id is different we would want to update the navigation and footer by calling `useCustomElements`.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -483,24 +484,32 @@ foam.CLASS({
...
+          // legacy support
           if ( this.webApp ) {
             predicates.push(this.EQ(this.Theme.APP_NAME, this.webApp));
+            try {
+              this.theme = await themeDAO.find(predicate);
+              if ( ! lastTheme || lastTheme.id != this.theme.id ) this.useCustomElements();
+              return;
+            } catch (err) {
+              console.warning(err);
+              // continue
```
zahnno: In this logic the application controller will consider `personalTheme` if none is found then use `webApp`. If the `personalTheme` is found then we use the `domainTheme`. Seems kind of odd since we're disregarding the personalTheme altogether.

I think the cascade applied should be personalTheme -> webApp? (Do we need this if we have domain themes?) -> domainTheme

The domain may have multiple themes associated to it so this would return the first one found.

src/foam/nanos/theme/ThemeDomain.js
```diff
@@ -0,0 +1,42 @@
...
+      label: 'Domain'
+    },
+    {
+      name: 'theme',
+      class: 'Reference',
+      of: 'foam.nanos.theme.Theme',
+      tableCellFormatter: function(value, obj, axiom) {
+        this.__subSubContext__.themeDAO
+          .find(value)
+          .then((theme) => {
```
zahnno, line 30:
```javascript
value$find.then((theme) => {
```
Will do the same.

src/foam/nanos/controller/ApplicationController.js
jlhughes: yes, and that is further down.

src/foam/nanos/controller/ApplicationController.js
jlhughes: webapp has been removed. 
personal theme is tested first and we fall through if it doesn't exist, just like it was before. 
A domain will only have one theme.

src/foam/nanos/theme/ThemeDomain.js
jlhughes: this was cut and paste, I don't know anything about u2 - unfortunately.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -481,26 +482,24 @@ foam.CLASS({
           // If the user has a personal theme, use that.
           this.theme = await this.user.personalTheme$find;
```
zahnno: Considering the potential that the user personalTheme cannot be found from the personalTheme$find (Missing in themeDAO), We could remove the else below and `return` if personalTheme$find is found to support further fall back.

jlhughes: Updated.

src/foam/nanos/theme/ThemeDomain.js
jlhughes: upated.

src/foam/nanos/session/Session.js
```diff
@@ -233,6 +235,19 @@ foam.CLASS({
...
+            Theme theme = (Theme) ((DAO) x.get("themeDAO")).find(td.getTheme());
+            if ( theme != null ) {
+              AppConfig themeAppConfig = theme.getAppConfig();
+              if ( themeAppConfig != null ) {
+                appConfig.copyFrom(themeAppConfig);
+              }
+              rtn = rtn.put("theme", theme);
+            }
+          }
+
```
jlhughes: if ThemeDomain or Theme not found, stick a default Theme    'new Theme()' into the context.

PR #3414, opened 2020-04-30
put diff hashmap into a treemap so the props can be iterated thru in order

noodlemoodle:
the keys of the diff map should be sorted to avoid inconsistencies in the final result

### DIFF
--- a/src/foam/nanos/approval/ApprovableAware.js
+++ b/src/foam/nanos/approval/ApprovableAware.js
@@ -52,8 +52,11 @@
               }
 
               Map diffHashmap = oldObj == null ? null : oldObj.diff(obj);
-              TreeMap diff = new TreeMap<>();
-              diff.putAll(diffHashmap);
+              TreeMap diff = null;
+              if ( diffHashmap != null ) {
+                diff = new TreeMap<>();
+                diff.putAll(diffHashmap);
+              }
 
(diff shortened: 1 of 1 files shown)


src/foam/nanos/approval/ApprovableAware.js
```diff
@@ -49,7 +50,11 @@
                 Logger logger = (Logger) x.get("logger");
                 logger.error("Error instantiating : ", obj.getClass().getSimpleName(), e);
               }
-              Map diff = oldObj == null ? null : oldObj.diff(obj);
+
+              Map diffHashmap = oldObj == null ? null : oldObj.diff(obj);
+              TreeMap diff = new TreeMap<>();
+              diff.putAll(diffHashmap);
```
nanoscott: skip this if the map is null

nanoscott (approved):

src/foam/nanos/approval/ApprovableAware.js
```diff
@@ -49,7 +50,14 @@
                 Logger logger = (Logger) x.get("logger");
                 logger.error("Error instantiating : ", obj.getClass().getSimpleName(), e);
               }
-              Map diff = oldObj == null ? null : oldObj.diff(obj);
+
+              Map diffHashmap = oldObj == null ? null : oldObj.diff(obj);
```
kgrgreer, line 56: Please add a comment to explain why we do this.

PR #3423, opened 2020-05-05
[NP-843] - Give RichChoiceView searchBy Support

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-843
#TimeSpent: 4h

### DIFF
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -327,10 +327,12 @@ foam.CLASS({
       postSet: function(oldValue, newValue) {
         this.sections.forEach((section) => {
           if ( newValue ) {
-            var pred = this.KEYWORD(newValue);
             if ( section.searchBy.length > 0 ) {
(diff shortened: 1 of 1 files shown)


zahnno (approved):

kennyqiyenkan (approved):

src/foam/u2/view/RichChoiceView.js
```diff
@@ -321,9 +326,17 @@ foam.CLASS({
       documentation: 'The text that the user typed in to search by.',
       postSet: function(oldValue, newValue) {
         this.sections.forEach((section) => {
-          section.filteredDAO = newValue
-            ? section.dao.where(this.KEYWORD(newValue))
-            : section.dao;
+          if ( newValue ) {
+            var pred = this.KEYWORD(newValue);
```
kgrgreer: Could set 'pred' in the 'else' of the below 'if' rather than create a KEYWORD which gets thrown out. Would also make the code easier to understand.

PR #3426, opened 2020-05-05
re-throw the exception

hchoura:


src/foam/box/SessionServerBox.java
```diff
@@ -108,7 +108,7 @@ public void send(Message msg) {
       logger.error("Error throw in SessionServerBox: " + t, " ,service: " + spec.getName(), t);
       t.printStackTrace();
       msg.replyWithException(t);
-      return;
+      throw t;
```
jlhughes, line 111: @hchoura why are you doing this? The exception is added the 'msg' being returned. what good is rethrowing it as this point?

PR #3429, opened 2020-05-06
add app configue into email template

yij793:
address : https://nanopay.atlassian.net/browse/NP-959
Note: add theme arg to email template
Nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/9756

### DIFF
--- a/src/foam/util/Emails/EmailsUtility.java
+++ b/src/foam/util/Emails/EmailsUtility.java
@@ -58,7 +58,7 @@ public class EmailsUtility {
       Theme theme = (Theme) themeDAO.find(
             MLang.EQ(Theme.SPID, user.getSpid())
           );
-      templateArgs.put("theme", theme.appConfig);
+      templateArgs.put("theme", theme);
       emailMessage.setTemplateArguments(templateArgs);
     }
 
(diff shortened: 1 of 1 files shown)


jlhughes (changes requested):
This is not necessary
See NP-65/Themes - this PR already ensures the appconfig in the context takes into account the Theme.appConfig, group appconfig settings and the baseline. 
Please close this PR and use the appconfig in the context. 
Also, theme will not be managing spid in any way.

src/foam/util/Emails/EmailsUtility.java
```diff
@@ -52,6 +54,11 @@ public static void sendEmailFromTemplate(X x, User user, EmailMessage emailMessa
         templateArgs = new HashMap<>();
         templateArgs.put("template", templateName);
       }
+      DAO themeDAO = (DAO) x.get("themeDAO");
+      Theme theme = (Theme) themeDAO.find(
+            MLang.EQ(Theme.SPID, user.getSpid())
+          );
+      templateArgs.put("theme", theme);
```
jlhughes, line 61: This is inappropriate.  we'll add theme to the context (via Session), then this is not necessary.  And having an entire modeled object as an argument is silly.   These should be key/value pairs.

PR #3432, opened 2020-05-06
NP-998 Phone Number View Hide

carl-zzz:
Just hide some phone properties, and later need to be removed
https://nanopay.atlassian.net/browse/NP-998
https://nanopay.atlassian.net/browse/CPF-3137
https://nanopay.atlassian.net/browse/CPF-3362

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -239,6 +239,7 @@ foam.CLASS({
       },
       view: { class: 'foam.u2.detail.VerticalDetailView' },
       createVisibility: 'HIDDEN',
+      hidden: true,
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/User.js
```diff
@@ -239,6 +239,7 @@ foam.CLASS({
       },
       view: { class: 'foam.u2.detail.VerticalDetailView' },
       createVisibility: 'HIDDEN',
+      hidden: true,
```
jlhughes: replace both createVisibility and hidden with 
visibility: 'HIDDEN'
here and elsewhere.

jlhughes (approved):

PR #3435, opened 2020-05-07
add theme to context

yij793:
address: https://nanopay.atlassian.net/browse/NP-65
foam: https://github.com/foam-framework/foam2/pull/3413

### DIFF
--- a/src/foam/util/Emails/EmailsUtility.java
+++ b/src/foam/util/Emails/EmailsUtility.java
@@ -2,10 +2,12 @@ package foam.util.Emails;
 
 import foam.core.X;
 import foam.dao.DAO;
+import foam.nanos.app.AppConfig;
 import foam.nanos.auth.User;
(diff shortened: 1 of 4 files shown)


src/foam/util/Emails/EmailsUtility.java
```diff
@@ -52,6 +56,8 @@ public static void sendEmailFromTemplate(X x, User user, EmailMessage emailMessa
         templateArgs = new HashMap<>();
         templateArgs.put("template", templateName);
       }
+      templateArgs.put("appConfig", appConfig);
+      templateArgs.put("theme", theme);
```
jlhughes: again, both of these are incorrect.
the args must be key/value pairs. 
so if you need phonenumber then you would add
put("phoneNumber", theme.getSupportUser().getPhoneNumber())
for example.

jlhughes (approved):

PR #3437, opened 2020-05-08
NotificationSettings Update

nanoscott:
- Changing the notificationDAO decorators to use the NotificationSettings
- Adding the ability to enable and disable notifications settings and email settings
- Removing disabledTopicsEmail since it is not used and reusing disabledTopics to prevent email sending
- Marking in-app notifications that are disabled as read automatically. If we were to prevent them from being written altogether, then the other notifications (e.g. email, slack, sms) wouldn't be triggered either, so we have to at least write them to the DAO

linked nanopay PR: https://github.com/nanoPayinc/NANOPAY/pull/9872


src/foam/nanos/notification/NotificationSetting.js
```diff
@@ -62,7 +85,18 @@ foam.CLASS({
...
         notification.setBroadcasted(false);
         notification.setGroupId(null);
-        notification.setEmailIsEnabled(false);
+
+        // We cannot permanently disable in-app notifications, so mark them read automatically
+        if ( ! getEnabled() ) {
+          notification.setRead(true);
+        }
+        else if ( user.getDisabledTopics() != null ) {
+          List disabledTopics = Arrays.asList(user.getDisabledTopics());
```
kgrgreer: This seems expensive. It will on average take 1.5*n operations plus generate a throw-away garbage List. Could we write a helper method to just find in the array, or better, have the javaPreSet for the property sort the array before storing then call Arrays.binarySearch() which will be log n? Same with above.

src/foam/nanos/notification/EmailSetting.js
```diff
@@ -54,34 +54,53 @@ foam.CLASS({
...
+        if ( ! getEnabled() ) 
+          return;
 
-            if ( notification.getEmailArgs() != null ) {
-              Map<String, Object> emailArgs = resolveNotificationArguments(x, notification.getEmailArgs(), user);
-              notification.setEmailArgs(emailArgs);
-            }
+        // Skip sending email messages for disabled topics
+        if ( user.getDisabledTopics() != null ) {
+          List disabledTopics = Arrays.asList(user.getDisabledTopics());
```
kgrgreer: Here also.

src/foam/nanos/notification/NotificationSetting.js
```diff
@@ -44,10 +46,31 @@ foam.CLASS({
...
+      name: 'doNotify',
+      args: [
+        { name: 'x', type: 'Context' },
+        { name: 'user', type: 'foam.nanos.auth.User' },
+        { name: 'notification', type: 'foam.nanos.notification.Notification' }
+      ],
+      javaCode: `
+        // Disable email sending when performing a doNotify to prevent duplicate emails
+        notification = (Notification) notification.fclone();
+        notification.setEmailIsEnabled(false);
```
jlhughes: Just noticing this here, but why do we need medium/message specific flags on the notification model now that we have notificationSettings?

src/foam/nanos/notification/SendEmailNotificationDAO.java
```diff
@@ -31,47 +36,40 @@ public SendEmailNotificationDAO(X x, DAO delegate) {
...
+    Notification notification    = (Notification) obj;
+    User         user            = (User) userDAO.find(notification.getUserId());
+    Notification oldNotification = (Notification) getDelegate().find(obj);
 
-    if ( oldNotif != null ) return super.put_(x, obj);
+    if ( oldNotification != null ) 
+      return super.put_(x, obj);
 
-    if ( ! notif.getEmailIsEnabled() || user == null ) return super.put_(x, obj);
+    if ( ! notification.getEmailIsEnabled() || user == null )
```
jlhughes: This should not be required anymore.

src/foam/nanos/notification/SendNotificationDAO.java
```diff
@@ -51,24 +51,52 @@ public void put(Object o, Detachable d) {
...
     }
 
     if ( SafetyUtil.isEmpty(notif.getGroupId()) && ! notif.getBroadcasted() ) {
-      return super.put_(x, notif);
+      Notification notification = notif;
+      User nofifUser = (User) userDAO.find(notif.getUserId());
+      if ( nofifUser != null ) {
+        NotificationSetting notificationSetting = getNotificationSetting(x, nofifUser);
+
+        // We cannot permanently disable in-app notifications, so mark them read automatically
```
jlhughes: curious why the InAppNotificationSettings can't handle setting 'read' during doNotify.

src/foam/nanos/notification/NotificationSetting.js
nanoscott: legacy thing .. i'll remove it .. it'll break nanopay though without a merge there too, so i'll have to prepare a PR on that branch as well.

src/foam/nanos/notification/SendEmailNotificationDAO.java
nanoscott: i'll remove it.

src/foam/nanos/notification/SendNotificationDAO.java
nanoscott: sometimes we don't go through InAppNotificationSettings .. specifically when we save a user's notification directly to the dao, hence the logic here as well as in the InAppNotification .. I can't really replace this with InAppNotificationSettings or I'll get an infinite loop.

src/foam/nanos/notification/NotificationSetting.js
nanoscott: i'll see what I can do.

zahnno (approved):

src/foam/nanos/notification/NotificationAction.js
```diff
@@ -0,0 +1,22 @@
+foam.ENUM({
```
jlhughes: This should be in a DAO, these will change as we add more are replace things. 
The notification settings themselves could register an ID/name to use elsewhere.

jlhughes (changes requested):

src/foam/nanos/notification/NotificationAction.js
nanoscott: @jlhughes I'm not sure what you mean here. This is an enum used by the Notification object to control whether the SendEmailNotificationDAO actually sends a notification. The confusion here comes by us having two ways of performing actions:

1. DAO.put
2. NotificationSetting.sendNotification (and the subclasses of NotificationSetting)

Since the NotificationSetting puts to the notificationDAO, there is duplicate logic when performing a user.doNotify call. Specifically, it calls the EmailSettings.sendNotification and also performs logic in the SendEmailNotificationDAO .. so unless there is a way to control whether a Notification instance should send an email (i.e. getEmailIsEnabled() call or the getAction() call), then we will have duplicate emails sent to the user.
I'm not sure how adding this to a DAO will help, or what it would even mean to have this is a DAO. Are you saying we should have Action types in a DAO and this is a reference to an action type .. and the NotificatSetting should have that same reference to the action type? I'm not sure why you would want this configurable since the *Setting object would have to be coded in order to enable it (e.g. SlackSetting) in which case the value could also be added to the enum. Unless you have a way to add NotificationSetting subtypes on the fly, I can't see why we would do this. 
I can add some logic here to match the enum instead of having it manually set per setting type.

src/foam/nanos/auth/User.js
```diff
@@ -448,14 +450,21 @@ foam.CLASS({
...
-      name: 'disabledTopicsEmail',
-      documentation: 'Disables types for email notifications.',
-      createVisibility: 'HIDDEN',
-      section: 'administrative'
+      class: 'Object',
+      /** @private */
+      name: 'disabledTopicSet',
+      javaType: 'java.util.HashSet',
+      hidden: true,
+      factory: function() { return {}; }
```
kgrgreer: Make transient.

src/foam/nanos/auth/User.js
```diff
@@ -448,14 +450,21 @@ foam.CLASS({
       name: 'disabledTopics',
       documentation: 'Disables types for notifications.',
       createVisibility: 'HIDDEN',
-      section: 'administrative'
+      section: 'administrative',
+      javaPostSet: `
```
kgrgreer, line 454: Better to create in the factory: for disabledTopicSet() and just clear that property in the post set here. That prevents you from having to generate the set if not used. I would rather just use a for loop to loop over val rather than create an intermediate Array which then needs to be garbate collected.

src/foam/nanos/notification/NotificationExpansionDAO.js
```diff
@@ -0,0 +1,81 @@
...
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.notification',
+  name: 'NotificationExpansionDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  documentation: `Expands a notification from a group or broadcast notification to a user specific notification. `,
```
kgrgreer: Doesn't SendNotificationDAO already do this?

src/foam/nanos/notification/NotificationExpansionDAO.js
```diff
@@ -0,0 +1,81 @@
...
+
+  implements: [
```
kgrgreer: This isn't used is it?

src/foam/nanos/notification/NotificationTemplateDAO.js
```diff
@@ -50,27 +52,38 @@ the notification will be handled. `,
...
+            .select(new ArraySink()))
+            .getArray();
+
           if ( templates.size() > 1 ) {
-            System.err.println("Multiple Notification templates for "+notification.getTemplate()+" found.");
+            logger.error("Multiple Notification templates for " + notification.getTemplate() + " found.");
             return notification;
           } else if ( templates.size() == 1 ) {
             template = (Notification) ((FObject)templates.get(0)).fclone();
-            // REVIEW/FIXME: no copyFrom in our java
```
kgrgreer, line 65: There is a copyFrom in FObject now.

kgrgreer: Just an FYI, this is fine.

src/foam/nanos/notification/SlackSetting.js
```diff
@@ -0,0 +1,61 @@
...
+        HttpPost httpPost = new HttpPost(slackWebhook);
+        httpPost.addHeader("Content-type", "application/json");
+        
+        // Add the slack message to the post
+        StringEntity params = new StringEntity("{\"text\" : \"" + slackMessage + "\"}", "UTF-8");
+        params.setContentType("application/json");
+        httpPost.setEntity(params);
+        
+        try {
+            CloseableHttpResponse response =  HttpClients.createDefault().execute(httpPost);
```
kgrgreer: indentation is off

src/foam/nanos/notification/NotificationExpansionDAO.js
jlhughes: SendNotificationDAO and all other Notification DAO decorators have been removed.

src/foam/nanos/notification/NotificationExpansionDAO.js
```diff
@@ -0,0 +1,81 @@
...
+        if ( notif.getBroadcasted() ) {
+          userDAO.select(new AbstractSink() {
+            @Override
+            public void put(Object o, Detachable d) {
+              User user = (User) o;
+              user.doNotify(x, notif);
+            }
+          });
+        } else if ( ! SafetyUtil.isEmpty(notif.getGroupId()) ) {
+          DAO receivers = userDAO.where(
```
jlhughes, line 49: This is an example where group should implement doNotify, then it can take care of notifying it's users.

nanoscott: Could do in the future .. i'll create a follow-up FOAM ticket for it.

src/foam/nanos/notification/NotificationTemplateDAO.js
nanoscott: got it.

jlhughes (approved):

src/foam/nanos/notification/NotificationExpansionDAO.js
nanoscott: https://nanopay.atlassian.net/browse/NP-1176 created.

PR #3438, opened 2020-05-08
add new theme() to context if theme is null

yij793:
changes : 1. adding new theme to context if theme is null
                 2. remove visibility from theme.supportUser
origin: https://github.com/foam-framework/foam2/pull/3413


src/foam/nanos/session/Session.js
```diff
@@ -288,6 +291,13 @@ foam.CLASS({
           .put("twoFactorSuccess", getContext().get("twoFactorSuccess"))
           .put(CachingAuthService.CACHE_KEY, getContext().get(CachingAuthService.CACHE_KEY));
 
+        if ( theme != null ) {
+          rtn = rtn.put("theme", theme);
+        } else {
+          theme = new Theme();
```
jlhughes: move this up to line 237, then none of this extra checking is required.

PR #3449, opened 2020-05-11
add context to test

hchoura:
related to https://github.com/nanoPayinc/NANOPAY/pull/9869

### DIFF
--- a/src/foam/test/TestUtils.java
+++ b/src/foam/test/TestUtils.java
@@ -10,6 +10,8 @@ import foam.core.ClassInfo;
 import foam.core.X;
 import foam.dao.DAO;
 import foam.dao.MDAO;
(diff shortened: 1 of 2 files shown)


src/foam/test/TestUtils.java
```diff
@@ -89,7 +92,10 @@ public static boolean testThrows(
         System.out.println("EXPECTED: \""+exceptionType.getName()+"\"");
         System.out.println("ACTUAL  : \""+t.getClass().getName()+"\"");
         t.printStackTrace();
-        throw t;
+        
+        AppConfig appConfig = (AppConfig) x.get("appConfig");
```
jlhughes, line 96: This should already be the case so would be ok to throw.

PR #3450, opened 2020-05-11
Persist data after sub class select

zahnno:
Persisting data through FObject Strategizer class selection.

This is allows factory data on properties or set values that use the FObjectView to retain data when selecting a new sub class.

Example: ExampleModel isEnabled field is set to true in factory. User creating ExampleModal selects new subclass of ExampleModel through FObjectView, isEnable will continue to be true even if isEnabled defaults to false.

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -35,11 +35,13 @@ foam.CLASS({
         };
       },
       postSet: function(oldValue, newValue) {
-        if ( newValue !== oldValue && oldValue !== '') {
+        if ( newValue !== oldValue && oldValue !== '' ) {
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

zahnno:
Thinking that this should be the default behaviour instead of explicitly setting a flag @kgrgreer @jlhughes

nanoNeel:
If its default behaviour, I'm pretty sure it will break something.
`  m.create(null, this).copyFrom(this.data) :`
I think something like this was removed by Troy before since it was breaking something in one of the strategizers.

edit:
Found Troy's PR : https://github.com/foam-framework/foam2/pull/2838

zahnno:
Thanks @nanoNeel Let's keep the flag based on the analysis on 2838

PR #3457, opened 2020-05-12
Added strategizer predicate on query & Add persistantData to FObject

zahnno:
- Enables strategizer query to utilize custom predicate. Useful for limiting strategies based on existing data. Ex. (IN(this.StrategyReference.STRATEGY, this.classInfoArray))

- Add persistantData to FObjectView to persist data through strategizer class selection. 
Solves both use cases mentioned in https://github.com/foam-framework/foam2/pull/2838. Data continues to persist and default values of selected classes will be applied.

### DIFF
--- a/src/foam/strategy/BasicStrategizer.js
+++ b/src/foam/strategy/BasicStrategizer.js
@@ -50,6 +50,10 @@ foam.CLASS({
               )
             );
 
+        if ( strategyPredicate != null  ) {
+          predicate = (Predicate) strategyPredicate;
+        }
+
         DAO strategyDAO = ((DAO) getStrategyDAO()).inX(x);
         List refs = ((ArraySink) strategyDAO
           .where(predicate)
--- a/src/foam/strategy/StrategyReference.js
+++ b/src/foam/strategy/StrategyReference.js
@@ -16,7 +16,6 @@ foam.CLASS({
   tableColumns: [
     'desiredModelId',
     'target',
-    'strategyPredicate',
     'strategy'
   ],
 
--- a/src/foam/strategy/StrategyReference.js
+++ b/src/foam/strategy/StrategyReference.js
@@ -16,6 +16,7 @@ foam.CLASS({
   tableColumns: [
     'desiredModelId',
     'target',
+    'strategyPredicate',
     'strategy'
   ],
 
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -39,9 +39,7 @@ foam.CLASS({
           if ( this.data && this.data.cls_.id === newValue ) return;
           var m = this.__context__.lookup(newValue, true);
           if ( m ) {
-            this.data = this.persistData ?
-                m.create(null, this).copyFrom(this.data) :
-                m.create(null, this);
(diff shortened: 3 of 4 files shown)


src/foam/strategy/StrategyReference.js
```diff
@@ -16,6 +16,7 @@ foam.CLASS({
   tableColumns: [
     'desiredModelId',
     'target',
+    'strategyPredicate',
```
jlhughes: is predicate appropriate for a table column? what does it look like.

src/foam/u2/view/FObjectView.js
```diff
@@ -89,9 +87,13 @@ foam.CLASS({
       value: true
     },
     {
-      class: 'Boolean',
-      name: 'persistData',
-      documentation: 'Allows data to persist to selected class instances.'
+      class: 'Object',
```
jlhughes, line 90: do you have an example of how this will be used/set?

nanoNeel:
LGTM, didn't see any broken strategizer views.

src/foam/strategy/StrategyReference.js
zahnno: No this was accidentally left in from some experimenting. I've removed it.

src/foam/u2/view/FObjectView.js
zahnno: Sure, I'll include the use of a predicate as well. Here's an example.:
```javascript
view: function(_, X) {
  let predicate = expr.AND(
      expr.EQ(foam.strategy.StrategyReference.DESIRED_MODEL_ID, 'foam.nanos.auth.User'),
      expr.IN(foam.strategy.StrategyReference.STRATEGY, [foam.lookup('foam.nanos.auth.SomeUserClass'), foam.lookup('foam.nanos.auth.AnotherUserClass') ])
  );
  return foam.u2.view.FObjectView.create({
    data: myData,
    of: foam.nanos.auth.User,
    persistantData: { website: X.data.user.website },
    predicate: predicate
 }, X);
```
- The FObjectView will persist X.data.user.website throughout any class selections.
- Strategizer class options will include SomeUserClass and AnotherUserClass excluding all other user subclasses. Granted the user has permissions to the strategies and if those classes exist.

jlhughes (approved):

src/foam/strategy/BasicStrategizer.js
```diff
@@ -50,6 +50,10 @@ foam.CLASS({
               )
             );
 
+        if ( strategyPredicate != null  ) {
```
kgrgreer, line 53: The code above to build the predicate should be put in the 'else' block of this 'if', since there is no reason to do the above work if you aren't going to use it in this case.

src/foam/u2/view/FObjectView.js
kgrgreer: Spelling: persistent

kgrgreer: I still don't get it from your explanation. Can you elaborate?

zahnno: @kgrgreer 
`persistentData` will be copied over to any objects created after a strategizer class selection.
`predicate` can omit or include additional strategizer class options.

PR #3460, opened 2020-05-12
DO NOT MERGE: Approvable Rework FOAM

tharmaman:
Rework includes:
1. Approvable ID is a guID and a new prop lookup ID which is the long lookup hash in order to account for repeated requests being submitted if the prior one has been rejected or approved
2. Create uses approvables instead of storing a object in PENDING state
3. Addition of ServerDAOKey to approvalRequest which is an optional property used in the case of the generic approval system if you require storing a clientDAO (which will be stored in the normal daoKey prop) for viewing purposes and a server-side DAO key for executing purposes
4. Update, Create and Remove will all throw an Exception saying (An approval request has been sent out.) to be replaced by the UserFeedbackException @nick-prat  is working on
5. Adjust the DAOCreateView and DAOUpdateViews due to 4. as they now do not return objects when creating and updating under the approval system
6. Removed pointless instanceOf ApprovalRequest check as we now only use the ApprovalRequest model and not RoleApprovalRequest anymore
7. When generating the approvable hash key @noodlemoodle  made, removes lifecycleState, network transient properties and storage transient properties from generating the hash key based on the map of the object
8. DAOSummaryView that gets passed to View Reference needed to hide the create, edit and delete buttons
9. Wired up the approval request to use a summary view when rendering a view reference that refers to an approvable being used for CREATE


src/foam/nanos/approval/ApprovableAware.js
```diff
@@ -74,19 +75,46 @@
...
+                List<String> networkTransientPropertyNames = new ArrayList<>();
+        
+                while ( allProperties.hasNext() ){
+                  PropertyInfo prop = (PropertyInfo) allProperties.next();
+        
+                  if ( prop.getStorageTransient() ){
+                    storageTransientPropertyNames.add(prop.getName());
+                  }
+
+                  if ( prop.getNetworkTransient() ){
```
nanoscott: shouldn't we be keeping network transient properties? these properties are stored in the database and would contribute to a valid change in the object, no?

src/foam/nanos/approval/ApprovableAware.js
```diff
@@ -74,19 +75,46 @@
...
               }
-
+              
               StringBuilder hash_sb = new StringBuilder(obj.getClass().getSimpleName());
               if ( operation == Operations.UPDATE && obj instanceof ApprovableAware ) 
                 hash_sb.append(String.valueOf(obj.getProperty("id")));
 
               if ( diff != null ) {
-                // remove ids, timestamps and userfeedback
+                // remove ids, timestamps and storageTransient properties
```
jlhughes: Why are we removing 'id's?

src/foam/nanos/approval/ApprovableAware.js
jlhughes: Not following this, all of these properties are important regardless of transient type.  Any downstream decorator may make use of them.

tharmaman: This is for when generating the lookupKey as the transient properties would interfere with the lookup. This doesn't effect the items in the propertiesToUpdate map.

However on the approvableAwareDAO I remove only storageTransient properties during updates, as when I'm updating lets say just a name on user, when viewing the reference for what properties were updated I would see userFeedback and on account I would see balance (local) and balance (home). All items which they did not personally edit and seem to come from no where to the user.

What we can do then is instead is:
1. HIDE storage transient and network transient properties in the PropertiesToUpdateView.
2. Keep the removal of storage transient and network transient properties when generating the lookup hash key, but have both storage transient and network transient properties present when generating the maps which get used to create the final object put in the DAO.

@nanoscott @jlhughes

src/foam/nanos/approval/ApprovableAware.js
tharmaman: So again this is for generating the lookup key its not being removed in the PropertiesToUpdate map on approvable, its so one user doesn't spam create requests for duplicate objects pending to be created. The first create request for an exactly identical object would have to be resolved first before they can submit a duplicate request. Was a requirement @noodlemoodle  implemented. 

@nanoscott @jlhughes

src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -375,112 +324,52 @@ foam.CLASS({
...
+            throw new RuntimeException(e);
+          }
+        }
+
+        Map propertiesToApprove = objectToDiffAgainst.diff(obj);
+        
 
         for ( int i = 0; i < storageTransientPropertyNames.size(); i++ ){
-          updatedProperties.remove(storageTransientPropertyNames.get(i));
+          propertiesToApprove.remove(storageTransientPropertyNames.get(i));
```
tharmaman: @jlhughes @nanoscott as mentioned above I can remove this part which takes the storageTransient properties out of the propertiesToUpdate map. But keep the removal of network and storage transient when generating the lookup key and hide storage transient and network transient properties in the PropertiesToUpdateView.

src/foam/nanos/approval/ApprovableAware.js
jlhughes: 1. OK
2. OK.

src/foam/nanos/approval/ApprovableAware.js
jlhughes: OK.

src/foam/nanos/approval/ApprovableAwareDAO.js
jlhughes: Yes, remove for lookup but keep for update.  got it.

src/foam/nanos/approval/ApprovableAware.js
nanoscott: I'm good with it too.

src/foam/nanos/approval/ApprovableAware.js
```diff
@@ -74,19 +75,46 @@
...
+
+                List<String> storageTransientPropertyNames = new ArrayList<>();
+                List<String> networkTransientPropertyNames = new ArrayList<>();
+        
+                while ( allProperties.hasNext() ){
+                  PropertyInfo prop = (PropertyInfo) allProperties.next();
+        
+                  if ( prop.getStorageTransient() ){
+                    storageTransientPropertyNames.add(prop.getName());
+                  }
```
lchanmann: can we remove the prop from the diff here instead of adding to storageTransientPropertyNames? 

Same for networkTransientPropertyNames below.

src/foam/nanos/approval/ApprovalDAO.java
```diff
@@ -73,21 +73,18 @@ public FObject put_(X x, FObject obj) {
   }
 
   private void rePutObject(X x, ApprovalRequest request) {
-    DAO dao = (DAO) x.get(request.getDaoKey());
+    String daoKey = request.getServerDaoKey() != null && ! request.getServerDaoKey().isEmpty() ? request.getServerDaoKey() : request.getDaoKey();
```
lchanmann: String property eg. serverDaoKey has "" as its default value but if for some reason needed to check for null or empty you can use SafetyUtil.isEmpty.

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -659,14 +667,40 @@
...
+              if ( obj.operation === foam.nanos.ruler.Operations.CREATE ){
+                var temporaryNewObject = obj.of.create({}, X);
+
+                var propsToUpdate = obj.propertiesToUpdate;
+
+                var keyArray = Object.keys(propsToUpdate);
+
+                for ( var i = 0; i < keyArray.length; i++ ){
+                  temporaryNewObject[keyArray[i]] = propsToUpdate[keyArray[i]];
+                }
```
lchanmann: I wonder if we can use copyFrom.
```
var temporaryNewObject = obj.of.create();
temporaryNewObject.copyFrom(obj.propertiesToUpdate);
```

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -659,14 +667,40 @@
...
+
+                X.stack.push({
+                  class: 'foam.comics.v2.DAOSummaryView',
+                  data: temporaryNewObject,
+                  of: temporaryNewObject.cls_,
```
lchanmann: if we store obj.of in a variable and it can be reused.

src/foam/nanos/approval/FulfilledApprovableRule.js
```diff
@@ -36,21 +40,38 @@
...
+            }
 
             Map propsToUpdate = approvable.getPropertiesToUpdate();
 
             Object[] keyArray = propsToUpdate.keySet().toArray();
 
             for ( int i = 0; i < keyArray.length; i++ ){
-              objToUpdate.setProperty((String) keyArray[i],propsToUpdate.get(keyArray[i]));
+              objectToPut.setProperty((String) keyArray[i],propsToUpdate.get(keyArray[i]));
             }
```
lchanmann: can we just iterate the map?

src/foam/nanos/approval/FulfilledApprovableRule.js
```diff
@@ -36,21 +40,38 @@
...
 
-            dao.put_(getX(), objToUpdate);
+            User createdBy = (User) ((DAO) x.get("bareUserDAO")).find(approvable.getCreatedBy());
+
+            X createdX = x.put("user", createdBy);
+
+            dao.put_(createdX, objectToPut);
```
lchanmann:
```suggestion
            dao.inX(createdX).put(objectToPut);
```

src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -202,166 +192,125 @@ foam.CLASS({
...
+      if ( fulfilledRequests.size() == 1 ) {
+        ApprovalRequest fulfilledRequest = (ApprovalRequest) fulfilledRequests.get(0);
+        fulfilledRequest.setIsFulfilled(true);
+
+        User lastModifiedBy = (User) ((DAO) x.get("bareUserDAO")).find(fulfilledRequest.getLastModifiedBy());
+        if ( lastModifiedBy == null ) lastModifiedBy = new User.Builder(x).setId(fulfilledRequest.getLastModifiedBy()).build();
+        X approvalX = getX().put("user", lastModifiedBy);
           
-          approvalRequestDAO.put_(approvalX, fulfilledRequest);
+        approvalRequestDAO.put_(approvalX, fulfilledRequest);
```
lchanmann:
```suggestion
        approvalRequestDAO.inX(approvalX).put(fulfilledRequest);
```

src/foam/nanos/approval/ApprovalDAO.java
tharmaman: @lchanmann does `request.getServerDaoKey().isEmpty()` not account for that?

src/foam/nanos/approval/ApprovableAware.js
tharmaman: 1. and 2. just added.

src/foam/nanos/approval/ApprovalRequest.js
tharmaman: done

src/foam/nanos/approval/ApprovableAware.js
tharmaman: Done.

src/foam/nanos/approval/FulfilledApprovableRule.js
tharmaman: Done.

src/foam/nanos/approval/ApprovalRequest.js
tharmaman: Yeah so new version we have to iterate over the properties check if transient and then remove it out anyways. So we can stick with the for loop because I'd have to do it anyways and iterate over the propsToUpdate to check if its transient to do copyFrom

src/foam/nanos/approval/ApprovalDAO.java
lchanmann: yeah, that will do.

jlhughes: this just bit us in production the other day. please change to SafetyUtil.isEmpty.

tharmaman: Done.

nanoscott (approved):

PR #3461, opened 2020-05-12
add exception for the test

hchoura:
To catch exception in the rule engine and in the contextAgent. 

It will cause several exception for other project.


src/foam/core/CompoundContextAgency.js
```diff
@@ -40,6 +40,7 @@ for ( Runnable agent : getAgents() ) {
   } catch (Throwable t) {
     logger.error(t.getMessage(), t);
     e.add(t);
+    throw t;
```
jlhughes: @hchoura the throwable is added to the compoundexception. 
You can't throw it.  This is nonsense.

jlhughes (changes requested):

src/foam/core/CompoundContextAgency.js
kgrgreer: Yes, this doesn't make sense. If there is an exception, then we'll rethrow it on line 45 below:
e.maybeThrow();

src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -84,6 +84,7 @@ public void execute(List<Rule> rules, FObject obj, FObject oldObj) {
       compoundAgency.execute(x_);
     } catch (Exception e) {
       logger.error(e.getMessage(), e);
+      throw e;
```
jlhughes, line 87: this is occuring asynchronously, there is nothing to catch the exception. Reporting is all we can do.  Perhaps we can add it to the rule - so a report of what failed. but we can't throw from here.

src/foam/core/CompoundContextAgency.js
```diff
@@ -42,7 +42,11 @@ for ( Runnable agent : getAgents() ) {
     e.add(t);
   }
 }
-e.maybeThrow();`
+try {
+  e.maybeThrow();
+} catch ( Throwable e1 ) {
+  e1.printStackTrace();
+}`
```
jlhughes: if e.maybeThrow() throws then the e1 needs to be rethrown.

src/foam/core/CompoundException.java
```diff
@@ -11,15 +11,16 @@
...
   public void add(Throwable t) {
     exceptions_.add(t);
   }
 
-  public void maybeThrow() {
-    if ( exceptions_.size() != 0 ) {
-      throw this;
+  public void maybeThrow() throws Throwable {
+    for ( int i = 0 ; i < exceptions_.size() ; i++ ) {
+      System.out.println( "This is an error =======> " + ( (Throwable) exceptions_.get(i) ) );
```
jlhughes: We only want this in TEST mode.

src/foam/core/CompoundException.java
```diff
@@ -11,15 +11,16 @@
...
+      throw (Throwable) exceptions_.get(i);
```
jlhughes: and why are we throwing on the first?

src/foam/core/ContextAgentRunnable.java
```diff
@@ -24,6 +24,11 @@ public String toString() {
   }
 
   public void run() {
-    agent_.execute(x_);
+    try {
+      agent_.execute(x_);
+    } catch ( Throwable e ) {
+      e.printStackTrace();
```
jlhughes: Again only in TEST mode.

src/foam/core/ContextAgentRunnable.java
```diff
@@ -24,6 +24,11 @@ public String toString() {
...
+      throw e;
```
jlhughes: and this is pointless. Who is catching it?

src/foam/core/CompoundContextAgency.js
```diff
@@ -42,7 +42,11 @@ for ( Runnable agent : getAgents() ) {
     e.add(t);
   }
 }
-e.maybeThrow();`
+try {
```
kgrgreer: What's the point of throwing the exception if we're just going to catch it.

src/foam/core/CompoundException.java
kgrgreer: This will only throw the first exception. We want to throw them all.

src/foam/core/ContextAgentRunnable.java
kgrgreer: Why is this needed? This will just pollute the stdout.

src/foam/core/CompoundContextAgency.js
```diff
@@ -42,7 +43,8 @@ for ( Runnable agent : getAgents() ) {
     e.add(t);
   }
 }
-e.maybeThrow();`
+e.maybeThrow((AppConfig) getX().get("appConfig"));
```
jlhughes, line 46: How about passing a context - that would be consistent with the rest of FOAM.

src/foam/core/CompoundException.java
```diff
@@ -7,19 +7,24 @@
...
     exceptions_.add(t);
   }
 
-  public void maybeThrow() {
-    if ( exceptions_.size() != 0 ) {
-      throw this;
+  public void maybeThrow(AppConfig appConfig) {
+    if ( Mode.TEST == appConfig.getMode() ) {
+      for ( int i = 0 ; i < exceptions_.size() ; i++ ) {
+        System.out.println( "This is an error =======> " + ( (Throwable) exceptions_.get(i) ) );
```
jlhughes, line 25: Please implement toString on CompoundException which can generate an appropriate message.

src/foam/core/CompoundException.java
```diff
@@ -7,19 +7,24 @@
...
+        throw (RuntimeException) exceptions_.get(i);
```
jlhughes, line 26: and this still makes no sense.

jlhughes (changes requested):

jlhughes:
Superceded by another PR.

PR #3462, opened 2020-05-13
Exception Java Support

nick-prat:
Reworked implementation of https://github.com/foam-framework/foam2/pull/2647

### Changes

- Instead of `foam.EXCEPTION({...})` to signify a native exception needs to be made, use the flag `generateNativeException: true`
- Foam Exception will be generated as they were previously, retaining their to/from CSV, Builders, etc...
- A `javaException()` method is injected into the foam model before creation, this method returns a new instance of the native exception class


nick-prat:
As an example, we can look at UserFeedbackException: 

Given the model:
```
foam.CLASS({
  package: 'foam.comics.v2.userfeedback',
  name: 'UserFeedbackException',
  extends: 'foam.core.AbstractException',
  generateNativeException: true,

  documentation: `
   ...
  `,

  properties: [
    {
      class: 'FObjectProperty',
      name: 'userFeedback'
    }
  ]
});
```

It generates a regular `UserFeedbackException` class, but also an accompanying java exception class `UserFeedbackNativeException` containing:

```
package foam.comics.v2.userfeedback;


public class UserFeedbackNativeException extends foam.core.AbstractNativeException {

  protected foam.comics.v2.userfeedback.UserFeedback userFeedback_;

  private boolean userFeedbackIsSet_ =     false
;


  public foam.comics.v2.userfeedback.UserFeedback getUserFeedback() {
    if ( ! userFeedbackIsSet_ ) {
     return null;
    }
    return userFeedback_;
  }

  public void setUserFeedback(foam.comics.v2.userfeedback.UserFeedback val) {
    userFeedback_ = val;
    userFeedbackIsSet_ = true;
    
  }

  public void clearUserFeedback() {
    
    userFeedbackIsSet_ = false;
    
  }

  public UserFeedbackNativeException(UserFeedbackException data) {
    super(data);
    this.setUserFeedback(data.getUserFeedback());
    this.setMessage(data.getMessage());
  }
}
```

This java exception class extends `AbstractNativeException` and it's generated class file is:
```
package foam.core;


public class AbstractNativeException extends java.lang.RuntimeException implements foam.core.MessageException {

  foam.core.FObject foamException;

  protected String message_;

  private boolean messageIsSet_ =     false
;


  public foam.core.FObject getFoamException() {
    return foamException;
  }

  public String getMessage() {
    if ( ! messageIsSet_ ) {
     return "";
    }
    return message_;
  }

  public void setMessage(String val) {
    message_ = val;
    messageIsSet_ = true;
    
  }

  public void clearMessage() {
    
    messageIsSet_ = false;
    
  }

  public AbstractNativeException(AbstractException data) {
    super();
    this.setMessage(data.getMessage());
    this.foamException = data;
  }
}
```
An example of how this will be used is:
```
throw new UserFeedbackException.Builder(x)
        .setUserFeedback(new UserFeedback.Builder(x)
        .setMessage("An approval request has been sent out.")
        .setStatus(UserFeedbackStatus.SUCCESS)
        .build()
      ).build().javaException();
```
The idea is that given a class `someFoamException` with `generateNativeException: true` you can throw it via `throw someFoamException.javaException();`

src/foam/comics/v2/userfeedback/UserFeedbackException.js
```diff
@@ -7,7 +7,9 @@
 foam.CLASS({
   package: 'foam.comics.v2.userfeedback',
   name: 'UserFeedbackException',
-  implements: ['foam.core.Exception'],
+  extends: 'foam.core.AbstractException',
+  generateNativeException: true,
```
jlhughes: @nick-prat @KernelDeimos is 'native' java when generating java, swift for swift, ... ?

nick-prat: @jlhughes I'm not very familiar with swift, but from what I've read it seems that the throwable type is an interface instead of a class, so it renders this whole solution redundant.  It might be a good idea to change this to `generateJavaException` instead, as it's a java specific problem & solution.

KernelDeimos: I agree with @nick-prat that `generateJavaException` makes more sense, but now I'm thinking exceptions should just always do that unless they're excluded from java generation alltogether, right?

nick-prat: I think keeping it opt-in is good as I'd rather have the compile load lower then a have a large amount of unused java exception classes.

jlhughes: @nick-prat @KernelDeimos re: opt-in - yes, I was referring to what 'native' meant.  So change to generateJavaException is that's all it can do at the moment. and also newNative to javaException() or something like that.

nick-prat: @jlhughes Done

KernelDeimos:
@googlebot I consent.

nick-prat:
@googlebot I fixed it.

PR #3465, opened 2020-05-13
Support function section title and subtitle

zahnno:
Enables dynamic section titles and subtitles. Functions placed on title and subtitle will account for data changes using expression slots. Applied if function is implied on section property. String titles and subtitles will continue to behave as expected.

Example: 
```javascript
sections: [
 {
    name: 'mySection',
    title: function(propData) {
      return propData ? 'Set this title if propData true' : 'Set this title if propData false';
    }
 }
]
```

### DIFF
--- a/src/foam/layout/Section.js
+++ b/src/foam/layout/Section.js
@@ -8,6 +8,42 @@ foam.CLASS({
   package: 'foam.layout',
   name: 'Section',
 
+  documentation: `
+    Provides model data sectioned viewing when using the Section Views.
+    Iterates over models sections and displays each section and it's associated properties.
+
+    Example:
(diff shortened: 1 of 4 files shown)


src/foam/layout/Section.js
```diff
@@ -15,19 +15,17 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'String',
```
kgrgreer, line 18: Add comment that this can be a String or a Function.

jlhughes (approved):

jlhughes:
@zahnno Can you add the PR description as documentation on Section or SectionView please.

PR #3472, opened 2020-05-14
Np 1065/agent auth refactor

smirnovaka:


smirnovaka:
most of this is just renaming
`x.get("user")` -> ` x.get("subject").getUser()`
`x.get(“agent”)` -> `x.get("subject").getRealUser()`

These files have very slight logic changes:
```src/foam/nanos/auth/CreatedByAwareDAO.js 
src/foam/nanos/auth/LastModifiedByAwareDAO.js 
src/foam/nanos/auth/UserAndGroupAuthService.js 
src/foam/nanos/auth/twofactor/GoogleTOTPAuthService.js 
src/foam/nanos/logger/LogMessageDAO.js 
src/foam/nanos/session/Session.js 
src/foam/nanos/ticket/TicketCommentOwnerDAO.js 
 src/foam/nanos/ticket/TicketOwnerDAO.js

File: src/foam/nanos/auth/twofactor/TwoFactorAuthService.js [hunk 1]
```diff
@@ -32,7 +33,7 @@ foam.CLASS({
       name: 'check',
       javaCode: `
         Session session = x.get(Session.class);
-        User user = (User) x.get("user");
+        User user = ((Subject) x.get("subject")).getUser();
```
jlhughes, line 36: Interesting, even before I thought this would have been the real user.

jlhughes (approved):

src/foam/nanos/auth/twofactor/TwoFactorAuthService.js
smirnovaka: x.get("user") is the real user only in case when agent is null. If user is acting as somebody else then the real user is stored in "agent", the acting user is stored in "user". So currently whenever code does x.get("user") it only cares for acting user. There are places with code like `agent != null ? agent : user` or just `x.get("agent")` -  thats when people want to get the real user(the logged in one). 
`subject.getUser()` - returns the last acting user; if user never acted as somebody else, returns logged in user.
`subject.getRealUser()` - always returns logged in user

src/foam/nanos/boot/Boot.java
```diff
@@ -94,6 +95,21 @@ public Object create(X x) {
       }
     });
 
+    root_ = root_.putFactory("user", new XFactory() {
+      public Object create(X x) {
+        logger.warning("Deprecated use of x.get(\"user\")");
```
jlhughes:
```suggestion
        logger.warning(new Exception("Deprecated use of x.get(\"user\")"));
```
So it's easy to track down where it's occurring.

src/foam/nanos/boot/Boot.java
```diff
@@ -94,6 +95,21 @@ public Object create(X x) {
...
+        return ((Subject) x.get("subject")).getUser();
+      }
+    });
+
+    root_ = root_.putFactory("agent", new XFactory() {
+      public Object create(X x) {
+        logger.warning("Deprecated use of x.get(\"agent\")");
```
jlhughes:
```suggestion
        logger.warning(new Exception("Deprecated use of x.get(\"user\")"));
```

PR #3473, opened 2020-05-15
make GoogleSheetsExportService EnabledAware

olhabn:
@jlhughes @kgrgreer 
Please take a look.

Jira ticket:
NP-871/make_GoogleSheets_and_PDFGoogleSheets_enabledAware


jlhughes (commented):
The next part is to make use of this when choosing to show google export as an export option.  Either this or don't register the export driver until google oauth has been configured for a deployment.

jlhughes (approved):

PR #3478, opened 2020-05-19
Add Browser Subtitle Support To DAOControllerConfig

tharmaman:
Currently is opt-out where the user can pass an empty string in the config to render no subtitle at all or a customized string to render a subtitle that differs from the generic one (View all xyzs)


src/foam/comics/v2/DAOBrowseControllerView.js
```diff
@@ -115,18 +122,27 @@ foam.CLASS({
...
+              .start(self.Rows)
+                .addClass(self.myClass('header-container'))
+                .start(self.Cols)
+                  .start()
+                    .addClass(self.myClass('browse-title'))
+                    .add(config$browseTitle)
+                  .end()
+                  .startContext({ data: self }).tag(self.CREATE).endContext()
+                .end()
+                .callIf(config$browseTitle.length > 0, function() {
```
nanoscott: this should be checking browserSubtitle, not browserTitle.

kgrgreer:
LGTM except for Scott's comment.

src/foam/comics/v2/DAOBrowseControllerView.js
tharmaman: Whoops just pushed fix.

tharmaman:
> LGTM except for Scott's comment.

Just pushed fix to scott's comment

nanoscott (approved):

src/foam/comics/v2/DAOControllerConfig.js
```diff
@@ -56,6 +56,11 @@ foam.CLASS({
       name: 'browseTitle',
       expression: function(of) { return foam.String.pluralize(of.model_.label); }
     },
+    {
+      class: 'String',
+      name: 'browseSubtitle',
+      expression: function(of) { return "View all " + foam.String.pluralize(of.model_.label.toLowerCase()) +'.' }
```
kgrgreer: There is a 'plural' property on Model, which you could/should use instead.

tharmaman: Fixed.

PR #3481, opened 2020-05-19
Added Summary View Config Plus Newline

tharmaman:
Basically enables us to pass in DAOList view to render for notifications instead of a ScrollTableView.


nanoscott (approved):

src/foam/comics/v2/DAOControllerConfig.js
```diff
@@ -61,6 +61,20 @@ foam.CLASS({
       name: 'browseSubtitle',
       expression: function(of) { return "View all " + of.model_.plural.toLowerCase() +'.' }
     },
+    {
+      class: 'foam.u2.ViewSpecWithJava',
```
kgrgreer, line 65: Why is this needed? Can't we just refine rather than extend ViewSpec and then just use the regular ViewSpec?

PR #3482, opened 2020-05-19
Styled notifications rows

blakegreen32:
Added margins and styling to notification rows, removed unread css. Will have updates on that later. 

nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/9914


nanoscott (approved):

src/foam/nanos/notification/NotificationRowView.js
```diff
@@ -77,7 +78,7 @@
             ],
             obj: this.data
           })
-          .tag(this.NotificationCitationView, {
+          .tag(this.LiquidNotificationCitationView, {
```
nanoscott: You can't reference this in FOAM .. it's in the NANOPAY code and doesn't make sense ...

nanoscott (changes requested):

src/foam/nanos/notification/NotificationRowView.js
blakegreen32: Ok, not sure which files should be moved in this case, should NotificationRowView be in nanopay?

blakegreen32: Wasn't exactly sure how this should be handled

nanoscott: The default (FOAM) row view should just use the NotificationCitationView. There should be a NANOPAY version of the row view that overrides the original and uses the LiquidNotificationCitationView.

Is there no way to dynamically lookup the proper ciatation view depending on the object being rendered . .. like Faceted view?

kgrgreer: You can to context.register(net.nanopay.liquid.LiquidNotificationCitationView, 'foam.nanos.notification.NotificationCitationView') to replace the view that gets used by default. Just like we do to replace the default ActionView with the nanopay ActionView.

kgrgreer: But also, I don't understand why Liquid would have a different notification view.

nanoscott (approved):

src/foam/nanos/notification/NotificationCitationView.js
```diff
@@ -13,40 +13,119 @@ foam.CLASS({
...
+      
+      // display notification type instead of initiationDescription if action or entity don't exist
+      var initiationDescription = this.data.initiationDescription;
+      if ( this.data.action == '' || this.data.entity == '' ) {
+        initiationDescription = this.data.notificationType;
+      }
+
+      // truncate string if it is too long
+      var description = this.data.description;
+      if ( description.length > 70 ) {
```
nanoscott: check if this is undefined before checking the length.

jlhughes (approved):

src/foam/nanos/notification/NotificationCitationView.js
```diff
@@ -13,40 +13,119 @@ foam.CLASS({
...
+      line-height: 17px;
+    }
+    ^ .initiationDescription {
+      font-family: IBMPlexSans;
+      font-size: 14px;
+      font-weight: 600;
+      color: #1e1f21;
+      margin-left: 16px;
+    }
+    ^ .created {
```
kgrgreer, line 27: Better to do ^created instead of ^ .created because if there are any nested views that also use the 'created' class, then it will conflict.

PR #3483, opened 2020-05-19
Fixed daokey in ApprovalAwareDAO

nick-prat:


src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -302,8 +302,10 @@ foam.CLASS({
 
       // remove doesn't need an approvable to be generated since we already have the object
       if ( operation == Operations.REMOVE ){
+        String dk = getOf().getObjClass().getSimpleName();
+        dk = Character.toLowerCase(dk.charAt(0)) + dk.substring(1) + "DAO";
```
jlhughes: This seems absurd that we are dependent on this name mangling.

tharmaman: We should probably add some sort of camelCase function to string, problem is that simple name returns "LiquiditySettings" and label or pluralize returns "Liquidity Settings". Since we standardize all our client DAOs to have this camel cased format we should probably look on adding it in if we don't already have it because we need it to generate "liquiditySettings". @jlhughes @kgrgreer

jlhughes: @tharmaman not so much the name calculation, but at this point is there no other way to know what DAO is to be used?  
Regarding the name, we already have String.daoize - see dao/daoUtils.js:273.

tharmaman: So only other way is if the DAO gets passed manually to the approvableAwareDAO decorator, would have to do this by adding clientDAO property to the easyDAO which gets passed along to the approvableAwareDAO.

jlhughes: Please do the following:

1. add 'serviceName' to the approvableAwareDAO model, 
2. have 'serviceName' default to the String.daoize of the 'of'.   
3. Call x.get(getServiceName()) to get the DAO.  
4. Override it in services.jrl where needed.

PR #3487, opened 2020-05-20
Filter System V2

kennyqiyenkan:
# Filter System V2
Filter System V2 does what it implies. It adds improvements and updates to the current Filter System. 

### Reciprocal Results
Reciprocal results means that for every selected option, the other filters would then narrow down their available options.

### Result Counters
Firstly, when filters have been set, the results would be shown along with the total possible results of the DAO. Please note, the numbers are not formatted in a convenient human readable way.
Secondly, a result counter is added next to the options for Strings, and Reference filters. This shows how many results there are for that option since it comes free with GroupBy.

### String & Reference Filters
These filters now use GroupBy limits. This means that even if there are a million groups, it would only return 101 groups only. If this is the case, it will notify the user to narrow down their search criteria by filtering other parts out.

### Clearing
Sometimes, you just want to reset your filters and burn it to the ground. You don't have to go that far, but you can clear it out now.

### Advanced Mode
Advanced mode brings the capability to filter a DAO using multiple criteria. The way this works is by OR( AND (criteria), AND (criteria), ... ). Each Criteria comes equipped with a TextSearchField and the various property filters. In Advanced mode however, the property filters are not limited to the ones defined in search columns of a model. All properties that are not hidden and have a search view would be displayed. Please note: permissioned properties are still being shown.

The first criteria will always be the reflection of the filters set in simple mode if the user has set any.

Advanced mode also runs in preview mode until the user confirms it. This means that if the user chooses to leave without confirming the filters, no filtering will be done. If a user confirms the filters and wishes to modify them, they can.

### Reference Filter View
Fixes an issue with Reference Filter View not displaying any options.

### Modal Header
Modal header was modified to reflect the newer design changes meant for the platform.

### Refactors
A number of files have been moved around. This may cause some pathing issues. I've tried my best to ensure that it was minimally done.

### Missing Property Filters
Some property types have not been updated to use this newer filter system. Therefore you may encounter some filters which may not be similar as the others. They also will not be restored properly when moving between modes and UI redraws. A good example of this is Currency (the amount not the model).

### Deprecating
Some files need deprecating as well as cleanup.


src/foam/u2/filter/properties/FloatFilterView.js
```diff
@@ -1,11 +1,11 @@
 /**
  * @license
- * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
```
Nauna, line 3: this shouldn't be changed

src/foam/u2/filter/properties/BooleanFilterView.js
```diff
@@ -1,16 +1,17 @@
 /**
 * @license
-* Copyright 2019 The FOAM Authors. All Rights Reserved.
+* Copyright 2020 The FOAM Authors. All Rights Reserved.
```
Nauna: should not change year - this is done a few times in pr. I believe all changes to copyright years should be reverted.

Slack comment by Kgr about a month or so ago was to this pnt.

src/foam/u2/filter/properties/DateFilterView.js
```diff
@@ -1,15 +1,15 @@
 /**
  * @license
- * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
```
Nauna, line 3: year

Nauna (approved):
looks clean and nice. Did not test.

jlhughes (commented):
Reviewed all files for syntax and such.  Did not build or test.

src/foam/u2/filter/properties/BooleanFilterView.js
kennyqiyenkan: hmm. fair enough. Will fix.

PR #3495, opened 2020-05-25
Np 1248/client refactor agent auth

smirnovaka:


src/foam/nanos/auth/CreatedByAwareDAO.js
```diff
@@ -20,8 +20,8 @@ foam.CLASS({
         }
         return this.delegate.find_(x, obj).then(function(result) {
           if ( result == null ) {
-            obj.createdBy = x.user.id;
-            obj.createdByAgent = x.agent != null ? x.agent.id : x.user.id;
+            obj.createdBy = x.subject.user.id;
+            obj.createdByAgent = x.subject.user.id;
```
jlhughes: Why are we not using the 'realUser'?

smirnovaka: you are right, updating

jlhughes:
@penzital Everywhere property 'user' has been added, it needs to be marked 'transient'.

smirnovaka:
> @penzital Everywhere property 'user' has been added, it needs to be marked 'transient'.

will update. Hidden as well?

jlhughes:
re: hidden - yes.  
visibility: 'HIDDEN'

src/foam/dao/history/HistoryRecord.js
```diff
@@ -14,11 +14,15 @@ foam.CLASS({
...
-    'user',
+    'subject',
     'updates'
   ],
 
   properties: [
+    {
+      name: 'user',
+      getter: function() { return this.subject.user; }
+    },
```
zahnno: This is unnecessary since we're setting the user and agent info on create of history record.

src/foam/dao/history/HistoryRecord.js
```diff
@@ -14,11 +14,15 @@ foam.CLASS({
   tableColumns: [
     'timestamp',
     'objectId',
-    'user',
+    'subject',
```
zahnno: Subject doesn't exist here and isn't being set so this table column will always be missing, let's leave this class the same - please see next comment.

zahnno (changes requested):
Odd that we're defining user as a property so frequently - I think it'd be effective to set the user and realUser in the context - remove the console warning and continue to export user. Instead of exporting and setting agent within the context replace it with realUser. Solely depending on subject seems redundant and doesn't have any additional client side functionality.

jlhughes:
@zahnno I believe the intent of adding 'user' was so not to have to modify the logic of every file replacing user with subject.getUser().  Going forward we'll use subject accordingly.

src/foam/dao/history/HistoryRecord.js
smirnovaka: yes, this change is wrong. I'll revert this file.

smirnovaka: updated.

PR #3496, opened 2020-05-25
[NP-172] - SMEController Refactor

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-172
#TimeSpent: 5d

### DIFF
--- a/src/foam/nanos/menu/VerticalMenu.js
+++ b/src/foam/nanos/menu/VerticalMenu.js
@@ -30,7 +30,7 @@ foam.CLASS({
 
   css: `
   ^ input {
-    width: 220px !important;
+    width: 220px;
   }
 
   ^ .side-nav-view {
--- a/src/foam/nanos/menu/VerticalMenu.js
+++ b/src/foam/nanos/menu/VerticalMenu.js
@@ -30,7 +30,7 @@ foam.CLASS({
 
   css: `
   ^ input {
-    width: 220px;
+    width: 220px !important;
   }
 
   ^ .side-nav-view {
@@ -50,7 +50,7 @@ foam.CLASS({
 
   .foam-u2-search-TextSearchView {
     text-align: center;
-    margin: 4px 0;
+    margin: 14px 0 0;
   }
 
   ^ .foam-u2-view-TreeViewRow-label {
--- a/src/foam/nanos/theme/Theme.js
+++ b/src/foam/nanos/theme/Theme.js
@@ -105,8 +105,19 @@ foam.CLASS({
       name: 'loginDAO'
     },
     {
-      class: 'String',
-      name: 'loginImage'
+      class: 'Image',
+      name: 'loginImage',
+      displayWidth: 60,
+      view: {
+        class: 'foam.u2.MultiView',
+        views: [
+          {
+            class: 'foam.u2.tag.TextArea',
+            rows: 4, cols: 80
+          },
+          { class: 'foam.u2.view.ImageView' },
+        ]
+      },
     },
     {
       class: 'String',
--- a/src/foam/nanos/theme/Theme.js
+++ b/src/foam/nanos/theme/Theme.js
@@ -101,10 +101,6 @@ foam.CLASS({
       section: 'infoSection',
     },
     {
-      class: 'String',
-      name: 'loginDAO'
-    },
-    {
       class: 'Image',
       name: 'loginImage',
       displayWidth: 60,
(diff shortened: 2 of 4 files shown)


src/foam/nanos/theme/Theme.js
```diff
@@ -100,6 +100,23 @@ foam.CLASS({
       name: 'appName',
       section: 'infoSection',
     },
+    {
+      class: 'String',
+      name: 'loginDAO'
```
jlhughes: So this can be removed correct?

src/foam/nanos/theme/Theme.js
```diff
@@ -100,6 +100,23 @@ foam.CLASS({
...
+    },
+    {
+      class: 'String',
```
jlhughes:
```suggestion
      class: 'Image',
```
Also add a dualView with a textbox view, look elsewhere in Theme,  so there is somewhere to add the byte encoded image, and the dual view to view it.

src/foam/nanos/theme/Theme.js
```diff
@@ -100,6 +100,23 @@ foam.CLASS({
...
+      name: 'loginImage'
+    },
+    {
+      class: 'String',
+      name: 'defaultGroup'
```
jlhughes: Where is this used?

src/foam/nanos/theme/Theme.js
```diff
@@ -100,6 +100,23 @@ foam.CLASS({
...
+    },
+    {
+      class: 'Array',
+      of: 'String',
+      name: 'loginCountryChoices'
```
jlhughes: Drop this. We'll have Crunch and CountryCorridors working before the next deployment.

src/foam/nanos/menu/VerticalMenu.js
```diff
@@ -30,7 +30,7 @@ foam.CLASS({
 
   css: `
   ^ input {
-    width: 220px;
+    width: 220px !important;
```
jlhughes: Where is real offending width?

src/foam/nanos/theme/Theme.js
nanopay-arthur: We still need this as a reference for token sign in put in SignIn.js.

src/foam/nanos/theme/Theme.js
nanopay-arthur: SignUp.js uses this for setting the group on the newly created User.

src/foam/nanos/theme/Theme.js
nanopay-arthur: How would I drop this without breaking the SignUp? Wouldn't it be better to keep it as is and then refactor when Crunch/CountryCorridors is good to go?

src/foam/nanos/menu/VerticalMenu.js
nanopay-arthur: SMEStyles.js has a .foam-u2-search-TextSearchView input sets the width and takes precedent. I'll check if setting the width to 220px looks ok.

src/foam/nanos/theme/Theme.js
nanopay-arthur: Okay, will do.

src/foam/nanos/menu/VerticalMenu.js
nanopay-arthur: I figured out what the issue was, the SMEStyles.js was set after the StackView was set causing the funky css precedence.

src/foam/nanos/theme/Theme.js
jlhughes: make the default the smebusinessregistrationdao for now - this will all have to change soon.  One will create an 'ablii' account - not associated with a business, then select an Ablii 'App' such as invoicing, which will require business creation. 
And assuming signup is on foam side, make it the normal userDAO (or whatever) then refine it on the nanopay side. or extend.

src/foam/nanos/theme/Theme.js
jlhughes: This should be done elsewhere - rules based on spid and urls.  just default the code to 'sme'.  This also will have to change based on Ablii apps.

src/foam/nanos/theme/Theme.js
jlhughes: on foam, don't restrict.  On nanopay refine or extend and hard code.  It will be fixed by 'application' wide country corridor settings. if we add it to Theme it will never be removed.

src/foam/nanos/theme/Theme.js
nanopay-arthur: Right, makes sense.

src/foam/nanos/theme/Theme.js
nanopay-arthur: Sounds good.

src/foam/nanos/theme/Theme.js
nanopay-arthur: Got it!

jlhughes (approved):

PR #3500, opened 2020-05-26
Revert "put notification to notificationTemplateDAO rather than notificationD…"

jlhughes:
Reverts foam-framework/foam2#3499


jlhughes:
This is incorrect.

PR #3502, opened 2020-05-26
Optimize basicStrategizer, elaborate on documentation

zahnno:
Optimize basicStrategizer, elaborate on documentation, grammar correction

### DIFF
--- a/src/foam/strategy/BasicStrategizer.js
+++ b/src/foam/strategy/BasicStrategizer.js
@@ -37,7 +37,10 @@ foam.CLASS({
           throw new RuntimeException("A desired model id must be specified.");
         }
 
-        Predicate predicate = SafetyUtil.isEmpty(target)
+        if ( strategyPredicate != null  ) {
+          predicate = (Predicate) strategyPredicate;
+        } else {
+          Predicate predicate = SafetyUtil.isEmpty(target)
           ? AND(
               EQ(StrategyReference.DESIRED_MODEL_ID, desiredModelId),
               EQ(StrategyReference.TARGET, "")
@@ -49,9 +52,6 @@ foam.CLASS({
                 EQ(StrategyReference.TARGET, target)
               )
             );
-
-        if ( strategyPredicate != null  ) {
-          predicate = (Predicate) strategyPredicate;
         }
 
         DAO strategyDAO = ((DAO) getStrategyDAO()).inX(x);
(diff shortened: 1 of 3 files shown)


src/foam/strategy/BasicStrategizer.js
```diff
@@ -37,7 +37,10 @@ foam.CLASS({
           throw new RuntimeException("A desired model id must be specified.");
         }
 
-        Predicate predicate = SafetyUtil.isEmpty(target)
+        if ( strategyPredicate != null  ) {
+          predicate = (Predicate) strategyPredicate;
+        } else {
+          Predicate predicate = SafetyUtil.isEmpty(target)
```
kgrgreer: Does this compile? Since you use the variable on line 41 that you don't declare until line 43?

src/foam/u2/view/FObjectView.js
```diff
@@ -88,12 +88,18 @@ foam.CLASS({
     },
     {
       class: 'Object',
-      name: 'persistantData',
-      documentation: 'Allows passed in data to persist to selected class instances.'
+      name: 'persistentData',
+      documentation: `
+        Allows passed in data to persist to selected class instances.
```
kgrgreer, line 93: I still don't understand. Why don't you just copy over all parameters?

zahnno: default values and factories of the new class won't be applied by copying.

kgrgreer: Couldn't you just walk all of the axioms and copy them yourself to avoid this problem?

        for ( PropertyInfo p : props ) {
          try {
            if ( p.isSet(obj) ) p.set(this, p.get(obj));
          } catch (ClassCastException e) {
            try {
              PropertyInfo p2 = (PropertyInfo) obj.getClassInfo().getAxiomByName(p.getName());
              if ( p2 != null ) {
                p.set(this, p2.get(obj));
              }
            } catch (ClassCastException e2) {}
          }
        }

zahnno: java? this is client side code.

kgrgreer: Then even simpler on the client, just use obj.copyFrom(obj2).

src/foam/strategy/StrategizerService.js
```diff
@@ -36,7 +36,7 @@ foam.INTERFACE({
       return foam.u2.view.FObjectView.create({
         data: X.data.exampleProp,
         of: foam.nanos.auth.User,
-        persistantData: { website: X.data.user.website },
+        persistentData: { website: X.data.user.website },
```
kgrgreer, line 39: This is no longer used. Use copyData instead.

kgrgreer:
Is this still needed? If so, please resolve conflicts and update to use copyData. Thx

PR #3504, opened 2020-05-26
add loop over exception

hchoura:
@lchanmann Please can you review.

### DIFF
--- a/src/foam/core/CompoundException.java
+++ b/src/foam/core/CompoundException.java
@@ -32,7 +32,10 @@ public class CompoundException
 
   public void maybeThrow() {
     if ( exceptions_.size() != 0 ) {
-      throw this;
+      for ( int i = 0 ; i < exceptions_.size() ; i++ ) {
+        System.out.println( "This is an error =======> " + ( (Throwable) exceptions_.get(i) ) );
+        throw (RuntimeException) exceptions_.get(i);
+      }
     }
   }
 


src/foam/core/CompoundException.java
```diff
@@ -32,7 +32,10 @@ public void add(Throwable t) {
 
   public void maybeThrow() {
     if ( exceptions_.size() != 0 ) {
-      throw this;
+      for ( int i = 0 ; i < exceptions_.size() ; i++ ) {
```
jlhughes, line 35: @hchoura no.
Do this output when AppConfig.MODE is TEST.
after this block 
'throw this' as before.

src/foam/core/CompoundException.java
```diff
@@ -32,7 +32,10 @@ public void add(Throwable t) {
...
+        System.out.println( "This is an error =======> " + ( (Throwable) exceptions_.get(i) ) );
+        throw (RuntimeException) exceptions_.get(i);
```
jlhughes, line 37: this makes no sense.

jlhughes (changes requested):

src/foam/core/CompoundException.java
```diff
@@ -32,7 +32,10 @@ public void add(Throwable t) {
 
   public void maybeThrow() {
     if ( exceptions_.size() != 0 ) {
-      throw this;
```
lchanmann, line 35: `throw this;` makes more sense than looping the exception list then throws the first one.

jlhughes:
superseded by another PR

PR #3507, opened 2020-05-27
[NP-813] custom table columns add ability to display property of property reference

olhabn:
ticket: https://nanopay.atlassian.net/browse/NP-813
nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/10121


src/foam/comics/v2/userfeedback/UserFeedback.js
```diff
@@ -26,7 +26,15 @@ foam.CLASS({
     {
       class: 'FObjectProperty',
       of: 'foam.comics.v2.userfeedback.UserFeedback',
-      name: 'next'
+      name: 'next',
+      javaToCSV: `
+        outputter.outputValue("next");
+      `,
+      javaToCSVLabel: `
+      //to remove "infinite" loop
```
olhabn: this changes were made as PrefixedCSVOutputter was in infinite loop

src/foam/dao/CSVSink.js
```diff
@@ -68,14 +68,19 @@ foam.CLASS({
           .setProps(getProps())
           .build();
       `
-    }
+    },
```
olhabn, line 71: please see CSVTableExportDriver, which is alternative for current CSVDriver. the difference between CSVTableExportDriver and CSVDriver is that CSVTableExportDriver uses Projection ( which returns array of values for provided properties ) while CSVDriver uses CSVSink which retrieves values using CSVOutputterImpl which calls property's toCSV method on objects, which are returned on DAO

so please take a look at CSVTableExportDriver, and let me know which you prefer
thanks

src/foam/lib/csv/CSVOutputterImpl.js
```diff
@@ -14,6 +14,9 @@ foam.CLASS({
 
   javaImports: [
     'foam.core.*',
+    'foam.nanos.column.ColumnConfigToPropertyConverter',
```
olhabn, line 17: same here 
please see comment https://github.com/foam-framework/foam2/pull/3507/files#r431170996

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
```diff
@@ -0,0 +1,370 @@
...
+      name: 'filterExportedProps',
+      documentation: 'returns propNames that are properties of the class and are networkTransient',
+      code: function(of, propNames) {
+        var props = of.getAxiomsByClass(foam.core.Property);
+        var allColumnNames = props.map(p => p.name);
+        if ( ! propNames ) {
+          return props.filter(p => p.networkTransient).map(p => p.name);
+        } props = [];
+        return propNames.filter(n => { 
+          return allColumnNames.includes(n.split('.')[0]);// && self.returnProperty(of, n).networkTransient;
```
olhabn: networkTransient is commented out as not a lot of properties are networkTransient.

src/foam/nanos/column_config/NestedPropertiesExpression.js
```diff
@@ -0,0 +1,220 @@
...
+        if ( !e ) return null;
+        return e.f(obj);
+      },
+      javaCode: `
+        Expr e = returnDotExprForNestedProperty(getObjClass(), getNestedProperty().split("\\\\."), 0);
+        if ( e == null )
+          return null;
+        //have no idea why but this context allows to execute find method, but object's context doesn't
+        //need to investigate
+        ((FObject)obj).setX(getX());
```
olhabn, line 52: is it legal to do such a thing?

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
olhabn: should we review networkTransient of some of the columns? or should I just uncomment that?

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -278,10 +307,17 @@ foam.CLASS({
               }).
 
               // Render the table headers for the property columns.
-              forEach(columns_, function([axiomOrColumnName, overrides]) {
-                var column = typeof axiomOrColumnName === 'string'
-                  ? view.of.getAxiomByName(axiomOrColumnName)
-                  : axiomOrColumnName;
+              forEach(columns_, function([property, overrides]) {
```
olhabn: Should we use projection instead of retrieving value for a property for UnstyledTableView?

src/foam/mlang/mlang.js
```diff
@@ -3240,10 +3246,32 @@ foam.CLASS({
     {
       name: 'f',
       code: function(o) {
```
olhabn, line 3248: works only for reference being arg1

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
olhabn: should we set networkTransient default value to true maybe?

src/foam/nanos/column_config/NestedPropertiesExpression.js
```diff
@@ -0,0 +1,217 @@
...
+      code: function(obj) {
+        var e =  this.returnDotExprForNestedProperty(this.objClass, this.nestedProperty.split('.'), 0);
+        if ( !e ) return null;
+        return e.f(obj);
+      },
+      javaCode: `
+        Expr e = returnDotExprForNestedProperty(getObjClass(), getNestedProperty().split("\\\\."), 0);
+        if ( e == null )
+          return null;
+        return e.f(obj);
```
olhabn: this expression context is different from next expression's context  ( which is the same as obj's context ). and this expression context allows to execute find methods but object context doesn't. and I'm not sure what's going on

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
jlhughes: I would include 'networkTransient' so that these columns are ignored.

src/foam/nanos/column_config/TableColumnOutputter.js
```diff
@@ -0,0 +1,110 @@
...
+          if ( foam.Array.isInstance(val) ) {
+            var stringArr = [];
+            for ( var i = 0 ; i < val.length ; i++ ) {
+              stringArr.push(await this.valueToString(val[i]));
+            }
+            return stringArr.join(' ');
+          }
+          if ( foam.core.UnitValue.isInstance(prop) ) {
+            if ( unitPropName ) {
+              if ( prop.unitPropName === "destinationCurrency" ) {
```
olhabn: is there anyway to find targetDAO?

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -349,157 +409,212 @@ foam.CLASS({
...
-                    var checked = newValueSlot.get();
+                tbodyElement.forEach(values.array, function(val) {
+                  var thisObjValue;
+                  var tableRowElement = this.E();
+                  tableRowElement.
+                  addClass(view.myClass('tr')).
+                  on('mouseover', function() {
+                    if ( !thisObjValue ) {
+                      dao.find(val[0]).then(v => {
+                      thisObjValue = v;
```
olhabn, line 453: I have a problem with context in `find` here (CABankAccount.js) for Bank Accounts and I have some problems with figuring it out. Please let me know if you have any suggestions, I'll keep investigating meanwhile

src/foam/nanos/column_config/TableColumnOutputter.js
```diff
@@ -0,0 +1,110 @@
...
+      var columnConfig = x.columnConfigToPropertyConverter;
+      var values = [];
+      for ( var propName of  propNames ) {
+        values.push(await columnConfig.returnValueForPropertyName(x, of, propName, obj));
+      }
+      return values;
+    },
+    {
+      name: 'objectToTable',
+      code: async function(x, of, props, obj) {
```
kgrgreer: Why is this async?

olhabn: fixed

src/foam/mlang/mlang.js
```diff
@@ -2696,6 +2701,30 @@ foam.CLASS({
 
         getArray().add(a);
       `
+    },
```
olhabn, line 2684: the reason this changes was made is that TTLSelectCachingDAO performs toString on projection for building cache key. and if projection returns only class name on toString method call the cache for the previous projection was used despite the fact that expressions changed

src/foam/u2/view/UnstyledTableView.js
jlhughes: BankAccounts are still stored in accountsDAO - is that what you were needing?

src/foam/nanos/column_config/TableColumnOutputter.js
jlhughes: @nanoMichal is there a way to find targetDAO for unitPropName and if there is any place where we map UnitValue to the target DAO, for which unitPropName is primary key?

nanoMichal: No I dont think so. There is no UnitDAO, but we talked about making a unitDAO which would just redirect to the proper dao (currency, security etc) by having a decorator which would check instanceof and direct appropriatly. Would this help?

src/foam/mlang/mlang.js
```diff
@@ -2696,6 +2701,30 @@ foam.CLASS({
...
+    {
+      name: 'toString',
+      code: function() {
+        var projectionString = this.cls_.name + '(';
+        var exprString = [];
+        for ( var expr of this.exprs ) {
+          exprString.push(expr.toString());
```
kgrgreer: I don't think this step is necessary. I think you could just join() this.exprs.

src/foam/mlang/mlang.js
```diff
@@ -2696,6 +2701,30 @@ foam.CLASS({
...
+        }
+        projectionString += exprString.join(', ') + ')';
+        return projectionString;
+      },
+      javaCode: `
+        StringBuilder sb = new StringBuilder(getClassInfo().getId() + "(");
+        StringJoiner joiner = new StringJoiner(", ");
```
kgrgreer: You can also specify the start and end of the string int the StringJoiner which would avoid having to create the StringBuilder and having to copy the data to it.

src/foam/mlang/mlang.js
```diff
@@ -3240,10 +3277,40 @@ foam.CLASS({
...
         return this.arg2.f(this.arg1.f(o));
       },
       javaCode: `
-        return getArg2().f(getArg1().f(obj));
+        StringBuilder sb = new StringBuilder("find");
+        PropertyInfo p1 = (PropertyInfo) getArg1();
+        FObject obj1;
+        if ( p1 instanceof AbstractFObjectPropertyInfo ) {
+          Object val = getArg1().f(obj);
+          if ( val == null )
```
kgrgreer: A bit nicer: return val == null ? : getArg2().f(val);

But your code, so you can choose.

src/foam/mlang/mlang.js
```diff
@@ -3240,10 +3277,40 @@ foam.CLASS({
...
+        sb.append(arr);
+        try {
+          obj1 = (FObject)obj.getClass().getMethod(sb.toString(), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
+        } catch (Exception e) {
+          return null;
+        }
+        if ( obj1 == null ) return null;
+        try {
+          return getArg2().f(obj1);
+        } catch( Exception e ) {
```
kgrgreer: Would be better to catch Throwable?

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
```diff
@@ -0,0 +1,223 @@
...
+          if ( ( p == null && i != 0 ) || ci == null )
+            break;
+          p = (PropertyInfo) ci.getAxiomByName(propNames[i]);
+          if ( i != propNames.length - 1 ) {
+            try {
+              Class cls;
+              if ( p instanceof foam.core.AbstractFObjectPropertyInfo ) {
+                cls = p.getValueClass();
+              } else {
+                char[] arr = p.getName().toCharArray();
```
kgrgreer: On the JS side we have foam.util.String.capitalize(). I think it would make the code easier to understand if we add a similar function to foam.util.StringUtil.java on the java side.

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
```diff
@@ -0,0 +1,223 @@
...
+                arr[0] = Character.toUpperCase(arr[0]);
+                sb.append(arr);
+                String s = sb.toString();
+                Method m = ci.getObjClass().getMethod(s, foam.core.X.class);
+                cls = m.getReturnType();
+                sb.setLength(4);
```
kgrgreer: What is this line for?

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
```diff
@@ -0,0 +1,223 @@
...
+            }
+          }
+        }
+        return p;
+      `
+    },
+    {
+      name: 'returnPropertyAndObject',
+      type: 'foam.nanos.column.ColumnPropertyValue',
+      documentation: `returns property and object on which such property's functions as f or toCSV can be called`,
```
kgrgreer: No need for ` when a single-line string without interpolation. Just use ' which parses faster.

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
```diff
@@ -0,0 +1,223 @@
...
+          p = (PropertyInfo) ci.getAxiomByName(propNames[i]);
+
+          if ( i != propNames.length - 1 ) {
+            try {
+              Class cls;
+              if ( p instanceof foam.core.AbstractFObjectPropertyInfo ) {
+                obj1 = (FObject) p.f(obj1);
+                cls = p.getValueClass();
+              } else {
+                char[] arr = p.getName().toCharArray();
```
kgrgreer: Create a capitalize() helper function.

src/foam/nanos/column_config/NestedPropertiesExpression.js
```diff
@@ -0,0 +1,222 @@
...
+          return p;
+        }
+        
+        StringBuilder sb = new StringBuilder("find");
+        try {
+          Class cls;
+          if ( p instanceof foam.core.AbstractFObjectPropertyInfo ) {
+            cls = p.getValueClass();
+          } else {
+            char[] arr = p.getName().toCharArray();
```
kgrgreer: same

src/foam/nanos/column_config/TableColumnOutputter.js
kgrgreer: Couldn't we use polymorphism to let the property format the value without us having to know the details? Would work for UnitValue but also for the other types as well and would be mor extensible.

src/foam/nanos/column_config/TableColumnOutputter.js
kgrgreer: It still says async.

olhabn: it now awaits for another method

src/foam/nanos/column_config/ColumnConfigPropertyValue.js
olhabn: it's for StringBuilder clean up and setting it to "find" for next property to use

kgrgreer: Okay, add a comment to make this clear. Thx

src/foam/nanos/column_config/TableColumnOutputter.js
olhabn: > Couldn't we use polymorphism to let the property format the value without us having to know the details? Would work for UnitValue but also for the other types as well and would be mor extensible.

I'm not sure if I get it correctly. But my understanding was that unitValue property contains value and fk to unitProp, which stored in different daos. We can use unitValue property the way it is on line 38, but the only object that knows how to format unitValue well is unitProp ( please see line 36 ). 
eg Account has home balance, which is unit value, and it has value, which is 30, and denomination, which is "CAD". Only "CAD" Currency object, retrieved from currency DAO, knows how to format it well.
Let me know if I misunderstood
So I'm not sure about polymorphism and if I get it correctly

olhabn: > No I dont think so. There is no UnitDAO, but we talked about making a unitDAO which would just redirect to the proper dao (currency, security etc) by having a decorator which would check instanceof and direct appropriatly. Would this help?

That would be great. not sure though if we need it that much

src/foam/nanos/column_config/CSVTableOutputter.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.column',
```
jlhughes: The file system path mush match the package.
please change the directory name to 'column' from column_config

jlhughes (changes requested):

src/foam/nanos/column_config/CSVTableOutputter.js
olhabn: fixed

src/foam/nanos/column/ColumnConfigPropertyValue.js
```diff
@@ -0,0 +1,218 @@
...
+  methods: [
+    {
+      name: 'filterExportedProps',
+      documentation: 'returns propNames that are properties of the class and aren\'t networkTransient',
+      code: function(of, propNames) {
+        var props = of.getAxiomsByClass(foam.core.Property);
+        var allColumnNames = props.map(p => p.name);
+        if ( ! propNames ) {
+          return props.filter(p => ! p.networkTransient ).map(p => p.name);
+        } props = [];
```
kgrgreer: This is odd formatting. Why is there a statement on the same line as a }?

src/foam/mlang/mlang.js
```diff
@@ -2696,6 +2701,22 @@ foam.CLASS({
 
         getArray().add(a);
       `
+    },
+    {
+      name: 'toString',
+      code: function() {
+        var projectionString = this.cls_.name + '(' + this.exprs.join(',') + ')';
```
kgrgreer: Just combine these two lines since you don't use the var.

src/foam/nanos/export/GoogleSheetsExportDriver.js
```diff
@@ -30,34 +30,43 @@ foam.CLASS({
...
       stringArray.push(metadata.map(m => m.columnLabel));
-      var values = await self.outputter.outputArray(sink.array, metadata);
+      
+      var values = [ await this.outputter.objToArrayOfStringValues(X, obj.cls_, [ obj ], metadata.map(p => p.propName)) ];
       stringArray = stringArray.concat(values);
 
+      sheetId = await X.googleSheetsDataExport.createSheet(X, stringArray, metadata);
+      if ( ! sheetId || sheetId.length === 0)
+        return '';
+      var url = `https://docs.google.com/spreadsheets/d/${sheetId}/edit#gid=0`;
```
kgrgreer: Could just return.

src/foam/nanos/column/ColumnConfigPropertyValue.js
olhabn: fixed

src/foam/mlang/mlang.js

src/foam/nanos/export/GoogleSheetsExportDriver.js
olhabn: fixed

PR #3509, opened 2020-05-27
remove issued date,leave to created

yij793:
Nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/9953
notes: we have createAware which will set Created in notification when notification get generated


zahnno (approved):

jlhughes (approved):

jlhughes:
There are no UI components that depend or show issueDate?

PR #3513, opened 2020-05-27
Currency Input View

kennyqiyenkan:
# Currency Input View
An input view that automatically formats the user's input as they type it in according to the provided currency, or currency selection.

## Warning
Currently, the handling of DisplayMode seems to be broken. This affects this view as well as it depends on DisplayMode to display the value or input correctly.

### Clean user input
Allows a user to easily read the value they are inputting based on their choice of, or provided currency. **Values provided by the user can only be numerical.**

### RW / RO Modes
If the DisplayMode is set to RW, the input is formatted without symbols. This is from modifying the format function in Currency. RO would format the value with symbols with respect to direction as well based on the implementation in Currency.

### Currency
I modified Currency's format function to include a second property, `hideId`. This is to allow for formatting of the value without adding the ID and symbol of the currency to it. That said, this may have made `formatPrecision` unnecessary depending on its usage. It has been added in such a way that existing usage of format would not be broken.

## Exciting! How Do I Use It?
To use this view, just instantiate it like any other view. However, `currencyId` or `contingentProperty` must be provided for the view to work as intended. `contingentProperty` is the name of the property that will handle the currency selection. 
As an example: There exists a model containing `currency` and `amount` properties. `currency` uses a Rich Choice View for the user to select a currency. `amount` is using the Currency Input View but the currency is contingent on `currency`. To enable this, just include `contingentProperty: 'currency'` as a parameter when instantiating the Currency Input View.

If the currency is fixed, you can provide the currency's id (eg: "CAD") instead. Just include `currencyId: 'CAD'` as a parameter when instantiating the Currency Input View. It will still depend on your currencyDAO having that currency as a record for it to work.

### DIFF
--- a/src/foam/core/Currency.js
+++ b/src/foam/core/Currency.js
@@ -161,7 +161,8 @@
         },
         {
           class: 'Boolean',
-          name: 'hideId'
+          name: 'hideId',
+          documentation: 'If true, will not add symbol or ID to formatted currency'
         }
       ],
       type: 'String',
--- a/src/foam/core/Currency.js
+++ b/src/foam/core/Currency.js
@@ -202,6 +202,28 @@
 
         return formatted;
       `
+    },
+    {
+      name: 'formatPrecision',	
+      documentation: `
+        Given a number, display it as a currency using the appropriate
+        precision. Use a period '.' for the decimal place and do not
+        include commas or currency symbols.
+        Suitable for use when exporting to CSV.
+      `,
+      code: function(amount) {
+        return (amount/Math.pow(2, this.precision)).toFixed(this.precision);
+      },
+      args: [
+        {
+          class: 'foam.core.UnitValue',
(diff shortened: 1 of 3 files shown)


nanoscott (approved):

src/foam/core/Currency.js
```diff
@@ -154,6 +158,10 @@
         {
           class: 'foam.core.UnitValue',
           name: 'amount'
+        },
+        {
+          class: 'Boolean',
+          name: 'hideId'
```
kgrgreer: Please add documentation:

src/foam/u2/view/CurrencyInputView.js
```diff
@@ -0,0 +1,161 @@
+foam.CLASS({
```
kgrgreer, line 7: Add copyright header.

PR #3543, opened 2020-06-03
DO NOT MERGE: Add XLIFFTranslation

hchoura:


src/foam/i18n/Locale.js
```diff
@@ -0,0 +1,57 @@
+/**
```
jlhughes, line 1: This is the old license/header - is it not @kgrgreer

src/foam/i18n/Locale.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+foam.CLASS({
+  package: 'foam.i18n',
+  name: 'Locale',
+
+  properties: [
+    {
+      class: 'String',
+      name: 'locale',
```
jlhughes, line 18: can we not make Enums for these three?  They are never going to change - or very very infrequently.  Simplify human setup - can't make a mistake.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -512,6 +513,14 @@ foam.CLASS({
       if ( ! lastTheme || lastTheme.id != this.theme.id ) this.useCustomElements();
     },
 
+    //TODO add the option to the user to select a language
```
jlhughes, line 14583: default considerations - we can provide a default for a domain Theme, then Group, and finally the User.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -285,7 +288,28 @@ foam.CLASS({
     },
     'currentMenu',
     'lastMenuLaunched',
-    'webApp'
+    'webApp',
+    {
+      name   : 'XLIFFTranslation',
```
jlhughes, line 294: Is this XLIFFTTranslation or LocaleTranslation? 
Is there something XLIFF speciific here?

jlhughes: We should have a LocaleTranslation Interface that our XLIFFTransaction implements.

PR #3544, opened 2020-06-03
[NP-641] step crunch wizard

KernelDeimos:
The following changes are in this PR
- Add StepWizardletView
- Add CircleIndicator element (used by StepWizardletView and StepWizardletStepsView)
- Add SimpleActionDialog (used by StepWizardletView)
- Move CRUNCH wizard to a popup
- Move intercept view to a popup
- Cleanup intercept view

### DIFF
--- a/src/foam/u2/borders/MarginBorder.js
+++ b/src/foam/u2/borders/MarginBorder.js
@@ -1,0 +1,41 @@
+
+foam.CLASS({
+  package: 'foam.u2.borders',
+  name: 'MarginBorder',
+  extends: 'foam.u2.Element',
+
+  documentation: `
+    A border which adds margin to all sides except the bottom.
+  `,
+
+  properties: [
+    {
+      name: 'margin',
+      class: 'Int',
+      value: 24
+    },
+    {
+      name: 'marginUnit',
+      class: 'String',
+      value: 'px'
+    },
+
+    {
+      name: 'marginExpr',
+      class: 'String',
+      expression: function (margin, marginUnit) {
+        return '' + margin + marginUnit;
+      }
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this
+        .style({
+          'margin': this.marginExpr$
+        })
+        ;
+    }
+  ]
+});
--- a/src/foam/u2/borders/MarginBorder.js
+++ b/src/foam/u2/borders/MarginBorder.js
@@ -10,35 +10,22 @@ foam.CLASS({
   extends: 'foam.u2.Element',
 
   documentation: `
-    A border which adds margin to all sides except the bottom.
+    A border which adds equal margin to all sides
   `,
 
   properties: [
     {
       name: 'margin',
-      class: 'Int',
-      value: 24
-    },
-    {
-      name: 'marginUnit',
       class: 'String',
-      value: 'px'
+      value: '24px'
     },
-
-    {
-      name: 'marginExpr',
-      class: 'String',
-      expression: function (margin, marginUnit) {
-        return '' + margin + marginUnit;
-      }
-    }
   ],
 
   methods: [
     function initE() {
       this
         .style({
-          'margin': this.marginExpr$
+          'margin': this.margin$
         })
         ;
     }
--- a/src/foam/u2/borders/MarginBorder.js
+++ b/src/foam/u2/borders/MarginBorder.js
@@ -1,3 +1,8 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
 
 foam.CLASS({
   package: 'foam.u2.borders',
--- a/src/foam/u2/dialog/SimpleActionDialog.js
+++ b/src/foam/u2/dialog/SimpleActionDialog.js
@@ -54,27 +54,26 @@ foam.CLASS({
     function initE() {
       var self = this;
       this.SUPER();
-      this
-        .startContext(this.context)
+      this.startContext(this.context)
         .start().addClass(this.myClass())
-        .start().addClass('Container')
-          .start().addClass('headerTitle').add(this.title).end()
-          .start().addClass('content')
-            .add(this.body)
-          .end()
-          .start().addClass('actions')
-            .forEach(this.actions, function (action) {
-              var a = action.clone();
-              a.code = function (...args) {
-                self.closeDialog();
-                return action.code.call(action, ...args);
-              }
-              this.add(a);
-            })
+          .start().addClass('Container')
+            .start().addClass('headerTitle').add(this.title).end()
+            .start().addClass('content')
+              .add(this.body)
+            .end()
+            .start().addClass('actions')
+              .forEach(this.actions, function (action) {
+                var a = action.clone();
+                a.code = function (...args) {
+                  self.closeDialog();
+                  return action.code.call(action, ...args);
+                }
+                this.add(a);
+              })
+            .end()
           .end()
         .end()
-        .end()
-        .end();
+      .end();
     },
   ]
 });
--- a/src/foam/u2/dialog/SimpleActionDialog.js
+++ b/src/foam/u2/dialog/SimpleActionDialog.js
@@ -1,0 +1,80 @@
+foam.CLASS({
+  package: 'foam.u2.dialog',
+  name: 'SimpleActionDialog',
+  extends: 'foam.u2.Controller',
+
+  imports: [
+    'closeDialog'
+  ],
+
+  css: `
+    ^ .headerTitle {
+      font-family: Lato;
+      font-size: 24px;
+      font-weight: 900;
+      font-style: normal;
+      font-stretch: normal;
+      line-height: 1.5;
+      letter-spacing: normal;
+    }
+
+    ^ .Container > *:not(:last-of-type) {
+      margin-bottom: 24px;
+    }
+    
+    ^ {
+      margin: 24px;
+    }
+  `,
+
+  properties: [
+    {
+      name: 'title',
+      class: 'String'
+    },
+    {
+      name: 'body',
+      class: 'String'
+    },
+    {
+      name: 'context'
+    },
+    {
+      name: 'actions',
+      class: 'Array'
+    },
+    {
+      name: 'closeOnChoice',
+      class: 'Boolean',
+      value: true
+    }
+  ],
+
+  methods: [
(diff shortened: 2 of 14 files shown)


KernelDeimos:
Previous PR comments available here: https://github.com/Nauna/foam2/pull/78

src/foam/u2/borders/MarginBorder.js
```diff
@@ -0,0 +1,41 @@
+
+foam.CLASS({
+  package: 'foam.u2.borders',
+  name: 'MarginBorder',
```
Nauna, line 9: A little uncertain about using this. 

The intercept popUp needs a UI upgrade - at the very least we will need some ux that will identify what is happening with 'pending' status capabilities. Which makes me think this is excessive for a temp solution.

Then I also wonder - considering our most recent dev-channel conversations over margins. Its just another reason I feel this is excessive.

Also - I don't see the reason to not add this as css - why a model.

Nauna: ^ not a blocking comment

KernelDeimos: > considering our most recent dev-channel conversations over margins

This is the reason I added this. `MarginBorder` allows me to take the concern of margin outside of the intercept view.

> why a model

CSS axiom would work too, but I figure if there's a CardBorder it makes sense to have a MarginBorder seeing as CardBorder also could be a CSS axiom.

src/foam/u2/dialog/SimpleActionDialog.js
```diff
@@ -0,0 +1,79 @@
+foam.CLASS({
+  package: 'foam.u2.dialog',
+  name: 'SimpleActionDialog',
+  extends: 'foam.u2.Controller',
+
```
Nauna, line 16: This needs a foam header and documentation for this as well.

src/foam/u2/borders/MarginBorder.js
KernelDeimos: Also just remembered, if this is a CSS axiom then the margin has to be defined statically. Consider GMail's different options for the user to change padding - if we ever did something similar it would be trivial to do with MarginBorder but rather complicated to do with a MarginCSSAxiom

src/foam/u2/wizard/StepWizardletController.js
```diff
@@ -0,0 +1,195 @@
+foam.CLASS({
```
Nauna, line 7: needs header

src/foam/u2/dialog/SimpleActionDialog.js
KernelDeimos: Added

src/foam/u2/wizard/StepWizardletStepsView.js
```diff
@@ -0,0 +1,121 @@
+foam.CLASS({
```
Nauna, line 7: header

src/foam/u2/wizard/StepWizardletView.js
```diff
@@ -0,0 +1,207 @@
...
+      }
+    },
+    {
+      name: 'close',
+      label: 'Save for Later',
+      code: function(x) {
+        this.data.saveProgress().then(() => {
+          this.onClose(x);
+        }).catch(e => {
+          console.error(e);
```
Nauna, line 179: this is a good idea, but based on convo in standup this morning - looks like we will defaulting toastNotifications to console.log().

So either we remove this now and rely on future implementation of 'this.notify()' or we do something to remember to remove this redundancy.

personally I'd vote for removing this console.log.

src/foam/u2/wizard/StepWizardletView.js
```diff
@@ -0,0 +1,207 @@
...
+      name: 'goNext',
+      label: 'next',
+      code: function(x) {
+        this.data.next().then((isFinished) => {
+          console.log('finished?', isFinished);
+          if ( isFinished ) {
+            this.onClose(x);
+          }
+        }).catch(e => {
+          console.error(e);
```
Nauna, line 200: same comment as above - recommend removing this

src/foam/u2/wizard/StepWizardletView.js
```diff
@@ -0,0 +1,207 @@
...
+      code: function() {
+        this.data.back();
+      }
+    },
+    {
+      name: 'goNext',
+      label: 'next',
+      code: function(x) {
+        this.data.next().then((isFinished) => {
+          console.log('finished?', isFinished);
```
Nauna: think this console.log() is unnecessary

src/foam/u2/wizard/util/CircleIndicator.js
```diff
@@ -0,0 +1,135 @@
...
+        this.start('img')
+          .attr('src', this.icon$)
+          .style({
+            'width': '' + this.size + 'px',
+            'height': '' + this.size + 'px',
+          })
+        .end();
+      }
+
+      if ( this.label !== '' ) {
```
Nauna: I believe `if ( this.label ) { ... }` is equivalent to `if ( this.label !== '' ) { ... }`

js empty string is falsy

src/foam/u2/wizard/StepWizardletView.js
KernelDeimos: I think a `console.error` is a good redundancy because we'll never display `e` in a notification to the user.

src/foam/u2/wizard/StepWizardletView.js
KernelDeimos: oops, good catch

src/foam/u2/wizard/util/CircleIndicator.js
KernelDeimos: It's not equivalent but it is a superset. Seems reasonable though since other falsy values shouldn't be accepted either

src/foam/u2/wizard/StepWizardletView.js
jlhughes: we should console.log/warn/error appropriately.  If we know it's an error then console.error.

src/foam/u2/borders/MarginBorder.js
```diff
@@ -0,0 +1,46 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.borders',
+  name: 'MarginBorder',
+  extends: 'foam.u2.Element',
+
+  documentation: `
+    A border which adds margin to all sides except the bottom.
```
kgrgreer: Why not the bottom?

src/foam/u2/borders/MarginBorder.js
```diff
@@ -0,0 +1,46 @@
...
+      class: 'Int',
+      value: 24
+    },
+    {
+      name: 'marginUnit',
+      class: 'String',
+      value: 'px'
+    },
+
+    {
```
kgrgreer: Why not just let the user type in whatever they want. They they can specify just one margin or 2 or 4 margins, like with regular CSS?

KernelDeimos: If I understand, you're suggesting a `.style({ margin: '24px' })` can do the same thing, right?

I was thinking if margin size is added to themes then this model could be modified to reflect that.

src/foam/u2/borders/MarginBorder.js
kgrgreer: I'm cool with this approach.

src/foam/u2/wizard/StepWizardletController.js
```diff
@@ -0,0 +1,201 @@
...
+          this.highestIndex = wizardletIndex;
+          this.subStack.push({
+            class: 'foam.u2.detail.SectionView',
+            section: section,
+            data$: wizardlet.data$,
+          })
+
+          // Automatically push the next section if this one is
+          // unavailable.
+          let slot = this.sectionAvailableSlots
```
kgrgreer: Why is this split onto two lines?

src/foam/u2/wizard/util/CircleIndicator.js
```diff
@@ -0,0 +1,135 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.wizard.util',
```
kgrgreer: Is this reall wizard specific? Would a more general package be better?

src/foam/u2/wizard/util/CircleIndicator.js
```diff
@@ -0,0 +1,135 @@
...
+          'color': this.textColor$,
+          'border': '' + this.borderThickness + 'px solid',
+          'cursor': this.ExpressionSlot.create({
+            obj: this,
+            code: function (clickable) {
+              return clickable ? 'pointer' : 'default';
+            }
+          }),
+        })
+        .on('mouseover', () => {
```
kgrgreer: Would it be easier to just use the :hover CSS selector?

src/foam/u2/borders/MarginBorder.js
kgrgreer: Just let them specify 'margin' as a string, rather than breaking into unit an int and a unit. Then the user can also specify: 'auto', 'inherit', or something like: '25px 50px 75px 100px;' or '75px 100px;'
It's just more flexible if you try to be less helpful.

src/foam/u2/wizard/util/CircleIndicator.js
KernelDeimos: This is me forgetting to move this model.

How is `foam.u2.tag.CircleIndicator`?

src/foam/u2/borders/MarginBorder.js
KernelDeimos: oops, that's not true after I found a better way to solve something else

src/foam/u2/wizard/StepWizardletController.js
KernelDeimos: don't remember, will fix

src/foam/u2/wizard/util/CircleIndicator.js
KernelDeimos: Not sure if I can use a selector here since I want the colour properties attached to slots. For example you can use CircleIndicator as a button in which case you may want the `:hover` colour to be the same as the default if the button turns disabled.

Nauna (approved):

PR #3548, opened 2020-06-03
Np 360/json formatter

smirnovaka:


src/foam/dao/F3FileJournal.js
```diff
@@ -0,0 +1,93 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2020

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2020

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+
+          public void executeUnderLock() {
+            old = dao.find_(x, id);
+            dao.put_(x, obj);
+          }
+
+          public void executeJob() {}
+
+          public void endJob() {
+            try {
```
kgrgreer, line 174: Here to line 199 we want to move to executeJob() so that this code is done in parallel. The problem with the old Outputter was that it couldn't be used by multiple threads at once, but Formatter doesn't have that problem.

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+  ],
+
+  axioms: [
+    {
+      name: 'javaExtras',
+      buildJavaClass: function (cls) {
+        cls.extras.push(`
+          protected static Pattern COMMENT = Pattern.compile("(/\\\\*([^*]|[\\\\r\\\\n]|(\\\\*+([^*/]|[\\\\r\\\\n])))*\\\\*+/)|(//.*)");
+
+          protected static ThreadLocal<StringBuilder> sb = new ThreadLocal<StringBuilder>() {
```
kgrgreer: Better than having a thread-local StringBuilder is to have a thread-local Formatter so we can format in parallel.

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+            old = dao.find_(x, id);
+            dao.put_(x, obj);
+          }
+
+          public void executeJob() {}
+
+          public void endJob() {
+            try {
+              record_ = ( old != null ) ?
+                getOutputter().stringifyDelta(old, obj) :
```
kgrgreer: stringifying creates garbage, just reset the thread-local formatter then on line 207, instead of outputting record_ we can output formatter.builder(). This let's us output directly from the StringBuilder without ever generating a String / garbage.

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+        }
+
+        public void executeJob() {
+          try {
+            // TODO: Would be more efficient to output the ID portion of the object.  But
+            // if ID is an alias or multi part id we should only output the
+            // true properties that ID/MultiPartID maps too.
+            FObject toWrite = (FObject) obj.getClassInfo().newInstance();
+            toWrite.setProperty("id", obj.getProperty("id"));
+            record_ = getOutputter().stringify(toWrite);
```
kgrgreer: Don't need to stringify. Just use the Formatter directly.

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+        writer.newLine();
+      `
+    },
+    {
+      name: 'writeComment_',
+     // synchronized: true,
+      javaThrows: [
+        'java.io.IOException'
+      ],
+      args: [
```
kgrgreer: You can use the new args syntax:
args: [ 'Context x', 'foam.core.FObject obj' ]

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+    },
+    {
+      name: 'write_',
+     // synchronized: true,
+      javaThrows: [
+        'java.io.IOException'
+      ],
+      args: [
+        {
+          class: 'String',
```
kgrgreer: Make this be a CharSequence instead of a String. Then you can still specify String but also a StringBuilder which avoids the extra step of converting SB's to Strings.
https://docs.oracle.com/javase/7/docs/api/java/lang/CharSequence.html

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+        }
+      ],
+      javaCode: `
+        write_(sb.get()
+          .append(prefix)
+          .append("p(")
+          .append(record)
+          .append(")")
+          .append(c)
+          .toString());
```
kgrgreer: See comment below about making write_() take a CharSequence then you don't need this .toString().

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+      `
+    },
+    {
+      name: 'remove',
+      type: 'FObject',
+      args: [
+        { name: 'x',      type: 'Context' },
```
kgrgreer: Can convert to new format. Here and elsewhere.

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+          type: 'String'
+        }
+      ],
+      javaCode: `
+        write_(sb.get()
+          .append(prefix)
+          .append("r(")
+          .append(record)
+          .append(")")
+          .toString());
```
kgrgreer: .toString() not needed

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,485 @@
...
+          if ( line == null ) return null;
+          if ( ! line.equals("p({") && ! line.equals("r({") ) return line;
+          StringBuilder sb = new StringBuilder();
+          sb.append(line);
+          while( ! line.equals("})") ) {
+            if ( (line = reader.readLine()) == null ) break;
+            sb.append("\\n");
+            sb.append(line);
+          }
+          return sb.toString().trim();
```
kgrgreer: We should make the StringPStream take a CharSequence instead. Then we don't need to call toString() here.

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,383 @@
...
+          .append(" (")
+          .append(user.getId())
+          .append(") at ")
+          .append(getTimeStamper().createTimestamp()));
+      `
+    },
+    {
+      name: 'getEntry',
+      documentation: 'retrieves a meaningful unit of text from the journal',
+      type: 'String',
```
kgrgreer: Switch to CharSequence so you don't have to call toString on line 331.

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,383 @@
...
+          }
+
+          public void executeUnderLock() {
+            old = dao.find_(x, id);
+            dao.put_(x, obj);
+          }
+
+          public void executeJob() {
+            try {
+              if ( old != null ) form.outputDelta(old, obj);
```
kgrgreer: format if/else "properly"

src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -0,0 +1,383 @@
...
+  ],
+
+  methods: [
+    {
+      name: 'put',
+      type: 'FObject',
+      args: [ 'Context x', 'String prefix', 'DAO dao', 'foam.core.FObject obj' ],
+      javaCode: `
+        final Object id = obj.getProperty("id");
+        JSONFObjectFormatter form = formatter.get();
```
kgrgreer: I found the name 'form' strange. I was expecting a form. How about 'fmt'?

src/foam/dao/AbstractF3FileJournal.js
kgrgreer: It violates the FOAM style guide which is based on the Google style guide:

https://github.com/foam-framework/foam/wiki/CodingStyle

if (old != null ) {
  form.outputDelta(old, obj);
} else {
  form.output(obj);
}

PR #3549, opened 2020-06-03
Dig Updates

nanoscott:
Lots of minor fixes for DIG


jlhughes (approved):
Thanks.

src/foam/nanos/dig/DIG.js
```diff
@@ -11,14 +11,19 @@ foam.CLASS({
...
+  implements: [
+    'foam.mlang.Expressions',
+  ],
 
   tableColumns: [
     'id',
     'daoKey',
     'cmd',
-    'format',
-    'owner'
```
kgrgreer, line 21: Why removing owner?

src/foam/nanos/dig/DIG.js
```diff
@@ -31,68 +36,80 @@ foam.CLASS({
...
 
+  imports: [
+    'nSpecDAO'
+  ],
+
   properties: [
-    'id',
     {
-      class: 'String',
+      name: 'id',
```
kgrgreer, line 45: I think we should add class: 'String'. How was this working on the java side without this?

src/foam/nanos/dig/DIG.js
```diff
@@ -31,68 +36,80 @@ foam.CLASS({
...
+        return ( cmd == 'SELECT' || cmd == 'REMOVE' ) ? foam.u2.DisplayMode.RW : foam.u2.DisplayMode.HIDDEN;
+      }
+  },
+  {
       class: 'String',
       name: 'q',
-      label: 'Query'
+      label: 'Select Query',
+      visibility: function(cmd) {
+        return (cmd == 'SELECT') ? foam.u2.DisplayMode.RW : foam.u2.DisplayMode.HIDDEN;
```
kgrgreer, line 98: Should be visible for 'REMOVE" also, shouldn't it?

kgrgreer: Or do we only accept remove by primary key? (Which is less useful, but safer.)

src/foam/nanos/dig/DIG.js
```diff
@@ -128,33 +145,18 @@ foam.CLASS({
...
-            url += "subject=" + encodeURIComponent(subject);
-          }
-        }
         if ( q ) {
           url += query ? "&" : "?";
           query = true;
           url += "q=" + encodeURIComponent(q);
         }
-        if ( limit >= 0 && limit != Number.MAX_SAFE_INTEGER ) {
+        if ( limit > 0 && limit != Number.MAX_SAFE_INTEGER ) {
```
kgrgreer: If they don't submit the limit in the url connection, what value will limit be?

src/foam/nanos/dig/DIG.js
nanoscott: owner isn't set or used. not sure who put it in there but it should actually be createdBy.

src/foam/nanos/dig/DIG.js
nanoscott: ID is already defined in the parent class .. DefaultHttpParameters.

src/foam/nanos/dig/DIG.js
nanoscott: yeah .. we require a PK.

src/foam/nanos/dig/DigSelectOperation.java
```diff
@@ -0,0 +1,70 @@
...
+import foam.mlang.MLang;
+import foam.mlang.predicate.Predicate;
+import foam.nanos.http.*;
+import foam.util.SafetyUtil;
+
+import java.io.PrintWriter;
+import java.util.List;
+import javax.servlet.http.HttpServletResponse;
+
+public class DigSelectOperation extends DigOperation
```
kgrgreer: I would split these along the other axis. I would have DIGFormatDrivers (or something like that) and then have each driver support all of the operations required for that format: find, selelct, remove, etc. Then, if I wanted to add a new format it would be self-contained rather than having to update each operation class. If the driver were modelled, then it could also have format specific fields (ie. pretty-print, indent, quote-keys, etc.).

src/foam/nanos/dig/DIG.js
nanoscott: unlimited.

src/foam/nanos/dig/DIG.js
```diff
@@ -104,7 +132,7 @@ foam.CLASS({
       label: 'Snippet',
       documentation: 'show a specific type of request would look like in a given language.',
       view: { class: 'foam.nanos.dig.DigSnippetView' },
-      expression: function(key, data, email, subject, daoKey, cmd, format, q, limit, dataFile) {
+      expression: function(key, data, daoKey, cmd, format, q, limit, skip) {
```
jlhughes, line 135: Did we lose 'email' support - if provided we send the results to the email. 
Actually, we don't need it here as we are not changing the snippet because of the email.

jlhughes (approved):
Nice.

src/foam/nanos/dig/drivers/DigFormatDriver.js
```diff
@@ -0,0 +1,267 @@
...
+  ],
+
+  methods: [
+    {
+      name: 'parseFObjects',
+      type: 'List',
+      javaThrows: [
+        'java.lang.Exception'
+      ],
+      args: [
```
kgrgreer, line 59: You could use the new args: format:
  args: [ 'X x', 'DAO dao', 'String data' ] 
if you like.

PR #3551, opened 2020-06-04
if all pre-req caps granted or pending - then top cap pending

Nauna:
Consider Canadian Onboarding,
When Onboarding pre-req capabilities all either granted or pending, this changes the Onboarding capability to pending.


noodlemoodle:
thinking this logic can be merged with checkPrereqs since we are doing pretty similar operations twice.
maybe we can have only one method that return the "lowest" status out of all the prereqs, in an order like [expired, action_required, pending, granted]. so that we can call it once in put_when checking prereqs and save the returned status, then later we can use the saved variable to check if should be granted or pending or action required?

Nauna:
So now instead of `checkPrereqs` we have `checkPrereqsChainedStatus`

So now checkPrereqsChainedStatus returns one of 3 status.
1) GRANTED: if all pre-reqs are in granted status
2) PENDING: if all pre-reqs are either granted or pending status
3) ACTION_REQUIRED: if not any of the above

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -311,40 +317,42 @@ foam.CLASS({
           type: 'foam.core.FObject'
         }
       ],
-      type: 'Boolean',
+      type: 'CapabilityJunctionStatus',
       documentation: `Check if prerequisites of a capability is fulfilled`,
```
noodlemoodle: please update

noodlemoodle:
haven't tested yet, but logic of it looks good to me

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
Nauna: done - thx for this!

jlhughes:
> So now instead of `checkPrereqs` we have `checkPrereqsChainedStatus`
> 
> So now checkPrereqsChainedStatus returns one of 3 status.
> 
> 1. GRANTED: if all pre-reqs are in granted status
> 2. PENDING: if all pre-reqs are either granted or pending status
> 3. ACTION_REQUIRED: if not any of the above

so is 2 - granted and at least one pending?

src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -311,40 +317,47 @@ foam.CLASS({
           type: 'foam.core.FObject'
         }
       ],
-      type: 'Boolean',
-      documentation: `Check if prerequisites of a capability is fulfilled`,
+      type: 'CapabilityJunctionStatus',
+      documentation: `
+        Check statuses of all preRequist capabilities - returning:
+        GRANTED: if all pre-reqs are in granted status
+        PENDING: if all pre-reqs are either granted or pending status
```
jlhughes: At least one pre-req is still pending.

jlhughes (approved):

src/foam/nanos/crunch/CapabilityCategory.js
```diff
@@ -75,7 +75,6 @@ foam.CLASS({
       javaThrows: ['AuthorizationException'],
       javaCode: `
         if ( getVisibilityCondition() == null ) {
-          ( getDefaultAuthorizer()).authorizeOnRead(x, this);
           return;
```
Nauna, line 78: ### Just a note on this:
Either we added **capabilitycategory.read.*** to each group that needed to see crunch categories - or we accept that everyone should see categories.

Latter makes more sense - discussed originally by @noodlemoodle - just appending this to this pr

PR #3553, opened 2020-06-04
[NP-1426] fix duplicate intercepts

KernelDeimos:


src/foam/u2/crunch/CapabilityIntercept.js
```diff
@@ -0,0 +1,42 @@
+foam.CLASS({
```
Nauna, line 7: header

src/foam/u2/crunch/CrunchController.js
```diff
@@ -36,6 +38,25 @@ foam.CLASS({
...
   ],
 
+  properties: [
+    {
+      name: 'activeIntercepts',
+      class: 'Array',
+      documentation: `
+        Since permissions may be checked during asynchronous calls,
+        it is possible that the same intercept view will be requested
+        twice in a short period of time. Keeping a a map of active
```
Nauna:
```suggestion
        twice in a short period of time. Keeping a map of an active
```

Nauna: would recommend not commiting - and just reading over this text to fix - above is only one suggestion but 
> intercept **view** is done to prevent two intercept views being
> open**ed** for the same permission (as this would be confusing for

NOT BLOCKING

KernelDeimos: Fixed the typo of typing `a` twice,

but regarding the rest of the text I wrote it in plural form intentionally. If two unrelated intercepts occur then both should open, and in this case both would be considered "active"

jlhughes (approved):

Nauna (approved):

PR #3554, opened 2020-06-04
[Merge after 3553] [NP-641] Implement clickable sections in StepWizardletView

KernelDeimos:
Probably do not merge until #3553 is merged. It still works without that change but it's much better with it.

### DIFF
--- a/src/foam/u2/wizard/StepWizardletController.js
+++ b/src/foam/u2/wizard/StepWizardletController.js
@@ -143,12 +143,12 @@ foam.CLASS({
       });
 
       // If the first screen has no available sections, move next and repeat.
-      var f;
-      f = () => {
+      var skipEmptyWizardlets;
+      skipEmptyWizardlets = () => {
         var currentWizardletIndex =
           this.screenIndexToSection(this.subStack.pos)[0];
         if ( this.countAvailableSections(currentWizardletIndex) < 1 ) {
-          return this.next().then(f);
+          return this.next().then(skipEmptyWizardlets);
         }
       };
       f();
@@ -232,13 +232,13 @@ foam.CLASS({
       return true;
     },
     function skipTo(screenIndex) {
-      var f;
-      f = () => {
+      var skipToScreenRecur;
+      skipToScreenRecur = () => {
         if ( this.subStack.pos !== screenIndex ) {
-          return this.next().then(f);
+          return this.next().then(skipToScreenRecur);
         }
       };
-      return f();
+      return skipToScreenRecur(); // call recursive function
     },
     function back() {
       this.subStack.back();
@@ -256,7 +256,9 @@ foam.CLASS({
         }
         i += nSections;
       }
-      return null;
+      throw new Error(
+        `Tried to get wizard screen at index ${i} but it doesn't exist`
+      );
     },
 
     function sectionToScreenIndex(wizardletIndex, sectionIndex) {
--- a/src/foam/u2/wizard/StepWizardletController.js
+++ b/src/foam/u2/wizard/StepWizardletController.js
@@ -151,7 +151,7 @@ foam.CLASS({
           return this.next().then(skipEmptyWizardlets);
         }
       };
-      f();
+      skipEmptyWizardlets();
     },
     function saveProgress() {
       var p = Promise.resolve();
--- a/src/foam/u2/wizard/StepWizardletController.js
+++ b/src/foam/u2/wizard/StepWizardletController.js
@@ -98,6 +98,33 @@ foam.CLASS({
         return sections.reduce(
           (sum, wizardletSections) => sum + wizardletSections.length, 0);
       }
+    },
+    {
+      name: 'canGoBack',
+      class: 'Boolean',
+      documentation: `
+        If the first screen has no available sections, then the back button
+        should be disabled.
+      `,
+      expression: function (subStack$pos) {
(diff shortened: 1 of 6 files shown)


src/foam/u2/wizard/StepWizardletController.js
```diff
@@ -178,6 +216,30 @@ foam.CLASS({
...
+        (total, sectionAvailable$) =>
+          sectionAvailable$.get() ? total + 1 : total,
+        0
+      );
+    },
+    function canSkipTo(wizardletIndex) {
+      let currentWizardletIndex =
+        this.screenIndexToSection(this.subStack.pos)[0];
+      for ( let w = currentWizardletIndex ; w <= wizardletIndex ; w++ ) {
+        if ( ! this.wizardlets[w].readyToSubmit() ) return false;
```
jlhughes: Where is readyToSubmit set?  Seems like this is just 'validated'?

KernelDeimos: I did this with the idea that an implementor of Wizardlet can choose what it means for something to be ready to submit. Actually this could even come up soon if we want to let the user click `next` for incomplete data but not for invalid values (both of which wouldn't pass the validation predicate)

Nauna: So this is a working concept for something we don't need and may use.

Eric is correct - this has come up in the next action, however we've decided to not do anything for next - allow no validation check, which became the necessity of https://github.com/foam-framework/foam2/pull/3560/files

@jlhughes please let us know your thoughts - I'm ok with having this level of complexity for potential future use. AND Joel - nice catch, thank you!

src/foam/u2/wizard/StepWizardletController.js
```diff
@@ -98,6 +98,33 @@ foam.CLASS({
...
+    },
+    {
+      name: 'canGoBack',
+      class: 'Boolean',
+      documentation: `
+        If the first screen has no available sections, then the back button
+        should be disabled.
+      `,
+      expression: function (subStack$pos) {
+        let indices = this.screenIndexToSection(subStack$pos);
```
Nauna, line 110: screenIndexToSection(...) can return null - what are you doing to handle this call - not only here but everywhere this(screenIndexToSection) call happens?

src/foam/u2/wizard/StepWizardletController.js
```diff
@@ -114,6 +141,17 @@ foam.CLASS({
         section: this.sections[0][0],
         data$: this.wizardlets[0].data$,
       });
+
+      // If the first screen has no available sections, move next and repeat.
```
Nauna, line 145: I'd actually like to see the function f - be named something more relevant and remove the comment on 145

src/foam/u2/wizard/StepWizardletController.js
```diff
@@ -178,6 +216,30 @@ foam.CLASS({
...
+    function canSkipTo(wizardletIndex) {
+      let currentWizardletIndex =
+        this.screenIndexToSection(this.subStack.pos)[0];
+      for ( let w = currentWizardletIndex ; w <= wizardletIndex ; w++ ) {
+        if ( ! this.wizardlets[w].readyToSubmit() ) return false;
+      }
+      return true;
+    },
+    function skipTo(screenIndex) {
+      var f;
```
Nauna: a function that returns a function? Where is this used? and why is this formatting necessary? If this isn't just an oversight, please document.

src/foam/u2/wizard/StepWizardletController.js
jlhughes: 'readyToSubmit' will morph into lots of custom code.  'validate' is our standard on models.   The wizard steps don't have validate but I would like to stick to the calling convention.   Open to debate.

src/foam/u2/wizard/StepWizardletController.js
KernelDeimos: Should never happen (requires invalid subStack position) but I'll modify screenIndexToSection to throw an error if the stack position isn't okay.

src/foam/u2/wizard/StepWizardletController.js
KernelDeimos: `return f()` - it's not returning a function, it's a recursive call (needs to be because of the promise). I'll add a comment

src/foam/u2/wizard/StepWizardletController.js
KernelDeimos: I'll add a better function name so that the stack trace looks better if there's ever an error here, but why remove the comment?

src/foam/u2/wizard/StepWizardletController.js
KernelDeimos: Unable to test any solutions because of an unrelated regression. Adding `[DNM]` tag and investigating now. :/

KernelDeimos: nvm I just forgot to rename a function in a place after addressing another PR comment

KernelDeimos: Moved this to `validate` and added a `mustBeValid` property - so if validation is needed before clicking next that will be set to `true`.

src/foam/u2/wizard/StepWizardletController.js
Nauna: Just felt the comment could be virtually defined with a good name. Your call

PR #3558, opened 2020-06-04
Number Shortener

kennyqiyenkan:
# Number Shortener
Ever had numbers so long they broke your views? Meet the Number Shortener.

## Features
Shortens a number if needed, to the provided precision. If a currency is provided, it will take that currency's precision into account as well and formats accordingly. Brackets are as follows:

- Thousands = K
- Millions = M
- Billions = B
- Trillions = T

## How-To
Just add the path to requires and use `this.NumberShortener.shortenNumber(...)`


nanoscott (approved):

jlhughes (approved):
Will be useful for PM and OM counts.

PR #3559, opened 2020-06-04
[NP-1436] Add UCJ put for no-data capabilities

KernelDeimos:
This fixes an issue where a most-parent capability doesn't get a UCJ stored despite needing one to track status.

### DIFF
--- a/src/foam/u2/crunch/CrunchController.js
+++ b/src/foam/u2/crunch/CrunchController.js
@@ -19,7 +19,7 @@ foam.CLASS({
     'ctrl',
     'prerequisiteCapabilityJunctionDAO',
     'stack',
-    'user',
+    'subject',
     'userCapabilityJunctionDAO'
   ],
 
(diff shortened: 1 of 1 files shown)


src/foam/u2/crunch/CrunchController.js
```diff
@@ -19,6 +19,7 @@ foam.CLASS({
     'ctrl',
     'prerequisiteCapabilityJunctionDAO',
     'stack',
+    'user',
```
Nauna: since here- can you make this import 'subject'

this.user == this.subject.user

src/foam/u2/crunch/CrunchController.js
```diff
@@ -89,14 +90,24 @@ foam.CLASS({
...
+          ).map(
+            cap => this.CapabilityWizardlet.create({
+              capability: cap
+            }).updateUCJ()
+          ))
+
+        ]);
+      }).then(capabilitiesSectionsTuple => {
+        let capabilities = capabilitiesSectionsTuple[0];
+        let sections = capabilitiesSectionsTuple[1];
```
jlhughes, line 111: possible out of bounds?

KernelDeimos: shouldn't be, call above to Promise.all always returns an array of 2 elements. I'll add a comment to document this

PR #3560, opened 2020-06-04
[DMN (yet)] [NP-641 Update] Keep next enabled

KernelDeimos:
This PR keeps the wizard's next button enabled

### DIFF
--- a/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
+++ b/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
@@ -100,7 +100,7 @@ foam.CLASS({
       }
     },
     {
-      name: 'readyToSubmit',
+      name: 'validate',
       code: function () {
         return true;
       }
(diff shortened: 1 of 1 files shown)


Nauna:
We are assuming this is a quick fix till we identify possible functionality - given the option to not check anything on 'next' or to check partial validation on 'next'.

realized that we can't check full validation - given the discussion surrounding the 'reviewed' prop explicit to SigningOfficers.

I approve this.

jlhughes (approved):

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -98,6 +98,12 @@ foam.CLASS({
           return this;
         });
       }
+    },
+    {
+      name: 'readyToSubmit',
```
jlhughes: This was renamed to mustBeValid on the foam side was it not?

KernelDeimos: Oops, you're right - will update

KernelDeimos: Adding DNM tag - need to check something

KernelDeimos:
This isn't needed anymore since `mustBeValid` defaults to false

PR #3567, opened 2020-06-05
[DO_NOT_MERGE] Help message

noodlemoodle:
add help message to property. 
usage
```
    {
      section: 'signingOfficerQuestionSection',
      name: 'isSigningOfficer',
      class: 'Boolean',
      helpMessage: 'IS_SIGNING_OFFICER_HELP', // name of message
    },
```

### DIFF
--- a/src/foam/u2/detail/SectionedDetailPropertyView.js
+++ b/src/foam/u2/detail/SectionedDetailPropertyView.js
@@ -231,7 +231,6 @@ foam.CLASS({
           var errorSlot = prop.validateObj && prop.validationTextVisible ?
             this.data.slot(prop.validateObj) :
             foam.core.ConstantSlot.create({ value: null });
-          // debugger;
 
           return self.E()
             .start(self.Rows)
--- a/src/foam/u2/detail/SectionedDetailPropertyView.js
+++ b/src/foam/u2/detail/SectionedDetailPropertyView.js
@@ -250,7 +250,7 @@ foam.CLASS({
                   })
                 .end()
                 .callIf(prop.help || ( prop.helpMessage && prop.sourceCls_[prop.helpMessage] ), function() {
-                  var help = prop.helpMessage && prop.sourceCls_[prop.helpMessage] ? prop.sourceCls_[prop.helpMessage] : prop;
+                  var help = prop.helpMessage && prop.sourceCls_[prop.helpMessage] ? prop.sourceCls_[prop.helpMessage] : prop.help;
                   this.start()
                     .addClass(self.myClass('tooltip'))
                     .start({
--- a/src/foam/u2/detail/SectionedDetailPropertyView.js
+++ b/src/foam/u2/detail/SectionedDetailPropertyView.js
@@ -231,6 +231,7 @@ foam.CLASS({
           var errorSlot = prop.validateObj && prop.validationTextVisible ?
             this.data.slot(prop.validateObj) :
             foam.core.ConstantSlot.create({ value: null });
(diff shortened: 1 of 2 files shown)


src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -231,6 +231,7 @@ foam.CLASS({
           var errorSlot = prop.validateObj && prop.validationTextVisible ?
             this.data.slot(prop.validateObj) :
             foam.core.ConstantSlot.create({ value: null });
+          // debugger;
```
noodlemoodle:
```suggestion
```

Nauna:
SO Im thinking eventually after this merge - we systematically remove 'help' from sections and use 'helpMessages' solely.

Other comment - this looks good to me - although I'd wonder if 'helpMessage' shouldn't have an adapt or something to help and avoid the UI update in SectionedDetailPropertyView

jlhughes:
Is it help or helpMessage which is defined in foam for the FObject or FObjectProperty?

noodlemoodle:
I'm not sure what you mean.. help message was added in this pr, and help existed on property but not on fojbect @jlhughes

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -249,7 +249,8 @@ foam.CLASS({
                     this.enableClass(self.myClass('error'), errorSlot);
                   })
                 .end()
-                .callIf(prop.help, function() {
+                .callIf(prop.help || ( prop.helpMessage && prop.sourceCls_[prop.helpMessage] ), function() {
+                  var help = prop.helpMessage && prop.sourceCls_[prop.helpMessage] ? prop.sourceCls_[prop.helpMessage] : prop;
```
kgrgreer: Why return 'prop' if the expression is false?

noodlemoodle:
```suggestion
                  var help = prop.helpMessage && prop.sourceCls_[prop.helpMessage] ? prop.sourceCls_[prop.helpMessage] : prop.help;
```

noodlemoodle: sorry, should be prop.help. updating @kgrgreer

jlhughes:
> I'm not sure what you mean.. help message was added in this pr, and help existed on property but not on FObject @jlhughes

'help' is defined on foam.core.Property.  Can we use this property property please or is there a reason for new property?

noodlemoodle:
> > I'm not sure what you mean.. help message was added in this pr, and help existed on property but not on FObject @jlhughes
> 
> 'help' is defined on foam.core.Property. Can we use this property property please or is there a reason for new property?

yes, but as far as I know 'help' will only take a String value as argument, whereas this will take the name of a message defined on the source class of the property, which we want for translation purposes right?

jlhughes:
> > > I'm not sure what you mean.. help message was added in this pr, and help existed on property but not on FObject @jlhughes
> > 
> > 
> > 'help' is defined on foam.core.Property. Can we use this property property please or is there a reason for new property?
> 
> yes, but as far as I know 'help' will only take a String value as argument, whereas this will take the name of a message defined on the source class of the property, which we want for translation purposes right?

@hchoura @kgrgreer are we planning on support property properties such as 'help' in our locale work - now or future?

kgrgreer:
Do not merge.

kgrgreer:
Let's not do this way.

PR #3570, opened 2020-06-05
Move script per run event capturing to own model and dao.

jlhughes:
Move per run script info such as lastRun date to another model/DAO so that the script model itself is not 'updated' on each run.


src/foam/nanos/script/Script.js
```diff
@@ -198,7 +200,8 @@ foam.CLASS({
       }
```
nanoscott, line 200: is max output length still necessary to limit? since it's all in memory now it shouldn't matter.

src/foam/nanos/script/Script.js
```diff
@@ -318,6 +321,7 @@ foam.CLASS({
           setLastDuration(pm.getTime());
           ps.flush();
           setOutput(baos.toString());
+          ((DAO) x.get("scriptEventDAO")).put_(x, this);
```
nanoscott, line 324: Test.js overrides this .. did you want this to happen at the end of a test too?

nanoscott: so does TestRunnerScript, but i can see why we wouldn't save those.

src/foam/nanos/script/ScriptEvent.js
```diff
@@ -0,0 +1,67 @@
...
+  ],
+
+  properties: [
+    {
+      name: 'id',
+      class: 'String',
+      visibility: 'RO'
+    },
+    {
+      name: 'type',
```
kgrgreer: Can you document what this field means?

kgrgreer: And/or, use a more meaningful name.

src/foam/nanos/script/Script.js
jlhughes: If I remember correctly, I believe this max output has more to do with HTTP limits rather than memory/disk concerns.

src/foam/nanos/script/Script.js
jlhughes: Re: save event at end of test - no.  And since it overrides runScript and does not save, I'm good with current implementation.

src/foam/nanos/script/ScriptEvent.js
jlhughes: Added. This is similar to the use of 'type' in some other models - useful in tableviews to distinguish subclasses of models shown in the same DAO.

nanoscott (approved):

src/foam/nanos/cron/services
```diff
@@ -3,12 +3,13 @@ p({
   "name":"cronDAO",
   "serve":true,
   "serviceScript":"""
+    mdao = new foam.dao.MDAO(foam.nanos.cron.Cron.getOwnClassInfo());
     return new foam.dao.EasyDAO.Builder(x)
       .setOf(foam.nanos.cron.Cron.getOwnClassInfo())
+      .setMdao(mdao)
+      .setInnerDAO(new net.nanopay.security.HashingJDAO(x, "SHA-256", true, false, mdao, "cronjobs"))
```
lchanmann: looks like net.nanopay.security.HashingJDAO is from nanopay side.

jlhughes: crap crap crap.  we need to use hashing dao where we may have the need to support instance specific scripts, crons. ...  i have having to duplicate on the nanopay side for this stuff. ok.

src/foam/nanos/script/ScriptEventDAO.js
```diff
@@ -0,0 +1,32 @@
...
+      name: 'put_',
+      javaCode: `
+      if ( obj instanceof Script ) {
+        Script script = (Script) obj;
+        ScriptEvent event = new ScriptEvent();
+        event.setOwner(script.getId());
+        event.setLastRun(script.getLastRun());
+        event.setLastDuration(script.getLastDuration());
+        event.setOutput(script.getOutput());
+        event.setType(script.getClass().getSimpleName());
```
lchanmann: will event.type be lost on replay since it's storageTrasient:true?

src/foam/nanos/script/services
```diff
@@ -3,16 +3,48 @@ p({
...
+    mdao = new foam.dao.MDAO(foam.nanos.script.Script.getOwnClassInfo());
     return new foam.dao.EasyDAO.Builder(x)
-      .setPm(true)
-      .setGuid(true)
       .setOf(foam.nanos.script.Script.getOwnClassInfo())
-      .setJournalType(foam.dao.JournalType.SINGLE_JOURNAL)
-      .setJournalName("scripts")
-      .setDecorator(new foam.nanos.script.ScriptRunnerDAO(null))
+      .setMdao(mdao)
+      .setInnerDAO(new net.nanopay.security.HashingJDAO(x, "SHA-256", true, false, mdao, "scripts"))
```
lchanmann: net.nanopay.security.HashingJDAO

src/foam/nanos/script/services
```diff
@@ -3,16 +3,48 @@ p({
...
+p({
+  "class":"foam.nanos.boot.NSpec",
+  "name":"scriptEventDAO",
+  "serve":true,
+  "serviceScript":"""
+    mdao = new foam.dao.MDAO(foam.nanos.script.ScriptEvent.getOwnClassInfo());
+    return new foam.dao.EasyDAO.Builder(x)
+      .setOf(foam.nanos.script.ScriptEvent.getOwnClassInfo())
+      .setMdao(mdao)
+      .setInnerDAO(new net.nanopay.security.HashingJDAO(x, "SHA-256", true, false, mdao, "scriptEvents"))
```
lchanmann: net.nanopay.security.HashingJDAO

src/foam/nanos/cron/services
jlhughes: Replaced with JDAO.  I'll duplicate service entries on the nanopay side.

src/foam/nanos/script/ScriptEventDAO.js
jlhughes: ah, yes, good call.  copy and paste.

src/foam/nanos/script/services
jlhughes: yes, all replaced with JDAO.

lchanmann (approved):

src/foam/nanos/script/ScriptEvent.js
jlhughes: renamed to scriptType.

PR #3571, opened 2020-06-05
Notification Message Updates

blakegreen32:
Updated NotificationMessage to new designs. Added description property that will add a description under the message that is optional when creating notification or calling this.notify(message, type, description);

Ticket: https://nanopay.atlassian.net/browse/NP-1010
nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/10044


kgrgreer:
LGTM

src/foam/u2/dialog/NotificationMessage.js
```diff
@@ -20,111 +20,204 @@ foam.CLASS({
...
+    }
+    ^close-icon:hover {
+      opacity: 1;
     }
   `,
 
+  properties: [
+    {
+      class: 'String',
+      name: 'type'
```
jlhughes, line 140: Can this use the same enum from foam.log.LogLevel?

blakegreen32: Its similar but I don't think so due to the fact that we need a success state, also the coloring isn't the same unless we plan to unify them. I'm up for the discussion though.

src/foam/u2/images/round-check-circle-green.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20">
```
Nauna, line 1: so I've looked this up and actually found - we currently have 4 green check-marks - all the same ... :(
/foam/u2/images/checkmark-small-green.svg
/net/nanopay/images/checkmark-large-green.svg
/net/nanopay/images/checkmark-small-green.svg
/foam/nanos/controller/checkmark-small-green.svg

image

blakegreen32: haha oh wow, ok i'll remove the others thanks for checking that

src/foam/u2/dialog/NotificationMessage.js
jlhughes: re: colours - I don't really care what shade of green log info messages are. 
re: success - for notification? I don't see this anywhere - and  clearly shows the issue with free form text for state.  How would anyone to know what the 'type's are supposed to be.  I see notifications states identical to log messages: info, warning, error.  Where/how is success used?   @nanoscott tagging you here to comment as well if possible. Thanks.

nanoscott: info and success are basically the same .. scenarios where this is something we would want to display:
- Record was successfully created
- Approval request was approved
- etc.
All of these things are fine being shown as info messages. So using LogLevel should be fine.

src/foam/u2/dialog/NotificationMessage.js
```diff
@@ -20,111 +20,204 @@ foam.CLASS({
...
       var self = this;
 
+      var img;
+      if ( this.type === 'error' ) {
+        img = 'images/round-error-red.svg';
+      } else if ( this.type === 'warning' ) {
+        img = 'images/baseline-warning-yellow.svg';
+      } else if ( this.type === 'info' ) {
+        img = 'images/round-error-blue.svg';
+      } else {
```
nanoscott, line 167: what is this else supposed to represent? should the default be 'info'? so the fall through should be to images/round-error-blue.svg?

src/foam/u2/dialog/NotificationMessage.js
```diff
@@ -20,111 +20,204 @@ foam.CLASS({
...
+              .attrs({ src: img })
+            .end()
+            .start().addClass(this.myClass('content'))
+              .enableClass(this.myClass('error-content'), this.type === 'error')
+              .enableClass(this.myClass('warning-content'), this.type === 'warning')
+              .enableClass(this.myClass('info-content'), this.type === 'info')
+              .callIfElse(foam.String.isInstance(this.message), function() {
+                this.add(self.message);
+              }, function() {
+                this.tag(self.message);
```
nanoscott, line 180: add a console.log(self.message); below this tag .. and maybe below the description below as well if it is not empty.

jlhughes:
@blakegreen32 if you could update LogLevel:
1. convert it's custom colour config to the new colour support Kevin added to Enums
2. change the colours themselves to those we've selected for toast notifications.  Use the blue for debug.

src/foam/u2/dialog/NotificationMessage.js
nanoscott: So we are going to do the LogLevel migration in a follow-up task, not in this PR. We can resolve this comment in this PR and address it later.

nanoscott (approved):

PR #3572, opened 2020-06-07
Add HTTP Authorization Type support - Bearer Token

jlhughes:
Add Authorization Types such a BEARER in HTTPBox.
Part of Medusa, but required earlier to test Medusa Nodes from pre-Medusa instances.
Also involves a refactoring of HTTPBox to allow for subclasses to override operations such as getReplyBox.

### DIFF
--- a/src/foam/box/HTTPAuthorizationType.js
+++ b/src/foam/box/HTTPAuthorizationType.js
@@ -1,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.ENUM({
+  package: 'foam.box',
+  name: 'HTTPAuthorizationType',
+
+  documentation: 'Types of HTTP Athorization',
+
+  values: [
+    {
+      name: 'NONE',
+      label: 'None',
+      ordinal: 0
+    },
+    // {
+    //   name: 'BASIC',
+    //   label: 'Basic',
+    //   ordinal: 1
+    // },
+    {
+      name: 'BEARER',
+      label: 'Bearer',
+      ordinal: 2
+    }
+  ]
+});
--- a/src/foam/box/HTTPBox.js
+++ b/src/foam/box/HTTPBox.js
@@ -45,6 +45,10 @@ foam.CLASS({
     'foam.box.Message',
   ],
 
+  javaImports: [
+    'javax.servlet.http.HttpServletRequest'
+  ],
+
   imports: [
     'creationContext',
     {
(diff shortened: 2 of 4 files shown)


src/foam/box/HTTPAuthorizationType.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2019 The FOAM Authors. All Rights Reserved.
```
nanoscott, line 3:
```suggestion
 * Copyright 2020 The FOAM Authors. All Rights Reserved.
```

src/foam/box/HTTPAuthorizationType.js
```diff
@@ -0,0 +1,30 @@
...
+
+  documentation: 'Types of HTTP Athorization',
+
+  values: [
+    {
+      name: 'NONE',
+      label: 'None',
+      ordinal: 0
+    },
+    // {
```
nanoscott, line 19: wondering why this is commented out?

src/foam/box/HTTPAuthorizationType.js
jlhughes: This really was created in 2019.

src/foam/box/HTTPAuthorizationType.js
jlhughes: Can't decide if None and Basic are the same.

src/foam/box/HTTPBox.js
```diff
@@ -86,6 +90,27 @@ foam.CLASS({
...
+      documentation: 'Calling http url, used when explicitly making cross site requests.',
+      name: 'origin',
+      class: 'String',
+      factory: function() {
+        return this.window && this.window.location.origin;
+      },
+      javaFactory: `
+      HttpServletRequest req = getX().get(HttpServletRequest.class);
+      if ( req != null ) {
+        return req.getScheme()+"://"+req.getServerName();
```
lchanmann, line 103: do we need to include port number? as `window.location.orgin` also have port number in it.

src/foam/box/HTTPBox.js
```diff
@@ -225,70 +270,108 @@ let task = URLSession.shared.dataTask(with: request) { data, response, error in
...
-  conn = (java.net.HttpURLConnection)url.openConnection();
-  conn.setDoOutput(true);
-  conn.setRequestMethod("POST");
-  conn.setRequestProperty("Accept", "application/json");
-  conn.setRequestProperty("Content-Type", "application/json");
+      // TODO: Go async and make request in a separate thread.
 
-  java.io.OutputStreamWriter output = new java.io.OutputStreamWriter(conn.getOutputStream(),
-                                                                     java.nio.charset.StandardCharsets.UTF_8);
+      java.net.HttpURLConnection conn;
```
lchanmann, line 275: Java 11 comes with new http client api, not sure if we could use that instead.

src/foam/box/HTTPBox.js
jlhughes: This is the default, you can override by explicitly setting the 'url'.  In medusa delegate this to a URLBuilder.

src/foam/box/HTTPBox.js
jlhughes: no, not changing this now.  We are going to replace anything with significant throughput with TCP client/server anyway.

lchanmann: ok, thanks for the elaboration.

PR #3575, opened 2020-06-08
Np 360/json formatter

smirnovaka:


src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -270,11 +283,11 @@ try {
...
       javaCode: `
-        write_(formatter.get().builder()
-          .append(prefix)
-          .append("r(")
-          .append(record)
-          .append(")"));
+      write_(sb.get()
+        .append(prefix)
+        .append("r(")
+        .append(record)
```
kgrgreer, line 289: Is it faster if we break this line up into three write_()'s? Write everything before 'record', write 'record', then write the ')'?
This would avoid copying 'record' into sb.

PR #3578, opened 2020-06-09
DUG Updates

nanoscott:
Only showing services the user has access to in DIG, DUG and SUGAR
Fixing properties that can be viewed in DUG
Adding titles to DIG, DUG and SUGAR

### DIFF
--- a/src/foam/nanos/dig/DIG.js
+++ b/src/foam/nanos/dig/DIG.js
@@ -37,7 +37,7 @@ foam.CLASS({
   ],
 
   imports: [
-    'nSpecDAO'
+    'AuthenticatedNSpecDAO'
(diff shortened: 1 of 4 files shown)


src/foam/nanos/dig/DIG.js
```diff
@@ -37,7 +37,7 @@ foam.CLASS({
   ],
 
   imports: [
-    'nSpecDAO'
+    'AuthenticatedNSpecDAO'
```
kgrgreer, line 40: Isn't the nSpecDAO already authenticated?

nanoscott: nope ..it's not

PR #3579, opened 2020-06-09
[DO_NOT_MERGE] [concept] theme controlled action spacing

KernelDeimos:
Interesting idea to eliminate margin-left on buttons and make button spacing theme-controlled. Unfortunately right now it doesn't work because the actions don't run correctly.

### DIFF
--- a/src/foam/u2/view/ActionArrayView.js
+++ b/src/foam/u2/view/ActionArrayView.js
@@ -1,0 +1,100 @@
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'ActionArrayView',
+  extends: 'foam.u2.View',
+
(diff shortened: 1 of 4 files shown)


src/foam/u2/view/ActionArrayView.js
```diff
@@ -0,0 +1,100 @@
...
+      name: 'actions',
+      title: 'Action Buttons'
+    }
+  ],
+
+  properties: [
+    {
+      class: 'String',
+      name: 'actionHorizontalPadding',
+      value: '8px'
```
jlhughes, line 76: why not model 'padding' and account for all. then use it wherever you need it like here.  the type/classes are String so can be null for not set.

PR #3580, opened 2020-06-09
Fix approval requests relationship

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-1430

## Changes
- Move user and approval request relationship from nanopay
- Add support for setting up approval request relationship DAO

### DIFF
--- a/src/foam/nanos/approval/ApprovableAwareDAO.js
+++ b/src/foam/nanos/approval/ApprovableAwareDAO.js
@@ -69,7 +69,8 @@ foam.CLASS({
       class: 'Object',
       javaType: 'foam.dao.DAO',
       name: 'approvalRequestDAO',
-      transient: true
+      transient: true,
+      javaFactory: 'return (DAO) getX().get("approvalRequestDAO");'
     }
   ],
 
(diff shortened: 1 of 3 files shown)


src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -165,7 +206,7 @@ foam.CLASS({
       ApprovableAware approvableAwareObj = (ApprovableAware) obj;
       LifecycleAware lifecycleObj = (LifecycleAware) obj;
 
-      DAO approvalRequestDAO = (DAO) x.get("approvalRequestDAO");
+      DAO approvalRequestDAO = obtainApprovalRequestDAO(x, obj);
```
jlhughes: This doesn't appear to be used.

jlhughes: getApprovalRequestDAO() should setup the dao.  The javaFactory of 'approvalRequestDAO' can have the logic of obtainApprovalRequestDAO - using getX(). 

The only other alternative is to make this DAO  implement foam.nanos.NanoService, implement 'start',  and then aquire the approvalRequestDAO.

lchanmann: We want the relationship DAO here and it needs the object `obj` which can be a different object on each put.

`obtainApprovalRequestDAO` is the first call to resolve the approval request relationship DAO if it exists, subsequently we can either use obtainApprovalRequestDAO or getApprovalRequest.

lchanmann: Updated to resolveApprovalRequestDAO which will resolve the relationship DAO and set it as the approvalRequestDAO and added javaFactory on approvalRequestDAO property - https://github.com/foam-framework/foam2/pull/3580/commits/2a4a1e9db3dd9f8c1758a06079ae7f1f3f10bf61.

src/foam/nanos/approval/ApprovableAwareDAO.js
```diff
@@ -165,7 +203,7 @@ foam.CLASS({
       ApprovableAware approvableAwareObj = (ApprovableAware) obj;
       LifecycleAware lifecycleObj = (LifecycleAware) obj;
 
-      DAO approvalRequestDAO = (DAO) x.get("approvalRequestDAO");
+      DAO approvalRequestDAO = resolveApprovalRequestDAO(x, obj);
```
jlhughes, line 219: Still don't understand this line. 
Why is this method not the logic of 'approvalRequestDAO' javafactory?

lchanmann: it needs `obj` to setup the relationship DAO and javaFactory doesn't have access to `obj` on each put.

jlhughes (approved):

PR #3582, opened 2020-06-10
[NP-1443] Refactor AbliiActionView + UI Button Fixes

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-1443
#TimeSpent: 3d


src/foam/u2/ActionView.js
```diff
@@ -76,43 +80,50 @@ foam.CLASS({
      */
 
     ^primary {
-      border-color: #355bc4;
+      border-color: #4a33f4;
```
jlhughes, line 83: Can we add border colour to Theme or whatever category these colours are and then specific them with the /* */. 
Ok if you just create a ticket to capture this work needed.

jlhughes (approved):

PR #3585, opened 2020-06-10
[NP-1475] Un-oops the LiveScriptBundler

KernelDeimos:
I missed an important instruction for the `pollEvents` method in Java's file watcher api
> Note that this method does not wait if there are no events pending.

This PR fixes LiveScriptBusyloop by adding a sleep interval.

### DIFF
--- a/src/foam/nanos/http/LiveScriptBundler.java
+++ b/src/foam/nanos/http/LiveScriptBundler.java
@@ -100,7 +100,7 @@ public class LiveScriptBundler
         }
         try {
           // Since each directory has a watcher thread, 1/100ms is sufficiently frequent
-          Thread.sleep(100);
+          Thread.sleep(500);
         } catch (InterruptedException e) {
           return;
         }
--- a/src/foam/nanos/http/LiveScriptBundler.java
+++ b/src/foam/nanos/http/LiveScriptBundler.java
@@ -98,6 +98,12 @@ public class LiveScriptBundler
 
           if ( ! key.reset() ) break;
         }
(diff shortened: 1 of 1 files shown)


src/foam/nanos/http/LiveScriptBundler.java
```diff
@@ -98,6 +98,12 @@ public void run() {
 
           if ( ! key.reset() ) break;
         }
+        try {
+          // Since each directory has a watcher thread, 1/100ms is sufficiently frequent
+          Thread.sleep(100);
```
kgrgreer: Would it hurt to bump this up to 500 or 1000?

KernelDeimos: I think it would be noticed, it's effectively the amount of time (maximum) between saving a file and the result showing up if you reload in the browser. However just this 100ms wait should drastically improve performance.

kgrgreer: I think we could go with 500ms. By the time you save in your browser then switch windows, I don't think 400ms would make much difference.

jlhughes: I would suspect developers watch the log files and wait for the Bundler to finish before reloading.  So a higher time like 500 or 1000s should be acceptible.

KernelDeimos: Done. Btw I added a lock to the web agent earlier, so now it's sufficient to reload the moment the bundler starts

PR #3590, opened 2020-06-10
Jetty Threadpool

KolombetM:
Just to get an idea


src/foam/nanos/Threadpool.js
```diff
@@ -0,0 +1,22 @@
+foam.CLASS({
```
jlhughes: license header.

jlhughes (commented):
Yep, just like this.

src/foam/nanos/Threadpool.js
```diff
@@ -0,0 +1,38 @@
+/**
```
jlhughes: /**
 * @license
 */

jlhughes (commented):
1. Rename Threadpool to ThreadpoolConfig
2. Move Threadpool to jetty package (package in source is correct).
3. Move service nspec to foam/nanos/jetty/services.jrl - and name the nspec (context key) 'jettyThreadpoolConfig' 
4. In HttpServer get ThreadpoolConfig from context. 
Thanks.

src/foam/nanos/jetty/ThreadpoolConfig.js
```diff
@@ -0,0 +1,27 @@
+/**
+* @license
+* Copyright 2020 The FOAM Authors. All Rights Reserved.
+* http://www.apache.org/licenses/LICENSE-2.0
+*/
+foam.CLASS({
+  package: 'foam.nanos.jetty',
+  name: 'ThreadpoolConfig',
```
kgrgreer: Could we rename this JettyThreadPoolConfig so it isn't confused with the NANOS threadpool?

jlhughes: Yes, that is in my notes above.

jlhughes: the class itself is already in the jetty package. the renaming was at the service (nspec) level.

jlhughes (approved):

PR #3593, opened 2020-06-11
makeing subject.user a slot

yij793:
address:https://nanopay.atlassian.net/browse/NP-1376
NP-1346 time 2h
using subject instead of user and agent

### DIFF
--- a/src/foam/nanos/u2/navigation/UserInfoNavigationView.js
+++ b/src/foam/nanos/u2/navigation/UserInfoNavigationView.js
@@ -16,9 +16,8 @@ foam.CLASS({
   ],
 
   imports: [
-    'agent',
-    'user',
-    'theme'
+    'theme',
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

src/foam/nanos/u2/navigation/UserInfoNavigationView.js
```diff
@@ -16,9 +16,8 @@ foam.CLASS({
   ],
 
   imports: [
-    'agent',
-    'user',
-    'theme'
+    'theme',
```
jlhughes, line 19: Is this used?

yij793: yes, line67
this.Menu.create({ id: this.theme.settingsRootMenu })

PR #3594, opened 2020-06-11
Use JSONFObjectFormatter in HTTPReplyBox

smirnovaka:


src/foam/box/HTTPReplyBox.js
```diff
@@ -37,7 +37,11 @@ try {
   javax.servlet.http.HttpServletResponse response = (javax.servlet.http.HttpServletResponse)getX().get("httpResponse");
   response.setContentType("application/json");
   java.io.PrintWriter writer = response.getWriter();
-  writer.print(new foam.lib.json.Outputter(getX()).setPropertyPredicate(new foam.lib.NetworkPropertyPredicate()).stringify(msg));
+  foam.lib.formatter.JSONFObjectFormatter formatter = new foam.lib.formatter.JSONFObjectFormatter(getX());
```
kgrgreer: Can we make this ThreadLocal to avoid creating and then growing a new one every time?

src/foam/box/HTTPBox.js
```diff
@@ -174,6 +175,11 @@ protected class Outputter extends foam.lib.json.Outputter {
     super.outputFObject(o);
   }
 }
+  protected foam.lib.formatter.JSONFObjectFormatter getFormatter(X x) {
```
jlhughes: re: ThreadLocal - yes, will add.  I'll add support in the lib.formatter package as we need this everywhere we use one.

src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -53,6 +54,18 @@ protected SimpleDateFormat initialValue() {
   protected boolean outputReadableDates_ = true;
   protected boolean outputDefaultClassNames_ = true;
 
+  public static ThreadLocal<FObjectFormatter> getThreadLocal(X x, Boolean quoteKeys, Boolean outputClassNames, PropertyPredicate predicate) {
```
jlhughes: Plan to add more of these as needed. This is for HTTPBox.

kgrgreer: If this is for the HTTPBox, then why not define it in the HTTPBox class? Then everyone can have their ThradLocal JSONFObjectFormatter their way?

jlhughes: I've used this 6 times in medusa and all have this same requirement.  I would say this is the default.

PR #3595, opened 2020-06-11
Use only one thread in LiveScriptBundler

KernelDeimos:
This didn't improve performance very much, but I gathered some useful information (and LiveScriptBundler probably shouldn't be creating a thread for each directory anyway).

If the answer on this StackOverflow thread is correct, the Java's FileWatcher api uses polling on Mac OS: https://stackoverflow.com/questions/9588737/is-java-7-watchservice-slow-for-anyone-else

This would explain the overhead, and why my previous commit adding `Thread.sleep` didn't improve performance as much as I hoped.

The good news is that there are alternatives available. I'm trying this next: https://github.com/gjoseph/barbarywatchservice - looks like it's almost a drop-in replacement for the WatcherService currently being used

### DIFF
--- a/src/foam/nanos/http/LiveScriptBundler.java
+++ b/src/foam/nanos/http/LiveScriptBundler.java
@@ -26,8 +26,7 @@ public class LiveScriptBundler
 
   // Filesystem
   protected String path_;
-  protected WatchService watcher_;
-  protected LinkedBlockingQueue<String> fileNames_;
+  protected Set<String> fileNames_;
 
   // Caching
   protected String javascriptBuffer_;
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

src/foam/nanos/http/LiveScriptBundler.java
```diff
@@ -64,42 +63,59 @@ public WatcherThread(
...
+          log_("IGNORE", foamPath);
+        }
+
+        if ( ! key.reset() ) break;
+      }
+    }
+  }
 
-          if ( ! key.reset() ) break;
+  private class WatcherThread implements Runnable {
```
jlhughes, line 98: If LiveScriptBundler implements NanoService, then it will call 'start' and from here start your thread or timer.  Then startup is consistent.

jlhughes (approved):

KernelDeimos:
Closing this - next pr, based on this one, is better

PR #3597, opened 2020-06-12
add permission to ucjDao checkOwnership

Nauna:
https://nanopay.atlassian.net/browse/NP-1478

added ucj.addPermission permission - so groups that aren't super privileged can still approve an ApprovalRequest that updates a UCJ

### DIFF
--- a/src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
+++ b/src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
@@ -25,6 +25,10 @@ foam.CLASS({
     'static foam.mlang.MLang.*'
   ],
 
+  messages: [
+    { name: 'ERROR_MSG', message: 'Error on UserCapabilityJunction checkOwnership create UCJ denied to user ' }
+  ],
+
   methods: [
     {
       name: 'getUser',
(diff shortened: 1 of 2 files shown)


src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -25,6 +25,10 @@ foam.CLASS({
     'static foam.mlang.MLang.*'
   ],
 
+  messages: [
+    { name: 'ERROR_MSG', message: 'Error on UserCapabilityJunction checkOwnership create UCJ denied to user ' }
```
jlhughes, line 29: Does this message make it back to the user? 
Can Crunch throw a subclass of Auth exception and always intercept?

Nauna: > Does this message make it back to the user?

No, only available in console. But took me awhile to find this cause there wasn't an error msg.

> Can Crunch throw a subclass of Auth exception and always intercept?

I don't know if I see the value of doing that - with what we currently have but not sure. Below I've described the intercept and the expense of our current implementation - may require some re-thought.

_So a CapabilityException (CapabilityRuntimeException) will be thrown in CapabilityAuthService which is a decorator on auth service. This check for capabilities happens with every permission check. And if a capability is found to intercept this permission then a CapabilityException is created._

_I then went and considered the efficiency of checking capabilities on each permission - and tbh it seems a little expensive._

_So for each permission check it does this (done with auth.check() or auth.checkUser()):_
1-  check if permission is a cached result - if yes return cache
2-  check if there is a capability with the permission listed in a Capability.PERMISSION_INTERCEPTED - if yes then return true. 
3- check is the permission is a Capability.NAME - if yes check if UCJ granted. 
4- check if a permission is implied by some ucj granted to user||realUser - if yes return true.
5- and finally - this is incorrect and should be removed - if all above fails, check if anything intercepts again

jlhughes: Thanks.  Yes, let's discuss. but in general with caching, the expensive lookup should only be once.

jlhughes (approved):

PR #3598, opened 2020-06-12
[DO_NOT_MERGE] [NP-94] adding drivers configuration

olhabn:
@kgrgreer @jlhughes 

https://nanopay.atlassian.net/browse/NP-94


olhabn:
still working on it

PR #3601, opened 2020-06-14
Implement StorageOptional which will supress output if the only prope…

jlhughes:
…rties to be output are optional - such as LastModified.


src/foam/lib/formatter/AbstractFObjectFormatter.java
```diff
@@ -116,6 +118,32 @@ protected synchronized List getProperties(ClassInfo info) {
     return propertyMap_.get(of);
   }
 
+  protected synchronized List getDelta(FObject oldFObject, FObject newFObject) {
```
kgrgreer: Why synchronized?

src/foam/lib/formatter/AbstractFObjectFormatter.java
```diff
@@ -116,6 +118,32 @@ protected synchronized List getProperties(ClassInfo info) {
...
+    ClassInfo info           = oldFObject.getClassInfo();
+    String    of             = info.getObjClass().getSimpleName();
```
kgrgreer: Why not call toLowerCase() here to avoid calling for each property?

src/foam/lib/formatter/AbstractFObjectFormatter.java
jlhughes: Ah, it initially didn't use getProperties itself. So this can now be removed.

src/foam/lib/formatter/AbstractFObjectFormatter.java
jlhughes: cut and paste, I'll fix where it came from as well.

src/foam/lib/formatter/AbstractFObjectFormatter.java
jlhughes: fixed.

src/foam/lib/formatter/AbstractFObjectFormatter.java
jlhughes: fixed.  Also in getProperties.

src/foam/dao/StorageOptionalDAO.js
```diff
@@ -0,0 +1,64 @@
...
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.dao',
+  name: 'StorageOptionalDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  documentation: `Discard DAO updates which result in only storageOptional properties changes.`,
```
jlhughes:
```suggestion
  documentation: `Discard DAO updates which result in only storageOptional property changes.`,
```

kgrgreer:
Has conflict.

PR #3606, opened 2020-06-16
[DO_NOT_MERGE] add thread.dumpstack

yij793:
using Thread.dumpStack to trace notification

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -646,6 +646,8 @@ foam.CLASS({
     {
       name: 'doNotify',
       javaCode: `
+        Thread.dumpStack();
+        System.out.println("doNotify method is triggered");
         // Get the default settings for the user if none are already defined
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/User.js
```diff
@@ -646,6 +646,8 @@ foam.CLASS({
     {
       name: 'doNotify',
       javaCode: `
+        Thread.dumpStack();
+        System.out.println("doNotify method is triggered");
```
nanoscott, line 650: why not write to the log?

jlhughes (approved):

src/foam/nanos/auth/User.js
yij793: you are right, I should

jlhughes: no. This is during replay and logging may not yet be available.

jlhughes: we are not merging this, it's just so QA has a branch to build against

PR #3610, opened 2020-06-16
[NP-1517] Update authCapability efficiency

KolombetM:
https://nanopay.atlassian.net/browse/NP-1517

### DIFF
--- a/src/foam/nanos/auth/CapabilityAuthService.js
+++ b/src/foam/nanos/auth/CapabilityAuthService.js
@@ -241,21 +241,19 @@ foam.CLASS({
 
         if ( capabilities.size() < 1 ) return;
 
-        List<Capability> filteredCapabilities = new java.util.ArrayList<Capability>(capabilities);
-
         for ( Capability c : capabilities ) {
           if ( ! c.getInterceptIf().f(x) ) {
-            filteredCapabilities.remove(c);
+            capabilities.remove(c);
           }
         }
 
         // Do not throw runtime exception if there are no intercepts
-        if ( filteredCapabilities.size() < 1 ) return;
+        if ( capabilities.size() < 1 ) return;
 
         // Add filteredCapabilities to a runtime exception and throw it
         CapabilityRuntimeException ex = new CapabilityRuntimeException(
           "Permission [" + permission + "] denied. Filtered Capabilities available.");
-        for ( Capability cap : filteredCapabilities ) ex.addCapabilityId(cap.getId());
+        for ( Capability cap : capabilities ) ex.addCapabilityId(cap.getId());
         throw ex;
       `
     }
--- a/src/foam/nanos/auth/CapabilityAuthService.js
+++ b/src/foam/nanos/auth/CapabilityAuthService.js
@@ -72,7 +72,7 @@ foam.CLASS({
           return;
 
         DAO userCapabilityJunctionDAO = (DAO) getX().get("userCapabilityJunctionDAO");
-        DAO capabilityDAO = (getX().get("capabilityDAO") == null ) ? (DAO) getX().get("capabilityDAO") : (DAO) getX().get("localCapabilityDAO");
+        DAO capabilityDAO = (DAO) getX().get("localCapabilityDAO");
         if ( capabilityDAO == null || userCapabilityJunctionDAO == null )
           return;
 
@@ -144,8 +144,8 @@ foam.CLASS({
         if ( x.get(Session.class) == null ) return false;
         if ( user == null || ! user.getEnabled() ) return false;
         User agent = ((Subject) x.get("subject")).getRealUser();
-        String agentKey = agent.getId() == user.getId() ? 
-          null : 
+        String agentKey = agent.getId() == user.getId() ?
+          null :
           user.getId() + ":" + agent.getId() + permission;
         String userKey = user.getId() + permission;
         this.initialize(x);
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -163,81 +163,36 @@ foam.CLASS({
...
-              if ( ucj.getStatus() == CapabilityJunctionStatus.GRANTED && 
+              if ( ucj.getStatus() == CapabilityJunctionStatus.GRANTED &&
                    c != null && ! c.isDeprecated(x) && c.implies(x, permission) ) {
                 return true;
               }
               return false;
             }
           };
+
+          // 1. check if the user has a capability that grants the permission
```
Nauna: I'd suggest either removing the numbers or consider making the top checks to the cache and the maybeIncercept #1. from lines 154-158

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -163,81 +163,36 @@ foam.CLASS({
...
+          Predicate userPredicate = EQ(UserCapabilityJunction.SOURCE_ID, user.getId());
           if ( userCapabilityJunctionDAO.find(AND(userPredicate, capabilityScope, predicate)) != null ) {
             result = true;
           }
           // Add the result to the cache
+          (( Map<String, Boolean> ) getCache()).put(userKey, result);
```
Nauna, line 189: this line is fine - but consider the result = true on line 186 (can't place comment on line that is not adjusted). We set result to true and then we check if true - this seems un-necessary. Consider removing the if check and just setting the cache.put and the return in the above if - line 186

src/foam/nanos/auth/CapabilityAuthService.js
Nauna: Also would suggest a comment update:
//# Check if a ucj implies the subject.user(business) has this permission

src/foam/nanos/auth/CapabilityAuthService.js
Nauna: commnet on line 188 is useless - in my mind

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -163,81 +163,36 @@ foam.CLASS({
...
           if ( result ) {
-            (( Map<String, Boolean> ) getCache()).put(userKey, result);
             return true;
           }
 
-          // 3. check if the agent has a capability that grants the permission
-          if ( agent != null ) {
+          // 2. check if the agent has a capability that grants the permission
```
Nauna: suggest a comment update:
//# Check if a ucj implies the subject.realUser(agent) has this permission

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -163,81 +163,36 @@ foam.CLASS({
...
                 return true;
               }
               return false;
             }
           };
+
+          // 1. check if the user has a capability that grants the permission
+          Predicate userPredicate = EQ(UserCapabilityJunction.SOURCE_ID, user.getId());
           if ( userCapabilityJunctionDAO.find(AND(userPredicate, capabilityScope, predicate)) != null ) {
             result = true;
```
Nauna, line 186: this line is fine - but consider the result = true. We set result to true and then we check if true - this seems unnecessary. Consider removing the 'if' check(line:190) and just setting the cache.put and the return in with this - line 186

also recommend same thing for code between line 202-209

noodlemoodle:
looks good to me but agree with Anna's suggestions.
Tested in liquid nothing broke, but not yet in ablii due to blocking bug after I pulled dev, will come back to this

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -248,17 +203,16 @@ foam.CLASS({
...
+            (( Map<String, Boolean> ) getCache()).put(agentKey, result);
             if ( result ) {
-              if ( agentKey != null ) (( Map<String, Boolean> ) getCache()).put(agentKey, result);
               return true;
             }
           }
-        } catch (Exception e) {
+        } catch ( Exception e ) {
           Logger logger = (Logger) x.get("logger");
           logger.error("check", permission, e);
```
Nauna: While here, can you consider changing 'check' to 'capabilityCheck'. I'm thinking e will give us stack trace, but 'check' seems useless. This function can get called from either auth.check or auth.checkUser

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -248,17 +203,16 @@ foam.CLASS({
...
         }
 
-        if ( result ) return true;
         maybeIntercept(x, permission);
```
Nauna, line 215: Do you think this is necessary? Since we do call this online 156?

src/foam/nanos/auth/CapabilityAuthService.js
```diff
@@ -248,17 +203,16 @@ foam.CLASS({
             }
 
             // Add the result to the cache
```
Nauna: comment 'add the result to cache' ... here and on line 188 seems useless

src/foam/nanos/auth/CapabilityAuthService.js
KolombetM: But then false results will not be cached

src/foam/nanos/auth/CapabilityAuthService.js
KolombetM: If we reached line 216 then we didn't call this method on line 156. Also, it means that there were no cashed false result before.

noodlemoodle:
There is one more thing that could be changed if you don't mind @Mykola97 - Please remove the line creating variable `filteredCapabilities` in the maybeIntercept method as well, and change references of the variable to `capabilities`, since this seems to be a duplicate to `capabilities`.

PR #3613, opened 2020-06-17
fix class names in formatter

smirnovaka:


src/foam/core/AbstractFObjectPropertyInfo.java
```diff
@@ -127,7 +128,13 @@ public boolean isDefaultValue(Object o) {
   }
   
   public void format(foam.lib.formatter.FObjectFormatter formatter, foam.core.FObject obj) {
-    formatter.output(get_(obj));
+    Object propObj = get_(obj);
+    if ( propObj instanceof FObject && ! (propObj instanceof UnknownFObject) ) {
```
kgrgreer: When is it not an FObject? When it is null?

src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -86,7 +86,7 @@ public JSONFObjectFormatter() {
     super();
   }
 
-  protected void outputUndefined() {
+  protected void outpvutUndefined() {
```
kgrgreer: typo "output" not "outpvut"

src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -74,7 +74,7 @@ protected SimpleDateFormat initialValue() {
   protected boolean outputShortNames_    = true;
   protected boolean outputDefaultValues_ = false;
   protected boolean multiLineOutput_     = false;
-  protected boolean outputClassNames_    = true;
+//  protected boolean outputClassNames_    = true;
```
kgrgreer: Why is this commented out?

src/foam/core/AbstractFObjectPropertyInfo.java
smirnovaka: in cases like this:
```{
     class: 'FObjectProperty',
     javaType: 'java.util.Timer',
     name: 'timer',
     javaFactory: 'return new Timer();',
   }

File: src/foam/lib/formatter/JSONFObjectFormatter.java (hunk 3 above)
Comment (smirnovaka) [reply src/foam/lib/formatter/JSONFObjectFormatter.java]: it's not used. Removed.

File: src/foam/core/AbstractFObjectPropertyInfo.java (hunk 1 above)
Comment (kgrgreer) [reply src/foam/core/AbstractFObjectPropertyInfo.java]: Then why is it an FObjectProperty? This seems wrong. If it isn't an FObject then it shouldn't be an FObjectProperty. java.util.Timer isn't going to be serializable anyway.

File: src/foam/lib/formatter/JSONFObjectFormatter.java [hunk 4]
```diff
@@ -74,7 +73,6 @@ protected SimpleDateFormat initialValue() {
   protected boolean outputShortNames_    = true;
   protected boolean outputDefaultValues_ = false;
   protected boolean multiLineOutput_     = false;
-  protected boolean outputClassNames_    = true;
```
kgrgreer, line 77: Why is this removed?

src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -303,7 +302,16 @@ protected Boolean maybeOutputProperty(FObject fo, PropertyInfo prop, boolean inc
...
-    ClassInfo info           = oldFObject.getClassInfo();
+    outputDelta(oldFObject, newFObject, null);
+  }
+
+  public void outputDelta(FObject oldFObject, FObject newFObject, ClassInfo defaultClass) {
+    ClassInfo info = newFObject.getClassInfo();
+
+    boolean outputClass = false;
+
+    if ( outputClassNames_ && ( info != defaultClass || outputDefaultClassNames_ ) ) outputClass = true;
```
kgrgreer: You could combine this line and the one above:
boolean outputClass = outputClassNames_ && ( info != defaultClass || outputDefaultClassNames_ );

src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -370,25 +370,18 @@ public void output(FObject[] arr) {
...
 
-    if ( info == defaultClass ) outputDefaultClassNames_ = false;
+    boolean outputClass = false;
 
-    output(o);
-  }
-
-  public void output(FObject o) {
-    ClassInfo info = o.getClassInfo();
+    if ( outputClassNames_ && ( info != defaultClass || outputDefaultClassNames_ ) ) outputClass = true;
```
kgrgreer: same as above comment

PR #3614, opened 2020-06-17
set short names to false in http box

zahnno:
Not sure how this was reverted but setting it back to false.


kgrgreer:
not needed, already fixed

PR #3620, opened 2020-06-17
isSortable

olhabn:
@jlhughes @Nauna @kgrgreer 
please review
https://nanopay.atlassian.net/browse/CPF-2517

### DIFF
--- a/src/foam/core/Property.js
+++ b/src/foam/core/Property.js
@@ -287,7 +287,12 @@ foam.CLASS({
     },
     {
       name: 'type',
-    }
+    },
+    {
+      class: 'Boolean',
+      name: 'isSortable',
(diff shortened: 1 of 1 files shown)


src/foam/core/Property.js
```diff
@@ -287,7 +287,12 @@ foam.CLASS({
     },
     {
       name: 'type',
-    }
+    },
+    {
+      class: 'Boolean',
+      name: 'isSortable',
```
jlhughes: I would say 'is' is not necessary, as 'ble' seems like a verb as well.

olhabn: fixed

jlhughes (approved):

PR #3621, opened 2020-06-17
NP-1475 improve and simplify live script bundler

KernelDeimos:
`io.methvin.watcher` supports native Mac OS file watching, and it also allowed me to remove WatcherThread and related logic since it has built-in support for recursive directory watching.

While I was here I decided to get rid of the files.js parsing code I wrote earlier and instead just check that a changed file ends with `.js` before running `build.js`, which is simpler and more reliable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,8 @@ foam2_MAVEN_DEPS = \
 	org.eclipse.jetty.websocket:websocket-server:9.4.8.v20171121 \
 	org.eclipse.jetty.websocket:websocket-servlet:9.4.8.v20171121 \
 	com.google.guava:guava:23.6-jre \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -75,7 +75,8 @@ foam2_MAVEN_DEPS = \
 	org.eclipse.jetty.websocket:websocket-server:9.4.8.v20171121 \
 	org.eclipse.jetty.websocket:websocket-servlet:9.4.8.v20171121 \
 	com.google.guava:guava:23.6-jre \
-	com.google.appengine:appengine-api-1.0-sdk:1.9.24
+	com.google.appengine:appengine-api-1.0-sdk:1.9.24 \
+	io.methvin:directory-watcher:0.9.10
```
jlhughes, line 79: please also update build/pom.xml.

PR #3623, opened 2020-06-17
[DO NOT MERGE] Np 1347/reverse google sheets integration

olhabn:
Draft pr


olhabn:
was trying to make smaller PRs so there is two branches one of them is another's one branch, but it probably won't work out that well. so closing this pr for now not to confuse anybody

PR #3624, opened 2020-06-18
MLang formulas

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/CPF-2699

## Changes
- Add MLang formulas for Add, Subtract, Multiply, Divide, MinFunc, and MaxFunc


src/foam/mlang/mlang.js
```diff
@@ -3775,6 +3776,137 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
```
jlhughes, line 3779: @kgrgreer @KernelDeimos Any requirement that mlang classes be in the same source/model file?

src/foam/mlang/mlang.js
```diff
@@ -3775,6 +3776,137 @@ foam.CLASS({
...
+  package: 'foam.mlang.expr',
+  name: 'Divide',
+  extends: 'foam.mlang.Formula',
+  implements: [ 'foam.core.Serializable' ],
+
+  methods: [
+    {
+      name: 'reduce',
+      abstract: false,
+      javaCode: 'return arg1 / arg2;'
```
jlhughes, line 3876: can we trap and report division by zero?

lchanmann: Addressed in https://github.com/foam-framework/foam2/pull/3628.

PR #3627, opened 2020-06-18
NumberShortener Usage for CurrencyInputView RO Mode

kennyqiyenkan:
Adds support for NumberShortener when displaying a value in RO DisplayMode for CurrencyInputView.

### DIFF
--- a/src/foam/u2/view/CurrencyInputView.js
+++ b/src/foam/u2/view/CurrencyInputView.js
@@ -19,6 +19,10 @@ foam.CLASS({
     'foam.mlang.Expressions'
   ],
 
+  requires: [
+    'foam.u2.util.NumberShortener'
+  ],
+
   imports: [
     'ctrl',
     'currencyDAO',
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/CurrencyInputView.js
```diff
@@ -86,6 +90,10 @@ foam.CLASS({
       name: 'currency',
       documentation: 'The actual currency object that can format the value'
     },
+    {
+      name: 'shortenToPrecision',
+      documentation: 'Expects an integer as value. Used for RO mode'
+    },
```
jlhughes, line 96: do we need a default?

kennyqiyenkan: Nope. Not having a value would just mean the displayed version of the value wont be shortened.

jlhughes (approved):

nanoscott (approved):

PR #3630, opened 2020-06-19
Add set property

KernelDeimos:
This is just a start - still needs Java support


kgrgreer:
Is this still needed?

KernelDeimos:
Nah I'll close this, it makes much more sense to add these methods to `Map` which already implements `$remove` and have `$add` set a key to `true`.

PR #3633, opened 2020-06-19
[NP-1552] ternary switch

KernelDeimos:
Made this three-state switch as an experiment in FOAM over the weekend. This version is refined based on suggestions from Joël and Kevin so it fits the theme better and allows all state transfers with only one click.

Additionally, since this is a new kind of button, I did some things to make it really obvious to the user what will happen. When you mouse over either side, that side darkens very slightly, as well as an indicator appears to show you where the switch will go if you click that side.

These are the state transfers:

Neutral state -> left for false, right for true
False state -> left for neutral, right for true
True state -> left for false, right for neutral

### DIFF
--- a/src/foam/u2/filter/properties/BooleanFilterView.js
+++ b/src/foam/u2/filter/properties/BooleanFilterView.js
@@ -19,7 +19,7 @@ foam.CLASS({
   ],
 
   requires: [
-    'foam.u2.CheckBox'
+    'foam.u2.tag.TernarySwitch'
   ],
 
   css: `
@@ -42,20 +42,15 @@ foam.CLASS({
       margin-bottom: 20px;
     }
 
-    ^container .foam-u2-md-CheckBox-label {
-      position: relative;
-      margin-top: 0;
-    }
-
-    ^container .foam-u2-md-CheckBox {
-      border-color: #9ba1a6;
-    }
-
-    ^container .foam-u2-md-CheckBox:checked {
-      background-color: #406dea;
-      border-color: #406dea;
+    ^container .foam-u2-tag-TernarySwitch {
+      flex-grow: 1;
+      text-align: center;
     }
   `,
+
+  messages: [
+    { name: 'help', message: 'Click the switch to modify the filter.' }
+  ],
 
   properties: [
     {
@@ -66,18 +61,9 @@ foam.CLASS({
       required: true
     },
     {
-      class: 'Boolean',
-      name: 'boolT',
-      label: 'True',
-      documentation: 'Filter property for True, mostly for UI/UX purposes',
-      value: false
-    },
-    {
-      class: 'Boolean',
-      name: 'boolF',
-      label: 'False',
-      documentation: 'Filter property for False, mostly for UI/UX purposes',
-      value: false
+      class: 'Int',
+      name: 'ternaryState',
+      value: 2 // 2 is neutral
     },
     {
       name: 'predicate',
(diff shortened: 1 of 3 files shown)


jlhughes:
@KernelDeimos can you add some screen shots, please. Thanks.

src/foam/u2/filter/properties/BooleanFilterView.js
```diff
@@ -104,33 +90,23 @@ foam.CLASS({
...
         .start().addClass(this.myClass('container'))
-          .start({
-            class: 'foam.u2.md.CheckBox',
-            data$: this.boolT$,
-            showLabel: true,
-            label: this.BOOL_T.label
-          }).end()
+          .tag(this.TernarySwitch, {
+            ternaryState$: this.ternaryState$,
+            width: '100%',
```
jlhughes: any way to pull this out into normal/regular CSS?

KernelDeimos: Just updated for width. Height would be non-trivial to move

src/foam/u2/tag/TernarySwitch.js
```diff
@@ -0,0 +1,237 @@
+foam.CLASS({
```
jlhughes, line 7: license header.

KernelDeimos: I always forget that... I gotta make a git plugin or something that checks my new files

KernelDeimos:
Screen Shot 2020-06-19 at 4 39 46 PM
Screen Shot 2020-06-19 at 4 40 20 PM
Screen Shot 2020-06-19 at 4 41 09 PM
Screen Shot 2020-06-19 at 4 41 31 PM

jlhughes (approved):

src/foam/u2/filter/properties/BooleanFilterView.js
```diff
@@ -104,33 +94,24 @@ foam.CLASS({
...
         .start().addClass(this.myClass('container'))
-          .start({
-            class: 'foam.u2.md.CheckBox',
-            data$: this.boolT$,
-            showLabel: true,
-            label: this.BOOL_T.label
-          }).end()
+          .start(this.TernarySwitch, {
+            ternaryState$: this.ternaryState$,
+            height: '30pt',
```
jlhughes, line 101: is it possible to move these to CSS?

src/foam/u2/filter/properties/BooleanFilterView.js
```diff
@@ -66,18 +65,9 @@ foam.CLASS({
...
-      documentation: 'Filter property for True, mostly for UI/UX purposes',
-      value: false
-    },
-    {
-      class: 'Boolean',
-      name: 'boolF',
-      label: 'False',
-      documentation: 'Filter property for False, mostly for UI/UX purposes',
-      value: false
+      class: 'Int',
```
kgrgreer, line 68: Does this mean I can no longer bind it directly to a boolean property? Couldn't you use three values as true, false and undefined, which FOAM supports?

jlhughes: Completely overlooked; I agree completely.

KernelDeimos: I wasn't able to get this working. The adapt function on Boolean converts `undefined` to `false`. If we want to use these value we would need a `Ternary` property that allows undefined.

PR #3639, opened 2020-06-22
[NP-1524] Add FormattedTextField

KernelDeimos:
This adds a generalized version of Kenny's CurrencyInputView. FormattedTextField will use the `tableCellFormatter` property of the property its editing/viewing to format the text as a user enters it.

### DIFF
--- a/src/foam/u2/FormattedTextField.js
+++ b/src/foam/u2/FormattedTextField.js
@@ -100,7 +100,7 @@ foam.CLASS({
     },
     // TODO: move this to property of foam.core.String
     function sanitizeString(s) {
-      return s.replace(/[\D\s\._\-\/]+/g, "");
+      return s.replace(/[\s\._\-\/]+/g, "");
     }
   ],
 })
--- a/src/foam/u2/FormattedTextField.js
+++ b/src/foam/u2/FormattedTextField.js
@@ -1,0 +1,107 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2',
+  name: 'FormattedTextField',
+  extends: 'foam.u2.View',
+
+  implements: [
+    'foam.mlang.Expressions'
+  ],
+
+  imports: [
+  ],
+
+  css: `
+    ^ {
+      display: flex;
+    }
+
+    ^container-selection {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      text-align: center;
+
+      box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


src/foam/u2/FormattedTextField.js
```diff
@@ -0,0 +1,107 @@
...
+        }));
+    },
+    function fromProperty(prop) {
+      this.SUPER(prop);
+
+      this.prop = prop;
+    },
+    // TODO: move this to property of foam.core.String
+    function sanitizeString(s) {
+      return s.replace(/[\D\s\._\-\/]+/g, "");
```
kennyqiyenkan: This would remove anything that isn't a number. Is this the behaviour you want for a generalized FormattedTextView?

KernelDeimos: tbh I assumed `\D` was for unicode dashes but I just looked it up and it's non-digits. I'll remove that token - thanks for the catch!

KernelDeimos: A property property called `sanitize` would be great for this because then you can override the answer to "what's a formatting character?" on any property, but I'll wait for @kgrgreer 's input on this

kennyqiyenkan: I like the idea of being able to provide a regex for it. It definitely makes it more dynamic.

However, I am wondering about the side effects in regards to UX. The CurrencyInputView has a side effect of moving the cursor back to the end if you inserted a value midway and it reformats. It's not a major thing for number input since I intended it to behave more like an atm or calculator, but I wonder about other things. Just food for thought.

KernelDeimos: I think I know an elegant way to fix all of that - adding `[DNM]` tag, I'll commit a fix soon

KernelDeimos: Got it working pretty well now

KernelDeimos: So playing around with it some more, it's probably good enough but has some undesirable quirks like deleting forwards (like PC del key or fn+delete on osx) while at a syntax character moves the cursor in the wrong spot. Also the code I wrote to make it work is a little complicated which is kind of a double-edged sword - like I feel like there's gotta be a much simpler way to do this but at the same time now we have a neat string diff utility if I factor it out.

jlhughes:
@KernelDeimos Is this replaced by String.format PR?

KernelDeimos:
It's a separate thing but would need some more work to be useful. Maybe if I add a comment saying it's incomplete then if anybody ever needs it it can be fixed later, either that or we can just close this

jlhughes:
@KernelDeimos I'd rather it be closed then merged in an incomplete state. 
@kgrgreer thoughts?

PR #3640, opened 2020-06-22
[NP-1559] - FObjectView ChoiceView Fix

nanopay-arthur:
Refs: 
https://nanopay.atlassian.net/browse/NP-1559
https://nanopay.atlassian.net/browse/NP-351

#TimeSpent: 1d

Affected Screens:
Screen Shot 2020-06-29 at 11 29 25 AM
Screen Shot 2020-06-29 at 11 23 49 AM

### DIFF
--- a/src/foam/nanos/boot/NSpec.js
+++ b/src/foam/nanos/boot/NSpec.js
@@ -118,9 +118,9 @@ foam.CLASS({
     {
       class: 'FObjectProperty',
       name: 'service',
-      view: 'foam.u2.DetailView',
-      // Why was this set to be a SectionedDetailView? Put back to DetailView. KGR
-      //      view: { class: 'foam.u2.detail.SectionedDetailView' },
+      view: {
+        class: 'foam.u2.view.FObjectView'
+      },
       readPermissionRequired: true,
       writePermissionRequired: true
     },
--- a/src/foam/nanos/boot/NSpec.js
+++ b/src/foam/nanos/boot/NSpec.js
@@ -118,9 +118,7 @@ foam.CLASS({
     {
       class: 'FObjectProperty',
       name: 'service',
-      view: {
-        class: 'foam.u2.view.FObjectView'
-      },
+      view: 'foam.u2.view.FObjectView',
       readPermissionRequired: true,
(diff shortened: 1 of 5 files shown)


jlhughes (approved):

src/foam/nanos/boot/NSpec.js
```diff
@@ -118,9 +118,7 @@ foam.CLASS({
     {
       class: 'FObjectProperty',
       name: 'service',
-      view: 'foam.u2.DetailView',
-      // Why was this set to be a SectionedDetailView? Put back to DetailView. KGR
-      //      view: { class: 'foam.u2.detail.SectionedDetailView' },
+      view: 'foam.u2.view.FObjectView',
```
kgrgreer, line 121: Don't we need to allowCustom: true for this?

src/foam/nanos/menu/Menu.js
```diff
@@ -53,7 +53,6 @@
       documentation: 'View initialized when menu is launched.',
       view: {
         class: 'foam.u2.view.FObjectView',
-        allowCustom: true,
```
kgrgreer: Don't we need this?

nanoscott (approved):

src/foam/nanos/boot/NSpec.js
nanopay-arthur: I don't believe we need to explicitly pass on a boolean for allowCustom since FObjectView automatically computes the boolean based on the length of this.choices via expression.

src/foam/nanos/menu/Menu.js
nanopay-arthur: Same as above.

lchanmann:
It's used to render the default selected view without needing to change dropdown to other option first.

Screen Shot 2020-06-22 at 2 44 23 PM

Expected:
Screen Shot 2020-06-22 at 2 48 00 PM

src/foam/nanos/menu/Menu.js
kgrgreer: But it only sets allowCustom: true if there are no choices, but there are choices in this case.

PR #3645, opened 2020-06-23
Np 1527/parsing error

smirnovaka:


src/foam/dao/AbstractF3FileJournal.js
```diff
@@ -355,14 +363,17 @@ try {
           String line = reader.readLine();
           if ( line == null ) return null;
           if ( ! line.equals("p({") && ! line.equals("r({") ) return line;
-          StringBuilder sb = new StringBuilder();
-          sb.append(line);
+          strb.setLength(0);
```
jlhughes: You have a thread local string builder - what is this one?

kgrgreer: Is used for reading, and is shared across threads, so threadlocal wouldn't work. Should add a comment and maybe give a more meaningful name.

smirnovaka: string builder is not thread local

src/foam/dao/F3FileJournal.js
```diff
@@ -39,53 +41,63 @@ foam.CLASS({
       ],
       javaCode: `
         // count number of entries successfully read
-        int successReading = 0;
-        JSONParser parser = getParser();
+        AtomicInteger successReading = new AtomicInteger();
```
kgrgreer, line 44: This can just be a regular 'int'. It is only updated in endJob() which is single-threaded so doesn't need synchronization.

PR #3647, opened 2020-06-23
Add FragmentedTextField

KernelDeimos:
A more user-friendly view that can be used instead of #3639, but less suitable for complex formatting situations.

This view works well when the formatting is statically delimited, such as what the applyFormat function from #3644 supports.

Here's an example:
Screen Shot 2020-06-23 at 3 43 59 PM
```
.tag(this.FragmentedTextField, {
  delegates: [
    {
      class: 'foam.u2.TextField',
      attributes: [ { name: 'maxlength', value: 2 } ]
    },
    '-',
    {
      class: 'foam.u2.TextField',
      attributes: [ { name: 'maxlength', value: 3 } ]
    },
    '-',
    {
      class: 'foam.u2.TextField',
      attributes: [ { name: 'maxlength', value: 3 } ]
    },
  ],
  data$: this.data$,
})
```

### DIFF
--- a/src/foam/u2/FragmentedTextField.js
+++ b/src/foam/u2/FragmentedTextField.js
@@ -20,7 +20,6 @@ foam.CLASS({
       margin-left: -1px;
       margin-right: -1px;
       line-height: 100%;
-      z-index: -1;
 
       /* TODO: get from theme when available */
       padding-left: 8px;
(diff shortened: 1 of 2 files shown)


src/foam/u2/FragmentedTextField.js
```diff
@@ -0,0 +1,74 @@
+foam.CLASS({
```
jlhughes, line 7: FOAM license header.

PR #3649, opened 2020-06-24
Cross Session Notifications / NotificationMessage Refactor

blakegreen32:
nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/10134
Ticket: https://nanopay.atlassian.net/browse/NP-1290


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -496,13 +499,15 @@ foam.CLASS({
...
-    function notify(data, type, description) {
-      /** Convenience method to create toast notifications. */
-      this.add(this.NotificationMessage.create({
-        message: data,
-        type: type,
-        description: description
-      }));
+    function notify(toastMessage, toastSubMessage, severity, transient) {
+      var notification = this.Notification.create();
+      notification.userId = this.user.id;
```
jlhughes: Who is user? does this need to be realUser?

blakegreen32: Needs to be the logged in user, since this notify method is for showing in app notification messages. When a Notification is created from an approval request it creates a notification and sets the id to the approver which would send them a notification message once I get websockets working for the notificationDAO.

src/foam/nanos/u2/navigation/SignIn.js
```diff
@@ -126,7 +131,10 @@ foam.CLASS({
                     this.user.copyFrom(updatedUser);
                     this.nextStep();
                   }).catch((err) => {
-                    this.notify(err.message || this.ERROR_MSG, 'error');
+                    this.ctrl.add(this.NotificationMessage.create({
```
nanoscott, line 134: why not use the notify method here?

src/foam/nanos/u2/navigation/SignUp.js
```diff
@@ -280,7 +282,10 @@ foam.CLASS({
             this.user.copyFrom(user);
             this.updateUser(x);
           }).catch((err) => {
-            this.notify(err.message || this.ERROR_MSG, 'error');
+            this.ctrl.add(this.NotificationMessage.create({
```
nanoscott, line 285: why not here as well?

src/foam/nanos/u2/navigation/SignIn.js
blakegreen32: so for sign in and sign up the user is not logged in yet. So I have to just display a NotificationMessage as we need the logged in user to send notification messages through the notificationDAO.

nanoscott: ok.

nanoscott:
You need to add some way that notifications are displayed by default for FOAM. So if an application is using the default ApplicationController, you need to show the notifications. So you'll need something like the sub method on the other PR. It should be able to be overriden by other controllers so that they can do their own display of notifications.

src/foam/nanos/controller/ApplicationController.js
jlhughes: yes, so where is it coming from? I kept expanding, but wasn't sure. Are you using subject.realUser or is it being set elsewhere?

nanoscott (approved):

src/foam/nanos/controller/ApplicationController.js
nanoscott: Does this work?

```suggestion
      notification.userId = this.subject.realUser.id;
```

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -351,6 +351,21 @@ foam.CLASS({
       window.addEventListener('resize', this.updateDisplayWidth);
       this.updateDisplayWidth();
 
+      this.__subSubContext__.notificationDAO.where(
+        this.EQ(this.Notification.USER_ID, this.subject.user.id)
```
nanoscott:
```suggestion
        this.EQ(this.Notification.USER_ID, this.subject.realUser.id)
```

src/foam/nanos/controller/ApplicationController.js
nanoscott: or 
```suggestion
      notification.userId = this.subject && this.subject.realUser ? this.subject.realUser.id : this.user.id;
```

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -348,6 +351,24 @@ foam.CLASS({
       window.addEventListener('resize', this.updateDisplayWidth);
       this.updateDisplayWidth();
 
+      var userNotificationQueryId = this.subject && this.subject.realUser ?
+      this.subject.realUser.id : this.user.id;
```
nanoscott, line 355:
```suggestion
        this.subject.realUser.id : this.user.id;
```

nanoscott (approved):

jlhughes (approved):

PR #3653, opened 2020-06-24
mlang formulas JS implementation

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-1564

## Changes
- Add DotF.f() and Formula.f() JS implementations and tests
- Cleanup unused foam.mlang.expr.Mul

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -2448,25 +2448,6 @@ foam.CLASS({
   ]
 });
 
-
-foam.CLASS({
-  package: 'foam.mlang.expr',
-  name: 'Mul',
-
-  extends: 'foam.mlang.predicate.Binary',
-
-  implements: [
-    'foam.core.Serializable'
-  ],
-
-  documentation: 'Multiplication Binary Expression.',
-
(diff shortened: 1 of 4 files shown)


jlhughes (approved):

src/foam/mlang/mlang.js
```diff
@@ -1695,7 +1695,7 @@ foam.CLASS({
       return typeof x === 'number' ? '' + x :
         typeof x === 'string' ? '"' + x + '"' :
         Array.isArray(x) ? '[' + x.map(this.toString_.bind(this)).join(', ') + ']' :
-        x.toString ? x.toString() :
+        (x && x.toString) ? x.toString() :
```
kgrgreer: Don't all JS objects have toString(), since it's in the Object class?

lchanmann: x here is the value of foam.mlang.Constant and it can be null then throws on x.toString.

    foam.mlang.Constant.create({value: null}).toString();  // throws Uncaught TypeError: Cannot read property 'toString' of null

PR #3657, opened 2020-06-26
Contextual DotF predicate

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-1379

## Changes
- ~Add contextualizing support to mlang expression~
- ~Add ContextAwarePredicate model~
- Contextualize DotF predicate
- Fix for strategizer choice view choice selection in update view (included here as the change is just 2 lines)


nanoscott (approved):

src/foam/mlang/mlang.js
```diff
@@ -3956,6 +3956,74 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
```
jlhughes: @kgrgreer going forward can we move new entries into their own file? Makes debugging in browser possible.

jlhughes (approved):

src/foam/core/ContextAware.js
```diff
@@ -17,5 +17,21 @@ foam.INTERFACE({
...
       args: [ { name: 'x', type: 'Context' } ]
     }
+  ],
+
+  axioms: [
+    {
+      name: 'javaExtras',
+      buildJavaClass: function(cls) {
+        cls.methods.push(`
+          static <T> T maybeContextualize(X x, T obj) {
```
kgrgreer, line 27: FObject extends ContextAware, so when would obj ever not be an instanceof ContextAware?

src/foam/mlang/MLang.java
```diff
@@ -37,9 +39,19 @@ public static Comparator DESC(Comparator c) {
   }
 
   public static Expr prepare(Object o) {
-    return o instanceof Expr ? (Expr) o :
+    var expr = o instanceof Expr ? (Expr) o :
```
kgrgreer: I thought you had messed up JS and Java for  a minute when you used 'var'.

src/foam/mlang/mlang.js
kgrgreer: Sure.

src/foam/core/ContextAware.js
lchanmann: obj can be of any type, doesn't have to be FObject.

src/foam/mlang/MLang.java
lchanmann: Java 'var' correctly guesses that it is a `Expr` if not the compiler would yell on line 45.

lchanmann:
Updated @kgrgreer @jlhughes please take another look. Thanks

PR #3658, opened 2020-06-29
Added stack pop after userFeedback exception to DAOCreateView and DAOUpdateView

tharmaman:


nick-prat:
Do we also want this for `NotificationRowView` `DeleteModal` and `AccountDAOCreateView`?

edit: just saw you have `AccountDAOCreateView` in a different PR.

tharmaman:
> Do we also want this for `NotificationRowView` `DeleteModal` and `AccountDAOCreateView`?

Added to AccountDAOCreateView in the PR on dev, don't think it makes sense for NotificationRowView since its actioned in the table and DeleteModal uses an x.closeDialog instead since its a modal so we're good

nick-prat (approved):

PR #3665, opened 2020-07-01
Move duration formatting logic from TableCellFormatter to Duration model

jlhughes:


src/foam/core/Duration.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * @license
+ * Copyright 2016 The FOAM Authors. All Rights Reserved.
```
jlhughes, line 3: @kgrgreer this is the year of the original types.js that Duration was pulled from.

nanoNeel (approved):

PR #3668, opened 2020-07-05
fixing cleanup

olhabn:
Ticket: https://nanopay.atlassian.net/browse/NP-1610

moving call to deleteGoogleSheet method to finally block on ExportDriver


src/foam/nanos/export/ExportDriver.js
```diff
@@ -24,6 +24,12 @@ foam.INTERFACE({
       args: ['X', 'dao'],
       documentation: 'Exports data in a DAO',
       flags: ['js']
+    },
+    {
+      name: 'cleanup',
```
jlhughes: Often this would be called 'tearDown' (in testing frameworks) as it is the opposite of 'setUp', should there be one. Unless you are really tied to 'cleanup'.

olhabn: good to know! thanks.
fixed

jlhughes (approved):

PR #3672, opened 2020-07-07
Association Capability

noodlemoodle:
There are 3 types of use cases for capability
1. capability that belong to subject.user
2. capability that belong to subject.realuser
3. capability that belong to the real user when acting as user

to cover the first 2 use cases : added associatedEntity as a prop capability model
to cover third use case : added AssociationCapability subclass, when the capability is instance of this, instead of UserCapabilityJunction, we create AgentCapabilityJunction (ext. ucj) which has an extra prop called effectiveUser

nanopay side - https://github.com/nanoPayinc/NANOPAY/pull/10168

**changes after discussion :**
1. make associated entity enum : https://github.com/foam-framework/foam2/pull/3672/commits/ca943db79776be03d7cdc3554298e32c85877cd1
2. remove associationcapability and denote this using ACTING_USER enum in associatedEntity property https://github.com/foam-framework/foam2/pull/3672/commits/9a08f7402c35395b86b7fdb97d96ba40997f904a


src/foam/nanos/crunch/UserCapabilityJunctionDAO.js
```diff
@@ -59,11 +43,13 @@ foam.CLASS({
...
-        User agent = ((Subject) x.get("subject")).getRealUser();
+        Subject subject = (Subject) x.get("subject");
+        User user = subject.getUser();
+        User realUser = subject.getRealUser();
+
         AuthService auth = (AuthService) x.get("auth");
-        boolean isOwner = obj.getSourceId() == user.getId() || ( agent != null && obj.getSourceId() == agent.getId() );
-        if ( ! isOwner && ! auth.check(x, "ucj.addPermission") ) throw new AuthorizationException(ERROR_MSG + agent.getId()+":"+user.getId());
+        boolean isOwner = obj.getSourceId() == user.getId() || obj.getSourceId() == realUser.getId();
+        if ( ! isOwner && ! auth.check(x, "*") ) throw new AuthorizationException();
```
Nauna, line 52: Don't think we should revert these changes - consider this https://github.com/foam-framework/foam2/pull/3597

think this is more correct: 'if ( ! isOwner && ! auth.check(x, "ucj.addPermission") ) throw new AuthorizationException(ERROR_MSG + realUser.getId()+":"+user.getId());'

noodlemoodle: huh must be a weird git thing, I was seeing "service.*" for the auth check locally so I changed to something that made more sense. will change, thanks

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -13,7 +13,8 @@ foam.CLASS({
   ],
 
   imports: [
-    'user',
+    'crunchController',
+    'subject',
```
Nauna: not used - remove?

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -22,6 +23,7 @@ foam.CLASS({
     'foam.core.Property',
     'foam.layout.Section',
     'foam.layout.SectionAxiom',
+    'foam.nanos.crunch.AgentCapabilityJunction',
```
Nauna: none of these required classes are used

src/foam/u2/crunch/CrunchController.js
```diff
@@ -243,6 +264,51 @@ foam.CLASS({
...
+            self.AND(
+              self.INSTANCE_OF(self.AgentCapabilityJunction),
+              self.EQ(self.UserCapabilityJunction.SOURCE_ID, associatedEntity.id),
+              self.EQ(self.AgentCapabilityJunction.EFFECTIVE_USER, self.subject.user.id)
+            )
+          ),
+          self.EQ(self.UserCapabilityJunction.TARGET_ID, self.capability.id)
+        )
+      ).then(ucj => {
+        self.ucj = ucj;
```
Nauna: hmmm - self.ucj ? 

CrunchController should not be saving any state thus should not be storing a ucj.

noodlemoodle: self refers to the wizardlet passed in line 269, working on a commit to use imports and requires from crunchcontroller instead of those from capabilitywizardlet and rename self back to wizardlet

src/foam/nanos/crunch/AssociationCapability.js
```diff
@@ -0,0 +1,25 @@
+/**
```
Nauna: Don't think this subclass is necessary. Actually complicates this a bit more than necessary.

Consider:
1) CODE:
```var isAssociationCapability = self.AssociationCapability.isInstance(self.data);```
```var associatedEntity = isAssociationCapability ? self.subject.realUser :```
```self.data.associatedEntity === 'user' ? self.subject.user : self.subject.realUser;```

above code used in multiple locations (CapabilityCardView, CapabilityIntercept and CrunchController)

If you got rid of this subclass then the code you need becomes:

```var associatedEntity = self.subject[self.data.associatedEntity];```

We can set the associatedEntity on the capability jrl entry.

2) Also this affects the CrunchController.launchWizard. Now it takes in a capability instead of an id. I think this now adds a layer of complication to the call - instead of just needing the id of the capability you want to call, you'll have to query the ucj for the capability before calling the launchWizard ... seems messy if this is going to be called through api(DIG).

Comment (noodlemoodle) [reply src/foam/nanos/crunch/AssociationCapability.js]: using self.subject[self.data.associatedEntity] covers the first 2 cases in my description above but not the third one for "capability that belong to the real user when acting as user"
as for crunchcontroller.launchwizard, I think even without this, the capability needs to be passed in to get its associatedEntity.

File: src/foam/u2/crunch/CrunchController.js (hunk 4 above)
Comment (noodlemoodle) [reply src/foam/u2/crunch/CrunchController.js]: resolved in https://github.com/foam-framework/foam2/pull/3672/commits/df543b6c18e30c6cf5e56c0a5b918af15ca5ef13

File: src/foam/u2/crunch/CrunchController.js [hunk 6]
```diff
@@ -243,6 +264,50 @@ foam.CLASS({
...
+    },
+    function save(wizardlet) {
+      var isAssociationCapability = foam.nanos.crunch.AssociationCapability.isInstance(wizardlet.capability);
+      var associatedEntity = isAssociationCapability ? this.subject.realUser : 
+      wizardlet.capability.associatedEntity === 'user' ? this.subject.user : this.subject.realUser;
+
+      return this.updateUCJ(wizardlet, associatedEntity).then(() => {
+        var ucj = wizardlet.ucj;
+        if ( ucj === null ) {
+          ucj = isAssociationCapability ?
```
Nauna: isAssociationCapability means we are saving cap to ucj on realUser ... does that really mean this will be an agentCapabilityJunction?

noodlemoodle: associatedEntity determines whether we should save to user/realuser. is associationcapability is a special case of capability where we want to grant to "realuser when acting as user", which uses agentcapabilityjunction instead of usercapabilityjunction

src/foam/nanos/crunch/AssociationCapability.js
Nauna: hmmm I see it now - so AssociationCapability is marker for the AgentCapabilityJunction.
Thx for clarifying.

Nauna: as for the wizard - how about a find in crunchController.launchWizard(id) that first searches for the capability id - then don't need to change the calls or do a find on each call to the launchWizard

Nauna (approved):

src/foam/nanos/crunch/AssociationCapability.js
```diff
@@ -0,0 +1,25 @@
...
+  name: 'AssociationCapability',
+  extends: 'foam.nanos.crunch.Capability',
+
+  documentation: `
+    A capability that a user has in relation to another entity.
+  `,
+
+  properties: [
+    {
+      name: 'associatedEntity',
```
KernelDeimos: Maybe this should be an enum

Nauna (approved):

jlhughes (approved):

PR #3673, opened 2020-07-07
OrCapability

KernelDeimos:
This PR will add OrCapability. OrCapability will work once status update rules are updated.

Current plan
- [x] Add capability-defined method to check if prerequisites are satisfactory (`hasPrerequisitesSatisfied`)
- [x] Add tests for Capability `hasPrerequisitesSatisfied`
- [x] Add OrCapability

### DIFF
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -17,7 +17,9 @@ foam.CLASS({
     'foam.dao.ArraySink',
     'foam.dao.DAO',
     'foam.dao.Sink',
+    'foam.mlang.predicate.Predicate',
     'foam.mlang.sink.Count',
+    'java.util.ArrayList',
     'java.util.Date',
     'java.util.List',
     'static foam.mlang.MLang.*'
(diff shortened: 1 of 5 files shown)


src/foam/nanos/crunch/Capability.js
```diff
@@ -282,6 +284,65 @@ foam.CLASS({
...
+        { name: 'x', type: 'Context' },
+        { name: 'ucjDaoKey', type: 'String' },
+        { name: 'ucjUserPredicate', type: 'Predicate' },
+      ],
+      documentation: `
+        Check if this capability is satisfied with its prerequisites.
+        Default implementation requires all prerequisites.
+      `,
+      javaCode: `
+        // Get a list of prerequisite capabilities
```
Nauna, line 301: Should we not be adjusting the logic for the UserCapabilityJunctionDAO decorator with this PR?

Then I have to wonder when you are iterating over the UCJ's here - we then are duplicating the iteration in the above named decorator.

I'd like to see how this all plays out with the changes to ucjDAO

KernelDeimos: > Should we not be adjusting the logic for the UserCapabilityJunctionDAO decorator

This would be replacing the similar code in UserCapabilityJunctionDAO (well, UserCapabilityJunctionDAO would be calling on this code). The reason doing it here instead of there would be that now it's the subclass of Capability's job to determine how its prerequisites are satisfied.

The reason I didn't remove the code in UserCapabilityJunctionDAO in this PR was I was hoping it could get merged as-is and then we have some opportunity to play with the idea and see if we run into hurdles.

KernelDeimos:
Deprecated by #3806

PR #3676, opened 2020-07-08
[DO_NOT_MERGE] Templated group on accessControlModal

minsunny:
temp open til the current ablii bugs are fixed.

nanopaySide :  https://github.com/nanoPayinc/NANOPAY/pull/10176

https://nanopay.atlassian.net/browse/NP-1545
using templated groups not hard coding on AccessControlModal

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -352,16 +352,6 @@ foam.CLASS({
 
         return ancestor;
       `
-    },
-    {
-      name: 'toSummary',
-      documentation: `When using a reference to the roleDAO, the labels associated
-        to it will show a chosen property rather than the first alphabetical string
-        property. In this case, we are using the name.
-      `,
-      code: function(x) {
-        return this.name || this.id;
-      }
     }
   ]
 });
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -133,8 +133,8 @@ foam.CLASS({
     */
     {
       class: 'Boolean',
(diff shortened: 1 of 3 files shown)


src/foam/nanos/auth/Group.js
```diff
@@ -131,6 +131,19 @@ foam.CLASS({
...
     }
     */
+    {
+      class: 'Boolean',
+      name: 'flag',
+      documentation: 'flag to filter sme specific groups'
+    },
+    {
+      class: 'String',
+      name: 'label',
```
zahnno, line 141: We should be using toSummary to display object data.

src/foam/nanos/auth/Group.js
```diff
@@ -131,6 +131,19 @@ foam.CLASS({
       documentation: 'Custom authentication settings for this group.'
     }
     */
+    {
+      class: 'Boolean',
+      name: 'flag',
+      documentation: 'flag to filter sme specific groups'
```
zahnno: Let's rename this property to **template** and change the documentation to _Used to specify that the group is used as a reference for the creation of other groups._ This provides a bit more concise documentation and naming in regards to the use of this property.

src/foam/nanos/auth/Group.js
minsunny: Just used a label property not toSummary method to set a label value when it is needed. For example, currently 'smeBusinessAdmin', 'smeBusinessApprover', 'smeBusinessEmployee' are used as id but we would like to display as Admin, Approver, Employee. For non sme specific groups, just use its id as default.

PR #3678, opened 2020-07-08
Notification transient check on authorize methods

blakegreen32:
Added transient check to authorizeOnCreate and authorizeOnUpdate methods of Notification model

### DIFF
--- a/src/foam/nanos/notification/Notification.js
+++ b/src/foam/nanos/notification/Notification.js
@@ -197,14 +197,14 @@ foam.CLASS({
       name: 'authorizeOnCreate',
       javaCode: `
       AuthService auth = (AuthService) x.get("auth");
-      if ( ! checkOwnership(x) && ! auth.check(x, createPermission("create")) ) throw new AuthorizationException("You don't have permission to create this notification.");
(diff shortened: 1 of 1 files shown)


src/foam/nanos/notification/Notification.js
```diff
@@ -197,14 +197,14 @@ foam.CLASS({
       name: 'authorizeOnCreate',
       javaCode: `
       AuthService auth = (AuthService) x.get("auth");
-      if ( ! checkOwnership(x) && ! auth.check(x, createPermission("create")) ) throw new AuthorizationException("You don't have permission to create this notification.");
+      if ( ! checkOwnership(x) && ! auth.check(x, createPermission("create")) && ! getTransient() ) throw new AuthorizationException("You don't have permission to create this notification.");
```
zahnno, line 200: Objects can be transient?

jlhughes: @zahnno This is a 'model' property on Notification. Not related to property property 'transient'.

jlhughes (approved):

src/foam/nanos/notification/Notification.js
zahnno: Ahh I see it now. What's the purpose of it?

PR #3679, opened 2020-07-09
Templated group on accessControlModal

minsunny:
nanopaySide : https://github.com/nanoPayinc/NANOPAY/pull/10176

https://nanopay.atlassian.net/browse/NP-1545

using templated groups not hard coding on AccessControlModal

Screen Shot 2020-07-08 at 9 31 34 PM
Screen Shot 2020-07-08 at 9 31 41 PM

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -127,14 +127,6 @@ foam.CLASS({
       class: 'Boolean',
       name: 'template',
       documentation: 'Used to specify that the group is used as a reference for the creation of other groups.'
-    },
-    {
-      class: 'String',
-      name: 'label',
-      documentation: 'Display name of the Group, if required.',
-      factory: function() {
-        return this.id;
-      }
     }
     /*
       FUTURE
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/Group.js
```diff
@@ -122,6 +122,11 @@ foam.CLASS({
         class: 'foam.u2.view.FObjectPropertyView',
         readView: { class: 'foam.u2.detail.VerticalDetailView' }
       }
+    },
+    {
+      class: 'Boolean',
+      name: 'template',
```
jlhughes, line 128: @zahnno this only works for this specific case.  To be honest, I'd rather hard coded groups for this ablii access thing, rather than this.   There would be no way to distinguish which group of templates were to be used together. 
The idea of access levels for abliii apps should be DAO based or perhaps we can use the maker/checker stuff that liquid is using. 
And each ablii app may have some app specific hierarchy.  So this is specific to payables for example.

jlhughes (changes requested):
See comments below, but let's stick with static/hard-coded list in ablii payables itself for now.

minsunny:
close this since decided not to use group.template property

PR #3682, opened 2020-07-09
create a section Axioms

hchoura:


src/foam/layout/SectionAxiom.js
```diff
@@ -43,6 +43,16 @@ foam.CLASS({
...
       value: function() { return true; }
+    },
+    {
+      class: 'String',
+      name: 'section',
+      getter: function() { return this.section_ ; },
+      setter: function(m) { this.section_ = m; }
+    },
+    {
+      class: 'Simple',
```
jlhughes, line 54: This skips the add to prototype.  If we just have 'section', can 'section_' be removed and then installInClass below may not be necessary. 
If we already have a lot of code dependent on this setup, then leave it.  But curious.

hchoura: what do you mean?

PR #3686, opened 2020-07-10
DIG/SUGAR CreatedByAware

nick-prat:
Adding this to make them SPID Aware


src/foam/nanos/dig/DIGOwnerRuleAction.js
```diff
@@ -0,0 +1,41 @@
+/**
```
jlhughes: Is this needed, can we not use 'createdBy' as that has both createdBy and createdByAgent.

jlhughes (commented):
See inline comment regarding using createdBy and CreatedByAgent. 
what is the purpose of the 'owner' property?

src/services
```diff
@@ -516,7 +516,33 @@ p({
...
+  "class":"foam.nanos.boot.NSpec",
+  "name":"digDAO",
+  "serve":true,
+  "serviceScript":
+  """
+  return new foam.dao.EasyDAO.Builder(x)
+    .setPm(true)
+    .setGuid(true)
+    .setOf(foam.nanos.dig.DIG.getOwnClassInfo())
+    .setCreatedByAware(true)
```
jlhughes: This is not necessary.

src/foam/nanos/dig/DIG.js
```diff
@@ -17,6 +17,7 @@ foam.CLASS({
 
   implements: [
     'foam.mlang.Expressions',
+    'foam.nanos.auth.CreatedByAware'
```
jlhughes: Didn't realize we didn't have createdBy, created, lastModified, ... 
Please add the other three.

src/foam/nanos/dig/DIG.js
```diff
@@ -40,10 +44,66 @@ foam.CLASS({
...
       name: 'id',
-      label: 'Request Name'
+      label: 'Request Name',
+      section: 'details'
+    },
+    {
+      class: 'Reference',
+      of: 'foam.nanos.auth.User',
+      name: 'createdBy',
+      storageOptional: true,
```
nanoscott: created* properties shouldn't be storageOptional, only lastModified* properties.

nanoscott (changes requested):

jlhughes (approved):

nanoscott (approved):

PR #3691, opened 2020-07-10
Np 1644/fix dot method

olhabn:
fixing NestedPropertiesExpression

@lchanmann @kgrgreer @jlhughes 
please review

### DIFF
--- a/src/foam/mlang/expr/Ref.js
+++ b/src/foam/mlang/expr/Ref.js
@@ -10,10 +10,9 @@ foam.CLASS({
   extends: 'foam.mlang.AbstractExpr',
   implements: [ 'foam.core.Serializable' ],
 
-  documentation: 'A Binary Predicate which returns reference property object',
+  documentation: 'An Unary Predicate which returns reference property object',
 
   javaImports: [
-    'foam.core.AbstractFObjectPropertyInfo',
     'foam.core.FObject',
     'foam.core.PropertyInfo',
     'foam.nanos.logger.Logger',
@@ -38,7 +37,8 @@ foam.CLASS({
         PropertyInfo p1 = (PropertyInfo) getArg1();
         FObject refObj = null;
         try {
-          refObj = (FObject)obj.getClass().getMethod("find" + StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
+          refObj = (FObject)obj.getClass().getMethod("find" + StringUtil.capitalize(p1.getName()), foam.core.X.class)
+            .invoke(obj, ((FObject)obj).getX());
         } catch ( Throwable t ) {
           Logger logger = (Logger) getX().get("logger");
           if ( logger == null ) {
@@ -51,11 +51,7 @@ foam.CLASS({
     },
 
     function comparePropertyValues(o1, o2) {
-      /**
-         Compare property values using arg2's property value comparator.
-         Used by GroupBy
-      **/
-      return this.arg2.comparePropertyValues(o1, o2);
+      return this.arg1.comparePropertyValues(o1, o2);
     }
   ]
 });
--- a/src/foam/mlang/expr/Ref.js
+++ b/src/foam/mlang/expr/Ref.js
@@ -10,7 +10,7 @@ foam.CLASS({
   extends: 'foam.mlang.AbstractExpr',
   implements: [ 'foam.core.Serializable' ],
 
-  documentation: 'An Unary Predicate which returns reference property object',
+  documentation: 'An Unary Expression which returns reference property object',
 
   javaImports: [
     'foam.core.FObject',
--- a/src/foam/mlang/expr/Ref.js
+++ b/src/foam/mlang/expr/Ref.js
@@ -23,14 +23,8 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'Class',
-      javaInfoType: 'foam.core.AbstractObjectPropertyInfo',
-      javaType: 'foam.core.ClassInfo',
-      name: 'arg1'
-    },
-    {
       class: 'foam.mlang.ExprProperty',
-      name: 'arg2'
+      name: 'arg1'
     }
   ],
 
@@ -38,13 +32,13 @@ foam.CLASS({
     {
       name: 'f',
       code: async function(o) {
-        return o[arg2.name + '$find'];
+        return o[arg1.name + '$find'];
       },
       javaCode: `
-        PropertyInfo p1 = (PropertyInfo) getArg2();
+        PropertyInfo p1 = (PropertyInfo) getArg1();
         FObject refObj = null;
         try {
-          refObj = (FObject)getArg1().getObjClass().getMethod("find" + StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
+          refObj = (FObject)obj.getClass().getMethod("find" + StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
         } catch ( Throwable t ) {
           Logger logger = (Logger) getX().get("logger");
           if ( logger == null ) {
--- a/src/foam/mlang/expr/Ref.js
+++ b/src/foam/mlang/expr/Ref.js
@@ -1,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.mlang.expr',
+  name: 'Ref',
+  extends: 'foam.mlang.AbstractExpr',
+  implements: [ 'foam.core.Serializable' ],
+
+  documentation: 'A Binary Predicate which returns reference property object',
+
+  javaImports: [
+    'foam.core.AbstractFObjectPropertyInfo',
+    'foam.core.FObject',
+    'foam.core.PropertyInfo',
+    'foam.nanos.logger.Logger',
+    'foam.nanos.logger.StdoutLogger',
+    'foam.util.StringUtil'
+  ],
+
+  properties: [
+    {
+      class: 'Class',
+      javaInfoType: 'foam.core.AbstractObjectPropertyInfo',
+      javaType: 'foam.core.ClassInfo',
+      name: 'arg1'
+    },
+    {
+      class: 'foam.mlang.ExprProperty',
+      name: 'arg2'
+    }
+  ],
+
+  methods: [
+    {
+      name: 'f',
+      code: async function(o) {
+        return o[arg2.name + '$find'];
+      },
+      javaCode: `
+        PropertyInfo p1 = (PropertyInfo) getArg2();
+        FObject refObj = null;
+        try {
+          refObj = (FObject)getArg1().getObjClass().getMethod("find" + StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
+        } catch ( Throwable t ) {
+          Logger logger = (Logger) getX().get("logger");
+          if ( logger == null ) {
+            logger = new StdoutLogger();
+          }
+          logger.error(t);
+        }
+        return refObj;
+      `
+    },
+
+    function comparePropertyValues(o1, o2) {
+      /**
+         Compare property values using arg2's property value comparator.
+         Used by GroupBy
+      **/
+      return this.arg2.comparePropertyValues(o1, o2);
+    }
+  ]
+});
--- a/src/foam/mlang/expr/Ref.js
+++ b/src/foam/mlang/expr/Ref.js
@@ -30,7 +30,7 @@ foam.CLASS({
   methods: [
     {
       name: 'f',
-      code: async function(o) {
+      code: function(o) {
         throw new Error('Ref is not supported');
       },
       javaCode: `
--- a/src/foam/mlang/expr/Ref.js
+++ b/src/foam/mlang/expr/Ref.js
@@ -31,7 +31,7 @@ foam.CLASS({
     {
       name: 'f',
       code: async function(o) {
-        return o[arg1.name + '$find'];
+        throw new Error('Ref is not supported');
       },
       javaCode: `
         PropertyInfo p1 = (PropertyInfo) getArg1();
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -3488,7 +3488,7 @@ foam.CLASS({
     function ENDS_WITH(a, b) { return this._binary_("EndsWith", a, b); },
     function FUNC(fn) { return this.Func.create({ fn: fn }); },
     function DOT(a, b) { return this._binary_("Dot", a, b); },
-    function REF(a, b) { return this._binary_("Ref", a, b); },
+    function REF(a) { return this._unary_("Ref", a); },
     function DOT_F(a, b) { return this._binary_("DotF", a, b); },
     function ADD() { return this._nary_("Add", arguments); },
     function SUB() { return this._nary_("Subtract", arguments); },
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -3255,34 +3255,29 @@ foam.CLASS({
         return this.arg2.f(this.arg1.f(o));
       },
       javaCode: `
-        if ( getArg1() instanceof PropertyInfo && getArg2() instanceof PropertyInfo ) {
-          StringBuilder sb = new StringBuilder("find");
-          PropertyInfo p1 = (PropertyInfo) getArg1();
-          FObject obj1;
-          if ( p1 instanceof AbstractFObjectPropertyInfo ) {
-            Object val = getArg1().f(obj);
-            return val == null ? null : getArg2().f(val);
-          }
-          try {
-            obj1 = (FObject)obj.getClass().getMethod(StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
-          } catch ( Throwable t ) {
-            return null;
-          }
-          if ( obj1 == null ) return null;
-          try {
-            return getArg2().f(obj1);
-          } catch ( Throwable t ) {
-            Logger logger = (Logger) getX().get("logger");
-            if ( logger == null ) {
-              logger = new StdoutLogger();
-            }
-            logger.error(t);
-            return null;
-          }
+        StringBuilder sb = new StringBuilder("find");
+        PropertyInfo p1 = (PropertyInfo) getArg1();
+        FObject obj1;
+        if ( p1 instanceof AbstractFObjectPropertyInfo ) {
+          Object val = getArg1().f(obj);
+          return val == null ? null : getArg2().f(val);
         }
-        Object val = getArg1().f(obj);
-        if ( val == null ) return null;
-        return getArg2().f(val);
+        try {
+          obj1 = (FObject)obj.getClass().getMethod(StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
+        } catch ( Throwable t ) {
+          return null;
+        }
+        if ( obj1 == null ) return null;
+        try {
+          return getArg2().f(obj1);
+        } catch ( Throwable t ) {
+          Logger logger = (Logger) getX().get("logger");
+          if ( logger == null ) {
+            logger = new StdoutLogger();
+          }
+          logger.error(t);
+          return null;
+        }
       `
     },
 
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -3266,7 +3266,7 @@ foam.CLASS({
   extends: 'foam.mlang.AbstractExpr',
   implements: [ 'foam.core.Serializable' ],
 
-  documentation: 'A Binary Predicate which applies arg2.f() to arg1.f().',
+  documentation: 'A Binary Predicate which returns reference property object',
 
   javaImports: [
     'foam.core.AbstractFObjectPropertyInfo',
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -3261,69 +3261,6 @@ foam.CLASS({
 
 
 foam.CLASS({
-  package: 'foam.mlang.expr',
-  name: 'Ref',
-  extends: 'foam.mlang.AbstractExpr',
-  implements: [ 'foam.core.Serializable' ],
-
-  documentation: 'A Binary Predicate which returns reference property object',
-
-  javaImports: [
-    'foam.core.AbstractFObjectPropertyInfo',
-    'foam.core.FObject',
-    'foam.core.PropertyInfo',
-    'foam.nanos.logger.Logger',
-    'foam.nanos.logger.StdoutLogger',
-    'foam.util.StringUtil'
-  ],
-
-  properties: [
-    {
-      class: 'Class',
-      javaInfoType: 'foam.core.AbstractObjectPropertyInfo',
-      javaType: 'foam.core.ClassInfo',
-      name: 'arg1'
-    },
-    {
-      class: 'foam.mlang.ExprProperty',
-      name: 'arg2'
-    }
-  ],
-
-  methods: [
-    {
-      name: 'f',
-      code: async function(o) {
-        return o[arg2.name + '$find'];
-      },
-      javaCode: `
-        PropertyInfo p1 = (PropertyInfo) getArg2();
-        FObject refObj = null;
-        try {
-          refObj = (FObject)getArg1().getObjClass().getMethod("find" + StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
-        } catch ( Throwable t ) {
-          Logger logger = (Logger) getX().get("logger");
-          if ( logger == null ) {
-            logger = new StdoutLogger();
-          }
-          logger.error(t);
-        }
-        return refObj;
-      `
-    },
-
-    function comparePropertyValues(o1, o2) {
-      /**
-         Compare property values using arg2's property value comparator.
-         Used by GroupBy
-      **/
-      return this.arg2.comparePropertyValues(o1, o2);
-    }
-  ]
-});
-
-
-foam.CLASS({
   package: 'foam.mlang.predicate',
   name: 'DotF',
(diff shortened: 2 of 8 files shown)


lchanmann:
@olhabn would be great if you could implement REF and use REF in the nested property expression instead of DOT. Please follow the discussion on dev-team channel for the distinction between DOT and REF.

olhabn:
> @olhabn would be great if you could implement REF and use REF in the nested property expression instead of DOT. Please follow the discussion on dev-team channel for the distinction between DOT and REF.

will do. thanks!

src/foam/mlang/mlang.js
```diff
@@ -3248,39 +3293,23 @@ foam.CLASS({
...
-          try {
-            return getArg2().f(obj1);
-          } catch ( Throwable t ) {
-            Logger logger = (Logger) getX().get("logger");
-            if ( logger == null ) {
-              logger = new StdoutLogger();
-            }
-            logger.error(t);
-            return null;
+        StringBuilder sb = new StringBuilder("find");
```
olhabn: let me know if i should refactor it to use invokeFinderMethod as Chanmann suggested for Dot in https://nanopay.slack.com/archives/G0336TYKV/p1594412722054800

src/foam/mlang/mlang.js
```diff
@@ -3225,6 +3225,49 @@ foam.CLASS({
...
+
+  methods: [
+    {
+      name: 'f',
+      code: function(o) {
+        return this.arg2.f(this.arg1.f(o));
+      },
+      javaCode: `
+        Object e1 = getArg1().f(obj);
+        if ( e1 == null ) return e1;
```
olhabn: the reason for this change is that 'Ref' could return null

src/foam/nanos/column/ColumnConfigPropertyValue.js
```diff
@@ -81,7 +81,8 @@ foam.CLASS({
               if ( p instanceof foam.core.AbstractFObjectPropertyInfo ) {
                 cls = p.getValueClass();
               } else {
```
olhabn, line 83: have no idea why it looked like that. maybe I reverted something on my columnconfig branch or something

src/foam/util/StringUtil.java
```diff
@@ -76,6 +76,6 @@ public static String capitalize(String s) {
     if ( SafetyUtil.isEmpty(s) ) return s;
     char[] chars = s.toCharArray();
     chars[0] = Character.toUpperCase(chars[0]);
-    return chars.toString();
```
olhabn, line 79: same here

src/foam/mlang/mlang.js
```diff
@@ -3225,6 +3225,49 @@ foam.CLASS({
...
+         Compare property values using arg2's property value comparator.
+         Used by GroupBy
+      **/
+      return this.arg2.comparePropertyValues(o1, o2);
+    }
+  ]
+});
+
+
+foam.CLASS({
```
jlhughes: Can you move this into its own source file, please - we are starting to break up these large js files.

src/foam/mlang/mlang.js
```diff
@@ -3248,39 +3293,22 @@ foam.CLASS({
...
-            Logger logger = (Logger) getX().get("logger");
-            if ( logger == null ) {
-              logger = new StdoutLogger();
-            }
-            logger.error(t);
-            return null;
+        PropertyInfo p1 = (PropertyInfo) getArg2();
+        FObject refObj = null;
+        try {
+          refObj = (FObject)getArg1().getObjClass().getMethod("find" + StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
```
jlhughes: Can you check if Relationship has a helper method to do this, so we only have method name creation logic in one place, please. 
if it doesn't, please add.

olhabn: yes, it does, but it's not really clear to me where we could retrieve a relationship for a reference property. is there some kind of relationship dao? thanks
is there Relationship class for backend?

src/foam/mlang/mlang.js

src/foam/mlang/mlang.js
olhabn: my understanding of relationship is that we use them for java code generation. so I'm wondering if we maybe create something like AbstractReferencePropertyInfo and copy info from relationship to it?

jlhughes: @olhabn ok, leave that for now and create another ticket to capture the support method - and ideas.

src/foam/mlang/expr/Ref.js
```diff
@@ -0,0 +1,61 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.mlang.expr',
+  name: 'Ref',
+  extends: 'foam.mlang.AbstractExpr',
+  implements: [ 'foam.core.Serializable' ],
+
+  documentation: 'A Binary Predicate which returns reference property object',
```
lchanmann: please update the documentation.

src/foam/mlang/expr/Ref.js
```diff
@@ -0,0 +1,61 @@
...
+
+  javaImports: [
+    'foam.core.AbstractFObjectPropertyInfo',
```
lchanmann: unused import

src/foam/mlang/expr/Ref.js
```diff
@@ -0,0 +1,61 @@
...
+    {
+      name: 'f',
+      code: async function(o) {
+        return o[arg1.name + '$find'];
+      },
+      javaCode: `
+        PropertyInfo p1 = (PropertyInfo) getArg1();
+        FObject refObj = null;
+        try {
+          refObj = (FObject)obj.getClass().getMethod("find" + StringUtil.capitalize(p1.getName()), foam.core.X.class).invoke(obj, ((FObject)obj).getX());
```
lchanmann: can you break this line into two? one for getMethod and another to invoke.

src/foam/mlang/expr/Ref.js
```diff
@@ -0,0 +1,61 @@
...
+        return refObj;
+      `
+    },
+
+    function comparePropertyValues(o1, o2) {
+      /**
+         Compare property values using arg2's property value comparator.
+         Used by GroupBy
+      **/
+      return this.arg2.comparePropertyValues(o1, o2);
```
lchanmann: there is no arg2

src/foam/mlang/expr/Ref.js
olhabn: fixed. thanks

src/foam/mlang/expr/Ref.js
olhabn: fixed.

src/foam/mlang/expr/Ref.js
olhabn: fixed.

src/foam/mlang/expr/Ref.js
olhabn: fixed.

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,95 @@ foam.CLASS({
...
-        }
+          name: 'prop',
+          javaType: 'foam.core.PropertyInfo',
+          javaInfoType: 'foam.core.AbstractObjectPropertyInfo'
+        },
       ],
-      code: function (prop1, prop2) {
-        return foam.mlang.Expressions.create().DOT(prop1, prop2);
+      code: function (of, propName, i, expr) {
+        return false;
```
lchanmann: is it always false for js?

olhabn: it was kinda not implemented.  sorry about that I've cleaned it up.

src/foam/mlang/expr/Ref.js
```diff
@@ -0,0 +1,57 @@
...
+    {
+      class: 'foam.mlang.ExprProperty',
+      name: 'arg1'
+    }
+  ],
+
+  methods: [
+    {
+      name: 'f',
+      code: async function(o) {
```
kgrgreer: JS predicates can't be async. Better to throw an exception.

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,92 @@ foam.CLASS({
...
-      code: function (of, propName, i) {
-        if ( i === propName.length - 1 )
-          return of.getAxiomByName(propName[i]);
+      code: function (of, propName, i, expr) {
         var prop = of.getAxiomByName(propName[i]);
-        return foam.mlang.Expressions.create().DOT(prop, this.returnPropExpr(prop, this.returnDotExprForNestedProperty(prop.of, propName, ++i)));
+        if ( ! prop ) return null;
+
+        var thisPropExpr; 
+        if ( foam.core.Reference.isInstance(prop) ) thisPropExpr = foam.mlang.Expressions.create().REF(prop);
```
kgrgreer: Please format this if/else and the next one normally (for FOAM). Thx.

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,92 @@ foam.CLASS({
...
+        if ( prop == null ) return null; 
+
+        Boolean isPropAReference = isPropertyAReference(ci, prop);
+
+        Expr thisPropExpr = null;
+        if ( isPropAReference )
+          thisPropExpr = REF(prop);
+        else
+          thisPropExpr = prop;
+        if ( expr == null) expr = thisPropExpr;
```
kgrgreer: same here.

src/foam/mlang/expr/Ref.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

lchanmann:
You can extract `buildPropertyExpr` method, something like
```
function buildPropertyExpr(prop, expr) {
  if ( foam.core.Reference.isInstance(prop) ) {
    prop = foam.mlang.Expressions.create().REF(prop);
  }

  return expr == null ? prop :
    foam.mlang.Expressions.create().DOT(expr, thisPropExpr);
}
```
and `returnDotExprForNestedProperty` would be shorter/easier to comprehend
```
function returnDotExprForNestedProperty(of, propName, i, expr) {
  var prop = of.getAxiomByName(propName[i]);
  if ( ! prop ) return null;

  var propExpr = buildPropertyExpr(prop, expr);

  if ( i === propName.length - 1 )
    return propExpr;

  return this.returnDotExprForNestedProperty(prop.of, propName, ++i, propExpr);
}
```
Same goes for Java and also extract `getPropertyClassInfo`.

src/foam/mlang/mlang.js
lchanmann: rename e1 to receiver and return null instead of the variable.

src/foam/mlang/expr/Ref.js
```diff
@@ -0,0 +1,57 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.mlang.expr',
+  name: 'Ref',
+  extends: 'foam.mlang.AbstractExpr',
+  implements: [ 'foam.core.Serializable' ],
+
+  documentation: 'An Unary Predicate which returns reference property object',
```
lchanmann: REF is an expression not predicate.

olhabn: noted and fixed. thanks

olhabn:
> You can extract `buildPropertyExpr` method, something like
> 
> ```
> function buildPropertyExpr(prop, expr) {
>   if ( foam.core.Reference.isInstance(prop) ) {
>     prop = foam.mlang.Expressions.create().REF(prop);
>   }
> 
>   return expr == null ? prop :
>     foam.mlang.Expressions.create().DOT(expr, thisPropExpr);
> }
> ```
> 
> and `returnDotExprForNestedProperty` would be shorter/easier to comprehend
> 
> ```
> function returnDotExprForNestedProperty(of, propName, i, expr) {
>   var prop = of.getAxiomByName(propName[i]);
>   if ( ! prop ) return null;
> 
>   var propExpr = buildPropertyExpr(prop, expr);
> 
>   if ( i === propName.length - 1 )
>     return propExpr;
> 
>   return this.returnDotExprForNestedProperty(prop.of, propName, ++i, propExpr);
> }
> ```
> 
> Same goes for Java and also extract `getPropertyClassInfo`.

fixed. please review

src/foam/mlang/mlang.js

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,118 @@ foam.CLASS({
...
       ],
-      code: function (of, propName, i) {
-        if ( i === propName.length - 1 )
-          return of.getAxiomByName(propName[i]);
+      code: function (of, propName, i, expr) {
         var prop = of.getAxiomByName(propName[i]);
-        return foam.mlang.Expressions.create().DOT(prop, this.returnPropExpr(prop, this.returnDotExprForNestedProperty(prop.of, propName, ++i)));
+        if ( ! prop ) return null;
+
+        expr = this.buildPropertyExpr(prop, expr);
```
lchanmann: Please create a new variable
```suggestion
        var propExpr = this.buildPropertyExpr(prop, expr);
```

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,118 @@ foam.CLASS({
...
-        PropertyInfo p = (PropertyInfo) ci.getAxiomByName(propName[i]);
-        if ( i == propName.length - 1 ) {
-          return p;
-        }
+        PropertyInfo prop = (PropertyInfo) ci.getAxiomByName(propName[i]);
+
+        if ( prop == null ) return null;
+
+        Boolean isPropAReference = isPropertyAReference(ci, prop);
+        expr = buildPropertyExpr(prop, expr, isPropAReference);
```
lchanmann: buildPropertyExpr should take a property and expression same as the JS one and in it you check if the property is a reference property.

I think you might not need classInfo argument for isPropertyAReference method.

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,118 @@ foam.CLASS({
...
+
+        if ( i == propName.length - 1 )
+          return expr;
         
-        StringBuilder sb = new StringBuilder("find");
         try {
+          StringBuilder sb = new StringBuilder("find");
```
lchanmann: can you extract the logic in try to method as well?

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,118 @@ foam.CLASS({
...
+        if ( isPropertyAReference )
+          thisPropExpr = REF(prop);
+        else
+          thisPropExpr = prop;
+        if ( expr == null )
+          expr = thisPropExpr;
+        else
+          expr = DOT(expr, thisPropExpr);
+
+        return expr;
```
lchanmann:
```suggestion
        if ( isPropAReference(prop) ) {
          prop = REF(prop);
        }
        
        return expr == null ? prop : DOT(expr, prop);
```

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,118 @@ foam.CLASS({
...
         }
       ],
-      code: function (prop1, prop2) {
-        return foam.mlang.Expressions.create().DOT(prop1, prop2);
+      code: function(prop, expr) {
+        if ( foam.core.Reference.isInstance(prop) ) {
+          prop = foam.mlang.Expressions.create().REF(prop);
+        }
+      
+        return expr == null ? prop :
```
lchanmann: expr is `undefined` when not given
```suggestion
        return ! expr ? prop :
```

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed. thanks

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: have problem with casting it back to Property as of now

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,120 @@ foam.CLASS({
...
         ClassInfo ci = of;
-        PropertyInfo p = (PropertyInfo) ci.getAxiomByName(propName[i]);
-        if ( i == propName.length - 1 ) {
-          return p;
-        }
+        PropertyInfo prop = (PropertyInfo) ci.getAxiomByName(propName[i]);
+
+        if ( prop == null ) return null;
+
+        Boolean isPropAReference = isPropertyAReference(prop);
```
lchanmann: unused variable

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,120 @@ foam.CLASS({
...
+      `
+    },
+    {
+      name: 'isPropertyAReference',
+      javaType: 'Boolean',
+      args: [
+        {
+          name: 'prop',
+          javaType: 'foam.core.PropertyInfo',
+          javaInfoType: 'foam.core.AbstractObjectPropertyInfo'
```
lchanmann: javaInfoType not needed

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,120 @@ foam.CLASS({
...
+        if ( ci == null )
+          return null; 
         
-        StringBuilder sb = new StringBuilder("find");
+        return returnDotExprForNestedProperty(ci, propName, ++i, expr);
+      `
+    },
+    {
+      name: 'getPropertyValueClass',
+      javaType: 'ClassInfo',
```
lchanmann: it returns a class info, should it be getPropertyClassInfo?

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,120 @@ foam.CLASS({
...
           Class cls;
-          if ( p instanceof foam.core.AbstractFObjectPropertyInfo ) {
-            cls = p.getValueClass();
-          } else {
-            sb.setLength(4);
-            Method m = ci.getObjClass().getMethod(StringUtil.capitalize(p.getName()), foam.core.X.class);
+          if ( ! isPropAReference )
+            cls = prop.getValueClass();
+          else {
+            Method m = ci.getObjClass().getMethod("find" + StringUtil.capitalize(prop.getName()), foam.core.X.class);
```
lchanmann: if prop.getClassInfo() and ci are the same we can spare taking ci as argument.

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +81,120 @@ foam.CLASS({
...
+          javaType: 'foam.core.ClassInfo'
+        },
+        {
+          name: 'prop',
+          javaType: 'foam.core.PropertyInfo',
+        }
+      ],
+      javaCode: `
         try {
+          Boolean isPropAReference = isPropertyAReference(prop);
```
lchanmann: isPropertyAReference does a reflection call and this method does the same reflection call again on line 141. If we have something like `getFinderMethod(prop)` then reuse the result there would be one reflection call.

src/foam/mlang/mlang.js
```diff
@@ -3223,16 +3223,7 @@ foam.CLASS({
...
-
-  javaImports: [
-    'foam.core.AbstractFObjectPropertyInfo',
-    'foam.core.FObject',
-    'foam.core.PropertyInfo',
-    'foam.nanos.logger.Logger',
-    'foam.nanos.logger.StdoutLogger',
-    'foam.util.StringUtil'
-  ],
+  documentation: 'A Binary Expression which applies arg2.f() to arg1.f().',
```
lchanmann: while you are here, can you also change this to
```suggestion
  documentation: `
    A Binary Expression which evaluates arg1 and passes the result to arg2.
    In other word, the output of arg1 is the receiver of arg2.
    
    For example, to get city from user address:
    
        DOT(User.ADDRESS, Address.CITY).f(user); // return user.address.city
        
    `,
```

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed. thanks

src/foam/mlang/expr/Ref.js
jlhughes: The 'async' should be removed, correct? @kgrgreer

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +83,124 @@ foam.CLASS({
...
-        ClassInfo ci = of;
-        PropertyInfo p = (PropertyInfo) ci.getAxiomByName(propName[i]);
-        if ( i == propName.length - 1 ) {
-          return p;
-        }
+        PropertyInfo prop = (PropertyInfo) of.getAxiomByName(propName[i]);
+
+        if ( prop == null ) return null;
+
+        Method m = getFinderMethod(prop);
```
lchanmann: unused variable

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +83,124 @@ foam.CLASS({
...
+        expr = buildPropertyExpr(prop, expr);
```
lchanmann: please create new variable
```suggestion
        var propExpr = buildPropertyExpr(prop, expr);
```

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -81,61 +83,124 @@ foam.CLASS({
...
         }
+      `
+    },
+    {
+      name: 'getPropertyClassInfo',
+      javaType: 'ClassInfo',
+      javaThrows: [
+        'NoSuchMethodException',
+        'IllegalAccessException',
+        'InvocationTargetException'
```
lchanmann: should be `javaThrows: [ 'java.lang.Exception' ]` because line 141 can return null.

olhabn: fixed

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed.

src/foam/nanos/column/NestedPropertiesExpression.js
olhabn: fixed.

lchanmann (approved):
Looks good. Thanks @olhabn

olhabn:
I was fixing an exception. The cause of exception is that I didn't handle use case of retrieving something like "address.countryId", which looks totally legit to me, but we were using REF for countryId which caused the exception. so I added lastPropName check and if reference property is the last one in nested property's name string then it would be treated as not reference property.
please let me know if it makes sense or if I shouldn't allow such properties
I'd really appreciate a review with my last commit included @lchanmann 
thanks!

lchanmann:
@olhabn you don't need that lastPropName argument and just return DOT in `if ( i === propName.length - 1)` block and move buildPropertyExpr call below it likes so
```
function (of, propName, i, expr) {
  var prop = of.getAxiomByName(propName[i]);
  if ( ! prop ) return null;

  if ( i === propName.length - 1 )
    return ! expr ? prop :
      foam.mlang.Expressions.create().DOT(expr, prop);

  var propExpr = this.buildPropertyExpr(prop, expr);
  return this.returnDotExprForNestedProperty(prop.of, propName, ++i, propExpr);
}
```

olhabn:
> @olhabn you don't need that lastPropName argument and just return DOT in `if ( i === propName.length - 1)` block and move buildPropertyExpr call below it likes so
> 
> ```
> function (of, propName, i, expr) {
>   var prop = of.getAxiomByName(propName[i]);
>   if ( ! prop ) return null;
> 
>   if ( i === propName.length - 1 )
>     return ! expr ? prop :
>       foam.mlang.Expressions.create().DOT(expr, prop);
> 
>   var propExpr = this.buildPropertyExpr(prop, expr);
>   return this.returnDotExprForNestedProperty(prop.of, propName, ++i, propExpr);
> }
> ```

fixed. thanks!

lchanmann (approved):
@jlhughes @kgrgreer I'm good with this. Please take another look.

src/foam/mlang/expr/Ref.js
kgrgreer: Don't make "async".

olhabn: fixed

jlhughes (approved):

PR #3692, opened 2020-07-13
SVG Graph and CrunchLab

KernelDeimos:
PR adds a Graph model, and an SVG graph renderer. A u2 controller called CrunchLab is added to this pr as an example of rendering the graph.


src/foam/graph/graph.js
```diff
@@ -0,0 +1,145 @@
+foam.CLASS({
```
jlhughes: license header please.

jlhughes (commented):
Can you move all these classes into their own files. Otherwise it's impossible to debug in the browser.

src/foam/graph/graph.js
```diff
@@ -0,0 +1,80 @@
+/**
```
jlhughes: Thanks for splitting. One last request; to be consistent, can you Capitalize this source file please.

src/foam/u2/crunch/lab/CrunchLab.js
```diff
@@ -0,0 +1,68 @@
+foam.CLASS({
```
jlhughes, line 7: license headers please.

src/foam/graph/graph.js
KernelDeimos: Looks like Git didn't let me do it the first time. Found a way around it

jlhughes (approved):

PR #3693, opened 2020-07-13
Make SUGAR Work

nanoscott:
Currently SUGAR does not work as intended. I am simplifying it so that it will work for now and I will add tasks to improve it and bring it back up to standard after the next deployment.

image


src/foam/nanos/dig/SUGAR.js
```diff
@@ -248,29 +67,25 @@ foam.CLASS({
...
-            });
-            return temp;
-          }, async function(error) {
-            var temp = await error.payload.then(function(result) {
-              return result;
-            });
-            return temp;
+        var req = this.HTTPRequest.create({
+          url: window.location.protocol + '//' + window.location.hostname + ':' + window.location.port + '/service/sugar?sessionId=' + localStorage.defaultSession,
+          method: 'POST',
```
jlhughes, line 135: Future - since we are POSTing we can put the SessionID in as a bearer token.

jlhughes (approved):

src/foam/nanos/dig/SUGAR.js
```diff
@@ -41,205 +39,26 @@ foam.CLASS({
       name: 'id',
       displayWidth: 40
     },
-    {
-      class: 'String',
-      name: 'serviceKey',
```
kgrgreer, line 109: Why removing the serviceKey? How will the user know what services are available?

src/foam/nanos/dig/SUGAR.js
```diff
@@ -41,205 +39,26 @@ foam.CLASS({
...
-          this.argumentInfo = null;
-          this.currentMethod = '';
-        }
-
-        this.postData = '';
-      }
-    },
-    {
-      class: 'String',
-      name: 'method',
```
kgrgreer, line 168: Why removing?

src/foam/nanos/dig/SUGAR.js
nanoscott: this doesn't work, that's why I'm removing it.

src/foam/nanos/dig/SUGAR.js
nanoscott: same thing here.

src/foam/nanos/dig/SUGAR.js
nanoscott: They will have to find out what is available through documentation

kgrgreer: I would rather just fix it, since the dynamically generated list which is based on their current permissions and capabilities will always be correct, whereas documentation may not be.

nanoscott:
@jlhughes .. conflicts have been resolved. Please go ahead with the merge to the sugar branch now.

PR #3694, opened 2020-07-13
Np 1661/fix bugs table columns expression bug and account balance and home balance properties

olhabn:
fixing errors

### DIFF
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -75,7 +75,7 @@ foam.CLASS({
       expression: function(columns, of, editColumnsEnabled, selectedColumnNames, allColumns) {
         if ( ! of ) return [];
         var cols;
-        if ( ! editColumnsEnabled ) cols = columns;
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -75,7 +75,7 @@ foam.CLASS({
       expression: function(columns, of, editColumnsEnabled, selectedColumnNames, allColumns) {
         if ( ! of ) return [];
         var cols;
-        if ( ! editColumnsEnabled ) cols = columns;
+        if ( ! editColumnsEnabled ) cols = of.getAxiomByName('tableColumns').columns || allColumns;
         else cols = selectedColumnNames;
```
olhabn, line 79: for some reason 'columns' expression is unreachable that's why code moved here

olhabn: so I'm wondering if we could change unitPropName to array of names or maybe there is a better solution

kgrgreer: The 'columns' expression would not be called if that field's value were explicitly set. Why an array?

PR #3700, opened 2020-07-15
Add DAOWizardlet

KernelDeimos:
This PR adds a wizardlet which puts its data into the corresponding DAO (or one specified by a `daoKey`).

The following example, which can be run in the Javascript console, creates a wizard for adding Cron jobs:

```
ctrl.add(foam.u2.dialog.Popup.create({ closeable: false }, ctrl.__subContext__).tag({
  class: 'foam.u2.wizard.StepWizardletView',
  data: foam.u2.wizard.StepWizardletController.create({
    wizardlets: [ foam.u2.wizard.DAOWizardlet.create({
      title: 'Cron Wizard',
      of: 'foam.nanos.cron.Cron',
    }, ctrl.__subContext__) ]
  }),
  onClose: (x) => {
    x.closeDialog();
  }
}));
```

(note that `ctrl.__subContext__` arguments are only needed if you run this from the javascript console; in a regular view it should work as expected without those)

### DIFF
--- a/src/foam/u2/wizard/DAOWizardlet.js
+++ b/src/foam/u2/wizard/DAOWizardlet.js
@@ -1,0 +1,43 @@
+foam.CLASS({
+  package: 'foam.u2.wizard',
+  name: 'DAOWizardlet',
+  extends: 'foam.u2.wizard.BaseWizardlet',
+
+  properties: [
+    {
+      name: 'of',
+      class: 'Class'
+    },
+    {
+      name: 'data',
+      class: 'FObjectProperty',
+      of: 'foam.core.FObject',
+      factory: function () {
+        return this.of.create({}, this.__context__);
+      }
+    },
+    {
+      name: 'daoKey',
+      class: 'String',
+      expression: function (of) {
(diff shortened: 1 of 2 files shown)


KernelDeimos:
~This maybe works fine, but right now if it's used in StepWizardletView the submit button doesn't work. I'm currently investigating the cause of this.~ I was testing it wrong

src/foam/u2/wizard/DAOWizardlet.js
```diff
@@ -0,0 +1,43 @@
+foam.CLASS({
```
kgrgreer, line 7: copyright

KernelDeimos: One day I'll submit a PR with all the license headers already added. Maybe not today, but one day

KernelDeimos: It's fixed on my branch but Github isn't updating this PR for some reason

KernelDeimos:
...

PR #3705, opened 2020-07-15
adding homeDenominationSessionService

olhabn:
ticket https://nanopay.atlassian.net/browse/NP-1683
nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/10211

Please review @jlhughes @kgrgreer


src/foam/nanos/session/Session.js
```diff
@@ -136,6 +136,11 @@ foam.CLASS({
       javaFactory: 'return reset(getX());',
       hidden: true,
       transient: true
+    },
```
kgrgreer: Let's not add this direction to the Session, otherwise we would be adding stuff to the session all of the time and it would be specific to our use case rather than generic. Instead, the Session has a Context, which is where the value can be added generically. My mistake. When I say to "add something to the Session", what I really mean is "add to the Session Context". Otherwise, everything looks good.

olhabn: do you mind me using just a "homeDenomination" string? if not, that it's fixed

src/foam/core/Currency.js
```diff
@@ -27,7 +27,7 @@
     {
       buildJavaClass: function(cls) {
         cls.extras.push(`
-          public String format(Long amount) {
+          public String format(long amount) {
```
olhabn, line 30: @jlhughes please let me know if you don't mind such changes ( you are the last person who worked on Currency according to git blame). thanks

jlhughes:
I would prefer renaming all of this to LocalSettings / LocalSettingsService or UserSettings ... service, which pushes all known localStorage values back to the server on login.  LocalSettings is owned by User.  
homeDenomination being just one of the possibly many local storage values. 
Could have a model with a map of key/value pairs, or the service just puts the key/value pairs in map which is added to the context.

src/foam/core/Currency.js
jlhughes: How did this work before?

olhabn:
> LocalSettingsService

fixed

src/foam/core/Currency.js
olhabn: pretty much the same

src/foam/nanos/session/services/LocalSettingsService.js
```diff
@@ -0,0 +1,19 @@
+foam.INTERFACE({
```
jlhughes: This service is for setting/getting a particular settings.
so two options.
LocalSetting is an model of 
id String
value String
and then this is just a DAO.
The client does normal DAO puts on startup. 
The client can also select from the server and do a sync - update client with those on server and update server with those on client. 

If you go the model/DAO route then you don't need this 'service', just a couple DAO decorators for client/server and something on the client side (might be something in EasyDA) - which runs when the DAO is instantiated.

olhabn: are we gonna store homeDenomination in DAO? I thought we just need to use it for homeBalance calculation.

or is it a DAO to store the name of local settings we set in Session Context to make a bulk set( read all of the settings from localStorage and set them all for session )?

jlhughes: This is about all localStorage properties, homeDenomination is just one of many.
The DAO is localSetting - id,value pair. 
The id in this case is 'homeDenomination' and the value is whatever it needs to be. 
Retrieving - two options: 
1. explicitly query localSettingDAO for some id 
2. place all localSetting entries into the Session, the lookup via context key.

src/foam/core/Currency.js
kgrgreer: java auto-boxing will convert, just less efficient.

src/foam/nanos/session/services/LocalSettingsService.js
kgrgreer: But we don't want the home denomination to be global for all of a user's logins, since they could have a different one for each device they're logged into. Would it just be a transient DAO in their context?

jlhughes: yes, all transient. 
The dao can be in their session or each entry can be added to the session (context). 
To deal with Chrome deleting localStorage, we can store a device id with the setting if we want to store for the user (permanently). 
So when the client instance starts it loads localStorage then sends all to the server - and those entries are added to the users localSettingDAO- in their session context. 
And future, when localStorage is empty, the client can query for last saved settings for that device and recreate localStorage. 
And perhaps we should just call it localStorage (model) and localStorageDAO .

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -549,6 +549,8 @@ foam.CLASS({
       } else if ( this.theme ) {
         this.window.location.hash = this.theme.defaultMenu;
       }
+
+      this.__subContext__.localSettingsDAO.put(foam.nanos.session.local_settings.LocalSetting.create({name: 'homeDenomination', value: 'USD'}));
     },
```
olhabn, line 554: localSettingsDAO not accessible here.

src/foam/nanos/session/Session.js
```diff
@@ -249,6 +249,9 @@ foam.CLASS({
 
           rtn = rtn.put("appConfig", appConfig);
 
+          DAO localSettingsDAO = new foam.dao.MDAO(foam.nanos.session.local_settings.LocalSetting.getOwnClassInfo());
+          rtn.put("localSettingsDAO", localSettingsDAO);
```
jlhughes:
```suggestion
          rtn = rtn.put("localSettingsDAO", localSettingsDAO);
```

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -549,6 +549,8 @@ foam.CLASS({
       } else if ( this.theme ) {
         this.window.location.hash = this.theme.defaultMenu;
       }
+
+      this.__subContext__.localSettingsDAO.put(foam.nanos.session.local_settings.LocalSetting.create({name: 'homeDenomination', value: 'USD'}));
```
jlhughes: local_settings -> localSettings (for package name).
And in this case fine will just putting it in the Session package directly.

src/foam/nanos/session/Session.js
jlhughes: Since the model is singular, make the dao name singular. 
localSettingDAO

src/foam/nanos/session/local_settings/LocalSetting.js
```diff
@@ -0,0 +1,22 @@
+foam.CLASS({
+  package: 'foam.nanos.session.local_settings',
+  name: 'LocalSetting',
+  properties: [
+    {
+      name: 'id',
```
jlhughes: just make the name the id.  What was your intent for ID? 
If we plan to store these permanently later then each will have a User reference, and yes, then we would need an ID - but it can be an id of name, user reference and device id.  
so in general wouldn't worry about that just now.

src/foam/nanos/session/services/ClientLocalSettingsServiceImpl.js
```diff
@@ -0,0 +1,14 @@
+foam.CLASS({
```
jlhughes: So none of this is required now, correct?

src/foam/nanos/u2/navigation/SignIn.js
```diff
@@ -124,6 +124,11 @@ foam.CLASS({
           this.auth.login(X, this.identifier, this.password).then(
             (logedInUser) => {
               if ( ! logedInUser ) return;
+
+              try {
+                X.localSettingsService.addHomeDenomination(X, localStorage.getItem('homeDenomination'));
```
jlhughes: Changes in the application controller should do this 'add' for you.

src/services
```diff
@@ -1122,4 +1122,26 @@ p({
...
+  "authenticate": false,
+  "serviceScript":"""
+    import foam.dao.DAO;
+    dao = x.get("localLocalSettingDAO");
+
+    if(dao == null)
+      dao = new foam.dao.NullDAO(x, foam.nanos.session.LocalSetting.getOwnClassInfo());
+    return new foam.dao.EasyDAO.Builder(x)
+      .setInnerDAO(dao)
+      .setAuthorize(false)
```
olhabn, line 1117: I will fix setAuthorize later

src/foam/nanos/session/services/ClientLocalSettingsServiceImpl.js
olhabn: yep. cleaned up. I don't know what I did wrong but DAO doesn't work the way it should ( DAO that used on put is not a Session context DAO ). I spent half of a day trying to figure out how to make it work but it was unsuccessful. I'd really appreciate your thoughts on my latest changes https://github.com/foam-framework/foam2/pull/3705/commits/8eba18631a66bcd017bb0821f3453afedcb8c964 . thanks!

src/services
jlhughes: authorize false is fine, it's only in the logged in users context.

src/foam/nanos/session/LocalSetting.js
```diff
@@ -0,0 +1,14 @@
+foam.CLASS({
```
jlhughes, line 7: license header.

src/foam/nanos/session/LocalSettingSessionDAO.js
```diff
@@ -0,0 +1,45 @@
+foam.CLASS({
```
jlhughes, line 7: license header.

src/foam/nanos/session/LocalSetting.js
olhabn: fixed

src/foam/nanos/session/LocalSettingSessionDAO.js
olhabn: fixed

jlhughes (approved):

PR #3715, opened 2020-07-23
Fix adding contact from scratch - FObjectVIew

minsunny:
nanopay : https://github.com/nanoPayinc/NANOPAY/pull/10276

Screen Shot 2020-08-19 at 12 57 20 PM

https://nanopay.atlassian.net/browse/NP-214


zahnno:
What happened to the predicate for the strategizer? We need it to omit classes somehow.

minsunny:
> What happened to the predicate for the strategizer? We need it to omit classes somehow.

@zahnno  those were removed by @kgrgreer in these changes https://github.com/foam-framework/foam2/pull/3622/files
I am just trying to fix based on the latest version but we need to see his review and opinion.

kgrgreer:
Let's try doing with copyFrom() or our own version of copyFrom() instead.

src/foam/u2/view/FObjectView.js
```diff
@@ -164,23 +169,28 @@ foam.CLASS({
 
       var classToData = function(c) {
         var m = c && this.__context__.lookup(c, true);
-        return m ? m.create(null, this) : null;
+        return m ? m.create(this.persistantData ? this.persistantData : null, this) : null;
```
kgrgreer: Does it also work if you just do: return m.create(this.data, this);

PR #3717, opened 2020-07-23
[NP-1712] Update step wizardlet controller

KernelDeimos:
This PR removes the StackView from StepWizardletView.

The position of a substack was originally being used to keep track of the wizard screen, but this wasn't a good approach because
- the screen is really identified by two indices (wizardlet and section within wizardlet's data model), and
- unavailable sections had to be pushed to the stack as blank screens before reaching the next available one

This also removes logic from `canSkipTo` which needs to be re-evaluated.

### DIFF
--- a/src/foam/u2/wizard/StepWizardConfig.js
+++ b/src/foam/u2/wizard/StepWizardConfig.js
@@ -1,0 +1,17 @@
+foam.CLASS({
+  package: 'foam.u2.wizard',
+  name: 'StepWizardConfig',
+
+  properties: [
+    {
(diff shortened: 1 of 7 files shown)


src/foam/u2/wizard/StepWizardConfig.js
```diff
@@ -0,0 +1,17 @@
+foam.CLASS({
```
jlhughes, line 7: license header please.

jlhughes (approved):

jlhughes (commented):
license header on StepWizardConfig.

PR #3726, opened 2020-07-28
[NP-1632] Capability View

jinnnn9:
- addresses: https://nanopay.atlassian.net/browse/NP-1632
Screen Shot 2020-07-28 at 1 02 58 AM

Screen Shot 2020-07-28 at 10 45 22 AM

- UI seems a bit off, but we should be able to fix this issue by adjusting css for the view that embeds the capability view.

### DIFF
--- a/src/foam/nanos/crunch/ui/CapabilityView.js
+++ b/src/foam/nanos/crunch/ui/CapabilityView.js
@@ -2,7 +2,7 @@ foam.CLASS({
   package: 'foam.nanos.crunch.ui',
   name: 'CapabilityView',
   extends: 'foam.u2.View',
-  documentation: 'All purpose capability view that takes in an array of capaiblity ids and display these capabilities',
+  documentation: 'All purpose capability view that takes in an array of capaiblity ids and displays these capabilities',
 
   imports: [
     'crunchController',
@@ -37,17 +37,17 @@ foam.CLASS({
     async function initE() {
       const self = this;
 
-      // get the wizardlets and sections for all of the capabilties
+      // gets the wizardlets and sections for all of the capabilties
       for ( let cap of this.capabilities ) {
         // wizardlets for current capability
         const curWizardlets = (await this.crunchController.getCapsAndWizardlets(cap)).wizCaps;
 
-         // add listener on curWizardlets
+         // adds listener on curWizardlets
          for ( let wizardlet of curWizardlets ) {
           this.addListeners(wizardlet.data);
         }
 
-        // get all the sections associated with curWizardlets
+        // gets all the sections associated with curWizardlets
         const curWizardletSectionsList = this.crunchController.generateSections(curWizardlets);
 
         this.wizardlets = this.wizardlets.concat(curWizardlets);
@@ -76,7 +76,7 @@ foam.CLASS({
 
       obj.sub(this.updateWizardlet);
 
-      // add listeners on inner fobjects
+      // adds listeners on inner fobjects
       for ( let value of Object.values(obj.instance_) ) {
         if ( this.isFObject(value) ) {
           this.addListeners(value);
--- a/src/foam/nanos/crunch/ui/CapabilityView.js
+++ b/src/foam/nanos/crunch/ui/CapabilityView.js
@@ -17,6 +17,16 @@ foam.CLASS({
     'foam.u2.detail.SectionView'
   ],
 
+  constants: [
+    {
+      name: 'MAX_DEPTH',
+      type: 'Integer',
+      value: 5,
+      documentation: 'max. depth addListeners can move down on recursion'
+    }
+
+  ],
+
   properties: [
     {
       class: 'StringArray',
@@ -52,11 +62,11 @@ foam.CLASS({
         const { caps: curCaps, wizCaps: curWizardlets } =
           await this.crunchController.getCapsAndWizardlets(capID);
 
-        // pre-populate curWizardlets' data
-        this.populateData(curWizardlets);
-
-        // add listeners on curWizardlets
-        curWizardlets.forEach(wizardlet => this.addListeners(wizardlet, wizardlet.data));
+        // pre-populate curWizardlets' data and add listeners to it
+        for ( let wizardlet of curWizardlets ) {
+          this.populateData(wizardlet);
+          this.addListeners(wizardlet, wizardlet.data);
+        }
 
         // get all the sections associated with curWizardlets
         const curWizardletSectionsList = this.crunchController.generateSections(curWizardlets);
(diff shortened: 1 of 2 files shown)


src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }
+  ],
+
+  listeners: [
+    {
+      name: 'updateWizardlet',
+      code: function() {
+        // TODO: can we only update the wizardlet that has new data instead of updating all the wizardlets?
+        // TODO: when/where do we call finalOnClose?
```
Nauna: we need to make some mods to finalOnClose - and will keep you posted. However I'm under the impression that it can be called here, after save.

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,105 @@
...
+        return true;
+      }
+    }
+  ],
+
+  listeners: [
+    {
+      name: 'updateWizardlet',
+      code: function() {
+        // TODO: can we only update the wizardlet that has new data instead of updating all the wizardlets?
```
Nauna: think what you have works. The listener listens to changes. When the data changes we can assume new data.

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,105 @@
...
+                ));
+              });
+            }));
+          });
+        }))
+      .end();
+    },
+
+    function addListeners(obj) {
+      if ( ! this.isFObject(obj) ) return;
```
Nauna: // This adds a listener to each wizardlet.data obj for the purpose of saving all user inputs. Releasing calling views from the responsibility.

^ want to suggest a bit of documentation to this function

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,105 @@
...
+      const self = this;
+
+      this.start().addClass(this.myClass())
+        .add(this.slot(function(capabilities) {
+          return this.E().forEach(capabilities, async function(cap) {
+            // get wizardlets for the capability and all of its dependent capabilities
+            self.wizardlets = (await self.crunchController.getCapsAndWizardlets(cap)).wizCaps;
+
+            // add listeners on wizardlets
+            for ( const wizardlet of self.wizardlets ) {
```
Nauna: sorry for being a burden, but const is an interesting choice. It works, however const creates a new block scope with each iteration ... 'let' would be my choice.

src/foam/nanos/crunch/ui/CapabilityView.js
jinnnn9: Don't we want to call the function at the end of the process (thus the name)? It looks like an overkill to call the function each time listeners are hit.

src/foam/nanos/crunch/ui/CapabilityView.js
jinnnn9: What I am doing here is if one wizardlet is updated, we save all the wizardlets which seemed inefficient to me. I was wondering if there is a way to know on which wizardlet the listener was triggered so that I can only save that wizardlet.

src/foam/nanos/crunch/ui/CapabilityView.js
Nauna: I agree it does seem messy, however I would argue that this should be a minimal call due to the nature of this view. 
1) This view is not for a full feature, it is for a few capabilities that are dependencies of legacy objects. 
2) This view is also for getting additional requirements from within a feature ( example: our invoice flow concept with paymentProvidorCapabilities), however again I would argue that these capabilities will be for additional information, simple dynamic capabilities that will have little more than a few properties. Therefore the call to the listener will be minimal.

src/foam/nanos/crunch/ui/CapabilityView.js
Nauna: First good catch, sorry I missed that.

So take a look at **com.google.foam.demos.sweeper.Board**, it will give an example of passing in a value to a listener. I would suggest adjusting code to pass in wizardlet into the listener then just saving that wizardlet.

Also think adding an init method and moving the adding the listeners to that. initE can get called multiple times.

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,111 @@
...
+            ));
+          });
+        }))
+      .end();
+    },
+
+    // adds a listener to each wizardlet.data obj for the purpose of saving
+    // all user inputs releasing calling views from the responsibility
+    function addListeners(obj) {
+      if ( ! this.isFObject(obj) ) return;
```
zahnno: why not just check for null? You're checking for isFObject below and the first iteration through the function will be an object or null since it references the "of" of a capability.

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,111 @@
...
+
+      obj.sub(this.updateWizardlet);
```
zahnno:
```suggestion
      obj.sub(this.updateWizardlet.bind(this, obj));
```
To reference wizardlet in listener.

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,111 @@
...
+      if ( typeof obj === 'object' && obj.instance_ ) {
+        return true;
+      }
+    }
+  ],
+
+  listeners: [
+    {
+      name: 'updateWizardlet',
+      code: function() {
```
zahnno:
```suggestion
      code: function(wizardlet) {
```
In regards to optimization - avoid saving all wizardlets.

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,111 @@
...
+        // TODO: can we only update the wizardlet that has new data instead of updating all the wizardlets?
+        // TODO: when/where do we call finalOnClose?
+        for ( let wizardlet of this.wizardlets ) {
+          try {
```
zahnno: In regards to optimization comments - remove for loop and call `wizardlet.save();` in conjunction with previous comment suggestions.

src/foam/nanos/crunch/ui/CapabilityView.js
jinn9: @zahnno Yes, if we call the function inside the loop, it will guarantee that the argument passed to the function will be fobject. But if we call the function for the first time, there is no protection against passing a non fobject to the function. In this case, calling sub on `obj` will raise an error.

jinn9: @zahnno Since we know `wizardlet.data` is an fobject, it will run into the case above, but I still added there for extra layer of protection.

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,166 @@
+foam.CLASS({
+  package: 'foam.nanos.crunch.ui',
+  name: 'CapabilityView',
+  extends: 'foam.u2.View',
+  documentation: 'All purpose capability view that takes in an array of capaiblity ids and displays these capabilities',
```
zahnno:
```suggestion
  documentation: 'All purpose capability view that takes in an array of capability ids and displays those capabilities',
```

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,166 @@
+foam.CLASS({
```
zahnno, line 7: Add license headers.

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,166 @@
...
+        }))
+      .end();
+    },
+
+    // add a listener to wizardlet.data and each wizardlet.data obj
+    // for the purpose of saving all user inputs releasing
+    // calling views from the responsibility
+    function addListeners(wizardlet, obj, depth=0) {
+      // Some obj (e.g., dao) have a long chain of nested objects within them
+      // which casues a stack overflow. Such objects don't need listeners on them
```
zahnno:
```suggestion
      // which causes a stack overflow. Such objects don't need listeners on them
```

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+  ],
+
+  properties: [
+    {
+      class: 'StringArray',
+      name: 'capabilityIDs'
+    },
+    {
```
Nauna, line 41: I would really recommend using definitions of variable types in these properties.

capabilityIDs - no problems
capabilities - should be class: 'FObjectArray', of: 'Capabilities',
wizardlets - 'FObjectArray',of: 'BaseWizardlet',
wizardletSectionsList -  class: 'FObjectArray', of: 'foam.layout.Section',
showTitle - class: 'String'

src/foam/nanos/crunch/ui/CapabilityView.js
```diff
@@ -0,0 +1,172 @@
...
+    },
+  ],
+
+  listeners: [
+    {
+      name: 'updateWizardlet',
+      code: function(wizardlet) {
+        try {
+          wizardlet.save();
+          this.saveNoDataCaps(this.capabilities);
```
Nauna, line 170: I'd recommend removing the saveNoDataCaps - and using the crunchController.finalOnClose() function here.

I realize finalOnClose isn't perfect with it's console.log and x.closeDialog() - however neither are going to affect you.

Nauna (approved):
Overall looks nice!

@kgrgreer can you take a look at the MAX_DEPTH and the use of isFObject.

src/foam/nanos/crunch/ui/CapabilityView.js
jinnnn9: wizardletSectionsList is an array of arrays so I had to give 'Array' class.

src/foam/nanos/crunch/ui/CapabilityView.js
jinnnn9: I ran into a few issues using finalOnClose. First, if there is no modal to close, x.closeDialog() will raise an error. Also, it closes the modal right away when the user clicks the agreement.

zahnno (approved):

PR #3727, opened 2020-07-28
Add ChooseNView

KernelDeimos:
ChooseNView allows selecting up to `n` of the specified choices. Each choice is rendered using `CheckBox` by default, but this can be overridden using the `booleanView` property.

This change is related to rendering an "AtLeastCapability" (tentative name, referred to as `OrCapability` in tickets) capabilities in the step wizard. This view helps to guarantee the maximum number of selections isn't surpassed while a validation elsewhere can prevent continuation until the minimum number of selections is met.

### DIFF
--- a/src/foam/u2/view/UnstyledChooseNView.js
+++ b/src/foam/u2/view/UnstyledChooseNView.js
@@ -1,0 +1,83 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.view',
+  name: 'UnstyledChooseNView',
+  extends: 'foam.u2.View',
+
+  requires: [
+    'foam.core.SimpleSlot'
+  ],
+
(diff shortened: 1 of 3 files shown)


src/foam/u2/view/UnstyledChooseNView.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+    },
+    {
+      name: 'data',
+      class: 'Map',
+      value: {},
+      factory: function () {
+        return {};
+      },
+      adapt: function (_, n) {
```
jlhughes: This sets the first 'n' to true/on - for initial display?

KernelDeimos: That's there to allow you to pass an array but... actually right now setting `data` doesn't update the checkboxes so there's no point in adapting this. (originally I had a getter convert the set into an array but the getter conflicted with the `$remove` and `$set` property methods)

I guess my options are either remove this or allow checkboxes to be set based on a provided value for `data`. I think I should do the second thing but that requires some clever thinking

jlhughes (approved):

PR #3734, opened 2020-07-29
refine src/tgt id of cap junction to show cap name

noodlemoodle:
refine src/target ids of capability junctions so that capability name is displayed in tableview instead of id for easier viewing

### DIFF
--- a/src/foam/nanos/crunch/CapabilityCapabilityJunctionRefine.js
+++ b/src/foam/nanos/crunch/CapabilityCapabilityJunctionRefine.js
@@ -1,0 +1,44 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.crunch',
+  name: 'CapabilityCapabilityJunctionRefine',
+  refines: 'foam.nanos.crunch.CapabilityCapabilityJunction',
+
+  documentation: `
+    Refine capabilitycapabilityjunction to add tablecellformatters for source and target id
+  `,
+
+  properties: [
+    {
+      class: 'Reference',
+      of: 'foam.nanos.crunch.Capability',
+      name: 'sourceId',
+      tableCellFormatter: function(value, obj, axiom) {
+        this.__subSubContext__.capabilityDAO
+          .find(value)
+          .then((capability) => this.add(capability.name))
+          .catch((error) => {
+            this.add(value);
+          });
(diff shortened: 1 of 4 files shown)


Nauna (approved):

src/foam/nanos/crunch/CapabilityCapabilityJunctionRefine.js
```diff
@@ -0,0 +1,44 @@
...
+  documentation: `
+    Refine capabilitycapabilityjunction to add tablecellformatters for source and target id
+  `,
+
+  properties: [
+    {
+      class: 'Reference',
+      of: 'foam.nanos.crunch.Capability',
+      name: 'sourceId',
+      tableCellFormatter: function(value, obj, axiom) {
```
Nauna, line 21: Just curious, did you try to add these tableCellFormatters to target and source property of the relationship?

src/foam/nanos/crunch/CapabilityCapabilityJunctionRefine.js
```diff
@@ -0,0 +1,44 @@
...
+
+  documentation: `
+    Refine capabilitycapabilityjunction to add tablecellformatters for source and target id
+  `,
+
+  properties: [
+    {
+      class: 'Reference',
+      of: 'foam.nanos.crunch.Capability',
+      name: 'sourceId',
```
jlhughes, line 20: With regard to names with references, we've standardized on non 'id' in the name. 
So 
name: 'source' as the findSource(x) returns the object, not the id. 
Similar with targetId. 
as long as you don't have a lot of code dependent on this.

noodlemoodle: ok sounds good, but these ones are generated by the relationship, should those be migrated to source/target?

jlhughes: got it, ok, leave them.

jlhughes (approved):

src/foam/nanos/crunch/CapabilityCapabilityJunctionRefine.js
noodlemoodle: ya I tried that at first but i think those are actually for defining property properties of the property on the source/target models, instead of being related to the source/target properties of the junction
so an example of when you'd use source/targetproperty is : user's daosummaryview should show their capabilities but capability's daosummaryview should not show its users. so we set something like `sourceProperty:{visibility:'RO'}` to make the capabilities show up in the user model's daosummaryview

PR #3736, opened 2020-07-29
fix search

olhabn:
@kgrgreer @jlhughes @hchoura 
please review


olhabn:
I think this pr pretty much fixes bugs mentioned by @hchoura as TreeView works as expected because tree view won't display same results as table view:
let's use 'url' example for Menus:
table view will display all the items that contain 'url',
tree view will display first-level parents that contain 'url' because we search for first-level menus and the load children for it so if child contains 'url' but parent doesn't than it won't be displayed. please correct me if i'm wrong @kgrgreer

PR #3743, opened 2020-07-30
Added availability predicates to capabilities

zahnno:
Capabilities will be returned depending on the predicate associated to the availabilityPredicate.

### DIFF
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -170,10 +170,20 @@ foam.CLASS({
     {
       class: 'foam.mlang.predicate.PredicateProperty',
       name: 'interceptIf',
+      networkTransient: true,
       javaFactory: `
       return foam.mlang.MLang.TRUE;
       `,
       documentation: 'condition under which the permissions that may be intercepted by this capability will be intercepted.'
+    },
+    {
+      class: 'foam.mlang.predicate.PredicateProperty',
+      name: 'availabilityPredicate',
+      networkTransient: true,
+      javaFactory: `
+      return foam.mlang.MLang.TRUE;
+      `,
+      documentation: 'Predicate used to omit or include capabilities from capabilityDAO'
     },
     {
       name: 'lastModified',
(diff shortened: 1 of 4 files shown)


Nauna:
As far as I am aware a few of us are using/planning to use the capabilityCategory predicate for visibility -visibilityCondition. Think this is redundant.

note: visibilityCondition, also plays a role in categorizing capabilities in predicates.

zahnno:
CapabilityCategory visibility predicates are great and I think they have a lot of uses however they provide an additional layer of complexity to solving underlying exposure issues. The suggestion is a work around to a problem that I can simplify to - **"What capabilities can users see and why"** - from both the client and server. Having this be dictated by a combination of capability and category setup along with GUI changes or awareness of these categories - I think - is a bit more tedious then it has to be and can be solved with a simple predicate awareness applied to a decorated layer on the capabilityDAO.

The availabilityPredicate solves configuration issues on capabilities where all capabilities are exposed to a user. A select on the capabilityDAO from any user on the system will return all capabilities on the system - currently - which I believe to be over exposure. Why show all capabilities to a user if they cannot ever satisfy them? We see this as being the case where a user with certain attributes won't require to fill out the same capability as another to be granted accessibility to the system. Also - the decorated layer placed on the capability allows availability to be dictated by permissions if needed. So if a user doesn't satisfy the availabilityPredicate we can still permission them to access the capability through a permission granted to them. This implementation tunes the solution towards a configuration solution rather than a development work around.

A capabilityCategory exclusion would be dictated by a query or a specific use case involving the capabilityCategory, while the availabilityPredicate provides availability support based on context configuration.

Views dictated by all entries of the capabilityDAO will show all capabilities with out additional client code. (Which will need to be accounted for in every view we present all capabilities that require some filtering.)

I appreciate the suggested work around and think the capabilityCategory is a useful system feature and has extensible use cases however I believe it adds complexity to solving the over exposure of capabilities - Also utilizing the availabilityPredicate is a configuration placed on a capability - developers can refrain from using it if they choose. It doesn't effect the system negatively and only adds additional support.

jlhughes (approved):

PR #3746, opened 2020-08-04
[NP-1777] Add crunch service

KernelDeimos:
This adds an early implementation of CrunchService, where `getGrantPath` will return all capabilities rather than just those that do not already have ACTIVE status. Making the change to only report necessary capabilities will require another change (I'll PR two options I came up with tomorrow) to avoid duplicating logic already present in CapabilityAuthService.


jlhughes (approved):

KernelDeimos:
Merged in #3753

PR #3747, opened 2020-08-04
Np 1726/refactor unstyled table view changes fix selected columns

olhabn:
FObject or Reference property displayed as 'Selected' even if no child property is selected

@jlhughes @kgrgreer please review

### DIFF
--- a/src/foam/u2/view/ColumnConfigView.js
+++ b/src/foam/u2/view/ColumnConfigView.js
@@ -704,15 +704,7 @@ foam.CLASS({
       this.expanded = false;
     },
     function returnSelectedProps() {
-      if ( ! this.hasSubProperties ) {
-        if ( this.level === 0 ) {
-          if ( foam.Array.isInstance(this.rootProperty) )
-            return [[this.rootProperty[0]]];
-          else
-            return [this.rootProperty];
-        } else 
-          return [this.rootProperty[0]];
-      } else {
+      if ( this.hasSubProperties ) {
         var arr = [];
         for ( var i = 0 ; i < this.subColumnSelectConfig.length ; i++ ) {
           if ( this.subColumnSelectConfig[i].isPropertySelected ) {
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/ColumnConfigView.js
```diff
@@ -726,8 +713,16 @@ foam.CLASS({
...
           }
         }
-        return arr;
+        if ( arr && arr.length > 0 )
+          return arr;
       }
+      if ( this.level === 0 ) {
+        if ( foam.Array.isInstance(this.rootProperty) )
+          return [[this.rootProperty[0]]];
+        else
```
kgrgreer: No need for 'else' if you've already returned inn the 'if'.

src/foam/u2/view/ColumnConfigView.js
```diff
@@ -726,8 +713,16 @@ foam.CLASS({
...
+          return [this.rootProperty];
+      } else
```
kgrgreer: No need for 'else' if you've already returned inn the 'if'.

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -293,6 +281,7 @@ foam.CLASS({
 
             return this.E().
               addClass(view.myClass('tr')).
+              style({ 'min-width': this.tableWidth_$ }).
```
kgrgreer, line 284: Does tableWidth_ ever change its value?

olhabn: it should be on column selection change as it's the sum of the columns tableWidth

src/foam/u2/view/ColumnConfigView.js
olhabn: fixed

src/foam/u2/view/ColumnConfigView.js
olhabn: fixed

PR #3749, opened 2020-08-04
[ME#3753] [NP-1777] Capability.getJunction(x)

KernelDeimos:
PR depends on #3746
PR is mutually exclusive to #3753 

This is one solution for determining a capabilities status based on context without repeating the logic of determining which user (i.e. `subject.realUser` vs `subject.user`) and which junction type (i.e. if it's an AgentCapabilityJunction)

### DIFF
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -17,7 +17,11 @@ foam.CLASS({
     'foam.dao.ArraySink',
     'foam.dao.DAO',
     'foam.dao.Sink',
+    'foam.mlang.predicate.Predicate',
     'foam.mlang.sink.Count',
+    'foam.nanos.auth.Subject',
+    'foam.nanos.auth.User',
+    'foam.nanos.logger.Logger',
     'java.util.Date',
     'java.util.List',
     'static foam.mlang.MLang.*'
@@ -306,6 +310,73 @@ foam.CLASS({
       Date capabilityExpiry = getExpiry();
 
       return today.after(capabilityExpiry);
+      `
+    },
+    {
+      name: 'getJunction',
+      type: 'foam.nanos.crunch.UserCapabilityJunction',
+      documentation: `
+        answers the question "does the context have this capability granted?"
+      `,
+      args: [
+        { name: 'x', type: 'Context' }
+      ],
+      javaCode: `
+        User user = ((Subject) x.get("subject")).getUser();
+        User realUser = ((Subject) x.get("subject")).getRealUser();
+
+        Predicate acjPredicate = INSTANCE_OF(AgentCapabilityJunction.class);
+        Predicate targetPredicate = EQ(UserCapabilityJunction.TARGET_ID, getId());
+        try {
+          DAO capabilityDAO = ( x.get("localCapabilityDAO") == null ) ? (DAO) x.get("capabilityDAO") : (DAO) x.get("localCapabilityDAO");
+          DAO userCapabilityJunctionDAO = (DAO) x.get("userCapabilityJunctionDAO");
+          
+          {
+            // Check if a ucj implies the subject.user has this permission
+            Predicate userPredicate = AND(
+              NOT(INSTANCE_OF(AgentCapabilityJunction.class)),
+              EQ(UserCapabilityJunction.SOURCE_ID, user.getId())
+            );
+            UserCapabilityJunction ucj = (UserCapabilityJunction)
+              userCapabilityJunctionDAO.find(AND(userPredicate,targetPredicate));
+            if ( ucj != null ) {
+              return ucj;
+            }
+          }
+
+          // Check if a ucj implies the subject.realUser has this permission
+          if ( realUser != null ) {
+            Predicate userPredicate = AND(
+              NOT(INSTANCE_OF(AgentCapabilityJunction.class)),
+              EQ(UserCapabilityJunction.SOURCE_ID, realUser.getId())
+            );
+            UserCapabilityJunction ucj = (UserCapabilityJunction)
+              userCapabilityJunctionDAO.find(AND(userPredicate,targetPredicate));
+            if ( ucj != null ) {
+              return ucj;
+            }
+          }
+
+          // Check if a ucj implies the subject.realUser has this permission in relation to the user
+          if ( realUser != null ) {
+            Predicate userPredicate = AND(
+              INSTANCE_OF(AgentCapabilityJunction.class),
+              EQ(UserCapabilityJunction.SOURCE_ID, realUser.getId()),
+              EQ(AgentCapabilityJunction.EFFECTIVE_USER, user.getId())
+            );
+            UserCapabilityJunction ucj = (UserCapabilityJunction)
+              userCapabilityJunctionDAO.find(AND(userPredicate,targetPredicate));
+            if ( ucj != null ) {
+              return ucj;
+            }
+          }
+
+        } catch ( Exception e ) {
+          Logger logger = (Logger) x.get("logger");
+          logger.error("getJunction", getId());
+        }
+
+        return null;
       `
     }
   ]
--- a/src/foam/nanos/crunch/JavaCrunchService.java
+++ b/src/foam/nanos/crunch/JavaCrunchService.java
@@ -3,6 +3,7 @@ package foam.nanos.crunch;
 import foam.core.X;
 import foam.dao.ArraySink;
 import foam.dao.DAO;
+import foam.nanos.logger.Logger;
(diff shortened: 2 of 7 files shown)


KernelDeimos:
Added `[DNM]` - going to put up another PR soon with a solution of adding `getJunction(x, capabilityId)` to `crunchService` which seems infinitely more sensible to me at the moment

Nauna (commented):
If we migrate agentCapabilityJunction to include an association to not only include a user but also an object ex bankAccount then do you still see getJunction working?

src/foam/nanos/crunch/Capability.js
```diff
@@ -307,6 +311,73 @@ foam.CLASS({
...
+        { name: 'x', type: 'Context' }
+      ],
+      javaCode: `
+        User user = ((Subject) x.get("subject")).getUser();
+        User realUser = ((Subject) x.get("subject")).getRealUser();
+
+        Predicate acjPredicate = INSTANCE_OF(AgentCapabilityJunction.class);
+        Predicate targetPredicate = EQ(UserCapabilityJunction.TARGET_ID, getId());
+        try {
+          DAO capabilityDAO = ( x.get("localCapabilityDAO") == null ) ? (DAO) x.get("capabilityDAO") : (DAO) x.get("localCapabilityDAO");
```
Nauna, line 331: I know this is repeated from a separate location but this doesn't make the most sense. 
If localCapabilityDAO is null, which it should never be, and should rightfully throw a null pointer exception but also capabilityDAO is built on top of local - thus it too would be null.

Nauna, line 331:
```suggestion
          DAO capabilityDAO = (DAO) x.get("localCapabilityDAO");
```

KernelDeimos:
> If we migrate agentCapabilityJunction to include an association to not only include a user but also an object ex bankAccount then do you still see getJunction working?

The answer to that question would probably be revealed by the answer to this question: how will those per-object associations affect the logic in CapabilityAuthService?

src/foam/nanos/crunch/Capability.js
```diff
@@ -307,6 +311,73 @@ foam.CLASS({
...
+          DAO userCapabilityJunctionDAO = (DAO) x.get("userCapabilityJunctionDAO");
+          
+          {
```
Nauna, line 334:
```suggestion

```

src/foam/nanos/crunch/Capability.js
```diff
@@ -307,6 +311,73 @@ foam.CLASS({
...
+            Predicate userPredicate = AND(
+              NOT(INSTANCE_OF(AgentCapabilityJunction.class)),
+              EQ(UserCapabilityJunction.SOURCE_ID, user.getId())
+            );
+            UserCapabilityJunction ucj = (UserCapabilityJunction)
+              userCapabilityJunctionDAO.find(AND(userPredicate,targetPredicate));
+            if ( ucj != null ) {
+              return ucj;
+            }
+          }
```
Nauna, line 345:
```suggestion

```

KernelDeimos:
btw I think #3753 is a better solution since it could hypothetically call a `Capability.getJunction` (like in this PR) if we ever need it

src/foam/nanos/crunch/Capability.js
Nauna, line 331: could even do this to simplify the code a bit - by removing targetPredicate
```suggestion
          DAO capabilityDAO = (DAO) x.get("localCapabilityDAO").where(EQ(UserCapabilityJunction.TARGET_ID, getId()));
```

src/foam/nanos/crunch/Capability.js
```diff
@@ -307,6 +311,73 @@ foam.CLASS({
...
+
+          // Check if a ucj implies the subject.realUser has this permission
+          if ( realUser != null ) {
+            Predicate userPredicate = AND(
```
Nauna, line 349: I wonder:
If we have a payments cap which could be used to grant a user payment access or perhaps invoice access, then
If a business has this caps ucj that allows for payments, and user has this ucj - would we be able to rely on this function?

How likely is this type of capability, such that could be for user and business? I would argue we have none right now, but considering a lot of the recent work has gone on around separating user and business to allow different functionality ... it may become increasingly frequent.

I would argue that since this is used in the crunchService to put this function there. 

I'm thinking that perhaps we should just be checking the subject to see if they have a ucj. So context subject is checked to see if realUser and user exist and if so check for acj or ucj on business ... then if subject only has realUser - just check realUser

src/foam/nanos/crunch/Capability.js
```diff
@@ -307,6 +311,73 @@ foam.CLASS({
...
+        answers the question "does the context have this capability granted?"
+      `,
+      args: [
+        { name: 'x', type: 'Context' }
+      ],
+      javaCode: `
+        User user = ((Subject) x.get("subject")).getUser();
+        User realUser = ((Subject) x.get("subject")).getRealUser();
+
+        Predicate acjPredicate = INSTANCE_OF(AgentCapabilityJunction.class);
```
Nauna, line 328: never used

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+import static foam.mlang.MLang.*;
+
+public class JavaCrunchService implements CrunchService {
```
Nauna, line 16: JavaCrunchService? hmmm Java seems unrelated

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,82 @@
...
+  public List getGrantPath(X x, String rootId) {
+    Logger logger = (Logger) x.get("logger");
+    logger.info("Handling CrunchService grantPath request: " + rootId);
```
Nauna, line 19: I'm thinking this logger is excessive. If we log every start for every user - this seems like it would clutter the logger dao and make finding anything relevant difficult.

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,82 @@
...
+
+    DAO prerequisiteDAO = (DAO) x.get("prerequisiteCapabilityJunctionDAO");
+    DAO capabilityDAO = (DAO) x.get("capabilityDAO");
+    // TODO:
+    // DAO junctionDAO
```
Nauna, line 24: can you be more specific on this and the below todo - please - has me scratching my head

Nauna:
> btw I think #3753 is a better solution since it could hypothetically call a `Capability.getJunction` (like in this PR) if we ever need it

Seriously :(
Just spent the last hr reviewing this.

plz close and https://github.com/foam-framework/foam2/pull/3746 - pr still exists if you want to talk about it

but not complete loss - some comments still relevant.

src/foam/nanos/crunch/Capability.js
KernelDeimos: targetPredicate is re-used

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: How? This is the Java-side CrunchService

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: Agreed, will fix in other PR

src/foam/nanos/crunch/Capability.js
KernelDeimos: Moved to CrunchService in other PR, re-reading the rest (logic is copied from CapabilityAuthService, which a later PR could simplify by using this method inside CapabilityAuthService)

KernelDeimos:
I kept both open in case anyone had a good argument to prefer this PR. Fortunately both PRs are nearly identical since it's the location of the implementation that differs. I'll address this comments in a commit to that PR

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: oops, todo comment was for me and no longer applies

src/foam/nanos/crunch/Capability.js
KernelDeimos: Scope limiter here makes the variable declarations consistent (also just removing the scope limiter without changing those will result in a compiler error)

PR #3753, opened 2020-08-04
[ME#3749] [NP-1777] CrunchService getJunction(x, capabilityId)

KernelDeimos:
Mutually exclusive to #3749. I think this is a better alternative but I kept the other PR open for discussion.


Nauna (commented):
So I've left comments on https://github.com/foam-framework/foam2/pull/3749 not realizing that this was the pr that should be reviewed. 

Most comment should be applied here - except in regards to the association comments - we can go without that concept for now and work on it all together, when that task starts.

src/foam/nanos/crunch/CrunchService.js
```diff
@@ -0,0 +1,37 @@
+foam.INTERFACE({
+  package: 'foam.nanos.crunch',
+  name: 'CrunchService',
+
```
kgrgreer, line 14: Add documentation:

KernelDeimos:
Fixed merge conflicts (sorry about delay; green checkmark made me think everything was fine here)

src/foam/nanos/crunch/CrunchService.js
```diff
@@ -0,0 +1,49 @@
...
+  documentation: `
+    CrunchService provides common logic used by the client and other CRUNCH
+    services.
+  `,
+
+  methods: [
+    {
+      name: 'getGrantPath',
+      documentation: `
+        getGrantPath provides a an array of capability objects representing
```
zahnno:
```suggestion
        getGrantPath provides an array of capability objects representing
```

src/foam/nanos/crunch/CrunchService.js
```diff
@@ -0,0 +1,49 @@
...
+      `,
+      async: true,
+      type: 'List',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'sourceId',
```
zahnno, line 30: Would a better suited name be capabilityId, since we're getting the prerequisites of a capability?

src/foam/nanos/crunch/CrunchService.js
```diff
@@ -0,0 +1,49 @@
...
+      ]
+    },
+    {
+      name: 'getJunction',
+      documentation: `
+        getJunction provides the correct UserCapabilityJunction based on the
+        context provided.
+      `,
+      async: true,
+      type: 'UserCapabilityJunction',
```
zahnno, line 42: I believe this needs to change to a list considering the GUID changes to UCJ's.

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,140 @@
...
+public class JavaCrunchService implements CrunchService {
+  public List getGrantPath(X x, String rootId) {
+    Logger logger = (Logger) x.get("logger");
+
+    DAO prerequisiteDAO = (DAO) x.get("prerequisiteCapabilityJunctionDAO");
+    DAO capabilityDAO = (DAO) x.get("capabilityDAO");
+
+    // Lookup for indices of previously observed capabilities
+    Map<String,Integer> alreadyListed = new HashMap<String,Integer>();
+    // Return list (list is reversed at the end to put prerequisites first)
```
zahnno: Why the ordering? - maybe add to documentation explaining the reasoning for it.

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,140 @@
...
+    List grantPath = new ArrayList<>();
+
+    Queue<String> nextSources = new ArrayDeque<String>();
+    nextSources.add(rootId);
+
+    // Doing this instead of "this" could prevent unexpected behaviour
+    // in the incident CrunchService's getJunction method is decorated
+    CrunchService crunchService = (CrunchService) x.get("crunchService");
```
zahnno: Why grab the service from the context instead of referencing its self? If the getJunction method was decorated on the crunchService - crunchService will use the decorated method.

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,140 @@
...
+        );
+        UserCapabilityJunction ucj = (UserCapabilityJunction)
+          userCapabilityJunctionDAO.find(AND(userPredicate,targetPredicate));
+        if ( ucj != null ) {
+          return ucj;
+        }
+      }
+
+      // Check if a ucj implies the subject.realUser has this permission
+      {
```
zahnno: Where's the if condition for this block?

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+      }
+
+      // Check if a ucj implies the subject.realUser has this permission in relation to the user
+      {
+        Predicate userPredicate = AND(
+          acjPredicate,
+          EQ(UserCapabilityJunction.SOURCE_ID, realUser.getId()),
+          EQ(AgentCapabilityJunction.EFFECTIVE_USER, user.getId())
+        );
```
zahnno: Your conditions could build a predicate and then you could call the ucj DAO once at the end. You could also build out the predicate through ternary expressions if you really wanna shorten it.

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,140 @@
...
+    // in the incident CrunchService's getJunction method is decorated
+    CrunchService crunchService = (CrunchService) x.get("crunchService");
+
+    while ( nextSources.size() > 0 ) {
+      String sourceId = nextSources.poll();
+
+      UserCapabilityJunction ucj = crunchService.getJunction(x, sourceId);
+      if ( ucj != null && ucj.getStatus() == CapabilityJunctionStatus.GRANTED ) {
+        continue;
+      }
```
zahnno: I thought this method was returning a capabilities requirements - why is the ucj referenced here?

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,140 @@
...
+        // Remove previously stored index of capability
+        alreadyListed.remove(sourceId);
+
+        // Shift remembered indexes, now that grantList has been shifted
+        Map<String,Integer> newAlreadyListed = new HashMap<String,Integer>();
+        for ( Map.Entry<String,Integer> entry : alreadyListed.entrySet() ) {
+          int newIndex = entry.getValue();
+          if ( newIndex > previousIndex ) newIndex--;
+        }
+        alreadyListed = newAlreadyListed;
```
zahnno: Assign alreadylisted to new hashmap? is this intended?

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,140 @@
...
+        .where(EQ(CapabilityCapabilityJunction.SOURCE_ID, sourceId))
+        .select(new ArraySink()) ).getArray();
+      for ( Object prereqObj : prereqs ) {
+        CapabilityCapabilityJunction prereq =
+          (CapabilityCapabilityJunction) prereqObj;
+        nextSources.add(prereq.getTargetId());
+      }
+    }
+
+    Collections.reverse(grantPath);
```
zahnno: Is an arrayList of capabilities sufficient here? I think a tree structure would be better suited.

src/foam/nanos/crunch/JavaCrunchService.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+import static foam.mlang.MLang.*;
+
+public class JavaCrunchService implements CrunchService {
```
zahnno: JavaCrunchService seems a bit odd lol. 

It's up to you but I would have called the implemented class "CrunchServiceInterface" and this class "CrunchService" - Since it is the logic of the "CrunchService".

src/foam/nanos/crunch/CrunchService.js
KernelDeimos: That's why I went for `sourceId` since the direction is more clear, although now that you mention it `capabilitySourceId` may be more clear.

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: What if we added a NodeCrunchService, or an open source contributor wanted to add PythonCrunchService?

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: I'm planning on making a `.flow` document to explain how to use CRUNCH, maybe that would be best suited there. The prerequisites are returned first because `getGrantPath` returns capabilities in the same order which a user is expected to acquire them.

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: > If the getJunction method was decorated on the crunchService - crunchService will use the decorated method.

I don't think this is true. If this CrunchService was a delegate of another CrunchService (extends ProxyCrunchService) that overrides getJunction, calling `this.getJunction` in this CrunchService will not call the decorated method.

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: Granted capabilities are not included in the return value of `getGrantPath`

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: This is intended... but I think I missed a step in that for loop

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: I lost that debate several times over, but I eventually realized it makes sense as long as `Or` capabilities can be nested lists to help the wizard logic.

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: This is a scope limiter. IMO they're really useful for making variable declarations consistent and making code easier to read. Some people don't like them because of their obscurity, hence their obscurity.

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: Sick! Definitely doing that, thanks!

src/foam/nanos/crunch/CrunchService.js
KernelDeimos: Oh shoot, okay thanks for letting me know

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: resolved

src/foam/nanos/crunch/CrunchService.js
Nauna: If you are pulling the context for ucj's that the business has with a bank account - then it would be a list.

Good catch @zahnno

KernelDeimos: @Nauna We'll need to discuss this. Right now I don't know how the wizard is supposed to handle capabilities like this (checking the UCJ status and rendering the UCJ data), or how other code checking UCJ status is meant to do so.

Nauna: - Hmmm - It is a good point as well.

#### So here are my thoughts:
If list length == 1 or 0, all good. However if list.length > 1 then we need to know the reference for look up. 
What we can do is leave the return type to be an individual ucj, but adjust the logic here a little. 

#### Consider this: 
adding an input variable to getJunction - be reference id - then if list.length > 1, and reference(input variable) is empty then throw an exception for missing reference.
This can allow for some distinction in calling functions, and give us time to implement the functionality of these associations.

@zahnno @jlhughes @kgrgreer your thoughts?

KernelDeimos: How would this affect `getGrantPath`? While UCJs can be associated with references to other objects like this, getting a grant path with __0 or 1__ reference id seems like an arbitrary limitation; it may be desired to get the grant path in consideration of two different references for example.

KernelDeimos: I have an idea where getGrantPath is passed something like:
```
[
  ["bankAccount", "asdf"],
  ["favouriteColour", "red"]
]
```

also indicating references are for specific capabilities may be useful

```
[
  ["capability-number-1", "favouriteColour", "red"],
  ["capability-number-2", "favouriteColour", "blue"]
]
```

KernelDeimos: To clarify, I'm thinking of the two above as not mutually exclusive, so for example you could specify:
```
[
  ["bankAccount", "asdf"],
  ["capability-number-1", "favouriteColour", "red"],
  ["capability-number-2", "favouriteColour", "blue"]
]
```

KernelDeimos: IMO we should merge this now and add this functionality as needed

src/foam/nanos/crunch/JavaCrunchService.java
jlhughes: CrunchService is fine for the interface. 
and we can change the server side (java only) implementation name later if we like/need. 
so leave for now.

src/foam/nanos/crunch/JavaCrunchService.java
zahnno: Then which one do you want? 
If you want to call this classes getJunction then call getJunction(x, sourceId). 
If you want to use the decorated getJunction method call x.get("crunchService").getJunction();

src/foam/nanos/crunch/JavaCrunchService.java
zahnno: @KernelDeimos We have yet to call a service based on the language it was written in so for consistency this stands out.

src/foam/nanos/crunch/JavaCrunchService.java
zahnno: Your comment is explaining what the code is, not why. Some one reviewing your code could see what this comment describes by looking at the code. It's helpful to have documentation explaining why the code is doing what it is doing.

src/foam/nanos/crunch/JavaCrunchService.java
zahnno: Ahh - I guess for the immediate use case this is what we need so ok.

src/foam/nanos/crunch/CrunchService.js
zahnno: For @Nauna  solution we'd also have to pass in a DAOkey since we'd want to have the id reference any object on the system. We could have both a getJunction and getJunctions method

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: Oh I misunderstood what you meant. I figured that comment would be helpful as a warning since the reversal is easy to miss. I'll update it.

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: I was thinking that from the API user's perspective getGrantPath should always be consistent with getJunction, so if the getJunction method was overridden by a decorator then the base CrunchService should use that.

Although I'm willing to change it out for `this` if you think that's better.

src/foam/nanos/crunch/JavaCrunchService.java
KernelDeimos: Renamed it to `ServerCrunchService` since I realized this name can be overloaded for other languages.

src/foam/nanos/crunch/JavaCrunchService.java
zahnno: It's up to you. Your comment says that you don't want to use a decorated getJunction - "Doing this instead of "this" could prevent unexpected behaviour in the incident CrunchService's getJunction method is decorated" From what I understand, you're trying to avoid using decorated getJunction methods.

KernelDeimos: oh, no exactly the opposite. If somebody overrides getJunction I want that change to be "noticed" by the getGrantPath method

Nauna (approved):

Nauna:
There are good ideas here, and these should be taken in when going through and changing the scope of agentCapabilityJunction objects - to the association of FObjects.

I'd suggest merge now, refactor later - keeping this pr as system notes when designing the above mentioned changes.

PR #3762, opened 2020-08-05
[NP-1796] Allow wizard-invoking capabilities to configure the wizard

KernelDeimos:
Reviewers beware: this PR contains #3753 as it's a dependancy. Merge IFF #3753 is merged.


src/foam/u2/crunch/CapabilityRequirementView.js
```diff
@@ -161,7 +169,8 @@ foam.CLASS({
     {
       name: 'getStarted',
       code: function(x) {
-        x.crunchController.generateAndDisplayWizard(this.functionData);
+        console.log('uhh', this.capability)
```
zahnno, line 172: console log?

src/foam/nanos/crunch/Capability.js
```diff
@@ -223,6 +223,14 @@ foam.CLASS({
       javaFactory: `
         return foam.nanos.crunch.AssociatedEntity.USER;
       `
+    },
+    {
+      name: 'wizardConfig',
+      class: 'FObjectProperty',
+      of: 'foam.u2.wizard.StepWizardConfig',
+      documentation: `
+        Allow this capability to configure the wizard when it is the root capability.
```
zahnno, line 235: Does this not need a default value or factory? I was also thinking this would be a generic class reference string, which could then be looked up and utilized by the stepWizardletController.

src/foam/u2/crunch/CrunchController.js
```diff
@@ -135,11 +135,15 @@ foam.CLASS({
...
+    function generateAndDisplayWizard(capability, capabilitiesSections) {
+      var maybeConfig = capability.wizardConfig
+        ? { config: capability.wizardConfig } : {};
+
       // called in CapabilityRequirementView
       return ctrl.add(this.Popup.create({ closeable: false }).tag({
         class: 'foam.u2.wizard.StepWizardletView',
         data: foam.u2.wizard.StepWizardletController.create({
+          ...maybeConfig,
           wizardlets: capabilitiesSections.wizCaps
```
zahnno, line 147: Where in the stepwizardletController is the onclose function being called and passing in the context and ucj?

I was thinking that the config would be created within the stepWizardletController (since it would be referenced on the capability) then from there the context and ucj passed along. 

Is this PR completed or am I missing something?

src/foam/nanos/crunch/Capability.js
KernelDeimos: I can be set as `{ class: 'foam.u2.wizard.RefreshingStepWizardConfig' }` which also allows setting other properties

src/foam/u2/crunch/CrunchController.js
KernelDeimos: Not completed. Adding `[DNM]` because I'll need #3753 to be merged first to do this properly (otherwise passing the ucj would be non-trivial). As for why this is in CrunchController I can explain on Zoom tomorrow - the short version is StepWizardController is independant of CRUNCH

KernelDeimos:
PR force-pushed and is now based on #3753

KernelDeimos:
Same functionality merged in #3802

PR #3771, opened 2020-08-07
dnm - Clear ucj data on expiry

noodlemoodle:
todo https://nanopay.atlassian.net/browse/NP-1806


src/foam/nanos/audit/AuditDAO.js
```diff
@@ -63,9 +63,13 @@ foam.CLASS({
         Logger  logger   = (Logger) x.get("logger");
         FObject current  = this.find_(x, obj);
         Object  objectId = obj.getProperty("id");
-    
-        logger.info("CHANGE", objectId, user.getId(), formatMessage(current, obj));
-    
+
+        if ( current == null ) {
```
kgrgreer, line 67: This seems like itt should be in a separate PR. No need to split now. Just saying for future.

PR #3773, opened 2020-08-07
Do not set context to null in XLocator

nanoscott:


kgrgreer:
I think this is a bug and can let someone's credentials be gained by someone else. What problem are you trying to fix? Rules not have the XLocator set? Didn't I already fix that problem earlier this week?

PR #3774, opened 2020-08-08
Fix rule authorization

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-1311

## Issue
- Ablii fee leaks into Treviso transaction due to failed spid enforcement in RulerDAO

## Changes
- Fix rule authorization
- Move serviceProviderAware after ruler so that ruler can enforce spid using ServiceProviderAwareSupport action command

PR #3777, opened 2020-08-10
Fix rule authorization

lchanmann:
## Ref
- https://github.com/foam-framework/foam2/pull/3774#issuecomment-670824666

## Changes
- Let ServiceProviderAwareDAO.put_ handle rule create and update
- Fix ServiceProviderAwareDAO.select_ to return objects when user has "spid.read.{obj.spid}" permission


nanoscott:
LGTM .. thanks for update it @lchanmann

PR #3778, opened 2020-08-10
Cpf 4391/report generation using google sheets api

olhabn:
ticket https://nanopay.atlassian.net/browse/CPF-4391
nanopay https://github.com/nanoPayinc/NANOPAY/pull/10384

@kgrgreer @jlhughes please review


src/foam/nanos/export/PDFGoogleSheetsExportDriver.js
```diff
@@ -31,44 +31,75 @@ foam.CLASS({
     {
       class: 'String',
       name: 'title'
```
olhabn, line 33: as most all of the code is the same for GoogleSheetsExportDriver.js and PDFGoogleSheetsExportDriver.js will create some kind of base class a little bit later

src/foam/nanos/export/GoogleSheetsExportDriver.js
```diff
@@ -57,18 +79,30 @@ foam.CLASS({
...
+        var expr1 = foam.mlang.Expressions.create();
+        var template = await X.reportTemplateDAO.find(expr1.EQ(foam.nanos.export.report.Template.ID, this.template));
+        propNames = template && template.columnNames && template.columnNames.length > 0 ? template.columnNames : X.filteredTableColumns ? X.filteredTableColumns : this.outputter.getAllPropertyNames(dao.of);
+      }
+       
       propNames = columnConfig.filterExportedProps(dao.of, propNames);
 
       var metadata = await self.outputter.getColumnMethadata(X, dao.of, propNames);
 
+      // to backend
```
olhabn, line 95: I think that it would be better to do projection on backend so we don't send it back and forth as we need this data only on backend, but we do formatting on front-end so I will add java version of unitPropValueToString.

olhabn: or should I just override toString for those classes?

src/foam/nanos/auth/Address.js
```diff
@@ -272,18 +272,18 @@ foam.CLASS({
...
-      sb.append(", ");
-      sb.append(this.getCountryId());
-      sb.append(", ");
-      sb.append(this.getPostalCode());
-      String rtn = sb.toString();
-      return rtn.equals(", , , , ") ? "" : rtn;
+        StringBuilder sb = new StringBuilder();
+        sb.append(getShortAddress());
+        sb.append(", ");
+        sb.append(this.getCity());
```
kgrgreer, line 587: Why do you need the 'this.'?

src/foam/nanos/auth/Address.js
```diff
@@ -307,7 +307,7 @@ foam.CLASS({
       javaCode: `
       StringBuilder sb = new StringBuilder();
       if ( this.getStructured() ) {
```
kgrgreer: Why using 'this.' in Java?

src/foam/nanos/export/report/Template.js
```diff
@@ -0,0 +1,24 @@
+foam.CLASS({
```
kgrgreer, line 7: Copyright header.

src/foam/nanos/export/report/Template.js
```diff
@@ -0,0 +1,24 @@
...
+  package: 'foam.nanos.export.report',
+  name: 'Template',
+  properties: [
+    {
+      name: 'id',
+      class: 'String',
+      documetation: 'file id',
```
kgrgreer: No trailing commas.

src/foam/nanos/export/report/Template.js
olhabn: fixed

src/foam/nanos/auth/Address.js
olhabn: fixed

src/foam/nanos/auth/Address.js
olhabn: fixed

src/foam/nanos/export/report/Template.js
olhabn: fixed

PR #3782, opened 2020-08-11
Pm alarm

KolombetM:
@jlhughes


src/foam/nanos/alarming/AlarmConfigName.js
```diff
@@ -0,0 +1,64 @@
+foam.CLASS({
```
jlhughes, line 5: License header please.

jlhughes (approved):

PR #3783, opened 2020-08-11
[CPF-504] set admin to private

jinnnn9:
addresses: https://nanopay.atlassian.net/browse/CPF-504
nanopay side pr: https://github.com/nanoPayinc/NANOPAY/pull/10367


jlhughes:
isPublic is a nanopay side refinement, so this update to the foam 'admin' must occur on the nanopay side. 
The nanopay users.jrl can set this to isPublic.

nanoscott (approved):

nanoscott (changes requested):
looks like you need to move this to the NP side instead of within FOAM, as per Joel's comments

jinn9:
@jlhughes Should I leave or remove 'admin' on the foam side?

PR #3784, opened 2020-08-11
MultiChoiceView + CardSelectView

tharmaman:
Wraps a tag that represents multiple choices.

The choices are in [value, label, isSelected, choiceMode ] quartets.

However the client can simply pass in [value, label] and it will adapt to a [value, label, isSelected, choiceMode ] format

Or the client can instead pass in a DAO to the MultiChoiceView.dao and the choices list will be automatically generated

Calling the following methods: 
    
1. MultiChoiceView.outputSelectedChoicesInValueLabelFormat() - will return an array containing the 
    selected choices in [ value, label ] format only if the minSelected, maxSelected criteria is respected
2. MultiChoiceView.outputSelectedChoicesInValueLabelFormat() - will return a predicated dao containing thee 
    selected choices in only if the minSelected, maxSelected criteria is respected

MultiChoiceView.data will be automatically set to a predicated dao based on the choices selected only if the minSelected, maxSelected criteria is respected, it will be foam.dao.NullDAO othrewise

this.booleanView is a ViewSpec for each choice. It defaults to foam.u2.CheckBox

Thinking of looking into abstracting a generic choice model which can be used for single choice (traditional choice view) and the multi choice view since the logic with wiring up the dao is similar


KernelDeimos:
Screenshots demonstrating behaviour as desired:

Screen Shot 2020-08-12 at 1 52 20 PM
Screen Shot 2020-08-12 at 1 52 46 PM

Error is displayed if there are too few suggestions, and remaining checkboxes become disabled after reaching the maximum number of selections.

Thanks @tharmaman, this is awesome!

src/foam/u2/view/CardSelectView.js
```diff
@@ -0,0 +1,70 @@
...
+  `,
+
+  properties: [
+    {
+      class: 'String',
+      name: 'label',
+      factory: function(){
+        return String(this.value);
+      }
+    },
```
kgrgreer: No trailing commas.

tharmaman: Fixed

PR #3789, opened 2020-08-12
cap int view update

Nauna:
addresses: https://nanopay.atlassian.net/browse/NP-1467

#### call example:
```.tag(foam.u2.view.EnumLegendView, { of: foam.nanos.crunch.CapabilityJunctionStatus, enumValueToHide: ['APPROVED'] })```

image

#### available capabilities will be listed on left - and will scroll down if overflow in container
image

### CODE REVIEW

File: src/foam/u2/view/EnumLegendView.js [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+  ^container {
+    border-style: solid;
+    border-width: thin;
+  }
+  ^eachValue {
+    display: inline-flex;
+    padding: 12px;
+    text-align: justify;
+  }
+  ^badge {
```
KernelDeimos: badge class is duplicated here. It may be better to use that `Style` model to keep the CSS is one place (we could try changing that to a CSS axiom later). In fact this class needs to be updated to add `^badge > * { background-color: none }` because the coloured rectangle on the inside is creating an artifact that doesn't look very clean.

src/foam/u2/crunch/CapabilityInterceptView.js
```diff
@@ -65,31 +103,40 @@ foam.CLASS({
...
+            }))
+          .end()
+          .start().addClass(this.myClass('legendSize')).tag(foam.u2.view.EnumLegendView, { of: foam.nanos.crunch.CapabilityJunctionStatus, enumValueToHide: ['APPROVED'] }).end()
         .end()
         .startContext({ data: this })
           .tag(this.CANCEL, { buttonStyle: 'SECONDARY' })
         .endContext();
     },
+
     function checkStatus(cap) {
```
KernelDeimos: Now that CrunchService exists a call to `this.crunchService.getJunction(null, cap.id)` is better. We currently have this logic duplicated in a lot of places so IMO we should work on getting those to use the CrunchService method also.

KernelDeimos (changes requested):

src/foam/u2/view/EnumLegendView.js
Nauna: adjustments coming

src/foam/u2/crunch/CapabilityInterceptView.js
Nauna: adjustments coming

KernelDeimos:
What if the legend was a slotted view which displays only one of these descriptions when a question mark is clicked?

Screen Shot 2020-08-12 at 1 23 51 PM

jlhughes:
Future - hovering over status gives the description of that status and option of expanding the full legend.  Just a thought.

src/foam/nanos/crunch/CapabilityJunctionStatus.js
```diff
@@ -10,46 +10,45 @@ foam.ENUM({
...
       label: 'granted',
-      documentation: `This is for capabilities that have passed all checks.`,
+      documentation: `This is for capabilities that have passed all checks. Capabilities in this status have their features unlocked.`,
       background: '#32bf5e'
     },
     {
       name: 'EXPIRED',
       label: 'expired',
-      documentation: `A cron runs to establish capabilities that are no longer valid and potentially expires a ucj.`,
+      documentation: `A capability can expire due to a number of factors. The data that was previously collected will be rest and will require to be re-completed to gain access.`,
```
jlhughes:
```suggestion
      documentation: `A capability can expire due to a number of factors. The data that was previously collected will be removed or reset and must be re-added or re-approved to gain access.`,
```

src/foam/u2/view/EnumLegendView.js
```diff
@@ -0,0 +1,59 @@
+foam.CLASS({
```
jlhughes, line 7: license header.

jlhughes, line 7:
```suggestion
/**
 * @license
 * Copyright 2020 The FOAM Authors. All Rights Reserved.
 * http://www.apache.org/licenses/LICENSE-2.0
 */

foam.CLASS({
```

PR #3791, opened 2020-08-12
Fix rule dao unindexed select

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-1688

## Changes
- Add daoKey and enabled indexes to Rule
- Fix scan plan cost function
- Fix simplifying True/False mlangs

### DIFF
--- a/src/foam/dao/index/ScanPlan.java
+++ b/src/foam/dao/index/ScanPlan.java
@@ -56,7 +56,7 @@ public class ScanPlan
         order_   = null;
       } else {
         sortRequired = true;
-        cost *= Math.log(cost) / Math.log(2);
+        cost = (long) (Math.log(cost) / Math.log(2));
       }
     }
 
--- a/src/foam/dao/index/ScanPlan.java
+++ b/src/foam/dao/index/ScanPlan.java
@@ -56,7 +56,7 @@ public class ScanPlan
         order_   = null;
       } else {
         sortRequired = true;
-        cost = (long) (Math.log(cost) / Math.log(2));
+        cost *= Math.log(cost) / Math.log(2);
       }
     }
 
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -510,6 +510,7 @@ foam.CLASS({
     },
     {
       name: 'partialEval',
+      code: function() { return this },
       javaCode: 'return foam.mlang.MLang.TRUE;'
     }
   ]
@@ -540,6 +541,7 @@ foam.CLASS({
     },
     {
       name: 'partialEval',
+      code: function() { return this },
       javaCode: 'return foam.mlang.MLang.FALSE;'
     }
   ]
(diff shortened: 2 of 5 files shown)


src/foam/dao/index/ScanPlan.java
```diff
@@ -56,7 +56,7 @@ public long calculateCost(PropertyInfo propertyInfo) {
         order_   = null;
       } else {
         sortRequired = true;
-        cost *= Math.log(cost) / Math.log(2);
+        cost = (long) (Math.log(cost) / Math.log(2));
```
kgrgreer: Why this change? Isn't the sorting cost n * log2(n)?

src/foam/mlang/mlang.js
```diff
@@ -508,6 +508,11 @@ foam.CLASS({
       swiftCode: 'return true',
       javaCode: 'return true;'
     },
+    {
+      name: 'partialEval',
+      code: function() { return this },
+      javaCode: 'return foam.mlang.MLang.TRUE;'
```
kgrgreer, line 514: Doesn't this already 'return this', which is the same?

src/foam/mlang/mlang.js
```diff
@@ -533,6 +538,11 @@ foam.CLASS({
       type: 'String',
       javaCode: 'return " 1 <> 1 ";',
       code: function() { return "1 <> 1"; }
+    },
+    {
+      name: 'partialEval',
+      code: function() { return this },
+      javaCode: 'return foam.mlang.MLang.FALSE;'
```
kgrgreer, line 545: Isn't this what already happens?

src/foam/dao/index/ScanPlan.java
lchanmann: Is it a search instead of sort? When searching the indexed plan this cost is being used.

src/foam/mlang/mlang.js
lchanmann: No, the singleton axiom doesn't work on java.

    foam.mlang.MLang.TRUE == new True();  // false

src/foam/mlang/mlang.js
lchanmann: No, the singleton axiom doesn't work on java.

src/foam/mlang/mlang.js
kgrgreer: I see. Then that's the real bug.  We need java Singleton support.

src/foam/dao/index/ScanPlan.java
kgrgreer: But we said sortRequired = true;
So we will need to sort the values first.

src/foam/mlang/mlang.js
kgrgreer: We can fix this in the future.

src/foam/mlang/mlang.js
kgrgreer: We can fix this in the future.

src/foam/dao/index/ScanPlan.java
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/3791/commits/00294323a4d50cc6a268b2564b4edcd352ddc90e.

PR #3793, opened 2020-08-13
Add additional helper constructors and create methods.

jlhughes:
Report when pmLogger is not found, except in Production.

### DIFF
--- a/src/foam/nanos/pm/PM.js
+++ b/src/foam/nanos/pm/PM.js
@@ -21,9 +21,9 @@ foam.CLASS({
     'foam.core.FObject',
     'foam.core.X',
     'foam.dao.DAO',
+    'static foam.mlang.MLang.EQ',
     'foam.nanos.alarming.Alarm',
-    'foam.nanos.alarming.AlarmConfig',
-    'static foam.mlang.MLang.EQ'
+    'foam.nanos.alarming.AlarmConfig'
   ],
 
   ids: [ 'key', 'name', 'startTime' ],
@@ -88,6 +88,19 @@ foam.CLASS({
     PMLogger pmLogger = (PMLogger) x.get(DAOPMLogger.SERVICE_NAME);
     if ( pmLogger != null ) {
       pmLogger.log(this);
+    } else {
+      foam.nanos.app.AppConfig app = (foam.nanos.app.AppConfig) x.get("appConfig");
+      if ( app != null &&
+           app.getMode() != foam.nanos.app.Mode.PRODUCTION ) {
+        Exception e = new Exception("PMLogger not found");
+        foam.nanos.logger.Logger logger = (foam.nanos.logger.Logger) x.get("logger");
+        if ( logger != null ) {
+          logger.error(e.getMessage(), this.toString(), e);
+        } else {
+          System.err.println(e.getMessage());
+          e.printStackTrace();
+        }
+      }
     }
       `
     },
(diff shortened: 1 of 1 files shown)


src/foam/nanos/pm/PM.js
```diff
@@ -88,6 +88,19 @@ foam.CLASS({
     PMLogger pmLogger = (PMLogger) x.get(DAOPMLogger.SERVICE_NAME);
     if ( pmLogger != null ) {
       pmLogger.log(this);
+    } else {
+      foam.nanos.app.AppConfig app = (foam.nanos.app.AppConfig) x.get("appConfig");
```
kgrgreer: Is this block really needed? If someone doesn't want PM's, they could just not install the PMLogger and they shouldn't be bombarded with exceptions because of it. Thoughts?

src/foam/nanos/pm/PM.js
```diff
@@ -112,16 +125,38 @@ foam.CLASS({
...
         setIsError(true);
+        setEndTime(new java.util.Date());
         StringBuilder sb = new StringBuilder();
         for (Object obj: args) {
           if ( obj instanceof Exception ) {
             setException(obj);
-            sb.append(((Exception) obj).getMessage()).append(",");
+            sb.append(((Exception) obj).getMessage());
+          } else {
+            sb.append(obj.toString());
```
kgrgreer: Do you need the '.toString()'?

src/foam/nanos/pm/PM.js
```diff
@@ -112,16 +125,38 @@ foam.CLASS({
...
         }
-        if ( sb.length() > 0 )
+        if ( sb.length() > 0 ) {
           setErrorMessage(sb.deleteCharAt(sb.length() - 1).toString());
-      `
+        }
+        PMLogger pmLogger = (PMLogger) x.get(DAOPMLogger.SERVICE_NAME);
+        if ( pmLogger != null ) {
+          pmLogger.log(this);
+        } else {
```
kgrgreer: same comment as above

src/foam/nanos/pm/PM.js
```diff
@@ -203,24 +250,47 @@ foam.CLASS({
...
               return sb.deleteCharAt(sb.length() - 1).toString();
             }
+
+            private static String combine(Object... args) {
+              if ( args == null ) {
+                return "";
+              }
+              StringBuilder sb = new StringBuilder();
+              for ( Object o: args) {
+                sb.append(o.toString()).append(":");
```
kgrgreer: Is .toString() needed? Then, do we need a String... version if we already have an Object... version?

src/foam/nanos/pm/PM.js
jlhughes: Having it has revealed a few places where we were trying log PMs and it was not working. If we don't want PMs or Logs then we can install the appropriate Null logger.  On the fence but I keep adding this anytime I'm working/debugging.

KolombetM (approved):

src/foam/nanos/pm/PM.js
```diff
@@ -203,24 +224,47 @@ foam.CLASS({
...
+              }
+              if ( args.length > 1 ) {
+                setName(combine(java.util.Arrays.copyOfRange(args, 1, args.length)));
               }
-              if ( sb.length() > 0 )
-                setName(sb.deleteCharAt(sb.length() - 1).toString());
               init_();
             }
 
             private static String combine(String... args) {
```
KolombetM: Do you need combine method for String..., if you have one for Object...?

jlhughes: @Mykola97 I couldn't compile without it. The calls with Object... and String... did not compile with just combine taking Object.

PR #3802, opened 2020-08-15
Define Wizardlets per capability

zahnno:
Capabilities can define wizardlets and crunch wizard config used by client crunch wizards.

Use case:  Defining whether wizard can skip to other steps or disabling next button if cap props have satisfied validation predicates.


jlhughes:
Reviewed by Eric.

PR #3803, opened 2020-08-15
[NP-1875] Move reference to Capability from Theme

KernelDeimos:
This fixes an error message that's appearing in DocBrowser. Two changes were made:

1. Moved admissionCapability from Theme to a refinement in CRUNCH

    This may not be the root cause, but this change should be made anyway as referring to CRUNCH from Theme (which is loaded earlier) is incorrect.

2. Changed `Reference` property to `String`

    In spite of resolving the dependancy issue, an error still persists if `admissionCapability` is a Reference of foam.nanos.crunch.Capability when using `src/foam/doc/index.html`.

### DIFF
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -344,6 +344,23 @@ foam.CLASS({
   sections: [{ name: 'capabilities' }]
 });
 
+foam.CLASS({
+  package: 'foam.nanos.crunch',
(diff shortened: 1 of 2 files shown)


nanopay-arthur (approved):
LGTM. I think we will need to sync up with kgr about why foam2/src/foam/doc/index.html is throwing errors when setting admissionCapability's class to Reference and how we can fix it.

PR #3805, opened 2020-08-16
Refactor FObjectFormatter to route all output through a single append…

jlhughes:
… call to simplify child classes intercept all buffer updates.

NP-1828 #time 4h


src/foam/lib/json/JSONParser.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * @license
+ * Copyright 2017 The FOAM Authors. All Rights Reserved.
```
smirnovaka, line 3: 2020

smirnovaka (approved):

src/foam/lib/json/JSONParser.js
jlhughes: This was converted from java to js - so it did have a copyright already.

PR #3809, opened 2020-08-17
[NP-1844] Tie CrunchService logic into FlatCapability

KernelDeimos:
Additionally, this PR adds the method `updateJunction` to CrunchService

### DIFF
--- a/src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
+++ b/src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
@@ -42,7 +42,6 @@ foam.CLASS({
     {
       name: 'put_',
       javaCode: `
-        // TODO: add logic to disprese FObjectArray to lower level capabilities
         DAO capabilityDAO = (DAO) x.get("capabilityDAO");
         ConnectedCapability sentCap = (ConnectedCapability) obj;
         FlatCapability flatCap = (FlatCapability)
--- a/src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
+++ b/src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
@@ -12,8 +12,12 @@ foam.CLASS({
 
   javaImports: [
     'foam.nanos.crunch.Capability',
+    'foam.nanos.crunch.CrunchService',
+    'foam.core.ClassInfo',
+    'foam.core.FObject',
     'foam.core.X',
-    'foam.dao.DAO'
+    'foam.dao.DAO',
+    'java.util.List'
   ],
 
   documentation: `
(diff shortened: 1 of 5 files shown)


src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
```diff
@@ -39,6 +43,36 @@ foam.CLASS({
       name: 'put_',
       javaCode: `
         // TODO: add logic to disprese FObjectArray to lower level capabilities
```
nanoscott: remove this todo

src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
```diff
@@ -39,6 +43,36 @@ foam.CLASS({
...
+        DAO capabilityDAO = (DAO) x.get("capabilityDAO");
+        ConnectedCapability sentCap = (ConnectedCapability) obj;
+        FlatCapability flatCap = (FlatCapability)
+          capabilityDAO.find(sentCap.getFlatCapabilityId());
```
nanoscott, line 48: use the sentCap.findFlatCapability(x) call instead here (if this is a reference).

src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
```diff
@@ -39,6 +43,36 @@ foam.CLASS({
...
+          }
+        }
+
+        CrunchService crunchService = (CrunchService) x.get("crunchService");
+
+        // Second pass: store UCJs
+        for ( int i = 0 ; i < classes.length ; i++ ) {
+          FObject dataObj = (FObject) flatData[i];
+          String targetId = capabilities[i];
+          crunchService.updateJunction(x, targetId, dataObj);
```
nanoscott, line 73: So, I'm assuming granting the capability happens automatically if all the data is filled in correctly, right?

src/foam/nanos/crunch/connection/FlatCapabilityDAO.js
```diff
@@ -90,6 +93,26 @@ foam.CLASS({
       name: 'put_',
       javaCode: `
         // TODO: add logic to retrieve capabilities in order of precendence
```
nanoscott: removed TODO

src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
KernelDeimos: sentCap is the ConnectedCapability object. It looks like it doesn't implement that method - should it?

src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
KernelDeimos: That's correct. This is done by a rule that @noodlemoodle added called SetUCJStatusOnPut.

src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
nanoscott: If you look at the generated code, you can see this method:

`  public foam.nanos.crunch.connection.FlatCapability findFlatCapabilityId(foam.core.X x) {
    return (foam.nanos.crunch.connection.FlatCapability)((foam.dao.DAO) x.get("flatCapabilityDAO")).find_(x, (Object) getFlatCapabilityId());
  }`

Maybe the field should be named flatCapability instead of flatCapabilityId so this method makes sense.

nanoscott (approved):

nanoscott:
@KernelDeimos .. renamed to flatCapability from flatCapabilityId .. you'll need to merge in the latest and update this PR so build doesn't break.

KernelDeimos:
@tharmaman  Remaining work to do after this PR:
- capabilityDAO should be decorated with FlatCapabilityDAO
- service should be added for ConnectedCapabilityDAO. This just does the UCJ puts so doesn't need a journal.
- Test 0: Putting a FlatCapability into capabilityDAO should automatically populate `classes` and `capabilities`
- Test 1: updateUCJ should be tested before ConnectedCapabilityDAO
- Test 2: Putting an object into ConnectedCapabilityDAO should cause the payloads to be distributed to their respective UCJS

tharmaman:
Already done by scott


> * capabilityDAO should be decorated with FlatCapabilityDAO
> * service should be added for ConnectedCapabilityDAO. This just does the UCJ puts so doesn't need a journal.

PR #3811, opened 2020-08-17
Fix simplifying True predicate

lchanmann:
## Issues
- simplifyPredicate() doesn't nullify True predicate.

## Changes
- Move nullifying True predicate before returning

### DIFF
--- a/src/foam/dao/index/TreeIndex.java
+++ b/src/foam/dao/index/TreeIndex.java
@@ -94,12 +94,11 @@ public class TreeIndex
           ((And) predicate).getArgs()[i] = new True();
         }
       }
+      // use partialEval to simplify predicate themselves.
+      predicate = predicate.partialEval();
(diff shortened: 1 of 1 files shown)


src/foam/dao/index/TreeIndex.java
```diff
@@ -96,9 +96,9 @@ public Object bulkLoad(FObject[] a) {
       }
       // use partialEval to simplify predicate themselves.
       predicate = predicate.partialEval();
-      if ( predicate instanceof True ) return new Object[]{state, null};
     }
 
+    if ( predicate instanceof True ) predicate = null;
```
jlhughes, line 101: I so hate this style of no {} with the if.  Had to go look at the entire source to understand where the final return went.  An empty line would help in future.

jlhughes (approved):

PR #3814, opened 2020-08-17
Define property type

zahnno:
Added Property type enum and reference on Address


jlhughes:
Where is this coming from?

jlhughes:
ok for now. this is the kind of thing that would be country specific when we refactor address.

PR #3816, opened 2020-08-18
create supportTypeDAO, and move fileDropZon to foam

yij793:
address: https://github.com/nanoPayinc/NANOPAY/pull/10390
note: create new DAO call supportFileTypeDAO use to store file types. and create decorator for fileDAO in order to grab file type and save it


src/foam/nanos/fs/fileDropZone/FileDropZone.js
```diff
@@ -0,0 +1,335 @@
+/**
```
jlhughes, line 1: Replace with Foam license header, please.  And others.

src/foam/nanos/fs/fileDropZone/FileDropZone.js
```diff
@@ -0,0 +1,335 @@
...
+  extends: 'foam.u2.Controller',
+
+  documentation: 'A default zone to drag & drop files',
+
+  requires: [
+    'foam.log.LogLevel',
+    'foam.blob.BlobBlob',
+    'foam.nanos.fs.File',
+    'foam.nanos.fs.FileArray',
+    'net.nanopay.ui.modal.ModalHeader'
```
jlhughes: These also need to move.  else this can't move.

src/services
```diff
@@ -1168,3 +1168,21 @@ p({
...
+  "name":"supportFileTypeDAO",
+  "serve":true,
+  "serviceScript":
+  """
+    return new foam.dao.EasyDAO.Builder(x)
+      .setPm(true)
+      .setPipelinePm(true)
+      .setGuid(true)
+      .setOf(foam.nanos.fs.FileType.getOwnClassInfo())
+      .setJournalName("supportFileType")
```
jlhughes:
```suggestion
      .setJournalName("fileTypes")
```

src/foam/nanos/fs/SupportFileTypeDAO.js
```diff
@@ -0,0 +1,72 @@
...
+            setX(x);
+            setDelegate(delegate);
+          }  
+        `
+        );
+      }
+    }
+  ],
+
+  methods: [
```
jlhughes, line 43: I would prefer we don't do this.  If we do do this, then there is no point - we allow anything. 
Our current supported list would be the default - in a repo provided journal - filetypes.jrl  
what was the intent?

src/services
```diff
@@ -1168,3 +1168,21 @@ p({
   """,
   "client": "{\"of\":\"foam.nanos.auth.TimeZone\"}"
 })
+
+p({
+  "class":"foam.nanos.boot.NSpec",
+  "name":"supportFileTypeDAO",
```
jlhughes:
```suggestion
  "name":"fileTypeDAO",
```

src/foam/nanos/fs/SupportFileTypeDAO.js
```diff
@@ -0,0 +1,68 @@
...
+            setDelegate(delegate);
+          }  
+        `
+        );
+      }
+    }
+  ],
+
+  messages: [
+    { name: 'FILE_TYPE_NOT_SUPPORT', message: 'Such file type are not supported' },
```
jlhughes:
```suggestion
    { name: 'FILE_TYPE_NOT_SUPPORT', message: 'File type not supported.' },
```

src/foam/nanos/fs/FileType.js
```diff
@@ -0,0 +1,27 @@
...
+
+  properties: [
+    {
+      class: 'Long',
+      name: 'id',
+      visibility: 'RO'
+    },
+    {
+      class: 'String',
+      name: 'fileName',
```
jlhughes: what is 'fileName' - can it just be 'name'? 
and 'id' should be string and GUID.  or is name and mimeType a unique combination? if so then the id of this model can be 
ids['name', 'mimeType'] 
And this is really the model for MimeType
with properties: 
type  and
subType
is it not?

src/foam/nanos/fs/SupportFileTypeDAO.js
```diff
@@ -0,0 +1,70 @@
...
+        return super.put_(x, obj);
+      }
+
+      File file = (File) obj;
+      DAO fileTypeDAO = (DAO) x.get("fileTypeDAO");
+      FileType fileType = (FileType) fileTypeDAO.find(
+        EQ(FileType.MIME, file.getMimeType())
+      );
+
+      // if fileType not store in fileTypeDAO, create new one and store it
```
jlhughes: fix or remove doc - no longer matches.

src/services
```diff
@@ -1168,3 +1168,22 @@ p({
   """,
   "client": "{\"of\":\"foam.nanos.auth.TimeZone\"}"
 })
+
+p({
```
jlhughes: move this foam/nanos/fs/services.jrl

src/services
```diff
@@ -561,7 +561,7 @@ p({
     dao = new foam.dao.KeyValueDAO.Builder(x)
       .setDelegate(new foam.dao.NullDAO(x, foam.nanos.fs.File.getOwnClassInfo()))
       .build();
-
+    dao = new foam.nanos.fs.SupportFileTypeDAO(x, dao);
```
jlhughes: why is this on demoObjectDAO? 
Should this not decorate the FileDAO stack?

src/foam/nanos/fs/FileType.js
yij793: you are right, gona rewrite this

src/services
yij793: this is on FileDAO stack, not demoObjectDAO

jlhughes: ok, last thing, can you move this to fs/services.jrl as well please.  Realize you are just updating, but since we are moving stuff around this is the moment. Thanks.   It just helps down the road with less merge conflicts when the nspecs are in separate service files rather than one big one.

jlhughes (approved):

PR #3819, opened 2020-08-18
Fixing empty of properties in FlatCapabilities

nanoscott:


KernelDeimos:
Good catch, although if an `of`less capability is a prerequisite of something it'll still need a UCJ put, so maybe FlatCapability needs a list of capabilities to put empty UCJs for (this can happen at the very end though, doesn't need to happen in order)

(consider this a non-blocking review comment because it might be easier to handle this when it comes up)

src/foam/nanos/crunch/connection/ConnectedCapabilityDAO.js
```diff
@@ -51,22 +52,40 @@ foam.CLASS({
...
-          FObject dataObj = (FObject) flatData[i];
+        int index = 0;
+        for ( var data : flatData ) {
+          // Skip capability where 'of' property is empty
+          while ( index < classes.length && SafetyUtil.isEmpty(classes[index]) ) {
+            index++;
+          }
+
+          if ( index >= classes.length ) {
+            throw new RuntimeException("Unexpected data provided: " + data);
```
jlhughes, line 63: This seems like a developer/configuration error as well so should generate Log warning/error.  Similar with below.

PR #3828, opened 2020-08-19
[NP-1834] Add server implementation for Capable

KernelDeimos:
I haven't included a property for UCJ capability requirements yet as I'm thinking there may be a better place for it. We can add a property to CapabilityRuntimeException for Capable objects and handle them as an ordinary intercept situation, with potentially minor changes to CrunchController to iterate over the Capables and create CapableCapabilityWizardlets

### DIFF
--- a/src/foam/nanos/crunch/CrunchService.js
+++ b/src/foam/nanos/crunch/CrunchService.js
@@ -33,6 +33,52 @@ foam.INTERFACE({
       ]
     },
     {
+      name: 'getCapabilityPath',
+      documentation: `
+        getGrantPath provides an array of capability objects representing
+        the list of capabilities required to grant the desired capability.
+      `,
+      async: true,
+      type: 'List',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'sourceId',
+          type: 'String'
+        },
+        {
+          name: 'filterGrantedUCJ',
+          type: 'boolean'
+        }
+      ]
+    },
+    {
+      name: 'getMultipleCapabilityPath',
+      documentation: `
+        getGrantPath provides an array of capability objects representing
+        the list of capabilities required to grant the desired capability.
+      `,
+      async: true,
+      type: 'List',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'capabilities',
+          type: 'String[]'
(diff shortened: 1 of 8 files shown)


src/foam/nanos/crunch/lite/Capable.js
```diff
@@ -0,0 +1,150 @@
+foam.INTERFACE({
```
jlhughes, line 7: license header.

src/foam/nanos/crunch/ui/CapableObjectWizardlet.js
```diff
@@ -0,0 +1,70 @@
...
+
+  properties: [
+    // Properties specific to CapabilityWizardSection
+    {
+      name: 'capability'
+    },
+    {
+      name: 'targetObject',
+      class: 'FObjectProperty',
+      of: 'foam.nanos.crunch.Capable'
```
jlhughes:
```suggestion
      of: 'foam.nanos.crunch.lite.Capable'
```

src/foam/nanos/crunch/CrunchService.js
```diff
@@ -32,6 +32,52 @@ foam.INTERFACE({
...
+      `,
+      async: true,
+      type: 'List',
+      args: [
+        {
+          name: 'x',
+          type: 'Context'
+        },
+        {
+          name: 'capabilities',
```
jlhughes: is this a list of capabilityId ? 
If so then perhaps the above could be 
capabilityId and this capabilityIds 
or is a capabilityId a capability and I'm using the wrong term.  Then the above could be capability, and this one capabilities.   Just trying to be consistent.

src/foam/nanos/crunch/lite/CapableObjectData.js
```diff
@@ -0,0 +1,13 @@
+foam.CLASS({
```
jlhughes, line 7: license header.

src/foam/nanos/crunch/ui/CapableObjectWizardlet.js
KernelDeimos: Oops, I didn't even realize that was here. Pushed some deletions to that since some of those methods were just copied from CapabilityWizardlet.

jlhughes (approved):

PR #3829, opened 2020-08-20
update views for Crunch Lab user and capability selection

Nauna:


src/foam/u2/crunch/lab/CrunchLab.js
```diff
@@ -42,12 +45,36 @@ foam.CLASS({
...
       class: 'Reference',
-      of: 'foam.nanos.auth.User'
+      of: 'foam.nanos.auth.User',
+      view: function(_, X) {
+        return {
+          class: 'foam.u2.view.RichChoiceView',
+          search: true,
+          sections: [
+            {
+              heading: 'Services',
```
jlhughes: is this the label you want for users?

Nauna: Good catch!! Adjusted

PR #3833, opened 2020-08-20
Added validation on streetNumber

minsunny:
https://nanopay.atlassian.net/browse/NP-1868

Added a validation on streetNumber itself so that wizard view's next/save buttons become disabled when error message thrown.

Screen Shot 2020-08-20 at 3 02 03 PM

### DIFF
--- a/src/foam/nanos/auth/Address.js
+++ b/src/foam/nanos/auth/Address.js
@@ -26,8 +26,8 @@ foam.CLASS({
       class: 'Boolean',
       name: 'structured',
       value: true,
-      documentation: `Determines whether the address is shown in the following structure: 
-        Street Number, Street Name, Suite Number. For an unstructured address field, 
+      documentation: `Determines whether the address is shown in the following structure:
+        Street Number, Street Name, Suite Number. For an unstructured address field,
         use address1 and/or address2.
       `,
       hidden: true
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/Address.js
```diff
@@ -155,9 +155,7 @@ foam.CLASS({
           predicateFactory: function(e) {
             return e.OR(
               e.EQ(foam.nanos.auth.Address.STRUCTURED, false),
-              e.GTE(foam.mlang.StringLength.create({
-                arg1: foam.nanos.auth.Address.STREET_NUMBER
-              }), 1)
+              e.REG_EXP(foam.nanos.auth.Address.STREET_NUMBER, /^[0-9]{1,16}$/)
```
zahnno, line 158: House numbers/street numbers can consist of alphabet characters, surprisingly considering the number part of the noun.
Here's a bit of a blurb about it: https://www.architecturelab.net/whats-a-street-number-and-whats-an-address/

minsunny: @patrick ohh thank. I will fix it in another way

minsunny:
no necessary

src/foam/nanos/auth/Address.js
jlhughes: Around here, we have 27A and B, and also 27 1/2  for example.  why are we bothering with this?

minsunny: in Contact.js there already checks Pattern streetNumber = Pattern.compile("^[0-9]{1,16}$"); but it is wrong. I will change to compare the length of input.

PR #3835, opened 2020-08-20
update z-index for view and status colour

Nauna:
https://nanopay.atlassian.net/browse/NP-1901
addresses pts 4-5 in above ticket


src/foam/nanos/crunch/CapabilityJunctionStatus.js
```diff
@@ -29,7 +29,7 @@ foam.ENUM({
       name: 'ACTION_REQUIRED',
       label: 'action required',
       documentation: `A capability started and requires further action to complete.`,
-      background: 'pink'
+      background: '#cf6f0a'
```
jlhughes, line 32: Do we not have these in Theme?

PR #3837, opened 2020-08-24
Handle null string and report unexpected types

jlhughes:


jlhughes:
captured in another PR.

PR #3838, opened 2020-08-24
Flat + ConnectedCapability Rework into CapabilityPayload with FObjects

tharmaman:
Steps:
1. On **CapabilityPayloadDAO**, do a find where the id is of the capability or a select for a list of the capabilities that you want
2. Based  on  **crunchService.getGrantPath()**, the  **CapabilityPayloadDAO** will return a **CapabilityPayload** object, which has stores a map based on getGrantPath where the key is  **capability.getName()**  and value is **an instance of the capability.getOf()**. if .getOf() is null then the key will map to a null.
3. Fill out **CapabilityPayload.capabilityDataObjects** and then make a put of the filled out **CapabilityPayload** into the **CapablityPayloadDAO**
4. The backend will call grantPath once again and cross reference the grantPath values with CapabilityPayload.capabilityDataObjects to find the data objects for their respective capabilities in order of the grantPath
4. The **UCJs** should be populated accordingly after calling **crunchService.updateJunction** on each cross reference

IMPORTANT: Basically had to made the Of for CapabilityPayloadDAO service be of foam.nanos.crunch.Capability  so that the predicate will work nicely since this dao is basically a facade for the capabilityDAO

Example:
1. Querying the id of the capability I need, in this case it's Eric's Capability with an ID of "ff1b2cbc-4e35-4f90-a76a-0ef3ff13824f-100"
1a

The resulting output:
1b

2. Make a put to the CapabilityPayloadDAO  with basically a modified version of the JSON from above:
2a

The resulting output:
2b

3. Checking userCapabilityJunctionDAO if ucjs got updated correctly:
3a

The resulting output:
3b


src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
```diff
@@ -0,0 +1,147 @@
...
+          CrunchService crunchService = (CrunchService) x.get("crunchService");
+
+          List grantPath = crunchService.getGrantPath(x, idToString);
+          List<Object> capabilityDataObjects = new ArrayList<>();
+          List<String> capabilities = new ArrayList<>();
+          for ( Object item : grantPath ) {
+            // TODO: Verify that MinMaxCapability lists can be ignored here
+            if ( item instanceof Capability ) {
+              Capability cap = (Capability) item;
+              if ( cap.getOf() != null ){
```
nanoscott, line 131: you aren't handling empty 'of' objects. the UCJ still needs to be created for these even though they have empty ofs .. they can't be ignored.

tharmaman: It still gets created, its added to the capabilities array just not the capabilityDataObjects array. It's like how you guys wired it up before where Capabilities with no of don't get added to the classes array

tharmaman: I think before was the assumption where the  empty 'of' objects would always be  at the end, so I think  we  can  also push null into the capabilityDataObjects array for empty 'of' objects in the case that empty of objects appear somewhere in the middle of the grant path

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
```diff
@@ -0,0 +1,147 @@
...
+    },
+    {
+      name: 'put_',
+      javaCode: `
+        CapabilityPayload sentCapPayload = (CapabilityPayload) obj;
+
+        Object[] capabilityDataObjects = sentCapPayload.getCapabilityDataObjects();
+        String[] capabilities = sentCapPayload.getCapabilities();
+
+        // TODO: Don't think this block is needed anymore
```
nanoscott: remove dead code .. we can always refer to the history if we really need this.

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
```diff
@@ -0,0 +1,147 @@
...
+            if ( item instanceof Capability ) {
+              Capability cap = (Capability) item;
+              if ( cap.getOf() != null ){
+                FObject newOfInstance;
+                try {
+                  newOfInstance = (FObject) cap.getOf().newInstance();
+                } catch (Exception e){
+                  throw new RuntimeException(e);
+                }
+                capabilityDataObjects.add(newOfInstance);
```
nanoscott: the indexes of this array have to correspond exactly to the indexes of the capabilities array in order for the put method to work properly, or else you'll be putting the wrong data object into the wrong UCJ

tharmaman: Re new fix

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
tharmaman: Done.

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
tharmaman: Fixed.

PR #3840, opened 2020-08-24
[NP-1885] add country regions

jinnnn9:
addresses: https://nanopay.atlassian.net/browse/NP-1885


src/foam/nanos/auth/Address.js
```diff
@@ -219,27 +223,326 @@ foam.CLASS({
...
+            );
+          },
+          errorMessage: 'INVALID_POSTAL_CODE_ERR_MSG'
+        },
+        // Austria
+        {
+          args: ['postalCode', 'countryId'],
+          predicateFactory: function(e) {
+            return e.OR(
+              e.NEQ(foam.nanos.auth.Address.COUNTRY_ID, 'AT'),
```
jlhughes, line 236: @kgrgreer can we use mmethod here for switching on country code to get the appropriate reg-ex?

src/foam/nanos/auth/Address.js
```diff
@@ -219,27 +223,326 @@ foam.CLASS({
...
                 foam.nanos.auth.Address.POSTAL_CODE,
-                /^^\d{5}(?:[-\s]\d{4})?$/i)
+                /^\d{5}(?:[-\s]\d{4})?$/i)
+            );
+          },
+          errorMessage: 'INVALID_POSTAL_CODE_ERR_MSG'
+        },
+        // Austria
+        {
+          args: ['postalCode', 'countryId'],
```
jlhughes, line 233: validationPredicates - on each call we are walking them all?

src/foam/nanos/auth/Address.js
kgrgreer: No, mmethod only works with types. We could make a switch/case mlang to make this easier, but we don't current have such a thing.

src/foam/nanos/auth/Address.js
jinnnn9: @jlhughes now you mention it, it doesn't look very efficient. I will make changes to it if we can go with your suggestion above.

jlhughes:
Merging as is. Will track improving validation check logic in another PR.

PR #3841, opened 2020-08-24
[NP-1834] Add CrunchService maybeIntercept method

KernelDeimos:
Adds a method any service can use to require a capability or an option from an array of capabilities.

Example:
```
CrunchService crunchService = (CrunchService) x.get("crunchService");
crunchService.require(new String[]{"554af38a-8225-87c8-dfdf-eeb15f71215f-49"});
```

### DIFF
--- a/src/foam/nanos/crunch/CrunchService.js
+++ b/src/foam/nanos/crunch/CrunchService.js
@@ -115,7 +115,12 @@ foam.INTERFACE({
       ],
     },
     {
-      name: 'require',
+      name: 'maybeIntercept',
+      documentation: `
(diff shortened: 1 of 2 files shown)


kgrgreer:
What does it mean to "require" a Capability? Could you use a different term to not overload FOAM's current use of "require"? Please provide more docs/comments.

KernelDeimos:
Updated to `maybeIntercept` as a possibly better name and added some docs. This method allows a block of code to throw CapabilityRuntimeException if a capability isn't satisfied, without necessarily creating a permission.

PR #3842, opened 2020-08-24
[DO_NOT_MERGE] [NP-1916] Simplify UCJ status checks

KernelDeimos:
This PR adds a BasicCapabilityJunctionStatus. This makes it possible to add statuses for edge cases without updating lots of rules. Additionally, this fixes a bug in CrunchService caused by only checking `GRANTED` status without checking `GRACE_PERIOD` (which is effectively granted); this also serves as an example of using this new status.

Credit to @noodlemoodle for coming up with the idea and @kgrgreer for showing me an easy way to implement it.

### DIFF
--- a/src/foam/nanos/crunch/CapabilityJunctionStatus.js
+++ b/src/foam/nanos/crunch/CapabilityJunctionStatus.js
@@ -6,49 +6,65 @@
 foam.ENUM({
   package: 'foam.nanos.crunch',
   name: 'CapabilityJunctionStatus',
+
+  properties: [
+    {
+      name: 'basicStatus',
+      class: 'Enum',
+      of: 'foam.nanos.crunch.BasicCapabilityJunctionStatus'
+    }
+  ],
+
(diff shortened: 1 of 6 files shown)


src/foam/nanos/crunch/CapabilityJunctionStatus.js
```diff
@@ -6,49 +6,65 @@
...
+      background: 'grey',
+      basicStatus: 'GRANTED'
     },
     {
       name: 'APPROVED',
       label: 'approved',
       documentation: `- not seen by users - Denoting a UCJ requiring review has been approved. It would need to go through the rules of the ucjDAO before 
       being set to granted.`,
-      background: '#bfae32'
+      background: '#bfae32',
```
jlhughes, line 66: So this enum property should have originally been named 'APPROVAL', rather than APPROVED.

jlhughes (approved):

PR #3843, opened 2020-08-25
js implementation for timeofday

xuerongNanopay:


src/foam/mlang/expr/TimeOfDay.js
```diff
@@ -45,6 +45,38 @@ foam.CLASS({
...
+      code: function() {
+        var typeToIdx = {
+          year: 0,
+          month: 1,
+          day: 2,
+          hour: 3,
+          minute: 4,
+          second: 5
+        };
+        var zonetime = new Intl.DateTimeFormat("en-US", {
```
jlhughes: @lchanmann @hchoura do we want/need to consider locale here?

src/foam/mlang/expr/TimeOfDay.js
```diff
@@ -45,6 +45,38 @@ foam.CLASS({
...
     {
       name: 'f',
+      code: function() {
+        var typeToIdx = {
+          year: 0,
+          month: 1,
+          day: 2,
+          hour: 3,
+          minute: 4,
+          second: 5
```
lchanmann: not being used.

src/foam/mlang/expr/TimeOfDay.js
```diff
@@ -45,6 +45,38 @@ foam.CLASS({
...
+        var p = [];
+        for ( var i = 0 ; i < formatted.length ; i++ ) {
+          var { type, value } = formatted[i];
+          var offset = typeToIdx[type];
+
+          if ( ! (offset === undefined) ) {
+            p[offset] = parseInt(value, 10);
+          }
+        }
+        var d = new Date(p[0], p[1], p[2], this.hour, this.minute, thie.second);
```
lchanmann: can we use object literal for capturing the date parts?
```suggestion
        var d = new Date(p.year, p.month, p.day, this.hour, this.minute, thie.second);
```

lchanmann: then `typeToIdx` can simply be an array of all the date parts of interest. Eg.

    var dateParts = ['year', 'month', 'day'];
    ...
    for ( var i = 0 ; i < formatted.length ; i++ ) {
        var { type, value } = formatted[i];
        if ( dateParts.indexOf(type) > -1 ) {
            p[type] = parseInt(value, 10);
        ...

src/foam/mlang/expr/TimeOfDay.js
```diff
@@ -45,6 +45,32 @@ foam.CLASS({
   methods: [
     {
       name: 'f',
+      code: function() {
+        var typeToIdx = {
+          year: 0,
+          month: 1,
+          day: 2
+        };
+        var zonetime = new Intl.DateTimeFormat(undefined, {
```
lchanmann: Can we use 'default' instead of `undefined`? 
```suggestion
        var zonetime = new Intl.DateTimeFormat('default', {
```
@jlhughes, the default locale would suffice as the locale here is for formatting the data and time for the parser below (lines 61-69) so that we can extract the date parts appropriately.

src/foam/mlang/expr/TimeOfDay.js
```diff
@@ -45,6 +45,32 @@ foam.CLASS({
...
+          hour12: false,
```
lchanmann: hour12 option is not needed.

lchanmann (approved):

jlhughes (approved):

PR #3844, opened 2020-08-25
Cpf 4830/add search column selection

olhabn:
@kennyqiyenkan @kgrgreer @jlhughes 
please review.

for ticket https://nanopay.atlassian.net/browse/CPF-4830

those are suggestions for removing "infinite loop" of one expression or subscription triggering another one. please let me know what you think. thanks!


src/foam/u2/filter/properties/ReferenceFilterView.js
```diff
@@ -152,15 +152,18 @@ foam.CLASS({
...
+          var refObj = referenceObjectsArray.find(r => r.id == daoContents.groupKeys[i] );
+          if ( refObj ) {
+            var objectId = refObj.id;
+            var summary = refObj.toSummary();
             if ( summary ) {
               result[objectId] = summary;
             } else {
               result[objectId] = `ID: ${objectId}`;
             }
+          } else {
```
olhabn, line 165: this case is for reference property not being set

PR #3846, opened 2020-08-25
[DO NOT MERGE]adding java Keyword cupport for references

olhabn:
reference columns search backend support

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -2284,13 +2284,15 @@ foam.CLASS({
 
   javaImports: [
     'foam.core.PropertyInfo',
+    'foam.mlang.ReferenceExpressionHelper',
     'java.lang.reflect.Method',
     'java.text.DateFormat',
     'java.text.SimpleDateFormat',
     'java.util.Date',
     'java.util.Iterator',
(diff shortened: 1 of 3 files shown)


src/foam/mlang/mlang.js
```diff
@@ -2380,7 +2382,9 @@ while ( i.hasNext() ) {
 
   try {
     String s = "";
-    if ( prop instanceof foam.core.AbstractFObjectPropertyInfo ) {
+    if ( ReferenceExpressionHelper.isPropertyAReference((PropertyInfo)prop) ) {
```
olhabn, line 2385: I've added backend support but I'm not sure what to do with front-end as js side changes must synchronous and I'm not sure how to implement that.
@kgrgreer @jlhughes please let me know what you think

src/foam/mlang/mlang.js
```diff
@@ -2380,7 +2382,9 @@ while ( i.hasNext() ) {
...
+      if ( checkNestedFObject(REF(prop).f(obj)) ) return true;
```
olhabn, line 2386: may create infinite loop need to check for references that were checked. or it's overkill and i will close this pr later

PR #3865, opened 2020-08-27
add extends to junctions

noodlemoodle:
needed for usercapabiiltyjunction : want to extend a renewable class where all the expiry related info is tracked

e.g. 
```
foam.RELATIONSHIP({
  package: 'foam.nanos.crunch',
  extends:'foam.nanos.crunch.Renewable',
  sourceModel: 'foam.nanos.auth.User',
  targetModel: 'foam.nanos.crunch.Capability',
   ...
});

### DIFF
--- a/src/foam/dao/Relationship.js
+++ b/src/foam/dao/Relationship.js
@@ -173,6 +173,15 @@ foam.CLASS({
           .filter(f => global.FOAM_FLAGS[f]);
         return foam.util.flagFilter(enabledFlags)(this);
       }
+    },
+    {
+      class: 'String',
+      name: 'extends',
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/foam/dao/Relationship.js [hunk 1]
```diff
@@ -173,6 +173,15 @@ foam.CLASS({
           .filter(f => global.FOAM_FLAGS[f]);
         return foam.util.flagFilter(enabledFlags)(this);
       }
+    },
+    {
+      class: 'String',
+      name: 'extends',
```
kgrgreer, line 179: Isn't this a reserved word in Java?

PR #3871, opened 2020-08-28
Production Index File Jar Version Fix

nanopay-arthur:


jlhughes (approved):
Fine, that works for now.

PR #3879, opened 2020-08-28
[NP-1958] Fix wizardlet create args

KernelDeimos:
Danny and I were able to produce an error by adding sections to the data model of a wizardlet with a createView override. We found out `this` was being passed as the context in `wizardlet.cls_.create()` instead of `this.__subContext__`.

### DIFF
--- a/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
+++ b/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
@@ -42,8 +42,8 @@ foam.CLASS({
         if ( ! this.of ) return null;
 
         var ret = this.of.getAxiomByName('capability') ?
-          this.of.create({ capability: this.capability }, this) :
-          this.of.create({}, this);
+          this.of.create({ capability: this.capability }, this.__subContext__) :
+          this.of.create({}, this.__subContext__);
 
         if ( this.ucj === null ) return ret;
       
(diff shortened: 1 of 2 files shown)


src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -42,8 +42,8 @@ foam.CLASS({
         if ( ! this.of ) return null;
 
         var ret = this.of.getAxiomByName('capability') ?
-          this.of.create({ capability: this.capability }, this) :
-          this.of.create({}, this);
+          this.of.create({ capability: this.capability }, this.__subContext__) :
+          this.of.create({}, this.__subContext__);
```
kgrgreer, line 46: How is this different than just using 'this'?

KernelDeimos: Is it not meant to be? After I saw the error I thought I did that by accident and this was supposed to be a context rather than an element. If elements are meant to work as the second arg then this may be a deeper issue.

kgrgreer:
You can pass either an explicit context or an object as the second argument to create. If you pass an object, it takes its __subContext__, so the two versions should be identical.
I would rather get to the root cause of this bug than making a unknown change that works for some unknown reason, that if someone sees, they might be tempted to revert.

kgrgreer:
Are you sure that this fixes the problem. What 'of' is being used?

KernelDeimos:
This fix is no longer needed

PR #3880, opened 2020-08-29
Simplify formula that contains only constant operands

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-1941

## Changes
- Simplify formula that contains only constant operands


jlhughes (commented):
@lchanmann was Double missed, or are we missing similar partialEval on Long, int, ... ?

lchanmann:
Yeah @jlhughes , Long and Integer couldn't be casted to Double.

PR #3883, opened 2020-08-31
Np 1726/refactor unstyled table view changes with projection objects

olhabn:
- refactoring
- actions displayed
- building objs for FObjectProperty and Reference property in unstylesTableView to call tablleCellFormatter on them
- using obj's class to find correct property
- some small fixes


src/foam/mlang/mlang.js
```diff
@@ -2735,7 +2735,7 @@ foam.CLASS({
             Object    o   = ci.newInstance();
 
             for ( int j = 0 ; j < es.length ; j++ ) {
-              PropertyInfo e = (PropertyInfo) es[j];
+              PropertyInfo e = (PropertyInfo) es[j];//probably not all the expressions can be casted to PropInfo eg NestedPropExpr
```
olhabn: please take a look at comment

src/foam/nanos/column/CommonColumnHandler.js
```diff
@@ -1,8 +1,11 @@
 foam.CLASS({
   package: 'foam.nanos.column',
   name: 'CommonColumnHandler',
+  requires: [
+    'foam.nanos.column.ColumnConfigToPropertyConverter'
+  ],
   methods: [
-    function returnColumnNameForNLevelName(context, col, n) {
```
olhabn, line 5: just a small refactoring here

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -37,8 +37,32 @@ foam.CLASS({
     }
   ],
   methods: [
-    function set(o, val) {
-      o.cls_.getAxiomByName(this.nestedProperty).set(o, val);
+    {
```
olhabn, line 40: should we maybe add void set method to AbstractExpr ? or create some kind of PropertyBasedExpression and add method there?

src/foam/nanos/export/CSVTableExportDriver.js
```diff
@@ -9,7 +9,10 @@ foam.CLASS({
   name: 'CSVTableExportDriver',
   implements: [ 'foam.nanos.export.ExportDriver' ],
 
+  extends: 'foam.nanos.export.TableExportDriver',
+
```
olhabn, line 13: re-using code using TableExportDriver

src/foam/nanos/export/GoogleSheetsBasedExportDriver.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.export',
+  name: 'GoogleSheetsBasedExportDriver',
+  implements: [
```
olhabn: re-using code for GoogleSheetsExportDriver and PDFGoogleSheetsExportDriver

src/foam/u2/view/ColumnConfigView.js
```diff
@@ -37,10 +37,10 @@ foam.CLASS({
         //selectedColumnNames misleading name cause it may contain objects
         data.selectedColumnNames = data.selectedColumnNames.map(c => 
           {
```
olhabn, line 39: refactoring

src/foam/mlang/mlang.js
```diff
@@ -2751,7 +2751,7 @@ foam.CLASS({
     {
       name: 'put',
       code: function put(o, sub) {
-        var a = [o.cls_];
+        var a = [o.cls_.id];
```
olhabn, line 2833: or wouldn't it be better to do check for a[0] being instance of class on line 2717?

src/foam/nanos/column/NestedPropertiesExpression.js
```diff
@@ -37,8 +37,35 @@ foam.CLASS({
...
+          name: 'o',
+          type: 'Object'
+        },
+        {
+          name: 'value',
+          type: 'Object'
+        }
+      ],
+      code: function(o, val) {
+        if ( this.nestedProperty.includes('.') ) return;
```
olhabn, line 53: not building FObjectProperty and Reference property obj's here as if one of the properties in chain is reference property we will not be able to set object

src/foam/mlang/mlang.js
```diff
@@ -2714,7 +2714,7 @@ foam.CLASS({
       documentation: 'An array of full objects created from the projection. Only properties included in exprs/the-projection will be set.',
       factory: function() {
         return this.projectionWithClass.map(p => {
-          var o = foam.lookup(p[0]).create();
+          var o = foam.lookup(p[0]).create(null, this.__context__);
```
kgrgreer: You should be able to just use 'this' instead of this.__context__

src/foam/mlang/mlang.js
kgrgreer: No, but those are the only ones useful here because they need to have a set() method. Better to wrap the whole 'for' body in a try catch to exclude those that aren't PropertyInfo's without skipping over those that are.

src/foam/mlang/mlang.js
kgrgreer: I'm fine with this change, but was the old code causing a problem?

src/foam/nanos/column/NestedPropertiesExpression.js
kgrgreer: I don't think so. You wouldn't normally think of set() as being part of an Expression. Better to just catch ClassCastException or check if InstanceOf to make code safe. Or Make a Settable or Setter interface or SettableExpression.

src/foam/mlang/mlang.js
olhabn: I don't remember, but probably on line 2717. I changed line 2754 to keep it consistent with backend

src/foam/mlang/mlang.js

src/foam/mlang/mlang.js

src/foam/mlang/mlang.js
olhabn: but I can change 2754 instead

src/foam/core/Property.js
```diff
@@ -293,7 +293,11 @@ foam.CLASS({
       name: 'sortable',
       value: true
     },
-    'unitPropValueToString'
+    'unitPropValueToString',
+    {
+      name: 'dependsOnPropertiesWithNames',
```
kgrgreer, line 298: Please add documentation:.

olhabn: fixed

PR #3884, opened 2020-08-31
[NP-1958] Make sequence-invoked wizard more configurable

KernelDeimos:
Allows the following configurations in the wizard sequence:
- It is possible to choose between popup and view stack wizards (or, provide any alternative implementation)
- StepWizardAgent now accepts a config for the wizard controller, and a ViewSpec for the wizard view
- You can specify a value for `location.hash` to be set to after the wizard is displayed. It is also possible to display multiple views in a sequence with different location hashes by preceding each one with an appropriate configuration of `ConfigureFlowAgent`.

### DIFF
--- a/src/foam/u2/crunch/wizardflow/ConfigureFlowAgent.js
+++ b/src/foam/u2/crunch/wizardflow/ConfigureFlowAgent.js
@@ -1,0 +1,73 @@
+foam.CLASS({
+  package: 'foam.u2.crunch.wizardflow',
+  name: 'ConfigureFlowAgent',
+  implements: [ 'foam.core.ContextAgent' ],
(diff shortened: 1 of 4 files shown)


src/foam/u2/crunch/wizardflow/ConfigureFlowAgent.js
```diff
@@ -0,0 +1,73 @@
+foam.CLASS({
```
jlhughes, line 7: license header, please.

KernelDeimos: I just realized there's a script I can copy...

PR #3886, opened 2020-08-31
add support for different locale variation

hchoura:


src/foam/u2/Element.js
```diff
@@ -1638,7 +1638,13 @@ foam.CLASS({
             let d =  this.__subContext__.localeDAO;
             this.add(this.PromiseSlot.create({
               promise://TODO support more that one language (add language to the id)
-                d.where(expr.AND(expr.EQ(foam.i18n.Locale.LOCALE, foam.locale),expr.EQ(foam.i18n.Locale.ID, c.data.id+'.'+c.clsInfo))).select().then(function(a){
+                d.where(
+                  expr.AND(
+                    expr.OR(
+                      expr.EQ(foam.i18n.Locale.LOCALE, foam.locale),
+                      expr.EQ(foam.i18n.Locale.LOCALE, foam.locale.substring(0,foam.locale.indexOf('-')))),
```
jlhughes, line 1645: Should this order by the x-y form, so it's selected, if it exists, in the arr[0] ?

hchoura: In this solution, we can ask for en-CA, but if it did not exist, it will return the available en axiom.

PR #3888, opened 2020-09-01
approvalrequestutil.getState returning requested for rejected approvals

noodlemoodle:
ApprovalRequestUtil is returning status REQUESTED on the first REQUESTED it sees, and rejecting was not working if the request happens to show up later in the list returned by approvalrequestdao

### DIFF
--- a/src/foam/nanos/approval/ApprovalRequestUtil.java
+++ b/src/foam/nanos/approval/ApprovalRequestUtil.java
@@ -77,6 +77,7 @@ public class ApprovalRequestUtil {
     } catch (Exception e){
       return tester.getState();
     }
-    return tester.getState();
+    ApprovalStatus state = tester.getState();
+    return state == null ? ApprovalStatus.REQUESTED : ApprovalStatus.APPROVED;
   }
 }
--- a/src/foam/nanos/approval/ApprovalRequestUtil.java
+++ b/src/foam/nanos/approval/ApprovalRequestUtil.java
@@ -78,6 +78,6 @@ public class ApprovalRequestUtil {
       return tester.getState();
     }
     ApprovalStatus state = tester.getState();
-    return state == null ? ApprovalStatus.REQUESTED : ApprovalStatus.APPROVED;
+    return state == null ? ApprovalStatus.REQUESTED : state;
   }
 }
--- a/src/foam/nanos/approval/ApprovalTester.java
+++ b/src/foam/nanos/approval/ApprovalTester.java
@@ -59,8 +59,7 @@ public class ApprovalTester {
    * @return approval status of the test
    */
   public ApprovalStatus test(ApprovalRequest request) {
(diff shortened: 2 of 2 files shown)


src/foam/nanos/approval/ApprovalRequestUtil.java
```diff
@@ -77,6 +77,7 @@ public void put(Object obj, Detachable sub) {
     } catch (Exception e){
       return tester.getState();
     }
-    return tester.getState();
+    ApprovalStatus state = tester.getState();
+    return state == null ? ApprovalStatus.REQUESTED : state;
```
lchanmann, line 81: if the collection is empty it would return REQUESTED status, is it intended?

src/foam/nanos/approval/ApprovalTester.java
```diff
@@ -59,8 +59,7 @@
    * @return approval status of the test
    */
   public ApprovalStatus test(ApprovalRequest request) {
-    if ( request.getStatus() == ApprovalStatus.REQUESTED
-      || incrRejected(request) >= request.getRequiredRejectedPoints()
+    if ( incrRejected(request) >= request.getRequiredRejectedPoints()
```
lchanmann, line 62: would check `request.getStatus() == ApprovalStatus.REQUESTED` after the rejection check instead of removing
```suggestion
    if ( incrRejected(request) >= request.getRequiredRejectedPoints()
      || request.getStatus() == ApprovalStatus.REQUESTED
```

src/foam/nanos/approval/ApprovalRequestUtil.java
noodlemoodle: good point, probably not .. will change this, thanks

src/foam/nanos/approval/ApprovalTester.java
noodlemoodle: would this lead to the same result though? if the first approvalrequest is status requested, then `request.getStatus()` will still be returned

lchanmann: Well yeah, it's an OR operation the order doesn't matter. But this `test` method is to early return when it see a REQUESTED or adequate rejection point collected.

Since the ApprovalDAO would cleanup pending approval requests once rejected/approved, the ApprovalTester class doesn't expect the sequence [REQUESTED, ..., REJECTED] where you have a rejection and there remain pending approval requests before it.
https://github.com/foam-framework/foam2/blob/c09f7b86068250a7cad45bd27ee0a234709870e6/src/foam/nanos/approval/ApprovalDAO.js#L61

noodlemoodle: i see now.. actually the problem here is that the approvalrequest are different classifications so it is not getting cleaned up. But the dao i passed into getstate does not filter the classification. 
I will close this pr and re-fix, thanks!

PR #3890, opened 2020-09-01
Crunch Wizard Config Refinements

zahnno:
Replaced individual wizardlet isValid logic with validate requirements placed on wizard config.


src/foam/u2/crunch/wizardflow/StepWizardAgent.js
```diff
@@ -37,34 +38,17 @@ foam.CLASS({
...
   methods: [
     function execute() {
-      let topCap = this.capabilities.slice(-1)[0];
-      let topCapConfig = topCap.wizardletConfig.cls_.create({
-        ...topCap.wizardletConfig.instance_
-      }, this);
-      let config = this.overrideTopCapabilityConfig
-        ? topCapConfig.copyFrom(this.config)
-        : this.config.copyFrom(topCapConfig)
-        ;
```
KernelDeimos, line 60: This shouldn't be removed, but since it's currently broken I'll re-implement a solution after this is merged.

KernelDeimos (approved):

PR #3900, opened 2020-09-02
using dao.find() instead of findTargetId()

yij793:
address: https://nanopay.atlassian.net/browse/NP-1988
time# 4h

notes: if business registration capability is targetId then ucj.findTargetId(x) will return null due to availabilityPredicate( only available when current user isn't a business). ucj.findTargetId(getX()) will throw NullpointerException. If there are better ways to fix this issue, please comment it. thank you.

### DIFF
--- a/src/foam/nanos/crunch/ruler/SetUCJStatusOnPut.js
+++ b/src/foam/nanos/crunch/ruler/SetUCJStatusOnPut.js
@@ -17,6 +17,7 @@ foam.CLASS({
     'foam.core.X',
     'foam.dao.ArraySink',
     'foam.dao.DAO',
+    'foam.dao.ProxyDAO',
     'foam.nanos.crunch.Capability',
     'foam.nanos.crunch.CapabilityCapabilityJunction',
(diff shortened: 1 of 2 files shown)


kgrgreer:
I would like to look into this a bit.

src/foam/nanos/crunch/ruler/SetUCJStatusOnPut.js
```diff
@@ -41,7 +42,8 @@ foam.CLASS({
             // the following should be checked if the result of previous rule ( validateUCJDataOnPut ) 
             // is not ACTION_REQUIRED. In the ACTION_REQUIRED case, the ucj should be put into the
             // dao without any additional checks
-            Capability capability = (Capability) ucj.findTargetId(x);
+            DAO capabilityDAO = (DAO) x.get("capabilityDAO");
+            Capability capability = (Capability) ucj.findTargetId(((ProxyDAO) capabilityDAO).getX());
```
kgrgreer: Could you please test and let me know if this would also work?
  Capability capability = (Capability) ucj.findTargetId(ruler.getX());

Thanks

PR #3908, opened 2020-09-03
on output, test for nulls first.

jlhughes:
on date output, test for null date before calling getTime.


src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -260,6 +262,7 @@ public void output(Object value) {
     } else if ( value instanceof List ) {
       output((List) value);
     } else /*if ( value == null )*/ {
```
jlhughes:
```suggestion
    } else {
```

PR #3911, opened 2020-09-03
[NP-2006] Add service to access files without authentication

KolombetM:
https://nanopay.atlassian.net/browse/NP-2006
@jlhughes

### DIFF
--- a/src/services
+++ b/src/services
@@ -521,10 +521,10 @@ p({
 
 p({
   "class": "foam.nanos.boot.NSpec",
-  "name": "unauthenticatedHttpFileService",
+  "name": "file",
   "authenticate": false,
   "serviceScript": """
-    fileService = new foam.nanos.http.UnauthenticatedFileService(x, "unauthenticatedHttpFileService", x.get("blobStore"));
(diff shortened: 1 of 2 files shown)


src/services
```diff
@@ -515,6 +519,16 @@ p({
   """
 })
 
+p({
+  "class": "foam.nanos.boot.NSpec",
+  "name": "unauthenticatedHttpFileService",
```
jlhughes: I would like to register this under just 'file'. 
unauthenticated is correct but thinking of the email templates and other links I would prefer
service/file over server/unauth....

kgrgreer: Or maybe something like "/pub". Thoughts @jlhughes ?

jlhughes: so longer term plans were to intercept /images and such in the http server and process them.  so eventually /images /files would be routed that way. 
or are you taking about /service/pub ?  but that would suggest anything public, or are suggesting /pub in place of service for anything unauthenticated - ok, yes, this is what you are talking about.    That would be fine, but we still need to add intercept of /images in the servlet, but it could then route/delegate through /pub.    
let's use /service/file today and setup /pub and the servlet intercept in another PR - as we need this support immediately.

PR #3913, opened 2020-09-03
NP-1967 Move UCJ.read Permission to Foam

carl-zzz:
Add UCJ.read permission to foam

Related Nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/10534


jlhughes:
Now correct, but obviously others should move. 
@carl-zzz can you create a ticket to capture moving the others.

PR #3916, opened 2020-09-04
Update SupportFileTypeDAO

KolombetM:
@yij793 
@jlhughes


jlhughes (approved):
Yeah, that's important!

src/foam/nanos/fs/SupportFileTypeDAO.js
```diff
@@ -45,7 +45,7 @@ foam.CLASS({
       name: 'put_',
       javaCode: `
       if (!(obj instanceof File)) {
-        return super.put_(x, obj);
+        return file;
```
jlhughes:
```suggestion
        return obj;
```

PR #3918, opened 2020-09-08
fix npe on isdataupdate predicate

noodlemoodle:


src/foam/nanos/crunch/ruler/IsDataUpdate.js
```diff
@@ -38,7 +38,10 @@ foam.CLASS({
           for ( foam.core.PropertyInfo prop : props ) {
             if ( prop.getStorageTransient() ) 
               continue;
-            if ( ! prop.f(oldData).equals(prop.f(newData)) ) 
+            Object newProp = prop.f(newData);
```
jlhughes: This will return true when a StorageTransient=false, StorageOptional=true property is the only remaining property (like LastModified).
An option is:
```
FObectFormatter formatter = new JSONFObjectFormatter();
formatter.setX(x);
formatter.setPropertyPredicate(new StoragePropertyPredicate());
if ( old != null ) {
  return formatter.outputDelta(old, nu).builder().length() > 0;
} else {
  return formatter.output(nu).builder().length() > 0;
}

```
Of course this should be ThreadLocal if possible.

jlhughes (approved):

PR #3920, opened 2020-09-08
[NP-2023] Tweak MinMax selection UI

KernelDeimos:
- Adds better help text
- Tweaks alignment and padding


src/foam/u2/view/MultiChoiceView.js
```diff
@@ -151,6 +166,22 @@ foam.CLASS({
       value: function(obj) {
         return [ obj.id, obj.toSummary() ];
       }
+    },
+    {
+      name: 'helpText_',
```
jlhughes, line 171: Please see my logic diagram from the ticket. 
But this will do for now.

jlhughes (approved):

PR #3923, opened 2020-09-09
added a null check

minsunny:
https://nanopay.atlassian.net/browse/NP-1989

### DIFF
--- a/src/foam/lib/json/Util.java
+++ b/src/foam/lib/json/Util.java
@@ -23,7 +23,7 @@ public class Util {
     int start = 0;
     int len = 0;
 
-    if ( src != null )
+    if ( src != null && ! src.equals("") )
       len = src.length();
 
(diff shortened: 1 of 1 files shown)


src/foam/lib/json/Util.java
```diff
@@ -21,7 +21,11 @@ protected static int copy(int start, int end, String src, StringBuilder dst) {
   /** Append src string to dst StringBuilder as an escaped JSON string. **/
   public static void escape(String src, StringBuilder dst) {
     int start = 0;
-    int len   = src.length();
+    int len = 0;
+
+    if ( src != null && ! src.equals("") )
```
jlhughes:
```suggestion
    if ( ! foam.util.SafetyUtil.isEmpty(src) )
```

jlhughes (approved):

PR #3924, opened 2020-09-09
[NP-2043] Support DAO interface and rules for payloads in Capable objects

KernelDeimos:
Adds a class `CapableAdapterDAO` which adapts any Capable to the DAO interface. Also adds a method to server-side Capable objects called `getCapablePayloadDAO` which returns a CapableAdapterDAO decorated with RulerDAO.


src/foam/nanos/crunch/lite/CapableAdapterDAO.js
```diff
@@ -0,0 +1,70 @@
...
+    },
+    {
+      name: 'select_',
+      javaCode: `
+        // TODO
+        throw new RuntimeException("TODO");
+      `
+    }
+  ]
+});
```
tharmaman, line 76:
```suggestion
});

```

src/foam/nanos/ruler/RulerDAO.js
```diff
@@ -310,6 +310,8 @@ for ( Object key : groups.getGroupKeys() ) {
       setX(x);
       setDelegate(delegate);
       setDaoKey(serviceName);
+      // This doesn't get called when using Builder,
```
tharmaman, line 313: lol i meant put this in capable when we call the constructor but I guess this works too

src/foam/nanos/crunch/lite/CapableAdapterDAO.js
```diff
@@ -0,0 +1,70 @@
+foam.CLASS({
```
jlhughes, line 7: header please.

src/foam/nanos/crunch/lite/Capable.js
```diff
@@ -93,6 +95,24 @@ foam.INTERFACE({
...
+          visibility: 'default',
+          args: [
+            { name: 'x', type: 'X' }
+          ],
+          body: `
+            DAO capableDAO = new CapableAdapterDAO.Builder(x)
+              .setCapable(this)
+              .setOf(CapablePayload.getOwnClassInfo())
+              .build();
+            x = x.put("capableDAO", capableDAO);
```
tharmaman, line 110: Do we need this still? It worked on mine without it unless it should be there.

tharmaman (approved):

src/foam/nanos/crunch/lite/CapableAdapterDAO.js
```diff
@@ -0,0 +1,70 @@
...
+  `,
+
+  properties: [
+    {
+      name: 'capable',
+      class: 'FObjectProperty',
+      of: 'foam.nanos.crunch.lite.Capable'
+    },
+    // {
+    //   name: 'of',
```
tharmaman, line 27: guess we also don't neeed the of as well since that wasn't the problem for  thee NPEs before?

tharmaman (approved):

src/foam/nanos/crunch/lite/Capable.js
KernelDeimos: I left it there by accident, but this is more future-proof since RulerDAO takes a daoKey so maybe we should leave it

src/foam/nanos/crunch/lite/CapableAdapterDAO.js
KernelDeimos: Yeah, also it's set in the code that constructs this now

PR #3925, opened 2020-09-10
Static validate

nick-prat:


src/foam/util/SafetyUtil.java
```diff
@@ -207,4 +209,11 @@ public static String trim(String s) {
     return s == null ? null : s.trim();
   }
 
+  public static void Validate(X x, Validatable v) {
```
kgrgreer: Rename to 'validate'.

nick-prat: right

PR #3928, opened 2020-09-10
Made StepWizardView scale with the whole wizard view on larger screen

nick-prat:
Changes Step Wizard view from Screen Shot 2020-09-10 at 3 56 41 PM to
Screen Shot 2020-09-10 at 4 17 47 PM


jlhughes (dismissed):

jlhughes:
If you have a really wide monitor, and your browser expanded, this may become unusable. 
Thoughts?  we could make the max a little bigger. There is probably a natural width for the three columns. 
@kgrgreer thoughts.

PR #3933, opened 2020-09-11
Np 1991/read many to many relationship property view changes

olhabn:
@kgrgreer @jlhughes  please review


olhabn:
something wrong with displaying Many-to-Many relationship

PR #3949, opened 2020-09-14
Prevent infinite recursion on StepWizardLetController

zahnno:
The for loop can lead to an invite recursion relating to the wizardposition's sectionindex

### DIFF
--- a/src/foam/u2/wizard/StepWizardletController.js
+++ b/src/foam/u2/wizard/StepWizardletController.js
@@ -186,6 +186,8 @@ foam.CLASS({
             ) {
             return p;
           }
+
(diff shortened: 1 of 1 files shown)


src/foam/u2/wizard/StepWizardletController.js
```diff
@@ -186,6 +186,8 @@ foam.CLASS({
             ) {
             return p;
           }
+
+          if ( p.sectionIndex < -1 ) return null;
```
KernelDeimos, line 190: Does this still work with `< 0`?

KernelDeimos (approved):

PR #3952, opened 2020-09-15
make obj displayed in view ref configurable

noodlemoodle:
Repurpose refObj (no longer used prop) for the view ref action, so that the rule creating the approvalrequest can configure the referenced obj when it doesnt want to display the default
for example in ucj approvalrequest for business onboarding, the ref obj can be set to the business

used for task  : "Discussion about top level Capability in an Approval Request - has no Data."

usage: https://github.com/nanoPayinc/NANOPAY/pull/10591


jlhughes (dismissed):

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -683,9 +682,17 @@
              return;
         }
 
-        var objId = X[self.daoKey_].of.ID.type === 'Long' ? parseInt(this.objId) : this.objId;
-
-        return X[this.daoKey_]
+        var objId, daoKey;
+        if ( self.refObj ) {
+          var res = self.refObj.split(':');
```
jlhughes: I don't like this. 
what about a 
refDaoKey like causeDaoKey. 
then
refDaoKey
refObjId
and refDaoKey can fallback to daoKey if not provided.

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -683,9 +682,17 @@
...
+          daoKey = res[0];
+          objId = X[daoKey].of.ID.type === 'Long' ? parseInt(res[1]) : res[1];
+        } else {
+          objId = X[self.daoKey_].of.ID.type === 'Long' ? parseInt(self.objId) : self.objId;
```
kgrgreer: More generically, this could be done as X[self.daoKey_].of.ID.adapt(null, self.objId), then it would work for any type.

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -683,9 +682,17 @@
...
         }
 
-        var objId = X[self.daoKey_].of.ID.type === 'Long' ? parseInt(this.objId) : this.objId;
-
-        return X[this.daoKey_]
+        var objId, daoKey;
+        if ( self.refObj ) {
+          var res = self.refObj.split(':');
+          daoKey = res[0];
+          objId = X[daoKey].of.ID.type === 'Long' ? parseInt(res[1]) : res[1];
```
kgrgreer: Same here, as below, showing I do code reviews from the bottom up :(.

src/foam/nanos/approval/ApprovalRequest.js
noodlemoodle: sounds good, i will update, thanks

src/foam/nanos/approval/ApprovalRequest.js
noodlemoodle: ok will address, thanks

src/foam/nanos/approval/ApprovalRequest.js
noodlemoodle: (: ʞo

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -668,9 +682,23 @@
             })
             .catch((err) => {
               console.warn(err.message || err);
-              return false;
+              if ( self.refObjId && self.refDaoKey && self.__subContext__[self.refDaoKey_] ) {
+                property = self.__subContext__[self.refDaoKey].of.ID;
+                objId = property.adapt.call(property, self.refObjId, self.refObjId, property);
+                return self.__subContext__[self.refDaoKey]
+                  .find(objId)
+                  .then((obj) => {
```
kgrgreer: Does it still work without this block?

kgrgreer: .then((obj) => { return !! obj; }), I mean, which could also have been written as just:
.then(obj => !!obj)

noodlemoodle: thanks, updating to `.then(obj => !! obj)` works

jlhughes (approved):

PR #3954, opened 2020-09-16
[DO NOT MERGE] Np 1764/issues with forgot password

olhabn:
@kgrgreer @jlhughes @blakegreen32 @Nauna please let me know what you think regarding comments below

ticket: https://nanopay.atlassian.net/browse/NP-1764

the thing is that there is no user when we resetting password so i have a couple of questions regarding that. 
an alternative is "Success password reset" view

### DIFF
--- a/src/foam/nanos/controller/ApplicationController.js
+++ b/src/foam/nanos/controller/ApplicationController.js
@@ -329,6 +329,23 @@ foam.CLASS({
       this.clientPromise.then(async function(client) {
         self.setPrivate_('__subContext__', client.__subContext__);
 
+        var userNotificationQueryId = self.subject && self.subject.realUser ?
+        self.subject.realUser.id : self.user.id;
+        self.__subContext__.notificationDAO.where(
+          self.EQ(this.Notification.USER_ID, userNotificationQueryId)
+        ).on.put.sub((sub, on, put, obj) => {
+          if ( obj.toastState == self.ToastState.REQUESTED ) {
+            self.add(self.NotificationMessage.create({
+              message: obj.toastMessage,
+              type: obj.severity,
+              description: obj.toastSubMessage
+            }));
+            var clonedNotification = obj.clone();
+            clonedNotification.toastState = self.ToastState.DISPLAYED;
+            self.__subContext__.notificationDAO.put(clonedNotification);
+          }
+        });
+
         await self.fetchSubject();
 
         // add user and agent for backward compatibility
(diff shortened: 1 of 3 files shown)


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -364,7 +364,7 @@ foam.CLASS({
 
       var userNotificationQueryId = this.subject && this.subject.realUser ?
       this.subject.realUser.id : this.user.id;
-
+      //is it called?
```
olhabn, line 384: is` src/foam/nanos/controller/ApplicationController.js initE` method or `nanopay/src/net/nanopay/ui/Controller.js onUserAgentAndGroupLoaded` method 
called for not authorized user?

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -575,6 +575,22 @@ foam.CLASS({
     },
 
     function notify(toastMessage, toastSubMessage, severity, transient) {
+      var userNotificationQueryId = this.subject && this.subject.realUser ?
```
olhabn, line 578: it obviously shouldn't be there but not sure where it should be

src/foam/nanos/notification/Notification.js
```diff
@@ -11,7 +11,7 @@ foam.CLASS({
   implements: [
     'foam.nanos.auth.Authorizable',
     'foam.nanos.auth.CreatedAware',
-    'foam.nanos.auth.CreatedByAware'
+    // 'foam.nanos.auth.CreatedByAware'
   ],
```
olhabn, line 15: what if there is no user?

src/foam/nanos/notification/services
```diff
@@ -28,11 +28,12 @@ p({
...
   "name": "notificationDAO",
   "lazy": false,
+  "authenticate": false,
   "serve": true,
   "serviceScript": """
     return new foam.dao.EasyDAO.Builder(x)
       .setOf(foam.nanos.notification.Notification.getOwnClassInfo())
-      .setAuthorize(true)
+      .setAuthorize(false)
       .setInnerDAO((foam.dao.DAO) x.get("localNotificationDAO"))
```
olhabn, line 37: what if there is no user?

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -329,6 +329,23 @@ foam.CLASS({
       this.clientPromise.then(async function(client) {
         self.setPrivate_('__subContext__', client.__subContext__);
```
olhabn, line 331: can we add subscription here?

jlhughes:
Resolved elsewhere.

PR #3956, opened 2020-09-16
Ruler on rule

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2078

## Changes
- Support setting ruler on localRuleDAO
- Add PropertyIsSetPredicate for checking if a property is set
- Add EnsurePropertyOnCreateRule for setting object property if unset on after create

### DIFF
--- a/src/foam/nanos/auth/ruler/EnsurePropertyOnCreateRule.js
+++ b/src/foam/nanos/auth/ruler/EnsurePropertyOnCreateRule.js
@@ -15,9 +15,8 @@ foam.CLASS({
     The rule is run after the object is put/created.
 
     The predicate of EnsurePropertyOnCreateRule will check for
-    1. NEW_OBJ.spid == rule.spid
-    2. NEW_OBJ is an instance of the "targetClass"
-    3. The "propName" property on NEW_OBJ is not set
+    1. NEW_OBJ is an instance of the "targetClass"
+    2. The "propName" property on NEW_OBJ is not set
 
     Then EnsurePropertyOnCreateRuleAction will set the object "propName"
     property the value of the "propValue" and re-put the object back to the DAO.
@@ -63,10 +62,6 @@ foam.CLASS({
       transient: true,
       visibility: 'HIDDEN',
       javaGetter: `
-        var spidPredicate = new PropertyEQValue();
-        spidPredicate.setPropName("spid");
-        spidPredicate.setPropValue(getSpid());
-
         var isInstanceOfPredicate = new IsInstancePredicate();
         isInstanceOfPredicate.setOf(getTargetClass());
 
@@ -74,7 +69,6 @@ foam.CLASS({
         propertyIsSetPredicate.setPropName(getPropName());
 
         return AND(
-          spidPredicate,
           isInstanceOfPredicate,
           NOT(propertyIsSetPredicate)
         );
--- a/src/foam/nanos/auth/ruler/EnsurePropertyOnCreateRule.js
+++ b/src/foam/nanos/auth/ruler/EnsurePropertyOnCreateRule.js
@@ -1,0 +1,145 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.auth.ruler',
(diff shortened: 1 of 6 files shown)


src/foam/nanos/auth/ruler/EnsurePropertyOnCreateRule.js
```diff
@@ -0,0 +1,145 @@
...
+        spidPredicate.setPropValue(getSpid());
+
+        var isInstanceOfPredicate = new IsInstancePredicate();
+        isInstanceOfPredicate.setOf(getTargetClass());
+
+        var propertyIsSetPredicate = new PropertyIsSetPredicate();
+        propertyIsSetPredicate.setPropName(getPropName());
+
+        return AND(
+          spidPredicate,
```
jlhughes: What happens when the object is not ServiceProviderAware? 
You could use the ServiceProvderAwarePredicate.

lchanmann: When object is not ServiceProviderAware the arg1 will be null since it doesn't have spid property and the arg2 is the spid of the current rule and is non-null therefore the rule predicate return false.

src/foam/nanos/ruler/predicate/PropertyIsSetPredicate.js
```diff
@@ -0,0 +1,46 @@
...
+  ],
+
+  properties: [
+    {
+      class: 'String',
+      name: 'propName'
+    },
+    {
+      class: 'Boolean',
+      name: 'isNew',
```
kgrgreer, line 27: Please document the purpose of this property.

src/foam/nanos/auth/ruler/EnsurePropertyOnCreateRule.js
lchanmann: ServiceProviderAwarePredicate will also check if the current user has permission to access the object which is what we don't want here. 

For example, there is rule1 which updates feeAccount of TransactionFeeRule if not set. We want rule1 to only apply to TransactionFeeRule with the same spid and not apply to TransactionFeeRule in other spids that user has access to.

jlhughes: I don't understand.

src/foam/nanos/ruler/predicate/PropertyIsSetPredicate.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/3956/commits/7ece73e7bada754b5e6fdb6a44d8eaef5041c2a3.

src/foam/nanos/auth/ruler/EnsurePropertyOnCreateRule.js
lchanmann: Well, ServiceProviderAwarePredicate will not check the current user permission if we explicitly set spid on the predicate. However, it still needs propertyInfos map which we don't have when constructing the predicate here.

jlhughes: Is this used to support spid specific rules?  And do we have to do this in all rules?

lchanmann: EnsurePropertyOnCreateRule relies on spid.
```
// Auto-fill feeAccount for TransactionFeeRule on create
p({
  class: "foam.nanos.auth.ruler.EnsurePropertyOnCreateRule",
  id: "2e92ac64-e331-4152-a076-682b0c981c70",
  name: "Auto-fill TransactionFeeRule feeAccount",
  daoKey: "localRuleDAO",
  ruleGroup: "TransactionFeeRule",
  targetClass: "net.nanopay.tx.fee.TransactionFeeRule",
  propName: "feeAccount",
  propValue: 2,
  spid: "treviso",
  lifecycleState: 1
})
```
So this rule will update TransactionFeeRule.feeAccount to 2 if feeAccount is not set when the TransactionFeeRule is created in treviso spid.

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/3956/commits/b893a287973bf560570cf2b0a5ae545ff7fe450f.

kgrgreer:
LGTM

jlhughes (approved):

PR #3957, opened 2020-09-16
removing CapabilityInterceptView

olhabn:
@Nauna @KernelDeimos @tharmaman 

please review


src/foam/u2/crunch/CrunchController.js
```diff
@@ -124,14 +123,7 @@ foam.CLASS({
       }
       // Register intercept for later occurances of the check above
       this.activeIntercepts.push(intercept);
-      // Pop up the popup
```
olhabn, line 127: let me know if somebody needs it and it's better to comment it out

PR #3963, opened 2020-09-17
Forgot password notification

hchoura:


Nauna (approved):

minsunny (approved):

src/foam/nanos/auth/RetrievePassword.js
```diff
@@ -64,11 +65,17 @@ foam.CLASS({
       code: function(X) {
         const user = this.User.create({ email: this.email });
         this.resetPasswordToken.generateToken(null, user).then((_) => {
-          this.notify(`${this.INSTRUC_ONE} ${this.email}. ${this.INSTRUC_TWO}`, '', this.LogLevel.INFO, true);
+          this.ctrl.add(this.NotificationMessage.create({
```
kgrgreer, line 68: The ctrl exports a notify method, which is what I think we should be using. Can't we just fix the notify method?

kgrgreer (changes requested):
I disagree with this approach. I think the old code is correct and we should be calling notify(). This change is too tightly coupled to 'ctrl'.

src/foam/nanos/auth/RetrievePassword.js
jlhughes: Yes, this.notify should work.
Also, the 'true - transient' property should be dropped so that this request is recorded for the user.

jlhughes: Talking with @hchoura this is the case where no user in the context.  So suspect doNotify ignores the request when no user in context.

jlhughes: Using the controller directly is standard practice for all views where the user is not logged in.  
We'll create a ticket to have 'notify' support this scenario.

jlhughes (approved):

PR #3966, opened 2020-09-18
Fix mobile problem

hchoura:
the mobile app did not use locale storage


smirnovaka (approved):

smirnovaka:
maybe you should add some warning saying that this is not the normal case

jlhughes (approved):

PR #3967, opened 2020-09-18
Enabling MinMaxCapability in CapabilityPayloadDAO

nanoscott:
This PR is required for: https://github.com/nanoPayinc/NANOPAY/pull/10527


src/foam/nanos/crunch/Capability.js
```diff
@@ -371,12 +371,16 @@ foam.CLASS({
...
+        Subject currentSubject = (Subject) x.get("subject");
 
         Subject subject = new Subject(x);
-        subject.setUser((User) userDAO.find(ucj.getSourceId()));
         if ( ucj instanceof AgentCapabilityJunction ) {
+          subject.setUser((User) userDAO.find(ucj.getSourceId()));
           AgentCapabilityJunction acj = (AgentCapabilityJunction) ucj;
           subject.setUser((User) userDAO.find(acj.getEffectiveUser())); // "user"
+        } else {
+          subject.setUser(currentSubject.getRealUser());
```
nanoscott, line 351: @noodlemoodle .. I need this to work in order to make my capability heirarchy to properly resolve prerequististes between businesses and users.

Specifically, I have a Business Signing Officer capability assigned to the business and a User Signing Officer capability assigned to the user and the User Signing Officer capability is a prerequisite of the Business Signing Officer capability.

Setting this subject is the only way to get the getJunctionForSubject method to work properly.

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
```diff
@@ -30,6 +30,7 @@ foam.CLASS({
     'java.util.HashMap',
     'java.util.List',
     'java.util.Map',
+    'java.util.TreeMap',
```
nanoscott, line 36: using a TreeMap so that the results are sorted and easier to read by map key.

nanoscott:
@KernelDeimos @noodlemoodle .. please review

src/foam/nanos/crunch/Capability.js
noodlemoodle: but the subject here does not necessarily contain the correct realUser right? it could be admin or someone approving the cap

nanoscott: @noodlemoodle .. we usually keep track of this kind of stuff in the subject itself, but since you are using it already, we need to somehow maintain that state if/when necessary. Having a condition checking to see if the current subject is the business should be enough to ensure that real user should be maintained when checking the junction for the subject.

noodlemoodle: ok, makes sense to me, thanks

PR #3972, opened 2020-09-21
prevent update to spid prop on user

noodlemoodle:
follow up of PreventMultipleSpidDAO that stops creation of User-ServiceProvider junctions if the user already has a ServiceProvider junction, this prevents the spid prop on user from being updated once set, to avoid differences in user's ServiceProvider capability and ServiceProviderAware reference


jlhughes (commented):
Hitherto only a super admin could affect this change.  This would preclude that as well.  Let me find it.

PR #3974, opened 2020-09-21
StepWizardletStepsView capabilities name translation

olhabn:
temp changes:
I was trying to fix “capability” name not being translated in StepWizardletStepsView. This is not a good solution as probably not for all wizardlets data is capability but I don’t have any other suggestions

### DIFF
--- a/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
+++ b/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
@@ -9,7 +9,8 @@ foam.CLASS({
   extends: 'foam.u2.wizard.BaseWizardlet',
 
   imports: [
-    'crunchController'
+    'crunchController',
+    'localeDAO'
   ],
 
   properties: [
@@ -50,8 +51,24 @@ foam.CLASS({
       name: 'title',
       class: 'String',
       expression: function(capability) {
-        if ( ! capability || ! capability.name ) return '';
-        return capability.name;
+        var expr = foam.mlang.Expressions.create();
+        return this.localeDAO.where(
+          expr.AND(
+            expr.OR(
+              expr.EQ(foam.i18n.Locale.LOCALE, foam.locale),
+              expr.EQ(foam.i18n.Locale.LOCALE, foam.locale.substring(0,foam.locale.indexOf('-')))),
+            expr.EQ(foam.i18n.Locale.ID, capability.id + '.name')))
+        .select().then(function(a){
+          let arr = a.array;
+          if ( arr.length > 0 ) {
+            let ea = arr[0];
+            return ea.target;
+          }
+          return capability.name;
+        })
+        .catch(function() {
+          return capability.name;
+        });
       }
     }
   ],
--- a/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
+++ b/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
@@ -21,23 +21,24 @@ foam.CLASS({
     // Properties specific to CapabilityWizardSection
     {
       name: 'capability',
-      postSet: function(capability) {
+      postSet: function() {
+        var self = this;
         return this.localeDAO.where(
           this.AND(
             this.OR(
               this.EQ(foam.i18n.Locale.LOCALE, foam.locale),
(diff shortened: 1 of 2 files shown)


jlhughes (approved):

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -50,8 +51,24 @@ foam.CLASS({
...
+              expr.EQ(foam.i18n.Locale.LOCALE, foam.locale),
+              expr.EQ(foam.i18n.Locale.LOCALE, foam.locale.substring(0,foam.locale.indexOf('-')))),
+            expr.EQ(foam.i18n.Locale.ID, capability.id + '.name')))
+        .select().then(function(a){
+          let arr = a.array;
+          if ( arr.length > 0 ) {
+            let ea = arr[0];
+            return ea.target;
+          }
+          return capability.name;
```
Nauna: since this is in a then function - I think it's safer to:
```suggestion
          this.title = capability.name;
```

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -50,8 +51,24 @@ foam.CLASS({
...
+        })
+        .catch(function() {
+          return capability.name;
```
Nauna: same as above

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -50,8 +51,24 @@ foam.CLASS({
       name: 'title',
       class: 'String',
       expression: function(capability) {
-        if ( ! capability || ! capability.name ) return '';
-        return capability.name;
+        var expr = foam.mlang.Expressions.create();
```
kgrgreer: Easier to do implements: [ ''foam.mlang.Expressions' ], then just do this. instead of expr.

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
kgrgreer: This is technically broken either way. Just returning the value doesn't work because expressions need to return the value directly, not a promise, which is what happens when you have an async DAO call. If you do what Anna says, it will work better, but it won't be an expression anymore because if capability changes again then it won't regenerate since you've explicitly set the value, which probably isn't an issue in this case because you're probably only setting the capability once, in which case, you could just set the postSet on Capability to set the title.

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
olhabn: fixed

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
kgrgreer: Or better, just use a factory.

olhabn: noted and fixed

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
olhabn: noted and fixed

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
olhabn: > Or better, just use a factory.

for some reason it doesn't work

src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -8,14 +8,39 @@ foam.CLASS({
...
+    'localeDAO'
   ],
 
   properties: [
     // Properties specific to CapabilityWizardSection
     {
-      name: 'capability'
+      name: 'capability',
+      postSet: function() {
+        var self = this;
```
kgrgreer, line 25: Do users ever set the title explicitly? If so, maybe we should do something like:
if ( this.title ) return;
So that we don't overwrite the explicitly specified capability.

olhabn: I don't think so, factory function is called but when displayed title is empty

PR #3979, opened 2020-09-23
Fix create ucj on spid set

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2120

## Changes
- Fix blank grant path due to filterGrantedUCJ flag
- Use capability as permission prefix for serviceProviderDAO

### DIFF
--- a/src/foam/nanos/auth/CreateUserCapabilityJunctionOnSpidSet.js
+++ b/src/foam/nanos/auth/CreateUserCapabilityJunctionOnSpidSet.js
@@ -61,7 +61,7 @@ foam.CLASS({
 
             ServiceProvider sp = (ServiceProvider) user.findSpid(x);
             if ( sp == null ) {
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/CreateUserCapabilityJunctionOnSpidSet.js
```diff
@@ -61,12 +61,12 @@ foam.CLASS({
 
             ServiceProvider sp = (ServiceProvider) user.findSpid(x);
             if ( sp == null ) {
-              logger.warning("Cannot find capability for service provider : ", sp.getId());
+              logger.warning("Cannot find capability for service provider : ", spid);
```
jlhughes:
```suggestion
              logger.error("Cannot find capability for service provider : ", spid);
```

noodlemoodle (approved):

PR #3982, opened 2020-09-23
Fix CapabilityCategory json serialization

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2139

## Changes
- Mark "defaultAuthorizer" property as transient


jlhughes (approved):
This has been puzzling me for some time, thanks.

PR #3983, opened 2020-09-24
[NP-2042] File-preview

KolombetM:
https://nanopay.atlassian.net/browse/NP-2042

Fix bug when image is not fit iframe


jlhughes (approved):

jlhughes:
@Mykola97 more future/feature - on a device in portrait mode, can the css be such that the image will show below?

KolombetM:
> @Mykola97 more future/feature - on a device in portrait mode, can the css be such that the image will show below?

Yes. It can be

PR #3989, opened 2020-09-25
some additions

olhabn:
performance screen clean and cleanAll buttons fixed


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -253,6 +253,10 @@ foam.CLASS({
           return foam.nanos.column.ColumnConfigToPropertyConverter.create();
         return this.__context__.columnConfigToPropertyConverter;
       }
+    },
+    {
+      name: 'onDAOUpdate',
```
kgrgreer: Perhaps a better choice of name. Normally I would expect something named onXYZ to be a callback/listener.

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -411,7 +418,7 @@ foam.CLASS({
 
           //with this code error created  slot.get cause promise return
           //FIX ME
-          return this.slot(function(data, data$delegate, order, updateValues) {
+          return this.slot(function(data, data$delegate, onDAOUpdate, order, updateValues) {
```
kgrgreer: Is there a better way to do this? Using data$proxy instead, maybe?

src/foam/u2/view/UnstyledTableView.js
olhabn: renamed

src/foam/u2/view/UnstyledTableView.js
olhabn: hmmm, probably, I don't have any suggestions as for now though. for some reason data$proxy doesn't work

olhabn: we're using this approach in TableView

src/foam/u2/view/ReciprocalSearch.js
```diff
@@ -150,6 +150,10 @@ foam.CLASS({
       var self = this;
 
       this.dao.on.sub(this.updateTotalCount);
+      this.dao.on.sub(function() {
```
olhabn, line 153: selectedCount wasn't updated on clearAll so I thought maybe this change will make some sense

PR #3992, opened 2020-09-28
Refactor PMLogger.java to javascript model.

jlhughes:
- Create Interface for PMLogger
- Move pm cron and services setup to package journal files.
- made PM.combine public - I call it while creating PMs for cluster hops
- set logger to stdoutlogger - so no need to test for null
- add a return;   on recursion.

### DIFF
--- a/src/foam/nanos/boot/NSpecFactory.java
+++ b/src/foam/nanos/boot/NSpecFactory.java
@@ -10,6 +10,7 @@ import foam.core.*;
 import foam.dao.ProxyDAO;
 import foam.nanos.*;
 import foam.nanos.logger.Logger;
+import foam.nanos.logger.StdoutLogger;
 import foam.nanos.pm.PM;
 import foam.util.SafetyUtil;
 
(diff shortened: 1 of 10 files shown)


KolombetM (approved):

src/foam/nanos/boot/NSpecFactory.java
```diff
@@ -27,42 +28,52 @@ public NSpecFactory(ProxyX x, NSpec spec) {
...
-      if ( logger != null ) logger.info("Creating Service", spec_.getName());
+      logger.info("Creating Service", spec_.getName());
       ns_ = spec_.createService(x_.getX().put(NSpec.class, spec_));
       Object ns = ns_;
       while ( ns != null ) {
-        if ( ns instanceof ContextAware ) ((ContextAware) ns).setX(x_.getX());
-        if ( ns instanceof NSpecAware )   ((NSpecAware) ns).setNSpec(spec_);
-        if ( ns instanceof NanoService )  ((NanoService) ns).start();
+        if ( ns instanceof ContextAware ) {
+          // REVIEW: System is presently dependent on this blind setX call. - Joel
```
kgrgreer, line 54: what do you want to review?

src/foam/nanos/pm/PM.js
```diff
@@ -84,7 +84,9 @@ foam.CLASS({
       javaCode: `
     if ( x == null ) return;
     if ( getIsError() ) return;
-    setEndTime(new java.util.Date());
+    if ( getEndTime() == null ) {
+      setEndTime(new java.util.Date());
```
kgrgreer, line 88: This reminds me that using a Date is much more expensive that using a long timestamp in millieseconds, BTW.

PR #3997, opened 2020-09-29
cap store fix

noodlemoodle:
fix an npe in IN fix, and also a capabilitystore query

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -1507,9 +1507,7 @@ foam.CLASS({
         StringBuilder b = new StringBuilder();
 
         Object[] a = null;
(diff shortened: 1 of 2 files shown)


lchanmann:
@noodlemoodle I also have a fix for the NPE in IN - https://github.com/foam-framework/foam2/pull/3998.

noodlemoodle:
@lchanmann lgtm, i can undo my fix and just leave the one for capa store, just left one comment

kgrgreer (changes requested):
I reject this PR. CONTAINS cannot be indexed but IN can. If IN has a bug then we should fix it, not switch to CONTAINS.

PR #3998, opened 2020-09-29
Fix NPE in IN mlang

lchanmann:
Arg2 of IN can be List, object array or string.

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -1513,6 +1513,8 @@ foam.CLASS({
           a = ((List) arg2).toArray();
         } else if ( arg2 instanceof Object[] ) {
           a = (Object[]) arg2;
+        } else if ( arg2 != null ) {
+          b.append(arg2.toString());
         }
 
         if ( a != null ) {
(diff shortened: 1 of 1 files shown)


src/foam/mlang/mlang.js
```diff
@@ -1508,10 +1508,11 @@ foam.CLASS({
 
         Object[] a = null;
         Object arg2 = getArg2().f(null);
```
noodlemoodle, line 1510: please add null check on this line

lchanmann: for the null case it would jump to line 1525 which doesn't access arg2 so I think we are fine.

noodlemoodle: i mean if getArg2() is null, would that ever be the case?

lchanmann: yeah, getArg2().f() can be null but it would not throw. The toString() would just return empty arg2 eg. `In(arg1, )`.

noodlemoodle: nvm just tried, i thought getArg2().f() would throw on null arg2 but it just returned constant. so lgtm, thanks!

PR #3999, opened 2020-09-29
Accommodate arg2 being a property

lchanmann:
since CapabilityStore is using `IN('featured', Capability.KEYWORDS)`.


Nauna:
fixes https://nanopay.atlassian.net/browse/NP-2175

PR #4001, opened 2020-09-29
Np 2075

noodlemoodle:
https://nanopay.atlassian.net/browse/NP-2075
nanopay pr - https://github.com/nanoPayinc/NANOPAY/pull/10672


kgrgreer:
This is okay for now, but we should fix the TableView not to call foam.comics.v2.DAOSummaryView, on double-click, which was a terrible terrible mistake, which only makes the TableView when used with the v2 comics. Instead, it should import the double-click handler like it used to.

PR #4008, opened 2020-09-30
[NP-2189] Allow text to be saved in fileDAO

KolombetM:
https://nanopay.atlassian.net/browse/NP-2189

Added possibility to save/retrieve text to/from fileDAO
Fixed bug with FileDAODecorator

@jlhughes

### DIFF
--- a/src/foam/nanos/fs/TextSaveView.js
+++ b/src/foam/nanos/fs/TextSaveView.js
@@ -17,7 +17,10 @@ foam.CLASS({
   ],
 
   properties: [
-    'data'
+    'data',
+    {
+      name: 'fileName'
+    }
   ],
 
   methods: [
@@ -30,9 +33,7 @@ foam.CLASS({
           .add('File Name')
         .end()
         .start('input')
-          .attrs({
-            name: 'fileName'
-          })
+          .on('blur', this.setName)
         .end()
         .br()
         .br()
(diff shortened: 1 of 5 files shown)


src/foam/nanos/fs/TextSaveView.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+  ],
+
+  listeners: [
+    function onBlur(e) {
+      let text = e.target.value;
+      let blob = new Blob([text], {
+          type: 'text/plain'
+      });
+      let name = document.getElementsByName('fileName');
```
kgrgreer: This will prevent you from using two of these views on the same screen. Why not just create a property named 'fileName', then access the property value directly.

KolombetM: Like that?
`.on('blur', this.setName)`

`function setName(e) {`
`      this.fileName = e.target.value;`
`    }`

src/foam/nanos/fs/TextSaveView.js
```diff
@@ -0,0 +1,69 @@
...
+    'data',
+    {
+      name: 'fileName'
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.SUPER();
+      var name = "";
```
kgrgreer: This isn't used, is it?

src/foam/nanos/fs/TextSaveView.js
```diff
@@ -0,0 +1,69 @@
...
+
+      this
+        .start('p')
+          .add('File Name')
+        .end()
+        .start('input')
```
kgrgreer: You could just do .add(this.FILE_NAME) and achieve the same result with having to add the onBlur listener. Just make this class extend foam.u2.Controller instead.

src/foam/nanos/fs/TextSaveView.js
```diff
@@ -0,0 +1,69 @@
...
+          .on('blur', this.setName)
+        .end()
+        .br()
+        .br()
+        .start('p')
+          .add('Text')
+        .end()
+        .start('input')
```
kgrgreer: Also, make this a property called 'text', then just do .add(this.TEXT)

PR #4010, opened 2020-10-01
Investigate capability performance issue

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2191

## Changes
- Add PM to crunch service and prerequisite capability junction DAO
- Add caching for prerequisite capabilities to crunch service
- Fix Capability.implies and Capability.getPrereqsChainedStatus to use cached prerequisite capabilities

### DIFF
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -332,18 +332,7 @@ foam.CLASS({
       javaCode: `
         // CrunchService used to get capability junctions
         CrunchService crunchService = (CrunchService) x.get("crunchService");
-
-        boolean allGranted = true;
         DAO userCapabilityJunctionDAO = (DAO) x.get("userCapabilityJunctionDAO");
-        DAO myPrerequisitesDAO = ((DAO)
-          x.get("prerequisiteCapabilityJunctionDAO"))
-            .where(
-              EQ(CapabilityCapabilityJunction.SOURCE_ID, getId()));
-
-        List<CapabilityCapabilityJunction> ccJunctions =
-          ((ArraySink) myPrerequisitesDAO.select(new ArraySink()))
-          .getArray();
-
         DAO userDAO = (DAO) x.get("userDAO");
         Subject currentSubject = (Subject) x.get("subject");
 
@@ -359,10 +348,10 @@ foam.CLASS({
           subject.setUser((User) userDAO.find(ucj.getSourceId()));
         }
 
-        for ( CapabilityCapabilityJunction ccJunction : ccJunctions ) {
-          Capability cap = (Capability) ccJunction.findSourceId(x);
-          if ( ! cap.getEnabled() ) continue;
-          UserCapabilityJunction ucJunction = crunchService.getJunctionForSubject(x, ccJunction.getTargetId(), subject);
+        var prereqs = crunchService.getPrereqs(getId());
+        for ( var capId : prereqs ) {
+          if ( ! this.getEnabled() ) continue;
+          UserCapabilityJunction ucJunction = crunchService.getJunctionForSubject(x, capId, subject);
 
           if ( ucJunction != null && ucJunction.getStatus() == CapabilityJunctionStatus.GRANTED )
             continue;
@@ -374,9 +363,11 @@ foam.CLASS({
                && ucJunction.getStatus() != CapabilityJunctionStatus.PENDING ) {
             return CapabilityJunctionStatus.ACTION_REQUIRED;
           }
-          if ( ucJunction.getStatus() == CapabilityJunctionStatus.PENDING ) allGranted = false;
+          if ( ucJunction.getStatus() == CapabilityJunctionStatus.PENDING ) {
+            return CapabilityJunctionStatus.PENDING;
+          }
         }
-        return allGranted ? CapabilityJunctionStatus.GRANTED : CapabilityJunctionStatus.PENDING;
+        return CapabilityJunctionStatus.GRANTED;
       `,
     }
   ]
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -325,11 +325,14 @@ foam.CLASS({
       ],
       documentation: `
         Check statuses of all prerequisite capabilities - returning:
+        GRANTED: If the capability is disabled
         GRANTED: If all pre-reqs are in granted status
         PENDING: At least one pre-req is still in pending status
         ACTION_REQUIRED: If not any of the above
       `,
       javaCode: `
+        if ( ! this.getEnabled() ) CapabilityJunctionStatus.GRANTED;
+
         // CrunchService used to get capability junctions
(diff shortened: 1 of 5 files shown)


kgrgreer:
In the future, something like:
- Add PM to crunch service and prerequisite capability junction DAO
Should be in its own PR.

src/foam/nanos/crunch/Capability.js
```diff
@@ -356,24 +348,29 @@ foam.CLASS({
...
-          if ( ucJunction == null ) {
-            return CapabilityJunctionStatus.ACTION_REQUIRED;
-          }
-          if ( ucJunction.getStatus() != CapabilityJunctionStatus.GRANTED
-               && ucJunction.getStatus() != CapabilityJunctionStatus.PENDING ) {
-            return CapabilityJunctionStatus.ACTION_REQUIRED;
+        var prereqs = crunchService.getPrereqs(getId());
+        if ( prereqs != null ) {
+          for ( var capId : prereqs ) {
+            // QUESTION if the capability is disabled what would be the status of the prereqs chain?
```
kgrgreer: Just ignore it. Maybe in the future we should prevent users from disabling capabilities which are prerequisits.

lchanmann: @KernelDeimos @noodlemoodle what is the CapabilityJunctionStatus for ignoring?

noodlemoodle: i think they are just skipped over if its disabled, like the prereq doesnt exist

lchanmann: so `cap.getPrereqsChainedStatus(x, ucj)` would return GRANTED if cap is disabled.

noodlemoodle: yes

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4010/commits/ec0b01c80288c3f553b265b7aac531120c918838.

src/foam/nanos/crunch/Capability.js
```diff
@@ -351,8 +354,6 @@ foam.CLASS({
         var prereqs = crunchService.getPrereqs(getId());
         if ( prereqs != null ) {
           for ( var capId : prereqs ) {
-            // QUESTION if the capability is disabled what would be the status of the prereqs chain?
-            if ( ! this.getEnabled() ) continue;
```
noodlemoodle: i think this is still needed, i thought the question was on how the status of a prereq might affect the result of checkPrereqsChainedStatus if the prereq were disabled, which is that it shouldn't affect the result and just be skipped over when iterating through the list of prereqs, so this continue if not enabled should still be checked.

but if you were asking what Capability.getPrereqsChainedStatus should return if the Capability itself is disabled, then I'm not really sure, it probably shouldn't just return GRANTED though and just return the normal result of this method.

lchanmann: The old code https://github.com/foam-framework/foam2/blob/master/src/foam/nanos/crunch/Capability.js#L360-L361 is to find the source capability from the junction which is just the current capability, right?

Do we want to check the target capability instead?

lchanmann: >  it probably shouldn't just return GRANTED though and just return the normal result of this method.

What is the normal result?

noodlemoodle: ooh okok i see now, the old code should have been finding the target capability, yes we want to be checking the prereq.

re:normal result - the check in https://github.com/foam-framework/foam2/pull/4010/commits/ec0b01c80288c3f553b265b7aac531120c918838#diff-7b510a01f4db89bd364afb66eed05ee5R334 shouldn't be necessary because whether or not this capability is enabled doesn't affect the result of its prereqchainedstatus

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4010/commits/6dede8c2287162db8f7133d62229c77b6ce30c39.

kgrgreer:
LGTM, but will give others a chance to review also

kgrgreer (approved):
LGTM, but let Ruby review before merging.

src/foam/nanos/crunch/Capability.js
```diff
@@ -356,24 +349,30 @@ foam.CLASS({
...
+              continue;
+
+            if ( ucJunction == null ) {
+              return CapabilityJunctionStatus.ACTION_REQUIRED;
+            }
+            if ( ucJunction.getStatus() != CapabilityJunctionStatus.GRANTED
+                 && ucJunction.getStatus() != CapabilityJunctionStatus.PENDING ) {
+              return CapabilityJunctionStatus.ACTION_REQUIRED;
+            }
+            if ( ucJunction.getStatus() == CapabilityJunctionStatus.PENDING ) {
```
noodlemoodle: I think this should be kept the same, the idea is that if there is one or more prereqs in ACTION_REQUIRED, we will return ACTION_REQUIRED, because it will allow user to open the capability in the store and make changes.
If there are none that are action_required, we return pending if there are any in pending or approved, and return granted otherwise

lchanmann:
```
documentation: `
        Check statuses of all prerequisite capabilities - returning:
        GRANTED: If all pre-reqs are in granted status
        PENDING : At least one pre-req is still in pending/approved status
        ACTION_REQUIRED: If not any of the above
`
```
How does it sound for `getPrereqsChainedStatus()`? @noodlemoodle @KernelDeimos

noodlemoodle: I think for PENDING documentation, there should be one more condition of `all other prereqs not in pending/approved are in granted status`, or swap pending and action_required so that : 
action_required = atleast one prereq is not in pending/approved/granted status
pending = none of the above
because if some prereqs are in pending and some are action_required, we would want to return action_required.

but everything else LGTm

lchanmann: Updated in https://github.com/foam-framework/foam2/pull/4010/commits/003a198c4e70f0fa24bf28672e1bcf4fbe23e27e.

src/foam/nanos/crunch/Capability.js
```diff
@@ -257,12 +257,15 @@ foam.CLASS({
           if ( this.stringImplies(permissionName, permission) ) return true;
         }
 
-        List<CapabilityCapabilityJunction> prereqs = ((ArraySink) this.getPrerequisites(x).getJunctionDAO().where(EQ(CapabilityCapabilityJunction.SOURCE_ID, (String) this.getId())).select(new ArraySink())).getArray();
+        CrunchService crunchService = (CrunchService) x.get("crunchService");
+        var prereqs = crunchService.getPrereqs(getId());
```
Nauna, line 261: var prereqs = crunchService.getPrereqs(getId());

var ??? in java?

It exists since Java 10 -- ok cool!

Nauna (approved):

noodlemoodle (approved):

PR #4019, opened 2020-10-02
fix the detail view

hchoura:


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -456,7 +456,7 @@ foam.CLASS({
                   }).
                   callIf(view.dblclick && ! view.disableUserSelection, function() {
                     tableRowElement.on('dblclick', function() {
-                      view.dblclick && view.dblclick(thisObjValue, val[0]);
+                      view.dblclick && view.dblclick(null, val[0]);
```
kgrgreer, line 459: What does this fix, and how does it fix it?

PR #3812, opened 2020-08-17
[Np 1866] alarm

KolombetM:
Add AlarmConfigName to service.jrl
If name not in alarmConfig, add it
Change put method to take String as an argument
@jlhughes

PR #4020, opened 2020-10-05
Remove GUID Id from Alarm and AlarmConfig, replace with 'name' which …

jlhughes:
…is unique.

Make Alarm ID multipart on name and hostname so Alarms can be clustered.


src/foam/nanos/alarming/AddAlarmNameDAO.js
```diff
@@ -12,32 +12,31 @@ foam.CLASS({
...
+      AlarmConfig config = (AlarmConfig) configDAO.find(alarm.getName());
+      if ( config != null ) {
+        if ( ! config.getEnabled() ) {
+          return alarm;
         }
-        return super.put_(x, obj);
+      } else {
+        config = new AlarmConfig();
+        config.setName(alarm.getName());
+        configDAO.put(config);
```
KolombetM: Should this be in the try/catch to prevent failing the Proxy put?
https://github.com/foam-framework/foam2/pull/3812#discussion_r478688015

jlhughes: indeed, good catch.

jlhughes:
corrupted

PR #4027, opened 2020-10-05
Return all invalid props on object validate

zahnno:
Calling the validate java method on an object will now throw an IllegalStateException containing all invalid properties including the error message associated to the props validity.

Currently an illegal state exception is thrown for the first invalid property of an object when its validate method is called. This can create a continuous back and forth between the server and the requestor until an object is fully validated. We can improve this by catching all prop validateObj exceptions and throwing them together.

### DIFF
--- a/src/foam/core/FObject.java
+++ b/src/foam/core/FObject.java
@@ -433,9 +433,9 @@ public interface FObject
       } catch (IllegalStateException e) {
         exceptions.add(prop.getName() + ": " + e.getMessage());
       }
-      if ( exceptions.size() > 0 ) {
-        throw new IllegalStateException(exceptions.toString());
-      }
+    }
+    if ( exceptions.size() > 0 ) {
+      throw new IllegalStateException(exceptions.toString());
     }
   }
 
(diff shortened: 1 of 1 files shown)


jlhughes (commented):
@zahnno if we model ValidationException (or similar) with property
'propertyInfo' or whatever is appropriate, along with message and then use a CompoundException to gather them all, then the UI could: 
1. generate a banner for all - as you intend
2. highlight all the properties and display the returned error message.
(2) we could do later, but if use a ValidationException and CompoundException now then (2) will simple later.

src/foam/core/FObject.java
```diff
@@ -424,12 +425,18 @@ default int compareTo(Object o) {
...
   default void validate(foam.core.X x) {
-
     List<PropertyInfo> props = getClassInfo().getAxiomsByClass(PropertyInfo.class);
+    List<String> exceptions = new ArrayList<String>();
     for ( PropertyInfo prop : props ) {
-      prop.validateObj(x, this);
+      try {
+        prop.validateObj(x, this);
+      } catch (IllegalStateException e) {
+        exceptions.add(prop.getName() + ": " + e.getMessage());
```
jlhughes, line 434: So here we would look for CompoundException in the context and add to it. 
or create it on 429.  But a CompoundException in the context would capture the model plus any nested fobjects... as well.

PR #4028, opened 2020-10-05
NP-1442/after approving the approval request, it redirects to the earlier screen

olhabn:
draft pr

I decided to add contextActions to use specific methods for contextMenu if required as in case of https://nanopay.atlassian.net/browse/NP-1442, where as Anna suggests in the comments we need to 'go back' to approval request if we use actions in DetailView but we don't need to 'go back' to approval request if we use context menu actions.


src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -570,13 +573,28 @@
           this.notify(this.SUCCESS_REJECTED, '', this.LogLevel.INFO, true);
 
           if ( this.stack.top.length > 0 ) {
-            this.stack.back();
+            if ( ! stackBackIgnore )
+              this.stack.back();
+            // else
```
olhabn: we need to update table as otherwise we will not be able to observe changes immediately. so I thought maybe we can do that with stack somehow, but it doesn't look good.

we have `this.data.sub('on', function()` in UnstyledTableView line 410 but that's not triggered (not sure if it works as expected)

kgrgreer (changes requested):
I would like to investigate alternatives when I return on Wednesday. I'm concerned about having actions vs contextActions, and would like to look into options of unifying the two, possibly with extra meta-information.

src/foam/core/Action.js
```diff
@@ -294,3 +294,23 @@ foam.CLASS({
     }
   ]
 });
+
+
+/** Add Context Action support to Model. */
+foam.CLASS({
```
Nauna: don't think you need this

olhabn:
used isContextMenuEnabled and isDetailViewEnabled instead of ContextMenuActions refinement just as an option for tomorrows discussion

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -596,10 +610,16 @@
       },
       code: function(X) {
         var objToAdd = X.objectSummaryView ? X.objectSummaryView : X.summaryView;
+        var onExecuteFunc;
+
+        if ( X.controllerMode && X.controllerMode.equals(foam.u2.ControllerMode.SUMMARY) )
```
kgrgreer: For Enums, you can use ==, instead of equals(). It's faster and shorter. This works because all enum values are singletons.

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -596,10 +610,16 @@
...
+          onExecuteFunc =  this.approveWithMemo.bind(this);
```
kgrgreer: You don't have to make this change, but since these are used as callbacks/listeners, if you moved them from methods: to listeners: then you wouldn't need to .bind(this) them, because listeners are pre-bound.

src/foam/nanos/approval/ApprovalRequest.js
olhabn: fixed

src/foam/nanos/approval/ApprovalRequest.js
olhabn: fixed

src/foam/comics/v2/DAOSummaryView.js
```diff
@@ -109,6 +109,12 @@ foam.CLASS({
       factory: function() { 
         return () => this.stack.back();
       }
+    },
+    {
+      class: 'Enum',
+      of: 'foam.u2.ControllerMode',
+      name: 'controllerMode',
+      value: foam.u2.ControllerMode.SUMMARY
```
kgrgreer: I think that if we decorate the stack, we no longer need this ControllerMode.

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -474,6 +478,15 @@
       class: 'String',
       name: 'approvableHashKey',
       hidden: true
+    },
```
kgrgreer: I don't think this should be specific to the ApprovalRequest, but the subStack should be installed in the TableView so that actions called from there, either from the context menu or directly, don't cause a stack.back().

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -795,5 +775,67 @@
...
+        var approvedApprovalRequest = this.clone();
+        approvedApprovalRequest.status = this.ApprovalStatus.APPROVED;
+        approvedApprovalRequest.memo = memo;
+
+        this.approvalRequestDAO.put(approvedApprovalRequest).then((req) => {
+          this.approvalRequestDAO.cmd(this.AbstractDAO.RESET_CMD);
+          this.finished.pub();
+          this.notify(this.SUCCESS_APPROVED, '', this.LogLevel.INFO, true);
+
+          if ( this.subStack.top.length > 0 ) {
```
kgrgreer: Do we need this 'if'? The back() method in stack already has:
      isEnabled: function(pos) { return pos > 0; },

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -795,5 +775,67 @@
...
+        }, (e) => {
+          this.throwError.pub(e);
+          this.notify(e.message, '', this.LogLevel.ERROR, true);
+        });
+      }
+    }
   ]
 });
+
+foam.CLASS({
```
kgrgreer: Should be in it's own file, or in an inner-class (classes:) in the class/model that instantiates it, which shouldn't be this one.

src/foam/u2/Element.js
```diff
@@ -38,9 +38,10 @@ foam.ENUM({
   ],
 
   values: [
-    { name: 'CREATE', label: 'Create', modePropertyName: 'createVisibility' },
-    { name: 'VIEW',   label: 'View',   modePropertyName: 'readVisibility'   },
-    { name: 'EDIT',   label: 'Edit',   modePropertyName: 'updateVisibility' }
+    { name: 'CREATE',  label: 'Create',  modePropertyName: 'createVisibility' },
+    { name: 'VIEW',    label: 'View',    modePropertyName: 'readVisibility'   },
+    { name: 'EDIT',    label: 'Edit',    modePropertyName: 'updateVisibility' },
+    { name: 'SUMMARY', label: 'Summary', modePropertyName: 'readVisibility'   }
```
kgrgreer: Not needed.

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -795,5 +775,67 @@
...
+  package: 'foam.nanos.approval',
+  name: 'NoBackStackProxy',
+  properties: [
+    'delegate',
+    {
+      name: 'top',
```
kgrgreer: Does this work? When you call nostackbackproxy.top() it will get the right method, but then when it is called it will be bound to the wrong 'this'.
Make it a method and have it just call delegate.push(args) or delgate.apply(this, arguments).

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -795,5 +775,67 @@
...
+          this.notify(e.message, '', this.LogLevel.ERROR, true);
+        });
+      }
+    }
   ]
 });
+
+foam.CLASS({
+  package: 'foam.nanos.approval',
+  name: 'NoBackStackProxy',
```
kgrgreer: Just name NoBackStack. It isn't common to name decorators Proxy, even though decorators are a form of Proxy.

src/foam/u2/Element.js
olhabn: how we find out where from the approve method is called from without mode?

src/foam/nanos/approval/ApprovalRequest.js
kgrgreer: My mistake, I thought this was push(), which is what you should implement. Why do you need top?

src/foam/nanos/approval/ApprovalRequest.js
olhabn: I don't need it anymore

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -39,7 +39,11 @@
     'foam.dao.AbstractDAO',
     'foam.u2.dialog.Popup',
     'foam.log.LogLevel',
-    'foam.nanos.approval.ApprovalStatus'
+    'foam.nanos.approval.ApprovalStatus',
+    {
```
kgrgreer, line 43: Is this still needed?

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -39,7 +39,11 @@
...
+      path: 'foam.u2.stack.Stack',
```
kgrgreer: Why is the long-form needed?

src/foam/u2/view/NoBackStack.js
```diff
@@ -0,0 +1,13 @@
+foam.CLASS({
```
kgrgreer, line 7: Add copyright header.

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -251,6 +252,15 @@ foam.CLASS({
     {
       name: 'dataChangedFlag',
       class: 'Boolean'
+    },
+    {
+      name: 'subStack',
+      transient: true,
+      flags: ['js'],
```
kgrgreer: Why is this needed?

kgrgreer: I mean flags:

src/foam/u2/view/NoBackStack.js
olhabn: fixed

src/foam/u2/view/UnstyledTableView.js
olhabn: fixed, all of those changes were needed for ApprovalRequest java class

src/foam/nanos/approval/ApprovalRequest.js
olhabn: fixed

src/foam/nanos/approval/ApprovalRequest.js
olhabn: I don't need it anymore, my mistake

PR #4031, opened 2020-10-06
Fix invoice view

noodlemoodle:
https://nanopay.atlassian.net/browse/NP-2227
nanopay pr https://github.com/nanoPayinc/NANOPAY/pull/10719

### DIFF
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -449,7 +449,7 @@ foam.CLASS({
                   }).
                   callIf(view.dblclick && ! view.disableUserSelection, function() {
                     tableRowElement.on('dblclick', function() {
-                      view.dblclick && view.dblclick(null, obj.id);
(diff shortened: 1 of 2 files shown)


jlhughes (approved):

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -449,7 +449,7 @@ foam.CLASS({
                   }).
                   callIf(view.dblclick && ! view.disableUserSelection, function() {
                     tableRowElement.on('dblclick', function() {
-                      view.dblclick && view.dblclick(null, obj.id);
+                      view.dblclick && view.dblclick(view.dataChangedFlag ? null : obj, obj.id);
```
olhabn, line 452: why do we need this check here?

olhabn, line 452: why do we need to change code here?

we don't have a whole object in table we have only properties set, which are table columns, that's way we were passing null

PR #4038, opened 2020-10-07
Return all invalid props on object validate

zahnno:
Calling the validate java method on an object will now throw an IllegalStateException containing all invalid properties including the error message associated to the props validity.

Currently an illegal state exception is thrown for the first invalid property of an object when its validate method is called. This can create a continuous back and forth between the server and the requestor until an object is fully validated. We can improve this by catching all prop validateObj exceptions and throwing them together.

old pr: https://github.com/foam-framework/foam2/pull/4027


src/foam/box/ValidationException.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+foam.CLASS({
+  package: 'foam.box',
```
jlhughes: can you move this to foam.core, please.

jlhughes (approved):

PR #4039, opened 2020-10-07
fixed broken contextMenu

minsunny:
https://nanopay.atlassian.net/browse/NP-2252

added this logic back to fix a contextMenu on table


src/foam/u2/view/OverlayActionListView.js
```diff
@@ -154,6 +154,9 @@ foam.CLASS({
     async function initializeOverlay() {
       var self = this;
 
+      if ( this.obj )
```
olhabn, line 157: I think we don't need this condition

PR #4040, opened 2020-10-07
More capable min max wizard fixes

tharmaman:
Bunch of unexpected wizard fixes to get the wizard working again after the slotting rework @KernelDeimos and I tackled here https://github.com/foam-framework/foam2/pull/4033


KernelDeimos:
"Jesus christ I fixed the slotting again" :joy:

I think we should do a static analysis of slot logic in the wizard in the near future. Not just here but also in StepWizardletController which I had to write some inventive code in (I'd like to get rid of my inventive code if possible)

PR #4042, opened 2020-10-08
figuring out optional projection

olhabn:
draft pr


src/foam/core/Model.js
```diff
@@ -85,6 +85,14 @@ foam.CLASS({
...
       name: 'methods'
+    },
+    {
+      class: 'Boolean',
+      name: 'queryObjectOnProjection',
+      documentation: `set to true if we cannot build object of the class's subclasses from projection results. 
+        e.g. the class's subclasses inherit properties, which require subclasses properties value for "calculation".
+        for example FixedFee 'formula' property requires 'fixedFee' property to calculate its result but 'fixedFee' property is not inherited from Fee class,
+        when using projection we cannot query fixedFee property of an object but we need it for 'formula' calculation when display Fee records from DAO`
     }
```
olhabn: what is the best place to place it?
should it be moved to some kind of menu configuration?

src/foam/mlang/mlang.js
```diff
@@ -2757,6 +2757,12 @@ foam.CLASS({
       name: 'exprs',
       documentation: 'The expressions to be evaluated and returned in the projection. Typically are Properties.',
     },
```
olhabn, line 2772: **the reason these changes are here:**
if we have columns that not only require the class's fields but also class's reference field's property value

src/foam/mlang/mlang.js
```diff
@@ -2757,6 +2757,12 @@ foam.CLASS({
...
+    {
```
olhabn, line 2760: so for example,
we class like FixedFee, but which also contains reference property value of which we need then we can gather all the data we need with one request with using Projection

olhabn: as alternative we could do 2 requests: 1. select (to gather all the objects) 2. projection to gather Reference property data

src/foam/mlang/mlang.js
olhabn: so for example,
we class like FixedFee, but which also contains reference property value of which we need then we can gather all the data we need with one request with using Projection

olhabn: as alternative we could do 2 requests: 1. select (to gather all the objects) 2. projection to gather Reference property data if needed for all the objects in table

src/foam/core/Model.js
kgrgreer: Better to go in DAOMenu2.js I think.

olhabn:
fixed, but I just realized that we check only "custom tableCellFormatter" on current class, but what is there are  "custom tableCellFormatter" on subclasses?

src/foam/mlang/mlang.js
kgrgreer: You could use a projection with one of the columns being the identity function which would return the whole object, so you would only do one request. So your result would be [ {whole object}, A, B, C ... Z ].

olhabn: the idea that I used here is that instead of building objects when we call 'array' we would just return the whole object whenever we have buildObjectsFromProjectionWithClass set to false. and if buildObjectsFromProjectionWithClass set to false then we just build objects from projection result. @kgrgreer what do you think about this approach?

olhabn: fixed, will revert if suggestion above is approved

olhabn:
so, I'm not sure if checking for custom tableCellFormatter will be enough

src/foam/u2/view/UnstyledTableView.js
```diff
@@ -425,9 +425,17 @@ foam.CLASS({
             var proxy = view.ProxyDAO.create({ delegate: dao });
 
             view.props = this.returnPropertiesForColumns(view, view.columns_);
+            var doPropertiesHaveCustomTableCellFormatter = false;
+
+            for ( var p of view.props ) {
+              if ( p.property.tableCellFormatter && ! p.property.cls_.TABLE_CELL_FORMATTER.isDefaultValue() ) {
```
olhabn: I'm not sure that isDefaultValue is sufficient here but I don't have any suggestions how to check if the class has its own table cell formatter other then isDefaultValue

src/foam/mlang/mlang.js
```diff
@@ -2774,7 +2780,7 @@ foam.CLASS({
         List result = new ArrayList();
         if ( getProjectionWithClass() != null ) {
           for ( Object list: (ArrayList)getProjectionWithClass() ) {
-            Object[] obj1 = Arrays.copyOfRange((Object[])list, 1, ((Object[])list).length);
+            Object[] obj1 = Arrays.copyOfRange((Object[])list, 2, ((Object[])list).length);
```
kgrgreer: I'm not too keen of having magic 0's, 1's, and 2's all throughout the code. We should make these constants.

src/foam/mlang/mlang.js
```diff
@@ -2757,10 +2757,16 @@ foam.CLASS({
...
+      class: 'Boolean',
+      name: 'buildObjectsFromProjectionWithClass',
+      documentation: 'if objects should be build from projectionWithClass (and they will not be retrieved with select)',
+      value: true
+    },
     {
       class: 'List',
       name: 'projectionWithClass',
-      documentation: 'The projection but with the class in position 0 with all other values offset by 1.',
+      documentation: 'The projection but with the class in position 1 and full object if requested at position 0 with all other values offset by 2.',
```
kgrgreer: We would only ever need the class or the object, but never both, so could we put either at position 0? Or, would that we awkward to deal with? We could just send an empty object, ie. {class: 'myClass'} with no properties set, which would be more consistent. I'm not saying we should, just thinking about the possibility.

src/foam/u2/view/UnstyledTableView.js
kgrgreer: How about:

p.property.hasOwnProperty('tableCellFormatter');

src/foam/mlang/mlang.js

src/foam/mlang/mlang.js
olhabn: sure, thanks. fixed

src/foam/u2/view/UnstyledTableView.js
olhabn: oh right

src/foam/mlang/mlang.js
```diff
@@ -2750,17 +2750,36 @@ foam.CLASS({
...
   properties: [
     {
       class: 'Array',
       type: 'foam.mlang.Expr[]',
       name: 'exprs',
       documentation: 'The expressions to be evaluated and returned in the projection. Typically are Properties.',
     },
+    {
+      class: 'Boolean',
+      name: 'buildObjectsFromProjectionWithClass',
```
kgrgreer: useProjection
or
enableProjection
or something else

PR #4044, opened 2020-10-08
[Np 2065] Support for wizardlets with no data models (allows removal of `of` from MinMax capabilities)

KernelDeimos:
Some background context: The step wizard view has a sidebar rendering each Wizardlet as a numerated step, and each `foam.layout.Section` of the Wizardlet's `of` as a sub-step. This means wizardlets like MinMaxCapabilityWizardlet, which controls the visibility of other wizardlets, need an arbitrary `of` so that it can be rendered.

This PR adds `WizardletSection` and a default factory for the BaseWizardlet's sections to be converted into WizardletSection instances. With this new model a wizardlet can generate pseudo-sections rather than requiring an arbitrary data model, and these can have titles generated by other means.

### DIFF
--- a/src/foam/core/stdlib.js
+++ b/src/foam/core/stdlib.js
@@ -1028,6 +1028,7 @@ foam.LIB({
         if ( ! opt_defaultMethod ) {
           foam.assert(type, 'Unknown type: ', arg1,
             'and no default method provided');
+          if ( ! type[uid] ) debugger;
           foam.assert(
             type[uid],
             'Missing ' + name + ' multi-method for type ', arg1, ' map: ', map,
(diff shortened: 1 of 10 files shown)


src/foam/core/stdlib.js
```diff
@@ -1028,6 +1028,7 @@ foam.LIB({
         if ( ! opt_defaultMethod ) {
           foam.assert(type, 'Unknown type: ', arg1,
             'and no default method provided');
+          if ( ! type[uid] ) debugger;
```
kgrgreer: remove

src/foam/u2/wizard/WizardletSection.js
```diff
@@ -0,0 +1,61 @@
+foam.CLASS({
```
jlhughes, line 7: header please.

jlhughes (approved):

PR #4049, opened 2020-10-08
Capability Per Item / Intercept UI Improvements

kennyqiyenkan:
- Removed legend section in intercept. Moved to tooltip hover on status pill
- Placed capabilities into card for horizontal version with effects
- Changed copy for Intercept
- Moved close button in intercept to top right with icon instead of text

### DIFF
--- a/src/foam/u2/crunch/CapabilityInterceptView.js
+++ b/src/foam/u2/crunch/CapabilityInterceptView.js
@@ -41,50 +41,62 @@ foam.CLASS({
   ],
 
   messages: [
-    { name: 'TITLE', message: 'Welcome to Capability unlock options' },
-    { name: 'CAP_TITLE', message: 'Capabilities Available' },
-    { name: 'SUBTITLE_1', message: 'You do not have access to undertake your previous selected action.' },
-    { name: 'SUBTITLE_2', message: 'Please select one of the following capabilities, to unlock full feature.' }
+    { name: 'TITLE', message: 'Permission Required' },
+    { name: 'LABEL_CAP_LIST', message: 'YOUR CAPABILITIES' },
+    { name: 'SUBTITLE_1', message: 'You currently do not possess the permissions to perform this action.' },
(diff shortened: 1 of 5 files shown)


src/foam/u2/crunch/CapabilityInterceptView.js
```diff
@@ -41,50 +41,78 @@ foam.CLASS({
   ],
 
   messages: [
-    { name: 'TITLE', message: 'Welcome to Capability unlock options' },
-    { name: 'CAP_TITLE', message: 'Capabilities Available' },
-    { name: 'SUBTITLE_1', message: 'You do not have access to undertake your previous selected action.' },
-    { name: 'SUBTITLE_2', message: 'Please select one of the following capabilities, to unlock full feature.' }
+    { name: 'TITLE', message: 'Permission Required' },
```
jlhughes: @kgrgreer please comment on renaming Capabilities to Permissions.   This does not fit all scenarios.

kgrgreer: We intercept permissions and then find a capability that grants it, correct? So, it is always a capability that kicks this off, isn't it?

jlhughes: Yes, we want to retain the idea of 'Capability' in the system.  @kennyqiyenkan Logan we need to rework this language.

kennyqiyenkan: I am thinking "Hold On - You are currently incapable of performing this action. Please refer to the following list on how to obtain it." It's a bit more agnostic, still implies it is an interrupt, and provides enough information. I think. I ran this past Logan. Will change when he gets back to me if good.

kgrgreer: Sounds good.

kennyqiyenkan: Changed it to what I mentioned above with the exception of the title. It's now "Missing Capability"

jlhughes (approved):

PR #4051, opened 2020-10-08
make capable view displayed in add bank account screen

RachaelDing:
https://nanopay.atlassian.net/browse/NP-2222
Screen Shot 2020-10-08 at 5 10 14 PM

### DIFF
--- a/src/foam/u2/crunch/CrunchController.js
+++ b/src/foam/u2/crunch/CrunchController.js
@@ -264,5 +264,11 @@ foam.CLASS({
 
       return p;
     },
+
+    function getCapableWizard(capable) {
+      return this.Sequence.create(null, this.__subContext__.createSubContext({
+        capable: capable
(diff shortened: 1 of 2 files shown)


src/foam/u2/crunch/CrunchController.js
```diff
@@ -264,5 +264,11 @@ foam.CLASS({
 
       return p;
     },
+
+    function getCapableWizard(capable) {
```
KernelDeimos: Now that I think about it `getWizardletsFromCapable` is a better name, and maybe there should be a comment mentioning that this is specifically for CapableView.

KernelDeimos (changes requested):

src/foam/u2/crunch/CrunchController.js
RachaelDing: Updated. Thank you :-)

jlhughes (approved):

PR #4052, opened 2020-10-08
Fixed up handling if user exits wizard

tharmaman:


jlhughes (approved):
looks sane.

src/foam/u2/crunch/CrunchController.js
```diff
@@ -133,14 +133,16 @@ foam.CLASS({
       if ( isCapable ) {
         p = p.then(() => self.launchCapableWizard(intercept));
       }
-
+
```
kgrgreer, line 136: remove trailing spaces

PR #4061, opened 2020-10-11
NP-1441 Split Script run/event into own DAO

jlhughes:
Refactoring which occurred during Medusa development. 
Also correct an interaction between StorageOptional and Cronjobs which prevented the cronjob scheduledTime updates to not be stored in the MDAO.


src/foam/nanos/boot/Boot.java
```diff
@@ -41,7 +41,7 @@ public Boot(String datadir) {
     XLocator.set(root_);
 
     Logger logger = new ProxyLogger(new StdoutLogger());
-    root_.put("logger", logger);
+    root_ = root_.put("logger", logger);
```
kgrgreer: Why is this necessary? root_ is a ProxyX, which does this automatically.

src/foam/nanos/boot/Boot.java
```diff
@@ -113,17 +117,35 @@ public Object create(X x) {
...
       Script script    = (Script) scriptDAO.find(startScript);
       if ( script != null ) {
-        ((Script) script.fclone()).runScript(root_);
+        logger.info("Boot,script", startScript);
+        script = (Script) script.fclone();
+        // REVIEW: want main to run in ReadOnlyContext, but tests to run normally
+        if ( "main".equals(startScript) ) {
+          try {
+            // NOTE: is read-only and will throw exception when it updates rundate.
+            script.runScript(new foam.core.ReadOnlyDAOContext(root_));
```
kgrgreer: Why is it ReadOnly?

src/foam/nanos/cron/CronScheduler.java
```diff
@@ -55,30 +67,34 @@ public void run() {
...
-          }
-        });
+        if ( getEnabled() ) {
+          Date now = new Date();
 
+          cronDAO_.where(
+                         MLang.AND(
+                                   MLang.LTE(Cron.SCHEDULED_TIME, now),
+                                   MLang.EQ(Cron.ENABLED, true),
+                                   MLang.OR(
```
kgrgreer: IN would be easier / more efficient here.

src/foam/nanos/cron/CronScheduler.java
```diff
@@ -55,30 +67,34 @@ public void run() {
...
+                                            MLang.EQ(Cron.STATUS, ScriptStatus.UNSCHEDULED),
+                                            MLang.EQ(Cron.STATUS, ScriptStatus.ERROR)
+                                            )
+                                   )
+                         ).select(new AbstractSink() {
```
kgrgreer: Should there also be an orderBy() so they're run in the scheduled order?

src/foam/nanos/cron/CronScheduler.java
```diff
@@ -89,9 +105,13 @@ public void put(Object obj, Detachable sub) {
           if ( delay > 5000 ) {
```
kgrgreer: Code simpler if this is just: delay > 5000 is ! getEnabled()

src/foam/nanos/boot/Boot.java
jlhughes: I don't want it to save back into the journals/bootscript file.  The bootscript is just that, boot only, can't have side effects for clustering.

src/foam/nanos/cron/CronScheduler.java
jlhughes: good idea.  Mind you, on 'put' the ScriptRunnerDAO calls agency.execute so order is lost.

src/foam/nanos/boot/Boot.java
jlhughes: updated.

src/foam/nanos/cron/CronScheduler.java
jlhughes: changed.

src/foam/nanos/cron/CronScheduler.java
jlhughes: added

src/foam/nanos/boot/Boot.java
kgrgreer: But we have other bootscripts when we want to run unit tests from the command-line, like the "TestReporter" script. Will this break scripts like those?

jlhughes: no, only 'main' is run in a ReadOnlyDAOContext.

jlhughes: Removed. 
This readonly restriction on main has been in the medusa branch a long time. Whatever it was protecting is no longer an issue.

PR #4062, opened 2020-10-12
Fix ClassNotFoundException on replay

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2246

## Changes
- Generate MinMaxCapabilityWizardlet.java as it's used by Capability.wizardlet
- Output ClassInfo as string instead of key to conform with the FObject json parser

### DIFF
--- a/src/foam/lib/formatter/JSONFObjectFormatter.java
+++ b/src/foam/lib/formatter/JSONFObjectFormatter.java
@@ -438,7 +438,7 @@ public class JSONFObjectFormatter
   }
 
   public void output(ClassInfo info) {
-    outputKey(info.getId());
+    output(info.getId());
 //    append('{');
 //    if ( quoteKeys_ ) append(beforeKey_());
 //    append("class");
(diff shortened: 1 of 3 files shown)


lchanmann:
@kgrgreer @KernelDeimos @penzital please review when you have a chance, thanks.

src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -438,7 +438,7 @@ public void outputJson(String str) {
   }
 
   public void output(ClassInfo info) {
-    outputKey(info.getId());
+    output(info.getId());
```
kgrgreer, line 441: The old code should be correct. Is quoteKeys_ set to true? KeyParser should handle either quoted or unquoted keys. Let me look into the JS parser...

KernelDeimos:
@lchanmann why was the `requires` removed in favour of using the class path explicitly? `requires` is usually better because it allows both genjava and the javascript builder to see a more accurate dependancy tree.

PR #4069, opened 2020-10-14
Update temp ui adjustments for Pending_Review status of capabilities

Nauna:
Ensure that it is only an 'onboarding' capability where we see this - and only for non-signing officers


kgrgreer:
This is just temporary? How do you plan to do this permanently eventually?

kennyqiyenkan:
So basically, this is explaining the situation better. On the second file, you're hard coding the specific capability to show this?

Nauna:
The real fix...would be for the app to know who the user is in the context. 

If a user does not have permissions needed for a capability than status should be 'Pending_review'. 

Therefore a better way to handle this would be to have a rule on the UCJDAO handle this status returning based on the context user and permission check on all the capabilities in the hierarchy to confirm availability of a user to action on this capability (ie status=ACTION_REQUIRED). 

The ucjDAO rule needs to return on finds the status for permissioned and non-permissioned users.

Another real fix will be to check if the capability is in a category and not just checking a hardcoded set of ids. Perhaps setting property on rule to input id's or categories to run logic on.

jlhughes:
@kennyqiyenkan can you create a ticket to capture the real fix that @Nauna describes.  Thanks

jlhughes (approved):
merging as we need this in the short term.

jlhughes:
https://nanopay.atlassian.net/browse/NP-2433 ticket capture future fix.

PR #4070, opened 2020-10-14
Added client runtime exception

tharmaman:


KernelDeimos (approved):

src/foam/core/ClientRuntimeException.js
```diff
@@ -29,23 +29,22 @@ foam.CLASS({
...
-          for CompoundException where the CompoundException itself
-          is not intended to be re-thrown but any of its child
-          exceptions might be.`,
-          type: 'RuntimeException',
-          visibility: 'public',
-          javaCode: `return this;`
-      }
+        Note that the exception returned by this property is the
+        one that should be re-thrown. This is particularly useful
+        for CompoundException where the CompoundException itself
```
kgrgreer: What do you mean by this line and the next?

jlhughes: @tharmaman This documentation was copied from CompoundException and makes send there, but not here.

tharmaman: CompoundException is another Exception that implements foam.core.ExceptionInterface which contains a list of exceptions. This line is basically saying it it's case, for **CompoundException.getClientRethrowException** instead of determining whether it should rethrow itself to client, iterates through its list of exceptions and could throw one of those elements depending on whether or not those elements themselves implementing **getClientRethrowException**. Feel free to add @nick-prat @KernelDeimos 

Kept those lines here in documentation because also exists in **CapabilityRuntimeException** but can just remove it and only keep it on **CompoundException**.

tharmaman: @jlhughes it was also in CapabilityRuntimeException thats why I thought to keep it here as well so should it be removed from both

tharmaman: Justt removed that bit documentation from CapabilityRuntimeException and ClientRuntimeException

jlhughes (approved):

PR #4072, opened 2020-10-14
Np 2346/use action view in overlay action list view

olhabn:


src/foam/u2/Element.js
```diff
@@ -672,7 +672,7 @@ foam.CLASS({
   constants: [
     {
       name: 'CSS_CLASSNAME_PATTERN',
-      factory: function() { return /^[a-z_-][a-z\d_-]*$/i; }
```
olhabn: added whitespace as an possible character for class name. was getting an error otherwise

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -164,25 +188,25 @@ foam.CLASS({
...
-              addClass(self.myClass('action')).
-              add(action.label).
-              on('click', function(evt) {
-                self.overlay_.close();
-                action.maybeCall(self.__subContext__, self.obj);
-              }).
-              attrs({
+
+      self.obj.sub(function() {
+        self.overlay_.close();
```
olhabn, line 191: didn't delete overlay_ property as it causes slot error

src/foam/u2/Element.js
kgrgreer: Not a valid class name, but probably someone was setting two classnames at once. Can you detect when there's a whitespace a the do a 'debugger' so we can find who is doing that?

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -164,25 +188,25 @@ foam.CLASS({
...
+      this.overlay_.startContext({ data: self.obj })
+      .forEach(self.data, function(action) {
+        this
+            .start()
+              .addClass(action.createIsEnabled$(self.__context__, self.obj).map( e => e ? self.myClass('button-container') : self.myClass('disabled-button-container')))
+              .add(action)
+              .attrs({
                 disabled: action.createIsEnabled$(self.__context__, self.obj).map(function(e) {
-                  return e ? false : 'disabled';
+                  return e ? false : true;
```
kgrgreer: Is this the same as: ! e
?

src/foam/u2/Element.js
olhabn: sure

src/foam/u2/view/OverlayActionListView.js
kgrgreer: Are you still getting the same state error?

src/foam/u2/view/OverlayActionListView.js
olhabn: oh right

src/foam/u2/view/OverlayActionListView.js
olhabn: no, it's something different. that error disappeared after css class name pattern was changed

PR #4078, opened 2020-10-15
Np2340/change text

RachaelDing:
https://nanopay.atlassian.net/browse/NP-2340

Screen Shot 2020-10-14 at 9 08 52 PM
Screen Shot 2020-10-14 at 9 50 01 PM
Screen Shot 2020-10-14 at 10 02 55 PM


jlhughes (approved):

KernelDeimos (changes requested):
While here change `INTRO_TEXT` to something more generic, such as
> Proceeding will open a wizard through which you can gain new capabilities.

(basically it describes the function of CRUNCH rather than what the wizard is for, since we don't know what the wizard is for)

PR #4080, opened 2020-10-15
sessionIDProperty

smirnovaka:


src/foam/core/types.js
```diff
@@ -65,6 +65,54 @@ foam.CLASS({
   ]
 });
 
+foam.CLASS({
+  package: 'foam.core',
```
kgrgreer: Move to the foam.box package.

src/foam/box/types.js
```diff
@@ -0,0 +1,48 @@
+
```
kgrgreer: Add copyright. Rename file to SessionIDProperty.js.

PR #4082, opened 2020-10-15
fix message location from RPCErrorMessage

Nauna:
real error fix and revert https://github.com/foam-framework/foam2/pull/4079

### DIFF
--- a/src/foam/box/RPCErrorMessage.js
+++ b/src/foam/box/RPCErrorMessage.js
@@ -23,6 +23,13 @@ foam.CLASS({
     {
       class: 'Object',
       name: 'data'
+    },
+    {
+      class: 'String',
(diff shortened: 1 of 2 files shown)


src/foam/box/RPCErrorMessage.js
```diff
@@ -23,6 +23,13 @@ foam.CLASS({
     {
       class: 'Object',
       name: 'data'
+    },
+    {
+      class: 'String',
+      name: 'message',
+      expression: function(data) {
```
kgrgreer: More efficient in this case to just implement getter: since nobody would ever want to set this directly. Also, make transient since we don't need to send across the network.

jlhughes (approved):

PR #4092, opened 2020-10-16
Added hasRequirement method to capable

nick-prat:


KernelDeimos (changes requested):
If this is a public method there should be a comment indicating that this method does not verify that the capability is granted.

src/foam/nanos/crunch/lite/Capable.js
```diff
@@ -73,16 +73,28 @@ foam.INTERFACE({
 
           var oldCapabilityPayloads = getCapablePayloads();
           
-          if (
-            Arrays.stream(oldCapabilityPayloads)
-                .map((cap) -> cap.getCapability().getId())
-                .noneMatch(capabilityId::equals)
-          ) {
+          if ( !hasRequirement(x, capabilityId) ) {
```
kgrgreer: Space after !

PR #4093, opened 2020-10-16
Add http request to rule engine's system context

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2368

## Changes
Add http request to rule engine's system context for use by email utility to find the theme and parameters for emails.

There are rules that need to update some restricted properties on object eg. User.COMPLIANCE then it will also send a notification to the user but system context doesn't have the http request which is used when finding the corresponding theme and parameters for the notification email.

### DIFF
--- a/src/foam/nanos/ruler/RuleEngine.java
+++ b/src/foam/nanos/ruler/RuleEngine.java
@@ -13,6 +13,7 @@ import foam.nanos.logger.Logger;
 import foam.nanos.pm.PM;
 import foam.util.SafetyUtil;
 
+import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

kgrgreer:
This is a duplicate of https://nanopay.atlassian.net/browse/NP-2270 which I’ve already sent Garfield a fix to test.
Also, this approach has two weaknessesses:
1. It will only work if it is an HTTP request which first accesses the RuleEngine, not if it is a non-lazy service or accessed by a cron-job first.
2. ServletRequests and responses aren't safe to use after the request is finished. This causes memory leaks and can corrupt future requests.

PR #4098, opened 2020-10-19
Fix email theme

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2368

## Changes
- Add spid property to Theme
- Support finding theme through user.SPID
- Update EmailUtility to try getting theme from the context before going through theme finding

## Cases
1. domain:localhost => load theme by user.SPID, then group and user
2. domain:other => load theme by domain, then group and user
    - when themeDomain not found, fallback to 1. domain:localhost
    - when themeDomain is found but associated theme not found => load theme by user.SPID, then group and user
3. when loading theme by `user.SPID: 'x.y.z'` not found, try `user.SPID: 'x.y'` then `user.SPID: 'x'` then `user.SPID: '*'`


src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -29,7 +29,7 @@
   private X userX_;
 
   public RuleEngine(X x, X systemX, DAO delegate) {
-    setX(systemX);
+    setX(systemX.put("theme", x.get("theme")));
```
lchanmann: Carry the current theme to the system context so that we won't need to change every rules that need the system context and will also trigger notification.

jlhughes (approved):

src/foam/nanos/ruler/RuleEngine.java
kgrgreer: If the RuleEngine is first accessed from something like a CRON Job, then 'x' will be the systemX and it won't contain a theme. If it's accessed from the web, then it will just be stuck with the theme of the first user to access this service, which will be difficult to debug in the future.

jlhughes: yes, if manually run a cron/script, then you'll get that users (admin) theme, which is not what we want for the rule run.

jlhughes: An approach discussed earlier:  add spid to Theme and only support one Theme per spid.  The ThemeDomains can continue to handle the many domains per whilelabel - all pointing to the same Theme - mind you, no way to enforce this.

lchanmann: RuleEngine is a new instance per execution, the subsequent won't be affected by the former.

kgrgreer: SPID's are meant to be hierarchical, so a company could have as many as they need if they want multiple themes.

jlhughes: @kgrgreer I'm good with this approach.

lchanmann: SPID is added to theme domain, this approach expects the users of a specific SPID to use/come from the same domain.

jlhughes: Our current ServiceProvider implementation does not support hierarchy.

jlhughes: If we just support one theme per spid then spid goes on Theme model.  We can still support Theme per group and this is independent of spid supporting hierarchy. 

web -> ThemeDomain -> theme
cron -> user (spid) [-> group] -> theme

src/foam/nanos/theme/Themes.js
```diff
@@ -96,31 +98,74 @@ Later themes:
...
-          ));
+
+      // Find theme from user via SPID
+      if ( user != null
+        && ( theme == null || "localhost".equals(td.getId()) )
+      ) {
+        var spid = user.getSpid();
+        while ( ! SafetyUtil.isEmpty(spid) ) {
+          theme = (Theme) themeDAO.find(
+            MLang.AND(
```
jlhughes, line 157: So we are looking a Theme with id that starts with the spid? 
Why not just by spid - Theme.SPID

lchanmann: yeah, just realized and updated.

jlhughes (approved):

PR #4101, opened 2020-10-19
remove the message in bottom

RachaelDing:
https://nanopay.atlassian.net/browse/NP-2286
remove the message shown in bottom in nature code review screen

### DIFF
--- a/src/foam/nanos/crunch/ui/CapableMinMaxCapabilityWizardlet.js
+++ b/src/foam/nanos/crunch/ui/CapableMinMaxCapabilityWizardlet.js
@@ -183,7 +183,6 @@ foam.CLASS({
         choices$: this.choices$,
         booleanView: this.CardSelectView,
         isValidNumberOfChoices$: this.isValid$,
-        showValidNumberOfChoicesHelper: false,
         minSelected$: this.min$,
         maxSelected$: this.max$,
         onSelect: this.adjustCapablePayloads.bind(this)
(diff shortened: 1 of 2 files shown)


jlhughes (changes requested):
@lchanmann @RachaelDing  Rachael where is this being called/setup from?  That is where this should be set to false.

RachaelDing:
@jlhughes 
This is only called in the initE() method of the same class. And nowhere setup it after initialization, so I changed the initial value of if. Should I set it in iniE() method instead? Thank you.

lchanmann:
@RachaelDing there are two places that call it and is bound to this.isValid$
Screen Shot 2020-10-20 at 12 02 47 PM

RachaelDing:
@lchanmann  I see what you meaned... Thank you

src/foam/nanos/crunch/ui/CapableMinMaxCapabilityWizardlet.js
```diff
@@ -182,6 +183,7 @@ foam.CLASS({
         choices$: this.choices$,
         booleanView: this.CardSelectView,
         isValidNumberOfChoices$: this.isValid$,
+        showValidNumberOfChoicesHelper: false,
```
jlhughes: remove this please.

RachaelDing: removed

PR #4103, opened 2020-10-20
NP-2281/file not saved for capable payloads

jlhughes:


src/foam/nanos/fs/FileArrayInCapablePayloadsDecorator.js
```diff
@@ -0,0 +1,44 @@
...
+    'fileDAO'
+  ],
+
+  requires: [
+    'foam.nanos.fs.File'
+  ],
+
+  properties: [
+    {
+      class: 'foam.nanos.fs.FileArray',
```
jlhughes: @Mykola97 what is this for?

PR #4107, opened 2020-10-20
DecorateApprovalRequest feature for CapableCreateApprovalsRuleActoin.js and PropertyModal

tharmaman:


src/foam/u2/PropertyModal.js
```diff
@@ -0,0 +1,126 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
```
jlhughes, line 7: This seems exactly like the work @RachaelDing and @olhabn were working on.

jlhughes (approved):

src/foam/u2/PropertyModal.js
olhabn: I think we were working on deactivate account modal

PR #4109, opened 2020-10-21
Np 2373/add confirmation modal to action

RachaelDing:
https://nanopay.atlassian.net/browse/NP-2373

An example for adding confirmation model -- https://github.com/nanoPayinc/NANOPAY/pull/10853


tharmaman:
@jlhughes seems like the ConfirmationModal binds to an action and in particular destructive actions. Whereas the PropertyModal binds to a property so that a user can edit a property in a separate modal (i.e. when approving an approval request so that i don't have to hit edit on a  request in order to fill things out but see the entire model). If further information for approval is needed to be store on the request, I just bind that property to the PropertyModal.

Tl;dr ConfirmationModal != PropertyModal

src/foam/u2/ActionView.js
```diff
@@ -313,6 +318,11 @@ foam.CLASS({
         this.buttonState = undefined;
       }
     },
+    {
+      name: 'hasConfirmationModal',
+      factory: function() { 
+        return this.action && this.action.view ? true : false; }
```
tharmaman:
```suggestion
        return this.action && this.action.view ? true : false;
      }
```

src/foam/u2/Element.js
```diff
@@ -2889,18 +2889,10 @@ foam.CLASS({
...
-        return { class: 'foam.u2.ActionView', action: this };
-      }
-    }
-  ],
-
   methods: [
     function toE(args, X) {
-      var view = foam.u2.ViewSpec.createView(this.view, args, this, X);
+      var view = foam.u2.ViewSpec.createView(
+        { class: 'foam.u2.ActionView', action: this }, args, this, X);
```
tharmaman:
```suggestion
        { class: 'foam.u2.ActionView', action: this }, args, this, X
      );
```

tharmaman (approved):

src/foam/core/Action.js
```diff
@@ -110,6 +110,11 @@ foam.CLASS({
       required: true,
       value: null
     },
+    {
+      class: 'Function',
+      name: 'view',
```
kgrgreer: I think a more informative name like confirmationView would be helpful.

src/foam/u2/ActionView.js
```diff
@@ -313,6 +318,12 @@ foam.CLASS({
         this.buttonState = undefined;
       }
     },
+    {
+      name: 'hasConfirmationModal',
+      factory: function() { 
+        return this.action && this.action.view ? true : false;
```
kgrgreer: BTW, this would be simpler to just write as:
return this.action && this.action.view

But I don't think this needs to be a property, since developers aren't going to set it directly, are they? Better to make a method or just inline this logic in the one place where it is used.

PR #4113, opened 2020-10-21
add msg to address

hchoura:


src/foam/nanos/auth/Address.js
```diff
@@ -22,16 +22,24 @@ foam.CLASS({
   ],
 
   messages: [
-    { name: 'INVALID_POSTAL_CODE_ERR_MSG', message: 'Invalid postal code' }
+    { name: 'INVALID_POSTAL_CODE_ERR_MSG', message: 'Invalid postal code' },
+    { name: 'INVALID_ADDRESS_1', message: 'Invalid value for address 1.' },
+    { name: 'PROVINCE_REQUIRED', message: 'Province required.' },
```
kgrgreer, line 27: Should this be REGION_REQUIRED?

src/foam/nanos/auth/Address.js
```diff
@@ -128,11 +136,11 @@ foam.CLASS({
         if ( typeof regionId !== 'string' || regionId.length === 0 ) {
           switch ( countryId ) {
             case 'CA':
-              return 'Province required.';
+              return this.PROVINCE_REQUIRED;
```
kgrgreer, line 139: This doesn't look correct. We should just be using this.REGION_REQUIRED and let the normal i18n system replace that message with the appropriate one.

jlhughes: Yes, all this should be removed now.  English - we can use region or province/state (to handle CA/US)  and then i18n for the rest.

jlhughes:
Addressed in another PR.

PR #4114, opened 2020-10-21
md

smirnovaka:


src/foam/u2/property/MDInput.js
```diff
@@ -0,0 +1,240 @@
...
+  extends: 'foam.u2.View',
+
+  properties: [
+//    [ 'nodeName', 'input' ],
+    {
+      name: 'data',
+      preSet: function(o, d) {
+        var f = ! d || typeof d === 'string' || typeof d === 'number' || typeof d === 'boolean' || foam.Date.isInstance(d);
+        if ( ! f ) {
+          this.__context__.warn('Set Input data to non-primitive:' + d);
```
kgrgreer: Why not import: 'warn', then do this.warn?

PR #4115, opened 2020-10-21
Skipping lookup when locale isn't defined

nanoscott:


kgrgreer:
What if the locale is just something like "en"?

kgrgreer:
I've replaced with another PR which removes manual translation. Eric will address by calling the new Element.tranlsate() method.

PR #4125, opened 2020-10-22
Allow theme.spid default to "*"

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2405

## Changes
- Allow theme.spid default to "*"

## Future work
- https://nanopay.atlassian.net/browse/NP-2457

### DIFF
--- a/src/foam/nanos/theme/Theme.js
+++ b/src/foam/nanos/theme/Theme.js
@@ -519,7 +519,18 @@ foam.CLASS({
       name: 'spid',
       class: 'Reference',
       of: 'foam.nanos.auth.ServiceProvider',
-      value: foam.nanos.auth.ServiceProviderAware.GLOBAL_SPID
+      documentation: `
+        Only set spid on the main theme so that it can be looked up via spid
+        when could not find theme via domain.
+
+        By default, theme.spid is empty to denote an auxiliary theme which is
+        used for augmenting the theme for specific users and groups.
+
+        There should be only one theme associated with a service provider. For
+        users in the same service provider to have a different styling the users
+        or group that the users belong to should be configured with an auxiliary
+        theme.
+      `,
     }
   ],
 
--- a/src/foam/nanos/theme/Theme.js
+++ b/src/foam/nanos/theme/Theme.js
@@ -519,18 +519,7 @@ foam.CLASS({
       name: 'spid',
       class: 'Reference',
       of: 'foam.nanos.auth.ServiceProvider',
-      documentation: `
-        Only set spid on the main theme so that it can be looked up via spid
-        when could not find theme via domain.
-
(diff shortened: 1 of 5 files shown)


src/foam/nanos/theme/Theme.js
```diff
@@ -519,7 +519,18 @@ foam.CLASS({
       name: 'spid',
       class: 'Reference',
       of: 'foam.nanos.auth.ServiceProvider',
-      value: foam.nanos.auth.ServiceProviderAware.GLOBAL_SPID
+      documentation: `
+        Only set spid on the main theme so that it can be looked up via spid
+        when could not find theme via domain.
+
+        By default, theme.spid is empty to denote an auxiliary theme which is
```
kgrgreer: Why is it called "auxiliary"?

src/foam/nanos/theme/Theme.js
```diff
@@ -519,7 +519,18 @@ foam.CLASS({
...
+        used for augmenting the theme for specific users and groups.
```
kgrgreer: What do you mean by "augmenting"?

src/foam/nanos/theme/Theme.js
lchanmann: auxiliary because it's only for augmenting the other theme.

src/foam/nanos/theme/Theme.js
lchanmann: Themes.findTheme(x) tries to find theme by domain, if not found then find by user.SPID, if cannot get one then make a new theme. User and group can be configured with a theme to augment the found/created theme previously.

lchanmann: It checks for user and group theme then does copyFrom instead of replacing. 

https://github.com/foam-framework/foam2/blob/927d3190f9c2b256b180a247ba1deb2ca3955b6f/src/foam/nanos/theme/Themes.js#L186-L203

src/foam/nanos/theme/Theme.js
lchanmann: Fixed https://github.com/foam-framework/foam2/pull/4125/commits/9a9d9c5c4f163a1545165f77e926b5ff08030465 - use "*" as default spid for theme.

src/foam/nanos/theme/Theme.js
lchanmann: Fixed https://github.com/foam-framework/foam2/pull/4125/commits/9a9d9c5c4f163a1545165f77e926b5ff08030465 - use "*" as default spid for theme.

lchanmann:
Ticket created for rule to theme them all https://nanopay.atlassian.net/browse/NP-2457, the detail will be provided by @kgrgreer.

PR #4126, opened 2020-10-22
[NP-2234] - Text Overlap Fix

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-2234
#TimeSpent: 4h

Screen Shot 2020-10-22 at 5 42 01 PM
Screen Shot 2020-10-22 at 5 42 09 PM
Screen Shot 2020-10-22 at 5 40 29 PM

### DIFF
--- a/src/foam/u2/detail/SectionedDetailPropertyView.js
+++ b/src/foam/u2/detail/SectionedDetailPropertyView.js
@@ -231,7 +231,7 @@ foam.CLASS({
         .add(this.slot(function(mode, prop, prop$label) {
 
           var isCheckBox = false;
-          if ( typeof prop.view === "function" ) {
+          if ( foam.Function.isInstance(prop.view) ) {
             var view = prop.view(null, this);
             isCheckBox = view.class === 'foam.u2.CheckBox' || ( view.cls_ && view.cls_.id === 'foam.u2.CheckBox' );
           }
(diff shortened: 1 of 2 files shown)


kennyqiyenkan (approved):

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -228,13 +230,22 @@ foam.CLASS({
         .addClass(`sectioned-detail-property-${this.prop.name}`)
         .add(this.slot(function(mode, prop, prop$label) {
 
+          var isCheckBox = false;
+          if ( typeof prop.view === "function" ) {
```
kgrgreer: foam.Function.isInstance() is more consistent.

Nauna:
I wonder is there anyway to make the validation section size itself based on the size of the top container. I think adding checks for Checkbox explicitly can be troublesome.

nanopay-arthur:
> I wonder is there anyway to make the validation section size itself based on the size of the top container. I think adding checks for Checkbox explicitly can be troublesome.

@Nauna So the issue wasn't really the validation section but rather the text section of the checkbox. The height of the top container wasn't being updated correctly due to the label text being position: absolute. I wasn't able to come up with another fix for this, but I agree, doing an explicit check for checkbox is not pretty.

kgrgreer:
Has merge conflict. Please resolve.

PR #4130, opened 2020-10-23
fix the name

hchoura:
@KernelDeimos

### DIFF
--- a/src/foam/u2/wizard/StepWizardletStepsView.js
+++ b/src/foam/u2/wizard/StepWizardletStepsView.js
@@ -112,7 +112,7 @@ foam.CLASS({
 
                 // Render title
                 .start('p').addClass(self.myClass('title'))
-                  .translate([wizardlet.id, 'title'], wizardlet.title)
+                  .translate(wizardlet.capability.id+'.name')
                   .style({
                     'color': isCurrent ? this.theme.black : this.theme.grey2
                   })


src/foam/u2/wizard/StepWizardletStepsView.js
```diff
@@ -112,7 +112,7 @@ foam.CLASS({
 
                 // Render title
                 .start('p').addClass(self.myClass('title'))
-                  .translate([wizardlet.id, 'title'], wizardlet.title)
+                  .translate(wizardlet.capability.id+'.name')
```
kgrgreer, line 115: I think: ['wizardlet', wizardlet.id, 'title']

KernelDeimos: Right now `wizardlet.id` will be something like `capability,UUID-STRING`, which will become `capability,UUID-STRING,title` after the code that's currently in place. Not all wizardlets are capability wizardlets, so I made determining the prefix the wizardlet's responsibility so the prefix could be `capability`. That said, if all our IDs are UUIDs then the prefix probably isn't needed at all.

@kgrgreer 's solution works, but if we translate the capability title outside of a wizardlet we might end up creating duplicate translation entries. (ex: `capability.UUID-STRING.title` and also `wizardlet.UUID-STRING.title`)

`wizardlet.capability.id+'.name'` should not be used because not all wizardlets are capability wizardlets.

PR #4143, opened 2020-10-26
add default language

hchoura:


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -400,6 +400,7 @@ foam.CLASS({
         let l = localStorage.getItem('localeLanguage');
         if ( l !== undefined ) foam.locale = l;
         //TODO manage more complicated language. 'en-CA'
+        foam.locale =  foam.locale ? foam.locale: 'en-US';
```
jlhughes:
```suggestion
        foam.locale =  foam.locale ? foam.locale : 'en-US';
```

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -400,6 +400,7 @@ foam.CLASS({
         let l = localStorage.getItem('localeLanguage');
         if ( l !== undefined ) foam.locale = l;
```
lchanmann, line 401: localStorage.getItem() can return null and it will be assigned to foam.locale, is it expected?

hchoura: Yes, the localStorage may not have that variable

lchanmann: shouldn't it also check for null

    if ( l ) foam.locale = l;

kgrgreer: In Boot.js we do:
      let l = global.localStorage && global.localStorage.getItem('localeLanguage');
      foam.locale = l || 'en';

Should we just change the default there to 'en-US'. How is it that if that code runs that locale isn't set here?

hchoura: The problem it run before, I didn't see this issue.
In the getItem, I will get null. it should be 'en' or other locale.

PR #4144, opened 2020-10-26
Changed copy for the capability statuses

kennyqiyenkan:
Changed copy as discussed

### DIFF
--- a/src/foam/nanos/crunch/CapabilityJunctionStatus.js
+++ b/src/foam/nanos/crunch/CapabilityJunctionStatus.js
@@ -10,31 +10,31 @@ foam.ENUM({
     {
       name: 'PENDING',
       label: { en: 'pending', pt: 'pendente'},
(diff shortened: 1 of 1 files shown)


src/foam/nanos/crunch/CapabilityJunctionStatus.js
```diff
@@ -10,31 +10,31 @@ foam.ENUM({
     {
       name: 'PENDING',
       label: { en: 'pending', pt: 'pendente'},
-      documentation: `This is awaiting verification. Capability access may take up to 24hrs.`,
+      documentation: `Our team is currently reviewing. Approvals may take up to 24 hours.`,
```
kgrgreer: Use regular single-quotes: ' instead of back-quotes ` when not using interpolation or multiple-lines since it parses faster and has less chance for unintended consequences.

kennyqiyenkan: Ah okay. changing it.

PR #4145, opened 2020-10-26
[NP-2483] Add version check for html file

KolombetM:
@kgrgreer

### DIFF
--- a/src/foam/nanos/blob/FileService.java
+++ b/src/foam/nanos/blob/FileService.java
@@ -11,6 +11,7 @@ import foam.blob.BlobService;
 import foam.blob.IdentifiedBlob;
 import foam.core.X;
 import foam.dao.DAO;
+import foam.nanos.app.AppConfig;
 import foam.nanos.auth.AuthService;
 import foam.nanos.auth.User;
 import foam.nanos.fs.File;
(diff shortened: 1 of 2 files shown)


src/foam/nanos/blob/FileService.java
```diff
@@ -54,6 +57,15 @@ protected void download(X x) {
         return;
       }
 
+      //Replace @version@ with actual foam version
+      if (file.getMimeType().equals("text/html")) {
```
kgrgreer: if ( "text/html".equals(file.getMimeType()) ) {
is slightly better because it avoids possibility of NPE (not likely in this case, but in general, use the literal as 'this' to avoid the possibility. Also, fixes the spacing with spaces after ( and before ).

src/foam/nanos/fs/File.js
```diff
@@ -177,7 +178,16 @@ foam.CLASS({
 
           reader.readAsText(this.data.blob)
         });
-      }
+      },
+      javaCode: `
+        if ( ! SafetyUtil.isEmpty(this.getDataString()) ) {
+          String          encodedString = this.getDataString().split(",")[1];
```
kgrgreer: Remove the extra spaces between types and variable names.

jlhughes (approved):

PR #4148, opened 2020-10-26
Allow object with spid="*" to be found

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2472

## Issues
- Treviso backoffice users need to find theme of "\*" spid defined in the parent group. The themeDAO has GlobalReadAuthorizer but ServiceProviderAwareSupport prevents it from reading the theme with "*" spid and fails the loading of group theme.

## Changes
- Allow object with spid="\*" to be found

### DIFF
--- a/src/foam/nanos/auth/ServiceProviderAwareSupport.js
+++ b/src/foam/nanos/auth/ServiceProviderAwareSupport.js
@@ -101,7 +101,7 @@ returning true if the spid or context users spid matches the current object.`,
       if ( obj != null &&
            obj instanceof ServiceProviderAware ) {
         ServiceProviderAware sp = (ServiceProviderAware) obj;
-        return sp.getSpid().equals("*") || sp.getSpid().startsWith(spid) ||
+        return sp.getSpid().startsWith(spid) || sp.getSpid().equals("*") ||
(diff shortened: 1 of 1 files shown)


kgrgreer:
LGTM

src/foam/nanos/auth/ServiceProviderAwareSupport.js
```diff
@@ -101,7 +101,7 @@ returning true if the spid or context users spid matches the current object.`,
       if ( obj != null &&
            obj instanceof ServiceProviderAware ) {
         ServiceProviderAware sp = (ServiceProviderAware) obj;
-        return sp.getSpid().startsWith(spid) ||
+        return sp.getSpid().equals("*") || sp.getSpid().startsWith(spid) ||
```
jlhughes: This will be the least common case, so perhaps swap the order of the first two tests.

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4148/commits/6026f7ada2b1a6f08141995c702099d6198db7d6.

jlhughes (approved):

PR #4149, opened 2020-10-27
add Default Language For Capability

hchoura:
add the default value for data from the journal

### DIFF
--- a/src/foam/u2/crunch/CapabilityRequirementView.js
+++ b/src/foam/u2/crunch/CapabilityRequirementView.js
@@ -42,7 +42,7 @@ foam.CLASS({
       display: inline-block;
       padding-bottom: 4vh;
     }
-    
+
     ^ p {
       display: inline-block;
       position: relative;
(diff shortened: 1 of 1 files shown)


src/foam/u2/crunch/CapabilityRequirementView.js
```diff
@@ -133,10 +133,13 @@ foam.CLASS({
 
         .start().addClass('titles')
           // title
-          .start().addClass('makeBold').translate(mainCapability[0].id+'.name').end()
+          .start().addClass('makeBold').translate(mainCapability[0].id+'.name', mainCapability[0].name).end()
           // subTitle
           .start().addClass('subTitle')
-            .translate(mainCapability[0].requirementViewTitle ? mainCapability[0].id + '.requirementViewTitle' : this.INTRO_TEXT)
+            .callIf( mainCapability[0].requirementViewTitle,
```
kgrgreer: Better might be just:
.translate(mainCapability[0].id + '.requirementViewTitle' , mainCapability[0].requirementViewTitle || this.INTRO_TEXT)
Then you can change the text using a locale msg even if the title wasn't set in the capability.

PR #4152, opened 2020-10-27
NP2465-And-NP2462 / UI fixes for Signup and User registration 

RachaelDing:
Addresses:
- https://nanopay.atlassian.net/browse/NP-2465
- https://nanopay.atlassian.net/browse/NP-2462

Changes:
- Added a wizardTitle to section and wizardSection, so that we can specify a wizard title -- For stepWizardlet, we use the wizardTitle to specify the subTtitle for each StepWizardletSteps
- Change String's default validation error messages for consistency
- Didn't manually add sentence cased labels, because this should be automatically handled after NP-2477 is finished

Screen Shot 2020-10-27 at 10 14 21 AM
Screen Shot 2020-10-27 at 10 14 29 AM
Screen Shot 2020-10-27 at 10 24 10 AM

### DIFF
--- a/src/foam/layout/Section.js
+++ b/src/foam/layout/Section.js
@@ -55,7 +55,10 @@ foam.CLASS({
       name: 'subTitle'
     },
     {
-      name: 'wizardTitle'
+      name: 'navTitle',
+      expression: function (title) {
+        return title;
+      }
     },
     {
       documentation: 'function and string',
--- a/src/foam/layout/Section.js
+++ b/src/foam/layout/Section.js
@@ -55,6 +55,9 @@ foam.CLASS({
       name: 'subTitle'
     },
     {
+      name: 'wizardTitle'
+    },
+    {
       documentation: 'function and string',
       name: 'help'
     },
--- a/src/foam/layout/SectionAxiom.js
+++ b/src/foam/layout/SectionAxiom.js
@@ -24,6 +24,9 @@ foam.CLASS({
       name: 'subTitle'
     },
     {
+      name: 'wizardTitle'
+    },
+    {
       name: 'help'
     },
     {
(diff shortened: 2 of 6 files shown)


src/foam/layout/Section.js
```diff
@@ -54,6 +54,9 @@ foam.CLASS({
       // Accepts function and string
       name: 'subTitle'
     },
+    {
+      name: 'wizardTitle'
+    },
```
KernelDeimos, line 62: Adding this expression might avoid the need to set wizardletTitle in some cases. (this is already handled by the expression in `WizardletSection`, but I just realized it's better over here in case something else uses this property in the future)
```
{
  name: 'wizardTitle',
  expression: function (title) {
    return title;
  }
},
```

src/foam/layout/SectionAxiom.js
```diff
@@ -23,6 +23,9 @@ foam.CLASS({
     {
       name: 'subTitle'
     },
+    {
+      name: 'wizardTitle'
```
KernelDeimos: @kgrgreer is this better as `wizardTitle`, `shortTitle`, or `navTitle`?

KernelDeimos (changes requested):

src/foam/layout/Section.js
kennyqiyenkan: Yeah, I agree. There might be cases where you may want a separate title on the nav, and others may be same.

src/foam/layout/SectionAxiom.js
kennyqiyenkan: personally, i prefer navTitle if it is being used on the navigation portion.

kgrgreer: Sections are more generic than wizards, so I agree with Kenny, something like navTitle or even just 'label' would be better.

PR #4153, opened 2020-10-27
Password Reactive Eye

zahnno:
- Toggles opacity of password visibility image when text element is focused.
- Smaller visibility image size.

### DIFF
--- a/src/foam/u2/view/PasswordView.js
+++ b/src/foam/u2/view/PasswordView.js
@@ -82,19 +82,22 @@ foam.CLASS({
           type: this.type,
           data$: this.data$,
           onKey: true,
-
         }, this.inputElement$)
-        .on('focus', () => this.opacity = 1)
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/PasswordView.js
```diff
@@ -68,24 +69,34 @@ foam.CLASS({
...
-        }, this.inputElement$).
-        addClass('full-width-input-password').end().
-        start('img').show(this.passwordIcon$).addClass('input-image').
-          attr('src', this.visibilityIcon$).on('click', this.visible).
-        end().
-      end();
+          onKey: true,
+
+        }, this.inputElement$)
+        .on('focus', () => this.opacity = 1)
```
kgrgreer: Couldn't this be done more easily with just CSS?

zahnno: You might be right

PR #4154, opened 2020-10-27
Adding CopyFromDAO

nanoscott:
Creating the CopyFromDAO which allows a subset of properties to be shown as the only properties on an entity.


src/files2.js
```diff
@@ -185,6 +185,7 @@ FOAM_FILES([
   { name: "foam/dao/ArraySink" },
   { name: "foam/dao/DAOInterceptor" },
   { name: "foam/dao/AbstractDAO" },
+  { name: "foam/dao/CopyFromDAO" },
```
nanoscott: Question: Is this necessary? I only added it because foam.dao.DAO is listed in both.

kgrgreer: I don't know what files2.js is.

kgrgreer:
I think that AdapterDAO does the same thing by default, but only on the JS side. Maybe better to just add javaCode: blocks to that class. Thoughts?

jlhughes (commented):
I would call this an Adapter.

src/foam/dao/Adapter.js
```diff
@@ -0,0 +1,61 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+foam.INTERFACE({
+  package: 'foam.dao',
+  name: 'Adapter',
+
+  documentation: `DAO for adapting between "of" input type and "to" delegate
```
kgrgreer: Documentation looks like for AdapterDAO, not DAOInterface.

nanoscott: Updated .. put the correct header too.

jlhughes (approved):

src/foam/dao/AdapterDAO.js
```diff
@@ -38,15 +52,37 @@ foam.CLASS({
...
-        function put(o, sub) {
-          this.delegate.put(this.adapt(o), sub);
+        {
+          name: 'put',
+          code: function put(o, sub) {
+            this.delegate.put(this.adapt(o), sub);
+          },
+          javaCode: `
+            FObject innerObject = (FObject) obj;
+            FObject outerObject = getAdapter().adaptFromDelegate(getX(), innerObject);
```
kgrgreer: Like this JS version, I would just call adapt(), instead of getAdapter().adaptFromDelegate() and then remove the 'adapter' property and have the adapt and unadapt methods just do what you want by default. Behaviour can be modified if needed by just subclassing and overriding those two methods.

PR #4158, opened 2020-10-27
Translate sub title

hchoura:


src/foam/u2/detail/SectionView.js
```diff
@@ -80,7 +80,7 @@ foam.CLASS({
                 obj$: self.data$,
                 code: section.subTitle
               }) : section.subTitle$;
-              this.start().addClass('subtitle').add(slot$).end();
+              this.start().addClass('subtitle').translate(slot$.obj.capability.id+'.description',slot$).end();
```
kgrgreer, line 83: .obj will only be there in the case that slot$ is an ExpressionSlot, and even then, it won't work because it won't be the text returned from the subTitle function.

PR #4161, opened 2020-10-27
add null check for title and navTtitle in WizardletSection

RachaelDing:


src/foam/u2/wizard/WizardletSection.js
```diff
@@ -28,7 +28,9 @@ foam.CLASS({
       name: 'title',
       class: 'String',
       expression: function(section) {
-        return section.title;
+        if (section) {
+          return section.title;
```
KernelDeimos: Oh wow it was missing from title too? Good catch

KernelDeimos (approved):

src/foam/u2/wizard/WizardletSection.js
kgrgreer: Or just: return section && section.title;

src/foam/u2/wizard/WizardletSection.js
```diff
@@ -49,7 +51,9 @@ foam.CLASS({
       name: 'navTitle',
       class: 'String',
       expression: function(section) {
-        return section.navTitle;
+        if (section) {
```
kgrgreer: same

PR #4163, opened 2020-10-28
Fix ServiceProvider permissionsGranted getter

KernelDeimos:
There were two problems causing NP-2471. A transient property was set in a journal entry, which caused my last fix (without the null check) to break the build. This PR changes the property to `networkTransient` and adds a null check.

### DIFF
--- a/src/foam/nanos/auth/ServiceProvider.js
+++ b/src/foam/nanos/auth/ServiceProvider.js
@@ -52,8 +52,16 @@ foam.CLASS({
     },
     {
       name: 'permissionsGranted',
-      transient: true,
-      javaGetter: 'return new String[] { "serviceprovider.read." + getId() };',
+      networkTransient: true,
+      javaGetter: `
+        int c = permissionsGranted_ != null ? permissionsGranted_.length : 0;
+        String[] perms = new String[c + 1];
+        for ( int i = 0 ; i < c ; i++ ) {
+          perms[i] = permissionsGranted_[i];
+        }
+        perms[c] = "serviceprovider.read." + getId();
+        return perms;
+      `,
       hidden: true
     }
   ],


src/foam/nanos/auth/ServiceProvider.js
```diff
@@ -52,8 +52,16 @@ foam.CLASS({
     },
     {
       name: 'permissionsGranted',
-      transient: true,
-      javaGetter: 'return new String[] { "serviceprovider.read." + getId() };',
+      networkTransient: true,
+      javaGetter: `
```
lchanmann, line 56: another idea to use `javaFactory: 'return new String[] { "serviceprovider.read" + getId() };'` instead then add "serviceprovider.read.treviso" permission to the treviso spid.

KernelDeimos: Wouldn't the extra permission in the journal entry be a bit redundant if there's already logic in here to do that?

lchanmann: no, javaFactory is only for when the service provider doesn't have granted permission.

KernelDeimos: What I mean is `"serviceprovider.read" + getId()` is already in the code somewhere, so I don't think it's good to expect that someone adding a ServiceProvider entry to the journal to know they have to add `serviceprovider.read.<id>` if and only if they happen to override the property. (or in other words, I would never expect adding `permissionsGranted: []` to a capability to remove a permission that's implied by the subclass)

lchanmann: Agreed but wonder do we have the case where someone else other than the admin is allowed to modify the granted permission on a service provider? We need a way to make the "serviceprovider.read.{spid}" permission inherent and permanent to a service provider.

lchanmann (approved):
Approved for now. Will need think about a way to make the "serviceprovider.read.{spid}" permission inherent and permanent to a service provider.

src/foam/nanos/auth/ServiceProvider.js
KernelDeimos: I believe no user will be able to remove that since it's always included by the getter. Although if users can edit these permissions then I should probably take out `networkTransient` too right?

PR #4178, opened 2020-10-28
ClientCachetranslationService getTranslation method to sync

nanopay-arthur:
#TimeSpent: 3h

### DIFF
--- a/src/foam/i18n/ClientCacheTranslationService.js
+++ b/src/foam/i18n/ClientCacheTranslationService.js
@@ -47,6 +47,13 @@ foam.CLASS({
       factory: function() {
         return (foam.locale || "").substring(3);
       }
+    },
+    {
+      class: 'Map',
+      name: 'localeEntries',
+      factory: function() {
+        return {};
+      }
     }
   ],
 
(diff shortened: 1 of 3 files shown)


src/foam/i18n/ClientCacheTranslationService.js
```diff
@@ -47,6 +47,13 @@ foam.CLASS({
       factory: function() {
         return (foam.locale || "").substring(3);
       }
+    },
```
kgrgreer, line 43: But then you should remove localeCache which isn't used anymore.

src/foam/i18n/ClientCacheTranslationService.js
```diff
@@ -84,6 +91,10 @@ foam.CLASS({
         // source: l.source,
         target: l.target
       }));
+      // populate map entries
```
kgrgreer: Remove the code above. Remove the check, otherwise you won't override language global messages with variant specific ones.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -62,6 +62,7 @@ foam.CLASS({
     'installCSS',
     'notificationDAO',
     'sessionSuccess',
+    'translationService',
```
kgrgreer: Is this needed?

PR #4180, opened 2020-10-29
Email subject translation

jinnnn9:
nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/10953

Screen Shot 2020-10-28 at 8 35 14 PM


src/foam/i18n/LocaleTranslationService.js
```diff
@@ -20,8 +29,24 @@ foam.CLASS({
...
+            EQ(Locale.SOURCE, source),
+            OR(
+              EQ(Locale.LOCALE, locale),
+              EQ(Locale.VARIANT, locale),
+              EQ(Locale.LOCALE_VARIANT, locale)
+            )
+          )
+        ).limit(1).select(new ArraySink())).getArray();
+        
+        String translation = translationInArray.size() > 0 ? translationInArray.get(0).getTarget() : originalText;
```
jinnnn9: Do we want to return the original english text or an empty string if no translation exists?

jlhughes: the original.

src/foam/i18n/LocaleTranslationService.js
```diff
@@ -20,8 +29,24 @@ foam.CLASS({
     {
       name: 'getTranslation',
       javaCode: `
-      System.err.println("*********************************** getTranslation");
-      return "foobar";
+        if ( locale.equals("en") ) return originalText;
```
jlhughes: Always translate.  Let the cache take care of this.

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -26,7 +30,7 @@ foam.CLASS({
 
   properties: [
     {
-      class: 'Long',
+      class: 'String',
```
jlhughes, line 33: This is a big change, any migration, side effects?

src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -148,7 +152,14 @@ foam.CLASS({
 
         // SUBJECT:
         if ( ! emailMessage.isPropertySet("subject") && ! SafetyUtil.isEmpty(getSubject()) ) {
```
jlhughes, line 154: Are we assuming the caller is translating when explicitly setting the property?   yeah, I think that's ok.

src/foam/nanos/notification/email/EmailTemplate.js
jlhughes, line 33: Also need to update the nspec for EmailTemplate, changing from setSeqNo to setGuid.

src/foam/i18n/LocaleTranslationService.js
kgrgreer: Yes, it is surprisingly common for people to translate english text.

src/foam/i18n/LocaleTranslationService.js
kgrgreer: This is too expensive and it is non-deterministic. If there are two translations, you don't know which one you will get from the limit(1). First look for with variant and then if you don't find look with just the locale.

src/foam/i18n/TranslationService.js
```diff
@@ -18,9 +18,26 @@ foam.INTERFACE({
     {
       name: 'getTranslation',
       async: true,
-      args: [ 'String locale', 'String source' ],
-//      javaType: 'foam.i18n.Locale'
-type: 'String'
+      args: [
+        {
```
kgrgreer, line 22: Why do we need 'x'?

jinnnn9: @kgrgreer I am calling x.get("localeDAO") in getTranslation method. Is there an alternative way that I don't need to have x as a parameter but still get localeDAO?

src/foam/i18n/TranslationService.js
```diff
@@ -18,9 +18,26 @@ foam.INTERFACE({
...
+          name: 'locale',
+          type: 'String'
+        },
+        {
+          name: 'source',
+          type: 'String'
+        },
+        {
+          name: 'originalText',
```
kgrgreer: Would "defaultText" be better?

src/foam/i18n/TranslationService.js
```diff
@@ -18,9 +18,26 @@ foam.INTERFACE({
     {
       name: 'getTranslation',
       async: true,
-      args: [ 'String locale', 'String source' ],
-//      javaType: 'foam.i18n.Locale'
-type: 'String'
+      args: [
```
kgrgreer, line 21: If you change the interface then you will also need to update the ClientCacheTranslationService implemenation.

jinn9: @kgrgreer implementation or parameters for the method on the client side?

jlhughes (approved):

src/foam/i18n/ClientCacheTranslationService.js
```diff
@@ -91,10 +91,15 @@ foam.CLASS({
     */
     {
       name: 'getTranslation',
-      args: [ 'String locale', 'String source' ],
+      args: [
+        { name: 'x', type: 'X' },
```
kgrgreer: Why is x needed?

src/foam/i18n/LocaleTranslationService.js
```diff
@@ -20,8 +29,23 @@ foam.CLASS({
     {
       name: 'getTranslation',
       javaCode: `
-      System.err.println("*********************************** getTranslation");
-      return "foobar";
+        DAO localeDAO = (DAO) x.get("localeDAO");
+
+        Locale localeEntry = (Locale) localeDAO.find(AND(
+          EQ(Locale.SOURCE, source),
+          EQ(Locale.LOCALE, locale)
```
kgrgreer: Need to also include the VARIANT here.

src/foam/i18n/LocaleTranslationService.js
```diff
@@ -20,8 +29,23 @@ foam.CLASS({
...
+        ));
+
+        if ( localeEntry == null ) {
+          localeEntry = (Locale) localeDAO.find(AND(
+            EQ(Locale.SOURCE, source),
+            EQ(Locale.VARIANT, locale)
```
kgrgreer: Need to look for LOCALE here, not VARIANT.

src/foam/i18n/TranslationService.js
kgrgreer: Call getX() instead.

src/foam/i18n/LocaleTranslationService.js
jinn9: @kgrgreer check if Locale.Locale and Locale.variant match locale? 

AND(EQ(Locale.SOURCE, source), EQ(Locale.LOCALE, locale), EQ(Locale.VARIANT, locale)) ?

src/foam/i18n/ClientCacheTranslationService.js
jinnnn9: @kgrgreer will remove

src/foam/i18n/LocaleTranslationService.js
```diff
@@ -20,8 +29,39 @@ foam.CLASS({
...
+        // following language-VARIANT format
+        boolean hasVariant = locale.contains("-");
+        String language = hasVariant ? locale.substring(0,2).toLowerCase() : locale.toLowerCase();
+        String variant = hasVariant ? locale.substring(3).toUpperCase() : "";
+        locale = hasVariant ? language + "-" + variant : locale;
+
+        Locale localeEntry = (Locale) localeDAO.find(
+          AND(
+            EQ(Locale.SOURCE, source),
+            OR(
```
kgrgreer: Don't do an OR. Then you don't know which matched. Just search for LOCALE+VARIANT first, then if you don't find, search for LOCALE + VARIANT = ''.

src/foam/i18n/LocaleTranslationService.js
```diff
@@ -20,8 +29,61 @@ foam.CLASS({
...
-      return "foobar";
+        if ( locale == "" || source == "" ) return defaultText;
+
+        DAO localeDAO = (DAO) getX().get("localeDAO");
+
+        // localeVariant - complete culture code ('en-CA', ‘fr-CA’, ‘de_AT’ …etc) 
+        // locale - determines locale type (en, fr, es …etc)        
+        // variant - Locale variation (CA for en-CA, CA for fr-CA, AT for de_AT …etc)
+        String variant = "";
+        String localeVariant = "";
```
kgrgreer: Don't use localeVariant since we are going to remove it. Just use locale and variant.

src/foam/i18n/LocaleTranslationService.js
```diff
@@ -20,8 +29,61 @@ foam.CLASS({
...
+              EQ(Locale.VARIANT, variant)
+            )
+          );
+
+          // search for locale variant
+          if ( localeEntry == null ) {
+            localeEntry = (Locale) localeDAO.find(
+              AND(
+                EQ(Locale.SOURCE, source),
+                EQ(Locale.LOCALE_VARIANT, localeVariant)
```
kgrgreer: search for AND(EQ(LOCALE, locale), EQ(VARIANT, ''))

kgrgreer: Actually, just delete this whole query, since you're doing the proper one below.

PR #4182, opened 2020-10-29
adding some code

olhabn:
so it looked like we needed to match front-end keyword to back-end keyword or otherwise on search  we find record with backend Keyword but we don't find a record with frontend Keyword on dblclick.
so lines 2422-2432 correspond to line 2478.

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -2421,13 +2421,7 @@ foam.CLASS({
             } else {
               s = prop.f(obj);
 
-              if ( ! foam.String.isInstance(s) ) {
-                if ( s.toSummary ) {
-                  s = s.toSummary();
-                } else {
-                  continue;
-                }
-              }
+              s = String(s);
             }
           } catch (err) {}
           if ( s.toLowerCase().includes(arg) ) return true;
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -2421,11 +2421,9 @@ foam.CLASS({
             } else {
               s = prop.f(obj);
 
-              if ( ! ( typeof s === "string" ) ) {
+              if ( ! foam.String.isInstance(s) ) {
                 if ( s.toSummary ) {
                   s = s.toSummary();
-                } else if ( s.toString ) {
-                  s = s.toString();
                 } else {
                   continue;
                 }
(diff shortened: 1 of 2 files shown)


src/foam/mlang/mlang.js
```diff
@@ -2418,10 +2418,18 @@ foam.CLASS({
...
             } else {
-              s = prop.f(obj).toLowerCase();
+              s = prop.f(obj);
+
+              if ( ! ( typeof s === "string" ) ) {
+                if ( s.toSummary ) {
+                  s = s.toSummary();
+                } else if ( s.toString ) {
+                  s = s.toString();
+                } else {
```
lchanmann: I don't think keyword should test toSummary() and toString()

src/foam/mlang/mlang.js
```diff
@@ -2418,10 +2418,18 @@ foam.CLASS({
               s = prop.f(obj).toString().toLowerCase();
             } else if ( this.Date.isInstance(prop) ) {
               s = prop.f(obj).toISOString().toLowerCase();
-            } else if ( ! this.String.isInstance(prop) ) {
-              continue;
             } else {
-              s = prop.f(obj).toLowerCase();
+              s = prop.f(obj);
+
+              if ( ! ( typeof s === "string" ) ) {
```
kgrgreer: foam.String.isInstance(s) is better

src/foam/mlang/mlang.js

src/foam/mlang/mlang.js

src/foam/mlang/mlang.js
```diff
@@ -2418,10 +2418,10 @@ foam.CLASS({
               s = prop.f(obj).toString().toLowerCase();
             } else if ( this.Date.isInstance(prop) ) {
               s = prop.f(obj).toISOString().toLowerCase();
-            } else if ( ! this.String.isInstance(prop) ) {
-              continue;
             } else {
-              s = prop.f(obj).toLowerCase();
+              s = prop.f(obj);
```
kgrgreer, line 2422: Would this be better:
s = String.valueOf(prop.f(obj)).toLowerCase();

olhabn: it doesn't work for me for some reason. but i can investigate

PR #4187, opened 2020-10-29
[DO_NOT_MERGE] [NP-2552] Add ucj listener to CachingAuthService

KernelDeimos:


jlhughes (approved):

src/foam/nanos/auth/CachingAuthService.java
```diff
@@ -12,6 +12,7 @@
 import foam.dao.Sink;
 import foam.mlang.predicate.Predicate;
 import foam.nanos.auth.User;
+import foam.nanos.crunch.CrunchService;
```
KernelDeimos: One thing I don't like about this is it makes CachingAuthService dependant on CrunchService. Maybe AuthService should have a `cmd_` method like DAOs do so that CachingAuthService could provide `purgeSink` to its delegates through a command. @kgrgreer

kgrgreer: Good thinking, but, the way it is now we have a predicate so we only purge when a specific user's cache needs to change. A generic purgeAuthCach CMD would lose this functionality and would constantly flood caches everytime a user/group/ucj was updated, which would be always. Could be done, but would still require some filtering.

jlhughes: Why can't the CapabilityJunctionDAO independently 'listen' on these DAOs.  Small duplication of code, but no dependencies.

KernelDeimos: CapabilityJunctionDAO is being listened to here by the AuthService

jlhughes: @KernelDeimos yes, our issue is 'here'.  We don't want CachingAuthService dependent/aware of Crunch.  There is a CapabilityAuthService is there not?

KernelDeimos: I've updated the PR, but leaving `DO_NOT_MERGE` because it doesn't work yet. I'm using `session.setContext` like in CachingAuthService, yet somehow when the same code gets run later `session.getContext` returns a stale version of the context.

KernelDeimos: Also merging master somehow broke the PR diff... I'll fix that tomorrow(/technically today)

PR #4189, opened 2020-10-29
Section label wizardlet

zahnno:
Fixes issue when title of capability payload is a function

### DIFF
--- a/src/foam/u2/wizard/StepWizardletStepsView.js
+++ b/src/foam/u2/wizard/StepWizardletStepsView.js
@@ -56,7 +56,7 @@ foam.CLASS({
   ],
 
   messages: [
-    { name: 'PART_LABEL', message: 'Part' }
+    { name: 'PART_LABEL', message: 'Part ' }
   ],
 
   methods: [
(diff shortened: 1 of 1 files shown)


src/foam/u2/wizard/StepWizardletStepsView.js
```diff
@@ -168,15 +167,22 @@ foam.CLASS({
         }))
     },
     function renderSectionLabel(elem, section, index, isCurrent) {
-      let title = section.navTitle;
-      if ( ! title || ! title.trim() ) title = "Part " + index;
-      return elem       
+      let title = section.title;
+      if ( ! title || ! foam.Function.isInstance(title) && ! title.trim() ) title = "Part " + index;
```
kgrgreer: Could you not call .add() in this 'if' block to avoid creating a slot when not needed?

kgrgreer: Should "Part" come from a Message?

zahnno: Yup

zahnno: A slot isn't created if title isn't a function. I've renamed the var to avoid confusion

PR #4190, opened 2020-10-29
Make getJunction never return null

KernelDeimos:


KernelDeimos:
If this is merged then revert https://github.com/foam-framework/foam2/pull/4191

src/foam/nanos/crunch/MinMaxCapability.js
```diff
@@ -71,6 +71,7 @@ foam.CLASS({
 
           UserCapabilityJunction ucJunction =
             crunchService.getJunction(x, ccJunction.getTargetId());
+          if ( ucJunction.getStatus() == AVAILABLE ) continue;
```
Nauna, line 74: think we can and should delete the line below that does the null check

src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -224,51 +224,32 @@ public UserCapabilityJunction getJunctionForSubject(
...
 
       // Check if a ucj implies the subject.realUser has this permission in relation to the user
-      UserCapabilityJunction ucj = (UserCapabilityJunction)
+      var ucj = (UserCapabilityJunction)
         userCapabilityJunctionDAO.find(AND(associationPredicate,targetPredicate));
-      if ( ucj != null ) {
-        return ucj;
+      if ( ucj == null ) {
+        ucj = buildAssociatedUCJ(x, capabilityId, subject);
+        ucj.setStatus(CapabilityJunctionStatus.AVAILABLE);
```
Nauna: I think it makes sense to have the buildAssociatedUCJ function set the ucj status to AVAILABLE. Then can avoid setting it after the calling code.

src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -224,51 +224,32 @@ public UserCapabilityJunction getJunctionForSubject(
...
 
+      return ucj;
     } catch ( Exception e ) {
       Logger logger = (Logger) x.get("logger");
       logger.error("getJunction", capabilityId, e);
-    }
 
-    return null;
+      // On failure, report that the capability is available
+      var ucj = buildAssociatedUCJ(x, capabilityId, subject);
```
Nauna, line 239: buildAssociatedUCJ throws an error. So my concern is the error thrown from the catch would not be caught and might cause unexpected behavior.

Perhaps you should remove blocks on line 230-231 and 240-241 and the ucj return - then run those commands in a separate try catch after the first.

Nauna (approved):

src/foam/nanos/crunch/MinMaxCapability.js
KernelDeimos: Agreed - thanks!

src/foam/nanos/crunch/ServerCrunchService.java
KernelDeimos: Good catch, I didn't think of that

src/foam/nanos/crunch/ServerCrunchService.java
KernelDeimos: I think if this is ever called with a capability that doesn't exist a runtime exception is exactly what's expected. I think catching the error and logging it would actually be worse in this situation because it would be necessary to return something, and then what would be returned? Could be a null UCJ, then we'd need to add the null checks back but then in every place we currently do that the check would just be a condition to throw a subsequent exception. Could be a UCJ with status AVAILABLE, but that implies that it's safe to overwrite that UCJ.

PR #4192, opened 2020-10-29
Inherit model actions and Table view item changes

zahnno:
InheritModelActions provides the option to inherit data model actions into the context menu actions.
This PR give dev's an option to include model actions as context menu actions in a tableview instead of being forced.

Also added - If no actions are available to a table view item, no vertical ellipsis will be shown.


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -123,6 +123,12 @@ foam.CLASS({
         model of the object being shown in the table.
       `
     },
+    {
+      class: 'Boolean',
+      name: 'inheritModelActions',
```
kgrgreer: When would we not want to do this?

src/foam/u2/view/OverlayActionListView.js
```diff
@@ -152,6 +152,8 @@ foam.CLASS({
     async function initE() {
       var self = this;
       this.onDetach(this.active_$.follow(this.overlay_.opened$));
+      this.data.map((action) => action.createIsAvailable$(this.__context__, this.obj));
+      if ( ! this.data.length > 0 ) return;
```
kgrgreer, line 156: Nice! This has been bothering me.

src/foam/u2/view/UnstyledTableView.js
zahnno: ~~Sub classes - In context of a User, if I subclass User to let's say a Contact and display those contacts within a table view, I may not want to show the actions of the User class.~~

kgrgreer: Is this a real or hypothetical requirement?

kgrgreer:
Has merge conflict.

src/foam/u2/view/UnstyledTableView.js
zahnno: Let me reiterate since I don't believe I explained this change properly.

The tableview defaults to including its data model actions _**and**_ contextMenuActions defined on the table view instance. This property changes that behaviour by dictating whether those should be combined or not. I could choose to opt out of my table view using its data model actions and solely use the contextMenuActions provided in the tableview instance.

I did it for a reason but I can't recall why since this was a ways back.

kgrgreer: So, it isn't really "inherit"ModelActions but "user"ModelActions?

zahnno: useModelActions, right.

zahnno: or how about utilizeModelActions HAH

PR #4200, opened 2020-10-30
CapabilityPayload REJECT + Cleanup of Approval Requests upon Status Change + Fixed MinMax ChainedStatus

tharmaman:


src/foam/nanos/crunch/lite/ruler/CapableStatusChangeAdjustApprovalsRuleAction.js
```diff
@@ -0,0 +1,135 @@
...
+
+        Capable capableNewObj = (Capable) obj;
+        Capable capableOldObj = (Capable) oldObj;
+
+        CapablePayload[] newCapablePayloads = capableNewObj.getCapablePayloads();
+        CapablePayload[] oldCapablePayloads = capableOldObj.getCapablePayloads();
+
+        List<CapablePayload> updatedApprovalPayloads = new ArrayList<>();
+
+        if ( newCapablePayloads.length != oldCapablePayloads.length ){
```
nick-prat, line 59: You have this as an after rule, so I think if this gets run twice, the second time it will accept it.

src/foam/nanos/crunch/lite/CapablePayloadStatusDAO.js
```diff
@@ -60,9 +60,33 @@ foam.CLASS({
         if ( oldStatus == APPROVED && newStatus == PENDING ) {
           newStatus = APPROVED;
         }
+
+        if ( oldStatus == REJECTED  ) {
+          // TODO: if rejected, need to change the status of all children (prereqs) as well to rejected
```
KernelDeimos: is this true? What if a prerequisite is common to another capability that is granted?

KernelDeimos (approved):

src/foam/nanos/crunch/lite/CapablePayloadStatusDAO.js
tharmaman: Yeah so lol this isn't a todo anymore forgot to remove it. It exists for line 66 onwards. This is only preqreqs on the capable object though right so it will only happen for that object. Regardless if any of the higher levels get rejected on the invoice then everything really is rejected.

src/foam/nanos/crunch/lite/MinMaxCapabilityRefinement.js
```diff
@@ -51,15 +52,31 @@ foam.CLASS({
...
+        int numberTotal = numberGranted + numberPending + numberRejected;
+
+        int numberGrantedOrPending = numberGranted + numberPending;
+
+        // TODO: right now this represents the capablePayloads haven't been processed for preqreqs yet
+        if ( numberTotal == 0 ){
+          return CapabilityJunctionStatus.ACTION_REQUIRED;
+        }
+
+        if ( getMin() > numberGrantedOrPending ){
```
jlhughes, line 87: Don't understand this one.

tharmaman: NumberGrantedorPending are basically the available Payloads which are GRANTED or can eventually be turned into GRANTED. If the MIN is greater than the number of available payloads which are GRANTED or can eventually be turned into GRANTED, then it is impossible for the total amount of GRANTED payloads to be greater than the MIN, thereby fulfilling the minimum requirement.

Imagine theres a min max which has 10 prerequisites attached and a min of 2.

If the user selected only 3 of those prereqs in the wizard, then he will get 3 approvals generated all in PENDING state.

If 1 of those 3 get rejected. Then we will have 2 numberGrantedOrPending which could still potentially satisfy the min requirement of 2  if those 2 get approved and thereby set to GRANTED.

If 2 of those 3 get rejected. then we will have 1 numberGrantedOrPending which would be impossible to satisfy the min requirement of 2 even if that 1 payload is GRANTED.

src/foam/nanos/crunch/lite/ruler/CapableStatusChangeAdjustApprovalsRuleAction.js
tharmaman: Resolved by CapableDAO cleanup

nick-prat (approved):

PR #4203, opened 2020-11-02
menuDAO order by ORDER and LABEL

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2569

## Changes
- Order JS menuDAO by order and label properties same as the server-side


kgrgreer:
Why are you doing this? Shouldn't it be ordered by order?

PR #4205, opened 2020-11-02
Java template

smirnovaka:


src/foam/core/templates.js
```diff
@@ -153,17 +116,112 @@ foam.CLASS({
...
     {
       name: 'simple',
       value: true
-    }
+    },
+    'args'
   ],
 
   methods: [
+    function getAbstractGrammar() {
```
kgrgreer: Can you rename this to createAbstractGrammar?

src/foam/core/templates.js
```diff
@@ -193,6 +251,16 @@ foam.CLASS({
       return f;
     },
 
+    function compileJava(t, name, args) {
+      this.args = args;
+      var result = this.javaGrammar.parseString(t);
+      if ( ! result ) throw "Error parsing template " + name;
+
+      var code =( result[0] ? t : result[1] );
```
kgrgreer: Is the result[0] check needed?

src/foam/core/templates.js
```diff
@@ -153,17 +116,112 @@ foam.CLASS({
...
+          },
+          'simple value': function(v) {
+            self.push("',\n self.",
+                v[1].join(''),
+                v[2],
+                ",\n'");
+          },
+          'raw values tag': function (v) {
+            var value = v[1].join('').replace(/\s/g, '');
+            var argExist = self.args.map(ar => ar.name).includes(value);
```
kgrgreer: Would be faster to create a property called argSet_ that contains the names of args, so you don't need to keep rebuilding this for every value.

PR #4206, opened 2020-11-02
Add object actions to context menu

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2595

## Issues
- Nature code approval request has its own actions override but the context menu on approval request table view doesn't include them

## Changes
- Add object actions to context menus on table view

### DIFF
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -406,15 +406,12 @@ foam.CLASS({
           var view = this;
 
           var actions = {};
-          var actionMapper = function(ret, action) {
-            ret[action.name] = action;
-            return ret;
-          };
+          var actionsMerger = action => { actions[action.name] = action; };
 
           // Model actions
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -405,10 +405,16 @@ foam.CLASS({
...
-            ? view.contextMenuActions.concat(modelActions)
-            : modelActions;
+          var actions = {};
+          var actionMapper = function(ret, action) {
+            ret[action.name] = action;
+            return ret;
+          };
+
+          // Model actions
+          view.of.getAxiomsByClass(foam.core.Action).reduce(actionMapper, actions);
```
kgrgreer: Why using reduce()?  Wouldn't forEach() be simpler since you don't actually replace 'actions'?

lchanmann: You're right. Fixed in https://github.com/foam-framework/foam2/pull/4206/commits/103c41c008c4a2a1215aea6603b24981e0313f7c.

PR #4211, opened 2020-11-03
DNM - fix certifydatareviewed not updating after permissions gained

noodlemoodle:
https://nanopay.atlassian.net/browse/NP-2552


noodlemoodle:
fixed in https://github.com/nanoPayinc/NANOPAY/pull/10996 instead

PR #4212, opened 2020-11-03
Approve without memo ApprovalRequest

blakegreen32:
Added approve without memo action to ApprovalRequest with listener method.

Ticket: https://nanopay.atlassian.net/browse/NP-2592


jlhughes:
.

jlhughes (approved):

PR #4219, opened 2020-11-04
NP-2607 BenchmarkRunner updates.

jlhughes:
- collect stats for passed and failed runs
- setup prefix logger
- correct per thread calculations.


src/foam/nanos/bench/BenchmarkRunner.java
```diff
@@ -145,43 +152,65 @@ public String getResult() {
...
-    if ( runPerThread_ ) {
-      threadCount_ = 1;
-      if ( reverseThreads_ ) {
-        threadCount_ = availableThreads;
-      }
+    int threads = 1;
+    if ( ! getRunPerThread() ) {
+      availableThreads = 1;
+    } else if ( reverseThreads_ ) {
+        threads = availableThreads;
```
kgrgreer, line 173: You can't reverseThreads unless you enable getRunPerThread()?

jlhughes: when runPerThread is false - we only do one run.  
But what's missing is  runPerThread = false, but still do one run with all threads.

nanoscott (approved):

PR #4223, opened 2020-11-04
Update ApprovalRequest.js

Nauna:
fix bug with not showing actions.
and
a bunch of syntax updates

### DIFF
--- a/src/foam/nanos/approval/ApprovalRequest.js
+++ b/src/foam/nanos/approval/ApprovalRequest.js
@@ -169,7 +169,7 @@
             slot.set(`User #${approver}`);
           }
         });
-        
+
         return {
           class: 'foam.u2.view.ValueView',
           data$: slot
@@ -206,7 +206,6 @@
         var userId = parseInt(objId);
         if ( !! userId ) {
           this.__subSubContext__.userDAO.find(userId).then(function(a) {
-
           if ( a != undefined ) {
             self.add(a.toSummary());
           } else {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -542,59 +540,54 @@
...
       },
       code: function(X) {
-        var objToAdd = X.objectSummaryView ? X.objectSummaryView : X.summaryView;
+        var objToAdd = X.objectSummaryView ?
+          X.objectSummaryView : X.summaryView;
         objToAdd.add(this.Popup.create({ backgroundColor: 'transparent' }).tag({
-          class: "foam.u2.MemoModal",
-          onExecute: this.approveWithMemo.bind(this, X)
+          class: 'foam.u2.MemoModal',
+          onExecute: this.approveWithMemoL.bind(this, X)
```
blakegreen32, line 574: approveWithMemoL ? Is this because the name of the listener can't be the same as the action?

Nauna: yes - I think it's terribly confusing having an action and listener with the same name.

kgrgreer: You can't have them with the same name. FOAM might not detect the error, but things won't work correctly.
You don't need to bind() listeners to this. They're pre-bound.

blakegreen32 (approved):

PR #4227, opened 2020-11-04
add place holder for country slection

yij793:
address: https://nanopay.atlassian.net/browse/NP-2586
nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/11017
Screen Shot 2020-11-04 at 3 43 21 PM

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -21,7 +21,7 @@ foam.CLASS({
   ],
 
   messages:[
-    { name: 'PLACEHOLDER', message: 'please select...'}
+    { name: 'PLACEHOLDER', message: 'Please select...'}
   ],
 
   properties: [
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -20,10 +20,6 @@ foam.CLASS({
     'foam.core.Latch'
   ],
 
-  messages:[
-    { name: 'PLACEHOLDER', message: 'Please select...'}
-  ],
-
   properties: [
     {
       class: 'foam.core.FObjectProperty',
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/FObjectView.js
```diff
@@ -20,6 +20,10 @@ foam.CLASS({
     'foam.core.Latch'
   ],
 
+  messages:[
+    { name: 'PLACEHOLDER', message: 'please select...'}
```
jlhughes:
```suggestion
    { name: 'PLACEHOLDER', message: 'Please select...'}
```

zahnno:
lgtm

src/foam/u2/view/FObjectView.js
```diff
@@ -203,7 +209,6 @@ foam.CLASS({
       );
 
       if ( this.data ) { this.objectClass = dataToClass(this.data); }
-      if ( ! this.data && ! this.objectClass && this.choices.length ) this.objectClass = this.choices[0][0];
```
kgrgreer, line 206: Why delete this?

yij793: @kgrgreer For my understanding, this to set objectClass select the first one in the array. I remove it  beacuse we don't want select anything by default

jlhughes: @yij793 'don't want to select anything by default' - for your use case or all use cases?  
We use default selection elsewhere. 
If you need this support then it's a feature we need to add.

yij793: @jlhughes hmm, in that case. I'll make change to make it flexible

src/foam/u2/view/FObjectView.js
```diff
@@ -97,6 +103,12 @@ foam.CLASS({
       name: 'dataWasProvided_',
       documentation: 'Set to true if data was initially provided. Used to implement classIsFinal.'
     },
+    {
+      class: 'Boolean',
+      name: 'selectByDefault',
+      documentation: 'If true, the choiceView will select the first element',
+      value: 'true'
```
kgrgreer: Make make the default value be:
expression: function(placeholder) { return ! placeholder; }
Or, get rid of this property entirely and just have it always select the first by default if placeholder isn't set.

PR #4229, opened 2020-11-05
Do not output id if it is storage transient

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2618

## Issues
- Multipart id property is generated as storage transient and should not be written to the journal

## Changes
- Check storage transient to output id property

### DIFF
--- a/src/foam/lib/formatter/JSONFObjectFormatter.java
+++ b/src/foam/lib/formatter/JSONFObjectFormatter.java
@@ -339,11 +339,14 @@ public class JSONFObjectFormatter
     }
     addInnerNewline();
     PropertyInfo id = (PropertyInfo) newInfo.getAxiomByName("id");
-    if ( ! id.getStorageTransient() ) outputProperty(newFObject, id);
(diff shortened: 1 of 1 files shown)


src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -339,8 +339,7 @@ public void outputDelta(FObject oldFObject, FObject newFObject, ClassInfo defaul
     }
     addInnerNewline();
     PropertyInfo id = (PropertyInfo) newInfo.getAxiomByName("id");
-    outputProperty(newFObject, id);
-
+    if ( ! id.getStorageTransient() ) outputProperty(newFObject, id);
```
kgrgreer: If we don't output the id, then are we going to get an extra "," output on line 346?

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4229/commits/fb36c41ba46989d1b99a2fdcdf9324a9684cf5c0.

PR #4232, opened 2020-11-05
[NP-2612] make username case insensitive

jinnnn9:
address: https://nanopay.atlassian.net/browse/NP-2612
nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/11024

summary:
- store usename in lowercase
- convert the username input into lowercase before checking against the usernames in the database


jlhughes (approved):

kgrgreer:
Wouldn't it be easier if we just put the toLowerCase on the preSet and javaPreSet?

PR #4242, opened 2020-11-07
Fix service provider permissionGranted containing duplicate "serviceprovider.read.<SPID>" permission

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2639

## Issues
- Service provider permissionGranted containing duplicate "serviceprovider.read.<SPID>" permission

## Changes
- Add inherent permissions property to capability
- Set inherent permissions for service provider to "serviceprovider.read.<SPID>"

### DIFF
--- a/src/foam/nanos/auth/ServiceProvider.js
+++ b/src/foam/nanos/auth/ServiceProvider.js
@@ -51,18 +51,9 @@ foam.CLASS({
       `
     },
     {
-      name: 'permissionsGranted',
-      networkTransient: true,
-      javaGetter: `
-        int c = permissionsGranted_ != null ? permissionsGranted_.length : 0;
-        String[] perms = new String[c + 1];
-        for ( int i = 0 ; i < c ; i++ ) {
-          perms[i] = permissionsGranted_[i];
-        }
-        perms[c] = "serviceprovider.read." + getId();
-        return perms;
-      `,
-      hidden: true
+      name: 'inherentPermissions',
+      javaGetter: 'return new String[] { "serviceprovider.read." + getId() };',
+      documentation: 'Service provider must have "serviceprovider.read.<SPID>" inherent permission.',
     }
   ],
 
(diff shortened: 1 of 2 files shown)


lchanmann:
@kgrgreer @jlhughes @KernelDeimos please review, thanks.

KernelDeimos:
Just to clarify, what's the use of this change? It seems it hides the inherent permission when you call `getPermissionsGranted()` but otherwise works the same.

kgrgreer:
LGTM

lchanmann:
@KernelDeimos the inherent permission e.g, "serviceprovider.read.treviso" was appended to the permission granted property several times - https://nanopay.atlassian.net/browse/NP-2639.

Step to reproduce:
1. Start the server
2. Login as admin
3. Create a server-side script to get permissionsGranted of treviso service provider.

Result:
Screen Shot 2020-11-08 at 4 18 53 PM


> It seems it hides the inherent permission when you call `getPermissionsGranted()`

The idea is to have "serviceprovider.read.\<SPID\>" to be the inherent permission on ServiceProvider and it is different from the permissionsGranted which can just be regular configuration.

Nauna:
Hmmm - I don't like the idea of this cause it adds yet another property on capability that needs to be maintain. Our models have a tendency to grow to quickly.

The issue is simply the function getter on the property permissionsGranted on the service provider model.

Every time the function is called it adds `perms[c] = "serviceprovider.read." + getId();`

I'm not suggesting not to accept this - but personally think there might be a better way. A property on ServiceProvider that controls the serviceProvider permissionsGranted.getter only adds the permission once is a simpler property to maintain.
Or change this from a getter to something that sets this property once.

jlhughes:
Continuing from @Anna, @KernelDeimos - we are flipping back and forth here, so need to stop and think about.

lchanmann:
IMO, I think it makes it harder to reason when we tamper the service provider to have a different logic for permission granted property than other capabilities and we might have to mess with another capability the same way if need something similar for that capability. Separating the permission granted which can be configured and the inherent ones would make the intent explicit and avoids the possible interferences when the permission granted is get or set.

PR #4243, opened 2020-11-08
Fix script output update

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2640

## Changes
- Run client-side polling on SCHEDULED status because script is run in async
- Turn off client-side caching for scriptDAO
- Update script output on UNSCHEDULED or ERROR statuses

### DIFF
--- a/src/foam/nanos/script/Script.js
+++ b/src/foam/nanos/script/Script.js
@@ -397,7 +397,7 @@ foam.CLASS({
         if ( this.server ) {
           this.__context__[this.daoKey].put(this).then(function(script) {
             self.copyFrom(script);
-            if ( script.status === self.ScriptStatus.RUNNING ) {
+            if ( script.status === self.ScriptStatus.SCHEDULED ) {
               self.poll();
             }
           });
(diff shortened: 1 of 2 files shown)


src/foam/nanos/script/Script.js
```diff
@@ -360,7 +360,9 @@ foam.CLASS({
         var self = this;
         var interval = setInterval(function() {
           self.__context__[self.daoKey].find(self.id).then(function(script) {
-            if ( script.status !== self.ScriptStatus.RUNNING ) {
+            if ( script.status === self.ScriptStatus.UNSCHEDULED
```
jlhughes, line 363: Is this necessary in combination with turning off client side caching?

lchanmann: Yes since the script might be still in SCHEDULED or RUNNING.

jlhughes: Ok, because async, may not be running yet. 
And we don't bother polling RUNNING, as there are no updates.

jlhughes (approved):

PR #4244, opened 2020-11-09
Clean up Locale Language 

xuerongNanopay:
- move languageChoiceView to foam.nano.auth
- Store prefer language to user.language
- Get language dropdown from languageDao


src/foam/nanos/auth/LanguageChoiceView.js
```diff
@@ -0,0 +1,237 @@
...
+        return language
+      }
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this.updateE()
+    },
+    async function updateE() {
```
kgrgreer: Why is this not just in initE()?

xuerongNanopay: initE() can not use async

jlhughes (approved):

src/foam/nanos/auth/LanguageChoiceView.js
kgrgreer: Why not?

xuerongNanopay: I got error when I have async in initE previously. But I forgot what the reason is.  I will try again, see if it will work with async in initE.

PR #4247, opened 2020-11-09
[NP-2564] help icon odd behaviour fix

jinnnn9:
address: https://nanopay.atlassian.net/browse/NP-2564
summary:
- when you hovered over the help icon, there was not enough space for the message to be displayed so the wizard was scrolled up and down.
- added enough padding to the entry bottom so that there is enough space for the message to be displayed
- removed padding to the button top so that the button and the last entry in the wizard are not too distant.
- make the help text left-justified

Screen Shot 2020-11-09 at 12 42 37 PM


kgrgreer:
Can you make it so that the text is also left-justified?

kennyqiyenkan (approved):
As kevin said. Otherwise I can agree with the approach.

PR #4248, opened 2020-11-09
[NP-2436] ImplyReviewedAction (+ CrunchController update)

KernelDeimos:
This rule action is somewhat untested because I haven't got the rule to work property yet


src/foam/nanos/crunch/extra/ImplyReviewedAction.js
```diff
@@ -0,0 +1,87 @@
...
+
+          // Find the review wizard corresponding to this capability.
+          //   A list is used in case there is more than one.
+          List reviewWizards = ((ArraySink) capabilityDAO.where(AND(
+            INSTANCE_OF(ReviewWizard.class),
+            EQ(ReviewWizard.CAPABILITY_TO_REVIEW, cap.getId())
+          )).select(new ArraySink())).getArray();
+          for ( Object o : reviewWizards ) {
+            var reviewWizard = (ReviewWizard) o;
+            grantReview(x, reviewWizard);
```
jlhughes, line 54: Is the predicate of this rule that this capability is Granted?

KernelDeimos: The predicate can be anything. (on NP-side it's the condition of having SO privilages prior to completing onboarding)

jlhughes (approved):

src/foam/nanos/crunch/UserCapabilityJunctionRefine.js
```diff
@@ -169,6 +179,56 @@ foam.CLASS({
...
+            "Tried to call getSubject() on UCJ with unrecognized capability");
+        }
+
+        if ( ucj.getSourceId() == currentSubject.getUser().getId() ) {
+          subject.setUser(currentSubject.getRealUser());
+          subject.setUser(currentSubject.getUser());
+          return subject;
+        }
+
+        if ( cap.getAssociatedEntity() == USER ) {
```
Nauna, line 224:
```suggestion
        if ( cap.getAssociatedEntity() == AssociatedEntity.USER ) {
```

src/foam/nanos/crunch/UserCapabilityJunctionRefine.js
```diff
@@ -169,6 +179,56 @@ foam.CLASS({
...
+          subject.setUser((User) userDAO.find(
+            0 != ucj.getLastUpdatedRealUser()
+              ? ucj.getLastUpdatedRealUser()
+              : ucj.getSourceId()
+          ));
```
Nauna, line 229: suggest remove:
          subject.setUser((User) userDAO.find(
            0 != ucj.getLastUpdatedRealUser()
              ? ucj.getLastUpdatedRealUser()
              : ucj.getSourceId()
          ));

replace with:
          if ( ucj.getLastUpdatedRealUser() != 0 ) {
            subject.setUser((User) userDAO.find(ucj.getLastUpdatedRealUser());
          }

This suggestion is easier to read

Further I wonder... getLastUdpatedRealUser() ...

So we are in the situation where ucj.sourceId is a business, and lastUpdatedRealUser is someone in this business or potentially a system user id. 

question - what if the last person to update this ucj is not related to the business?
And actually if the sourceId is for the business ... there isn't a way to get the realUser. If person A changed the ucj and person B is logged in ... this subject is falsey. Or what if system user is calling this method ... they would arbituraly get back a user of the business... which makes me wonder if this will be a problem.

Perhaps documentation...

src/foam/nanos/crunch/extra/ImplyReviewedAction.js
```diff
@@ -0,0 +1,90 @@
...
+
+          var ucj = (UserCapabilityJunction) obj;
+          var cap = (Capability) capabilityDAO.find(ucj.getTargetId());
+
+          // Create UCJ subject context
+          x = x.put("subject", ucj.getSubject(x));
+
+          // Find the review wizard corresponding to this capability.
+          //   A list is used in case there is more than one.
+          List reviewWizards = ((ArraySink) capabilityDAO.where(AND(
```
Nauna, line 48: naming suggestion:
ReviewWizard -> ReviewOtherCapCapability

ReviewWizard makes me think client ui.

src/foam/nanos/crunch/UserCapabilityJunctionRefine.js
KernelDeimos: There's a static import for this

src/foam/nanos/crunch/extra/ImplyReviewedAction.js
KernelDeimos: ReviewWizard is the top-level capability for a group of ReviewCapability, so in a way it does represent the wizard.

src/foam/nanos/crunch/UserCapabilityJunctionRefine.js
KernelDeimos: The intention is that this is only set when the capability will be granted by an approval but you still need to know something about the `realUser` when it's granted.

Specifically for NP, the condition we have to auto-grant the review capabilities requires information about this user, but it can only happen after approval has been granted.

KernelDeimos: In other words what we want is "what user last changed this capability as the business - do their privileges imply a review?" I haven't thought of a better solution to this problem yet.

Nauna (approved):

PR #4257, opened 2020-11-10
modified height on filter drop down

minsunny:
https://nanopay.atlassian.net/browse/NP-13

increasing the select size function works well but this below applied so it was prevented.
^ .foam-u2-tag-Select { height: /*%INPUTHEIGHT%*/ 40px;}
replaced this height to auto.

Screen Shot 2020-11-10 at 3 55 25 PM
Screen Shot 2020-11-10 at 3 57 46 PM
Screen Shot 2020-11-10 at 4 13 08 PM

### DIFF
--- a/src/foam/u2/search/GroupBySearchView.js
+++ b/src/foam/u2/search/GroupBySearchView.js
@@ -32,7 +32,9 @@ foam.CLASS({
   css: `
     ^ select {
       min-width: 220px;
-      min-height: 100px;
+    }
+    ^ .foam-u2-tag-Select {
+      height: auto;
     }
   `,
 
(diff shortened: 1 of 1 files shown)


kgrgreer:
When there is only one entry, like shown in your screenshot above, it looks odd that it has so much extra space. Can you try to find an approach which makes it larger when it has more entries, but isn't oversized when it doesn't? Maybe try increasing the select size attribute.

jlhughes:
Nice. 
Is there still a max height to deal with large lists?

minsunny:
> Nice.
> Is there still a max height to deal with large lists?

@jlhughes  : the max height - max entities can be shown up to 10 (placeholder inclusive)


Screen Shot 2020-11-10 at 4 31 52 PM
Screen Shot 2020-11-10 at 4 32 13 PM

kgrgreer:
Much better. Great.

PR #4272, opened 2020-11-11
translate wizardlet

hchoura:


jlhughes:
@hchoura can you update the wizard dismissal text: 

> Are you sure you want to leave this screen?
> Go back | Save and exit

Thanks - came out of the bug hunt today.

jlhughes (approved):

PR #4276, opened 2020-11-11
NSpec AltFactory

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2682

## Changes
- Add AltFactory to support composing service with multiple nspec keys and predicates
- Add ContextUserPredicate to support checking predicate against the current user in the context


jlhughes (approved):
I like it.

PR #4281, opened 2020-11-12
LanguageChoiceView Fixes

blakegreen32:
Ticket:  https://nanopay.atlassian.net/browse/NP-2669
Screen Shot 2020-11-12 at 4 15 12 PM

### DIFF
--- a/src/foam/nanos/auth/LanguageChoiceView.js
+++ b/src/foam/nanos/auth/LanguageChoiceView.js
@@ -168,7 +168,7 @@ foam.CLASS({
       let country = await this.countryDAO.find(this.lastLanguage.variant)
       let label = this.lastLanguage.variant != "" ? `${this.lastLanguage.name}(${this.lastLanguage.variant})` : `${this.lastLanguage.name}`
       if ( country && country.name != null ) {
-        label = `${this.lastLanguage.name}(${this.getCountryName(country.name, country.name)})`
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/LanguageChoiceView.js
```diff
@@ -183,7 +186,7 @@ foam.CLASS({
       let country = await this.countryDAO.find(language.variant)
       let label = language.variant != "" ? `${language.name}(${language.variant})` : `${language.name}`
       if ( country && country.name != null ) {
-        label = `${language.name}(${this.getCountryName(country.name, country.name)})`
+        label = `${language.name}\u00A0(${this.getCountryName(country.name, country.name)})`
```
kennyqiyenkan, line 188: Does this need unicode? A backtick should already detect a space.

kennyqiyenkan (changes requested):
Could you provide a small screenshot as well? Better to see the changes

blakegreen32:
Added screenshot

src/foam/nanos/auth/LanguageChoiceView.js
blakegreen32: Xuerong had suggested that but it caused the string to break and overlap. After looking online I found that unicode and it worked fine.

kennyqiyenkan (approved):

PR #4283, opened 2020-11-12
[DO_NOT_MERGE] [NP-2638] Filedao view is broken

KolombetM:
https://nanopay.atlassian.net/browse/NP-2638

### DIFF
--- a/src/foam/dao/KeyValueDAO.js
+++ b/src/foam/dao/KeyValueDAO.js
@@ -34,13 +34,6 @@ foam.CLASS({
       javaCode: 'return obj;'
     },
     {
-      name: 'select_',
-      code: function(sink) {
-        return Promise.resolve(sink);
-      },
-      javaCode: 'return sink;'
-    },
-    {
       name: 'removeAll_',
       code: function() {
         return Promise.resolve();


jlhughes:
@Mykola97 why are we supporting select on fileDAO - this makes no sense.

KolombetM:
> @Mykola97 why are we supporting select on fileDAO - this makes no sense.

Without select we can't check fileDAO data from backoffice. On the other hand, find operation is unauthorized, so it doesn’t make sense to allow select. Should we just ignore the fact, that we can't check fileDAO from UI?

jlhughes:
> > @Mykola97 why are we supporting select on fileDAO - this makes no sense.
> 
> Without select we can't check fileDAO data from backoffice. On the other hand, find operation is unauthorized, so it doesn’t make sense to allow select. Should we just ignore the fact, that we can't check fileDAO from UI?

'check fileDAO data from backoffice'?  that's ridiculous - to look a table view of hashcodes.   Select will not be supported.  If you have a hash, you can find it.  If they need 'find' support then we can add that, but unfiltered select will not be supported.  
An option is to implement select_ and check that predicate is not just true, and then call find_ with the predicate to only return one result.  Then the advanced filter one could add the file hash as the id they want to find, and pasting a file hash in the search bar would do the same.

PR #4286, opened 2020-11-13
DNM - fix blank screen after general admission

noodlemoodle:
https://nanopay.atlassian.net/browse/NP-2680


noodlemoodle:
fix is elsewhere

PR #4288, opened 2020-11-13
Until parser

smirnovaka:
current Js Until parser works as expected

### DIFF
--- a/src/foam/lib/parse/Until.java
+++ b/src/foam/lib/parse/Until.java
@@ -16,7 +16,7 @@ public class Until
   }
 
   public PStream parse(PStream ps, ParserContext x) {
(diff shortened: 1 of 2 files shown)


src/foam/lib/parse/Until.java
```diff
@@ -0,0 +1,22 @@
...
+  implements Parser {
+
+  protected Parser until_;
+
+  public Until(Parser until) {
+    until_ = until;
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
+    Parser repeat = new Not(until_, AnyChar.instance());
```
kgrgreer: This could be done in the constructor to avoid creating every time.

PR #4289, opened 2020-11-13
[NP-2668] Fix CapabilityIntercept id property

KernelDeimos:


src/foam/nanos/crunch/CapabilityIntercept.js
```diff
@@ -34,7 +34,7 @@ foam.CLASS({
     {
       class: 'String',
       name: 'id',
-      value: 'foam.box.CapabilityRequiredRemoteException'
+      value: 'foam.box.CapabilityIntercept'
```
jlhughes:
```suggestion
      value: 'foam.nanos.crunch.CapabilityIntercept'
```

PR #4290, opened 2020-11-13
[NP-2598] Add docs for CRUNCH and FLOW, add TryItSnippet

KernelDeimos:
Does anybody know where `ace.js` went?


src/foam/flow/Document.js
```diff
@@ -379,6 +383,58 @@ foam.CLASS({
           };
         },
 
+        'foam-with-contents': function(v) {
```
kgrgreer, line 386: Okay for now, but have a look at how foam1 let's you specify property values with sub-tags.

PR #4291, opened 2020-11-13
Fixes and updates

noodlemoodle:
- update check for CapabilityRuntimeException to CapabilityRequiredRemoteException since it was removed
- use crunchService getjunction in place of manually building ucjdao find predicate


kgrgreer:
A great title. Reusable for all circumstances.

PR #4301, opened 2020-11-16
Fix repeat log overwriting the original log

lchanmann:
## Ref
- https://github.com/foam-framework/foam2/pull/4287

## Changes
- Clear repeat log id after copyFrom


jlhughes (approved):
I didn't know we had that call!

PR #4303, opened 2020-11-16
Language Picker Changes

blakegreen32:
Ticket: https://nanopay.atlassian.net/browse/NP-2712

Concerned about the country name change if this is being explicitly checked anywhere we would have to update that as well.

Screen Shot 2020-11-16 at 1 19 18 PM

### DIFF
--- a/src/foam/nanos/auth/LanguageChoiceView.js
+++ b/src/foam/nanos/auth/LanguageChoiceView.js
@@ -62,8 +62,8 @@ foam.CLASS({
   }
   ^ .popUpDropDown > div {
     margin-left: -16px;
+    margin-right: -16px;
     width: 100%;
-    text-align: center;
     height: 25;
     padding: 0px 16px 5px 16px;
     font-size: 14px;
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/LanguageChoiceView.js
```diff
@@ -62,8 +62,8 @@ foam.CLASS({
   }
   ^ .popUpDropDown > div {
     margin-left: -16px;
+    margin-right: -16px;
```
kennyqiyenkan, line 65: Is there a particular reason we are using negative margins here?

blakegreen32: Due to the paddings of the divs I used the negative margins to properly align the inner divs.

jlhughes:
Re: Country - the Country name will be translated (believe this is still in dev) so Brazil will be Brasil for Portuguese, for example. 
@blakegreen32 what was your concern about the country name?

blakegreen32:
> Re: Country - the Country name will be translated (believe this is still in dev) so Brazil will be Brasil for Portuguese, for example.
> @blakegreen32 what was your concern about the country name?

My concern is the change from USA to US. If we are checking somewhere for "USA" it will break

PR #4307, opened 2020-11-16
installLanguage before login

xuerongNanopay:
- trigger installLanguage method  before login check, which allow translation using in non-login required views.
- add PT translation for SignIn and SignUp view.


src/foam/nanos/u2/navigation/SignUp.js
```diff
@@ -37,7 +37,8 @@ foam.CLASS({
     { name: 'EMAIL_ERR', message: 'Valid email required' },
     { name: 'EMAIL_AVAILABILITY_ERR', message: 'This email is taken. Please try another.' },
     { name: 'USERNAME_EMPTY_ERR', message: 'Username required' },
-    { name: 'USERNAME_AVAILABILITY_ERR', message: 'This username is taken. Please try another.' }
+    { name: 'USERNAME_AVAILABILITY_ERR', message: 'This username is taken. Please try another.' },
+    { name: 'PASSWORD_ERR', message: 'Password should be at least 6 characters.' }
```
jlhughes, line 42: What was the original error? And why is that not translated already?  Just curious. I don't like the validation value '6' hard coded into the message.

dictator-x: the original error comes from (I think) password property itself, which is code in foam.core.StringPropertyValidationRefinement(this error message is constructed inside), which is not being translated.

xuerongNanopay: The original error comes from password property itself, which is in 'foam.core.StringPropertyValidationRefinement'(the error message is constructed inside), which I don't think it is being translated.

src/foam/nanos/controller/ApplicationController.js
```diff
@@ -319,9 +319,9 @@ foam.CLASS({
       this.clientPromise.then(async function(client) {
         self.setPrivate_('__subContext__', client.__subContext__);
 
-        await self.fetchSubject();
         await client.translationService.initLatch;
         self.installLanguage();
+        await self.fetchSubject();
```
kgrgreer, line 339: After we get the subject, if their locale is different than foam.locale, then we need to re-installLanguage() again.

src/foam/nanos/u2/navigation/SignUp.js
jlhughes: Let's leave the the PASSWORD_ERR out for now. we need a better way to deal with these.

PR #4310, opened 2020-11-17
Np 2553 fragmentedtextfield paste

noodlemoodle:
add paste to fragmentedtextfield https://nanopay.atlassian.net/browse/NP-2553

nanopay update https://github.com/nanoPayinc/NANOPAY/pull/11130


src/foam/u2/FragmentedTextField.js
```diff
@@ -95,19 +97,36 @@ foam.CLASS({
       this.numOfParts = this.delegates.length;
 
       return this;
+    },
+    function processData(data, index, arr) {
```
KernelDeimos, line 101: So this looks like it also works if I'm in the second box and decide to "paste" the rest of the number, is this correct?

src/foam/u2/FragmentedTextField.js
```diff
@@ -95,19 +97,36 @@ foam.CLASS({
...
-          this.currentIndex = this.currentIndex + 2;
-          this.childNodes[this.currentIndex].focus();
+        var arr = this.processData(this.childNodes[this.currentIndex].data, this.currentIndex, []);
+        for ( var i = 0; i < arr.length; i++ ) {
+          this.childNodes[this.currentIndex].data = arr[i];
+          if ( next = this.childNodes[this.currentIndex + 2] ) {
+            if ( this.delegates[this.currentIndex].maxLength > arr[i].length ) break;
+
+            if ( ! ( next && next.data ) ) {
+              this.currentIndex = this.currentIndex + 2;
```
KernelDeimos, line 124: I don't remember why this was `+2`. What happens when it's `++`?

KernelDeimos (approved):

src/foam/u2/FragmentedTextField.js
noodlemoodle: the textfields in the delegates are separated by stuff like '-', now I'm wondering if those might be better if added as a suffix in the fragment model.. any thoughts?

src/foam/u2/FragmentedTextField.js
noodlemoodle: yeah it made sense to me if the data was half filled, and you paste from some empty box it would work similarly and fill the data starting from that box and onwards

src/foam/u2/FragmentedTextField.js
KernelDeimos: Oh, or maybe a typecheck would be better. IMO it's a later problem since we're always using it for this situation, but there should be a `// TODO` or something at least

noodlemoodle, line 124:
```suggestion
              // TODO : typecheck the delegates to determine what currentIndex should be set to next instead of '+2'
              this.currentIndex = this.currentIndex + 2;
```

noodlemoodle: yeah makes sense, commented/ticketed 
https://github.com/foam-framework/foam2/pull/4310/files#diff-2849a5a16bce0b829d0e6707e081a15d147abc768ac7c78c7b9cf898355b1c9cR123
https://nanopay.atlassian.net/browse/NP-2744

PR #4311, opened 2020-11-17
create lisntener to check UCJdao consequently

yij793:
address: https://nanopay.atlassian.net/browse/NP-2693
note: Create a listener that will check crunch service continuously and stoped when the it get approved and update ucjStatus in UI. I do have a question, do we need to stop this continuous function when we switch pages? How can I do that?

### DIFF
--- a/src/foam/u2/crunch/CapabilityFeatureView.js
+++ b/src/foam/u2/crunch/CapabilityFeatureView.js
@@ -183,8 +183,6 @@ foam.CLASS({
           } else {
             this.window.setTimeout(this.statusUpdate, 2000);
           }
-        }).catch(err => {
-          throw err;
         })
       }
     }
--- a/src/foam/u2/crunch/CapabilityFeatureView.js
+++ b/src/foam/u2/crunch/CapabilityFeatureView.js
@@ -18,7 +18,8 @@ foam.CLASS({
     'crunchService',
     'ctrl',
     'subject',
-    'userCapabilityJunctionDAO'
+    'userCapabilityJunctionDAO',
+    'window'
   ],
 
   requires: [
@@ -175,12 +176,12 @@ foam.CLASS({
       code: function() {
         if ( this.cjStatus != this.CapabilityJunctionStatus.PENDING ) {
           return;
-        };
+        }
         this.crunchService.getJunction(null, this.data.id).then(ucj => {
           if ( ucj && ucj.status === this.CapabilityJunctionStatus.GRANTED ) {
             this.cjStatus = this.CapabilityJunctionStatus.GRANTED;
           } else {
-            window.setTimeout(this.statusUpdate, 2000);
+            this.window.setTimeout(this.statusUpdate, 2000);
           }
         })
       }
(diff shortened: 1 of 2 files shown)


src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -85,6 +85,7 @@ foam.CLASS({
        this.SUPER();
        this.onDetach(this.crunchService.sub('updateJunction', this.daoUpdate));
        this.daoUpdate();
+       this.onDetach(this.cjStatus$.sub(this.statusUpdate));
```
kgrgreer, line 89: The onDetach() will remove the listener when the view is removed.

kgrgreer:
The onDetach() will remove the listener when the view is removed.

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -168,6 +169,21 @@ foam.CLASS({
           }
         });
       }
+    },
+    {
+      name: 'statusUpdate',
+      code: function() {
+        if ( this.cjStatus != this.CapabilityJunctionStatus.PENDING ) {
+          return;
+        };
```
kgrgreer: No ; needed.

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -168,6 +169,21 @@ foam.CLASS({
...
+        this.crunchService.getJunction(null, this.data.id).then(ucj => {
+          if ( ucj && ucj.status === this.CapabilityJunctionStatus.GRANTED ) {
+            this.cjStatus = this.CapabilityJunctionStatus.GRANTED;
+          } else {
+            window.setTimeout(this.statusUpdate, 2000);
```
kgrgreer: Better to import: 'window', then use this.window. This will allow this code to still work if the app has multiple windows open (like the translation console does).

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -168,6 +169,21 @@ foam.CLASS({
...
+          }
+        })
```
zahnno, line 186: Please add a catch incase the call fails.

zahnno (approved):

src/foam/u2/crunch/CapabilityFeatureView.js
kgrgreer: This doesn't do anything. If it runs in a .then(), then it is async, so throwing an exception after this code has already returned, can never be caught.

yij793: @kgrgreer I think it works as .then().catch(), not inside

zahnno: The service method call can still fail and the catch method is defined on the service call.

kgrgreer: Yes, it can fail, but what good with throwing an exception do? It will happen in the future after this method has returned and nobody can catch that exception.

zahnno: Then console.error the message

PR #4322, opened 2020-11-18
Create NOC Notification on ERROR Alarms.  Also fix AlarmConfig to ini…

jlhughes:
1. Create NOC Notification on ERROR Alarms
2. Create NOC Notification on admin logins
3. Fix AlarmConfig setup. 
4. Use Notification.toastMessage as template created Notification.template value - The Notification.template value was hijacked to support templates. The NotificationTemplateDAO looks up the template via the 'template' value, but then clears it to create the real Notification.  The toastMessage is set as the 'template' value, as the 'template' value becomes the subject of generated emails.

### DIFF
--- a/src/foam/nanos/alarming/AddAlarmNameDAO.js
+++ b/src/foam/nanos/alarming/AddAlarmNameDAO.js
@@ -32,6 +32,8 @@ foam.CLASS({
       } else {
         config = new AlarmConfig();
         config.setName(alarm.getName());
+        config.setSeverity(alarm.getSeverity());
+        config.setClusterable(alarm.getClusterable());
         try {
           configDAO.put(config);
         } catch ( Exception e ) {
--- a/src/foam/nanos/alarming/AddAlarmNameDAO.js
+++ b/src/foam/nanos/alarming/AddAlarmNameDAO.js
@@ -33,7 +33,6 @@ foam.CLASS({
         config = new AlarmConfig();
         config.setName(alarm.getName());
         config.setSeverity(alarm.getSeverity());
-        config.setClusterable(alarm.getClusterable());
         try {
           configDAO.put(config);
         } catch ( Exception e ) {
--- a/src/foam/nanos/alarming/AlarmLoggingDAO.js
+++ b/src/foam/nanos/alarming/AlarmLoggingDAO.js
@@ -21,7 +21,7 @@ foam.CLASS({
     {
       name: 'put_',
       javaCode: `
-      Alarm old = (Alarm) getDelegate().find_(x, obj.getProperty("id"));
+      Alarm old = (Alarm) getDelegate().find_(x, ((Alarm)obj).getId());
       Alarm alarm = (Alarm) getDelegate().put_(x, obj);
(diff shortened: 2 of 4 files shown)


src/foam/nanos/alarming/AddAlarmNameDAO.js
```diff
@@ -32,6 +32,8 @@ foam.CLASS({
       } else {
         config = new AlarmConfig();
         config.setName(alarm.getName());
+        config.setSeverity(alarm.getSeverity());
+        config.setClusterable(alarm.getClusterable());
```
jlhughes:
```suggestion
```

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -140,10 +142,38 @@ foam.CLASS({
 
         Session session = x.get(Session.class);
         session.setUserId(user.getId());
+
+        if ( Group.ADMIN_GROUP.equalsIgnoreCase(user.getGroup()) ||
+             check(x.put("user", user), "*") ) {
```
lchanmann: shouldn't it be "subject" instead of "user" in the context?

src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -140,10 +142,38 @@ foam.CLASS({
...
+        }
+
         ((DAO) getLocalSessionDAO()).inX(x).put(session);
         session.setContext(session.applyTo(session.getContext()));
-
         return user;
+      } catch ( AuthenticationException e ) {
+        if ( user != null &&
+             ( Group.ADMIN_GROUP.equalsIgnoreCase(user.getGroup()) ||
+               check(x.put("user", user), "*") ) ) {
```
lchanmann: same as above

src/foam/nanos/alarming/AlarmLoggingDAO.js
```diff
@@ -12,14 +12,22 @@ foam.CLASS({
...
+    'foam.dao.DAO',
     'foam.nanos.logger.Logger',
+    'foam.nanos.notification.Notification'
   ],
 
   methods: [
     {
       name: 'put_',
       javaCode: `
+      Alarm old = (Alarm) getDelegate().find_(x, obj.getProperty("id"));
```
kgrgreer: Is this code generic, or you always know the type of the object? If you know, you could just cast it and then do .getID(), which would be slightly faster.

kgrgreer: I mean, you could just cast 'obj' to Alarm.

src/foam/nanos/auth/UserAndGroupAuthService.js
jlhughes: @lchanmann is the userX, further up, correct?

lchanmann: yes

lchanmann (approved):

PR #4324, opened 2020-11-18
add head option for select view

yij793:
address: https://nanopay.atlassian.net/browse/NP-2634
nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/11154
add header for select view

Screen Shot 2020-11-18 at 9 58 27 AM

### DIFF
--- a/src/foam/u2/tag/Select.js
+++ b/src/foam/u2/tag/Select.js
@@ -38,7 +38,13 @@ foam.CLASS({
         return undefined;
       }
     },
-    'size'
+    'size',
+    {
+      name: 'optGroup',
+      factory: function() {
+        return undefined;
+      }
+    }
   ],
 
   methods: [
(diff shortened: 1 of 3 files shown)


kennyqiyenkan (changes requested):
Are they all strings to be used as dropdown titles? If so, I think you should name it better. I did not get that when I first read that. Either add documentation or change the property name or both. Thanks

kennyqiyenkan (approved):

jlhughes:
How does this differ from RichChoiceView sections?

kennyqiyenkan:
> How does this differ from RichChoiceView sections?

I believe this is just adding support for default html select.

kgrgreer:
I think people have been using the RichChoiceView, not because they needed RichChoices, but just to get this header.
But my question is how is this different than the placeholder?

yij793:
> How does this differ from RichChoiceView sections?

RichChoiceView only take Dao

src/foam/u2/tag/Select.js
```diff
@@ -38,7 +38,14 @@ foam.CLASS({
...
       }
     },
-    'size'
+    'size',
+    {
+      name: 'dropDownHeader',
+      factory: function() {
+        return undefined;
+      },
+      documentation: 'drop down header is use to indicate the category of choices. it located above placeholder'
```
kgrgreer: But unlike RichChoiceView, there is only one "category", which means it isn't really a category. So this is just a "header". Could we fix the documentation and rename it to something like just 'header'?

src/foam/u2/tag/Select.js
```diff
@@ -38,7 +38,14 @@ foam.CLASS({
         return undefined;
       }
     },
-    'size'
+    'size',
+    {
+      name: 'header',
+      factory: function() {
```
kgrgreer: I think you can remove this factory.

PR #4326, opened 2020-11-18
[NP-2387] hide arrows if there are 4 or less capability cards

jinnnn9:
address: https://nanopay.atlassian.net/browse/NP-2387

### DIFF
--- a/src/foam/u2/crunch/CapabilityStore.js
+++ b/src/foam/u2/crunch/CapabilityStore.js
@@ -118,6 +118,10 @@ foam.CLASS({
       margin-top: 24px;
     }
   `,
+  
+  constants: {
+    MAX_NUM_DISPLAYBLE_CARDS: 4
+  },
 
   properties: [
     {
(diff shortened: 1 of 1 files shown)


src/foam/u2/crunch/CapabilityStore.js
```diff
@@ -255,12 +259,16 @@ foam.CLASS({
...
                 .end());
             }
-            spot.start('span').start('img').addClass(self.myClass('left-arrow'))
-                .attr('src', 'images/arrow-back-24px.svg')
-                .on('click', function() {
-                  self.carouselCounter--;
-                })
-              .end().end();
+            self.callIf(self.totalNumCards > self.MAX_NUM_DISPLAYBLE_CARDS, function() {
+              return spot.start('span')
```
kgrgreer, line 263: Is the span needed?

jinnnn9: @kgrgreer yes. We need the arrow to be an inline element so that it is correctly positioned.

PR #4329, opened 2020-11-18
Augment CapabilityAvailabilityDAO predicate

lchanmann:
## Ref
- https://github.com/foam-framework/foam2/pull/3322

## Changes
- Augment CapabilityAvailabilityDAO predicate to replace proxy sink
- Fix test ids
- Optimize CapabilityStore predicate to not send "IN" mlang when the list is empty

### DIFF
--- a/src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
+++ b/src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
@@ -10,7 +10,13 @@ foam.CLASS({
   extends: 'foam.dao.ProxyDAO',
   flags: ['java'],
 
+  imports: [
+    'AuthService auth'
+  ],
+
   javaImports: [
+    'foam.mlang.predicate.AbstractPredicate',
+    'foam.mlang.predicate.Predicate',
     'foam.nanos.auth.AuthService',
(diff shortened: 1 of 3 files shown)


src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
```diff
@@ -62,19 +67,38 @@ foam.CLASS({
...
-          public void put(Object o, Detachable d) {
-            Capability capability = (Capability) o;
-            if ( capability.getAvailabilityPredicate().f(x) || auth.check(x, AVAILABILITY_PERMISSION + capability.getId()) ) {
-              getDelegate().put(capability, d);
-            }
+        return getDelegate().select_(x, sink, skip, limit, order, augmentPredicate(x, predicate));
+      `
+    },
+    {
+      name: 'augmentPredicate',
```
kgrgreer, line 74: Is this method needed? Couldn't you just AND() this predicate and the supplied 'predicate' together?

lchanmann: yes, it's needed since unlike the regular predicate we need both x and the capability obj.

PR #4332, opened 2020-11-18
[NP-2781] update the text for "pending review"

RachaelDing:
https://nanopay.atlassian.net/browse/NP-2781

### DIFF
--- a/src/foam/nanos/crunch/CapabilityJunctionStatus.js
+++ b/src/foam/nanos/crunch/CapabilityJunctionStatus.js
@@ -48,7 +48,7 @@ foam.ENUM({
     {
       name: 'PENDING_REVIEW',
       label: { en: 'pending review', pt: 'revisão pendente' },
-      documentation: 'The information you provided is pending signing officer review',
+      documentation: 'The information you provided is pending final review',
       background: '#bfae32'
     },
     {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/crunch/CapabilityJunctionStatus.js
```diff
@@ -48,7 +48,7 @@ foam.ENUM({
     {
       name: 'PENDING_REVIEW',
       label: { en: 'pending review', pt: 'revisão pendente' },
-      documentation: 'The information you provided is pending signing officer review',
+      documentation: 'The information you provided is pending final review',
```
jlhughes: @kennyqiyenkan why the use of 'final'?  It may suggest this is the last step, but it's granting may reveal another capability to data-fill.

kennyqiyenkan: The wording came from Simon. If it is wrong, then I would suggest discussing it with them for better wording or go with what we had before.

jlhughes: @RachaelDing @kennyqiyenkan Discussed with Simon and Francois and we've decided on:
"The information you provided is pending review"

kgrgreer: @RachaelDing Please remove the word "final". Thx

RachaelDing: updated. Thanks

PR #4334, opened 2020-11-19
Emailtemplate

smirnovaka:


src/foam/lib/parse/Seq3.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class Seq3
+  implements Parser
+{
+  private Parser[] parsers;
+  private int      index1;
+  private int      index2;
+  private int      index3;
+
+  public Seq3(int i, int j, int k,  Parser... args) {
```
kgrgreer: Why not make this generic and call it something like SeqI. Then take int[] indices. Then create an Object[] result of the same length. Then anyone can just specify any number of elements they want to pick out of the result. Or, just don't bother. Once I need to create an array for the result anyway, then it doesn't make too much difference if it is a few characters longer or not. It still needs to be allocated and garbage collected at that point, which is the majority of the cost.

src/foam/nanos/notification/email/EmailTemplateUtil.js
```diff
@@ -0,0 +1,396 @@
...
+          Literal.create("%}"));
+
+        grammar.addSymbol("IF", ifParser);
+        Action ifAction  = new Action() {
+          @Override
+          public Object execute(Object val, foam.lib.parse.ParserContext x) {
+            Object[] valArr = (Object[]) val;
+            StringBuilder ifCond = new StringBuilder();
+            Object[] val0 = (Object[]) valArr[0];
+            for ( int i= 0; i < val0.length; i++ ) {
```
kgrgreer: space before = and ;'s

src/foam/lib/parse/Grammar.js
```diff
@@ -47,6 +47,22 @@ StringPStream ps = new StringPStream();
...
+    },
+    {
+      name: 'parseSb',
+      type: 'foam.lib.parse.PStream',
+      args: [
+        { name: 'str', type: 'StringBuilder' },
+        { name: 'optName', type: 'String' }
+      ],
+      javaCode: `
+if ( optName == null || "".equals(optName) ) optName = "START";
```
kgrgreer: SafetyUtil.isEmpty(optName)

src/foam/lib/parse/SeqI.java
```diff
@@ -0,0 +1,33 @@
...
+    indices = indexes;
+  }
+
+  public PStream parse(PStream ps, ParserContext x) {
+    Object[] value = new Object[indices.length];
+
+    for ( int i = 0 ; i < parsers.length ; i++ ) {
+      ps = ps.apply(parsers[i], x);
+      if ( ps == null ) return null;
+      for ( int j = 0; j < indices.length; j++ ) {
```
kgrgreer: spacing, here and elsewhere

src/foam/nanos/notification/email/EmailTemplateUtil.js
```diff
@@ -0,0 +1,396 @@
...
+    'foam.dao.DAO',
+    'foam.lib.json.*',
+    'foam.lib.parse.*',
+    'foam.lib.parse.Action',
+    'foam.lib.parse.Grammar',
+    'java.lang.StringBuilder',
+    'java.util.HashMap',
+    'java.util.Map',
+    'static foam.mlang.MLang.EQ'
+
```
kgrgreer: remove blank line

src/foam/nanos/notification/email/EmailTemplateUtil.js
```diff
@@ -0,0 +1,396 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.notification.email',
+  name: 'EmailTemplateUtil',
```
kgrgreer: "Util" is vague. How about a more descriptive name like "Engine" (I don't know why, but template systems are often called engines) or Service or Compiler. Is there anything email specific about this?

src/foam/nanos/notification/email/EmailTemplateUtil.js
```diff
@@ -0,0 +1,396 @@
...
+      javaType: 'java.lang.StringBuilder',
+      name: 'blockContentSb',
+      javaFactory: `return new StringBuilder();`
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.lib.parse.Grammar',
+      name: 'grammar',
+      documentation: `grammar for parsing "{{ val }}", "{% if/else}}" symbols`,
+      javaFactory: `
```
kgrgreer: The problem with this approach is that it is fairly expensive to build a parser, and then the parser doesn't get fast until it is used, but because you never reuse it, it never gets fast. Could this be reused? There are a couple approaches: 1. move it to its own class and make it a singleton (many other parsers do this) 2. move this whole class to be a service and then only have one instance. Then you would have to remove the stringbuilder as a property.

src/foam/nanos/notification/email/EmailTemplateUtil.js
```diff
@@ -0,0 +1,396 @@
...
+              for ( int i = 0 ; i < val1.length ; i++ ) {
+                finalVal.append(val1[i]);
+              }
+            } else {
+              Object[] val2 = (Object[]) valArr[2];
+              for ( int i = 0 ; i < val2.length ; i++ ) {
+                finalVal.append(val2[i]);
+              }
+            }
+            StringPStream finalValPs = new StringPStream();
```
kgrgreer: What are you doing here?

src/foam/nanos/notification/email/EmailTemplateUtil.js
```diff
@@ -0,0 +1,396 @@
...
+      return getSb();
+      `
+    },
+    {
+      name: 'isMultiTemplated',
+      args: [
+        { name: 'body', type: 'String' }
+      ],
+      type: 'Boolean',
+      documentation: `checks if template extends another templates, if yes, joins two and returns true`,
```
kgrgreer: Can extended templates extend other templates?

src/foam/nanos/notification/email/EmailTemplateUtil.js
smirnovaka: I'm re-parsing the inner part in case there is other symbols inside if-else statements

src/foam/nanos/notification/email/EmailTemplateEngine.js
```diff
@@ -0,0 +1,425 @@
...
+        Grammar grammar = new Grammar();
+        grammar.addSymbol("START", grammar.sym("markup"));
+
+        // markup symbol defines the pattern for the whole string
+        grammar.addSymbol("markup", new Repeat0(new Alt(grammar.sym("IF_ELSE"),
+          grammar.sym("IF"), grammar.sym("SIMPLE_VAL"), grammar.sym("ANY_CHAR"))));
+        Action markup = new Action() {
+          @Override
+          public Object execute(Object value, ParserContext x) {
+            return ((StringBuilder) x.get("sb")).toString();
```
kgrgreer: Do we need to to toString() it? Couldn't we just return the StringBuilder directly? Isn't it just as good as the String itself because they both implement the CharSequence interface.

src/foam/nanos/notification/email/EmailTemplateEngine.js
```diff
@@ -0,0 +1,425 @@
...
+          Literal.create("</"),
+          Whitespace.instance(),
+          Literal.create("content"),
+          Whitespace.instance(),
+          Literal.create(">")
+        ));
+        Action superBlockAction = new Action() {
+          @Override
+          public Object execute(Object val, ParserContext x) {
+            ((StringBuilder) x.get("sb")).append((String) x.get("content"));
```
kgrgreer: Do you need to cast this to a String?

PR #4335, opened 2020-11-19
adjust the themes, if find a theme in group, no need keep looking for…

yij793:
address: https://nanopay.atlassian.net/browse/NP-2737
nanopay side:https://github.com/nanoPayinc/NANOPAY/pull/11174
notes: if we find theme though group, we no longer need find its parent group. In back office case, it will always be nanopay

### DIFF
--- a/src/foam/nanos/theme/Themes.js
+++ b/src/foam/nanos/theme/Themes.js
@@ -177,15 +177,17 @@ Later themes:
       if ( user != null ) {
         DAO groupDAO = (DAO) x.get("groupDAO");
         Group group = user.findGroup(x);
-        while ( group != null ) {
-          Theme groupTheme = group.findTheme(x);
+        Theme groupTheme = null;
+        while ( groupTheme == null &&  group != null) {
+          groupTheme = group.findTheme(x);
           if ( groupTheme != null ) {
             theme = (Theme) theme.fclone().copyFrom(groupTheme);
             if ( ! SafetyUtil.isEmpty(group.getDefaultMenu()) ) {
               theme.setDefaultMenu(group.getDefaultMenu());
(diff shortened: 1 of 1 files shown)


zahnno (approved):

src/foam/nanos/theme/Themes.js
```diff
@@ -177,15 +177,17 @@ Later themes:
...
+        Theme groupTheme = null;
+        while ( groupTheme == null &&  group != null) {
+          groupTheme = group.findTheme(x);
           if ( groupTheme != null ) {
             theme = (Theme) theme.fclone().copyFrom(groupTheme);
             if ( ! SafetyUtil.isEmpty(group.getDefaultMenu()) ) {
               theme.setDefaultMenu(group.getDefaultMenu());
             }
+          } else {
+            group = (Group) groupDAO.find(group.getParent());
```
lchanmann: can we just break at after applying groupTheme instead?

yij793: like this ?
```java
          if ( groupTheme != null ) {
            theme = (Theme) theme.fclone().copyFrom(groupTheme);
            if ( ! SafetyUtil.isEmpty(group.getDefaultMenu()) ) {
              theme.setDefaultMenu(group.getDefaultMenu());
            }
          break;
}
```

jlhughes: yes, just break at the end of the if ( groupTheme != null ) { block }

jlhughes: So this should be a one line delta. 
All you need is 'break;' 
You can remove everything else you added.

jlhughes (approved):

PR #4338, opened 2020-11-19
Add new logoImages model

carl-zzz:
The new logoImages model contains 6 different images to replace the current largeLogo:
- smallLightOnDark
- smallDarkOnLight
- mediumLightOnDark
- mediumDarkOnLight
- largeLightOnDark
- largeDarkOnLight
However, I think we should standardize the 'logo' (maybe with constant length and width) as well as other sizes of logoImages @jlhughes @kgrgreer 
And I keep largeLogo and largeLogoEnabled for now, and will be removed later, and I also keep the 'logo' because sometime companies/we only need the logo without the slogan or brand word

Related Nanopay changes: https://github.com/nanoPayinc/NANOPAY/pull/11175

### DIFF
--- a/src/foam/nanos/auth/ChangePasswordView.js
+++ b/src/foam/nanos/auth/ChangePasswordView.js
@@ -119,7 +119,7 @@ foam.CLASS({
   methods: [
     function initE() {
       const self = this;
-      const logo = this.theme.largeLogo || this.theme.logo;
+      const logo = this.theme.logo;
 
       this.addClass(this.myClass())
         // header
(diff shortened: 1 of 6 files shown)


jlhughes:
re: logo - with and without slogan, branding.  This is the intent of small, medium, large.  Small presumably without slogan, branding unless they provided something that works in a small space. 
Perhaps we should also add the favicon here as well. 
If we go the route of 'logo', then rather than small, ,,,  light/dark,  it's 
* banner
* login
* ... 
* email
...

PR #4339, opened 2020-11-20
Add CIDR Whitelist support

jlhughes:
Add Classless Inter-Domain Routing restrictions. 
Or in layman's terms - restrict logins to ip ranges. 

Example: 
To restrict group login to VPN: 172.0.0.0/24

### DIFF
--- a/src/foam/nanos/auth/Group.js
+++ b/src/foam/nanos/auth/Group.js
@@ -122,7 +122,13 @@ foam.CLASS({
         class: 'foam.u2.view.FObjectPropertyView',
         readView: { class: 'foam.u2.detail.VerticalDetailView' }
       }
-    }
+    },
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/Group.js
```diff
@@ -122,7 +122,13 @@ foam.CLASS({
         class: 'foam.u2.view.FObjectPropertyView',
         readView: { class: 'foam.u2.detail.VerticalDetailView' }
       }
-    }
+    },
+    {
+      documentation: 'Restrict members of this group to particular IP address range',
+      class: 'StringArray',
+      name: 'cidrWhiteList',
```
kgrgreer, line 131: Add documentation: on the format the strings should be.

PR #4344, opened 2020-11-20
fixing scrolling hopefully

olhabn:
https://nanopay.atlassian.net/browse/NP-2769


olhabn (commented):
'min-height': '424px' considering current css configuration makes onScroll event not fired

jlhughes (commented):
This was changed six months ago by @tharmaman, do you remember why?

olhabn:
both dashboard table and 1M relationship tables look fine to me as of now

PR #4349, opened 2020-11-21
Add ability to disable generation of default java constructors.  Refa…

jlhughes:
…ctor AuthenticationException to a modeled Exception


src/foam/nanos/auth/AuthenticationException.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * @license
+ * Copyright 2018 The FOAM Authors. All Rights Reserved.
```
jlhughes:
```suggestion
 * Copyright 2020 The FOAM Authors. All Rights Reserved.
```

PR #4352, opened 2020-11-23
Updated section docs

zahnno:
Added additional section documentation.

### DIFF
--- a/src/foam/layout/Section.js
+++ b/src/foam/layout/Section.js
@@ -31,8 +31,8 @@ foam.CLASS({
       sections: [
         {
           name: 'mainSection'
-          title: function(eval, data) {
-            return data.myProp ? eval(data.myMessage) : 'Set this title if myProp false';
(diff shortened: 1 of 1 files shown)


src/foam/layout/Section.js
```diff
@@ -12,14 +12,27 @@ foam.CLASS({
...
       name: 'myModel',
+      messages: [
+        { name: 'myMessage', message: 'This message will evaluate a template literal \${myProp}' }
+      ],
       sections: [
         {
           name: 'mainSection'
-          title: function(myProp) {
-            return myProp ? 'Set this title if myProp true' : 'Set this title if myProp false';
+          title: function(eval, data) {
```
kgrgreer: Eval is a poor name choice, because people and audit-tools will think we're calling the JS-eval method. How about 'expand' or something else?

zahnno: Ahh - yes very true, will update to evaluateMessage

PR #4353, opened 2020-11-23
NP-2771 User Business Contact DAO Tabbed Sections

nanoscott:


src/foam/nanos/auth/Relationships.js
```diff
@@ -49,7 +49,7 @@ foam.RELATIONSHIP({
   forwardName: 'notificationSettings',
   inverseName: 'owner',
   sourceProperty: {
-    hidden: true
+    section: 'systemInformation'
```
blakegreen32, line 52: We don't want this hidden anymore?

blakegreen32 (approved):

src/foam/nanos/auth/Relationships.js
nanoscott: i was trying to show as many properties as possible .. and then we can hide those properties by hiding the entire section. I am showing these notification settings in case the system administrator wants to see them in the future.

blakegreen32:
@googlebot I consent.

jlhughes (approved):

PR #4356, opened 2020-11-23
Add JS subject.toString()

lchanmann:
## Ref
- https://github.com/nanoPayinc/NANOPAY/pull/11189

## Change
- Add JS subject.toString() - used by AbliiOnboardingSupport.js

### DIFF
--- a/src/foam/nanos/auth/Subject.js
+++ b/src/foam/nanos/auth/Subject.js
@@ -50,10 +50,7 @@ foam.CLASS({
       javaType: 'java.util.ArrayList<User>',
       name: 'userPath',
       documentation: 'path from realUser to current user',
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/Subject.js
```diff
@@ -50,7 +50,10 @@ foam.CLASS({
       javaType: 'java.util.ArrayList<User>',
       name: 'userPath',
       documentation: 'path from realUser to current user',
-      javaFactory: 'return new ArrayList();'
+      javaFactory: 'return new ArrayList();',
+      factory: function() {
```
kgrgreer: Shouldn't we add this to 'List' in types.js?

PR #4364, opened 2020-11-24
[NP-2820] Fix error in handling non-async availability of first section

KernelDeimos:
There was a bug in the specific case that the first section of the first wizardlet is unavailable, but there was no asynchronous call determining the availability of the section (no slot update). This was the cause of NP-2820. The wizard now properly determines section availability for this case.


nanopay-arthur (approved):
Tested these changes on both admission capability and signing officer capability, works as intended.

PR #4367, opened 2020-11-24
[DO_NOT_MERGE]create new propery visiabily predicate

yij793:
address: https://nanopay.atlassian.net/browse/NP-2802
nanoside: https://github.com/nanoPayinc/NANOPAY/pull/11184
note: create new PredicateProperty for Capability called visibilityPredicate. this is use to decide if we want this capability card visible or not base on a predicate


src/foam/nanos/crunch/Capability.js
```diff
@@ -197,6 +197,13 @@ foam.CLASS({
       javaFactory: 'return foam.mlang.MLang.TRUE;',
       documentation: 'Predicate used to omit or include capabilities from capabilityDAO'
     },
+    {
+      class: 'foam.mlang.predicate.PredicateProperty',
+      name: 'visibilityPredicate',
```
jlhughes, line 196: I'd like to name this something specific to capabilities so it's not confused with visibility in general. . 
```suggestion
      name: 'capabilityVisibilityPredicate',
```

jlhughes: Actually looking at this, why is it needed? 
@nanoMichal I'm sure we have predicate to test for True already.

yij793: @zahnno @Nauna

Nauna: So this came from availabilityPredicate.
So the requirement up to now was sufficient to use availabilityPredicate to filter the capabilityDAO for capability in the store.
However new requirement is to have the signingOfficer question capability visible and accessible in Ablii, but not visible but accessible in Treviso.
So now we have a difference in visibility and availability.

AvailabilityPredicate uses the CapabilityAvailabilityDAO and if predicate false returns no capability (null)
VisibilityPredicate will be 100% ui 

Although note this was a quick fix so we can move on with other work.

Phase 2 we have an idea to make a Cardlet - similar to Wizardlet, but for capability cards instead of wizard - and store UI components for Capability cards in the store.

Nauna: FYI - this was revision of @noodlemoodle 's original work. Ruby's work originally had this setup on availabilityPredicate and we were going to use permissions for dao availability. However setting up permissions was tedious at some points so this predicate (availabilityPredicate) was wanted to be as it was. Thus Ruby's work was reverted.
HOWEVER as described above we do need this logic ... thus new visibilityPredicate

nanoMichal:
You could use PropertyEqValue predicate

src/foam/nanos/crunch/Capability.js
```diff
@@ -197,6 +197,13 @@ foam.CLASS({
       javaFactory: 'return foam.mlang.MLang.TRUE;',
       documentation: 'Predicate used to omit or include capabilities from capabilityDAO'
     },
+    {
+      class: 'foam.mlang.predicate.PredicateProperty',
+      name: 'capabilityVisibilityPredicate',
+      networkTransient: true,
+      javaFactory: 'return foam.mlang.MLang.TRUE;',
+      documentation: 'Predicate of the visiability for capabilities'
```
zahnno:
```suggestion
      documentation: 'Predicate of the visibility for capabilities in the capability store/keyword sections'
```

src/foam/nanos/crunch/CapabilityIsVisible.js
```diff
@@ -0,0 +1,33 @@
...
+
+  methods: [
+    {
+      name: 'f',
+      code: () => { return true; },
+      javaCode: `
+        if ( ! ( obj instanceof Capability) ) return false;
+        X x = XLocator.get();
+        Capability capability = (Capability) obj;
+        return capability.getVisible() ? capability.getCapabilityVisibilityPredicate() != null
```
zahnno: This could just be a single condition I believe - replace `?` with `&&` and remove the `: false`

src/foam/nanos/crunch/CapabilityIsVisible.js
```diff
@@ -0,0 +1,33 @@
...
+
+  methods: [
+    {
+      name: 'f',
+      code: () => { return true; },
+      javaCode: `
+        if ( ! ( obj instanceof Capability) ) return false;
+        X x = XLocator.get();
+        Capability capability = (Capability) obj;
+        return capability.getVisible() ? capability.getVisibilityPredicate() != null
```
nanoMichal: Why not just replace the "Visible" property with the visibilityPredicate property. 
Then the predicate check is always simple. EQ(visible, true); or PropertyEQValue(visible,true)
Or if its because you want the ui checkbox just do: 
PropertyEQValue(visible,true) && PropertyEQValue(visibilityPredicate,true). 
Then you dont need this custom predicate at all

jlhughes: @yij793 can you respond to @nanoMichal comments, please.

yij793: @jlhughes I did with slack, this doesn't works on client side

jlhughes: I still find it really odd that you need this just to check a boolean.  Is it a bug or missing support making it not possible?

yij793: @jlhughes In client side, PropertyEQValue's prop does not accept predicate, same as EQ.

src/foam/nanos/crunch/CapabilityIsVisible.js
```diff
@@ -0,0 +1,33 @@
...
+  methods: [
+    {
+      name: 'f',
+      code: () => { return true; },
+      javaCode: `
+        if ( ! ( obj instanceof Capability) ) return false;
+        X x = XLocator.get();
+        Capability capability = (Capability) obj;
+        return capability.getVisible() && capability.getCapabilityVisibilityPredicate() != null 
+        && capability.getCapabilityVisibilityPredicate().f(x);
```
lchanmann:
```suggestion
          && capability.getCapabilityVisibilityPredicate().f(XLocator.get());
```
and remove line 26.

src/foam/nanos/crunch/CapabilityIsVisible.js
```diff
@@ -0,0 +1,33 @@
...
+    'foam.nanos.crunch.Capability'
+  ],
+
+  methods: [
+    {
+      name: 'f',
+      code: () => { return true; },
+      javaCode: `
+        if ( ! ( obj instanceof Capability) ) return false;
+        X x = XLocator.get();
```
lchanmann: Please remove this line.

src/foam/nanos/crunch/CapabilityIsVisible.js
```diff
@@ -0,0 +1,33 @@
...
+foam.CLASS({
+  package: 'foam.nanos.crunch',
+  name: 'CapabilityIsVisible',
+  extends: 'foam.mlang.predicate.AbstractPredicate',
+  implements: [ 'foam.core.Serializable' ],
+
+  javaImports: [
+    'foam.core.FObject',
+    'foam.core.XLocator',
+    'foam.core.X',
```
lchanmann: FObject and X are unused, please remove.

lchanmann (approved):

src/foam/nanos/crunch/CapabilityIsVisible.js
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.crunch',
+  name: 'CapabilityIsVisible',
+  extends: 'foam.mlang.predicate.AbstractPredicate',
+  implements: [ 'foam.core.Serializable' ],
+
```
kgrgreer: Could add:
axioms: [ { class: 'foam.pattern.Singleton' } ],
to make more efficient.

yij793: @kgrgreer added

src/foam/nanos/crunch/Capability.js
```diff
@@ -197,6 +197,13 @@ foam.CLASS({
       javaFactory: 'return foam.mlang.MLang.TRUE;',
       documentation: 'Predicate used to omit or include capabilities from capabilityDAO'
     },
+    {
+      class: 'foam.mlang.predicate.PredicateProperty',
+      name: 'capabilityVisibilityPredicate',
```
KernelDeimos:
```suggestion
      name: 'visibilityPredicate',
```
@yij793 I think `visibilityPredicate` is better because it's consistent with `availabilityPredicate` and more generic

src/foam/nanos/crunch/Capability.js
```diff
@@ -197,6 +197,13 @@ foam.CLASS({
...
+      networkTransient: true,
```
KernelDeimos:
```suggestion
```

Based on our conversation with @jlhughes in Slack we shouldn't have this networkTransient

src/foam/nanos/crunch/Capability.js
jlhughes: yes, my initial comment on this was from misunderstanding.  So it should revert to visibilityPredicate.  Sorry about that.

src/foam/nanos/crunch/Capability.js
jlhughes: yes, should not be networkTransient.

src/foam/dao/daoUtils.js
```diff
@@ -265,6 +265,56 @@ foam.CLASS({
 });
 
 
+foam.CLASS({
```
jlhughes, line 268: Add some documentation please.

jlhughes:
Please pull foam master into this branch.

PR #4369, opened 2020-11-25
[NP-2830] Add more CRUNCH documentation

KernelDeimos:
- Started some documentation related to CrunchController, and CRUNCH-specific wizardlet documentation.
- Also generalized `PropertyShortSummary` to `AxiomShortSummary`, making the Property variant a subclass and adding `MethodShortSummary`.

As an interesting note, I learned that getting the argument names from a method axiom is not as trivial as I expected:
- Method axioms defined by javascript function objects have an empty `args` property.
- FOAM provides `Function.argNames` which MethodShortSummary (from this PR) calls manually

I wonder if this should be done automatically when a `foam.core.Method` axiom is created from a javascript function object.


jlhughes (approved):
Approving the document parts. I'll let Kevin comment on the axiom bits

PR #4372, opened 2020-11-25
Revert "[NP-2830] Add more CRUNCH documentation"

jlhughes:
Reverts foam-framework/foam2#4369


jlhughes:
Reverted as we wanted to check on an existing axiom support.

PR #4379, opened 2020-11-27
approvalRequestDAO Sections

blakegreen32:


jlhughes (approved):

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -64,17 +64,19 @@
...
     },
     {
-      name: 'supportDetails'
+      name: 'complianceInformation',
+      title: 'Compliance',
+      order: 1
     },
     {
-      name: '_defaultSection',
+      name: 'systemInformation',
```
tharmaman, line 77: '_defaultSection' used by system as section for properties without a section specified btw. So from what I see every property has a section now?

blakegreen32: Ya they do, I could add it back in incase we add other properties that aren't section specified.

jlhughes: re: _defaultSection - going forward we should be more conscious of new properties.

PR #4385, opened 2020-11-27
Revamp JSON output delta

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-2889

## Issues
- JSON output for model having multipart id doesn't contain the properties (of the multipart id) on update

## Changes
- Always output id although it can be transient or multipart id to ensure that the entry/record is identifiable on replay.
- Output delta without generating delta properties list
- Update HistoryDAO to remove the delta list dependency

### DIFF
--- a/src/foam/dao/history/HistoryDAO.js
+++ b/src/foam/dao/history/HistoryDAO.js
@@ -14,6 +14,9 @@ foam.CLASS({
     'foam.core.PropertyInfo',
     'foam.core.X',
     'foam.dao.DAO',
+    'foam.lib.PropertyPredicate',
+    'foam.lib.StorageOptionalPropertyPredicate',
+    'foam.lib.StoragePropertyPredicate',
     'foam.lib.formatter.FObjectFormatter',
     'foam.lib.formatter.JSONFObjectFormatter',
     'foam.nanos.auth.Subject',
@@ -40,12 +43,15 @@ foam.CLASS({
       name: 'javaExtras',
       buildJavaClass: function(cls) {
         cls.extras.push(
-          ` 
+          `
+            private static final PropertyPredicate propertyPredicate_ = new StoragePropertyPredicate();
+            private static final PropertyPredicate optionalPredicate_ = new StorageOptionalPropertyPredicate();
+
             protected static final ThreadLocal<FObjectFormatter> formatter_ = new ThreadLocal<FObjectFormatter>() {
               @Override
               protected JSONFObjectFormatter initialValue() {
                 JSONFObjectFormatter formatter = new JSONFObjectFormatter();
-                formatter.setPropertyPredicate(new foam.lib.StoragePropertyPredicate());
+                formatter.setPropertyPredicate(propertyPredicate_);
                 return formatter;
               }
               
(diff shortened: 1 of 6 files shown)


jlhughes (commented):
@lchanmann if we meta mark properties as 'id' fields then perhaps an approach similar to storageTransient. And then we can add an IdPropertyPredicate which would be used by the outputters to control which properties are output.  
In AbstractFObejctFormatter.getDelta getProperties() propertyPredicateCheck an IdPropertyPredicate would return true, ensuring that that property was always output.

kgrgreer:
Do not merge. I want to look into this a bit first.

src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -338,15 +338,14 @@ public void outputDelta(FObject oldFObject, FObject newFObject, ClassInfo defaul
       append(',');
     }
     addInnerNewline();
-    PropertyInfo id = (PropertyInfo) newInfo.getAxiomByName("id");
-    boolean outputId = ! id.getStorageTransient();
-    if ( outputId ) outputProperty(newFObject, id);
+
+    // Always output id although it can be transient or multipart id to ensure
+    // that the entry/record is identifiable on replay.
+    outputId(newFObject, newInfo);
```
kgrgreer: I think that this method should be re-written to include the functionality of getDelta() in-line. getDelta creates an ArrayList, which will generate a lot of garbage. Also, it caches the value in delta_, which doesn't seems safe. Once we do that, we can modify the delta behaviour to just exclude id properties, which will actually make it faster still because we won't have to bother with the comparison.

lchanmann: On update, getDelta() doesn't include the id property and the multipart-id's properties because id wasn't changed but  the output json would still need the identifier field(s) for replaying. The `outputId()` method ensures the id/multipart-id is written out.

src/foam/dao/history/HistoryDAO.js
```diff
@@ -43,12 +43,15 @@ foam.CLASS({
       name: 'javaExtras',
       buildJavaClass: function(cls) {
         cls.extras.push(
-          ` 
+          `
+            private static final PropertyPredicate propertyPredicate_ = new StoragePropertyPredicate();
```
kgrgreer: By convention, static variables should be capitalized if final, or end with two underscores __, if updatable or threadlocal. So these two should be capital/constantized and formatter_ should be formatter__.

See: https://github.com/foam-framework/foam2/blob/8ef0e293ad790ee1c73ef35f658f4f0e1ad52203/src/foam/util/StringUtil.java#L9

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4385/commits/a90d41f3afb9235337b2cbd99f3008009e34a303.

jlhughes (commented):
LGTM

src/foam/lib/formatter/JSONFObjectFormatter.java
```diff
@@ -306,53 +306,64 @@ protected Boolean maybeOutputProperty(FObject fo, PropertyInfo prop, boolean inc
...
+    String    of        = newInfo.getObjClass().getSimpleName().toLowerCase();
+    List      axioms    = getProperties(newInfo);
+    int       size      = axioms.size();
+    int       delta     = 0;
+    int       ids       = 0;
+    int       optional  = 0;
+
+    String before = builder().toString();
+    reset();
+    for ( int i = 0; i < size ; i++ ) {
```
kgrgreer, line 324: space before ;

PR #4386, opened 2020-11-28
[DNM] [NP-2802] Make predicates ContextAware

KernelDeimos:
Adding DNM because I want to test a serializable sink solution.

### DIFF
--- a/src/foam/dao/PredicateContextualizingDAO.js
+++ b/src/foam/dao/PredicateContextualizingDAO.js
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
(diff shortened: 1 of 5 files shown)


src/foam/dao/PredicateContextualizingDAO.js
```diff
@@ -0,0 +1,37 @@
+foam.CLASS({
```
jlhughes, line 7: header please.

lchanmann:
Discussed with Eric offline to try overriding 'deepClone' instead.

KernelDeimos:
The `deepClone` approach gives me the system context, but to check if preconditions are met the session context is needed. There is also a security concern with predicates accessing the system context.

PR #4387, opened 2020-11-29
visibilityPredicate, preconditions, flow fixes

KernelDeimos:


src/foam/dao/daoUtils.js
```diff
@@ -265,6 +265,56 @@ foam.CLASS({
...
+      javaCode: `
+        if ( sink == null ) sink = new ArraySink();
+        ProxySink refinedSink = new ProxySink(x, sink) {
+          @Override
+          public void put(Object obj, Detachable sub) {
+            super.put(update(x, (FObject) obj), sub);
+          }
+        };
+        return ((ProxySink) super.select_(x, refinedSink, skip, limit, order, predicate)).getDelegate();
+      `
```
lchanmann: update on select_ and find_? should be used sparingly.

src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
```diff
@@ -96,6 +97,10 @@ foam.CLASS({
         { type: 'Capability', name: 'capability' }
       ],
       javaCode: `
+        // TODO: Rules calling capabilityDAO under the system context are
+        //   breaking availability logic. This needs proper investigation.
+        if ( ((Subject) x.get("subject")).getUser().isAdmin() ) return true;
```
lchanmann: if the subject.user is admin, the permission check on line 105 would succeed, no?

src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -205,6 +207,51 @@ public UserCapabilityJunction getJunction(X x, String capabilityId) {
...
+    if ( capabilityDAO.inX(x).find(capabilityId) == null ) return false;
+
+    var preconditions = Arrays.stream(((CapabilityCapabilityJunction[]) ((ArraySink) ((DAO) x.get("prerequisiteCapabilityJunctionDAO"))
+      .where(AND(
+        EQ(CapabilityCapabilityJunction.SOURCE_ID, capabilityId),
+        EQ(CapabilityCapabilityJunction.PRECONDITION, true)
+      ))
+      .select(new ArraySink())).getArray()
+      .toArray(new CapabilityCapabilityJunction[0])))
+      .map(CapabilityCapabilityJunction::getTargetId).toArray(String[]::new);
```
lchanmann: please use MAP sink instead.

src/foam/nanos/crunch/predicate/CapabilityGranted.js
```diff
@@ -22,6 +22,11 @@ foam.CLASS({
     {
       name: 'capabilityId',
       class: 'String'
+    },
+    {
+      name: 'subjectFromUCJ',
+      class: 'Boolean',
+      value: true
```
lchanmann: please add documentation.

src/foam/nanos/crunch/predicate/CapabilityPreconditionPredicate.js
```diff
@@ -0,0 +1,37 @@
...
+        var cap = (Capability) obj;
+        cap = (Capability) cap.fclone();
+
+        X x = getX();
+
+        var crunchService = (CrunchService) x.get("crunchService");
+        if ( ! crunchService.hasPreconditionsMet(x, cap.getId()) ) {
+          return false;
+        }
+        return true;
```
lchanmann:
```suggestion
        return crunchService.hasPreconditionsMet(x, cap.getId());
```

src/foam/nanos/crunch/predicate/CapabilityPreconditionPredicate.js
```diff
@@ -0,0 +1,37 @@
...
+  ],
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: `
+        var cap = (Capability) obj;
+        cap = (Capability) cap.fclone();
+
+        X x = getX();
```
lchanmann: is x the system or session context?

src/foam/nanos/crunch/predicate/IsAgent.js
```diff
@@ -0,0 +1,36 @@
...
+  ],
+
+  methods: [
+    {
+      name:'f',
+      javaCode: `
+        if ( ! ( obj instanceof X ) ) return false;
+        X x = (X) obj;
+        return ((Subject) x.get("subject")).getUser().getId()
+          != ((Subject) x.get("subject")).getRealUser().getId();
```
lchanmann: isAgent might make a good method on the subject itself.

src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
zahnno: IsAdmin - This property raises a red flag for me - why and what defines an admin on a nanos system? Our authorization system is primarily based on permissions and user owned objects defining those permissions. Why are we not using permissions to resolve this?

zahnno (commented):
CapabilityPreconditionDAO.js - empty file, please remove.

src/foam/dao/daoUtils.js
```diff
@@ -265,6 +265,56 @@ foam.CLASS({
...
+    }
+  ]
+});
```
zahnno: What's the purpose of this proxy class?

src/foam/nanos/crunch/SendNotificationOnTopLevelCapabilityStatusUpdate.js
```diff
@@ -40,7 +40,7 @@ foam.CLASS({
 
           // visible checks if the capability is a top-level capability
           // and availabilitypredicate checks if the user has access to the capability
-          if ( cap == null || ! ( cap.getVisible() && cap.getAvailabilityPredicate().f(x) ) ) return;
+          if ( cap == null || ! ( cap.getAvailabilityPredicate().f(x) ) ) return;
```
zahnno: If this condition controls the notification for 'top level capabilities being updated' then I think we have some flawed logic here.

This condition could be checking if the cap has any dependants (Which might not apply correctly in all scenarios)
OR
A property on capability defines 'top level'. I'm assuming that was the mixed intent with visible.

src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
KernelDeimos: Glad you brought this up in review, I wanted to bring attention to it on Slack but I forgot.

> why and what defines an admin on a nanos system?

`IsAdmin` does - but in the documentation for that method I explicitly wrote that we shouldn't use it for this purpose because of exactly what you said: that's what the permission system is for.

The call to `capabilityDAO` from a system context is the real problem. `CapabilityAvailabilityDAO` is nonsense for any operation you'd expect to do under a system context, since its behaviour is entirely reliant on a meaningful `subject` in the context.

We could also create a permission for capability availability and explicitly check for that; my worry is if we did that it would encourage calling `capabilityDAO` from a system context, which may cause us other problems in the future.

Currently I'm working on a `PreventSystemDAO` that we can decorate the served `capabilityDAO` with to warn about or prevent this behaviour.

src/foam/dao/daoUtils.js
KernelDeimos: Yes this is true. This was going to be used for an `isVisible` storageTransient property it if wasn't possible to do without it. As it happens it was, and this is no longer being used, however we do have DAO decorators on the system which do this which could extend this class for simplicity, so I left it in the PR.

src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
KernelDeimos: Does this apply to a system context? That's where I was finding an issue.

The problem started when I added the `IsAgent` predicate to SigningOfficer's `availabilityPredicate`, which showed me that the predicate was behaving incorrectly and getting a system context. I'll try removing this in a few minutes and see if any issues persist.

src/foam/nanos/crunch/predicate/CapabilityPreconditionPredicate.js
KernelDeimos: This file has been removed

src/foam/nanos/crunch/ServerCrunchService.java
KernelDeimos: I created a ticket for this: https://nanopay.atlassian.net/browse/NP-2907

I think it might make sense to optimize this method later so I can consider caching possibilities at the same time.

src/foam/dao/daoUtils.js
jlhughes: To modify the obj on find and select, implement update.

src/foam/nanos/crunch/predicate/IsAgent.js
KernelDeimos: I agree with this. tagging @kgrgreer @penzital in case they have comments.

src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
zahnno: Great - I believe the permission approach would be best suited for this (as it's currently implemented). isAdmin raises security concerns such as who can update it, when and what will it be used for and the caveat of implementing additional requirements to satisfy those concerns.

src/foam/nanos/crunch/SendNotificationOnTopLevelCapabilityStatusUpdate.js
KernelDeimos: Yeah good thing we do PR reviews. I forgot to update this properly. As long as this rule has a session context it should be easy to fix.

src/foam/dao/daoUtils.js
```diff
@@ -265,6 +265,56 @@ foam.CLASS({
...
+      name: 'select_',
+      javaCode: `
+        if ( sink == null ) sink = new ArraySink();
+        ProxySink refinedSink = new ProxySink(x, sink) {
+          @Override
+          public void put(Object obj, Detachable sub) {
+            super.put(update(x, (FObject) obj), sub);
+          }
+        };
+        return ((ProxySink) super.select_(x, refinedSink, skip, limit, order, predicate)).getDelegate();
```
kgrgreer: How is this different than AdapterDAO?

src/foam/nanos/crunch/ServerCrunchService.java
kgrgreer: I agree with @lchanmann.

src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -205,6 +207,51 @@ public UserCapabilityJunction getJunction(X x, String capabilityId) {
     return this.getJunctionForSubject(x, capabilityId, subject);
   }
 
+  public boolean hasPreconditionsMet(
+    X x, String capabilityId
```
kgrgreer: Maybe a good practice would be to name 'x' something like either 'sysX' or 'sessionX', which would make reading code easier.

src/foam/nanos/crunch/predicate/IsAgent.js
kgrgreer: Agreed.

src/foam/dao/daoUtils.js
KernelDeimos: It wasn't built for the same reason, but it looks like AdapterDAO can still act as an ObjectMutationProxyDAO. I'll remove this from the PR.

src/foam/nanos/crunch/CapabilityAvailabilityDAO.js
KernelDeimos: @zahnno I've confirmed that this check isn't needed anymore

src/foam/dao/daoUtils.js
KernelDeimos: This has been removed

src/foam/nanos/crunch/predicate/CapabilityGranted.js
KernelDeimos: Added

src/foam/nanos/crunch/predicate/CapabilityPreconditionPredicate.js
KernelDeimos: This file has been removed

src/foam/nanos/crunch/predicate/IsAgent.js
KernelDeimos: Updated

src/foam/nanos/crunch/ServerCrunchService.java
lchanmann: @KernelDeimos can you add a TODO comment if we are going to defer this?

zahnno (approved):

src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -205,6 +207,51 @@ public UserCapabilityJunction getJunction(X x, String capabilityId) {
...
+      // Return false if capability does not exist or is not available
+      if ( capabilityDAO.find(preconditionId) == null ) return false;
+      var ucj = getJunction(sessionX, preconditionId);
+      if ( ucj.getStatus() != CapabilityJunctionStatus.GRANTED ) return false;
+    }
+
+    return true;
+  }
+
+  public ArraySink getEntryCapabilities(X x) {
```
Nauna, line 239: please add documentation to the purpose of this function and hasPreconditionsMet

src/foam/nanos/crunch/crunchtest/CapabilityTest.java
```diff
@@ -397,7 +396,7 @@ public void testCapability(X x) {
     c1.setNotes("noted");
     c1.setVersion("0.0.0.0");
     c1.setEnabled(true);
-    c1.setVisible(false);
+    c1.setVisibilityPredicate(TRUE);
```
Nauna, line 399: why set this at all? - not big deal

src/foam/nanos/crunch/predicate/CapabilityGranted.js
```diff
@@ -32,17 +42,19 @@ foam.CLASS({
         if ( ! ( obj instanceof X ) ) return false;
         var x = (X) obj;
 
-        var newUCJ = (UserCapabilityJunction) x.get("NEW");
-        if ( newUCJ != null ) {
-          x = x.put("subject", newUCJ.getSubject(x));
+        if ( getSubjectFromUCJ() ) {
+          var newUCJ = (UserCapabilityJunction) x.get("NEW");
+          if ( newUCJ != null ) {
+            x = x.put("subject", newUCJ.getSubject(x));
```
Nauna, line 48: your getting a UCJ, and getting the subject from it??

Nauna (commented):
there is bunch of isAgent additions that are never used.
In what scenario do you see these being used?

Nauna (approved):

yij793:
@googlebot I consent

src/foam/nanos/crunch/predicate/CapabilityGranted.js
KernelDeimos: If a UCJ rule or predicate is executed within a system context it will need to know what subject the UCJ applies to. The logic works like this:

```
associatedEntity=0 -> subject is the source id alone
associatedEntity=1 -> subject is the source id alone
associatedEntity=2 -> subject is the realUser + effectiveUser
```

I noticed this logic was already present in a few places so I added `getSubject` to UserCapabilityJunction. It's basically asking "what's the most specific subject we can get with respect to this UCJ?"

The way the system works right now seems to be described pretty well by:
- _checking_ an associatedEntity=2 UCJ, you can get information the user and business's UCJs
- _affecting_ other UCJ's should happen the other way - a change to a business UCJ can affect any associatedEntity=2 UCJ, but it should never happen the other way around.

src/foam/nanos/crunch/crunchtest/CapabilityTest.java
KernelDeimos: `setVisible` doesn't exist anymore so I had to update it. Didn't think about removing visibility... this is a good reminder that we should structure test cases so it's really easy to add more cases - we should be able to change that test case so it tries it both ways

src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -205,6 +207,54 @@ public UserCapabilityJunction getJunction(X x, String capabilityId) {
...
+
+  // see documentation in CrunchService interface
+  public ArraySink getEntryCapabilities(X x) {
+    var sink = new ArraySink();
+    var proxySink = new ProxySink(x, sink) {
+      @Override
+      public void put(Object o, Detachable sub) {
+        var cap = (Capability) o;
+        if (
+          ! cap.getVisibilityPredicate().f(x)
```
kgrgreer: This expression all fits nicely on one line and then would be easier to read.

src/foam/nanos/crunch/predicate/IsAgent.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.crunch.predicate',
+  name: 'IsAgent',
```
kgrgreer, line 9: Is this used anywhere?

KernelDeimos: Nanopay side - used in a predicate for signing officer privileges so it doesn't show up when your real user and effective user are the same

PR #4388, opened 2020-11-29
Fixing column config prop view bug 11292020

olhabn:
- fixing setting 'of' property 
- fixing passing selectionChanged property

### DIFF
--- a/src/foam/u2/view/ColumnConfigView.js
+++ b/src/foam/u2/view/ColumnConfigView.js
@@ -710,7 +710,7 @@ foam.CLASS({
       return this.showOnSearch;
     },
     function returnSubColumnSelectConfig(subProperties, level, expanded) {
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/ColumnConfigView.js
```diff
@@ -747,6 +708,59 @@ foam.CLASS({
       if ( query.length === 0 )
         this.expanded = false;
       return this.showOnSearch;
+    },
+    function returnSubColumnSelectConfig(subProperties, level, expanded) {
+      if ( ! this.of || ! this.of.getAxiomByName || subProperties.length === 0 || !expanded )
```
kgrgreer: space after last !

olhabn: fixed

PR #4390, opened 2020-11-30
[DO_NOT_MERGE] Translation to Portuguese for user registration when the signup is in Portuguese

xuerongNanopay:
link ticket: https://nanopay.atlassian.net/browse/NP-2870
solution: persist current language to user model when doing sign up solve this ticket.

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -297,7 +297,8 @@ foam.CLASS({
       name: 'phoneNumberVerified',
       writePermissionRequired: true,
       section: 'operationsInformation',
-      gridColumns: 6
+      gridColumns: 6,
+      order: 40
     },
     {
       class: 'PhoneNumber',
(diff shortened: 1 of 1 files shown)


jlhughes (commented):
So this is setting/resetting the user's account to whatever the browser is set to?
Also, I can attempt to log in to anyone's account and update their language, if this does what it looks like. 
Rather, in the application controller, can we not just query the browser language setting if there is no local data of ours?

src/foam/nanos/u2/navigation/SignUp.js
```diff
@@ -183,6 +183,17 @@ foam.CLASS({
           });
         }
       }
+    },
+    {
+      name: 'createUserDefaultLanguage',
+      code: function() {
+        let l = foam.locale.split("-")
+        let code = l[0]
+        let varient = l[1] ? l[1] : undefined
```
kgrgreer: Could just be varient = l[1] || undefined;, or because you just use it for its truth-iness, you could just leave it as variant = l[1]

kgrgreer: You could also, just grab foam.language, except that it seems to be set incorrectly.

PR #4391, opened 2020-11-30
Report and alarm on capability not found

jlhughes:


jlhughes:
@Nauna capability not found - does this occur when the capability does not satisfy its availabilityPredicate?

Nauna (approved):

PR #4396, opened 2020-11-30
Fix null pointer exception in CapableView 

jinnnn9:
fix null pointer exception and add comments for future reference

### DIFF
--- a/src/foam/nanos/crunch/ui/CapableView.js
+++ b/src/foam/nanos/crunch/ui/CapableView.js
@@ -107,9 +107,7 @@ foam.CLASS({
     // add listeners to payload data
     function listenOnPayloads() {
       for ( const payload of this.capableObj.capablePayloads ) {
-        if ( payload.data === null ) continue;
-
-        payload.data.sub(this.clonePayloads);
+        if ( payload.data ) payload.data.sub(this.clonePayloads);
       }
     }
   ],
--- a/src/foam/nanos/crunch/ui/CapableView.js
+++ b/src/foam/nanos/crunch/ui/CapableView.js
@@ -43,7 +43,7 @@ foam.CLASS({
       name: 'wizardlets',
       documentation: 'wizardlets for capable payloads',
       postSet: function() {
-        this.addListeners();
+        this.listenOnPayloads();
       }
     },
     {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/crunch/ui/CapableView.js
```diff
@@ -101,9 +104,11 @@ foam.CLASS({
       .end();
     },
 
-    // add listeners to payloads
-    function addListeners() {
-      for ( payload of this.capableObj.capablePayloads ) {
+    // add listeners to payload data
+    function listenOnPayloads() {
+      for ( const payload of this.capableObj.capablePayloads ) {
+        if ( payload.data === null ) continue;
```
kgrgreer: Takes a little less mental effort for people to figure out code if you write it as either:
if ( payload.data) payload.data.sub(...).  OR:
payload.data && payload.ata.sub(...)

jinnnn9: I want to make a clear distinction between when data is set to undefined and when data is set to null because when data is null, it is intended but when data is undefined, it is an error. How about

```
if (data === undefined) raise/notify an error
if (data === null) continue

payload.data.sub(...)
```

kgrgreer: If it is either null or undefined, you're still going to get a NPE. Is that what you want?

jinn9: @kgrgreer yes In both cases its is NPE and I dont want to add a listener. But when data is undefined, there is some flaw in logic so I want to log it for debugging purpose.

jinn9: How about 

 ```
if ( payload.data ) {
          payload.data.sub(this.clonePayloads);
 } else if ( payload.data === undefined ) {
          console.warn(`${payload} cannot have data as undefined`);
 }
```

jinnnn9: @kgrgreer Now I think about it, it makes more sense that the function responsible for setting payload data should deal with undefined data.

jinnnn9: @kgrgreer Addressed the top comment

PR #4407, opened 2020-12-02
Np 1347/fixing GoogleSheetsTransactionsDataImportServiceImpl

olhabn:
- using googleSheetsTransactionsDataImport if the model of the menu is transaction
- supporting payer and payee being both user and business

### DIFF
--- a/src/foam/nanos/google/api/sheets/import/GoogleSheetsDataImportServiceImpl.js
+++ b/src/foam/nanos/google/api/sheets/import/GoogleSheetsDataImportServiceImpl.js
@@ -28,6 +28,7 @@
     'java.time.LocalDateTime',
     'java.time.format.DateTimeFormatter',
     'java.util.ArrayList',
+    'java.util.Collections',
     'java.util.Date',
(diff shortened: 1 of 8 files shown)


src/foam/nanos/google/api/sheets/import/GoogleSheetsDataImportServiceImpl.js
```diff
@@ -229,7 +229,8 @@
           Object val = data.get(i).get(columnIndex);
           pasreAndSetValue(x, obj, importConfig.getColumnHeaderPropertyMappings()[j], data.get(i).get(columnIndex));
         }
-        postSetValues(x, obj);
+        if ( ! postSetValues(x, obj) )
+          return new ArrayList<>();
```
kgrgreer: Will this ArrayList ever be updated? Would an immutable List be sufficient.

olhabn: fixed

PR #4410, opened 2020-12-02
[NP-2946] Add atLeastOneInCategory to CrunchService

KernelDeimos:


src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -207,6 +207,29 @@ public UserCapabilityJunction getJunction(X x, String capabilityId) {
     return this.getJunctionForSubject(x, capabilityId, subject);
   }
 
+  public boolean atLeastOneInCategory(X x, String category) {
```
jlhughes, line 210: Could this just be getCategoryCount().  and the find is a COUNT
and/or the atLeastOne calls the getCategoryCount.

jlhughes: Ok, will do for now. 
We'll investigate speed difference later.

PR #4412, opened 2020-12-02
Capability Store Improvements

zahnno:
Removes the need for refresh wizardlets to update cap store. Capability wizardlet save method publishes an 'updatejunction'  on the crunchService which is now subbed to on the cap store allowing for a dynamic view without the need to refresh.

However, the capability store wasn't correctly rendering its capability cards, making it difficult to make use of the listener to update capabilities shown in store. Went ahead and fixed that and added a listener on the publish of 'updateJunction'.

### DIFF
--- a/src/foam/u2/crunch/CapabilityStore.js
+++ b/src/foam/u2/crunch/CapabilityStore.js
@@ -189,7 +189,7 @@ foam.CLASS({
     },
     function initE() {
       this.SUPER();
-      this.crunchService.sub('updateJunction', this.onChange);
+      this.onDetach(this.crunchService.sub('updateJunction', this.onChange));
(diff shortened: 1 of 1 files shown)


src/foam/u2/crunch/CapabilityStore.js
```diff
@@ -189,7 +189,7 @@ foam.CLASS({
     },
     function initE() {
       this.SUPER();
-
+      this.crunchService.sub('updateJunction', this.onChange);
```
kgrgreer: If you do:
this.onDetach(this..crunchService.sub('updateJunction', this.onChange));
then it will remove the listener automatically when this view is destroyed.

PR #4413, opened 2020-12-02
Allow overriding relationship ids

lchanmann:
## Changes
- Allow overriding relationship ids
- Override user capability junction ids

### DIFF
--- a/src/foam/dao/Relationship.js
+++ b/src/foam/dao/Relationship.js
@@ -182,6 +182,14 @@ foam.CLASS({
         Only used for many to many relationships.
         Name of the class that this junction class should inherit from.
       `
+    },
+    {
+      class: 'StringArray',
+      name: 'ids',
+      documentation: "Ids of the relationship.",
+      factory: function() {
+        return ['sourceId', 'targetId'];
+      }
     }
     /* FUTURE:
     {
@@ -300,7 +308,7 @@ foam.CLASS({
         package: this.package,
         name: name,
         extends: this.extends,
-        ids: ['sourceId', 'targetId'],
+        ids: this.ids,
         properties: [
           {
             class: 'Reference',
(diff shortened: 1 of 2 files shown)


jlhughes (approved):

src/foam/dao/Relationship.js
```diff
@@ -182,6 +182,14 @@ foam.CLASS({
         Only used for many to many relationships.
         Name of the class that this junction class should inherit from.
       `
+    },
+    {
+      class: 'StringArray',
+      name: 'ids',
+      documentation: "Ids of the relationship.",
+      factory: function() {
```
kgrgreer, line 190: Doesn't just using value: work?

src/foam/dao/Relationship.js
```diff
@@ -300,7 +308,7 @@ foam.CLASS({
         package: this.package,
         name: name,
         extends: this.extends,
-        ids: ['sourceId', 'targetId'],
+        ids: this.ids,
```
kgrgreer, line 311: How is this different?

src/foam/nanos/crunch/Capability.js
```diff
@@ -449,6 +449,7 @@ foam.RELATIONSHIP({
   extends:'foam.nanos.crunch.Renewable',
   sourceModel: 'foam.nanos.auth.User',
   targetModel: 'foam.nanos.crunch.Capability',
+  ids: [ 'id' ],
```
kgrgreer, line 452: I think this is the only important line.

src/foam/dao/Relationship.js
lchanmann: value works.

src/foam/dao/Relationship.js
lchanmann: from hardcoded to property.

src/foam/nanos/crunch/Capability.js
lchanmann: yes, that's when relationship allows setting 'ids' to a custom one. The default ids is ['sourceId', 'targetId'].

src/foam/dao/Relationship.js
lchanmann: Well, value doesn't work. StringArray property has a default factory and factory is evaluated before value.

PR #4415, opened 2020-12-02
Support nested string literals

zahnno:
Messages with references to nested object are now supported when using the title & subtitle helper function 
ex(' message: '${capability.name} ${capability.description}' )

@kgrgreer `getNestedPropValue` is there anything like this already? If not then I believe it'd be useful to have it on FObject or where ever you feel it makes sense.

### DIFF
--- a/src/foam/u2/detail/SectionView.js
+++ b/src/foam/u2/detail/SectionView.js
@@ -67,6 +67,10 @@ foam.CLASS({
     {
       class: 'Function',
       name: 'getNestedPropValue',
+      documentation: `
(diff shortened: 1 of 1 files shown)


kgrgreer:
We have similar code, but not exactly the same. Please add documentation to explain the feature, otherwise people won't know it's there. Thx

zahnno:
alright - I'll add the documentation. I'll leave it up to you to add anywhere else you think necessary. It's currently accessed in the sectionView but may be useful elsewhere

PR #4416, opened 2020-12-03
add local for pt and support to emailVerificationService

Nauna:


src/foam/nanos/auth/email/EmailVerificationWebAgent.js
```diff
@@ -65,25 +63,36 @@ foam.CLASS({
         String             redirect         = request.getParameter("redirect");
         User               user             = (User) userDAO.find(Long.valueOf(userId));
 
+        TranslationService ts = (TranslationService) x.get("translationService");
+        String local = user.getLanguage().getCode().toString();
```
jlhughes, line 67: Why is this required?

jlhughes: Ah, ok, alternative is creating a subcontext with subject set that is used in call to translation service.  
this is fine.

jlhughes (approved):

PR #4419, opened 2020-12-03
IBAN validtion

minsunny:
https://nanopay.atlassian.net/browse/NP-2857

nanopay : https://github.com/nanoPayinc/NANOPAY/pull/11226

Added IBAN validation from com.google.foam.demos.iban.IBAN so that we can validate IBAN front end

### DIFF
--- a/src/foam/nanos/iban/ValidationIBAN.js
+++ b/src/foam/nanos/iban/ValidationIBAN.js
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
 foam.CLASS({
   package: 'foam.nanos.iban',
   name: 'ValidationIBAN',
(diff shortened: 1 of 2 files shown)


src/foam/nanos/nanos.js
```diff
@@ -448,5 +448,8 @@ FOAM_FILES([
   { name: "foam/nanos/geocode/GoogleMapsCredentials" },
 
   { name: "com/google/foam/demos/u2/AllViews" },
-  { name: "com/google/foam/demos/u2/MDAllViews" }
+  { name: "com/google/foam/demos/u2/MDAllViews" },
```
jlhughes, line 451: @kgrgreer can we move this foam/u2 ? 
@minsunny what in this are we using?

src/foam/nanos/iban/ValidationIBAN.js
```diff
@@ -0,0 +1,237 @@
+foam.CLASS({
```
jlhughes, line 7: header.

src/foam/nanos/nanos.js
minsunny: this was added with this PR https://github.com/foam-framework/foam2/pull/4114

jlhughes: My point here is that we are dependent on a 'demo' path for production code.  
If there is something in MDAllViews we are using it should move to foam/u2

kgrgreer: Scott added that, this isn't really part of this PR, she just added a comma. It is just part of the documentation, so I think it's fine.

PR #4420, opened 2020-12-04
[Np-2042] File preview

KolombetM:
https://nanopay.atlassian.net/browse/NP-2042

Allow select on file preview image


src/foam/nanos/fs/fileDropZone/FileCard.js
```diff
@@ -97,12 +98,20 @@ foam.CLASS({
...
+      name: 'selected'
     }
   ],
 
   methods: [
     function initE() {
       this.SUPER();
+      if ( this.selected == this.index ) {
+        this.style({
+          "border-color": "blue"
```
jlhughes: Is it possible to use: 
/* PRIMARY1 */ #hexcode  (or whatever is appropriate)
as we do elsewhere? so the colour is configurable and matches.

KolombetM: Updated. Is /* PRIMARY1 */ just a comment, or it is meaningful for foam parser?

RachaelDing: /\*%PRIMARY1%\*/ #hexcode works. If theme CSS variable PRIMARY1 is available, it goes with PRIMARY1. If it is not available, /\*%PRIMARY1%\*/ will act like a comment.

jlhughes (approved):

PR #4422, opened 2020-12-04
Intuit/BackOffice/Support UI Updates

blakegreen32:
Updated Intuit/BackOffice/Support UI and CSS

Ticket: https://nanopay.atlassian.net/browse/NP-2755


jlhughes:
@blakegreen32 why are all the colour codes hard coded and not from our Theme? 
If we have a new series of colours then we need to manage them accordingly.

nanoscott (approved):
LGTM .. please address Joel's concerns about the colours being in the theme first though.

src/foam/u2/borders/CardBorder.js
```diff
@@ -13,8 +13,10 @@ foam.CLASS({
     ^ {
       background-color: #ffffff;
       border: solid 1px #e7eaec;
-      box-shadow: 0 4px 5px 0 rgba(0,0,0,0.14), 0 1px 10px 0 rgba(0,0,0,0.12), 0 2px 4px -1px rgba(0,0,0,0.20);
-      padding: 6px;
+      border-radius: 6px;
```
jlhughes, line 16: @blakegreen32 Please create a ticket to capture all these colours that need to be supported for Cards - I'm assuming.  Or update them if we have what we need.

jlhughes (approved):

PR #4425, opened 2020-12-05
Add server side support for choosing to write files based on size to …

jlhughes:
Add server side support for choosing to write files based on size to either the real file system or store directly in the File model.

Also, when BlobService is used the resulting 'id' must be set on the corresponding File model, else httpFileService requests, which use the blob id, will never be resolved.

### DIFF
--- a/src/foam/nanos/fs/FileDataDAO.js
+++ b/src/foam/nanos/fs/FileDataDAO.js
@@ -1,0 +1,87 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.fs',
+  name: 'FileDataDAO',
+  extends: 'foam.dao.ProxyDAO',
+
+  javaImports: [
+    'foam.blob.*'
+  ],
+
+  properties: [
+    {
+      class: 'Long',
+      name: 'maxStringDataSize',
+      value: 1024 * 3
+    }
+  ],
+
+  methods: [
+    {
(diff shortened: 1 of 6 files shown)


src/foam/nanos/fs/FileDataDAO.js
```diff
@@ -0,0 +1,87 @@
...
+        try {
+          java.io.ByteArrayOutputStream os = new java.io.ByteArrayOutputStream((int) file.getFilesize());
+          blob.read(os, 0, file.getFilesize());
+          String encodedString = java.util.Base64.getEncoder().encodeToString(os.toByteArray());
+          String type = file.getMimeType();
+          if ( foam.util.SafetyUtil.isEmpty(type) ) {
+            if ( ! foam.util.SafetyUtil.isEmpty(file.getFilename()) &&
+                 file.getFilename().lastIndexOf(".") != -1 &&
+                 file.getFilename().lastIndexOf(".") != 0 ) {
+              type = "text/" + file.getFilename().substring(file.getFilename().lastIndexOf(".")+1);
```
lchanmann, line 50: Should we have a map of extensions and mine types? https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types. So that for .txt file the type is "text/plain" instead of "text/txt".

jlhughes: @lchanmann Yeah, we have the start of it in fs/SupportFileTypeDAO and fs/FileType . 
The above is just copy of what we currently have. 
So a future ticket. https://nanopay.atlassian.net/browse/NP-3001

lchanmann (approved):

PR #4430, opened 2020-12-07
persist language when user sign up

xuerongNanopay:
link ticket: https://nanopay.atlassian.net/browse/NP-2870

Description: persist language that takes from foam.locale to user model, when user signup.

Solve problem: If we do not persist language when signup, then user will have default language which is 'en'. As a result, FrontEnd may have foam.locale = 'pt-BR' and user.language = 'en', which is inconsistent, which will have some texts showing in Portuguese and some int English  UI. 

Note: You can not modify other user's language by attempting signup, because each email only allow sign up once.


src/foam/nanos/u2/navigation/SignUp.js
```diff
@@ -183,6 +183,17 @@ foam.CLASS({
           });
         }
       }
+    },
+    {
+      name: 'defaultUserLanguage',
+      code: function() {
+        let l = foam.locale.split("-")
+        let code = l[0]
+        let variant = l[1] || undefined
```
kgrgreer: The "|| undefined" is redundant and doesn't do anything. What is the intent?

src/foam/nanos/u2/navigation/SignUp.js
```diff
@@ -183,6 +183,17 @@ foam.CLASS({
           });
         }
       }
+    },
+    {
+      name: 'defaultUserLanguage',
+      code: function() {
+        let l = foam.locale.split("-")
```
kgrgreer: All of your lines are missing semi-colons.

PR #4432, opened 2020-12-07
Fixing service definition for AuthenticatedNSpecDAO

nanoscott:


src/services
```diff
@@ -20,19 +20,16 @@ p({
...
+      .setOf(foam.nanos.boot.NSpec.getOwnClassInfo())
+      .build();
   """,
   "client":"""
     {
-      "of":"foam.nanos.boot.NSpec",
-      "cache":true,
-      "ttlSelectPurgeTime": 0,
-      "ttlPurgeTime": 0
+      "of":"foam.nanos.boot.NSpec"
```
kgrgreer: Why removing the cache?

nanoscott: Using the CacheDAO instead of the TTLCacheDAO causes an error that I do not understand. If you have time to fix it, please checkout the Medusa branches and then try to load the daoKey in a new DIG entity. Otherwise, this fixes it (we could check this in as an interm fix).

jlhughes: @nanoscott So issue is classes referenced in nspec configuration not found js bin.  foam-bin.js 
For example foam.box.socket.SocketServer.  This is a model intended for server side only.

jlhughes: @nanoscott if you add the caching back in, we'll merge the other changes.

nanoscott: re-added.

PR #4443, opened 2020-12-09
fix issue when there were 1 chocie, the selection are hidden

yij793:
address: https://nanopay.atlassian.net/browse/NP-3049

note: if there are placeholder, the objectclass will not be this.choice[0][0]. so should same as visibility of objectclass

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -32,8 +32,8 @@ foam.CLASS({
       class: 'String',
       name: 'objectClass',
       label: '',
-      visibility: function(allowCustom, classIsFinal, choices, data) {
-        if ( ! allowCustom && choices.length <= 1  ) return foam.u2.DisplayMode.HIDDEN;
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/FObjectView.js
```diff
@@ -32,8 +32,8 @@ foam.CLASS({
       class: 'String',
       name: 'objectClass',
       label: '',
-      visibility: function(allowCustom, classIsFinal, choices, data) {
-        if ( ! allowCustom && choices.length <= 1  ) return foam.u2.DisplayMode.HIDDEN;
+      visibility: function(allowCustom, classIsFinal, choices, data, placeholder) {
+        if ( ! allowCustom && choices.length <= 1 && !placeholder) return foam.u2.DisplayMode.HIDDEN;
```
kgrgreer: space after !

jlhughes:
```suggestion
        if ( ! allowCustom && choices.length <= 1 && ! placeholder ) return foam.u2.DisplayMode.HIDDEN;
```

PR #4445, opened 2020-12-10
[NP-3009] ValiationIBAN Java support

KolombetM:
https://nanopay.atlassian.net/browse/NP-3009


src/foam/nanos/iban/ValidationIBAN.js
```diff
@@ -150,98 +150,275 @@ foam.CLASS({
...
+        }
+      ],
+      code:  function(cc) {
+        var cInfo = this.COUNTRIES[cc];
+        return cInfo && cInfo[0];
+      },
+      javaCode: `
+        Object[] cInfo = (Object[]) COUNTRIES.get(cc);
+        if ( cInfo != null ) return (String) cInfo[0];
+        return "null";
```
jlhughes: why String 'null' and not null? 
The javascript returns undefined.

src/foam/nanos/iban/ValidationIBAN.js
```diff
@@ -150,98 +150,275 @@ foam.CLASS({
...
+        iban = trim(iban);
+        if ( "".equals(iban) ) throw new IllegalStateException(IBAN_REQUIRED);
 
-      if ( checksum != 1 ) return `${this.INVALID_CHECKSUM} ${checksum}`;
+        String cc = iban.substring(0, 2);
+        String format = getFormatForCountry(cc);
 
-      return 'passed';
-    },
+        if ( format == null ) throw new IllegalStateException(UNKNOWN_COUNTRY_CODE + cc);
```
jlhughes: The call getFormatForCountry presently returns "null" rather than null.

jlhughes: Throw foam.core.ValidationException rather than IllegalStateException, here and elsewhere.

KolombetM: "null" - too much following js code, will fix.

nanopay/bank/BRBankAccount.js expects IllegalStateException from all validations, should I change this as well?

jlhughes: really, that was unexpected.  yes, that would be nice.

src/foam/nanos/iban/ValidationIBAN.js
```diff
@@ -150,98 +150,275 @@ foam.CLASS({
...
-        case "W": return c >= 'a' && c <= 'z' || c >= '0' && c <= '9';
-      }
-      return true;
+    {
+      name: 'validateChar',
+      type: 'boolean',
+      args: [
+        {
+          name: 'format',
+          type: 'Char'
```
kgrgreer: Better to make this a 'char'.

src/foam/nanos/iban/ValidationIBAN.js
```diff
@@ -150,98 +150,275 @@ foam.CLASS({
...
+          case "C": return c >= 'A' && c <= 'Z' || c >= 'a' && c <= 'z';
+          case "F": return c >= '0' && c <= '9';
+          case "L": return c >= 'a' && c <= 'z';
+          case "U": return c >= 'A' && c <= 'Z';
+          case "W": return c >= 'a' && c <= 'z' || c >= '0' && c <= '9';
+        }
+        return true;
+      },
+      javaCode: `
+        String f = Character.toString(format);
```
kgrgreer: Faster to make a char instead of Char then switch off of that.

src/foam/nanos/iban/ValidationIBAN.js
kgrgreer: Unless the 'Char' type makes the javaType 'char' already.

src/foam/nanos/iban/ValidationIBAN.js
```diff
@@ -150,98 +150,275 @@ foam.CLASS({
...
+      name: 'mod',
+      type: 'Integer',
+      args: [
+        {
+          name: 'num',
+          type: 'String'
+        },
+        {
+          name: 'a',
+          type: 'Integer'
```
kgrgreer: Does this become 'int'?

jlhughes (approved):

src/foam/nanos/iban/ValidationIBAN.js
```diff
@@ -150,98 +154,274 @@ foam.CLASS({
...
+          }
+        }
+        return iban;
+      },
+      javaCode: `
+        iban = iban.substring(4) + iban.substring(0, 4);
+        for ( int l = iban.length() - 1 ; l >= 0 ; l--) {
+          char c = iban.charAt(l);
+          if ( c >= 'A' && c <= 'Z' ) {
+           iban = iban.substring(0, l) + (10 + (Character.codePointAt(Character.toString(c), 0) - Character.codePointAt(Character.toString('A'), 0))) + iban.substring(l+1);
```
kgrgreer, line 358: Builds a new string for each character which will be slow. Would be better to create a char[] at the beginning.

src/foam/nanos/iban/ValidationIBAN.js
```diff
@@ -150,98 +154,274 @@ foam.CLASS({
...
+          iban = iban.substring(0, i) + '0' + iban.substring(i+1);
+        }
+      },
+      javaCode: `
+        for ( int i = iban.length() - 1 ; i >= 2 ; i-- ){
+          char c = iban.charAt(i);
+          if ( c < '9' ) {
+            return setChecksum(iban.substring(0, i) + (Integer.parseInt("" + iban.charAt(i)) + 1 + "") + iban.substring(i+1));
+          }
+          iban = iban.substring(0, i) + '0' + iban.substring(i+1);
```
kgrgreer, line 421: Could be made faster with an char[].

PR #4446, opened 2020-12-10
[NP-3050] Add CapabilityGrantMode

KernelDeimos:
Set `grantMode` of a capability to `MANUAL` if you do not wish for the capability to be granted automatically by a user action. This is useful when an event outside of CRUNCH is expected to grant the capability.


jlhughes (commented):
Nice.

PR #4452, opened 2020-12-10
Make email validation into message

jinnnn9:
nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/11386

### DIFF
--- a/src/foam/core/Validation.js
+++ b/src/foam/core/Validation.js
@@ -366,6 +366,11 @@ foam.CLASS({
   name: 'EmailPropertyValidationRefinement',
   refines: 'foam.core.EMail',
 
+  messages: [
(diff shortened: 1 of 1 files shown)


jlhughes (commented):
Does the Portuguese already exist?

src/foam/core/Validation.js
```diff
@@ -366,6 +366,11 @@ foam.CLASS({
   name: 'EmailPropertyValidationRefinement',
   refines: 'foam.core.EMail',
 
+  messages: [
+    { name: 'EMAIL_REQUIRED', message: 'Email required' },
+    { name: 'VALID_EMAIL_REQUIRED', message: 'Valid email required' }
```
kgrgreer: Would this be better if it were something like: 'Valid email address required'?

PR #4453, opened 2020-12-10
[NP-2931] - Remove BankAccountController

nanopay-arthur:
Refs:
https://nanopay.atlassian.net/browse/NP-2960
https://nanopay.atlassian.net/browse/NP-2931

#TimeSpent: 4d

Screen Shot 2020-12-10 at 6 52 15 PM
Screen Shot 2020-12-10 at 6 52 23 PM
Screen Shot 2020-12-10 at 6 52 32 PM


src/foam/comics/v2/DAOBrowseControllerView.js
```diff
@@ -16,6 +16,7 @@ foam.CLASS({
 
   imports: [
     'auth',
+    'currentMenu',
```
kgrgreer: What if someone is using without menus? Like a single Controller app?

src/foam/comics/v2/DAOControllerConfig.js
```diff
@@ -81,6 +81,23 @@ foam.CLASS({
...
+      name: 'primaryAction',
+      documentation: `
+        The most important action on the page. The view for this controller may
+        choose to display this action prominently.
+      `,
+      value: null
+    },
+    {
+      class: 'Boolean',
+      name: 'disableBrowseViewDataSet',
```
kgrgreer: What does this do?

src/foam/comics/v2/DAOBrowseControllerView.js
nanopay-arthur: Right, we only use the currentMenu to fetch the full path so that we're able to translate the title and subtitle. In the case where currentMenu is not set, I'll add a condition where it will default to config$title and config$subtitle.

src/foam/comics/v2/DAOControllerConfig.js
nanopay-arthur: Now that you bring this up, I think we don't need this.

src/foam/comics/v2/DAOBrowseControllerView.js
kgrgreer: Then, if the currentMenu import is optional, this line should change to 'currentMenu?', otherwise it will generate an error if it is missing. But, does it really even need to know about menus? Couldn't it just translate the title and subtitle that are provided in its config? The .handler.config.browseTitle pattern seems a little too internal implementation specific. Couldn't I instead just use something like "Edit Bank Accounts" as the translation key?

nanopay-arthur: Right, that does seem like the better solution in this case. Will update.

PR #4456, opened 2020-12-11
[NP-3063] Add FilterGrantModeAgent

KernelDeimos:
`CapabilityWizardlet`s for capabilities with `grantMode` set to `MANUAL` will not be displayed in CRUNCH wizards. Refer to #4446 for more information about `grantMode`.

This PR also adds `foam.Array.filter()` and `arrayProperty$filter` as a byproduct.

During this I learned that `FilterGrantModeAgent`'s `wizardlets` property cannot be an expression because property expressions do not accept import arguments.


Nauna (approved):
Merged locally with master and latest dev - works nicely

PR #4458, opened 2020-12-11
Close Wizard Bug

hchoura:
when you select close from the wizard, we will not have immediately another popup.
@Nauna @zahnno


jlhughes (approved):

zahnno:
The removed code is responsible for updating the capability store when statuses change. FYI @jlhughes

PR #4472, opened 2020-12-13
[NP-3054] Make UCJ status updates more restrictive

KernelDeimos:


jlhughes:
@KernelDeimos Regarding capabilities with grant mode manual, other than system context, user will need this permission correct? 
Also with regard to test cases where we are calling updateJunction - the status is simply a requested status? 
@nanoscott tagging you so you are aware, thinking of API calls and the API user needing the new perm.

KernelDeimos:
Yes that's correct, this permission is needed for other users to grant manual capabilities. For `updateJunction` the status can be considered requested status - the use case is that a MinMax will set a choice to `AVAILABLE` as a means of cancelling the UCJ if the user changes their mind

src/foam/nanos/crunch/ruler/AuthorizeUCJStatusOnPut.js
```diff
@@ -0,0 +1,67 @@
...
+    'foam.nanos.auth.AuthService',
+    'foam.nanos.crunch.Capability',
+    'foam.nanos.crunch.UserCapabilityJunction',
+    'java.util.HashMap',
+    'static foam.nanos.crunch.CapabilityJunctionStatus.*'
+  ],
+
+  documentation: `
+    This rule prevents a UCJ status from being set above ACTION_REQUIRED if the
+    user in context does not have
```
kgrgreer: Do you need to finish this sentence?

src/foam/nanos/crunch/ruler/AuthorizeUCJStatusOnPut.js
```diff
@@ -0,0 +1,67 @@
...
+            var ol = (UserCapabilityJunction) oldObj;
+            var nu = (UserCapabilityJunction) obj;
+            var auth = (AuthService) x.get("auth");
+
+            if ( ol == null ) {
+              if ( ! auth.check(x, PERMISSION) ) nu.setStatus(ACTION_REQUIRED);
+              return;
+            }
+
+            if (
```
kgrgreer: Why format this 'if' funny?

src/foam/nanos/crunch/ruler/AuthorizeUCJStatusOnPut.js
KernelDeimos: Weird... it happened again. I think I have an editor shortcut I'm not aware of that does this :/

src/foam/nanos/crunch/ruler/AuthorizeUCJStatusOnPut.js
KernelDeimos: It was just going past 80 characters so it was making me scroll sideways - should I move that to one line?

KernelDeimos:
Added DNM tag because I'm getting a new error on here when I merge with dev+master that doesn't happen on dev+master itself

KernelDeimos:
Fixed it!

jlhughes (approved):

PR #4476, opened 2020-12-14
fixed incorrectly enabled next button

minsunny:
https://nanopay.atlassian.net/browse/NP-3075

fixed incorrectly enabled next button

 in stepWizardletController.js 's canGoNext always returning true.
validation message checking parts are fine but found out allowSkipping's default has been set as true. 

Screen Shot 2020-12-14 at 11 16 42 AM

### DIFF
--- a/src/foam/u2/crunch/EasyCrunchWizard.js
+++ b/src/foam/u2/crunch/EasyCrunchWizard.js
@@ -16,8 +16,7 @@ foam.CLASS({
   properties: [
     {
       class: 'Boolean',
-      name: 'allowSkipping',
-      value: true
+      name: 'allowSkipping'
     },
     {
       class: 'Boolean',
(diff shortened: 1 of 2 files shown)


minsunny:
@KernelDeimos please review thanks.

src/foam/u2/crunch/EasyCrunchWizard.js
```diff
@@ -16,8 +16,7 @@ foam.CLASS({
   properties: [
     {
       class: 'Boolean',
-      name: 'allowSkipping',
-      value: true
```
KernelDeimos, line 20: The default wasn't false before? - or is this changing the default?

Either way `allowSkipping: false` is a much more reasonable default IMO so I'm approving this.

KernelDeimos (approved):

PR #4477, opened 2020-12-14
Cap store update fix

zahnno:
Addresses issues with capability store requiring browser refresh to reflect appropriate capability changes. Fixes previous flickering issues as well. 

Also added visibility logic on some capableObject props since they were incorrectly displayed in the gui

### DIFF
--- a/src/foam/u2/crunch/CapabilityFeatureView.js
+++ b/src/foam/u2/crunch/CapabilityFeatureView.js
@@ -181,7 +181,7 @@ foam.CLASS({
         this.crunchService.getJunction(null, this.data.id).then(ucj => {
           if ( ucj && ucj.status === this.CapabilityJunctionStatus.GRANTED ) {
             this.auth.cache = {};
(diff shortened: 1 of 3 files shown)


jlhughes:
Looks good to me.

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -180,7 +181,7 @@ foam.CLASS({
         this.crunchService.getJunction(null, this.data.id).then(ucj => {
           if ( ucj && ucj.status === this.CapabilityJunctionStatus.GRANTED ) {
             this.auth.cache = {};
-            this.crunchService.pub('updateJunction');
+            this.crunchService.pub('grantedJunction');
```
KernelDeimos, line 184: So this ensures that the store items are reloaded after capabilities are granted - is this correct?

KernelDeimos (approved):

PR #4484, opened 2020-12-15
Extract ruled model and DAO support

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-3097

## Changes
- Extract Ruled model and add RuledDAO cmd and support
- Add RuledDAOTest


src/foam/nanos/ruler/RuledDAO.js
```diff
@@ -0,0 +1,48 @@
...
+      name: 'find_',
+      javaCode: `
+        var ret = getDelegate().find_(x, id);
+        if ( ret instanceof Ruled ) {
+          return ((Ruled) ret).f(x) ? ret: null;
+        }
+        return ret;
+      `
+    },
+    {
```
kgrgreer, line 36: Let's comment this out, otherwise we we won't be able to view/configure in the GUI.

src/foam/nanos/ruler/RuledDAO.js
```diff
@@ -0,0 +1,48 @@
...
+    'foam.mlang.predicate.Predicate',
+    'foam.nanos.ruler.Ruled'
+  ],
+
+  methods: [
+    {
+      name: 'find_',
+      javaCode: `
+        var ret = getDelegate().find_(x, id);
+        if ( ret instanceof Ruled ) {
```
kgrgreer: I don't think this is the correct behaviour since it doesn't take into account ruleGroup or priority.
What we want is that if someone does EQ(RULE_GROUP, 'some value') then we find the the first matching item in priority order.

lchanmann: This is after the find by id, priority won't have any effect, no?

src/foam/nanos/ruler/RuledDAO.js
```diff
@@ -0,0 +1,42 @@
...
+  ],
+
+  methods: [
+    {
+      name: 'cmd_',
+      javaCode: `
+        if ( obj instanceof FindRuledCommand ) {
+          var sink = (ArraySink) getDelegate()
+            .where(AND(
+              INSTANCE_OF(Ruled.getOwnClassInfo()),
```
kgrgreer: I don't think this is needed. It will just slow things down and it should always be the case anyway.

src/foam/nanos/ruler/RuledDAO.js
```diff
@@ -0,0 +1,42 @@
...
+              EQ(Ruled.RULE_GROUP, ((FindRuledCommand) obj).getRuleGroup()) ))
+            .orderBy(DESC(Ruled.PRIORITY))
+            .select(new ArraySink());
```
kgrgreer: Better to process directly in the sink, then, you can stop once you find the first match.

src/foam/nanos/ruler/RuledDAO.js
lchanmann: This is to safety check in case RuledDAO was added to a non-Ruled model/DAO.

kgrgreer: Better to do that just once then when we setOf(), rather than on every object in every select.

src/foam/nanos/ruler/RuledDAO.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4484/commits/d81b5809a06ff4d4eb5a0b2527a3f81bf6314cce.

src/foam/nanos/ruler/RuledDAO.js
lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4484/commits/43e617028b5148a6aab83bdae951667221c5c31d.

PR #4486, opened 2020-12-15
Np 2066/lazy load sectioned detail view

olhabn:


src/foam/u2/detail/TabbedDetailView.js
```diff
@@ -73,13 +73,16 @@ foam.CLASS({
...
                     this
-                      .start(self.Tab, { label$: title$ || self.defaultSectionLabel })
+                      .start(self.Tab, { label$: title$ || self.defaultSectionLabel }, tab)
                         .call(function() {
                           this.tag(self.SectionView, {
                             data$: self.data$,
                             section: s,
-                            showTitle: false
+                            showTitle: false,
+                            skipLoading$: tab.value.selected$.map(function(s) {return !s;})
```
kgrgreer: Space after !.
Does this then erase the tab from the UI when it is unselected, forcing us to rebuild if the user goes back?
If so, it would be better to make skipLoading a latch and rename it loadLatch. This could be done by adding a preSet so that it can toggle from false to true but not the reverse (making it a latch).

olhabn: space fixed.
got a small problem though:
skipLoading is reference which is shared between all the tabs, so if will do 
`
preSet: function(o, n) {
        if ( o && ! n ) {
          return o;
        }
        return n;
      }
`
then tabs after first will not be updated

olhabn: I got your point though. 
I'll try to fix the code but I have no idea how to do that with the approach mentioned above

src/foam/u2/detail/SectionView.js
```diff
@@ -78,6 +78,17 @@ foam.CLASS({
           return this.getNestedPropValue(obj[props.shift()], props.join('.'))
         }
       }
+    },
+    {
+      class: 'Boolean',
+      name: 'loadLatch',
+      preSet: function(o, n) {
+        if ( ! o && n ) {
+          return o;
```
olhabn, line 87: o || n

src/foam/u2/detail/SectionView.js
```diff
@@ -78,6 +78,14 @@ foam.CLASS({
           return this.getNestedPropValue(obj[props.shift()], props.join('.'))
         }
       }
+    },
+    {
+      class: 'Boolean',
+      name: 'loadLatch',
+      preSet: function(o, n) {
+        return ! ( o || n );
```
olhabn: thanks for a suggestion @kgrgreer 
looks  really pretty now

kgrgreer: Is that correct? Should it be just 'o || n'? I see. You start off with value: true. Normally latches go from false to true, so can you reverse the logic?

src/foam/u2/detail/SectionView.js
```diff
@@ -111,15 +121,28 @@ foam.CLASS({
...
-                  .tag(self.SectionedDetailPropertyView, {
-                    prop: p,
-                    data$: self.data$
-                  })
-                .end();
-              })
+              .add(this.slot(function(loadLatch) {
+                var view = this.E();
+                
+                if ( ! loadLatch ) {
```
kgrgreer: reverse here also

src/foam/u2/detail/SectionView.js
olhabn: okay, so setting it to true was a bug. as the default value of loadLatch would make section which are used elsewhere not only in TabbedDetailView blank.

src/foam/u2/detail/SectionView.js
olhabn: okay, so setting it to true was a bug. as the default value of loadLatch would make section which are used elsewhere not only in TabbedDetailView blank.

so no need to reverse here

olhabn: okay, so setting it to true was a bug. as the default value of loadLatch would make section which are used elsewhere not only in TabbedDetailView blank.

src/foam/u2/detail/SectionView.js
```diff
@@ -78,6 +78,17 @@ foam.CLASS({
           return this.getNestedPropValue(obj[props.shift()], props.join('.'))
         }
       }
+    },
+    {
+      class: 'Boolean',
+      name: 'loadLatch',
+      preSet: function(o, n) {
+        return this.sectionContentViewSet ? o || n : n;
```
kgrgreer: Couldn't this just be 'o || n' and then no need for sectionContent since this value will only change once, ie. from false -> true?

src/foam/u2/detail/SectionView.js
kgrgreer: Shouldn't this be: ' if ( loadLatch ) ' ?

src/foam/u2/detail/SectionView.js
```diff
@@ -86,6 +97,8 @@ foam.CLASS({
       var self = this;
       self.SUPER();
 
+      var sectionContent;
```
kgrgreer: Why is this needed?

src/foam/u2/detail/SectionView.js
```diff
@@ -78,6 +78,17 @@ foam.CLASS({
           return this.getNestedPropValue(obj[props.shift()], props.join('.'))
         }
       }
+    },
+    {
+      class: 'Boolean',
+      name: 'loadLatch',
+      preSet: function(o, n) {
+        return this.sectionContentViewSet ? o || n : n;
+      }
```
olhabn, line 87: so default value - false and we load the view on false because if SectionView.js is used elsewhere other then in TabbedDetailVIew we should load at

value is false but then changed to true in TabbedDetailVIew if tab isn't selected and we need to ignore it

src/foam/u2/detail/SectionView.js
olhabn: false will be changed to true if tab isn't selected in TabbedDetailView and we need to return true

src/foam/u2/detail/SectionView.js
olhabn: if we change `if ( ! loadLatch )` to  `if ( loadLatch )` then on default value which is false we will not display section content and that will ruin for example log in view

src/foam/u2/detail/SectionView.js
olhabn: so false is default and on false we load section content as views elsewhere other then TabbedDetailVIew will work with `loadLatch` set to false and we need content of the section loaded

olhabn: false -> true will be changed in src/foam/u2/detail/TabbedDetailView.js line 85 for tabs which are not selected and we return true and on true we do not load section content

olhabn: and then when we select it back from true ->  false we aren't loading it again as we have true || false -> true. on true we don't load

olhabn: but we cannot load on true value as it's not default and other usages of SectionView will not load the content

src/foam/u2/detail/SectionView.js
olhabn: cleaned up

src/foam/u2/detail/TabbedDetailView.js
```diff
@@ -78,13 +78,16 @@ foam.CLASS({
...
                     this
-                      .start(self.Tab, { label$: title$ || self.defaultSectionLabel, selected: self.memento.paramsObj.selectedTab && self.memento.paramsObj.selectedTab === s.title })
+                      .start(self.Tab, { label$: title$ || self.defaultSectionLabel, selected: self.memento.paramsObj.selectedTab && self.memento.paramsObj.selectedTab === s.title }, tab)
                         .call(function() {
                           this.tag(self.SectionView, {
                             data$: self.data$,
                             section: s,
-                            showTitle: false
+                            showTitle: false,
+                            loadLatch$: tab.value.selected$.map(function(s) {return ! s;})
```
kgrgreer: The loadLatch should just be a latch version of selected. It turns true but then never turns back. So, just add 'selected' to SectionView and on the postSet of it, set loadLatch to true if the new value is true. 
Then change line 138 to just if ( loadLatch )

src/foam/u2/detail/SectionView.js
```diff
@@ -78,6 +78,24 @@ foam.CLASS({
           return this.getNestedPropValue(obj[props.shift()], props.join('.'))
         }
       }
+    },
+    {
+      class: 'Boolean',
+      name: 'loadLatch',
+      expression: function(selected) {
```
kgrgreer: You don't need this expression and preSet and the postSet on 'selected'. Either approach will work. The postSet is more efficient.

olhabn: I think I need expression for such use case:
if `selected` was not changed then this expression will set `loadLatch` to true as `selected` is true otherwise loadLatch will be false, which we don't want in cases on SectionView being used elsewhere other then TabbedDetailView

kgrgreer: Ok, then you should document that, since it isn't going to be obvious to the casual observer, who might be tempted to "fix" your code.

PR #4487, opened 2020-12-15
class exception caused by trying to set enum as string

Nauna:


Nauna:
With Master at the time that I pulled this wasn't there - but this change from name to label is negligable thus closing the pr

PR #4488, opened 2020-12-15
[NP-3106] Section view fix

jinnnn9:
address: https://nanopay.atlassian.net/browse/NP-3106

summary:
Don't create html element if section title/subtitle function returns an empty string -> prevents creating extra space between properties

e.g.,
Screen Shot 2020-12-15 at 12 19 10 PM

### DIFF
--- a/src/foam/u2/detail/SectionView.js
+++ b/src/foam/u2/detail/SectionView.js
@@ -99,7 +99,7 @@ foam.CLASS({
                   code: section.title
                 });
                 if ( slot$.value ) {
-                  this.start('h2').add(slot$).end();
+                  this.start('h2').add(slot$.value).end();
                 }
               } else {
                 this.start('h2').add(section.title).end();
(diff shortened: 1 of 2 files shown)


src/foam/u2/detail/SectionView.js
```diff
@@ -93,22 +93,30 @@ foam.CLASS({
...
-                code: section.subTitle
-              }) : section.subTitle$;
-              this.start().addClass('subtitle').add(slot$).end();
+              if ( foam.Function.isInstance(self.section.subTitle) ) {
+                const slot$ = ExpressionSlot.create({
+                  args: [ self.evaluateMessage$, self.data$ ],
+                  code: section.subTitle
+                });
+                if ( slot$.value ) {
+                  this.start().addClass('subtitle').add(slot$).end();
```
kgrgreer: Can you just add slot$.value instead of slot$. It would be more efficient.

src/foam/u2/detail/SectionView.js
```diff
@@ -93,22 +93,30 @@ foam.CLASS({
...
+              if ( foam.Function.isInstance(self.section.title) ) {
+                const slot$ = foam.core.ExpressionSlot.create({
                   args: [ self.evaluateMessage$, self.data$ ],
-                  obj$: self.data$,
                   code: section.title
-                }) : section.title$;
-              this.start('h2').add(slot$).end();
+                });
+                if ( slot$.value ) {
+                  this.start('h2').add(slot$).end();
```
kgrgreer: Can you just add slot$.value instead of slot$. It would be more efficient.

PR #4512, opened 2020-12-17
Pending review included in cap store polling listener 

zahnno:
Cards on capability store update based on UCJ changes. 
Required behaviour explained here: https://nanopay.atlassian.net/browse/NP-3131

NOTE extended issues: 

Flickering is persistent on most status changes - onChange is called multiple times based on a published event of 'updateJunction'. Possible loop caused by ucj listeners or expression in conjunction with the fetches occurring in the onChange listeners.

Menus aren't updated when menu read permissions are granted.

### DIFF
--- a/src/foam/u2/crunch/CapabilityStore.js
+++ b/src/foam/u2/crunch/CapabilityStore.js
@@ -39,6 +39,7 @@ foam.CLASS({
     'capabilityDAO',
     'crunchController',
     'crunchService',
+    'menuDAO',
     'registerElement'
   ],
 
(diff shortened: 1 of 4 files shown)


src/foam/u2/crunch/CapabilityStore.js
```diff
@@ -391,16 +391,16 @@ foam.CLASS({
...
-      name: 'onChange',
-      isMerged: true,
-      code: function() {
-        this.crunchService.getEntryCapabilities().then(a => {
-          this.visibleCapabilityDAO = this.ArrayDAO.create({
-            array: a.array
-          });
-        });
-      }
+    async function onChange() {
```
kgrgreer: Why is the isMerged removed? If it is flickering, then increase the mergeDelay.

PR #4513, opened 2020-12-17
fixing View Reference blank bug

olhabn:
ticket https://nanopay.atlassian.net/browse/NP-3123

### DIFF
--- a/src/foam/nanos/approval/ApprovalRequest.js
+++ b/src/foam/nanos/approval/ApprovalRequest.js
@@ -742,6 +742,8 @@
                     createPredicate: foam.mlang.predicate.False.create(),
                     deletePredicate: foam.mlang.predicate.False.create()
                   }),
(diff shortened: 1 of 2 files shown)


src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -742,6 +742,8 @@
                     createPredicate: foam.mlang.predicate.False.create(),
                     deletePredicate: foam.mlang.predicate.False.create()
                   }),
+                  mementoHead: null,
+                  backLabel: 'Back'
```
jlhughes, line 746: Can we support translation on this?

kgrgreer: It will be used in an ActionView, which will translate it, as long as there is a message in the LocaleDAO.

PR #4515, opened 2020-12-17
adding onDetach

olhabn:


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -347,19 +347,19 @@ foam.CLASS({
       // Start Memento Support
       this.WindowHash.create({value$: this.memento.value$});
```
olhabn, line 349: these changes are probably pretty useless

PR #4521, opened 2020-12-17
[NP-2896] fix cap store cards cut off issue

RachaelDing:
address: https://nanopay.atlassian.net/browse/NP-2896

- Remove hard coding of MAX_NUM_DISPLAYBLE_CARDS
- Add/remove the arrows based on if the cards overflow
- When the cards overflow, add arrows and enable scroll bar

Screen Shot 2020-12-17 at 5 07 32 PM
Screen Shot 2020-12-17 at 5 07 24 PM

### DIFF
--- a/src/foam/u2/crunch/CapabilityStore.js
+++ b/src/foam/u2/crunch/CapabilityStore.js
@@ -124,10 +124,6 @@ foam.CLASS({
     }
   `,
 
-  constants: {
-    MAX_NUM_DISPLAYBLE_CARDS: 4
-  },
-
   properties: [
     {
       name: 'visibleCapabilityDAO',
--- a/src/foam/u2/crunch/CapabilityStore.js
+++ b/src/foam/u2/crunch/CapabilityStore.js
@@ -174,8 +174,7 @@ foam.CLASS({
     },
     {
       name: 'cardsOverflow',
-      class: 'Boolean',
-      value: false
+      class: 'Boolean'
     },
     'wizardOpened'
   ],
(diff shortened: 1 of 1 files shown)


src/foam/u2/crunch/CapabilityStore.js
```diff
@@ -174,6 +172,11 @@ foam.CLASS({
       value: [],
       documentation: 'stores the styling of each featureCapability'
     },
+    {
+      name: 'cardsOverflow',
+      class: 'Boolean',
+      value: false
```
kgrgreer: not needed, already the default

src/foam/u2/crunch/CapabilityStore.js
```diff
@@ -407,6 +408,14 @@ foam.CLASS({
         this.menuDAO.cmd_(this, foam.dao.CachingDAO.PURGE);
         this.menuDAO.cmd_(this, foam.dao.AbstractDAO.RESET_CMD);
       }
+    },
+    {
+      name: 'onResize',
+      code: function() {
+        var scrollWidth = this.document.querySelector(`.foam-u2-crunch-CapabilityStore-feature-column-grid`).scrollWidth;
```
kgrgreer: If there is more than one this will break. You stored it in the 'ele' variable, so you could access that the do ele.el().scrollWidth;

RachaelDing: updated. Thanks

kgrgreer:
Good job.

PR #4527, opened 2020-12-18
fixing bug

olhabn:
ticket: https://nanopay.atlassian.net/browse/NP-1096

### DIFF
--- a/src/foam/u2/search/GroupBySearchView.js
+++ b/src/foam/u2/search/GroupBySearchView.js
@@ -149,7 +149,7 @@ foam.CLASS({
     },
 
     function updatePredicate_(choice) {
-      var exists = typeof choice !== 'undefined' && choice !== '';
+      var exists = typeof choice !== 'undefined' && choice !== '' && choice != null;
       this.predicate = exists ? this.op.create({
         arg1: this.property,
         arg2: this.Constant.create({ value: choice })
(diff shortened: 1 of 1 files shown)


src/foam/u2/search/GroupBySearchView.js
```diff
@@ -149,7 +149,7 @@ foam.CLASS({
     },
 
     function updatePredicate_(choice) {
-      var exists = typeof choice !== 'undefined' && choice !== '';
+      var exists = typeof choice !== 'undefined' && choice !== '' && choice != null;
       this.predicate = exists ? this.op.create({
```
olhabn, line 153: so there are at least two methods `findChoiceByData` and `findChoiceByText` in `foam2/src/foam/u2/view/ChoiceView.js` which set choice to null so added a null check

olhabn: as an option we probably could change those methods to return `undefined`

kgrgreer: I think that's a better idea, to modify those methods to return undefined.
commit f05febbd89e8defeaa67cd679f5b7ec830e8f5ff
Date:   Fri Dec 18 10:34:45 2020 -0500

    ChoiceView findChoiceBy... methods now return undefined if not found instead of null.

Also, we should use foam.Undefined.isInstance() instead of typeof.

PR #4532, opened 2020-12-18
changing of.model_. to of.getAxiomByName

olhabn:
ticket: https://nanopay.atlassian.net/browse/NP-2040

### DIFF
--- a/src/foam/u2/filter/FilterView.js
+++ b/src/foam/u2/filter/FilterView.js
@@ -206,10 +206,14 @@ foam.CLASS({
 
         if ( searchColumns ) return searchColumns;
 
-        if ( of.model_.searchColumns ) return of.model_.searchColumns;
+        var columns = of.getAxiomByName('searchColumns');
+        columns = columns && columns.columns;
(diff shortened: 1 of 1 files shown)


src/foam/u2/filter/FilterView.js
```diff
@@ -206,10 +206,14 @@ foam.CLASS({
 
         if ( searchColumns ) return searchColumns;
 
-        if ( of.model_.searchColumns ) return of.model_.searchColumns;
-
-        if ( of.model_.tableColumns ) {
-          return of.model_.tableColumns.filter(function(c) {
```
olhabn, line 212: changing of.model_. to of.getAxiomByName
as it doesn't work for refinements I think
or at least for UserCapabilityJunctionRefine

jlhughes (approved):
Well, just seems correct in general.

src/foam/u2/filter/FilterView.js
kgrgreer: Your changes are correct.

PR #4533, opened 2020-12-18
Locale dao authorize on update

hchoura:
Screen Shot 2020-12-18 at 3 04 37 PM

### DIFF
--- a/src/foam/i18n/Locale.js
+++ b/src/foam/i18n/Locale.js
@@ -8,11 +8,24 @@ foam.CLASS({
   package: 'foam.i18n',
   name: 'Locale',
 
+  javaImports: [
+    'foam.nanos.auth.AuthorizationException',
+    'foam.nanos.auth.Subject',
+    'foam.nanos.auth.User',
+  ],
+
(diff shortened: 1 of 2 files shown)


jlhughes (commented):
two issues:
If adding this authorizeOn then this model should implement Authorizable, and you should also implement authorizeOnCreate and Delete
Or as @kgrgreer mentioned, just add an authorizer in easyDAO
      .setAuthorizer(new foam.nanos.auth.GlobalReadAuthorizer("locale"))

hchoura:
> two issues:
> If adding this authorizeOn then this model should implement Authorizable, and you should also implement authorizeOnCreate and Delete
> Or as @kgrgreer mentioned, just add an authorizer in easyDAO
> .setAuthorizer(new foam.nanos.auth.GlobalReadAuthorizer("locale"))

yes, 
`.setAuthorizer(new foam.nanos.auth.GlobalReadAuthorizer("locale"))`
solve the issue.

https://github.com/foam-framework/foam2/blob/3a5cb639d8e01dfabe04c50f2266b54b4511854b/src/foam/i18n/services.jrl#L31

PR #4547, opened 2020-12-21
Adding ability to fold arbitrary values in the OM logger

nanoscott:


src/foam/nanos/om/OMLogger.js
```diff
@@ -16,6 +16,17 @@ foam.CLASS({
   ],
 
   methods: [
+    {
+      name: 'fold',
+      type: 'Void',
+      args: [ 
+        { name: 'value', type: 'Float' },
```
kgrgreer, line 23: Why is it a Float instead of some integer type?

PR #4550, opened 2020-12-21
Np 1096/group keys empty

olhabn:


jlhughes (approved):

src/foam/mlang/mlang.js
```diff
@@ -2578,7 +2578,9 @@ foam.CLASS({
       hidden: true,
       name: 'groupKeys',
       javaFactory: 'return new java.util.ArrayList();',
-      factory: function() { return []; }
+      factory: function() {
```
kgrgreer, line 2582: Should also make the javaFactory do the same and then we can make this field transient to save bandwidth.

PR #4553, opened 2020-12-22
Np 3115/granted ucj data update

noodlemoodle:
general ucj rules cleanup and adjust rules so that GRANTED ucjs data can be updated

https://github.com/nanoPayinc/NANOPAY/pull/11522


jlhughes (approved):

src/foam/nanos/crunch/ruler/SaveUCJDataOnGranted.js
```diff
@@ -43,9 +43,8 @@ foam.CLASS({
               )
             ));
 
-            if ( ucj.getStatus() != CapabilityJunctionStatus.GRANTED || ucj.getIsRenewable()
-              || ( old != null && ! old.getIsRenewable() && old.getStatus() == CapabilityJunctionStatus.GRANTED ) ) 
-              return;
+            if ( ucj.getStatus() != CapabilityJunctionStatus.GRANTED || ucj.getIsRenewable() ) return;
+            if ( old != null && ! old.getIsRenewable() && old.getData().equals(ucj.getData()) ) return;
```
Nauna, line 47: Just thinking this out-
If ucj is not granted or if ucj is renewable return ... if old is not renewable and no data change return.
I don't understand what renewable has to do with saving the data here.

noodlemoodle: so the first case is for if ucj is renewable but not granted means that the uccj would've moved to state ingraceperiod or inrenewalperiod on the renewable interface, which doesn't change its granted status, in this case we don't want to save the data because there shouldn've been an update - this change in state is done by system
the second case is if the ucj is GRANTED not renewable but there was no data update whichalso means we dont need to save the data

PR #4556, opened 2020-12-22
update address postal code error message

yij793:
address: https://nanopay.atlassian.net/browse/NP-3147
nanopay side: https://github.com/nanoPayinc/NANOPAY/pull/11525
Screen Shot 2020-12-22 at 12 25 08 PM
Screen Shot 2020-12-22 at 12 25 13 PM
Screen Shot 2020-12-22 at 12 24 41 PM
Screen Shot 2020-12-22 at 12 36 39 PM


src/foam/nanos/auth/Address.js
```diff
@@ -219,6 +220,16 @@ foam.CLASS({
       },
       gridColumns: 6,
       validationPredicates: [
+        {
+          args: ['postalCode'],
+          predicateFactory: function(e) {
+            return e.GT(
+              foam.mlang.StringLength.create({
+                arg1: foam.nanos.auth.Address.POSTAL_CODE
+              }), 0);
```
zahnno, line 229: why the new validation predicate? You could include this GT expression within the validationPredicate below. I'm sure including 'Valid' will still get the message across to the user. The errorMessage in the vp below also includes both cases where zip code and postal code are mentioned.

Nauna: Ya was thinking this but confirmed on the ticket - it was what was desired. 'they' want different msg for this.

Nauna (approved):

PR #4564, opened 2021-01-04
[NP-3143] - Add actionContext Call to RichChoiceView

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-3143


src/foam/u2/view/RichChoiceView.js
```diff
@@ -524,8 +532,16 @@ foam.CLASS({
...
+                  .add(this.slot(function(action, actionContext) {
+                    if ( action && actionContext) {
+                      return this.E()
+                        .startContext({data: actionContext})
+                        .start(self.DefaultActionView, { action: action })
+                          .addClass(self.myClass('action'))
+                        .end()
+                        .endContext();
+                    }
+                    if ( action && ! actionContext) {
```
Nauna:
```suggestion
                    if ( action ) {
```
not that the code is incorrect but redundant check considering your returning above if 'actionContext' is truthy.

Nauna (approved):

PR #4574, opened 2021-01-05
save the payloads when it get generated

yij793:
address: https://nanopay.atlassian.net/browse/NP-3194
nanoSide: https://github.com/nanoPayinc/NANOPAY/pull/11559
Notes: Quick fix for unable add bank account. @KernelDeimos having an idea to correct the logic

### DIFF
--- a/src/foam/nanos/crunch/ui/CapableView.js
+++ b/src/foam/nanos/crunch/ui/CapableView.js
@@ -42,7 +42,10 @@ foam.CLASS({
       of: 'foam.u2.wizard.BaseWizardlet',
       name: 'wizardlets',
       documentation: 'wizardlets for capable payloads',
-      postSet: function() {
+      postSet: function(_, n) {
+        var promises = [];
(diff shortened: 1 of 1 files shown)


KernelDeimos (approved):
This is needed because CapableView listens for updates on payloads, therefore the payloads need to exist after initialization.

src/foam/nanos/crunch/ui/CapableView.js
```diff
@@ -42,7 +42,10 @@ foam.CLASS({
       of: 'foam.u2.wizard.BaseWizardlet',
       name: 'wizardlets',
       documentation: 'wizardlets for capable payloads',
-      postSet: function() {
+      postSet: function(_, n) {
+        var promises = [];
+        n.forEach(wizard => promises.push(wizard.save()))
+        Promise.all(promises).then(cps => console.log(cps))
```
Nauna, line 48: should we really be console.logging this? If you want to log this, think the logger should be used.

PR #4580, opened 2021-01-06
add condition to pass the check of postal code for countries that don…

yij793:
address: https://nanopay.atlassian.net/browse/NP-3151
Screen Shot 2021-01-06 at 1 02 54 PM

### DIFF
--- a/src/foam/nanos/auth/Address.js
+++ b/src/foam/nanos/auth/Address.js
@@ -223,10 +223,18 @@ foam.CLASS({
         {
           args: ['postalCode'],
           predicateFactory: function(e) {
-            return e.GT(
-              foam.mlang.StringLength.create({
-                arg1: foam.nanos.auth.Address.POSTAL_CODE
-              }), 0);
+            return e.OR(
+              e.EQ(foam.nanos.auth.Address.COUNTRY_ID, 'ZA'),
+              e.EQ(foam.nanos.auth.Address.COUNTRY_ID, 'TT'),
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/Address.js
```diff
@@ -223,10 +223,18 @@ foam.CLASS({
...
+              e.EQ(foam.nanos.auth.Address.COUNTRY_ID, 'TT'),
+              e.EQ(foam.nanos.auth.Address.COUNTRY_ID, 'MY'),
+              e.EQ(foam.nanos.auth.Address.COUNTRY_ID, 'MX'),
+              e.EQ(foam.nanos.auth.Address.COUNTRY_ID, 'JM'),
+              e.EQ(foam.nanos.auth.Address.COUNTRY_ID, 'LB'),
+              e.GT(
+                foam.mlang.StringLength.create({
+                  arg1: foam.nanos.auth.Address.POSTAL_CODE
+                }), 0)
+            )
```
minsunny, line 237: question! We already check the postal code validations on the each configured country. Then We need this here?

zahnno:
If we are missing regions from countries we support, we should add them. Along with zip code validations.

src/foam/nanos/auth/Address.js
yij793: yeah, they want 2 error message, one is if it is empty, one is invalid

jlhughes:
@zahnno I agree.  @yij793 can you create a ticket to capturing all the missing regions and code validations.  Then work on it slowly (lower priority, as time permits), country by country.

PR #4587, opened 2021-01-07
fix enum wrong default value

noodlemoodle:
assuming we always want to return ordinal 0 for default value of enum properties unless otherwise specified, then
this should return forordinal(0) instead of values[0] since the values are not necessarily installed in the correct order

fixes issue encountered where values[0] for paymentstatus is ordinal 6 when i built locally without setting default value explicitly but is ordinal 0 on ci's build, similar issue also happens for invoicestatus

### DIFF
--- a/src/foam/core/Enum.js
+++ b/src/foam/core/Enum.js
@@ -384,7 +384,7 @@ foam.CLASS({
         return n
       },
       expression: function(type) {
-        return type && foam.lookup(type).VALUES[0];
+        return type && ( foam.lookup(type).forOrdinal(0) || foam.lookup(type).VALUES[0] );
       }
     },
     {


src/foam/core/Enum.js
```diff
@@ -384,7 +384,7 @@ foam.CLASS({
         return n
       },
       expression: function(type) {
-        return type && foam.lookup(type).VALUES[0];
+        return type && ( foam.lookup(type).forOrdinal(0) || foam.lookup(type).VALUES[0] );
```
kgrgreer, line 387: Would this also work if you just used 'of' instead of 'foam.lookup(type)'?

noodlemoodle: no, not really sure why but installAxioms is changing the order in the sort even though they all have the same priority..

kgrgreer: I mean replace foam.lookup(type) to be replaced with just this.of?

kgrgreer:
If you change the code in Enum.js line 227 to:

          if ( def.ordinal ) {
            next = Math.max(def.ordinal + 1, next);
          } else {
            def.ordinal = next++;
          }

Does that fix the problem?

kgrgreer:
Fixed by
commit d28f4fad7783eda74d631a95e0ce1229fc2b0c64
Date:   Tue Jan 26 13:33:58 2021 -0500

    Fix Enum ordinals not getting output correctly and set priority based on ordinal.

PR #4588, opened 2021-01-08
using mql for memento predicate serialization

olhabn:
image

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -858,7 +858,7 @@ foam.CLASS({
       var mqlStringsArr = [];
       for ( var a in this.args ) {
         if ( ! this.args[a].toMQL )
-          throw new Error( 'Argument does not support toMQL' );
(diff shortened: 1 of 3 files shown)


src/foam/mlang/mlang.js
```diff
@@ -764,6 +764,20 @@ getArg2().prepareStatement(stmt);`
         getArg1().authorize(x);
         getArg2().authorize(x);
       `
+    },
+    function arg2ToMQL() {
+      if ( this.arg2 && this.arg2.value && foam.Date.isInstance(this.arg2.value) ) {
```
kgrgreer, line 769: This is when arg2 is a Constant? Couldn't this logic be moved to Constant and we just call toMQL() on arg2?

olhabn:
looks like 'AND' for strings with '<', '>' isn't parsed correctly 
(eg 'created>2020-12-28T00:00:00.000Z AND created<2021-01-23T00:00:00.000Z').
so closing pr until I figure out that

PR #4589, opened 2021-01-08
fix first historyRecord miss in medusa

xuerongNanopay:
link ticket: https://nanopay.atlassian.net/browse/NP-2883
when  obj.getProperty(“id”) is null or 0 then it’s a create operation.
perform the delegate put, then pass the original obj and result from delegate to historyDAO

### DIFF
--- a/src/foam/dao/history/HistoryDAO.js
+++ b/src/foam/dao/history/HistoryDAO.js
@@ -24,7 +24,8 @@ foam.CLASS({
     'foam.nanos.logger.Logger',
     'java.util.ArrayList',
     'java.util.Date',
-    'static foam.mlang.MLang.EQ'
+    'static foam.mlang.MLang.EQ',
+    'foam.util.SafetyUtil'
   ],
 
   messages: [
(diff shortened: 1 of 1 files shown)


src/foam/dao/history/HistoryDAO.js
```diff
@@ -128,34 +129,46 @@ foam.CLASS({
...
+      FObject current = this.find_(x, obj);
+      Object objectId = ((PropertyInfo) obj.getClassInfo().getAxiomByName("id")).f(obj);
+      boolean isCreate = objectId == null || 
+                          current == null ||
+                          SafetyUtil.isEmpty(String.valueOf(objectId)) ||
+                          new Long(0L).equals(objectId);
+      if ( isCreate ) {
+        // do "put" first if it is "create" action.
+        FObject persistObject = super.put_(x, obj);
+        objectId = ((PropertyInfo) persistObject.getClassInfo().getAxiomByName("id")).f(obj);
```
jlhughes:
```suggestion
        objectId = obj.getProperty("id");
```

jlhughes: Why are you not taking the objectId from the persistObject?  I though the issue is that the argument obj may not yet have an Id?

xuerongNanopay: oh, just a copy and paste mistake. I will update

src/foam/dao/history/HistoryDAO.js
```diff
@@ -128,34 +129,46 @@ foam.CLASS({
...
-        User agent = subject.getRealUser();
-        FObject current = this.find_(x, obj);
-    
+      Subject subject = (Subject) x.get("subject");
+      User user = subject.getUser();
+      User agent = subject.getRealUser();
+      FObject current = this.find_(x, obj);
+      Object objectId = ((PropertyInfo) obj.getClassInfo().getAxiomByName("id")).f(obj);
+      boolean isCreate = objectId == null || 
+                          current == null ||
```
kgrgreer: Why are the other checks needed? Isn't 'current == null' sufficient?

jlhughes: We were both  fighting with this. I have something similar in serviceproviderawaredao. 
It suggest bad data - we have a 0 entry.  reviewing again.

xuerongNanopay: Yes, make sense. If there is no data find from mdao, it should be "create".

jlhughes: Looking at notes, it was an optimization to avoid the find, then the order was changed, and in the end we both need the obj object regardless, so can be removed.

xuerongNanopay: @kgrgreer We can't. Because when the obj is "create" (first time put), there is no Id in obj (in medusa). So, it can't find anything.

xuerongNanopay: We have different easydao setup in development and medusa. "current == null" is used to make sure development working correctly. and the rest of checks is used to make sure medusa currectly.

src/foam/dao/history/HistoryDAO.js
```diff
@@ -128,34 +129,44 @@ foam.CLASS({
...
-        FObject current = this.find_(x, obj);
-    
+      Subject subject = (Subject) x.get("subject");
+      User user = subject.getUser();
+      User agent = subject.getRealUser();
+      FObject current = this.find_(x, obj);
+      Object objectId = obj.getProperty("id");
+      boolean isCreate = current == null;
+
+      if ( isCreate ) {
```
jlhughes:
```suggestion
      if ( current == null ) {
```

jlhughes (approved):

PR #4594, opened 2021-01-08
[NP-3230] Preparation for scrolling step wizard

KernelDeimos:
Reviewers please note: diff displays a lot of new code in StepWizardController but another file (Step**Wizardlet**Controller) was deleted (I renamed it)

StepWizardletController hasn't been updated in a while and I found there were changes I could make to improve maintainability. The changes have been tested in the onboarding flow as well as invoices.

Additionally, WizarldetSection now has its own `isValid` property, rather than a Wizardlet using the `currentSection` property to see which section to check validity for. This will allow the scrolling wizard to know the validity of adjacent sections.


src/foam/u2/crunch/wizardflow/StepWizardAgent.js
```diff
@@ -28,7 +28,8 @@ foam.CLASS({
   requires: [
     'foam.u2.dialog.Popup',
     'foam.u2.wizard.StepWizardConfig',
-    'foam.u2.wizard.StepWizardletController'
+    // 'foam.u2.wizard.StepWizardletController'
```
tharmaman: might as well just straight up remove?

tharmaman (approved):
LGTM gunna test

src/foam/u2/crunch/wizardflow/StepWizardAgent.js
KernelDeimos: Yeah, fixed it

tharmaman:
> LGTM gunna test

invoice flow works

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+      name: 'visitedWizardlets',
+      class: 'FObjectArray',
+      of: 'foam.u2.wizard.Wizardlet'
+    },
+
+    // Convenience properties
+    {
+      name: 'currentWizardlet',
+      expression: function(wizardlets, wizardPosition) {
+        var current = wizardlets[wizardPosition.wizardletIndex];
```
Nauna, line 84: If visitedWizardlets was an int array you can just compare ints instead of objects here.

Nauna: Or you can create a boolean array with length of wizardlets and set visited wizardlets as true and therefore no comparison here.

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+    },
+    {
+      name: 'submitted',
+      class: 'Boolean'
+    }
+  ],
+
+  methods: [
+    function setupWizardletListeners(wizardlets) {
+      console.log('step wizard', this, wizardlets);
```
Nauna, line 139: don't see a point to adding this console log here.

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+      if ( this.wsub ) this.wsub.detach();
+      this.wsub = this.FObject.create();
+
+      var wi = 0, si = 0;
+      wizardlets.forEach((w, wizardletIndex) => {
+
+        // Bind availability listener for wizardlet availability
+        var isAvailable$ = w.isAvailable$;
+        console.log(wizardletIndex, w, w.isAvailable$, isAvailable$);
```
Nauna, line 148: another console log - seems more like a debugging thing. Don't think a user will ever want to know or even understand the wizardlet object.

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+        this.wsub.onDetach(isAvailable$.sub(() => {
+          console.log('X');
```
Nauna, line 150: please remove console.log

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+
+        // Bind availability listener for each wizardlet section
+        w.sections.forEach((section, sectionIndex) => {
+          var sectionPosition = this.WizardPosition.create({
+            wizardletIndex: wizardletIndex,
+            sectionIndex: sectionIndex,
+          });
+          var isAvailable$ = section.isAvailable$;
+          this.wsub.onDetach(isAvailable$.sub(() => {
+            console.log('Y');
```
Nauna, line 162: please remove console.log

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+            this.onWizardPositionAvailabilityUpdate(
```
Nauna, line 163: are you sure you have the right function here... shouldn't it be 'onSectionAvailability'

makes me wonder why are you creating a new WizardPosition here. Shouldn't their only be one wizardPosition?

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+
+      });
+
+    },
+    function detach() {
+      this.wsub.detach();
+      this.SUPER();
+    },
+    function positionAfter(pos) {
+      let subWi = pos.wizardletIndex
```
Nauna, line 176: don't think assigning functional variables for the wizardPosition helps with the readability of logic. I personally would prefer to just read the variables from where they are coming from. Especially here where we have multiple objects with similar names.
Example sections and wizardlets both have isValid.
But this is more along the lines of understanding that here in this controller we are maintaining sections, wizardlets, capabilities, ucj's, wizardletPositions and various other helper objects/properties.

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+      let subSi = pos.sectionIndex;
+      if ( subSi >= this.wizardlets[subWi].sections.length - 1 ) {
+        if ( subWi >= this.wizardlets.length - 1 ) return null;
+        subSi = 0;
+        subWi++;
+      } else {
+        subSi++;
+      }
+      return this.WizardPosition.create({
```
Nauna, line 185: Why create a new object? Wouldn't reassigning the properties be enough?

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+        if (
+          wizardlet.sections.length > 0 &&
+          wizardlet.sections[p.sectionIndex].isAvailable
+        ) return p;
+      }
+
+      return null;
+    },
+    function saveProgress() {
+      return this.wizardlets.reduce((p, wizardlet) =>
```
Nauna, line 222: If we have visited a wizardlet - why are we doing things differently? Should we not save it again?

It would be nice to not try to force synchronization of js code with reduce and promises all over the place.

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+          return true;
+        }
+        this.wizardPosition = this.nextScreen;
+        return false;
+      });
+    },
+    function back() {
+      let previousScreen = this.previousScreen;
+      if ( previousScreen == null ) {
+        throw new Error('back() called without checking canGoBack');
```
Nauna, line 249: Where does this get called from? If this is a message a user can see this needs to be a message and need to add local support.

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+      var lastWizardletIndex = start.wizardletIndex;
+      for ( let p = start ; p != null ; p = iter(p) ) {
+        // Also check isValid on the wizardlet itself
+        if ( p.wizardletIndex != lastWizardletIndex ) {
+          if ( ! this.wizardlets[lastWizardletIndex].isValid ) {
+            return false;
+          }
+        }
+        if ( typeof p.compareTo != 'function' ) {
+          debugger;
```
Nauna, line 279: :( 
debugger!!!

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+      }
+    },
+    function skipTo(pos) {
+      this.wizardPosition = pos;
+    }
+  ],
+
+  listeners: [
+    function onWizardletAvailability(wizardletIndex, value) {
+      console.log('A');
```
Nauna, line 293: please remove

src/foam/u2/wizard/StepWizardController.js
```diff
@@ -0,0 +1,321 @@
...
+      // Force a position update so views recalculate state
+      this.wizardPosition = this.WizardPosition.create({
+        wizardletIndex: this.wizardPosition.wizardletIndex,
+        sectionIndex: this.wizardPosition.sectionIndex,
+      });
+    },
+    function onSectionAvailability(sectionPosition, value) {
+      console.log('B');
```
Nauna, line 301: please remove

src/foam/u2/wizard/StepWizardController.js
KernelDeimos: WizardPosition? We've been using that for a while now. It's simplified a lot of the wizard by making the wizard position a single slot you can listen two rather than two slots that can change independently.

src/foam/u2/wizard/StepWizardController.js
KernelDeimos: [this wasn't true; StepWizardAgent calls it. I knew that when I included it but somehow forgot] ~`saveProgress` isn't actually called. I copied it from the other wizard and updated it - now I'm just keeping there on the short term in case we need it.~

On your second point, if we don't save in order the saves don't happen in order, which I've been told on multiple occasions needs to happen. (I can see the rationale - it increases client time but saves on server load)

src/foam/u2/wizard/StepWizardController.js
KernelDeimos: This error case is a developer error so an exception makes sense here

src/foam/u2/wizard/StepWizardController.js
KernelDeimos: object comparisons in javascript are reference comparisons - they perform just as well as the int or bool comparisons you suggest

src/foam/u2/wizard/StepWizardController.js
KernelDeimos: > are you sure you have the right function here... shouldn't it be 'onSectionAvailability'

fixed, thanks

> makes me wonder why are you creating a new WizardPosition here. Shouldn't their only be one wizardPosition?

I'm not aware of any reason there should only be one instance of WizardPosition so this question is confusing to me. WizardPosition objects can be compared and having multiple objects to represent different positions is useful.

src/foam/u2/wizard/StepWizardController.js
KernelDeimos: What are functional variables? I can imagine what that term would mean but it's not what I think you mean to mean.

> would prefer to just read the variables from where they are coming from

I can't really tell what you want based on that. Where are they "coming from"? I think maybe you're proposing something like `positionAfter(wi, si)` over `positionAfter(pos)`. Adding the WizardPosition model has simplified a lot of code by reducing position updates to a single slot, enabling comparison of positions, and simplifying iteration of positions.

PR #4596, opened 2021-01-08
reverting some changes

olhabn:


src/foam/u2/filter/properties/DateFilterView.js
```diff
@@ -81,12 +81,7 @@ foam.CLASS({
       documentation: `
         A date selectable by the user. Used as the second argument to the
         predicate generated by this view.
-      `,
-      preSet: function(o, n) {
-        var n1 = new Date(n);
-        n1.setDate(n1.getDate() + 1);
-        return n1;
-      }
+      `
```
olhabn, line 84: this will cause a bug if setting date from existing predicate

PR #4601, opened 2021-01-11
adjust source of country pt local …

Nauna:
to work with RichChoiceView translation setup

var summary in below is the country.name

image


jlhughes (commented):
This doesn't make sense to me. 
I can see 'in addition to' the codes, but not replacing.

Nauna:
> This doesn't make sense to me.
> I can see 'in addition to' the codes, but not replacing.

TBH I was the one that added the codes/ids ... and that was for this ticket - country translation. I can't find where the code for checking the ids exists. And since the RichChoiceView and the StructuredAddressView use the name for look up I decided to just remove the code/id ... BUUT ... I think it is safer to have both. Will adjust

jlhughes (approved):

PR #4605, opened 2021-01-11
[NP-3232] Fix bugs from NP-3188

KernelDeimos:
Introduces the following changes:
- Fix issues caused by NP-3188
- Add a stdlib function to make reduction promises more readable
- Rename `dataController` property to `wao`

### DIFF
--- a/src/foam/core/stdlib.js
+++ b/src/foam/core/stdlib.js
@@ -126,7 +126,7 @@ foam.LIB({
 foam.LIB({
   name: 'foam.Promise',
   methods: [
-    function reduce(arry, fn, parentPromise) {
+    function inOrder(arry, fn, parentPromise) {
       parentPromise = parentPromise || Promise.resolve();
       return arry.reduce((p, v) => p.then(() => fn(v)), parentPromise);
     },
--- a/src/foam/core/stdlib.js
+++ b/src/foam/core/stdlib.js
@@ -124,6 +124,17 @@ foam.LIB({
 
 
 foam.LIB({
+  name: 'foam.Promise',
+  methods: [
+    function reduce(arry, fn, parentPromise) {
+      parentPromise = parentPromise || Promise.resolve();
+      return arry.reduce((p, v) => p.then(() => fn(v)), parentPromise);
+    },
+  ]
+});
+
+
+foam.LIB({
   name: 'foam.Function',
   methods: [
     function isInstance(o) { return typeof o === 'function'; },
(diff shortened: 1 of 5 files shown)


jlhughes (approved):

src/foam/core/stdlib.js
```diff
@@ -123,6 +123,17 @@ foam.LIB({
 });
 
 
+foam.LIB({
+  name: 'foam.Promise',
+  methods: [
+    function reduce(arry, fn, parentPromise) {
```
kgrgreer: A better name for parentPromise would be opt_parentPromise to indicate that it's optional.

kgrgreer: Add documentation.

src/foam/u2/crunch/FlatteningCapabilityWizardlet.js
```diff
@@ -56,6 +56,18 @@ foam.CLASS({
 
       // Return true to prevent wizardlet from being pushed normally
       return true;
+    },
+    async function save() {
+      await foam.Promise.reduce(this.delegates, d => d.save());
```
kgrgreer: Do you really want a 'reduce' operation? Do you really compare to combine the results? Or, would a forEach/forAll/all method be more understandable?

KernelDeimos: Yes, a reduce is desired. I've been using the following technique whenever I expect promises to execute after each previous promise is complete:
```
newPromise = objects.reduce((p, object) =>
  p.then(() => object.someAsyncMethod()), Promise.resolve());
```

I could use `Promise.all`, but then wizardlets are saved an an arbitrary order which we don't want here - we actually want these asynchronous behaviours to occur synchronously.

The reason for adding this to `stdlib` is that I find I need this a lot, but it's not very readable when I write it as I've done in the example above.

kgrgreer: Reduce doesn't imply that they are executed in sequence, so maybe something like inSeq() or inSequence() or aSeq() (the same as FOAM1)?

KernelDeimos: If I read `aSeq` I would believe it to mean "not sequential" but I like `inSeq`

KernelDeimos: Oh, how about just `inOrder`?

PR #4609, opened 2021-01-12
[NP-3230] Fix call to onSectionAvailability

KernelDeimos:


jlhughes (commented):
Still a debugger call in this file.

PR #4616, opened 2021-01-13
Remove url trailing slash

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-3111

## Issues
- The default appConfig.url = "http://localhost:8080/" when combining with path starting with slash eg. "/service/verifyEmail" the url redirect to 404 page

## Changes
- Remove url trailing slash

### DIFF
--- a/src/foam/nanos/auth/email/EmailTokenService.js
+++ b/src/foam/nanos/auth/email/EmailTokenService.js
@@ -35,6 +35,7 @@ foam.CLASS({
 DAO tokenDAO = (DAO) getX().get("localTokenDAO");
 DAO userDAO  = (DAO) getX().get("localUserDAO");
 AppConfig appConfig = user.findGroup(x).getAppConfig(x);
+String url = appConfig.getUrl().replaceAll("/$", "");
 
 Token token = new Token();
 token.setUserId(user.getId());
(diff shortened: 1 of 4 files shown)


src/foam/nanos/auth/email/EmailTokenService.js
```diff
@@ -35,6 +35,7 @@ foam.CLASS({
 DAO tokenDAO = (DAO) getX().get("localTokenDAO");
 DAO userDAO  = (DAO) getX().get("localUserDAO");
 AppConfig appConfig = user.findGroup(x).getAppConfig(x);
+String url = appConfig.getUrl().replaceAll("/$", "");
```
jlhughes: Why not do this in Group.getAppConfig?

lchanmann: Sure, we can but will still have to do this for places that are not getting appConfig from group.

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4616/commits/2c113592cc59709117d65160c87c8605b1415280.

jlhughes (approved):

PR #4617, opened 2021-01-13
Fix email notification link url

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-3111

## Changes
- Use group.url for notification email link
- Load theme via user spid when the loaded theme's (via domain) spid is not same as user's
- Reload theme when the context theme is available for use in EmailsUtility but the theme's spid is not same as user's


src/foam/nanos/notification/NotificationSetting.js
```diff
@@ -67,21 +67,8 @@ foam.CLASS({
         { name: 'notification', type: 'foam.nanos.notification.Notification' }
       ],
       javaCode: `
-        X userX = x;
```
jlhughes: @lchanmann why are you removing this? 
I intentionally added it recently, so all downstream - from notificationSettings - logic didn't have to worry about this.

jlhughes (commented):
.

src/foam/nanos/notification/NotificationSetting.js
lchanmann: Reverted in https://github.com/foam-framework/foam2/pull/4617/commits/816722dda2f442fd04c5ed9425fe5917936c42b5.

src/foam/nanos/theme/Themes.js
```diff
@@ -196,17 +203,6 @@ Later themes:
         }
       }
 
-      if ( theme == null ) {
```
jlhughes, line 199: Why did you remove this?

src/foam/nanos/theme/Themes.js
```diff
@@ -174,9 +175,15 @@ Later themes:
         }
       }
 
+      if ( theme == null ) {
```
jlhughes, line 178: Why did you add this back?

src/foam/nanos/theme/Themes.js
lchanmann: It was moved up to line 178.

src/foam/nanos/theme/Themes.js
lchanmann: It allows loading group theme when there is no spid theme.

jlhughes (approved):

PR #4630, opened 2021-01-15
Removed unnecessary params on signup

zahnno:
These fields are not used - group assignment should be handled by server

### DIFF
--- a/src/foam/nanos/controller/ApplicationController.js
+++ b/src/foam/nanos/controller/ApplicationController.js
@@ -198,8 +198,7 @@ foam.CLASS({
         return {
           dao_: client$userDAO || null,
           imgPath: '',
-          group: 'system',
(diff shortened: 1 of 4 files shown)


src/foam/nanos/controller/ApplicationController.js
```diff
@@ -198,8 +198,7 @@ foam.CLASS({
         return {
           dao_: client$userDAO || null,
           imgPath: '',
-          group: 'system',
-          countryChoices_: [] // empty defaults to entire countryDAO
+          group: 'system'
```
jlhughes: What is this used for? Very odd to see it on the client side.

zahnno: Nothing

jlhughes (approved):

PR #4633, opened 2021-01-15
Update User + HumanNameTrait

jinnnn9:
nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/11652

Summary
- updated toSummary so that the user string defaults to username
- fixed legal name getter so that legal name is not set to " " if first name and last name don't exist

### DIFF
--- a/src/foam/nanos/auth/HumanNameTrait.js
+++ b/src/foam/nanos/auth/HumanNameTrait.js
@@ -35,9 +35,9 @@ foam.INTERFACE({
         return [firstName, middleName, lastName].filter(name => name).join(' ');
       },
       javaGetter: `
-        String firstName = this.getFirstName();
-        String middleName = this.getMiddleName();
-        String lastName = this.getLastName();
+        String firstName = getFirstName();
+        String middleName = getMiddleName();
+        String lastName = getLastName();
 
         StringBuilder sb = new StringBuilder();
 
--- a/src/foam/nanos/auth/HumanNameTrait.js
+++ b/src/foam/nanos/auth/HumanNameTrait.js
@@ -8,10 +8,6 @@ foam.INTERFACE({
   package: 'foam.nanos.auth',
   name: 'HumanNameTrait',
 
-  javaImports: [
-    'foam.util.SafetyUtil'
-  ],
-
   properties: [
     {
       class: 'String',
@@ -36,34 +32,17 @@ foam.INTERFACE({
       documentation: 'Full legal name of user. Appends first, middle & last name.',
       transient: true,
       expression: function(firstName, middleName, lastName) {
-        if ( firstName && middleName && lastName ) {
-          return `${firstName} ${middleName} ${lastName}`;
-        }
-
-        if ( firstName && lastName ) {
-          return `${firstName} ${lastName}`;
-        }
-
-        return '';
+        return [firstName, middleName, lastName].filter(name => name).join(' ');
       },
       javaGetter: `
-        String firstName = this.getFirstName();
-        String middleName = this.getMiddleName();
-        String lastName = this.getLastName();
-
-        if ( ! SafetyUtil.isEmpty(firstName) && 
-             ! SafetyUtil.isEmpty(middleName) &&
-             ! SafetyUtil.isEmpty(lastName) ) {
-
-          return firstName + " " + middleName + " " + lastName;
-        }
-
-        if ( ! SafetyUtil.isEmpty(firstName) && ! SafetyUtil.isEmpty(lastName) ) {
-          return firstName + " " + lastName;
-        }
-
-        return "";
-      `,
+        java.util.List<String> filteredNameList = java.util.Arrays
+          .asList(this.getFirstName(), this.getMiddleName(), this.getLastName())
+          .stream()
+          .filter(name -> !name.isEmpty())
+          .collect(java.util.stream.Collectors.toList());
+            
+        return String.join(" ", filteredNameList);
+      `
     }
   ]
 });
--- a/src/foam/nanos/auth/HumanNameTrait.js
+++ b/src/foam/nanos/auth/HumanNameTrait.js
@@ -41,11 +41,17 @@ foam.INTERFACE({
 
         StringBuilder sb = new StringBuilder();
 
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/HumanNameTrait.js
```diff
@@ -32,12 +36,33 @@ foam.INTERFACE({
...
-        return middleName != '' ? firstName + ' ' + middleName + ' ' + lastName : firstName + ' ' + lastName;
+        if ( firstName && middleName && lastName ) {
+          return `${firstName} ${middleName} ${lastName}`;
+        }
+
+        if ( firstName && lastName ) {
+          return `${firstName} ${lastName}`;
+        }
+
+        return '';
```
kgrgreer: You could also just do: return [firstName, middleName, lastName].filter(n=>n).join(' ')

src/foam/nanos/auth/HumanNameTrait.js
```diff
@@ -32,12 +36,33 @@ foam.INTERFACE({
...
+             ! SafetyUtil.isEmpty(middleName) &&
+             ! SafetyUtil.isEmpty(lastName) ) {
+
+          return firstName + " " + middleName + " " + lastName;
+        }
+
+        if ( ! SafetyUtil.isEmpty(firstName) && ! SafetyUtil.isEmpty(lastName) ) {
+          return firstName + " " + lastName;
+        }
+
```
kgrgreer: Do we want to support just 'lastName'?

jinnnn9: @kgrgreer documentation says full legal name so I returned an empty string if both first name and last name didn't exist.

src/foam/nanos/auth/HumanNameTrait.js
jinnnn9: @kgrgreer this doesn't tell which field is not an empty string though. For instance, only firstName might be available.

src/foam/nanos/auth/HumanNameTrait.js
kgrgreer: I would think that some name is better than no name. If it's invalid to not specify both first and last name, then that should be a validation error.

src/foam/nanos/auth/HumanNameTrait.js
kgrgreer: What do you mean? It just concatenates the values and includes spaces between non-null values.

jinnnn9: @kgrgreer Yes, but shouldn't the values be concatenated under certain conditions? For example, if only lastName and middleName are provided (e.g., by editing user in backoffice), we might just want to return lastName instead of concatenating lastName and middleName, and returning the result.

jinn9: @kgrgreer Or we don't need to worry about the example case above because the example above is a validation error?

src/foam/nanos/auth/HumanNameTrait.js
```diff
@@ -32,13 +32,17 @@ foam.INTERFACE({
...
-          : getFirstName() + " " + getLastName();
-      `,
+        java.util.List<String> filteredNameList = java.util.Arrays
+          .asList(this.getFirstName(), this.getMiddleName(), this.getLastName())
+          .stream()
+          .filter(name -> !name.isEmpty())
+          .collect(java.util.stream.Collectors.toList());
+            
+        return String.join(" ", filteredNameList);
+      `
```
jinn9, line 45: had to use fullpath for the java libraries in the getter because javaImports didn't work for the getter in interface

src/foam/nanos/auth/HumanNameTrait.js
```diff
@@ -32,13 +32,17 @@ foam.INTERFACE({
...
       expression: function(firstName, middleName, lastName) {
-        return middleName != '' ? firstName + ' ' + middleName + ' ' + lastName : firstName + ' ' + lastName;
+        return [firstName, middleName, lastName].filter(name => name).join(' ');
       },
       javaGetter: `
-        return ! getMiddleName().equals("")
-          ? getFirstName() + " " + getMiddleName() + " " + getLastName()
-          : getFirstName() + " " + getLastName();
-      `,
+        java.util.List<String> filteredNameList = java.util.Arrays
```
kgrgreer: This seems a little heavy weight for such a simple thing. On the client we don't care because there's a whole computer for only one user, but we need to share a server with thousands of users.

src/foam/nanos/auth/HumanNameTrait.js
```diff
@@ -32,13 +32,21 @@ foam.INTERFACE({
...
-          ? getFirstName() + " " + getMiddleName() + " " + getLastName()
-          : getFirstName() + " " + getLastName();
-      `,
+        String firstName = this.getFirstName();
+        String middleName = this.getMiddleName();
+        String lastName = this.getLastName();
+
+        StringBuilder sb = new StringBuilder();
+
+        if ( ! firstName.isEmpty() ) sb.append(firstName + " ");
```
kgrgreer: This code creates 4 fewer objects:

if ( ! firstName.isEmpty() ) sb.append(firstName);
if ( ! middleName.isEmpty() ) {
  if ( sb.length > 0 ) sb.append(' ');
  sb.append(middleName);
}
if ( ! lastName.isEmpty() ) {
  if ( sb.length > 0 ) sb.append(' ');
  sb.append(lastName);
}
return sb.toString();

src/foam/nanos/auth/HumanNameTrait.js
```diff
@@ -32,13 +32,21 @@ foam.INTERFACE({
...
       expression: function(firstName, middleName, lastName) {
-        return middleName != '' ? firstName + ' ' + middleName + ' ' + lastName : firstName + ' ' + lastName;
+        return [firstName, middleName, lastName].filter(name => name).join(' ');
       },
       javaGetter: `
-        return ! getMiddleName().equals("")
-          ? getFirstName() + " " + getMiddleName() + " " + getLastName()
-          : getFirstName() + " " + getLastName();
-      `,
+        String firstName = this.getFirstName();
```
kgrgreer: No need for 'this.' in Java.

PR #4637, opened 2021-01-18
adding default placeholder for FObjectView - first item in Fobject list

Nauna:


src/foam/u2/view/FObjectView.js
```diff
@@ -107,6 +111,9 @@ foam.CLASS({
     {
       class: 'String',
       name: 'placeholder',
+      factory: function() {
```
kgrgreer, line 114: Couldn't this be set as value: 'select...'? Can't we set values with Locales?

Nauna: So had to retest -but creating a local entry with source: 'foam.u2.view.FObjectView.PLACEHOLDER.value' doesn't work. I remember seeing that @hchoura did a string translation this way and so copied the method.

But will try to debug and discuss this on Monday. This was not a priority task. But good to finish- point is no rush.

PR #4639, opened 2021-01-18
[NP-3322] Slot property for changes to `errors_` on members of an FObjectArray

KernelDeimos:
I realized in order to propagate `capablePayload` errors it is necessary to listen for array changes and re-bind to the `errors_$` slot of each item in the array. Since the behaviour itself is generic I figure having it on FObjectArray could be very helpful for similar cases in the future.

### DIFF
--- a/src/foam/core/FObjectArray.js
+++ b/src/foam/core/FObjectArray.js
@@ -35,9 +35,7 @@ foam.CLASS({
 
           // Safely combine error lists and update error_$
           var setErrors = errorLists => {
-            errors_$.set(errorLists
-              .map(v => v || [])
-              .flat(1));
+            errors_$.set(errorLists.flat(1).filter(v => v));
           };
           // Handle an array property update
           var subToArray = val => {
--- a/src/foam/core/FObjectArray.js
+++ b/src/foam/core/FObjectArray.js
@@ -22,6 +22,48 @@ foam.CLASS({
 
   documentation: "A Property which contains an array of 'of' FObjects.",
 
+  methods: [
+    function installInProto(proto) {
+      this.SUPER(proto);
+      var self = this;
+      Object.defineProperty(proto, self.name + '$' + 'errors', {
+        get: function getArrayErrorSlot() {
+          // errors_$ will be updated with an array member's error_$ slot
+          var errors_$ = foam.core.SimpleSlot.create({ value: [] });
+          // sub will be replaced when the array changes
+          var sub = foam.core.FObject.create();
+
+          // Safely combine error lists and update error_$
+          var setErrors = errorLists => {
+            errors_$.set(errorLists
+              .map(v => (v && Array.isArray(v)) ? v : [])
+              .reduce((arry, v) => arry.concat(v), []));
+          };
+          // Handle an array property update
+          var subToArray = val => {
+            setErrors(val.map(v => v.errors_));
+            sub.onDetach(foam.core.ArraySlot.create({
+              slots: val.map(v => v.errors_$)
+            }).map(errorLists => {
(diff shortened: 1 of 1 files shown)


KernelDeimos:
Note for review: probably easiest to read the code from the bottom (code after the last comment) to the top

src/foam/core/FObjectArray.js
```diff
@@ -22,6 +22,48 @@ foam.CLASS({
...
+          // errors_$ will be updated with an array member's error_$ slot
+          var errors_$ = foam.core.SimpleSlot.create({ value: [] });
+          // sub will be replaced when the array changes
+          var sub = foam.core.FObject.create();
+
+          // Safely combine error lists and update error_$
+          var setErrors = errorLists => {
+            errors_$.set(errorLists
+              .map(v => (v && Array.isArray(v)) ? v : [])
+              .reduce((arry, v) => arry.concat(v), []));
```
kgrgreer: Does this reduce do anything?

KernelDeimos: Yes - it turns a list of lists into one big list

kgrgreer: Then couldn't you just call .flat()?

KernelDeimos: Yes (fixed)

src/foam/core/FObjectArray.js
```diff
@@ -22,6 +22,48 @@ foam.CLASS({
...
+        get: function getArrayErrorSlot() {
+          // errors_$ will be updated with an array member's error_$ slot
+          var errors_$ = foam.core.SimpleSlot.create({ value: [] });
+          // sub will be replaced when the array changes
+          var sub = foam.core.FObject.create();
+
+          // Safely combine error lists and update error_$
+          var setErrors = errorLists => {
+            errors_$.set(errorLists
+              .map(v => (v && Array.isArray(v)) ? v : [])
```
kgrgreer: What is 'v' when it isn't null but isn't an array?

src/foam/core/FObjectArray.js
```diff
@@ -22,6 +22,48 @@ foam.CLASS({
...
+      Object.defineProperty(proto, self.name + '$' + 'errors', {
+        get: function getArrayErrorSlot() {
+          // errors_$ will be updated with an array member's error_$ slot
+          var errors_$ = foam.core.SimpleSlot.create({ value: [] });
+          // sub will be replaced when the array changes
+          var sub = foam.core.FObject.create();
+
+          // Safely combine error lists and update error_$
+          var setErrors = errorLists => {
+            errors_$.set(errorLists
```
kgrgreer: Could just be: errorLists.flat().map(v=>v) ?

KernelDeimos: I want this to be consistent with checking `errors_` on an FObject. The `map` is used to turn nulls into arrays so that `concat` doesn't add nulls as array elements (there are places where errors are checked by seeing if the array length is greater than zero)

For example this outputs `[null, null]` which would indicate there are errors to display (list length > 0)
```
[null, null].flat()
```

but this outputs `[]` which is what I want it to do
```
[null, null].map(v => v || []).flat()
```

kgrgreer: [null,null].flat().filter(v=>v);
also gives [] but with 1 fewer character ;-)

KernelDeimos: Oh, my brain flipped the filter to a map for some reason

KernelDeimos: fixed!

PR #4642, opened 2021-01-18
Https config with fileDAO record

olhabn:


src/foam/nanos/jetty/HttpServer.js
```diff
@@ -316,7 +336,7 @@ foam.CLASS({
           server.addConnector(sslConnector);
 
         } catch ( java.io.FileNotFoundException e ) {
-          logger.error("No KeyStore file found at path: " + this.getKeystorePath(),
+          logger.error("No KeyStore file found " + this.getKeyCertificateFileName(),
```
jlhughes:
```suggestion
          logger.error("Failed reading keystore", this.getKeyCertificateFileName(), e.getMessage(), 
```

src/foam/nanos/jetty/HttpServer.js
```diff
@@ -325,6 +345,7 @@ foam.CLASS({
           logger.error("Error when enable the https.");
```
jlhughes:
```suggestion
          logger.error("Failed configuring https", e);
```

jlhughes (approved):

src/foam/nanos/jetty/HttpServer.js
```diff
@@ -99,10 +107,15 @@ foam.CLASS({
       class: 'FObjectProperty',
       of: 'foam.nanos.logger.Logger',
       visibility: 'HIDDEN',
-      javaFactory: `
+      javaGetter: `
```
olhabn: NPE exception line 138:
ok, so getX().get("logger") is not null when we call javaFactory but:
in DAOLogger's delegate is null although it shouldn't be and it isn't in other places. It is localLogMessageDAO.

so logger == null check will not catch it it. javaGetter fixes it but it doesn't feel right.

jlhughes: Ok, just explicitly use a StdoutLogger from here, and we'll look into setup order later.

olhabn: the problem occur after http server record updated.

maybe in some put methods context there is no localLogMessageDAO?

PR #4647, opened 2021-01-19
[NP-3341] Use ProxySlot on WizardSection isAvailable

KernelDeimos:


src/foam/u2/wizard/WizardletSection.js
```diff
@@ -79,6 +79,17 @@ foam.CLASS({
   ],
 
   methods: [
+    function activateProxySlot() {
+      var proxy$ = foam.core.ProxySlot.create();
+      this.isAvailable$ = proxy$;
+      var f = () => {
+        var isAvailable$ = this.section.createIsAvailableFor(this.data$);
```
kgrgreer, line 86: Could these two lines be combined?

src/foam/u2/wizard/WizardletSection.js
```diff
@@ -79,6 +79,17 @@ foam.CLASS({
...
+        proxy$.delegate = isAvailable$;
+      };
+      this.data$.sub(f);
```
kgrgreer: Do you need to wrap this in an this.onDetach()?

src/foam/u2/wizard/WizardletSection.js
kgrgreer: Shouldn't createIsAvailableFor(slot) be listening for slot changes? Why do we need to use the ProxySlot?

KernelDeimos: I'll be honest, after several hours of debugging this was the best solution I could come up with. I have no idea why what you described doesn't work - I tried that first.

src/foam/u2/wizard/WizardletSection.js
KernelDeimos: yes

src/foam/u2/wizard/WizardletSection.js
KernelDeimos: It's possible that the slot gets broken, but I tried adding `} catch (e) { console.error(e);` to the try blocks in Slot.js and I didn't see any exceptions get thrown.

kgrgreer:
I'll merge, but I think there must be an underlying bug that is preventing a simpler solution from working. Please schedule some time with me to debug.

PR #4650, opened 2021-01-19
Configurable Authorizer

zahnno:
## ExtendedConfigurableAuthorizer
  * Allows for grouped object access based on object values, permission templates and permissions configured. 
  * PermissionTemplates have a DAOKey array detailing when to apply to an authorizer. The authorizer also defines a DAOKey. A permission template further defines properties used to construct the permission strings along with the operation to apply it to.
  * `ExtendedConfigurableAuthorizer` Provides runtime authorization configuration through updates to entries within permissionTemplateReferenceDAO.
  * The Authorizer requires configuration, if no permissionTemplateReferences are attributed to the authorizer, ~~all requests to the service will be permitted.~~ StandardAuthorizer logic will be applied.
 
 ### Example:
A PermissionTemplateReference with a daokey of ['userDAO'], operation "read" and PermissionTemplateProperties [{ class: "PermissionTemplateProperty", propertyReference: "language"}] would use the value of language using a User object attempting to be authorized on the userDAO and compare it to the authenticated permission list.

The requestor (User making the request to access objects) may have the following permission 'userdao.read.en' granting access to all users with the values of language 'en'.

In the case where conflicts may arise from properties holding similar values, a common one for example may be a property of color, you can set impliesValue on the PermissionTemplateProperty referenced in the list of your PermissionTemplateReference, using the example above, will check for a permission of userdao.read.language[en].

Benchmarking tests and comparison results applied when reading a unauthenticated, standardAuthorized and configurableAuthorized user MDAO. View ConfigurableAuthorizerBenchmark for more information.

### Issues:
  * Ranges are not supported. 
  * Conflicts arise if property values can be the same. Example: firstName and lastName being defined as properties on a templateReference. There's no distinction between them when constructing the permission and may lead to unintentional authorization. This can be resolved by extending both permission segments to expect object properties along with a value. Currently only the property is referenced and is translated to a value which constructs the permission to check against the requestors permission list. **RESOLVED**

### Note:
These changes do not change any existing functionality.


kgrgreer:
I would like to get a demo of this before we merge. Thx

zahnno:
@kgrgreer Cache support added.

jlhughes (approved):

src/foam/demos/clock/index.html
```diff
@@ -0,0 +1,12 @@
+<html>
```
kgrgreer, line 1: Looks like you have some extra files in this PR.

PR #4653, opened 2021-01-20
Np 2337/fixing on sorting not working on fitst click on arrow if set from memento

olhabn:
on setting order from memento fixes:
- arrow icon set up properly
- order set correctly on click on sorting arrow after sort order set from memento

### DIFF
--- a/src/foam/u2/view/UnstyledTableView.js
+++ b/src/foam/u2/view/UnstyledTableView.js
@@ -464,21 +464,6 @@ foam.CLASS({
           var view = this;
           view.props = this.returnPropertiesForColumns(view, view.columns_);
 
-          if ( this.memento && this.memento.paramsObj.c ) {
-            var columns = this.memento.paramsObj.c.split(',');
-            for ( var c of columns ) {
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/UnstyledTableView.js
```diff
@@ -673,6 +663,23 @@ foam.CLASS({
...
+              for ( var c of columns ) {
+                if ( this.shouldColumnBeSorted(c) && ! c.includes('.')) {
+                  var prop = view.props.find(p => p.fullPropertyName === c.substr(0, c.length - 1) );
+                  if ( prop ) {
+                    if ( c[c.length - 1] === this.DESCENDING_ORDER_CHAR )
+                      this.order = this.DESC(prop.property);
+                    else
+                      this.order = prop.property;
+                    dao = dao.orderBy(this.order);
+                  }
```
kgrgreer, line 678: Should you break out of the loop at this point?

olhabn: I'm not sure why should we. 
we are going through properties that have sorting order char for which we set order and do sorting and then we go to the next one

PR #4656, opened 2021-01-20
adding now and when

olhabn:


src/foam/dao/MDAO.java
```diff
@@ -242,4 +242,16 @@ public Object cmd_(X x, Object cmd) {
     }
     return super.cmd_(x, cmd);
   }
+
+  synchronized Object now() {
+    return state_;
+  }
+
+  Object when(Object state) {
+    synchronized ( state ) {
```
kgrgreer: Doesn't need to be synchronized.

PR #4657, opened 2021-01-20
PMs

nick-prat:
Added PMs to DUGRuleAction, DigWebAgent, SugarWebAgent, CapabilityPayloadDAO

### DIFF
--- a/src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
+++ b/src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
@@ -229,8 +229,7 @@ foam.CLASS({
     {
       name: 'put_',
       javaCode: `
-      var pm = new PM(CapabilityPayloadDAO.getOwnClassInfo().getId(), "capabilityPayloads.put");
-      pm.setStartTime(System.currentTimeMillis());
+      var pm = new PM(CapabilityPayloadDAO.getOwnClassInfo().getId(), "put");
 
       try {
         CapabilityPayload receivingCapPayload = (CapabilityPayload) obj;
@@ -244,7 +243,6 @@ foam.CLASS({
         processCapabilityList(x, grantPath, capabilityDataObjects, currentCapabilityDataObjects);
 
         var ret =  find_(x, receivingCapPayload.getId());
-        pm.setEndTime(System.currentTimeMillis());
         return ret;
       } catch(Throwable t) {
         pm.setErrorMessage(t.getMessage());
--- a/src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
+++ b/src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
@@ -245,8 +245,7 @@ foam.CLASS({
         var ret =  find_(x, receivingCapPayload.getId());
         return ret;
       } catch(Throwable t) {
-        pm.setErrorMessage(t.getMessage());
-        pm.setIsError(true);
+        pm.error(x, t.getMessage());
         throw t;
       } finally {
         pm.log(x);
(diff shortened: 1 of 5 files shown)


src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
```diff
@@ -238,7 +243,16 @@ foam.CLASS({
         List grantPath = ((CrunchService) x.get("crunchService")).getGrantPath(x, receivingCapPayload.getId());
         processCapabilityList(x, grantPath, capabilityDataObjects, currentCapabilityDataObjects);
 
-        return find_(x, receivingCapPayload.getId());
+        var ret =  find_(x, receivingCapPayload.getId());
+        pm.setEndTime(System.currentTimeMillis());
+        return ret;
+      } catch(Throwable t) {
+        pm.setErrorMessage(t.getMessage());
```
jlhughes: Why not just call pm.error(x, t.getMessage()); ?

src/foam/nanos/dig/DUGRuleAction.js
```diff
@@ -44,19 +44,28 @@ foam.CLASS({
...
-              getBearerToken(), 
-              getFormat(),
-              new foam.lib.AndPropertyPredicate(x, 
-                new foam.lib.PropertyPredicate[] {
-                  new foam.lib.ExternalPropertyPredicate(),
-                  new foam.lib.NetworkPropertyPredicate(), 
-                  new foam.lib.PermissionedPropertyPredicate()}),
-              true);
-              
+            var pm = new PM(DUGRuleAction.getOwnClassInfo().getId(), "dugrule." + dugRule.getId() + "." + dugRule.getName());
```
jlhughes: No need to do this here, 
```suggestion
            var pm = new PM(DUGRuleAction.getOwnClassInfo().getId(), "dugrule.", dugRule.getId(), ".", dugRule.getName());
```
And PMs already combine names with ':' like a path.

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
```diff
@@ -228,6 +229,10 @@ foam.CLASS({
     {
       name: 'put_',
       javaCode: `
+      var pm = new PM(CapabilityPayloadDAO.getOwnClassInfo().getId(), "capabilityPayloads.put");
+      pm.setStartTime(System.currentTimeMillis());
```
jlhughes: This is also not necessary. PM creation does this for you.

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
```diff
@@ -238,7 +243,16 @@ foam.CLASS({
         List grantPath = ((CrunchService) x.get("crunchService")).getGrantPath(x, receivingCapPayload.getId());
         processCapabilityList(x, grantPath, capabilityDataObjects, currentCapabilityDataObjects);
 
-        return find_(x, receivingCapPayload.getId());
+        var ret =  find_(x, receivingCapPayload.getId());
+        pm.setEndTime(System.currentTimeMillis());
```
jlhughes: This is also not necessary as pm.error and pm.log both do this for you.

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
```diff
@@ -228,6 +229,10 @@ foam.CLASS({
     {
       name: 'put_',
       javaCode: `
+      var pm = new PM(CapabilityPayloadDAO.getOwnClassInfo().getId(), "capabilityPayloads.put");
```
jlhughes: With regard to naming, in this case the first part is already CapabilityPayloadDAO, so the second 'capabilityPayloads.put can be just 'put'

src/foam/nanos/crunch/connection/CapabilityPayloadDAO.js
nick-prat: I saw that the `try { ... } catch (...) { ... } finally { ... pm.log(x) }` pattern was used elsewhere so I just stuck to that.

jlhughes: yes, the try/catch/finally is fine, 
It's the unnecessary calls to setErrorMessage, setIsError, setTime... 
just call 
pm.error(x, errorMesage) 
and
pm.log(x)

jlhughes (approved):

PR #4663, opened 2021-01-22
fix actions alingment

olhabn:
image

### DIFF
--- a/src/foam/u2/filter/FilterView.js
+++ b/src/foam/u2/filter/FilterView.js
@@ -177,6 +177,10 @@ foam.CLASS({
       float: right;
       padding-top: 8px;
     }
+
+    ^hide-overflow {
+      overflow: hidden;
+    }
   `,
 
   messages: [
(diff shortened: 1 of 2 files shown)


kgrgreer:
The advanced filter panel no longer right aligns with the search field and the padding for the "8 of 8 selected" isn't the same at the top and bottom.

olhabn:
fixed 
image

src/foam/u2/filter/FilterView.js
```diff
@@ -321,6 +325,7 @@ foam.CLASS({
             .end()
           .end()
           .start()
+            .addClass(this.myClass('hide-overflow'))
```
kgrgreer: given it is so directly mapped to a particular CSS attribute, you could just do .style({overflow: 'hidden'}) instead of addClass() and the defining the CSS for that class.

olhabn: fixed

PR #4667, opened 2021-01-22
[NP-3255] fix CapabilityDAO Table

jinnnn9:
Address: https://nanopay.atlassian.net/browse/NP-3255

Cause: 
 - capabilityWizardlet is created without setting capability property, and `capability` defaults to `undefined`. When `id` is accessed, its expression function is invoked and tries to access `capability.id`. This raises an error because it is trying to access `id` of `undefined`. 

Solution: 
- handle when capability is `undefined`.

----

Is UI supposed to look this way for the wizardlet column? Or will it be fixed?

Screen Shot 2021-01-22 at 2 26 22 PM

### DIFF
--- a/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
+++ b/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
@@ -25,7 +25,7 @@ foam.CLASS({
     {
       name: 'id',
       expression: function (capability) {
-        return 'capability,' + capability.id;
+        return capability ? 'capability,' + capability.id : 'capability undefined';
       }
     },
 
--- a/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
+++ b/src/foam/nanos/crunch/ui/CapabilityWizardlet.js
@@ -25,7 +25,7 @@ foam.CLASS({
     {
       name: 'id',
       expression: function (capability) {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/crunch/ui/CapabilityWizardlet.js
```diff
@@ -25,7 +25,7 @@ foam.CLASS({
     {
       name: 'id',
       expression: function (capability) {
-        return 'capability,' + capability.id;
+        return capability ? 'capability,' + capability.id : 'capability undefined';
```
kgrgreer: "why do we add the 'capability,' prefix? Could we just use the capability.id?
Could this just be?
return capability ? capability.id : '';  ?

Nauna: So I believe the reason this was set to 'capability' prefix was cause the wizardlet is itself an object and also stores the ucj (as a property) <- or use too ... now wizardlet stores UCJWAO object as property.

Therefore the pt was to specify the id.

However Kgr brings up a good pt. 
Either this uses its own id or it should just return cap.id and it will be known that this is its convention.
I vote for removing 'capability' prefix.

Nauna (approved):

PR #4668, opened 2021-01-25
Ref view ucj

Nauna:
THIS PR reveals some bugs.
- 
image
Which might be caused by having 2 properties with ModeAltView. Although commenting out one of the view property.property will get you somewhere - the bug of max call stack still persists.

So then comment out one of the property views ... you then get
image


kgrgreer:
What happens if you change line 52 of ModeAltView to:
        this.args = args;

PR #4669, opened 2021-01-25
[NP-3368] - ChoiceView Auto Select Single Option Update

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-3368

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -48,6 +48,13 @@ foam.CLASS({
       }
     },
     {
+      class: 'Boolean',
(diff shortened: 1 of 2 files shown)


src/foam/u2/view/FObjectView.js
```diff
@@ -47,6 +47,13 @@ foam.CLASS({
         };
       }
     },
+    {
+      class: 'Boolean',
+      name: 'choiceSelected',
+      expression: function(objectClass) {
```
kgrgreer: Why is this needed? Why not use objectClass?

PR #4672, opened 2021-01-26
DNM - Np 3358

noodlemoodle:
https://nanopay.atlassian.net/browse/NP-3358

- fix business ownership capability (and its prereqs) going into ACTION_REQUIRED if try to reopen after GRANTED
- fix granted minmaxes showing up in wizard when they shouldn't. e.g., after completing unlock domestic payments, business ownership will show up in the international payments wizard

### DIFF
--- a/src/foam/nanos/crunch/ServerCrunchService.java
+++ b/src/foam/nanos/crunch/ServerCrunchService.java
@@ -454,22 +454,27 @@ public class ServerCrunchService extends ContextAwareSupport implements CrunchSe
   public WizardState getWizardState(X x, String capabilityId) {
     var subject = (Subject) x.get("subject");
 
+    DAO capabilityDAO = (DAO) x.get("capabilityDAO");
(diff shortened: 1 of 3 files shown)


src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -513,7 +520,14 @@ private WizardState getWizardStateFor_(X x, Subject s, String capabilityId) {
       try {
         var ucj = getJunction(x, cap.getId());
         if ( IN(UserCapabilityJunction.STATUS, grantedStatuses).f(ucj) ) {
-          granted.add(cap.getId());
+          // if a minmax capability is in GRANTED status, its prerequisites should also be added
+          // as a part of granted list, so the wizrd will know to ignore the prerequisites
+          if ( isMinmax ) {
```
KernelDeimos, line 525: `isMinmax` should be `isPrerequisiteAware` - this condition is true for any capability that has a wizardlet implementing PrerequisiteAware.

PR #4674, opened 2021-01-26
[3378] - Add Support To Ignore Base Class As Choice When No Strategies Fetched

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-3378

### DIFF
--- a/src/foam/u2/view/FObjectView.js
+++ b/src/foam/u2/view/FObjectView.js
@@ -73,6 +73,12 @@ foam.CLASS({
     },
     {
       class: 'Boolean',
+      name: 'skipBaseClass',
+      documentation: 'If true, skips of base-class as a choice when no strategies fetched.',
+      value: false
+    },
+    {
+      class: 'Boolean',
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/FObjectView.js
```diff
@@ -71,6 +71,12 @@ foam.CLASS({
       class: 'Class',
       name: 'of'
     },
+    {
+      class: 'Boolean',
+      name: 'skipBaseClass',
+      documentation: 'If true, skips of base-class as a choice when no strategies fetched.',
```
kgrgreer, line 77: Can you expand on this to say when/why you would want to do this?

nanopay-arthur: When we are adding a bank account to a contact, but the user has yet to unlock any corridor strategies, FObjectView sets the choices to the base class of BankAccount.js. In this case it doesn't make sense to have the base class as an option, since we only allow the user to add subclasses of BankAccount.

PR #4677, opened 2021-01-26
Np-3279 medusa ssl connection

xuerongNanopay:
ticket: https://nanopay.atlassian.net/browse/NP-3279


jlhughes (commented):
Looks great.  Tested.  Works great.
Changes - give me your thoughts. 
Place the SslContextFactory in the context - via nspec setup.
Then simply pull the factory from the context and use it to:
- create the ssl context
- for all the keypath, password properties.   Then not duplicated.  
Where ever you need the passwords and such, just take them from the factory. 
Or split the SslContextFactory into two parts, the context factory itself, and 
SocketSslConfig with the keystore info.   
Both available in the context.

jlhughes (commented):
Great.

jlhughes (approved):

PR #4684, opened 2021-01-27
[NP-3319] Fix currentSection property

KernelDeimos:
[ vacation disclaimer: pull-request is for work done on a previous day ]


jinnnn9:
fixes NP-3319 and a bug where expression returns undefined

tharmaman (approved):

PR #4688, opened 2021-01-28
Support treviso backoffice

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-3386

## Changes
- Use menuDAO default order i.e. order, label
- Place user.group in administrative section to allow msp-admin to save user

### DIFF
--- a/src/foam/nanos/auth/User.js
+++ b/src/foam/nanos/auth/User.js
@@ -849,7 +849,7 @@ foam.RELATIONSHIP({
   },
   targetProperty: {
     hidden: false,
-    section: 'systemInformation',
+    section: 'administrative',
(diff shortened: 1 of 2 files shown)


jlhughes (approved):

src/foam/nanos/auth/User.js
```diff
@@ -849,7 +849,7 @@ foam.RELATIONSHIP({
   },
   targetProperty: {
     hidden: false,
-    section: 'systemInformation',
+    section: 'administrative',
```
nanoscott, line 852: @lchanmann .. there is no 'administive' section .. why did you change this? now the group cannot be seen

PR #4696, opened 2021-01-28
added showToast option

minsunny:
https://nanopay.atlassian.net/browse/NP-3152

added 'showToast' option in CheckPendingAgent.js to prevent from throwing non necessary toast message but throwing it when cap card clicked.

PR #4705, opened 2021-01-29
end sequence

minsunny:
additional fix that related to https://github.com/foam-framework/foam2/pull/4696/files
fix of onboarding pop up by making the crunch sequence end

### DIFF
--- a/src/foam/u2/crunch/wizardflow/CheckPendingAgent.js
+++ b/src/foam/u2/crunch/wizardflow/CheckPendingAgent.js
@@ -59,6 +59,7 @@ foam.CLASS({
       if ( ucj.status !== this.CapabilityJunctionStatus.AVAILABLE ) {
         var statusPending = ucj.status === this.CapabilityJunctionStatus.PENDING;
         var shouldReopen = await this.crunchService.maybeReopen(this.ctrl.__subContext__, ucj.targetId);
+
         if ( ! shouldReopen && this.showToast ) {
           var message = statusPending ? this.CANNOT_OPEN_PENDING : this.CANNOT_OPEN_GRANTED;
           this.ctrl.notify(message, '', this.LogLevel.INFO, true);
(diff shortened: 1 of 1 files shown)


Nauna (approved):

Nauna:
https://nanopay.atlassian.net/browse/NP-3350

src/foam/u2/crunch/wizardflow/CheckPendingAgent.js
```diff
@@ -73,6 +74,10 @@ foam.CLASS({
         this.cancelled = true;
         this.sequence.endSequence();
       }
+      if ( ! shouldReopen && statusPending ) {
```
KernelDeimos: It looks like the previous overall condition for ending the sequence was the following:

```
( status !== AVAILABLE && ! shouldReopen ) || ( shouldReopen && capabilities.length < 1 )
```

However here, when `showToast=false`, the condition is now this:
```
! shouldReopen && status == PENDING
```

I think a simpler condition is better so this is good, but I also wonder - why not just ` ! shouldReopen `? @Nauna might remember if there's a reason we had it more complicated to begin with.

src/foam/u2/crunch/wizardflow/CheckPendingAgent.js
```diff
@@ -73,6 +74,10 @@ foam.CLASS({
...
+        this.cancelled = true;
+        this.sequence.endSequence();
```
KernelDeimos: **Related to previous comment** - `this.sequence.endSequence()` appears three times in this method now, however I believe it can be placed into a single condition at the end.

KernelDeimos (changes requested):

src/foam/u2/crunch/wizardflow/CheckPendingAgent.js
Nauna: In early days - we had limited cases and may have overlooked the general scope. Think this is just legacy of building what we needed at the time.

src/foam/u2/crunch/wizardflow/CheckPendingAgent.js
minsunny: updated

KernelDeimos (approved):

PR #4706, opened 2021-01-29
Np 3321/fixing columns not being visible and table styling in DetailView

olhabn:
I don't think we need to filter networkTransient properties out as some of them are just set on front end

ticket: https://nanopay.atlassian.net/browse/NP-3321


jlhughes:
Removing networkTransient: would this not allow selection/viewing of these properties on the client even though they could not get server data?  So an expectation to export but it would not be useful?

jlhughes (approved):

PR #4708, opened 2021-01-30
Auto add an Index on spid for ServiceProviderAware DAOs.

jlhughes:
Also use MDAO.GET_MDAO_CMD to attempt to get proxied MDAO to add Indexes on.

### DIFF
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -225,7 +225,7 @@ foam.CLASS({
         }
 
         if ( getServiceProviderAware() ) {
-         delegate = new foam.nanos.auth.ServiceProviderAwareDAO.Builder(getX())
+          delegate = new foam.nanos.auth.ServiceProviderAwareDAO.Builder(getX())
             .setDelegate(delegate)
             .build();
 
--- a/src/foam/dao/EasyDAO.js
+++ b/src/foam/dao/EasyDAO.js
@@ -179,7 +179,11 @@ foam.CLASS({
             }
             delegate = getMdao();
             if ( getIndex() != null && getIndex().length > 0 ) {
-              ((foam.dao.MDAO) getMdao()).addIndex(getIndex());
+              if ( delegate instanceof foam.dao.MDAO ) {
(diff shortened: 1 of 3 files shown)


src/foam/dao/EasyDAO.js
```diff
@@ -221,9 +225,26 @@ foam.CLASS({
         }
 
         if ( getServiceProviderAware() ) {
-          delegate = new foam.nanos.auth.ServiceProviderAwareDAO.Builder(getX())
+         delegate = new foam.nanos.auth.ServiceProviderAwareDAO.Builder(getX())
```
jlhughes:
```suggestion
          delegate = new foam.nanos.auth.ServiceProviderAwareDAO.Builder(getX())
```

src/foam/dao/EasyDAO.js
```diff
@@ -1026,14 +1047,14 @@ model from which to test ServiceProvider ID (spid)`,
         return this;
       },
       javaCode: `
-        if ( getInnerDAO() == null && getMdao() != null ) {
-          ((foam.dao.MDAO) getMdao()).addIndex(props);
-        } else {
-          Logger logger = (Logger) getX().get("logger");
-          if ( logger == null ) {
-            logger = new foam.nanos.logger.StdoutLogger();
+        if ( getMdao() != null ) {
```
lchanmann: Are we no longer want to log when the nspec tries to add index to wrapper easy DAO? eg. no log for when adding index in localUserDAO?

jlhughes: The  cmd(MDAO.GET_MDAO_CMD) will traverse the delegates and return if it his another EasyDAO or a real MDAO, so we can attempt to add indexes on nspecs such as localUserDAO, and if we can't then it warns.

lchanmann: but we don't setMdao() when there is an innerDAO, do we? so getMdao() here will be null, no?

jlhughes: yeah, I can just call the cmd on the delegate, no concern for the mdao, good call.

lchanmann (approved):
LGTM

PR #4716, opened 2021-02-01
[NP-3436] MedusaAdapterDAO MDAO when/state

jlhughes:


src/foam/dao/MDAO.java
```diff
@@ -74,8 +74,40 @@ public void detach() {
   protected Object   writeLock_ = new Object();
   protected Set      unindexed_ = new HashSet();
 
-  public final static String GET_MDAO_CMD = "GET_MDAO_CMD"; 
-  
+  /**
+   * DAO Command issued on a DAO stack to return the MDAO.
+   * Used by Medusa to directly update the MDAO.
+   */
+  public final static String GET_MDAO_CMD = "GET_MDAO_CMD";
```
kgrgreer: If you're grabbing the MDAO directly then do you need the other commands?

src/foam/dao/MDAO.java
```diff
@@ -74,8 +74,40 @@ public void detach() {
...
+
+  /**
+   * DAO Command to retrieve current MDAO state. Intented
+   * to be used in Command WhenCmd
+   */
+  public final static String STATE_CMD = "STATE_CMD";
```
kgrgreer: Shouldn't this be the NOW_CMD, since that's the method you're calling? It's just an implementation detail that it actually returns its state.

src/foam/dao/MDAO.java
kgrgreer: Although, to be clear, I think it's a better idea to have the other commands than direct access to the MDAO, since the CMDs allow for decoration. So my question is really: do you need both? (And if not, the MDAO would be the one to remove.)

src/foam/nanos/medusa/MedusaAdapterDAO.js
```diff
@@ -96,6 +146,32 @@ It then marshalls it to the primary mediator, and waits on a response.`,
...
+    {
+      name: 'init_',
+      javaCode: `
+      getIndexStateDAO();
+      `
+    },
+    {
+      name: 'find_',
+      javaCode: `
+      DAO dao = getDelegate();
```
kgrgreer: Should this block of code be made into a helper method, since it appears in more than one spot?
Is this thread-safe?

kgrgreer: Maybe name it something like getReadDelegate().

src/foam/nanos/http/NanoRouter.java
```diff
@@ -93,9 +93,14 @@ protected void service(final HttpServletRequest req, final HttpServletResponse r
     resp.setHeader("X-Frame-Options", "deny");
 
     try {
-      if ( ! spec.getEnabled() ) {
+      if ( spec == null ) {
+        System.err.println("Service not found: " + serviceKey);
+        resp.sendError(resp.SC_NOT_FOUND, "No service found for: "+serviceKey);
+        return;
+      } else if ( ! spec.getEnabled() ) {
```
kgrgreer: No need for 'else' if you've already returned in the 'if'.

src/foam/nanos/medusa/MedusaAdapterDAO.js
```diff
@@ -92,13 +111,55 @@ It then marshalls it to the primary mediator, and waits on a response.`,
       return formatter;
     }
   };
+
+  // TODO: pool of reusable WriteLocks
+  protected Map<Object, WriteLock> findLocks_ = new HashMap();
```
kgrgreer, line 118: How is this collection synchronized?

jlhughes: yes, multiple puts of the same id need synchronization.

src/foam/nanos/http/NanoRouter.java
```diff
@@ -121,7 +127,7 @@ public Object create(X x) {
         agent.execute(requestContext);
       }
     } catch (Throwable t) {
-      System.err.println("Error serving " + serviceKey + " " + path);
+      System.err.println("Error serving: "+serviceKey+" "+path+" "+t.getMessage());
```
jlhughes:
```suggestion
      System.err.println("Error serving: " + serviceKey + " " + path + " " + t.getMessage());
```

PR #4721, opened 2021-02-01
DIG session context fix

nick-prat:
If a user logs in with Basic auth but also supplies their session id, it will wrap a new session context around the old one.  This causes the log error where it prints the user & userid multiple times.  To fix this I have this solution to reuse the old session context if it's been properly setup (old session context's user id == login user id).  

Another potential solution would be to check the session context in `AuthWebAgent.authenticate` for the userId and return the session if the ids match and the user's password is valid, but that would skip any login logic we currently use (Such as the LoginAttempAuthService).

### DIFF
--- a/src/foam/nanos/auth/UserAndGroupAuthService.js
+++ b/src/foam/nanos/auth/UserAndGroupAuthService.js
@@ -142,11 +142,15 @@ foam.CLASS({
           throw new AccessDeniedException(e);
         }
 
+        Session session = x.get(Session.class);
+        if ( session.getUserId() == user.getId() ) {
+          return user;
+        }
+
         // Freeze user
         user = (User) user.fclone();
         user.freeze();
-
-        Session session = x.get(Session.class);
+        
         session.setUserId(user.getId());
 
         if ( check(userX, "*") ) {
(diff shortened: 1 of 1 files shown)


src/foam/nanos/auth/UserAndGroupAuthService.js
```diff
@@ -142,11 +142,15 @@ foam.CLASS({
           throw new AccessDeniedException(e);
         }
 
+        Session session = x.get(Session.class);
+        if ( session.getUserId() == user.getId() ) {
```
kgrgreer, line 148: Please add comment to explain why.

jlhughes:
Why  is both session and 'Basic' auth being used?
but - this is the norm, as we have username/password, but also take the session that the browser generated. 
so what is different in the flow you are discussing?

nick-prat:
@jlhughes This affects the DIG basic auth flow (and the front end if you managed to get to the login screen while logged in), it looks like we're not preventing the user from logging in with basic auth if they're already logged in on the back end. This doesn't prevent the user from doing that, but will reuse the session context instead of creating a new one.

PR #4730, opened 2021-02-02
DO NOT MERGE: do not show validation for props that have its default value

noodlemoodle:
https://nanopay.atlassian.net/browse/NP-3391

### DIFF
--- a/src/foam/u2/detail/SectionedDetailPropertyView.js
+++ b/src/foam/u2/detail/SectionedDetailPropertyView.js
@@ -246,11 +246,10 @@ foam.CLASS({
           var errorSlot = prop.validateObj && prop.validationTextVisible ?
             this.data.slot([[prop.name], function() {
               var self = this;
-              var ax = self.cls_.getAxiomByName(prop.name);
-              if ( ax.isDefaultValue(self[prop.name]) ) return '';
-              if ( Array.isArray(ax.validateObj) ) 
-                return ax.validateObj[1].apply(self, ax.validateObj[0]);
-              return ax.validateObj.apply(self, foam.Function.argNames(ax.validateObj).map(function(arg) { return self[arg]; }));
+              if ( prop.isDefaultValue(self[prop.name]) ) return '';
+              if ( Array.isArray(prop.validateObj) ) 
+                return prop.validateObj[1].apply(self, prop.validateObj[0]);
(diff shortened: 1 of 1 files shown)


jlhughes:
@noodlemoodle Can you add a screen shot please. Thanks.

noodlemoodle:
sure so for example this postalcode field in brazil onboarding flow will not show validation when the value is the default empty string
Screen Shot 2021-02-02 at 2 52 24 AM
Screen Shot 2021-02-02 at 2 52 19 AM
Screen Shot 2021-02-02 at 2 52 10 AM

jlhughes:
Tested with Treviso onboarding, and apart from inputting extra data, as there is no way to know what is really required, it works fine.

jlhughes (approved):

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -244,7 +244,15 @@ foam.CLASS({
...
 
           var errorSlot = prop.validateObj && prop.validationTextVisible ?
-            this.data.slot(prop.validateObj) :
+            this.data.slot([[prop.name], function() {
+              var self = this;
+              var ax = self.cls_.getAxiomByName(prop.name);
+              if ( ax.isDefaultValue(self[prop.name]) ) return '';
+              if ( Array.isArray(ax.validateObj) ) 
+                return ax.validateObj[1].apply(self, ax.validateObj[0]);
+              else
```
kgrgreer: No need for 'else'.

src/foam/u2/detail/SectionedDetailPropertyView.js
```diff
@@ -244,7 +244,15 @@ foam.CLASS({
         .add(this.slot(function(mode, prop, prop$label) {
 
           var errorSlot = prop.validateObj && prop.validationTextVisible ?
-            this.data.slot(prop.validateObj) :
+            this.data.slot([[prop.name], function() {
+              var self = this;
+              var ax = self.cls_.getAxiomByName(prop.name);
```
kgrgreer: What is the difference between 'ax' and 'prop'?

noodlemoodle: they are the same thing, i was doing something else wrong when prop didn't work.
updated, thanks

kgrgreer: Much better. Let's wait until after UX meeting before merging.

PR #4731, opened 2021-02-02
[NP-3464] File select - table view

jlhughes:


KolombetM (approved):

src/services
```diff
@@ -357,8 +357,9 @@ p({
   "name": "httpBlobService",
   "authenticate": false,
   "serviceScript": """
-    httpBlobService = new foam.nanos.blob.HttpBlobService(x, x.get("blobStore"));
-    return new foam.nanos.http.SessionWebAgent("service.run.httpBlobService", httpBlobService);
+    return new foam.nanos.blob.HttpBlobService(x, x.get("blobStore"));
```
jlhughes, line 360: @Mykola97 let's leave httpBlob and httpFileService requiring login/authentication, and use service/file for unauthenticated requests.  So we'll revert these two changes.

PR #4733, opened 2021-02-02
Add is class predicate

smirnovaka:


lchanmann (approved):

src/foam/nanos/ruler/predicate/IsClassPredicate.js
```diff
@@ -0,0 +1,36 @@
...
+      class: 'Class',
+      name: 'of',
+      documentation: 'class that we want the object to be an class of'
+    }
+  ],
+  methods: [
+    {
+      name: 'f',
+      javaCode: `
+        return EQ(
```
kgrgreer, line 30: This is going to recreate all of the objects forming the mlang for each predicate check, which could be a lot.
Better to store this mlang in a property (or extend ProxyPredicate and use delegate).  This will generate a lot fewer objects and be much faster.

PR #4736, opened 2021-02-02
using urls to set service

olhabn:


src/foam/nanos/boot/NSpec.js
```diff
@@ -275,16 +279,7 @@ foam.CLASS({
...
-            class: 'foam.comics.v2.DAOBrowseControllerView',
-            data: service,
-            config: {
-              class: 'foam.comics.v2.DAOControllerConfig',
-              dao: service,
-              createPredicate: foam.mlang.predicate.True,
-              editPredicate: foam.mlang.predicate.True
-            }
-          });
+          this.ctrl.memento.value$.set('admin.data' + foam.nanos.controller.Memento.SEPARATOR + this.name);
```
olhabn: the least ugly approach I could come up with

kgrgreer: Could you just do this.ctrl.memento.value = 'admin.data' + foam.nanos.controller.Memento.SEPARATOR + this.name;

?

PR #4739, opened 2021-02-02
[NP-3411] Payables menu not displaying bug

jinnnn9:
remove `reset` as `purge` also does update on the dao.


jlhughes:
LGTM. Will wait on review by @zahnno

PR #4743, opened 2021-02-03
NP-2963 Random early session timeout / forced logout Medusa

xuerongNanopay:
link: https://nanopay.atlassian.net/browse/NP-2963
Can not test locally. Will test after redeploy.


src/foam/box/SessionServerBox.java
```diff
@@ -192,4 +192,13 @@ public void send(Message msg) {
       XLocator.set(null);
     }
   }
+
+  private String getRemoteHost(HttpServletRequest req) {
```
jlhughes:
```suggestion
  protected String getRemoteHost(HttpServletRequest req) {
```

jlhughes (commented):
Looks good

PR #4757, opened 2021-02-04
fixing using capability id instead of ucj id

olhabn:
ticket https://nanopay.atlassian.net/browse/NP-3486


src/foam/u2/view/ReadManyToManyRelationshipPropertyView.js
```diff
@@ -26,13 +26,20 @@ foam.CLASS({
     'foam.comics.v2.DAOControllerConfig'
   ],
 
+  property: [
```
olhabn, line 29: not related to current pr. just some refactoring

PR #4766, opened 2021-02-05
fixing tree view

olhabn:
image

nanopay pr: https://github.com/nanoPayinc/NANOPAY/pull/11880

### DIFF
--- a/src/foam/nanos/menu/VerticalMenu.js
+++ b/src/foam/nanos/menu/VerticalMenu.js
@@ -105,7 +105,7 @@ foam.CLASS({
               startExpanded: true,
               query: self.menuSearch$,
               onClickAddOn: function(data) { self.openMenu(data); },
-              selection$: self.currentMenu$,
+              selection: self.currentMenu,
               formatter: function(data) {
                 this.translate(data.id + '.label', data.label);
               }
--- a/src/foam/nanos/menu/VerticalMenu.js
+++ b/src/foam/nanos/menu/VerticalMenu.js
@@ -39,7 +39,6 @@ foam.CLASS({
     display: inline-block;
     position: absolute;
     height: calc(100vh - 80px);
-    width: 250px;
     overflow-x: hidden;
     z-index: 100;
     font-size: 26px;
@@ -53,13 +52,6 @@ foam.CLASS({
     margin: 14px 0 0;
   }
 
-  ^ .foam-u2-view-TreeViewRow-label {
-    font-weight: 300;
-  }
-
-  ^ .foam-u2-view-TreeViewRow {
-    width: 100%;
-  }
   ^ .tree-view-height-manager {
     margin-bottom: 40px;
   }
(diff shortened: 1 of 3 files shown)


src/foam/nanos/menu/VerticalMenu.js
```diff
@@ -114,7 +105,7 @@ foam.CLASS({
               startExpanded: true,
               query: self.menuSearch$,
               onClickAddOn: function(data) { self.openMenu(data); },
-              selection$: self.currentMenu$,
+              selection: self.currentMenu,
```
olhabn, line 108: it creates a bug for menus under 'Data Managment'

olhabn: and it's not a good fix but I don't have a better solution

src/foam/u2/view/TreeView.js
```diff
@@ -200,37 +207,44 @@ foam.CLASS({
...
               show(this.hasChildren$).
               style({
-                'visibility': 'visible',
-                'font-size':  '16px',
-                'transform':  this.expanded$.map(function(c) { return c ? 'rotate(180deg)' : 'rotate(90deg)'; })
+                'visibility':    'visible',
+                'font-size':     '16px',
+                'float':         'right',
+                'height':        '6px',
+                'padding':       self.expanded$.map(function(c) { return c ? '5px 0px 8px 3px' : '0px 0px 11px 4px'}),
```
olhabn: I don't have a better solution for making `toggle-icon` vertically aligned

src/foam/u2/view/TableView.js
```diff
@@ -10,6 +10,10 @@ foam.CLASS({
   extends: 'foam.u2.view.UnstyledTableView',
 
   css: `
+    ^ {
+      overflow-x: scroll;
```
kgrgreer: Would this be better if it were 'auto'?

src/foam/u2/view/TreeView.js
```diff
@@ -200,37 +207,44 @@ foam.CLASS({
...
-              }
-            }, self.level$, this.selection$, this.data$.dot('id'))).
-            start().
-              addClass(self.myClass('select-level')).
-              style({
-                'width': '100%',
-                'padding-right': '20px'
-              }).
-              addClass(self.myClass('label')).
+              else
```
kgrgreer: No need for 'else'.

src/foam/u2/view/TreeView.js
kgrgreer: That's fine. CSS isn't an elegant tool.

nanopay-arthur: What if we make toggle icon's width equal to its height? And if it's still off we can just add some padding to the bottom?

olhabn: thanks!

src/foam/u2/view/TableView.js
olhabn: fixed

src/foam/u2/view/TreeView.js
olhabn: fixed

PR #4767, opened 2021-02-05
Refactor compound exception

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-3526

## Changes
- Use Throwable.suppressed on CompoundException instead of array property
- Throw underlying exception for single-item compound exception
- Implement ExceptionInterface instead of Exception as it already has implemented getClientRethrowException()

### DIFF
--- a/src/foam/core/CompoundException.js
+++ b/src/foam/core/CompoundException.js
@@ -42,12 +42,6 @@ foam.CLASS({
   properties:  [
     {
       class: 'Object',
-      name: 'exceptions',
-      javaType: 'ArrayList',
-      javaFactory: 'return new ArrayList<RuntimeException>();'
-    },
-    {
-      class: 'Object',
       name: 'sb',
       javaType: 'ThreadLocal',
       javaFactory: `
@@ -69,6 +63,11 @@ foam.CLASS({
 
   methods:  [
     {
+      name: 'getExceptions',
+      type: 'Throwable[]',
+      javaCode: 'return getSuppressed();'
+    },
+    {
       // TODO: cloning this property from ExceptionInterface creates a bug.
       name: 'getClientRethrowException',
       documentation:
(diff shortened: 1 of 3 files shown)


src/foam/core/CompoundException.js
```diff
@@ -68,6 +62,11 @@ foam.CLASS({
   ],
 
   methods:  [
+    {
+      name: 'getExceptions',
```
kgrgreer, line 66: Do we need this? Could we just use getSuppressed directly?

src/foam/core/CompoundException.js
```diff
@@ -98,21 +96,28 @@ foam.CLASS({
     {
       name: 'add',
       args: [{ name: 't', javaType: 'Throwable' }],
-      javaCode: 'getExceptions().add(t);'
+      javaCode: 'addSuppressed(t);'
```
kgrgreer: If it is a CompoundException or just has suppressed exceptions, should we flatten it?

src/foam/core/CompoundException.js
lchanmann: Thought we could preserve the concept of a compound exception having a list exception and the client code could just stay the same eg. `compoundException.getExceptions()`. IMO, getSuppressed() is quite java sided.

src/foam/core/CompoundException.js
lchanmann: Yes, that's better since the nest suppressed exceptions will be surfaced in the log.
Fixed in https://github.com/foam-framework/foam2/pull/4767/commits/ecd62ad8f2e754522ddb7fe90555bf46bb54efe8.

src/foam/core/CompoundException.js
```diff
@@ -98,21 +96,36 @@ foam.CLASS({
     {
       name: 'add',
       args: [{ name: 't', javaType: 'Throwable' }],
-      javaCode: 'getExceptions().add(t);'
+      javaCode: `
+        for ( var e : t.getSuppressed() ) {
+          add(e);
+        }
+
+        if ( t.getSuppressed().length == 0 ) {
```
kgrgreer: Maybe we only want to flatten it if it is a compound exception. Other code may throw suppressing exceptions, in which case we probably want the extra information from the top one. So maybe:
if ( CompoundException ) { add all suppressed } else { add(t) }
Thoughts?

lchanmann: Fixed in https://github.com/foam-framework/foam2/pull/4767/commits/da5134435708559c452af6e1a55ba52b492d49ff. Thanks.

PR #4771, opened 2021-02-07
fixing memento

olhabn:


jlhughes (approved):
Tested, this fixes the issue with multipart keys.

olhabn:
only memento fix. adding multipart key support

src/foam/u2/view/ScrollTableView.js
```diff
@@ -224,7 +224,18 @@
 
         var id = mementoHead;
         if ( of ) {
-          id = of.ID.fromString(mementoHead);
+          if ( ! foam.core.MultiPartID.isInstance(of.ID) ) {
+            id = of.ID.fromString(mementoHead);
+          } else {
```
olhabn, line 229: so probably ugly approach but I don't have other options.
if id is object (for MultiPartID) we will get rid of '{', '}', ':' as if we will let them stay then there is no difference between head and parameters. so head will look like `name="myAlarm",hostname="admins-MacBook-Pro"`

kgrgreer:
I'm merging this as a temporary solution until @olhabn can implement the alternate design which we've discussed.

PR #4777, opened 2021-02-08
Np 3464/file select v1

KolombetM:
@jlhughes

### DIFF
--- a/src/foam/nanos/fs/fileDropZone/FilePreview.js
+++ b/src/foam/nanos/fs/fileDropZone/FilePreview.js
@@ -18,10 +18,10 @@ foam.CLASS({
   ],
 
   methods: [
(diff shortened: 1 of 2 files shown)


src/foam/nanos/fs/fileDropZone/FilePreview.js
```diff
@@ -70,7 +71,12 @@ foam.CLASS({
       if ( ! this.data[pos] ) {
         return;
       }
-      url = URL.createObjectURL(this.data[pos].data.blob);
+      let d = await this.data[pos].data;
```
jlhughes, line 75: Can you add documentation as to what is happening here, thanks.

jlhughes (approved):

PR #4778, opened 2021-02-09
NP-3504 Temporary Fix for Password Validator

carl-zzz:
https://nanopay.atlassian.net/browse/NP-3504

Temporary fix for password length check text. May need further regular expression check or fix for min/maxLength failure text.


src/foam/nanos/auth/ResetPassword.js
```diff
@@ -44,7 +44,17 @@ foam.CLASS({
...
-      minLength: 10
+      validationPredicates: [
+        {
+          args: ['newPassword'],
+          predicateFactory: function(e) {
+            return e.GTE(foam.mlang.StringLength.create({
+              arg1: foam.nanos.auth.ResetPassword.NEW_PASSWORD
+            }), 10);
+          },
+          errorString: 'Passwords must be at least 10 characters.'
```
jlhughes: Can you make these MESSAGEs and add translation for Portuguese as well please.

src/foam/nanos/auth/UpdatePassword.js
```diff
@@ -51,7 +60,17 @@ foam.CLASS({
         class: 'foam.u2.view.PasswordView',
         passwordIcon: true
       },
-      minLength: 10
```
jlhughes, line 54: Don't remove this, let's continue to figure out why it's not working or available.

jlhughes: I would expect this is accessible in validation calls.

jlhughes (approved):

PR #4783, opened 2021-02-09
Np 3557/min max capability data

tharmaman:
Solves the problem where previously exiting a Capable Payload Wizard didn't save the choices you made beforehand.

Also after talking with @KernelDeimos, it made more sense to use MinMaxCapability.data to store the choices that were selected. Aside from the saving/reopening case for Capable Payload, it allows us to also see what Capability  prereqs were granted as a result of direct selection from the user via the GUI.

Hopefully this shouldn't affect any ucjs which currently exist through a min max since those junctions will have already been granted based on the relationship alone and not through reading data. So don't think a migration script is needed as this data  property is primarily to help with the GUI, seeing as they are already past that case then it would not care.

In addition hopefully to @nanoscott's API case where there are no GUIs used.


jlhughes:
LGTM

KernelDeimos (approved):

PR #4784, opened 2021-02-09
[NP-3529] - Empty RegionDAO Fix

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-3529

Description: Even though the order of the regionDAO requests was correct, we noticed that the promise resolution occurred in reverse at times, LIFO, causing the correct choices to be overwritten with []. Adding a sequence number that increments every time onDAOUpdate gets called ensures that any straggler promises that resolve from a stale state get ignored.


kgrgreer:
Brilliant!

PR #4785, opened 2021-02-10
revert changes for dataIsSet

noodlemoodle:
revert change from following 2 prs related to dataIsSet
https://github.com/foam-framework/foam2/pull/4742
https://github.com/foam-framework/foam2/pull/4745

### DIFF
--- a/src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
+++ b/src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
@@ -56,7 +56,7 @@ public class ExpireUserCapabilityJunctionsCron implements ContextAgent {
         activeJunction.setIsInGracePeriod(true);
       }
       if ( activeJunction.getStatus() == CapabilityJunctionStatus.EXPIRED ) {
-        activeJunction.setData(null);
+        activeJunction.getPayload().clearData();
       }
       
       logger.debug("Moved UserCapabilityJunction : " + activeJunction.getId() + " into status :" + activeJunction.getStatus());
--- a/src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
+++ b/src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
@@ -55,7 +55,9 @@ public class ExpireUserCapabilityJunctionsCron implements ContextAgent {
       } else {
         activeJunction.setIsInGracePeriod(true);
       }
-      if ( activeJunction.getStatus() == CapabilityJunctionStatus.EXPIRED ) activeJunction.clearData();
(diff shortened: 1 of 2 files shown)


src/foam/nanos/crunch/UserCapabilityJunctionRefine.js
```diff
@@ -127,8 +127,8 @@ foam.CLASS({
       javaFactory: `
         var payload = new CapabilityJunctionPayload();
         // Temporary so outdated test journals work
-        if ( statusIsSet_ ) payload.setStatus(status_);
-        if ( dataIsSet_ ) payload.setData(data_);
+        if ( statusIsSet_ ) payload.setStatus(getStatus());
+        if ( dataIsSet_ ) payload.setData(getData());
```
jlhughes: what is the point of these two? getData() calls  getPayload().getData() 
or is there something else going on?
data and status are storageTransient, so any old journals with 'data' and 'status' properties will be ignored.

src/foam/nanos/crunch/ExpireUserCapabilityJunctionsCron.java
```diff
@@ -55,7 +55,9 @@ public void execute(X x) {
       } else {
         activeJunction.setIsInGracePeriod(true);
       }
-      if ( activeJunction.getStatus() == CapabilityJunctionStatus.EXPIRED ) activeJunction.clearData();
+      if ( activeJunction.getStatus() == CapabilityJunctionStatus.EXPIRED ) {
+        activeJunction.setData(null);
```
jlhughes: so this would be migrated to 
activeJucntion.getPayload().clearData()
yes?
and any reason you wouldn't do that now?

noodlemoodle: true, will update, thanks

src/foam/nanos/crunch/UserCapabilityJunctionRefine.js
noodlemoodle: need @KernelDeimos  to comment please, this was the code before the fix made here https://github.com/foam-framework/foam2/pull/4745, and since that fix was specifically for recursion caused by checking dataIsSet, I reverted it as well

KernelDeimos: This does cause a recursion issue, so `status_` and `data_` is still better. I'm assuming when outdated UCJs are read from the journal the proxying getter and setter aren't called, otherwise this can be removed. Or, if we don't have any outdated UCJs, this can definitely be removed.

noodlemoodle: reverted revert of fix

jlhughes (approved):
We'll continue to clean this up.

PR #4790, opened 2021-02-10
DNM [NP-3412] update help text ui

jinnnn9:
address: https://nanopay.atlassian.net/browse/NP-3412

Screen Shot 2021-02-10 at 1 19 13 PM


kgrgreer:
Fix padding so scrollbar doesn't extend all the way to top and bottom and so that top and bottom padding are the same on initial screen.

PR #4792, opened 2021-02-10
Add Mixin axiom

KernelDeimos:


src/foam/core/Mixin.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    'flags',
+    'path',
+    ['priority', 200]
+  ],
+
+  methods: [
+    function installInClass(cls) {
+      var m = this.__context__.lookup(this.path);
+      if ( ! m ) throw 'No such interface or trait: ' + this.path;
```
kgrgreer: "No such mixing." or "No such model."

src/foam/core/Mixin.js
```diff
@@ -0,0 +1,53 @@
...
+  `,
+
+  properties: [
+    {
+      name: 'name',
+      getter: function() { return 'mixin_' + this.path; }
+    },
+    'flags',
+    'path',
+    ['priority', 200]
```
kgrgreer, line 23: What is this used for?

src/foam/core/Mixin.js
```diff
@@ -0,0 +1,53 @@
...
+    properties behave as if they were defined on the model itself.
+  `,
+
+  properties: [
+    {
+      name: 'name',
+      getter: function() { return 'mixin_' + this.path; }
+    },
+    'flags',
+    'path',
```
kgrgreer, line 22: What is this used for?

src/foam/core/Mixin.js
KernelDeimos: I saw priority on the implements mixin and I assume it ensures this axiom will override all existing methods and properties

src/foam/core/Mixin.js
KernelDeimos: which one - path? That's the string that's passed into the AxiomArray

src/foam/core/Mixin.js
KernelDeimos: Fixed

PR #4803, opened 2021-02-11
[NP-3464] FileSelect_v1

KolombetM:
add SessionWebAgent back

### DIFF
--- a/src/foam/nanos/blob/FileService.java
+++ b/src/foam/nanos/blob/FileService.java
@@ -55,6 +55,7 @@ public class FileService
 
       File file = (File) fileDAO_.find_(x, id);
 
+      // file service don't check sessionId so only files without owner should be retrived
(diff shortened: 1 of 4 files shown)


jlhughes:
Ok, understand why spid is an issue. 
Do the following:
Add a Relationship with User, so a user had 'files' and a file has an 'owner'. 
Permissions will be needed for the fraud, paymentops groups to read    file.read.*  sort of thing. 
Then file in auth on read can be:  file.read.id  and the owner, system/admin, and fraud ops can read.   
auth on read will not do the auth check if the file.owner is null. 
If the file is reference when not logged in, then only files without an owner can be read. So we can use these for all the terms and conditions and other global or pre-login stuff. 
and the owner can be any type of user - (contact, business). ...

KolombetM:
add Owner check to fileService
remove async from data getter
fix getText method
update filePreview to show text

jlhughes (approved):

PR #4805, opened 2021-02-11
NOC email template + deleteModal translations

minsunny:
nanopay : nanoPayinc/NANOPAY#11878

https://nanopay.atlassian.net/browse/NP-3407
https://nanopay.atlassian.net/browse/NP-3564

added NOC template - NOC email notifications was used without no template, added basic templated to set its body and subject properly
DeleteModal translation

### DIFF
--- a/src/foam/nanos/alarming/AlarmLoggingDAO.js
+++ b/src/foam/nanos/alarming/AlarmLoggingDAO.js
@@ -48,7 +48,6 @@ foam.CLASS({
           if ( alarm.getIsActive() ) {
             Notification notification = new Notification.Builder(x)
               .setTemplate("NOC")
-              .setEmailName("NOC")
               .setToastMessage(alarm.getName())
               .setBody(alarm.getNote())
               .setClusterable(alarm.getClusterable())
--- a/src/foam/nanos/alarming/AlarmLoggingDAO.js
+++ b/src/foam/nanos/alarming/AlarmLoggingDAO.js
@@ -48,6 +48,7 @@ foam.CLASS({
           if ( alarm.getIsActive() ) {
             Notification notification = new Notification.Builder(x)
               .setTemplate("NOC")
+              .setEmailName("NOC")
               .setToastMessage(alarm.getName())
               .setBody(alarm.getNote())
               .setClusterable(alarm.getClusterable())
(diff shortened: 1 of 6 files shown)


src/foam/nanos/notification/EmailSetting.js
```diff
@@ -91,13 +91,12 @@ foam.CLASS({
...
           }
-        }    
+        }
 
         try {
           if ( foam.util.SafetyUtil.isEmpty(notification.getEmailName()) ) {
-            message.setSubject(notification.getTemplate());
             message.setBody(notification.getBody());
-            EmailsUtility.sendEmailFromTemplate(x, null, message, null, null);
+            EmailsUtility.sendEmailFromTemplate(x, null, message, "NOC", null);
```
jlhughes: This is just plain wrong.  If an email template is not provided, then emails are not generated. 
You can generate a warning if you like.. 
so in this first part do not call sendEmailFromTemplate.

src/foam/nanos/notification/EmailSetting.js
```diff
@@ -91,18 +93,19 @@ foam.CLASS({
           if ( config != null ) {
             notification.getEmailArgs().put("link", config.getUrl());
           }
-        }    
+        }
+
+        if ( "NOC".equals(notification.getEmailName()) && ! SafetyUtil.isEmpty(notification.getBody()) ) {
```
minsunny: NOC notification doesnt set an email args for now and some notifications not storing its body after notification translation stuffs updated. So temporarily added the condition if "NOC".equals(notification.getEmailName()) but in the near future when going through this
https://nanopay.atlassian.net/browse/NP-3353, might be considered here.

src/foam/nanos/alarming/AlarmLoggingDAO.js
```diff
@@ -48,6 +48,7 @@ foam.CLASS({
           if ( alarm.getIsActive() ) {
             Notification notification = new Notification.Builder(x)
               .setTemplate("NOC")
+              .setEmailName("NOC")
```
jlhughes: This is on the template.  It's not needed here. that's the whole point of the notification template.

src/foam/nanos/notification/EmailSetting.js
jlhughes: I don't care, simply remove all of this 'NOC' testing on the foam side. 
It's not the foam side's responsibility.

jlhughes (approved):

PR #4809, opened 2021-02-12
[NP-3418] update radioView

jinnnn9:
**address:** https://nanopay.atlassian.net/browse/NP-3418

**summary:** make each choice have the same width

Screen Shot 2021-02-11 at 11 45 30 PM

(an example of how it would look when there are more than 3 choices)

Screen Shot 2021-02-11 at 11 44 36 PM

### DIFF
--- a/src/foam/u2/view/RadioView.js
+++ b/src/foam/u2/view/RadioView.js
@@ -25,9 +25,12 @@ foam.CLASS({
   ],
 
   css: `
-    ^ { padding: 4px 0; }
-    ^ label { position: relative; color: #444; }
-    ^horizontal-radio { display: flex; }
+    ^horizontal-radio { 
+      display: flex;
+      align-items: center;
+      align-content: center;
+      flex-wrap: wrap;
+    }
   `,
 
   properties: [
@@ -76,8 +79,7 @@ foam.CLASS({
 
       this.add(this.choices.map(function(c) {
         return this.E('div').
-          // TODO: why is the radio item getting assigned the same class as the radio whole
-          addClass(this.myClass()).
+          addClass('choice').
           start('input').
             attrs({
               type: 'radio',
--- a/src/foam/u2/view/RadioView.js
+++ b/src/foam/u2/view/RadioView.js
@@ -25,11 +25,29 @@ foam.CLASS({
   ],
 
   css: `
+    ^ { 
+      padding: 4px 0;
+    }
+
     ^horizontal-radio { 
       display: flex;
       align-items: center;
       align-content: center;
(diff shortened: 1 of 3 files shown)


jlhughes:
Looks good.

src/foam/u2/view/RadioView.js
```diff
@@ -76,8 +79,7 @@ foam.CLASS({
 
       this.add(this.choices.map(function(c) {
         return this.E('div').
-          // TODO: why is the radio item getting assigned the same class as the radio whole
-          addClass(this.myClass()).
+          addClass('choice').
```
kgrgreer, line 100: Is this used? Can we just remove this line?

jinnnn9: If you are referring to `addClass('choice')`, then yes we do to style each choice

line 237 in SectionDetailedPropertyView to give 1/3 width:

```
^ .foam-u2-view-RadioView.foam-u2-view-RadioView-horizontal-radio .choice {
      flex-basis: calc(100% / 3);
 }
```

Nauna: agree - the extra spacing seems unnecessary - and ruins our standard ... but the font size change maybe necessary.

jinnnn9: @Nauna hmm, not sure what spacing you are referring to. Could you please elaborate?

Nauna: css for choice =
    ^ .foam-u2-view-RadioView .choice {
      margin-bottom: 16px;
      font-size: 16px;
    }
    
  Im referencing margin-bottom: 16px;
    
  But hmmm - perhaps that isn't the spacing in image...
  
image

it probably is - we just don't want it on the last choice row ... :|

Nauna: Still think it looks good - and approve these changes

kgrgreer: Edit AllViews.js to include a RadioView and see how it looks in a regular DetailView.

jinnnn9: @Nauna that margin-bottom: 16px was added to add space between choices when displayed vertically I assume. I can take that out when the choices are being displayed horizontally, or from the last choice if displayed vertically. Also there is extra spacing from the grid gap.

jinnnn9: @kgrgreer 

radioView look after moving css to RadioView:

Screen Shot 2021-02-12 at 11 47 15 AM

Nauna (approved):

PR #4814, opened 2021-02-15
RulerDAO Clone before After Rule, and possibly update if changes during After Rule.

jlhughes:
Perform delta between result of put from before rule and result of after rule.

NOTE: left log calls in until tested further. 
All NP test cases pass.

### DIFF
--- a/src/foam/nanos/ruler/RulerDAO.js
+++ b/src/foam/nanos/ruler/RulerDAO.js
@@ -23,8 +23,6 @@ foam.CLASS({
     'foam.dao.AbstractSink',
     'foam.dao.ArraySink',
     'foam.dao.DAO',
-    'foam.lib.formatter.FObjectFormatter',
-    'foam.lib.formatter.JSONFObjectFormatter',
     'foam.mlang.order.Desc',
     'foam.mlang.predicate.Predicate',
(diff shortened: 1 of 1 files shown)


lchanmann (approved):
LGTM

src/foam/nanos/ruler/RulerDAO.js
```diff
@@ -127,11 +129,25 @@ if ( oldObj == null ) {
...
+  ret = ret.fclone();
+  FObject before = ret.fclone();
+  if ( oldObj == null ) {
+    applyRules(x, ret, oldObj, (GroupBy) rulesList.get(getCreateAfter()));
+  } else {
+    applyRules(x, ret, oldObj, (GroupBy) rulesList.get(getUpdateAfter()));
+  }
+  // Test for changes during after rule
+  FObjectFormatter formatter = formatter_.get();
+  if ( formatter.maybeOutputDelta(before, ret) ) {
```
kgrgreer: You just want to know if the object has changed for journalling purposes?
If that's the case, then this would be way way faster if you just copied the maybeOutputDelta() method and kept the logic part without the outputting parts.

jlhughes: Replaced with FObject.diff.

PR #4815, opened 2021-02-15
add isCLassPredicate

smirnovaka:


src/foam/nanos/ruler/predicate/IsClassPredicate.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2021 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.nanos.ruler.predicate',
+  name: 'IsClassPredicate',
```
jlhughes, line 9: I can see we really screwed up on naming. 
All of the isInstance, isClass should have been named 
classOfPredicate, instanceOfPredicate.  
Ok, too late now.

smirnovaka: yes.. should I rename this one?

PR #4818, opened 2021-02-16
[NP-3559] Medusa Threading Cleanup, Primary serve last MDAO until put completes.

jlhughes:
Have Primary serve last MDAO state for 'find' and 'select' while 'put' operation in-flight.


src/foam/nanos/medusa/ElectoralServiceServer.js
```diff
@@ -263,6 +265,7 @@ foam.CLASS({
         }
       }
       getLogger().debug("execute", "exit", "state", getState().getLabel(), "election time", getElectionTime());
+      Thread.currentThread().setName(savedThreadName);
```
kgrgreer: Does this need to be in a finally?

kgrgreer:
I would like to see PM's added to things like:
Thread.currentThread().sleep(sleep);

Since relatively long sleeps can negatively affect performance, and it would be good to have an idea of how often they're happening.

PR #4820, opened 2021-02-16
[NP-3612] LogMessage DateTime TableView Filtering

jlhughes:
Repurpose 'created' as a DateTime property to provide date range queries on TableViews. 
To avoid creating a date on each log message creation, the 'created' property is only calculated on select, which, hitherto, has only been for admin log message viewing.


src/foam/nanos/logger/LogMessageDAO.js
```diff
@@ -40,6 +40,7 @@ foam.CLASS({
       javaCode: `
       LogMessage lm = (LogMessage) obj;
       lm.setHostname(getHostname());
+      lm.setLastModified(new java.util.Date());
```
kgrgreer: Do we never load from the DAO on restart?

src/foam/nanos/logger/LogMessage.js
```diff
@@ -125,5 +125,13 @@ foam.CLASS({
     //   view: { class: 'foam.u2.view.PreView' },
     //   updateVisibility: 'RO'
     // }
+    {
+      documentation: 'Support searching by date range.  Created is a human readable date string',
+      name: 'lastModified',
```
kgrgreer: So this is the same as the 'created' field, just as a Date instead of a String? Should document this point so it isn't confusing to people.

src/foam/nanos/logger/LogMessageDAO.js
jlhughes: Nope, the ID is also storageTransient. 
It's backed by WriteOnly and FixedSize DAOs

src/foam/nanos/logger/LogMessage.js
jlhughes: Um, that's what I thought said in the documentation.  I'll try to elaborate further.

kgrgreer: I see. Because this has a label: 'Created', I thought when it said "Created" in the docs it was talking about itself.
Having said: The "created" property is the same but in a human readable string format. would have been clearer to me.

src/foam/java/refinements.js
```diff
@@ -333,7 +338,8 @@ foam.CLASS({
         toCSV:                   this.javaToCSV,
         toCSVLabel:              this.javaToCSVLabel,
         fromCSVLabelMapping:     this.javaFromCSVLabelMapping,
-        sheetsOutput:            this.sheetsOutput
+        formatJSON:             this.javaFormatJSON,
```
kgrgreer: fix indentation

jlhughes:
```suggestion
        formatJSON:              this.javaFormatJSON,
```

PR #4822, opened 2021-02-16
DO  NOT MERGE: Reference summary on approvals

tharmaman:
Added a reference summary to help the user get information about the object they are approving without having to click view Reference so that they could filter requests out more easily with the table and have an idea of what they are looking at from the getgo.


nanoscott:
@tharmaman .. can you add a toSummary method to UCJs which proxies to the toSummary of the sourceId or else this PR won't really help for any of the UCJ approvals.

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -138,6 +138,11 @@
   ],
 
   properties: [
+    {
```
kgrgreer, line 141: Could this be transient and be calculated by an expression on objId and daoKey?

tharmaman: In the case of something like UCJ, the information wouldn't be enough. If that's the case then we would have to make something specific for UCJ where we want the sourceId of the UCJ and also query that to get the object we want as the referenceSummary and call toSummary() on it.

In terms of transient, I  believe it's better to store it because the back-end knows the true intent of what the underlying request is about since the various rules are what instantiate the requests in the first place. If we rely on having the front-end just attempt to calculate what the intent for reference summary is then I think that will restrict the flexibility of referenceSummary.

I'm not opposed to making default behaviour work with objId and daoKey however.

I think saying it's a direct summary of the View Reference doesn't really work with the UCJ cases.

tharmaman:
> @tharmaman .. can you add a toSummary method to UCJs which proxies to the toSummary of the sourceId or else this PR won't really help for any of the UCJ approvals.

Yeah so RE: np side, I had to manually call  .toSummary on the user/business for the rules where these approvals are instantiated. I think proxying the toSummary of UCJ will only work if moving forward we always want ucj.toSummary to proxy only to the sourceId and not actually be a toSummary of the UCJ itself with the targetId included @nanoscott

src/foam/nanos/approval/ApprovalRequest.js
tharmaman: I guess thinking of it as a second bit of classification data in addition to  the actual ApprovalRequest.classification property might be clearer.

tharmaman: Think if we go down that road, then ComplianceApprovalRequest.referenceSummary will override the expression to call .toSummary() on whatever the sourceId object is.

tharmaman: Yeah now I'm thinking and looking over all the rules I had to manually add the setting of referenceSummary upon approval request instantiation within their respective rules in NP side. The ComplianceApprovalRequests.referenceSummary override will work to handle the UCJ and we could make this transient where the browser can calculate what to show based on grabbing the object with objId and daoKey and calling toSummary on it. The referenceSummary should in the end always be directly connected to whatever its showing in viewReference otherwise something  is wrong so this is a good check to have.

PR #4823, opened 2021-02-17
[NP-3587] CIDR Addresses not calculating in js

jlhughes:
'expression' on notation is not being triggered.

### DIFF
--- a/src/foam/net/CIDR.js
+++ b/src/foam/net/CIDR.js
@@ -32,16 +32,24 @@ List entries are of the form: 172.0.0.0/24 - this would restrict logins to 172.x
       class: 'String',
       placeholder: '10.0.0.0/24',
       required: true,
-      onKey: false,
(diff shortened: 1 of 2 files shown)


src/foam/net/CIDR.js
```diff
@@ -8,51 +8,68 @@ foam.CLASS({
...
+      required: true,
+      onKey: false,
+      validateObj: function(notation) {
+        if ( ! notation ||
+             notation.length < this.NOTATION_MIN_LENGTH ||
+             notation.length > this.NOTATION_MAX_LENGTH ) {
+          return 'invalid notation';
+        }
+        var m = notation.match(/\d+/g); // [ '198', '162', '1', '1', '24' ]
+        if ( ! m || m.length < 5 ) {
```
kgrgreer: Could you just say != 5?

PR #4825, opened 2021-02-17
fix Assertion failed: Cannot stub non-void non-async

hchoura:
Those methods are used just in java.
Add java flags


kgrgreer:
I don't think we should be loading this code on the client, so then this check wouldn't be needed. Can we instead find out how to just not load this code on the client?

PR #4827, opened 2021-02-18
CachingAuthService - if subject.user null return empty cache.

jlhughes:
service/threads requires permission service.run.threads and prompts for login.
On admin login, access is denied. 
AuthWebAgent calls auth.check with session.getContext() which does not have the 'admin' user.
During troubleshooting - walk the auth proxy chain calling check on each auth service. The CachingAuthService fails with NPE on dereference of Subject.user, which is null.    If an empty cache is returned when user is null, then the admin access to service/threads succeeds.

### DIFF
--- a/src/foam/nanos/auth/CachingAuthService.java
+++ b/src/foam/nanos/auth/CachingAuthService.java
@@ -64,6 +64,10 @@ public class CachingAuthService
     Subject subject = (Subject) x.get("subject");
     User    user    = subject.getUser();
 
+    if ( user == null ) {
+      return new ConcurrentHashMap<String,Boolean>();
+    }
+
     // If the user in the context does not match the session user, do not use the
     // cache permission map, which belong to session user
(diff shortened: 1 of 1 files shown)


lchanmann:
AuthWebAgent calls authenticate before the auth check. The session and its context should have been setup. A null value on Subject.user shouldn't have gotten that far into the CachingAuthService, the earlier decorator in the chain should handle the case.

kgrgreer:
You don't need to always login. Consider a public website or a website like Amazon which lets you browse before logging in to purchase.

lchanmann:
If we don't require login to access service/threads, shouldn't the service be un-authenticated then it won't use AuthWebAgent.

https://github.com/foam-framework/foam2/blob/7c708de6aa46c679b4326ef70bc940e240c87996/src/foam/nanos/http/NanoRouter.java#L189-L191

PR #4830, opened 2021-02-18
Better Document Updates for Wizard

KernelDeimos:


src/foam/u2/wizard/WizardletAware.js
```diff
@@ -0,0 +1,19 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.INTERFACE({
+  package: 'foam.u2.wizard',
+  name: 'WizardletAware',
+
+  documentation: `
+    Wizardlet data implementing this interface can override the default
+    method of detecting changes.
```
kgrgreer: Add why they would want to do that.

PR #4831, opened 2021-02-18
Fix errors in UCJ edit view

KernelDeimos:


src/foam/nanos/crunch/CapabilityJunctionPayload.js
```diff
@@ -93,7 +93,9 @@ foam.CLASS({
     {
       name: 'toSummary',
       code: function(){
-        return this.capabilityDAO.find(this.capability).then(capability => `${capability.name}`);
+        return this.capability
+          ? this.capabilityDAO.find(this.capability).then(capability => `${capability.name}`)
```
kgrgreer: Why `${capability.name}` and not just capability.name?

KernelDeimos: Yes, it looks like there was something else there before: https://github.com/foam-framework/foam2/commit/34d7a0bbdafec31d273cfe653aae54dbfd58343b#diff-56b110a0a0becb1125deabfbecc542d8d9a72e603af43db82d114e1dd3737c27L121

Those properties never belonged on the payload so they were removed. I'll remove the backticks now that they aren't needed

PR #4832, opened 2021-02-18
Updating order and gridColumns for approval requests

nanoscott:


tharmaman (commented):
Think its a bit hard to track what actually changed because I think you reordered the properties themselves too? In addition to adding order, gridColumns and some sections to props?

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -64,19 +64,12 @@
 
   sections: [
     {
-      name: 'requestDetails',
-      title: 'Approval Request Information',
-      order: 0
-    },
-    {
-      name: 'complianceInformation',
```
tharmaman, line 72: Need to update this section reenaming for properties on ComplianceApprovalRequest np side I believe btw

tharmaman (changes requested):

src/foam/nanos/approval/ApprovalRequest.js
nanoscott: I'm have a PR ready for that.

PR #4833, opened 2021-02-18
Transient ReferenceSummary on ApprovalRequest

tharmaman:
This property is solely used on the client, therefore just needs to use a tableCellFormatter to provide the preview of the viewReference.


src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -138,6 +138,25 @@
   ],
 
   properties: [
+    {
+      class: 'String',
+      name: 'referenceSummary',
+      section: 'requestDetails',
```
nanoscott: can you update the section here to be 'approvalRequestInformation' and set the order to be 21 and the gridColumns to be 6

src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -138,6 +138,25 @@
...
+      transient: true,
+      tableWidth: 200,
+      tableCellFormatter: function(_,obj) {
```
nanoscott, line 133: the detailed view also has to be set, not only the table cell

nanoscott (changes requested):

nanoscott (approved):

src/foam/nanos/approval/ApprovalRequest.js
nanoscott: we can do this in a followup

src/foam/nanos/approval/ApprovalRequest.js
tharmaman: Done.

src/foam/nanos/approval/ApprovalRequest.js
tharmaman: Done.

PR #4835, opened 2021-02-18
Fixing google api credentials id 02182021

olhabn:
using domain name as id

### DIFF
--- a/src/foam/nanos/google/api/auth/GoogleApiAuthService.js
+++ b/src/foam/nanos/google/api/auth/GoogleApiAuthService.js
@@ -54,7 +54,9 @@ foam.CLASS({
         'java.io.IOException'
       ],
       javaCode: `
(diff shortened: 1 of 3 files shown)


src/foam/nanos/google/api/auth/GoogleApiAuthService.js
```diff
@@ -54,7 +55,9 @@ foam.CLASS({
         'java.io.IOException'
       ],
       javaCode: `
-        GoogleApiCredentials credentialsConfig = (GoogleApiCredentials) ((DAO)getX().get("googleApiCredentialsDAO")).find(((AppConfig)x.get("appConfig")).getUrl());
+        HttpServletRequest req = (HttpServletRequest)x.get(HttpServletRequest.class);
+        GoogleApiCredentials credentialsConfig = (GoogleApiCredentials) ((DAO)getX().get("googleApiCredentialsDAO")).find(req.getServerName());
```
jlhughes, line 59: so serverName is just 'localhost' rather than http://localhost:8080, for example. This is what you want?

olhabn: yep

jlhughes (approved):

PR #4839, opened 2021-02-19
optimize the ClientBuilder

hchoura:
**Compare in Chrome**
Chrome

**Compare after restart**
after restart

**Compare in FireFox** 
firefox

### DIFF
--- a/src/foam/nanos/client/ClientBuilder.js
+++ b/src/foam/nanos/client/ClientBuilder.js
@@ -92,49 +92,54 @@ foam.CLASS({
             }
           });
 
-          self.nSpecDAO.where(self.EQ(self.NSpec.SERVE, true)).select({
-            put: function(spec) {
-              if ( spec.client ) {
-                client.exports.push(spec.name);
+          self.nSpecDAO.where(self.EQ(self.NSpec.SERVE, true)).select(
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

src/foam/nanos/client/ClientBuilder.js
```diff
@@ -92,49 +92,54 @@ foam.CLASS({
...
           });
 
-          self.nSpecDAO.where(self.EQ(self.NSpec.SERVE, true)).select({
-            put: function(spec) {
-              if ( spec.client ) {
-                client.exports.push(spec.name);
+          self.nSpecDAO.where(self.EQ(self.NSpec.SERVE, true)).select(
+            foam.mlang.Expressions.create().PROJECTION(foam.nanos.boot.NSpec.NAME, foam.nanos.boot.NSpec.CLIENT))
+            .then(p => {
+              foam.dao.ArrayDAO.create({array: p.array})
```
kgrgreer, line 98: Why not just loop over the array? Why bother creating an ArrayDAO?

kgrgreer: Maybe move a lot of this code to a helper method to avoid so much nesting.

PR #4843, opened 2021-02-21
[NP-3666] Medusa Node Assignment Refactor

jlhughes:
Refactor bucket assignment and selection to account for AWS Availability Zones (geographical regions / DRs).  In AWS Mediators and Nodes are assigned to an Availability Zone and with auto bucket assignment, it's possible that all nodes receiving a Medusa Entry are in the same Availability Zone resulting in zero redundancy. 
Bucket assignment can be explicitly set in ClusterConfig, such that buckets span Availability Zones. 
Previously a Medusa Entry was written to one node in each bucket.  Now a Medusa Entry is written to all nodes of one bucket.


src/foam/nanos/medusa/MedusaBroadcast2NodesDAO.js
```diff
@@ -57,26 +57,19 @@ foam.CLASS({
 
   methods: [
     {
-      documentation: `
-  - mod by node groups to get node bucket
-  - create job for each client in bucket
-`,
+      documentation: `Distribute entry once to each bucket. Mod of entry.id and bucket.size selects the Node in the bucket to receive the entry.`,
```
jlhughes:
```suggestion
      documentation: `Distribute entry to each node in one bucket. Mod of entry.id and bucket.size selects the bucket to receive the entry.`,
```

PR #4846, opened 2021-02-22
[NP-3680] - ucjDAO fileDAO Sync

nanopay-arthur:
Ref: https://nanopay.atlassian.net/browse/NP-3680

### DIFF
--- a/src/foam/nanos/crunch/ClientCrunchService.js
+++ b/src/foam/nanos/crunch/ClientCrunchService.js
@@ -9,22 +9,22 @@ foam.CLASS({
   name: 'ClientCrunchService',
 
   implements: [
-    'foam.nanos.crunch.CrunchService',
-    'foam.mlang.Expressions'
+    'foam.mlang.Expressions',
+    'foam.nanos.crunch.CrunchService'
   ],
 
   requires: [
     'foam.nanos.crunch.AgentCapabilityJunction',
-    'foam.nanos.crunch.UserCapabilityJunction',
     'foam.nanos.crunch.AssociatedEntity',
-    'foam.nanos.crunch.CapabilityJunctionStatus'
+    'foam.nanos.crunch.CapabilityJunctionStatus',
+    'foam.nanos.crunch.UserCapabilityJunction'
   ],
 
   imports: [
     'capabilityDAO',
-    'userCapabilityJunctionDAO',
     'logger',
-    'subject'
+    'subject',
+    'userCapabilityJunctionDAO'
   ],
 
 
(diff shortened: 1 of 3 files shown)


src/foam/nanos/crunch/ClientCrunchService.js
```diff
@@ -9,15 +9,138 @@ foam.CLASS({
...
+          this.logger.error("getJunction", capabilityId, e);
+
+          // On failure, report that the capability is available
+          var ucj = await this.buildAssociatedUCJ(x, capabilityId, this.subject);
+          return ucj;
+        }
+      }
+    },
+    {
+      name: 'updateJunction',
```
kgrgreer, line 37: Isn't this the only method we need? What are the others for?

KolombetM (approved):

src/foam/nanos/crunch/ClientCrunchService.js
nanopay-arthur: I assumed we needed to define the other CrunchService methods updateJunction calls within itself, but now that you mention it, I don't see why client's updateJunction method can't just call ServerCrunchService's getJunction method. Will update.

src/foam/nanos/fs/FileArrayDAODecorator.js
```diff
@@ -25,42 +25,51 @@ foam.CLASS({
...
-      });
+    async function processFiles(obj) {
+      var props1 = obj.cls_.getAxiomsByClass(foam.nanos.fs.FileArray);
+      var values = await Promise.all(props1.map(prop => Promise.all(prop.f(obj).map(f => this.processFile(f)))));
+      values.forEach(f => f.forEach(f2 => {
+        f2.dataString = undefined;
+        f2.data = undefined;
+      }));
+      props1.forEach((prop, i) => prop.set(obj, values[i]));
+      var props2 = obj.cls_.getAxiomsByClass(foam.core.FObjectProperty).filter((p) => ! foam.dao.DAOProperty.isInstance(p));
```
jlhughes, line 54: Question: is a FileArray also a FObjectProperty? Curious if obj FileArray properties are processed twice?

nanopay-arthur: Yes, currently a nested FileArray would be processed twice... We can add another condition to the filter of props2 to be ! props1.includes(p).

nanopay-arthur: actually, I don't think my suggestion makes sense...

nanopay-arthur: actually, it does.

PR #4850, opened 2021-02-23
update EmailTemplate.body.view to see html as well as txt

Nauna:


src/foam/nanos/notification/email/EmailTemplate.js
```diff
@@ -48,7 +48,13 @@ foam.CLASS({
       class: 'String',
       name: 'body',
       documentation: 'Template body',
-      view: { class: 'foam.u2.tag.TextArea', rows: 40, cols: 150 }
+      view: {
+        class: 'foam.u2.MultiView',
+        views: [
+          { class: 'foam.u2.HTMLView' },
+          { class: 'foam.u2.tag.TextArea', rows: 40, cols: 130 }
```
Nauna, line 55:
```suggestion
          { class: 'foam.u2.tag.TextArea', rows: 40 }
```

Jin caught this:
just one thing, I don’t think we want to manually set cols for textarea
it will oveflow in detailview if the screen is too small

Nauna:
Jin is working on a better fix to catch the issue with the cols overflow on the textArea

PR #4853, opened 2021-02-23
toSummaryPromise in CitationView

blakegreen32:
Calling toSummaryPromise method if it exists in data of CitationView

### DIFF
--- a/src/foam/u2/CitationView.js
+++ b/src/foam/u2/CitationView.js
@@ -34,7 +34,17 @@ foam.CLASS({
       name: 'updateSummary',
       isFramed: true,
       code: function() {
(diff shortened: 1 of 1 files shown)


nanoscott (approved):

src/foam/u2/CitationView.js
```diff
@@ -34,7 +34,17 @@ foam.CLASS({
       name: 'updateSummary',
       isFramed: true,
       code: function() {
-        this.summary = this.data ? this.data.toSummary() : undefined;
+        if ( this.data ) {
```
jlhughes: what if toSummary is async and we await in it, so each model can decide what to do and we don't have two methods and in the case of transactions - they do completely different things.

PR #4861, opened 2021-02-24
Crunch Lab Dynamic Capability List for UCJs

zahnno:
Capability list is now filtered by capabilities referenced by the chosen users, capable of viewing all capabilities as well, retaining existing behaviour.

Updated UI
Screen Shot 2021-02-26 at 12 37 24 PM

^ Simply applied new detail property views to existing propertyInfo's displayed, did not bother with model driven gui. Also applied page title, labels and help messages which are supported by the SectionDetailPropertyViews applied.


src/foam/u2/crunch/lab/CrunchLab.js
```diff
@@ -45,41 +47,56 @@ foam.CLASS({
...
+      class: 'foam.dao.DAOProperty',
+      name: 'filteredCapabilityDAO',
+      expression: function(crunchUser) {
+        if ( crunchUser == 0 ) return this.capabilityDAO;
+        return this.PromisedDAO.create({
+          of: 'foam.nanos.crunch.Capability',
+          promise: this.userCapabilityJunctionDAO.where(this.EQ(this.UserCapabilityJunction.SOURCE_ID, crunchUser))
+            .select(this.MAP(this.UserCapabilityJunction.TARGET_ID))
+            .then((sink) => {
+              let capabilities = sink.delegate.array ? sink.delegate.array : [];
```
KernelDeimos: When does this case happen?

zahnno: I'm not completely sure, but I've encounter cases.

zahnno: ~~Not too sure why that case happens but I remember encountering it when similar logic was applied on the BankAccount currencies.~~

Edit: After taking a look at the Map expression it seems like this would happen if there are no objects in the serviceDAO which the select is being called on.

Which would be unlikely in this case - regardless the map should handle this.

Nauna:
Nadir mentioned this too - but I couldn't think of a way to get the ucj list relevant to a user. At least not yet. Its not as simple as filtering for ucj's the user already has.

For example consider the below image
image

In this case I would not be able to see the root capability for this user - to see where they are in the flow.
Therefore this fix really limits the use of crunch lab and I would NOT suggest the change

Nauna:
I have a thought - what if you queried the capabilityDAO via the user context - then the availabilityPredicate would be used to filter the capabilities.
The one thing we might also have to do is ensure the philosophy that all user relevant capabilities are specifically available to them and we use the visibilityPredicate to create the user flow.
I say this because we do mostly do that but have some oddities. The businessRegistration was/is an example. It was/is available to only users without a business - but that wouldn't be what we want here. The businessRegistration capability is relevant to the business users ... especially if they want to update their address or phone number.
Another example oddity is the use of 'precondition' will have to ensure this doesn't effect the availability of a capability - don't believe it does, but not 100%.

zahnno:
> Nadir mentioned this too - but I couldn't think of a way to get the ucj list relevant to a user. At least not yet. Its not as simple as filtering for ucj's the user already has.

The UCJ graph list isn't being filtered here, which I believe is what you're assuming, correct me if I'm wrong. The following will assume this assumption since that is how I interpreted your comment.

The selected users associated UCJs are being used to filter the **Capability** list that is chosen and displayed as the root, **nothing about the UCJ/Capability graph is being omitted**. The capabilities that are shown in the graph have no relation to the capabilities displayed in the **capability dropdown list**.

In the image you posted, the international payments capability would be available since the mlang used in the filter checks for the sourceId of the selected user UCJs - In the example you posted, if the business Buz One was selected, then the international payment capability would still be shown, since it would have a UCJ that has Buz One id as its source Id and International payments as its target Id.

Just a friendly suggestion: Please explain in detail an image when you post it. "in this case your fix doesn't work." was a bit ambiguous and caused me to assume your concern.

src/foam/u2/crunch/lab/CrunchLab.js
```diff
@@ -82,32 +103,80 @@ foam.CLASS({
...
+          sections: [
+            {
+              heading: 'Capabilities',
+              dao$: X.data.slot(function(showAllCapabilities, associatedUser, crunchUser) {
+                if ( crunchUser == 0 || showAllCapabilities) return this.capabilityDAO;
+                let predicate = associatedUser ?
+                    this.AND(
+                      this.EQ(this.AgentCapabilityJunction.SOURCE_ID, crunchUser),
+                      this.EQ(this.AgentCapabilityJunction.EFFECTIVE_USER, associatedUser)
+                    ) :
```
KernelDeimos:
```suggestion
                    this.OR(
                      this.AND(
                        this.EQ(this.AgentCapabilityJunction.SOURCE_ID, crunchUser),
                        this.EQ(this.AgentCapabilityJunction.EFFECTIVE_USER, associatedUser)
                      ),
                      this.EQ(this.UserCapabilityJunction.SOURCE_ID,associatedUser)
                    ) :
```
```suggestion
                    this.AND(
                      this.EQ(this.AgentCapabilityJunction.SOURCE_ID, crunchUser),
                      this.EQ(this.AgentCapabilityJunction.EFFECTIVE_USER, associatedUser)
                    ) :
```

This will allow better visibility of capabilities owned by the associated user

KernelDeimos (changes requested):
This works great! I commented a small tweak to make the results better. ~I don't see the signing officer question capability... is that no longer a prerequisite?~ [redacted: was looking in the wrong place]

KernelDeimos:
Just found an issue - associated user doesn't update DAGView because the slot isn't listening. I'm testing a fix now and I'll PR it to your fork branch if it works.

src/foam/u2/crunch/lab/CrunchLab.js
KernelDeimos: actually ignore this commit suggestion - I have a better idea

zahnno: This conversation was resolved on another communication channel. 

The suggested predicate above provides all combinations of capabilities that relate to the users chosen. Which includes more capabilities in the root capability list than desired. Suggestion was dismissed.

zahnno:
Updated predicate to populate additional ucj data on graph.

~~When https://github.com/foam-framework/foam2/pull/4879 is merged these [lines](https://github.com/foam-framework/foam2/pull/4861/files#diff-d958cf2127a2a6525d1c3a1027ac6d071f20a00b8085b413e19dbac7a4c7fd11R212-R215) can be commented out into code, replacing the predicate below them.~~

^ Disregarding suggestion to use client side crunchService method **getJunctionFor** in https://github.com/foam-framework/foam2/pull/4879. The call is more taxing comparing to the predicate currently placed due to multiple calls required on client for each capabilityId associated to root and logic getJunctionForSubject builds several predicates for each prerequisite capability associated to the selected root capability, also creates ucj/acjs if none are found. Although these ucjs are also created under the correct subject, the permissions associated to the context passed in, are not. @KernelDeimos @nanopay-arthur

KernelDeimos:
> Updated predicate to populate additional ucj data on graph.
> 
> ~When #4879 is merged these [lines](https://github.com/foam-framework/foam2/pull/4861/files#diff-d958cf2127a2a6525d1c3a1027ac6d071f20a00b8085b413e19dbac7a4c7fd11R212-R215) can be commented out into code, replacing the predicate below them.~
> 
> ^ Disregarding suggestion to use client side crunchService method **getJunctionFor** in #4879. The call is more taxing comparing to the predicate currently placed due to the logic in getJunctionForSubject which builds several predicates for each prerequisite capability associated to the selected root capability and also creates ucj/acjs if none are found. Although these ucjs are also created under the correct subject, the permissions associated to the context passed in, are not. @KernelDeimos @nanopay-arthur

Thanks for bringing this up. ClientCrunchService should cache the predicate instead of re-constructing it, so this is an issue with ClientCrunchService's implementation rather than it's use here. I'll ask Arthur if he's willing, otherwise I can take on that optimization.

KernelDeimos:
> also creates ucj/acjs if none are found

It will return a UCJ with status `AVAILABLE` if there's nothing in the DAO, but it doesn't store anything. This is done to avoid having a distinction between null and available UCJs, which could otherwise create subtle edge-cases. CrunchLab will then display the status as available for these UCJs, which is the correct behaviour.

zahnno:
@KernelDeimos I understand what it's doing, I read the code. I don't understand what purpose it serves, and believe it doesn't belong here.

Now I'm looking specifically at the GraphView when I say that. I am not aware of the additional use cases for the **getJunctionFor** method, however for this use case the graph plots the ucj data based on the capability targetIds that have already been previously plotted prior to the call. There is no requirement afaik for every capability to reference ucj data, aside from the logic within the CapabilityGraphNodes which rely on saved UCJ data ( the on click to ucj data requires an id reference ) which needs to be fixed regardless.

My point is this **getJunctionFor** is far more taxing than the predicate provided in the current suggested code.

KernelDeimos:
> I don't understand what purpose it serves, and believe it doesn't belong here.

This is the reason:
> This is done to avoid having a distinction between null and available UCJs

PR #4864, opened 2021-02-25
Add authentication to DIG and SUGAR

jlhughes:
* Authenticate DIG and SUGAR with the SessionWebAgent.
* In DIG and SUGAR add a PrintWriter to the context if not provided.
* In SessionWebAgent add the HttpParameters to the Session context passed to delegate WebAgents.  The HttpParameters is executed by the servlet before SessionWebAgent, but SessionWebAgent builds its own context to pass to the delegates, but the delegates WebAgents are expecting a HttpParameters in the context.

Tested with Medusa; prompted for login on service/dig and get appropriate output thereafter.


minsunny:
LGTM

minsunny (approved):

PR #4865, opened 2021-02-25
Np 3651/fixing scroll

olhabn:


src/foam/u2/view/ScrollTableView.js
```diff
@@ -296,7 +299,8 @@
           var scroll = this.memento.paramsObj.r * this.rowHeight;
           scroll = scroll >= this.rowHeight && scroll < this.scrollHeight ? scroll : 0;
 
-          document.getElementById(this.id).scrollTop = scroll;
+          if ( this.childNodes && this.childNodes.length > 0 )
```
olhabn, line 302: so, for some reason `this.id` is id of DAOBrowserView not of ScrollTableView

PR #4872, opened 2021-02-25
Added toSummary promise handling in referenceSummary view + tcf

tharmaman:


src/foam/nanos/approval/ApprovalRequest.js
```diff
@@ -135,25 +135,33 @@
         this.__subSubContext__[obj.daoKey].find(obj.objId).then(requestObj => {
           let referenceSummaryString = `ID:${obj.objId}`;
 
-          if ( requestObj && requestObj.toSummary() ){
-            referenceSummaryString = requestObj.toSummary()
-          }
+          if ( requestObj ){
+            Promise.resolve(requestObj.toSummary()).then(function(requestObjSummary) {
```
kgrgreer, line 140: If we make toSummary() able to return Promises, then we need to change everywhere in the system that we use toSummary(), not just here. FYI: @jlhughes

tharmaman: Think only place so far where it returns a promise is for `Transaction.js`, let me just double check if we actually need that Promise.resolve()

tharmaman: Confirmed `Transaction.toSummary()` returns a promise,

**TableCellFormatter** seems to magically handle the `Transaction.toSummary(`) and render its resolved Promise without having to directly resolve it, I’m guessing `Element.add` can take promises

But yeah for **View**, we just set a simple slot to `Transaction.toSummary()`, so its setting the slot as the actual Promise and it doesn’t render anything if you don’t resolve it.

tharmaman: Left TODO for ReferenceView

PR #4879, opened 2021-02-26
Add getJunctionFor Method to CrunchService

nanopay-arthur:


src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -193,6 +193,18 @@ public List getCapabilityPath(X x, String rootId, boolean filterGrantedUCJ) {
...
   }
+  
+  public UserCapabilityJunction getJunctionFor(X x, String capabilityId, User effectiveUser, User user) {
+    Subject subject = (Subject) x.get("subject");
+    foam.nanos.auth.AuthService auth = (foam.nanos.auth.AuthService) x.get("auth");
+    if ( auth.check(x, "service.crunchService.customSubject") ) {
+      var requestedSubject = new Subject();
+      requestedSubject.setUser(user);
+      requestedSubject.setUser(effectiveUser);
+      return this.getJunctionForSubject(x, capabilityId, requestedSubject);
```
zahnno, line 204: Context still holds requestors permissions and getJunctionForSubject also builds missing ucj's, capabilityDAO would be authorized for whoever is making the request not the subject users. This can lead to issues where ucj's are created when they shouldn't be.

@KernelDeimos Why is a **get** method creating capabilities junctions?

KernelDeimos: > Why is a get method creating capabilities junctions?

Where are you seeing it do this? It constructs an object to return if there's nothing in the DAO already, but it doesn't save any new data - it is still a get method.

nanopay-arthur: @zahnno Not sure I follow the issue with the context holding the requestor's permissions. We only overwrite the subject in the case that an admin is calling the method, otherwise its functionality is the same as getJunction.

zahnno: > It constructs an object to return if there's nothing in the DAO already

Right, construct/create and not saved, sure. What use case does this satisfy in the crunch system? @KernelDeimos 

> if ( auth.check(x, "service.crunchService.customSubject") ) {

This is **_not_** admin specific. Nothing hard coded into our system should be admin specific. This is a permission which can be provided to anyone as it should be. Similar logic dictate capability availability to a user/context.

[ Here's the code ](https://github.com/foam-framework/foam2/blob/9938c974676954fe5fccbedfa9350e9a6c7c80f2/src/foam/nanos/crunch/ServerCrunchService.java#L361-L395) where the issue occurs and where the intention or use case must be defined. There's a comment in the buildAssociatedUCJ that details intended behaviour of the call stating
``` If the subject in context doesn't have the capability available, we should act as though it doesn't exist; this is why inX is here.``` leading me to believe there is a reason to take heed of permissions.

This leads me back to my first question to Eric and the intention behind creating ucjs (... and not saving them)

KernelDeimos: `service.crunchService.customSubject` should be renamed `service.crunchService.getJunctionFor` to reflect that it's a permission any user can have to be able to call this method.

@zahnno one important point you made is that the context doesn't necessarily reflect the requested subject's capability availability. Since a user in CRUNCH lab would like to see as much of the tree as is available to them, I think this behaviour is good. An admin should see all relevant capabilities even if the subject can't. Additionally, an underprivileged user who happens to have permission to use this method should not be able to see additional capabilities just because the subject has them.

KernelDeimos: @nanopay-arthur the `auth.check` here isn't needed - this is already handled by userCapabilityJunctionDAO's checkOwnership rule. As a general rule, `auth.check` calls from the client should only be used for visual cues related to permissions rather than behaviour like this.

KernelDeimos: edit: was wrong about this - thought `getJunctionFor` was somewhere else

PR #4896, opened 2021-03-02
Np-3490 common passwords

xuerongNanopay:
ticket: https://nanopay.atlassian.net/browse/NP-3490
1. make dao for common words
2. make passwordPolicy use this common words dao
3. set the passwordPolicy to be used on the ui - set into password property validations
4. move passwordPolicy into theme and not on group => then will be able to use in Signup view when group is not set yet.


src/foam/nanos/auth/CommonPassword.js
```diff
@@ -0,0 +1,25 @@
...
+      name: 'id'
+    }
+  ],
+  methods: [
+    {
+      name: 'toString',
+      code: function() {
+        return this.id
+      }
+    },
```
kgrgreer: No trailing commas.

src/foam/nanos/auth/CommonPassword.js
```diff
@@ -0,0 +1,25 @@
...
+  ]
+})
```
kgrgreer, line 25: Missing ;

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -70,6 +82,10 @@ foam.CLASS({
           javaType: 'String'
         }
       ],
+      code: function(user, potentialPassword) {
+        let password = potentialPassword.toLowerCase();
+        return this.blackList.find( w => w.toString() === password ) === undefined ? true : false;
```
kgrgreer: Instead of going to the Array is is O(lgn), you scanning an array which is O(n).

src/foam/nanos/auth/PasswordPolicy.js
```diff
@@ -70,6 +82,10 @@ foam.CLASS({
           javaType: 'String'
         }
       ],
+      code: function(user, potentialPassword) {
+        let password = potentialPassword.toLowerCase();
```
kgrgreer: Better to set to lowerCase on the adapt methods for 'id' on CommonPassword so you only ever have to do it once.

src/foam/nanos/auth/PasswordPolicy.js
kgrgreer: Also, this will download the whole database to the client, which will probably only do one or two checks.

PR #4901, opened 2021-03-03
Fixed null ProxyX

tharmaman:
Throws NPEs while in debugger during tests.


jlhughes (approved):

kgrgreer:
I would rather find the root cause. We can look into this together tomorrow if you like.

PR #4904, opened 2021-03-03
add documentation for downloading document issue

xuerongNanopay:


src/foam/nanos/http/AuthWebAgent.java
```diff
@@ -73,6 +73,7 @@ public void execute(X x) {
     X requestX = session.getContext()
       .put(HttpServletRequest.class,  x.get(HttpServletRequest.class))
       .put(HttpServletResponse.class, x.get(HttpServletResponse.class))
+      // "lazy" the invoking of getWriter(). It prevents throwing exception from calling getOutputStream() later in the code.
```
jlhughes, line 76:
```suggestion
      // "lazy" the invoking of getWriter(). It prevents throwing exception from calling getOutputStream() later in the code.
      // The calling context provided a PrintWriter, but it needs to be explicitly passed on the requestX context, similar to HttpServletRequest/Response.
```

PR #4908, opened 2021-03-03
[DO NOT MERGE] add create label action

hchoura:


jlhughes (commented):
This is a common pattern of ours, but really, is this just allowing selection of the controller to use?  Can the boolean just be the controller to use?

kgrgreer:
Can you demo/explain this to me after next R&D standup?

PR #4915, opened 2021-03-04
Np 3764

xuerongNanopay:
ticket: https://nanopay.atlassian.net/browse/NP-3764
root cause: missing the case of null


jlhughes (approved):

src/foam/u2/view/PasswordView.js
```diff
@@ -132,7 +132,7 @@ foam.CLASS({
     function checkAvailability() {
       this.theme.passwordPolicy.validate(this.data)
         .then((pw) => {
-          if ( pw === undefined ) {
+          if ( ! pw ) {
```
kgrgreer, line 135: How about just:
  this.isAvailable = ! pw;

PR #4921, opened 2021-03-05
castObject mehod

smirnovaka:


src/foam/core/AbstractLongPropertyInfo.java
```diff
@@ -107,4 +107,13 @@ public boolean isDefaultValue(Object o) {
   public void format(foam.lib.formatter.FObjectFormatter formatter, foam.core.FObject obj) {
     formatter.output(get_(obj));
   }
+  @Override
+  public Object castObject(Object value) {
+    if ( value instanceof String ) {
+      return Long.valueOf((String) value);
+    } else if ( value instanceof Number ) {
```
kgrgreer: No need for 'else' since you've already returned in the 'if'.

src/foam/core/AbstractStringPropertyInfo.java
```diff
@@ -94,4 +94,11 @@ public boolean isDefaultValue(Object o) {
   public void format(foam.lib.formatter.FObjectFormatter formatter, foam.core.FObject obj) {
     formatter.output(get_(obj));
   }
+
+  public Object castObject(Object value) {
+    if ( value instanceof Number ) {
```
kgrgreer: Couldn't you just always return value.toString() (if it isn't null)? Since toString() on a String will just return itself. And then other types will work also.

PR #4928, opened 2021-03-05
[NP-3780] Fix attributes action for FLOW docs

KernelDeimos:


src/foam/flow/Document.js
```diff
@@ -438,7 +438,8 @@ foam.CLASS({
         },
 
         'attributes': function(v) {
-          return v.reduce(function(a, kv) {
+          console.log('v', v);
```
jlhughes:
```suggestion
```

PR #4930, opened 2021-03-05
[NP-3781] Fix clear all action on Performance page

KernelDeimos:


src/foam/nanos/pm/PMTableView.js
```diff
@@ -65,8 +65,8 @@ foam.CLASS({
     },
     {
       name: 'clearAll',
-      code: function() {
-        this.data.removeAll();
+      code: function(X) {
+        X.pmInfoDAO.removeAll();
```
kgrgreer, line 69: Is this different?

PR #4931, opened 2021-03-08
add if condition to check for array to fix ClassCastException

smirnovaka:


kgrgreer:
What is it when it isn't an array?

PR #4937, opened 2021-03-09
add rule engine missing action log

nanoMichal:


src/foam/nanos/ruler/RuleEngine.java
```diff
@@ -189,6 +189,10 @@ private boolean isRuleActive(Rule rule, RuleAction action) {
       isActive = ((LifecycleAware) rule).getLifecycleState() == LifecycleState.ACTIVE;
     }
 
+    if ( action == null ) {
+      Logger logger = (Logger) getX().get("logger");
+      logger.log("rule ID "+rule.getId()+" has a action set to null ");
```
kgrgreer, line 194: "an action"

lchanmann, line 194: `action` can be also be an async action.

nanoMichal: Yea, i see you use it to check both with same method.. meaning it will log on every rule except those with both set. 
my issue was i had an action filled out but there was an error in it which resulted in it not being loaded thus being null, and no logs as to why it wasn't working. Any idea for a better place to put some sort of notice or log for a null action?

PR #4940, opened 2021-03-10
Refactor agent auth service

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-3302

## Changes
- Refactor agentAuth.actAs() to return subject
- Move AgentJunctionStatus from nanopay to FOAM

### DIFF
--- a/src/foam/nanos/auth/AgentUserAuthService.java
+++ b/src/foam/nanos/auth/AgentUserAuthService.java
@@ -42,7 +42,7 @@ public class AgentUserAuthService
     sets "user" in the context to the passed in user. This allows users to
     act on behalf of others while retaining information on the user.
   */
-  public User actAs(X x, User entity) throws AuthenticationException {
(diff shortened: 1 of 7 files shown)


src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -51,7 +51,7 @@ public User actAs(X x, User entity) throws AuthenticationException {
       throw new AuthenticationException();
     }
```
zahnno, line 52: This condition block isn't necessary since subject realUser would always be available.

zahnno (commented):
Looks good one comment

src/foam/nanos/auth/AgentUserAuthService.java
lchanmann: agentAuth is not authenticated, still need this check.

PR #4942, opened 2021-03-10
[NP-3805] EnumView in AnyView

KernelDeimos:
Enum values displayed in AnyView are currently displayed with an FObjectView. Since changing a property within an enum value is an inconsistent way to update the property, AnyView should use EnumView for this case.


src/foam/dao/debug/putpath.js
```diff
@@ -0,0 +1,280 @@
+foam.CLASS({
```
kgrgreer: Please add copyright notice and documentation:

kgrgreer:
I don't see any mention of AnyView except in the title.

PR #4944, opened 2021-03-10
Add js fobject

hchoura:


src/files.js
```diff
@@ -97,6 +97,7 @@ FOAM_FILES([
   { name: "foam/i18n/Messages" },
   { name: "foam/core/Validation" },
   { name: "foam/core/Action" },
+  { name: "foam/u2/JSFObject" },
```
kgrgreer: I would put this in foam.core instead of foam.u2.

src/foam/u2/JSFObject.js
```diff
@@ -0,0 +1,25 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2',
+  name: 'JSFObject',
+  extends: 'foam.core.Property',
+
+  documentation: `
+    Add properties to an object allowing the server to handle it.
```
kgrgreer: Change to:
A Property which is only meaningful in JS. Is stored in Java as unresolved JSON.

PR #4948, opened 2021-03-11
History Record Improvements

zahnno:
- Added historyDAO config option to easyDAO
- Appended user reference ids to history record and made setting values on history records easier.

### DIFF
--- a/src/foam/nanos/auth/Subject.js
+++ b/src/foam/nanos/auth/Subject.js
@@ -98,7 +98,7 @@ foam.CLASS({
         User realUser = (User) getRealUser();
 
         return realUser.getId() != user.getId() ?
-            realUser.toSummary() + "(" + user.toSummary() + ")" :
+            user.toSummary() + "(" + realUser.toSummary() + ")" :
             user.toSummary();
       `
     }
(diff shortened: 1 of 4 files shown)


src/foam/nanos/auth/Subject.js
```diff
@@ -89,6 +89,15 @@ foam.CLASS({
       javaCode: `
         return getUser().getId() != getRealUser().getId();
       `
+    },
+    {
+      name: 'toSummary',
+      type: 'String',
+      javaCode: `
+        return isAgent() ?
```
jlhughes, line 97: Can this not just use toString above?  As it outputs the entire path.

zahnno: That's a good question. 

Would we want to have toSummary potentially reference a long list of user summaries? Example being "john doe >> jane doe >> john doe >> cookie doe co". This approach can extend a considerable amount I'd imagine and already has a place in the toString method.

I've always thought of toSummary on an object to provide a legible description that includes the objects top interest property values. In my uses of the subject I haven't cared much for the user path.

PR #4949, opened 2021-03-11
Np 3785/add mql

smirnovaka:


src/foam/mlang/MLang.java
```diff
@@ -249,6 +249,12 @@ public static Predicate DOT_F(Object o1, Object o2) {
     return dotF;
   }
 
+  public static Predicate MQL(String query) {
+    MQLExpr mql = new MQLExpr();
```
kgrgreer: Could we add and then use a constructor which takes query?

src/foam/mlang/mlang.js
```diff
@@ -3879,6 +3881,88 @@ foam.CLASS({
...
+    'foam.mlang.Constant',
+    'foam.parse.QueryParser'
+  ],
+
+  properties: [
+    {
+      class: 'Map',
+      name: 'specializations_',
+      factory: function() { return {}; },
+      javaFactory: 'return new java.util.HashMap<ClassInfo, foam.mlang.predicate.Predicate>();'
```
kgrgreer: Better to make a ConcurrentHashMap or synchronize elsewhere.

src/foam/nanos/ruler/predicate/MQL.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+  ],
+
+  methods: [
+    {
+      name: 'f',
+      javaCode: `
+        if ( getIsNew() ) {
+          FObject nu  = (FObject) NEW_OBJ.f(obj);
+          return MQL(getQuery()).f(nu);
```
kgrgreer: This is going to create/compile/specialize a new MQL for each call to f(), which is going to be very slow.

kgrgreer: Let's remove the MQL constructor and add a caching create() method, like many of the other parsers do. On the JS side we can make it a Multiton on the 'query' property. See docs in Multiton.js.

src/foam/mlang/mlang.js
```diff
@@ -3879,6 +3881,115 @@ foam.CLASS({
...
+        return (Predicate) getSpecializations_().get(model);
+      `
+    },
+    {
+      name: 'specialize',
+      args: [ { name: 'model', type: 'ClassInfo' } ],
+      type: 'Predicate',
+      code: function(model) {
+        var qp = foam.parse.QueryParser.create({of: model.id});
+        debugger;
```
kgrgreer: remove debugger

src/foam/nanos/ruler/predicate/MQL.js
```diff
@@ -0,0 +1,47 @@
+/**
```
kgrgreer: I would like to handle this using the DOT notation (which we didn't do yet). Can you remove this from the PR?

PR #4951, opened 2021-03-11
Updating CIDR Whitelisting

nanoscott:


src/foam/nanos/session/Session.js
```diff
@@ -220,6 +218,15 @@ List entries are of the form: 172.0.0.0/24 - this would restrict logins to the 1
           ((foam.nanos.logger.Logger) x.get("logger")).warning(this.getClass().getSimpleName(), "validateRemoteHost", remoteIp, e.getMessage());
         }
       }
+      
+      if ( cidrs.length > 0 ) return;
```
nanoscott: Question: @jlhughes .. is this right? Seems like we shouldn't return here since we already checked above if the IP was whitelisted as part of the CIDR whitelist, and if it didn't match we would've returned there, no? Why check the lenght again here?

jlhughes: ah, yes, the inRange, - in my head I had it reversed. so don't need length check.

jlhughes (approved):

PR #4958, opened 2021-03-12
[NP-3840] Refactor CronScheduler into modelled class

jlhughes:
Also
On start, delay 'execute' to allow clustering to startup.
And during execute, only check and update status if Medusa Primary. Crons only run on Primary, but each mediator was running this Scheduler which was updating statuses.


src/foam/nanos/cron/CronScheduler.js
```diff
@@ -0,0 +1,179 @@
...
+      name: 'execute',
+      javaCode: `
+    Logger logger = new PrefixLogger(new Object[] {
+        this.getClass().getSimpleName()
+      }, (Logger) x.get("logger"));
+    logger.info("execute");
+
+    try {
+      while ( true ) {
+        foam.nanos.medusa.ClusterConfigSupport support = (foam.nanos.medusa.ClusterConfigSupport) x.get("clusterConfigSupport");
```
kgrgreer, line 121: I'm not keen on tightly coupling CRON to MEDUSA, but can't think of a better way.

PR #4960, opened 2021-03-13
Authorizer Benchmarker

zahnno:
Generic Authorizer Benchmarker

Plug in the authorizer you'd like benchmark results on.
Specify the operation you'd like to benchmark and run it for results.
Added test context helper method in TestUtils.


src/foam/nanos/bench/benchmarks/AuthorizerBenchmark.java
```diff
@@ -0,0 +1,114 @@
...
+  If further setup or additional logic on each CRUD operation is required, extending this class and referencing the setup and CRUD method should suffice.
+ */
+
+public class AuthorizerBenchmark
+  implements Benchmark
+{
+  protected X userAuthorizedContext;
+  protected DAO dao = new MDAO(User.getOwnClassInfo());
+  protected Authorizer authorizer;
+  protected int recordAmount;
```
jlhughes, line 35: Believe this is the same as the BenchmarkRunner invocationCount. 
When BenchmarkRunner is configured for multiple threads, then it executes invocationCount per thread.

zahnno: I thought so. I made recordAmount so this benchmark is capable of benchmarking each thread with multiple records instead of one entry or a hard coded iteration per thread.

jlhughes (approved):

PR #4963, opened 2021-03-15
NP-3811 Strong Password UI is quirky

xuerongNanopay:
ticket: https://nanopay.atlassian.net/browse/NP-3811

### DIFF
--- a/src/foam/u2/view/PasswordView.js
+++ b/src/foam/u2/view/PasswordView.js
@@ -84,12 +84,16 @@ foam.CLASS({
     {
       class: 'Boolean',
       name: 'isAvailable',
+      value: true
     },
   ],
 
   methods: [
     function initE() {
       this.SUPER();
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/PasswordView.js
```diff
@@ -98,7 +102,14 @@ foam.CLASS({
...
         }, this.inputElement$)
           .addClass('full-width-input-password')
-          .on('blur', this.checkAvailability)
+          .on('keyup', () => {
+            clearTimeout(typingTimer);
+            typingTimer = setTimeout(this.checkAvailability, doneTypingInterval);
+          })
+          .on('keydown', () => {
+            clearTimeout(typingTimer);
+            self.isAvailable = true;
```
kgrgreer: Aren't self and this the same here, so you could get rid of 'self'?

PR #4967, opened 2021-03-16
Deprecate isAdmin

zahnno:
### Deprecated isAdmin

- Replaced with descriptive permission checks.
- Updated system auth to check for group implies on * to maintain existing behaviour.
- Added new permissions relating to checks mentioned earlier.


src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -161,36 +139,33 @@ foam.CLASS({
...
       }
-
       return super.put_(x, obj);
       `
     },
     {
       name: 'find_',
       javaCode: `
-      if ( isAdmin(x) ) {
+      if ( ((AuthService) x.get("auth")).check(x, "spid.obj.find.*") ) {
```
jlhughes: How about 
serviceprovider.read.*

src/permissions
```diff
@@ -34,5 +34,12 @@ p({"class":"foam.nanos.auth.Permission","id":"user.rw.onboarded","description":"
 p({"class":"foam.nanos.auth.Permission","id":"service.userCapabilityJunctionDAO","description":""})
 p({"class":"foam.nanos.auth.Permission","id":"ucj.addPermission","description":"Permission to add a UCJ, if not UCJ owner."})
 p({"class":"foam.nanos.auth.Permission","id":"usercapabilityjunction.read.*","description":"Permission to read usercapabilityjunction"})
+p({"class":"foam.nanos.auth.Permission","id":"usercapabilityjunction.create.*","description":"Permission to create usercapabilityjunction"})
```
jlhughes:
```suggestion
p({"class":"foam.nanos.auth.Permission","id":"usercapabilityjunction.create","description":"Permission to create usercapabilityjunction"})
```
to be consistent with create permissions.

src/foam/nanos/auth/ServiceProviderAwareDAO.js
```diff
@@ -115,7 +96,7 @@ foam.CLASS({
       }
       if ( SafetyUtil.isEmpty(spid) ||
            user != null &&
-           user.isAdmin() &&
+           auth.check(x, "default.to.theme.spid") &&
```
jlhughes:
```suggestion
           auth.check(x, "theme.read.*") &&
```
We don't have theme permissions, but if we had to add them in the future.

src/foam/nanos/crunch/ServerCrunchService.java
```diff
@@ -325,14 +326,15 @@ public UserCapabilityJunction updateJunction(
       ucj.setStatus(status);
     }
 
+    AuthService auth = (AuthService) x.get("auth");
     if (
-      subject.getRealUser().isAdmin()
+      auth.check(x, "warn.ucj.update")
```
jlhughes:
```suggestion
      auth.check(x, "usercapabilityjunction.warn.update")
```
like to keep the model first and thinking about other permissions like model.ro.property

src/foam/nanos/auth/ServiceProviderAwareDAO.js
zahnno: I thought of that also but it's already used for reading the serviceProviderDAO. Providing this method with its own permission decouples the authorization on the two functions.

src/foam/nanos/auth/ServiceProviderAwareDAO.js
zahnno: I don't think keeping this similar to our DOPs would benefit here since it's a condition placed on whether the user should default to the spid of the theme applied to context. It doesn't share much relevance to reading themes.

src/permissions
```diff
@@ -34,5 +34,12 @@ p({"class":"foam.nanos.auth.Permission","id":"user.rw.onboarded","description":"
...
 p({"class":"foam.nanos.auth.Permission","id":"menu.read.admin.googleApiCredentials","description":""})
-p({"class":"foam.nanos.auth.Permission","id":"data.import.googleSheets","description":""})
\ No newline at end of file
+p({"class":"foam.nanos.auth.Permission","id":"data.import.googleSheets","description":""})
+p({"class":"foam.nanos.auth.Permission","id":"default.to.theme.spid","description":"Permits user defaulting to theme spid when calling serviceAwareDAOs if no spid is associated to them."})
+p({"class":"foam.nanos.auth.Permission","id":"spid.object.find.*","description":"Permits finding all objects in DAO regardless of their spid"})
+p({"class":"foam.nanos.auth.Permission","id":"spid.object.remove.*","description":"Permits removing objects in a spid aware DAO regardless of their spid"})
+p({"class":"foam.nanos.auth.Permission","id":"spid.object.find.*","description":"Permits finding objects in spid aware DAO regardless of their spid"})
+p({"class":"foam.nanos.auth.Permission","id":"warn.ucj.update","description":"Log when a user with this permission updates a user capability junction."})
```
zahnno:
```suggestion
p({"class":"foam.nanos.auth.Permission","id":"usercapabilityjunction.warn.update","description":"Log when a user with this permission updates a user capability junction."})
```

jlhughes (approved):

PR #4968, opened 2021-03-17
Test and set getTime to zero if less than zero.  This occurs when tim…

jlhughes:
…ing between physical instances and clocks are slightly different.

### DIFF
--- a/src/foam/nanos/pm/PM.js
+++ b/src/foam/nanos/pm/PM.js
@@ -100,9 +100,7 @@ foam.CLASS({
       name: 'getTime',
       type: 'Long',
       javaCode: `
-    long t = getEndTime() - getStartTime();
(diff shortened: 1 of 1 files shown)


src/foam/nanos/pm/PM.js
```diff
@@ -100,7 +100,9 @@ foam.CLASS({
       name: 'getTime',
       type: 'Long',
       javaCode: `
-    return getEndTime() - getStartTime();
+    long t = getEndTime() - getStartTime();
```
kgrgreer: return Math.max(0, getEndTime() - getStartTime());

jlhughes: Funny, that's what I had originally, but was going for speed, will update.

PR #4973, opened 2021-03-17
Implement agent auth logout

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-3302

## Changes
- Implement agent auth logout

### DIFF
--- a/src/foam/nanos/auth/AgentUserAuthService.java
+++ b/src/foam/nanos/auth/AgentUserAuthService.java
@@ -122,4 +122,26 @@ public class AgentUserAuthService
       return false;
     }
   }
+
+  public void logout(X x) {
+    User agent = ((Subject) x.get("subject")).getRealUser();
+
+    // Check for current context user
+    if ( agent == null || agent.getLifecycleState() != LifecycleState.ACTIVE ) {
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


src/foam/nanos/auth/AgentUserAuthService.java
```diff
@@ -122,4 +122,26 @@ public boolean canActAs(X x, User agent, User entity) {
       return false;
     }
   }
+
+  public void logout(X x) {
+    User agent = ((Subject) x.get("subject")).getRealUser();
+
+    // Check for current context user
+    if ( agent == null || agent.getLifecycleState() != LifecycleState.ACTIVE ) {
```
jlhughes, line 130: Why the ACTIVE test? - how did they end up in the Subject? 
Assume this could occur if an admin deactivates them before they can logout. But any subsequent operation would test their status and deny them.

lchanmann: the ACTIVE test is just a defensive check not strictly necessary, we have a rule (LogoutUserAction) to logout the user when lifecycle is set to DELETED. Can remove the check if we don't want it, @jlhughes @kgrgreer?

PR #4979, opened 2021-03-18
Date - only adjust to noon when outside of foam.Date.MIN/MAX range

jlhughes:
Previously, all dates were being set to 'noon' time.

### DIFF
--- a/src/foam/core/types.js
+++ b/src/foam/core/types.js
@@ -157,7 +157,7 @@ foam.CLASS({
 
           d = ret;
         }
-        if ( d == foam.Date.MAX_DATE || d == foam.Date.MIN_DATE ) return d;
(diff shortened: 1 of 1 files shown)


src/foam/core/types.js
```diff
@@ -157,7 +157,7 @@ foam.CLASS({
 
           d = ret;
         }
-        if ( d == foam.Date.MAX_DATE || d == foam.Date.MIN_DATE ) return d;
+        if ( d <= foam.Date.MAX_DATE && d >= foam.Date.MIN_DATE ) return d;
```
kgrgreer, line 160: This is always true. I've fixed this in another PR.

kgrgreer:
This is always true. I've fixed this in another PR.

PR #4993, opened 2021-03-19
Return Element instead of slot

tharmaman:


src/foam/nanos/crunch/ui/MinMaxCapabilityWizardletSection.js
```diff
@@ -50,7 +50,7 @@ foam.CLASS({
           return vs;
         }
         return null;
-      })
+      }))
```
kgrgreer: Missing ;

tharmaman: added

PR #4994, opened 2021-03-19
[NP-3879] Update copyFrom

KernelDeimos:


src/foam/core/FObject.js
```diff
@@ -985,7 +985,13 @@ foam.CLASS({
           // for each object since they are of the exact same
           // type.
           if ( o.hasOwnProperty(name) || props[i].factory ) {
-            this[name] = o[name];
+            let handledByPropertyClass = false;
```
kgrgreer: This is an odd control-structure. Why not just:

if ( ! props[i].copyValueFrom || ! iprops[i].copyValueFrom(this, o) ) this[name] = o[name];

KernelDeimos: Oh neat. Usually I avoid that in case someone assumes the function has no side effect, but in this case it makes it look so much cleaner. Thanks! (fixed)

src/foam/core/types.js
```diff
@@ -741,6 +741,15 @@ foam.CLASS({
       value: function(value, cloneMap, opt_X) {
         cloneMap[this.name] = value && value.clone ? value.clone(opt_X) : value;
       }
+    },
+    {
+      name: 'copyTransient',
+      class: 'Boolean',
+      value: false,
```
kgrgreer: When/why would we want to make this configurable?

KernelDeimos: Originally I thought it was needed because the client build stopped working, but then making the change to put this here fixed the issue by itself. I left it here in case it turns out to cause a problem somewhere else (since it's a regression if any code expects the object to hold a reference to an FObject in a property).

Should I remove this?

PR #4998, opened 2021-03-22
Np 3806/updates locals

Nauna:
- locale changes are business requirement.
- uniformed the email label from e-mail to Email consistently
- simple white space makes crunch service function more legible
- removed disable save from general views - since ucj's can be saved with invalid data ... and don't believe this should be blocking any data. Back office have higher permissions so maybe able to bypass invalid fields in model. Letting the server validation handle whether this object is saved or not seems sufficient. Also brought this up in slack - no one objected.


src/foam/comics/v2/DAOUpdateView.js
```diff
@@ -111,9 +111,6 @@ foam.CLASS({
   actions: [
     {
       name: 'save',
-      isEnabled: function(workingData$errors_) {
```
kgrgreer: This is okay for UCJ's but not for most other types of data. We could either make this configurable or create a sub-class of DAOUpdateView for UCJs. Or find some other solution. But I don't think we should change global behaviour to accommodate UCJs.

jlhughes (approved):

PR #5012, opened 2021-03-23
Add lastModifiedByAgent

lchanmann:
## Ref
- https://nanopay.atlassian.net/browse/NP-3901

## Changes
- Add lastModifiedByAgent


jlhughes (approved):

zahnno:
We should serialize subject and replace createdBy lastmodified and this with the acting subject.

jlhughes:
@zahnno, yes that is a good idea.

PR #5013, opened 2021-03-23
do not check whether prerequisites imply a permission in the implies method

noodlemoodle:
Remove the check on whether prerequisites imply a permission in the implies method of Capability
This is redundant since a capability is only granted when its prerequisites are also granted

### DIFF
--- a/src/foam/nanos/crunch/Capability.js
+++ b/src/foam/nanos/crunch/Capability.js
@@ -330,17 +330,7 @@ foam.CLASS({
         for ( String permissionName : permissionsGranted ) {
           if ( this.stringImplies(permissionName, permission) ) return true;
         }
-
-        CrunchService crunchService = (CrunchService) x.get("crunchService");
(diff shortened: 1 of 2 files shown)


KernelDeimos (approved):

tharmaman (approved):

zahnno:
This changes the functionality of capability implies by quite a bit, considering it removes checking the capabilities prerequisites granted permissions as well.

How about adding another method to capability with the old implies method? This would be useful since we can retain the ability to check if a capability or it's prerequisites imply a permission.

noodlemoodle:
> This changes the functionality of capability implies by quite a bit, considering it removes checking the capabilities prerequisites granted permissions as well.
> 
> How about adding another method to capability with the old implies method? This would be useful since we can retain the ability to check if a capability or it's prerequisites imply a permission.

addressed in https://github.com/foam-framework/foam2/pull/5036 , tagged dnm for now since need update after https://github.com/foam-framework/foam2/pull/5034 merged

PR #5015, opened 2021-03-23
add translate to action and add en for aproval request

yij793:
address: https://nanopay.atlassian.net/browse/NP-3897
Screen Shot 2021-03-23 at 11 21 54 AM


src/foam/nanos/dao/Operation.js
```diff
@@ -21,28 +21,28 @@ foam.ENUM({
   values: [
     {
       name: 'CREATE',
-      label: 'Create',
+      label: { en: 'Create', pt: 'Crio'},
```
kgrgreer, line 24: If we include these here then everyone has to download the Portuguese text, even those that speak other languages. This makes the download larger for everyone and doesn't scale as we add support for more and more languages. Can't these just be defined in the locales journal?

yij793: @kgrgreer should able to, will make the changes

PR #5026, opened 2021-03-24
letting the user finish input before field adjust - so that decimal i…

Nauna:
…s read

### DIFF
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -2525,7 +2525,7 @@ foam.CLASS({
   requires: [ 'foam.u2.view.CurrencyView' ],
   properties: [
     [ 'displayWidth', 15 ],
-    [ 'view', { class: 'foam.u2.view.CurrencyView' } ]
+    [ 'view', { class: 'foam.u2.view.CurrencyView', onKey: false } ]
   ]
 });
 
--- a/src/foam/u2/Element.js
+++ b/src/foam/u2/Element.js
@@ -2525,7 +2525,7 @@ foam.CLASS({
   requires: [ 'foam.u2.view.CurrencyView' ],
   properties: [
     [ 'displayWidth', 15 ],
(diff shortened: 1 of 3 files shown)


src/foam/u2/Element.js
```diff
@@ -2525,7 +2525,7 @@ foam.CLASS({
   requires: [ 'foam.u2.view.CurrencyView' ],
   properties: [
     [ 'displayWidth', 15 ],
-    [ 'view', { class: 'foam.u2.view.CurrencyView' } ]
+    [ 'view', { class: 'foam.u2.view.CurrencyView', onKey: false } ]
```
kgrgreer: I would just change it in CurrencyView itself.

src/foam/u2/tag/Input.js
```diff
@@ -46,8 +46,9 @@ foam.CLASS({
     {
       class: 'Boolean',
       name: 'onKey',
+      documentation: 'When true, $$DOC{ref:".data"} is updated on every keystroke, rather than on blur.',
       attribute: true,
-      // documentation: 'When true, $$DOC{ref:".data"} is updated on every keystroke, rather than on blur.'
+      value: true
```
kgrgreer: Why is this being set to true?

src/foam/u2/tag/Input.js
```diff
@@ -137,8 +137,6 @@ foam.CLASS({
 
     function fromProperty(p) {
       this.SUPER(p);
-
-      if ( ! this.hasOwnProperty('onKey') && p.validateObj ) this.onKey = true;
```
kgrgreer: I don't think it is safe to remove this.

Nauna: We can't override the property onKey if we don't. The only thing that will work, is changing the call like for UnitValue in Elements.js

...
[ 'view', { class: 'foam.u2.view.CurrencyView' , onKey: false } ]

kgrgreer:
Let's discuss later.

PR #5027, opened 2021-03-24
Make MultiChoiceView use foam.util.equals

tharmaman:
For cases where data and choice have elements which contain the same value, but are different objects

### DIFF
--- a/src/foam/u2/view/MultiChoiceView.js
+++ b/src/foam/u2/view/MultiChoiceView.js
@@ -168,28 +168,17 @@ foam.CLASS({
     },
 
     function isChoiceSelected(data, choice){
-      var isSelected = false
-
       for ( var i = 0 ; i < data.length ; i++ ) {
-        isSelected = foam.util.equals(data[i], choice);
-
-        if ( isSelected ) break;
+        if ( foam.util.equals(data[i], choice) ) return true;
       }
-      
-      return isSelected;
+      return false;
     },
 
-    function getIndexOfChoice(data, choice){
-      var indexDataToAdd = -1;
-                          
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/MultiChoiceView.js
```diff
@@ -167,20 +167,45 @@ foam.CLASS({
       return this.dao.where(this.IN(of.ID, this.data));
     },
 
+    function isChoiceSelected(data, choice){
```
kgrgreer, line 170: This seems easier to understand:

function isChoiceSelected(data, choice) {
      for ( var i = 0 ; i < data.length ; i++ )
        if ( foam.util.equals(data[i], choice) ) return true;

      return false;
    },

src/foam/u2/view/MultiChoiceView.js
```diff
@@ -167,20 +167,45 @@ foam.CLASS({
...
+      
+      return isSelected;
+    },
+
+    function getIndexOfChoice(data, choice){
+      var indexDataToAdd = -1;
+                          
+      for ( var i = 0 ; i < data.length ; i++ ) {
+        if ( foam.util.equals(data[i], choice) ){
+          indexDataToAdd  = i;
```
kgrgreer: Again, why not just get rid of the variable and return i here directly?

tharmaman: Yeah my bad the code here previously wasn't in a method and was in-line so the direct returns make sense instead of breaking.

src/foam/u2/view/MultiChoiceView.js
tharmaman: RE: comment below

PR #5034, opened 2021-03-24
Capability StringImplies -> AuthPermission implies

zahnno:
While reviewing some code and going over the capabilityAuthService these suggestions came about. 

Capability @stringImplies and AuthPermission @implies do the same thing.

@stringImplies was our own custom method that didn't need to exist since AuthPermission @implies serves the same purpose.

Also cleaned up some odd values/string assignments in the capabilityAuthService from a ways back that aren't used anymore.


noodlemoodle (approved):
LGTM

PR #4950, opened 2021-03-11
Predicated prerequisitecapabilityjunction

noodlemoodle:
- add predicate fo capabilitycapabilityjunction
- add predicatedprerequisitecapabilityjunctiondao that filters the dao by predicate for non-admin users
- cache the pcjs map from crunchservice in the session and only use it when context user/realuser matches session user/agent

- updates on nanopay side : https://github.com/nanoPayinc/NANOPAY/pull/12122
- merge capabilities for signingofficer (US, CA, BR) : https://github.com/nanoPayinc/NANOPAY/pull/12123

PR #5036, opened 2021-03-25
DO_NOT_MERGE Implies refactor

noodlemoodle:
rename current implies method to grantPermission and add implies method which includes functionality of old implies method to check if prerequisites implies the permission
address https://github.com/foam-framework/foam2/pull/5013#issuecomment-806012385


noodlemoodle:
added to https://github.com/foam-framework/foam2/pull/4950

PR #5040, opened 2021-03-25
Replaces colors to match new designs and added a new badge view

sarthak-marwaha:
Screen Shot 2021-03-25 at 11 27 49 AM
Screen Shot 2021-03-25 at 11 29 24 AM

- Added a Badge View with support for badges attached to card
- Changed corner rounding to align with new designs 
- replaced colors in Treviso theme to match new designs

### DIFF
--- a/src/foam/nanos/crunch/CapabilityJunctionStatus.js
+++ b/src/foam/nanos/crunch/CapabilityJunctionStatus.js
@@ -11,52 +11,52 @@ foam.ENUM({
       name: 'PENDING',
       label: { en: 'pending', pt: 'pendente'},
       documentation: 'Our team is currently reviewing. Approvals may take up to 24 hours',
-      background: '#E7EAEC',
-      color: '#1E1F21'
+      background: '#FFFFFF',
+      color: '#865300'
     },
     {
       name: 'GRANTED',
       label: { en: 'granted', pt: 'concedida'},
       documentation: 'The information you provided has been approved',
-      background: '#E2F2DD',
-      color: '#19402E'
+      background: '#FFFFFF',
+      color: '#007328'
     },
     {
       name: 'EXPIRED',
       label: { en: 'expired', pt: 'expirada'},
       documentation: 'The information required has changed, or your inputs are no longer valid',
-      background: '#FBE88F',
-      color: '#816819'
+      background: '#A00B0B',
+      color: '#FFFFFF'
     },
     {
       name: 'ACTION_REQUIRED',
       label: { en: 'action required', pt: 'ação requerida'},
       documentation: 'Information is missing for required fields',
-      background: '#FBE88F',
-      color: '#816819'
+      background: '#D57D11',
+      color: '#FFFFFF'
     },
     {
       name: 'AVAILABLE',
       label: { en: 'available', pt: 'acessível'},
       documentation: 'You are ready to get started',
-      background: '#A7BEFF',
-      color: '#202341'
+      background: '#00B03E',
+      color: '#FFFFFF'
     },
     {
       name: 'APPROVED',
       label: { en: 'approved', pt: 'aprovada'},
       documentation: `- not seen by users - Denoting a UCJ requiring review has been approved. It would need to go through the rules of the ucjDAO before
       being set to granted.`,
-      background: '#E2F2DD',
-      color: '#19402E',
+      background: '#FFFFFF',
+      color: '#007328',
       ordinal: 6
     },
     {
       name: 'PENDING_REVIEW',
       label: { en: 'pending review', pt: 'revisão pendente' },
       documentation: 'The information you provided is pending review',
-      background: '#FBE88F',
-      color: '#816819'
+      background: '#FFFFFF',
+      color: '/*%WARNING1%*/ #865300'
     },
     {
       name: 'REJECTED',
@@ -64,8 +64,8 @@ foam.ENUM({
       documentation: `- not seen by users - Denoting a junction requiring review has been rejected. Meant to mark items in a FINAL rejected state where it is not 
       expected to go to EXPIRED and have the user fill out more info. Used in the 
       Capable object junctions.`,
-      background: '#F79393',
-      color: '#631414'
+      background: '#FFFFFF',
+      color: '/*%DESTRUCTIVE2%*/ #A61414'
     },
   ]
 });
--- a/src/foam/u2/crunch/CapabilityFeatureView.js
+++ b/src/foam/u2/crunch/CapabilityFeatureView.js
@@ -28,7 +28,7 @@ foam.CLASS({
     'foam.nanos.crunch.UserCapabilityJunction',
     'foam.u2.crunch.Style',
     'foam.u2.Tooltip',
-    'foam.u2.view.BadgeView',
+    'foam.u2.view.EnumBadgeView',
     'foam.u2.view.ReadOnlyEnumView'
   ],
 
@@ -120,11 +120,7 @@ foam.CLASS({
               .end()
               .start()
                 .addClass(this.myClass('badge'))
-                .add(foam.u2.view.BadgeView.create({
-                  data: cjStatus.label,
-                  color: cjStatus.color,
-                  backgroundColor: cjStatus.background
-                }))
+                .add(foam.u2.view.EnumBadgeView.create({ data: cjStatus }))
               .end();
           }))
           .add(this.slot(function(isRenewable) {
--- a/src/foam/u2/crunch/CapabilityFeatureView.js
+++ b/src/foam/u2/crunch/CapabilityFeatureView.js
@@ -41,10 +41,8 @@ foam.CLASS({
       position: relative;
     }
 
-    ^ .foam-u2-crunch-Style-badge {
-      position: absolute;
-      bottom: 12px;
-      right: 8px;
+    ^badge > * {
+      border-radius: 0px 11.2px 11.2px 0px
     }
 
     ^ .foam-u2-crunch-Style-renewable-description {
@@ -120,14 +118,14 @@ foam.CLASS({
                 .enableClass(style.myClass('tooltipDisabled'), self.tooltipEnabled, true)
                 .add(cjStatus.documentation)
               .end()
-              .start(foam.u2.view.BadgeView, {
-                label: cjStatus.label,
-                color: cjStatus.color,
-                backgroundColor: cjStatus.background,
-                isAttachedToCard: true
-              })
-                .addClass(self.myClass('badge'))
-              .end()
+              .start()
+                .addClass(this.myClass('badge'))
+                .add(foam.u2.view.BadgeView.create({
+                  label: cjStatus.label,
+                  color: cjStatus.color,
+                  backgroundColor: cjStatus.background
+                }))
+              .end();
           }))
           .add(this.slot(function(isRenewable) {
             return isRenewable ? this.E()
--- a/src/foam/u2/crunch/CapabilityFeatureView.js
+++ b/src/foam/u2/crunch/CapabilityFeatureView.js
@@ -27,7 +27,9 @@ foam.CLASS({
     'foam.nanos.crunch.CapabilityJunctionStatus',
     'foam.nanos.crunch.UserCapabilityJunction',
     'foam.u2.view.ReadOnlyEnumView',
-    'foam.u2.crunch.Style'
+    'foam.u2.crunch.Style',
+    'foam.u2.Tooltip',
+    'foam.u2.view.BadgeView'
   ],
 
   documentation: `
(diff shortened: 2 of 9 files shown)


src/foam/nanos/crunch/CapabilityJunctionStatus.js
```diff
@@ -11,61 +11,61 @@ foam.ENUM({
       name: 'PENDING',
       label: { en: 'pending', pt: 'pendente'},
       documentation: 'Our team is currently reviewing. Approvals may take up to 24 hours',
-      background: '#E7EAEC',
-      color: '#1E1F21'
+      background: '#FFFFFF',
```
Nauna, line 14: Is there anyway to have these use the theme colours? Like what was used in the ApprovalStatus. ITs. just everything is going to be configurable but these ... ok if we are ok with everything in app changing BUT the ucj.status.
Just being difficult - and commenting - as a food for thought.

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -110,18 +112,21 @@ foam.CLASS({
...
                 .addClass(style.myClass('tooltip-bottom'))
                 .enableClass(style.myClass('tooltipDisabled'), self.tooltipEnabled, true)
                 .add(cjStatus.documentation)
               .end()
-              .add(cjStatus.label).addClass(style.myClass('badge'))
-              .style({
-                'background-color': cjStatus.background,
-                'color': cjStatus.color
-              });
+              .start(foam.u2.view.BadgeView, {label: cjStatus.label, color: cjStatus.color, backgroundColor: cjStatus.background, isAttachedToCard: true}).addClass(self.myClass('badge')).end()
```
Nauna: If you are using VS code as your IDE you can turn on ESLint, which we have configured to yell at us for long lines. Our standard is generally 80char per line

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -110,18 +112,21 @@ foam.CLASS({
...
+              // .add(cjStatus.label).addClass(style.myClass('badge'))
+              // .style({
```
Nauna: If this is not going to be used any more, we should be deleting it. There will always be a history of it in github if you ever want to come back.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+  extends: 'foam.u2.View',
+  
+  css:`
+    ^Wrapper{
+      height: 24px;
+     
+      width: 79px;
+
+      padding: 0 8px;
+
```
Nauna: You are taking over the css ... but was all this space intentional. I am under the impression that uniformity is easier to read.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+    },
+    {
+        class: 'String',
+        name: 'color',
+    },
+  ],
+
+  methods: [
+    function initE(){
+        this
```
Nauna, line 67: 1) the indenting here, isn't quite what we traditionally do. For example the subsequent html tags are usually grouped together via indenting. So 
.start()
   .add()
.end()

2) you are the css guy - but I believe setting this in its own div would be a good idea. DIV is set by using .start('div') which is what we use in foam. NOTE .start() == .start('div')... div is the default.

src/foam/u2/crunch/CapabilityFeatureView.js
kgrgreer: What is says is: "Line lengths should be 80 characters or less, except for embedded data, like templates or sprites, or when modifying the code to make it fit in less than 80 characters would actually makes it less readable."
Source: https://github.com/foam-framework/foam/wiki/CodingStyle

In this case, you could definitely indent between start() and end() and improve readability.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
+/**
```
kgrgreer: Wrong copyright header.
Use:
/**
 * @license
 * Copyright 2021 The FOAM Authors. All Rights Reserved.
 */

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+
+      padding: 0 8px;
+ 
+
+      font-family: /*%FONT1%*/ Roboto, 'Helvetica Neue', Helvetica, Arial, sans-serif;
+      font-size: 10px;
+      font-weight: 500;
+      font-style: normal;
+      font-stretch: normal;
+      line-height: 24px;
```
kgrgreer: Sort by css names.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+        class: 'String',
+        name: 'label',
+    },
+    {
+        class: 'String',
+        name: 'backgroundColor',
+    },
+    {
+        class: 'String',
+        name: 'color',
```
kgrgreer: No trailing commas.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+    },
```
kgrgreer: same

src/foam/u2/view/BadgeView.js
kgrgreer: indentation is 2 spaces

src/foam/nanos/crunch/CapabilityJunctionStatus.js
sarthak-marwaha: These Status indicators don't use the theme colours so we kind of have to hard code them.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+            function(){this.style({'border-radius': '11.2px'})}
+          )
+          .style({
+            'background-color': this.backgroundColor,
+            'color': this.color
+          })
+          .start()
+          .addClass(this.myClass('label'))
+          .add(this.label)
+          .end()
```
kgrgreer: missing ;

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+          .callIfElse(this.isAttachedToCard,
+            function(){this.style({'border-radius': '0px 11.2px 11.2px 0px'})},
+            function(){this.style({'border-radius': '11.2px'})}
+          )
+          .style({
+            'background-color': this.backgroundColor,
+            'color': this.color
+          })
+          .start()
+          .addClass(this.myClass('label'))
```
kgrgreer: indent between start and end

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+          .add(this.label)
+          .end()
+          
+    }
+  ],
```
kgrgreer: no trailing commas

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,85 @@
...
+
+})
```
kgrgreer: missing ;

src/foam/u2/crunch/CapabilityFeatureView.js
sarthak-marwaha: Missed this somehow, will delete

src/foam/u2/view/BadgeView.js
sarthak-marwaha: Nope, Will remove

src/foam/nanos/crunch/CapabilityJunctionStatus.js
sarthak-marwaha: Could be something we could discuss with the design team regarding colour consistency. I believe it is done like this to ensure legibility over the different background colour. For now I'll leave it at this and we can come back to it.

kgrgreer: Then open a new ticket so we don't forget.

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -110,18 +112,22 @@ foam.CLASS({
...
-              .add(cjStatus.label).addClass(style.myClass('badge'))
-              .style({
-                'background-color': cjStatus.background,
-                'color': cjStatus.color
-              });
+              .start(foam.u2.view.BadgeView, {
+                label: cjStatus.label,
+                color: cjStatus.color,
+                backgroundColor: cjStatus.background,
+                isAttachedToCard: true
```
kgrgreer: I would like to try and remove this attribute and then adjust CSS from the CapabilityFeatureView.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,72 @@
...
+    {
+        class: 'String',
+        name: 'color'
+    }
+  ],
+
+  methods: [
+    function initE() {
+      this
+        .addClass(this.myClass('Wrapper'))
```
kgrgreer: Since this is top-level and for the whole view, you could just use this.myClass(), and the the CSS selector above would just be '^'.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,72 @@
...
+        .callIfElse(this.isAttachedToCard,
+          function() {
+            this.style({ 'border-radius': '0px 11.2px 11.2px 0px' });
```
kgrgreer: I don't think we should have the isAttachedToCard feature, but if we kept it, it would be easier to do this as:
this.style({'border-radius': this.isAttachedToCard ? '...' : '...'})

kgrgreer: And then it could be included into the .style() call below.

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -27,7 +27,9 @@ foam.CLASS({
     'foam.nanos.crunch.CapabilityJunctionStatus',
     'foam.nanos.crunch.UserCapabilityJunction',
     'foam.u2.view.ReadOnlyEnumView',
-    'foam.u2.crunch.Style'
+    'foam.u2.crunch.Style',
```
kgrgreer, line 29: sort requires

src/foam/u2/crunch/CapabilityFeatureView.js
```diff
@@ -110,18 +110,22 @@ foam.CLASS({
             'background-image': "url('" + self.data.icon + "')"
           })
           .add(this.slot(function(cjStatus) {
-            return this.E().addClass(style.myClass('tooltip'))
+            return this.E()
+            .addClass(style.myClass('tooltip'))
```
kgrgreer: indent this line two spaces

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,61 @@
...
+      line-height: 24px;
+      letter-spacing: normal;
+      padding: 0 8px;
+      text-align: center;
+      width: 79px;
+    }
+   
+  `,
+
+  documentation: `Creates badges with rounded/squared sides based on display context`,
```
kgrgreer: Use ' instead of `.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,61 @@
...
+
+  properties: [
+    {
+        class: 'String',
+        name: 'label'
+    },
+    {
+        class: 'String',
```
kgrgreer: class: 'Color',

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,61 @@
...
+        name: 'backgroundColor'
+    },
+    {
+        class: 'String',
```
kgrgreer: class: 'Color',

src/foam/u2/view/BadgeView.js
kgrgreer: Still not entirely sorted. letter comes before line.

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,61 @@
...
+  methods: [
+    function initE() {
+      this
+        .addClass(this.myClass())
+        .style({
+          'background-color': this.backgroundColor,
+          'color': this.color
+        })
+        .start()
+          .addClass(this.myClass('label'))
```
kgrgreer: Is this class used, or just for future use? Is this whole element (start()/end()) needed?

src/foam/u2/view/BadgeView.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+   
+  `,
+
+  documentation: `Creates badges with rounded/squared sides based on display context`,
+
+  properties: [
+    {
+        class: 'String',
+        name: 'label'
```
kgrgreer: Can we just use 'data' instead of 'label', to be consistent with all other Views? In which case, you can just remove this property since 'View' already has a data property.

PR #5050, opened 2021-03-26
createView on config of DAOMenu2/its Controllers

zahnno:
Allows to configure the createView for DAOMenu2 instead of the SectionedDetailView being hard coded. Still limited since it's a viewSpec and can't really manage the create view data very well, but is better than having the DAOMenu2 restricted to the SectionedDetailView create view.


jlhughes (approved):

zahnno:
@kgrgreer @jlhughes thoughts on defaulting to the AltView?

PR #5055, opened 2021-03-26
correct badgeView to EnumBadgeView

minsunny:
original here https://github.com/foam-framework/foam2/pull/5040/files


sarthak-marwaha:
Looks good

PR #5057, opened 2021-03-26
[Np 3713] Browser versioning

KolombetM:
https://nanopay.atlassian.net/browse/NP-3713
Add a border instead of message for loginView


src/foam/u2/borders/BrowserSupportBorder.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * @license
+ * Copyright 2020 The FOAM Authors. All Rights Reserved.
```
kgrgreer: 2021

src/foam/u2/borders/BrowserSupportBorder.js
```diff
@@ -0,0 +1,58 @@
...
+  documentation: 'Border to show browser support message',
+
+  messages: [
+    { name: 'BROWSER_SUPPORT', message: 'This application is optimized for '}
+  ],
+
+  methods: [
+    function init() {
+      if ( this.isChrome() ) {
+        this
```
kgrgreer, line 37: This part is the same for both cases, so only the part that's different needs to be in the if.

KolombetM: How do I break them without appending to context?
e.g. The part inside if will append to the context, but not to the end of the page.
```
function init() {
  this
    .start()
      .addClass(this.myClass())
      .start('div', null, this.content$)
        .addClass(this.myClass('content'))
      .end()
    .end();
  if ( ! this.isChrome() ) {
    this
      .start('div')
        .addClass(this.myClass("browser-message"))
        .add(this.BROWSER_SUPPORT)
        .start('a')
          .attrs({href: 'https://www.google.com/chrome/'})
          .add("Google Chrome")
        .end()
      .end();
  }
},
```

PR #5062, opened 2021-03-29
context agent implement removed

zahnno:
The implemented interface 'ContextAgent' removed here was causing build issues and as far as I'm aware doesn't need to be implemented


jlhughes:
But this is the interface for 'execute'. 
is the issue that it's missing the java implementation?

KernelDeimos:
@zahnno what kind of build issues? It would be nice to know why implementing this interface is causing an issue rather than just remove it.

zahnno:
The class was implementing an interface with an abstract method that wasn't being overridden.

If the ContextAgent's interface method was a default java method, it would have been fine, but ContextAgent execute is abstract and requires an override on classes implementing it.

PR #5070, opened 2021-03-29
adding some front end partial eval functionality for

olhabn:
In

### DIFF
--- a/src/foam/mlang/mlang.js
+++ b/src/foam/mlang/mlang.js
@@ -1704,7 +1704,7 @@ return false
         var value = this.arg2;
         if ( this.Constant.isInstance(this.arg2) ) value = this.arg2.value;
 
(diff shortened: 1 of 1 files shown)


src/foam/mlang/mlang.js
```diff
@@ -1701,10 +1701,18 @@ return false
     {
       name: 'partialEval',
       code: function partialEval() {
-        if ( ! this.Constant.isInstance(this.arg2) ) return this;
+        var value = this.arg2;
+        if ( this.Constant.isInstance(this.arg2) ) value = this.arg2.value;
 
-        return ( ! this.arg2.value ) || this.arg2.value.length === 0 ?
-            this.FALSE : this;
+        if ( ! value || value === 0 )
```
kgrgreer: I think this is the same as just ! value

PR #5072, opened 2021-03-29
adding isAvailable to export button to disable button if there is no export drivers available

olhabn:
nanopay: https://github.com/nanoPayinc/NANOPAY/pull/12252

### DIFF
--- a/src/foam/comics/v2/DAOBrowserView.js
+++ b/src/foam/comics/v2/DAOBrowserView.js
@@ -223,6 +223,10 @@ foam.CLASS({
       label: '',
       toolTip: 'Export Table Data',
       icon: 'images/export-arrow-icon.svg',
+      isAvailable: async function() {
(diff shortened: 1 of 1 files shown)


jlhughes (approved):

nanoscott (approved):

src/foam/comics/v2/DAOBrowserView.js
```diff
@@ -223,6 +223,10 @@ foam.CLASS({
       label: '',
       toolTip: 'Export Table Data',
       icon: 'images/export-arrow-icon.svg',
+      isAvailable: async function() {
+        var records = await this.__subContext__.exportDriverRegistryDAO.select();
```
kgrgreer: Why using __subContext__ instead of __context__, or better, import: 'exportDriverRegistryDAO', and then do this.exportDriverRegistryDAO?

olhabn: fixed

PR #5073, opened 2021-03-29
[NP-3970] Add Block border

KernelDeimos:


src/foam/u2/borders/Block.js
```diff
@@ -0,0 +1,18 @@
+foam.CLASS({
```
jlhughes, line 7: header.

src/foam/u2/borders/Block.js
```diff
@@ -0,0 +1,24 @@
...
+ * @license
+ * Copyright 2021 The FOAM Authors. All Rights Reserved.
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+foam.CLASS({
+  package: 'foam.u2.borders',
+  name: 'Block',
+  extends: 'foam.u2.Element',
+
```
kgrgreer, line 11: Please add documentation:

PR #5087, opened 2021-03-30
Move default language to theme

xuerongNanopay:


Nauna:
addresses: https://nanopay.atlassian.net/browse/NP-3687

PR #5090, opened 2021-03-31
add prop to configure rowview in treeview

noodlemoodle:


src/foam/u2/view/TreeView.js
```diff
@@ -391,6 +391,11 @@ foam.CLASS({
     {
       class: 'Function',
       name: 'onClickAddOn'
+    },
+    {
+      class: 'String',
```
kgrgreer: Better to make this class: 'ViewSpec'

PR #5091, opened 2021-03-31
[NP-3959] Add file preview to detail view

KolombetM:
https://nanopay.atlassian.net/browse/NP-3959


src/foam/nanos/fs/File.js
```diff
@@ -64,9 +64,30 @@ foam.CLASS({
...
       documentation: 'File converted to base64 string'
     },
+    {
+      class: 'String',
+      name: 'image',
+      createVisibility: 'HIDDEN',
+      updateVisibility: 'RO',
+      readVisibility: 'RO',
+      transient: true,
+      storageTransient: true,
```
jlhughes: transient is both networkTransient and storageTransient, so only one of these are needed.

jlhughes (approved):

PR #5097, opened 2021-03-31
Fix crash after making bad choices

KernelDeimos:


jlhughes (approved):
Good for now, thanks.

PR #5102, opened 2021-04-01
add error DateView2 css to SectionedDetailPropertyView

yij793:
address: https://nanopay.atlassian.net/browse/NP-4007
Screen Shot 2021-04-01 at 12 49 08 PM


jlhughes (commented):
Where/how is this used/needed?

kgrgreer:
Instead of doing this, can you remove all uses of DateView2? They are no longer needed. Thx

jlhughes (approved):

PR #5104, opened 2021-04-01
[NP-3972] Placeholder on BusinessOwnershipData.js not working

KolombetM:
https://nanopay.atlassian.net/browse/NP-3972

### DIFF
--- a/src/foam/u2/view/RichChoiceView.js
+++ b/src/foam/u2/view/RichChoiceView.js
@@ -720,7 +720,7 @@ foam.CLASS({
           });
 
           var summary = this.fullObject$.map(o => {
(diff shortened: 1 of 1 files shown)


src/foam/u2/view/RichChoiceView.js
```diff
@@ -720,7 +720,7 @@ foam.CLASS({
           });
 
           var summary = this.fullObject$.map(o => {
-            return o ? o.toSummary() : this.defaultSelectionPrompt;
+            return ( o && o.toSummary() ) ? o.toSummary() : this.defaultSelectionPrompt;
```
kgrgreer: Better to not compute toSummary twice. So:

return (o && o.toSummary()) || this.defaultSelection;

